PR #9, opened 2015-12-23
use python code blocks in readme

mrocklin:


alimanfoo:
Thank you!

alimanfoo:
For the record, the Python syntax highlighting was doing strange things for some value representations, turning some characters red with no sense. I've disabled for now.

Issue #21, opened 2016-04-14
Generalize to other storage systems


mrocklin:
I want something very similar to `zarr` on S3 and I'm pondering the easiest way to get there.  One approach is to generalize `zarr` to accept pluggable byte storage solutions.

Currently, I believe that `zarr` effectively treats the file system as a `MutableMapping` into which it can deposit and retrieve bytes.  If this is the case then what are your thoughts on actually using the `MutableMapping` interface instead of touching files directly?  That way I could provide MutableMappings that use file systems, zip files, s3, hdfs, etc..  This nicely isolates a lot of the "where do I put this block of bytes" logic from the array slicing and compression logic.

For concreteness, here is a `MutableMapping` that loads/stores data in a directory on the file system.  https://github.com/mrocklin/zict/blob/master/zict/file.py

alimanfoo:
In principle it sounds great. 

For the non-synchronized array/chunk implementations this sounds relatively straightforward, i.e., zarr really does just need a MutableMapping where it can store and retrieve bytes. For the synchronized implementations I'm not sure how to handle the locking. I guess the locking is at a level above the MutableMapping interface, because executing ``__setitem__()`` on a chunk includes both a retrieval operation (get existing bytes) and a store operation (put modified bytes). Currently for the persistent synchronized chunks zarr uses a [fasteners inter-process lock](http://fasteners.readthedocs.org/en/latest/examples.html#interprocess-locks), which depends on access to a file system.

Very happy to discuss.

mrocklin:
Perhaps we supply both a `MutableMapping` and a `Lock`-like object?  Current pairs might be `(dict, Lock)` or `(zict.File, fasteners.InterProcessLock)`?

When dealing with distributed storage/computation we rarely care about storing in place.  It's far more common to make a completely new dataset as output.

alimanfoo:
Right, but when you are storing the output, what happens depends on how
your storage operations are aligned with chunk boundaries. If they are not
exactly aligned then two storage operations may need to store data within
the same chunk. So you need to sync access to the chunk, even when writing
new outputs.

Yes, maybe a MutableMapping and some other interface that allows to acquire
and release a lock for a given key?

On Friday, 15 April 2016, Matthew Rocklin <notifications@github.com> wrote:

> Perhaps we supply both a MutableMapping and a Lock-like object? Current
> pairs might be (dict, Lock) or (zict.File, fasteners.InterProcessLock)?
>
[...]

PR #22, opened 2016-04-19
Refactoring WIP

alimanfoo:
This PR contains work on refactoring to achieve separation of concerns between compression/decompression, storage, synchronization and array/chunk data management. Primarily motivated by #21.

Still TODO:
- [x] Improve test coverage
- [x] Rework persistance/storage documentation to generalise beyond directory store
- [x] Improve docstrings, add examples where appropriate
- [x] Do some benchmarking/profiling
- [x] Fall back to pure python install


zarr/array.py
```diff
@@ -0,0 +1,428 @@
...
+
+        self._store = store
+
+    @property
+    def store(self):
+        return self._store
+
+    @property
+    def shape(self):
+        return self._store.meta['shape']
```
mrocklin: We may want to cache the metadata on the object itself for repeated reads.  Some `MutableMapping`s will have a non-trivial latency.

zarr/store/directory.py
```diff
@@ -0,0 +1,220 @@
...
+from zarr.mappings import frozendict, Directory
+
+
+METAPATH = '__zmeta__'
+DATAPATH = '__zdata__'
+ATTRPATH = '__zattr__'
+
+
+class DirectoryStore(ArrayStore):
+
```
mrocklin: Is it possible to remove the need to create a special subclass for each kind of `MutableMapping`?

zarr/store/zipfile.py
```diff
@@ -0,0 +1,32 @@
...
+    def attrs(self):
+        pass
+
+    @property
+    def cbytes(self):
+        pass
+
+    @property
+    def initialized(self):
+        pass
```
mrocklin: For me, ideally, zarr wouldn't need to know anything about zipfiles and we could reduce the entire interface down to just `MutableMapping`.  That way you wouldn't need to maintain or test anything about zip, s3, hdfs, etc. and that can all be handled by other libraries.

mrocklin:
I'm pretty excited that this is getting started so early.

mrocklin:
Are there particular places where others can be useful?

zarr/array.py
alimanfoo: Glad you said that, I was in two minds. There is a slight complication in that the shape needs to be updated if the array is resized, but I just need to remember to update the object field.

zarr/store/directory.py
alimanfoo: Not sure what you mean, could you elaborate?

zarr/store/zipfile.py
alimanfoo: You mean reduce the entire ArrayStore interface to MutableMapping? I'm not sure that could be done as configuration metadata, user attributes and compressed chunk data all need to be managed in different ways potentially. Also an array store needs to implement the cbytes property and resize() method.

alimanfoo:
Thanks @mrocklin, I think the most useful thing would be to do what you're doing which is review the ArrayStore interface, and possibly also try to implement another ArrayStore for S3, ZipFile or something else, just to validate that the interface is reasonable and sufficient.

alimanfoo:
The MemoryStore and DirectoryStore implementations are complete and the Array class is also complete, so the code in this branch is ready for someone else to try making a different ArrayStore implementation.

zarr/array.py
mrocklin: Perhaps we cache on read and invalidate the cache on write?

zarr/store/zipfile.py
mrocklin: That was my meaning, yes.  Though the interface looks manageable enough to not be a big deal to implement.  

I am still motivated to find a way to support completely foreign MutableMappings.  I wonder if we could manage this by creating an `ArrayStore` that consumes a couple of MutableMappings, perhaps one in-memory dict or zipfile that manages metadata and another that just manages bulk data out on some large data store.  We would pass around the small mapping which would contain a pointer to the larger one.

zarr/store/memory.py
```diff
@@ -0,0 +1,82 @@
...
+
+        # setup new chunks container
+        new_data = dict()
+
+        # delete any chunks not within range
+        for ckey in self.data:
+            cidx = map(int, ckey.split('.'))
+            if all(i < c for i, c in zip(cidx, new_cdata_shape)):
+                # keep the chunk
+                new_data[ckey] = self._data[ckey]
```
mrocklin: If this could be changed to operate in place, and if we don't actually care about implementing `cbytes`, then this class may be what I'm looking for regarding a pure `MutableMapping` interface.

mrocklin: I realize though that this will still be a pain for remote datastores because we'll be calling `__getitem__` and `__setitem__` many times moving the data between the local process.

How important is `resize` to normal operation of `zarr`?  Can I allocate an empty array and then fill it and read from it without ever touching `resize`?

zarr/store/zipfile.py
alimanfoo: Yes, you could implement a generic array store class that allows the user to pass in a Mapping for the metadata and a MutableMapping for the compressed chunk data (and optionally a MutableMapping for user attributes) allowing in principle any combination of storage for these three elements.

zarr/store/memory.py
alimanfoo: This could operate in-place, see implementation in DirectoryStore. 

Supporting resize() could be optional, it is not essential to normal operation of zarr at all. I think it would be reasonable for an ArrayStore implementation to raise NotImplemented from a call to resize().

zarr/store/zipfile.py
alimanfoo: The ``cbytes`` and ``initialized`` properties could also be optional, they are useful diagnostic information but currently only used within ``Array.__repr__()``.

zarr/store/memory.py
alimanfoo: I.e., you can fill and read an array without touching resize() if you know in advance what shape the array has. The resize() method is a public API method for if a user wants to change the size of one or more dimensions of an array.

alimanfoo:
Looking at the code again I can see how it would also be possible to get a cleaner separation of concerns between the Array code and the compression/decompression code. I.e., it would be possible to define a generic interface to a Compressor, being an object that supports compress() and decompress() methods, of which BloscCompressor would be one implementation, but there could be others. 

Given that blosc is the fastest and most flexible compressor currently available I'm not sure it's worth doing this now, but maybe something to consider in the future.

mrocklin:
I generally support swappable compression mostly because blosc isn't universally deployed.  LZ4, Snappy, and ZLib are all far more common.  I think that this is really a question of how Python-centric you want the data format to be.  Short term I think that anything is fine (no need to prematurely optimize) but long term supporting more common formats may be wise.

mrocklin:
I just tried this out.  Some feedback.

I have to admit that it took me a moment to realize that I needed to handle git submodules to get `c-blosc`.  Not a big issue for most people who aren't building from source, but install docs are sometimes handy.  

Need to add `'zarr.store'` to setup.py.

It would be nice for construction functions like `empty` to take in a different store rather than by hardcoded to `MemoryStore`.

mrocklin:
Success!

```python
In [1]: import zarr
In [2]: x = zarr.empty((1000, 1000), chunks=(500, 500), dtype='f8')

In [3]: from s3fs import S3FileSystem
In [4]: from s3fs.mapping import S3Map
In [5]: s3 = S3FileSystem(anon=False)
In [6]: d = S3Map(s3, 'mrocklin/zarr-test-2')

In [7]: x.store._data = d

In [8]: %time x[:] = 1
CPU times: user 84.6 ms, sys: 64 µs, total: 84.6 ms
Wall time: 3.62 s

In [9]: %time x[:5, :5]
CPU times: user 14.8 ms, sys: 129 µs, total: 15 ms
Wall time: 289 ms
Out[9]: 
array([[ 1.,  1.,  1.,  1.,  1.],
       [ 1.,  1.,  1.,  1.,  1.],
       [ 1.,  1.,  1.,  1.,  1.],
       [ 1.,  1.,  1.,  1.,  1.],
       [ 1.,  1.,  1.,  1.,  1.]])

In [10]: s3.ls('mrocklin/zarr-test-2')
Out[10]: 
['mrocklin/zarr-test-2/0.0',
 'mrocklin/zarr-test-2/0.1',
 'mrocklin/zarr-test-2/1.0',
 'mrocklin/zarr-test-2/1.1']
```

cc @martindurant

mrocklin:
It would also be quite convenient if I could pickle up a zarr file and pass it around very cheaply (assuming that the underlying mapping is also easily picklable.)

mrocklin:
I have a simple `DictStore` implementation here: https://github.com/alimanfoo/zarr/pull/23

alimanfoo:
Hooray! I'll follow up on other points ASAP.

On Thursday, 21 April 2016, Matthew Rocklin <notifications@github.com>
wrote:

> Success!
>
> In [1]: import zarr
> In [2]: x = zarr.empty((1000, 1000), chunks=(500, 500), dtype='f8')
>
> In [3]: from s3fs import S3FileSystem
> In [4]: from s3fs.mapping import S3Map
> In [5]: s3 = S3FileSystem(anon=False)
> In [6]: d = S3Map(s3, 'mrocklin/zarr-test-2')
>
> In [7]: x.store._data = d
>
> In [8]: %time x[:] = 1CPU times: user 84.6 ms, sys: 64 µs, total: 84.6 ms
> Wall time: 3.62 s
>
> In [9]: %time x[:5, :5]CPU times: user 14.8 ms, sys: 129 µs, total: 15 ms
> Wall time: 289 ms
> Out[9]:
> array([[ 1.,  1.,  1.,  1.,  1.],
>        [ 1.,  1.,  1.,  1.,  1.],
>        [ 1.,  1.,  1.,  1.,  1.],
>        [ 1.,  1.,  1.,  1.,  1.],
>        [ 1.,  1.,  1.,  1.,  1.]])
>
> In [10]: s3.ls('mrocklin/zarr-test-2')
> Out[10]:
> ['mrocklin/zarr-test-2/0.0',
>  'mrocklin/zarr-test-2/0.1',
>  'mrocklin/zarr-test-2/1.0',
>  'mrocklin/zarr-test-2/1.1']
>
> cc @martindurant <https://github.com/martindurant>
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/alimanfoo/zarr/pull/22#issuecomment-213059163>
>


-- 
Alistair Miles
The Wellcome Trust Centre for Human Genetics
Roosevelt Drive
Oxford
OX3 7BN
United Kingdom
Email: alimanfoo@googlemail.com <alimanfoo@gmail.com>
Web: http://purl.org/net/aliman
Tel: +44 (0)1865 287721

mrocklin:
@broxtronix here is the PR allowing `zarr` to use alternative storage, including S3.  If you look a few comments ahead you'll see a tiny proof of concept just from a single machine.  I can ping you again once we have a non-trivial example with a distributed cluster and dask.array.

mrocklin:
Also cc @freemanlab .  I recall you saying you had a custom format somewhere doing this with .npy files and JSON metadata.  You might find `zarr` of interest.

mrocklin:
Verified that I can pass through a serialization cycle in about a millisecond.

mrocklin:
I'm now a bit curious about how to package up pointers to all of the relevant MutableMappings.  One approach would be to store the attrs and meta mappings in the data mapping when we explicitly flush.  We can then pass around that single large mapping and, when we load it we unpack the attrs and metadata mappings.

If we want to only assume that the data mapping can store bytes (true for many important mappings) then we probably need to serialize things.  Looking at the metadata now we see that this is pretty doable with straight JSON (which feels like the right choice to me) with a couple of exceptions:

```python
In [22]: x.store.meta
Out[22]: 
{'chunks': (1000, 1000),
 'clevel': 5,
 'cname': b'blosclz',
 'dtype': dtype('float64'),
 'fill_value': None,
 'shape': (10000, 10000),
 'shuffle': 1}
```

The `cname` attribute will have to be encoded as text and the `dtype` attribute will have to be turned into some string or nested list of strings.  I vaguely recall there being a recent issue about doing this somewhere.

mrocklin:
Or actually, the current master version of zarr must do exactly this, right?  You store everything on the file system currently.

mrocklin:
In https://github.com/alimanfoo/zarr/pull/24 I flush attrs and meta into data.  Not sure if this is desired or not.

alimanfoo:
I've just pushed some work on refactoring of storage and compression. Some notes on what's changed...

All store classes have been removed. Arrays are now initialised with a single mutable mapping. Meta and attrs are flushed into the store as JSON.

I have also refactored the compression code, putting blosc behind a generic API, and adding an implementation of zlib as primary compressor. So, i.e., now you can use blosc (with any supported internal complib) or you can use zlib directly. There are no tests for this yet but some light checks in an interactive session seem to work.

I have also unified all compression-related arguments to use a more generic form. Now the arguments 'compression' and 'compression_opts' are expected throughout. If using blosc, the user has to do e.g. ``compression='blosc', compression_opts=dict(cname='lz4', clevel=1, shuffle=1)``. This is more verbose, but I figured it was better to have a single unified API everywhere in the code. Also these arguments parallel h5py.

There are still big gaps in the test coverage, but I figured I'd wait until the API settles then backfill tests.

Comments very welcome.

mrocklin:
I'm excited to read through this.

zarr/core.py
```diff
@@ -0,0 +1,466 @@
...
+    chunks = normalize_chunks(chunks, shape)
+    dtype = np.dtype(dtype)
+    compressor_cls = get_compressor_cls(compression)
+    compression_opts = compressor_cls.normalize_compression_opts(
+        compression_opts
+    )
+
+    # delete any pre-existing items in store
+    for key in list(store.keys()):
+        del store[key]
```
mrocklin: `store.clear()` might be faster in some cases.

zarr/attrs.py
```diff
@@ -1,74 +1,81 @@
...
-
 import json
-import os
-from threading import Lock
-import fasteners
+from collections import MutableMapping
 
 
-class PersistentAttributes(object):
+class Attributes(MutableMapping):
```
mrocklin: Not very important, but this class could probably be stripped down if we wanted to.

mrocklin:
This looks great to me.  Everything seems very sensible to my eye.

zarr/attrs.py
mrocklin: On second thought.  Not sure I would bother to do this.

mrocklin:
Wait, I'm not sure how to create a new array on top of a new mutablemapping.

mrocklin:
Playing with zip files.

```python
In [1]: from zarr import empty
In [2]: from zict import Zip
In [3]: z = Zip('myfile.zarr')
In [4]: x = empty(shape=(10000, 10000), chunks=(1000, 1000), dtype='f8', store=z)

In [5]: import numpy as np
In [6]: %time template = np.random.randint(0, 10000, size=(1000, 10000))
CPU times: user 152 ms, sys: 2.99 ms, total: 155 ms
Wall time: 156 ms

In [7]: %time for i in range(10): x[i*1000: (i+1)*1000:, :] = template
CPU times: user 2.54 s, sys: 296 ms, total: 2.84 s
Wall time: 1.71 s

In [8]: z.close()

In [1]: from zict import Zip
In [2]: from zarr import Array
In [3]: x = Array(Zip('myfile.zarr'))
In [4]: x
Out[4]: 
zarr.core.Array((10000, 10000), float64, chunks=(1000, 1000))
  compression: blosc; compression_opts: {'cname': 'blosclz', 'clevel': 5, 'shuffle': 1}
  nbytes: 762.9M; initialized: 100/100
  store: zict.zip.Zip

In [5]: %time x[:, 0]
CPU times: user 132 ms, sys: 3.43 ms, total: 135 ms
Wall time: 94.3 ms
Out[5]: 
array([[ 9101.],
       [ 7018.],
       [ 3672.],
       ..., 
       [ 3458.],
       [ 7467.],
       [  889.]])

In [6]: !du -h myfile.zarr
224M	myfile.zarr
```

alimanfoo:
Awesome! I guess you hacked empty() to add the store kwarg. I'll add that in for all top level creation functions.

alimanfoo:
I'm going to leave this for a couple of days, if we're still happy with the API design then next week I'll do some work to improve test coverage.

alimanfoo:
Check this out:

```
In [1]: import zarr

In [2]: import numpy as np

In [3]: import lzma

In [4]: filters = [dict(id=lzma.FILTER_DELTA, dist=8), dict(id=lzma.FILTER_LZMA2, preset=1)]

In [5]: z_lzma = zarr.array(np.arange(1000000), chunks=100000, compression='lzma', compression_opts=dict(filters=filters))

In [6]: z_lzma
Out[6]: 
zarr.core.Array((1000000,), int64, chunks=(100000,))
  compression: lzma; compression_opts: {'filters': [{'dist': 8, 'id': 3}, {'preset': 1, 'id': 33}], 'check': 0, 'preset': None, 'format': 1}
  nbytes: 7.6M; nbytes_stored: 9.1K; ratio: 860.4; initialized: 10/10
  store: builtins.dict
```

mrocklin:
Very cool.  I like how easy https://github.com/alimanfoo/zarr/pull/22/commits/b9a385c4a5af017503c3f4a32972a52f9e2b5730 was to accomplish.  Is there a way to store arrays without compression?

alimanfoo:
Well ``compression='zlib', compression_opts=0`` will do it, but probably not the most efficient way. I guess worth adding a null compressor.

alimanfoo:
Skeleton docs are up on RTFD: http://zarr.readthedocs.io/en/refactor/index.html

alimanfoo:
Hi @mrocklin, @shoyer, I've written some new documentation to go along with the refactoring work in this PR, it's up on RTFD at http://zarr.readthedocs.io/en/refactor/index.html. If you had any time to review I'd really appreciate any comments or suggestions. 

Btw I've finished technical work on this PR, my plan is to release this as version 1.0 (major version bump to communicate backwards-incompatible API change).

mrocklin:
In the zip example you need to `flush` the MutableMapping.  This is a Zip specific thing; it stores its metadata at the end of the file.

Is `store = zarr.init_store(...); x = Array(store)` the right way to handle mutablemappings?  Thoughts on `zarr.empty(..., store=d)`?  (Either is fine with me personally.)

In the spec, thoughts on switching from *order* to *strides* a la numpy?

mrocklin:
I think that this looks great.  Perhaps we should point a few others at the spec to get their thoughts.

cc @stefanv @freeman-lab @scopatz

http://zarr.readthedocs.io/en/refactor/index.html
http://zarr.readthedocs.io/en/refactor/spec/v1.html

alimanfoo:
>
> In the zip example you need to flush the MutableMapping. This is a Zip
> specific thing; it stores its metadata at the end of the file.
>
 Thanks, I'll fix that.

> Is store = zarr.init_store(...); x = Array(store) the right way to handle
> mutablemappings? Thoughts on zarr.empty(..., store=d)? (Either is fine
> with me personally.)
>
Both are equivalent, zarr.empty(..., store=d) is shorthand. I've used the
longhand version in a couple of places just to emphasise the steps involved
in creating an array.

> In the spec, thoughts on switching from *order* to *strides* a la numpy?
>
I thought strides was something different? For zarr 'order' is roughly
analogous to the 'order' kwarg you can pass into numpy array creation
routines, except that it specifies the memory layout to be used within each
chunk rather than the whole array.

shoyer:
@alimanfoo NumPy's internal memory model doesn't use order, but rather strides which indicate how to iterate through the array. For example, for an array with shape `(n, m)` and dtype `dtype`, `order='C'` implies `strides=(n * dtype.itemsize, dtype.itemsize)` whereas `order='F'` implies `strides=(dtype.itemsize, m * dtype.itemsize)`.

I actually disagree with @mrocklin about switching to strides from order in the spec. Strides are an important part of NumPy's memory model (they allow for transposing and slicing without copies), but they don't make sense for storing data. I can't think of any circumstances with strides different from C/F order.

I would, however, clarify what C and F order refer to in the spec (column vs row-major).

alimanfoo:
Thanks @shoyer, I've added some text clarifying the meaning of 'C' and 'F' order.

alimanfoo:
@mrocklin check this out, super simple benchmark with dask.array but interesting results...

http://nbviewer.jupyter.org/github/alimanfoo/zarr/blob/refactor/notebooks/dask_copy.ipynb

The notebook profiles copying data from one array to another via da.store(), using the threaded scheduler, on my 4 core laptop.

When copying numpy array -> numpy array, giving lock=False to da.store() increases CPU utilisation but actually slows the copy down overall, perhaps memory contention?

When copying zarr array -> zarr array I get ~200% CPU with lock=True and ~400% CPU with lock=False, so turning off the lock really does help when using zarr as a data sink.

When copying h5py dataset -> h5py dataset or bcolz carray -> bcolz carray CPU is stuck at 100%, presumably because these hold the GIL.

mrocklin:
Yeah, I believe that the CPU reports as busy when its waiting for memory.

It might also be interesting to compare against a straight memory copy with numpy

```python
In [5]: %time a2[:] = a1
CPU times: user 209 ms, sys: 175 µs, total: 209 ms
Wall time: 208 ms
```

There are around 200 tasks in the computational graph.  With the graph optimization, each task probably takes around 1ms.  Dask might actually have a fair amount of the overhead here.  There are some improvements in master that would help to reduce this a bit.

It might be interesting to look at other datasets that are less compressible.  @jwittenbach recently recommended the following random dataset as a proxy for sort-of-structured data: `(np.random.normal(0, 1, size=(...)) * 100).astype('u2')`

alimanfoo:
I've updated the [dask copy benchmark](http://nbviewer.jupyter.org/github/alimanfoo/zarr/blob/refactor/notebooks/dask_copy.ipynb?flush_cache=true) to use a random normal dataset. Results are qualitatively the same, dask + zarr gives close to full CPU utilisation.

I've also run another [benchmark on some genotype data](http://nbviewer.jupyter.org/github/alimanfoo/zarr/blob/refactor/notebooks/dask_count_alleles.ipynb), again dask + zarr is very effective. I feel very nerdy saying this but it's so satisfying to see all cores fully utilised at last!

mrocklin:
This is great.  Did any actual performance improvements result from this PR or was it just that they hadn't been updated since recent blosc changes?

alimanfoo:
No this PR does not improve performance, but I didn't expect it to. I just
hadn't properly benchmarked zarr with dask before in a way that fully
showed the benefits of using them together.

I have verified separately that this PR does not degrade performance
either, so I am happy that the refactoring work is all to the good.

Btw in this PR along the way I found some bugs in the __getitem__ and
__setitem__ code for handling certain types of slices. Those bugs are
present in the previous release, so that will be another improvement.

On Thursday, May 5, 2016, Matthew Rocklin <notifications@github.com> wrote:

> This is great. Did any actual performance improvements result from this PR
> or was it just that they hadn't been updated since recent blosc changes?
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/alimanfoo/zarr/pull/22#issuecomment-217169376>
>


-- 
Alistair Miles
The Wellcome Trust Centre for Human Genetics
Roosevelt Drive
Oxford
OX3 7BN
United Kingdom
Email: alimanfoo@googlemail.com <alimanfoo@gmail.com>
Web: http://purl.org/net/aliman
Tel: +44 (0)1865 287721

alimanfoo:
I've modified setup.py so that it will only attempt to compile the Blosc extension on posix platforms. It will fall back to a pure Python install on other platforms. Also, if compilation is attempted but fails for any reason, it will fall back to pure Python installation.

I've verified that Zarr will install via pip under PyPy on Linux and CPython on Windows, some very light testing suggests it works OK under both.

I've tagged up a beta release and uploaded to PyPI, can be installed via:

```
$ pip install --pre zarr
```

I've done all I wanted to do prior to a v1.0 release. I'll give it a week or so before making the release in case anyone has feedback on implementation or docs.

Btw I have also removed the dependency on cython so now only numpy is required for installation. Currently it is a hard requirement that numpy is already installed, setup.py will fail with an error message otherwise. I know there is a variety of workarounds out there for setup.py which mean that numpy will be installed automatically if not present, but I don't particularly like the way that numpy then gets installed via easy_install with a full compile, which won't work anyway on some systems. It just seems easier and simpler right now to ask people to install numpy first, then install zarr. If anyone has experience or strong feelings about this then happy to reconsider.

shoyer:
> I've modified setup.py so that it will only attempt to compile the Blosc extension on posix platforms. It will fall back to a pure Python install on other platforms.

This means you can use zarr without Blosc? (i.e., no compression?)

Is so, that's pretty cool.

alimanfoo:
Yes, zarr can use zlib, bz2 or lzma via the Python standard library, or can
use no compression at all (compression=None).

On Friday, May 6, 2016, Stephan Hoyer <notifications@github.com> wrote:

> I've modified setup.py so that it will only attempt to compile the Blosc
> extension on posix platforms. It will fall back to a pure Python install on
> other platforms.
>
> This means you can use zarr without Blosc? (i.e., no compression?)
>
> Is so, that's pretty cool.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/alimanfoo/zarr/pull/22#issuecomment-217304999>
>


-- 
Alistair Miles
The Wellcome Trust Centre for Human Genetics
Roosevelt Drive
Oxford
OX3 7BN
United Kingdom
Email: alimanfoo@googlemail.com <alimanfoo@gmail.com>
Web: http://purl.org/net/aliman
Tel: +44 (0)1865 287721

shoyer:
very cool!

alimanfoo:
cc @benjeffery, @hardingnj, new release of zarr in preparation, docs here: http://zarr.readthedocs.io/en/refactor/index.html

scopatz:
Sorry to jump in here a a couple of days late.  First off, this seems like a very cool project; glad that something like this exists. It seems to have many of the bells and whistles I would like.

My main question about this package has to do with the memory mapping.  It seems as though using `zarr.open()` will automatically create a memory mapped array.  Is this true? What are the rules for when things are flushed or read in?  Should these rules be implementation specific or part of the spec in some way? This stuff wasn't entirely clear to me from the documentation.  

Other than that, the overall structure reminds me a bit of ASDF (XML metadata describing an array). It is probably worth mentioning somewhere in the docs.

Also, in addition to the bcolz / carray mention, what are the advantages and disadvanatges of this approach vs theirs.  I can think of a few (human readable metadata that doesn't require a custom parser), but are there others?

Are there plans for implementing this spec in other languages? C, JS, Fortran?  

Also, what order of parallelism are we looking at? Is most of the parallelism coming from thread-level? This might be a round about way of asking if big MPI jobs have been run with zarr.

I don't know if this is quite the review that you were looking for @mrocklin. If not feel free to clarify :)

shoyer:
@mrocklin perhaps zict objects (at least `Zip`) should be context managers?

@alimanfoo The API docs are missing `Array.attrs`, which would be nice to include.

Also, any thoughts on zhdf (as [described here](https://github.com/alimanfoo/zarr/issues/21#issuecomment-210554651)? Would you be open to including this as a module in zarr itself, probably `zarr.hdf`? Obviously this isn't necessary for a 1.0 release. The specification itself would be pretty trivial -- just a hierarchical directory of zarr files, with `attrs` files providing metadata at various roots. The main work would be just putting together the basic API, probably something in between h5py/zarr. I hate to commit myself to more work that I don't have time to do, but this feels like something I might be driven to put together a PR for at some point...

alimanfoo:
Thanks @scopatz for the comments, much appreciated.

``zarr.open()`` is a convenience function for setting up an array backed by a directory store. I.e.:

```
z = zarr.open('example.zarr', mode='w', shape=100, chunks=10)
```

...is short-hand for...

```
store = zarr.DirectoryStore('example.zarr')
zarr.init_store(store, shape=100, chunks=10, overwrite=True)
z = zarr.Array(store)
```

The name "open" is perhaps not very appropriate, because no files are actually being maintained in an open state. Whenever any data in this array is modified (e.g., ``z[:] = 1``) the ``zarr.Array`` implementation will work one-by-one through the underlying chunks, compressing data and updating the store. What the array store then does with the updated chunk data depends on how the store is implemented and is flexible. The ``zarr.DirectoryStore`` class immediately writes data to files, so in this case data are immediately flushed to disk, and nothing is being memory-mapped (beyond what the operating system will do transparently). If a different store is used (e.g., ``zict.Zip``, providing a storage interface to a zip file) then the effect of modifying data in the array would be completely different in terms of changes to the file system. Part of the idea in the spec is to abstract the array store interface, so there is flexibility to implement different types of array storage.

The main advantages over ``bcolz.carray`` are: (1) Zarr chunks in multiple dimensions whereas Bcolz only chunks the first dimension, so with Zarr you can get good performance for multiple data access patterns; (2) Zarr (in this PR) has a layer of abstraction between the array and the storage, so there is flexibility to store Zarr arrays in memory, in a file system directory, in a zip file, on S3, on HDFS, etc; and (3) Zarr supports chunk-level locking and releases the GIL during compression/decompression operations, so you can safely use Zarr arrays in both multi-threaded and multi-process parallel programs, and can maximise the amount of work that can be done in parallel.

I personally don't have the time or skills to implement the Zarr spec in other languages, but thought that it was worth writing up in case anyone else found it worthwhile to do another implementation.

Regarding parallelism, Zarr arrays should work well in either multi-threaded or multi-process/distributed programs. Personally I have only benchmarked Zarr in multi-threaded programs. I believe @mrocklin has plans to try out Zarr in some larger distributed programs using S3 as an array store. No-one has tried using Zarr in an MPI program AFAIK.

Thanks again, any further comments or suggestions very welcome.

alimanfoo:
Thanks @shoyer, Array.attrs added to the docs.

I think zhdf is a great idea, I would definitely use it myself. I imagine it would be possible to emulate the h5py API quite closely if you wanted to. I think it would make sense as a separate module, e.g., zarr.hdf as you suggest. I don't have time to do it myself but would be very happy to review a PR at any point.

alimanfoo:
RTFD build has started failing, I don't think it's my fault, [issue raised](https://github.com/rtfd/readthedocs.org/issues/2197).

alimanfoo:
I've made a new beta release: 1.0.0b4. Two significant changes:

* Fix a memory leak in the blosc C extension.
* The blosc extension now automatically decides how it should call c-blosc (using _ctx methods or not) based on whether it is being called from the main Python thread or not. This saves the user having to do any switching if using a multi-threaded tool like dask.

alimanfoo:
Hi @mrocklin, I was slightly premature asking ``zict.Zip`` to support ``bytearray`` values. I did some further investigation and it turns out I can only achieve the optimisation I was looking for if I use the Python ``array`` module instead, because there is some overhead involved in initialising a ``bytearray``. 

As it currently stands in this PR, the Blosc extension ``compress()`` function performs a [copy of the compressed data](https://github.com/alimanfoo/zarr/blob/refactor/zarr/blosc.pyx#L171) in order to produce a Python bytes value containing the compressed data. This extra memory copy [can be avoided](https://github.com/alimanfoo/zarr/blob/refactor_pybuffer/zarr/blosc.pyx#L130) by using the Python ``array`` module, which gives a speedup of 10-20% in some simple benchmarks.

I haven't included any changes in this PR yet as I wanted to check with you. To make this optimisation possible we would have to widen the set of possible values that an array store ``MutableMapping`` will accept to include ``bytes`` and ``array.array``. The ``array.array`` class implements the buffer protocol and the Python docs say it can be used wherever a [bytes-like object](https://docs.python.org/3/glossary.html#term-bytes-like-object) is expected, so this seems ok to me. I checked with the zipfile module and you can pass an instance of ``array.array`` into ``writestr()``. Are you ok with this?

mrocklin:
Generally I agree that it's a good idea for key-value stores to support all
of bytes, bytearrays, memoryviews, etc..  At the same time I think that
Zarr should be robust to the absence of this support if possible, even if
it's somewhat inefficient.  Perhaps this is just a try-except block around
the setitem call on the Zarr side?

alimanfoo:
Thanks @mrocklin for modifying zict to accept other value types. I'm slightly reluctant to add some try-catch-retry logic to Zarr to handle storage that supports only bytes values. I think it would be better to surface the issue and encourage the store implementation to support all bytes-like values, which in most cases should be trivially possible. If it really isn't doable then they could always wrap the store in something like ``zict.func.Func``.

mrocklin:
OK

shoyer:
+1 for not adding fallback logic

mrocklin:
cc @martindurant 

Thoughts on supporting other byte-like structures than `bytes` in `hdfs3` and `s3fs`?

martindurant:
I am not immediately sure. I am pretty buried this week, but I might have time to look into it this evening.

alimanfoo:
If you could point me to the relevant parts of the code for writing data
out to S3 or HDFS I'd be happy to look into whether Python arrays or other
bytes-like values could be accepted.

On Tue, May 10, 2016 at 6:01 PM, Martin Durant <notifications@github.com>
wrote:

> I am not immediately sure. I am pretty buried this week, but I might have
> time to look into it this evening.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/alimanfoo/zarr/pull/22#issuecomment-218223028>
>



-- 
Alistair Miles
The Wellcome Trust Centre for Human Genetics
Roosevelt Drive
Oxford
OX3 7BN
United Kingdom
Email: alimanfoo@googlemail.com <alimanfoo@gmail.com>
Web: http://purl.org/net/aliman
Tel: +44 (0)1865 287721

mrocklin:
You would probably want to start from here:

*  https://github.com/dask/s3fs/blob/master/s3fs/mapping.py#L65
*  https://github.com/dask/hdfs3/blob/master/hdfs3/mapping.py#L65

The gritty details will be in the underlying libraries `boto3` and `libhdfs3`

shoyer:
S3FS buffers data in a BytesIO before writing. So that should already
handle writing bytesarray objects (possibly after some type checks are
relaxed). For HDFS3, the gritty details do indeed appear to be in libhdfs3.

alimanfoo:
Confirmed s3fs bottoms out in ``io.BytesIO.write()`` which will accept ``bytes``, ``bytearray`` or ``array.array``.

hdfs3 bottoms out in ``libhdfs3.hdfsWrite()`` which is [wrapped via ctypes](https://github.com/dask/hdfs3/blob/master/hdfs3/lib.py#L320). The ctypes docs say that "None, integers, bytes objects and (unicode) strings are the only native Python objects that can directly be used as parameters in these function calls." I did a simple experiment with ``libc.printf`` which confirmed that passing ``bytearray`` or ``array.array`` raises a TypeError. 

However, hdfs3 currently pushes all values through an [ensure_bytes()](https://github.com/dask/hdfs3/blob/master/hdfs3/core.py#L71) function before calling libhdfs3. This could be modified with another couple of checks, e.g.:

```
def ensure_bytes(s):
    """ Give strings that ctypes is guaranteed to handle """
    if isinstance(s, dict):
        return {k: ensure_bytes(v) for k, v in s.items()}
    if isinstance(s, str) and sys.version_info < (3,):
        return s
    if hasattr(s, 'encode'):
        return s.encode()
    if hasattr(s, 'tobytes'):
        return s.tobytes()
    if isinstance(s, bytearray):
        return bytes(s)
    else:
        return s
```

martindurant:
Agreed, can simply add bytearray tests to s3fs, and I think it's reasonable to edit hdfs3's `ensure_bytes` and add the same tests there too.

alimanfoo:
Great. Btw I am using array.array now instead of bytearray. Probably worth
covering both types.

On Wednesday, 11 May 2016, Martin Durant <notifications@github.com
<javascript:_e(%7B%7D,'cvml','notifications@github.com');>> wrote:

> Agreed, can simply add bytearray tests to s3fs, and I think it's
> reasonable to edit hdfs3's ensure_bytes and add the same tests there too.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/alimanfoo/zarr/pull/22#issuecomment-218341951>
>


-- 
Alistair Miles
The Wellcome Trust Centre for Human Genetics
Roosevelt Drive
Oxford
OX3 7BN
United Kingdom
Email: alimanfoo@googlemail.com <alimanfoo@gmail.com>
Web: http://purl.org/net/aliman
Tel: +44 (0)1865 287721

alimanfoo:
Hi @mrocklin, can I ask what your plans are for trying out zarr with S3 and/or HDFS? I'm just wondering if I should hold back on a 1.0.0 release until you've had a chance to try it out at scale or if I should go ahead with a release now. It's ready to go as far as I'm concerned, but it would of course be great to have some proof of principle that distributed storage is working as expected.

mrocklin:
Good question.  I've been pulled onto other things over the last week.

I'll carve out some time today to fool around.

mrocklin:
Performance experiment: https://www.youtube.com/watch?v=8WtaYvqhxHc

I'm pretty satisfied with it.  I did run into some issues with blosc.  Both with cname being a string, rather than bytes.  Also ran into this: 

```python
  File "/home/mrocklin/workspace/dask/dask/array/core.py", line 2041, in store
    out[index] = np.asanyarray(x)
  File "/home/mrocklin/workspace/zarr/zarr/core.py", line 457, in __setitem__
    value[value_selection])
  File "/home/mrocklin/workspace/zarr/zarr/core.py", line 573, in _chunk_setitem
    cdata = self._compressor.compress(chunk)
  File "/home/mrocklin/workspace/zarr/zarr/compressors.py", line 152, in compress
    return blosc.compress(array, self.cname, self.clevel, self.shuffle)
  File "zarr/blosc.pyx", line 118, in zarr.blosc.compress (zarr/blosc.c:2027)
    # perform decompression
TypeError: compress() takes exactly 5 positional arguments (4 given)
```

I got past this by just using `compression=None`, which was convenient.

mrocklin:
Non-basic explanation starts at around 2m20s

alimanfoo:
Way way cool, that made my week!

Could you give a small example of how the cname issue came up?

For the other error, is it possible that some of your workers have an older
version of zarr? The line number in blosc.pyx (118) doesn't correspond with
the current state of the code.

On Thursday, May 12, 2016, Matthew Rocklin <notifications@github.com> wrote:

> Performance experiment: https://www.youtube.com/watch?v=8WtaYvqhxHc
>
> I'm pretty satisfied with it. I did run into some issues with blosc. Both
> with cname being a string, rather than bytes. Also ran into this:
>
>   File "/home/mrocklin/workspace/dask/dask/array/core.py", line 2041, in store
>     out[index] = np.asanyarray(x)
>   File "/home/mrocklin/workspace/zarr/zarr/core.py", line 457, in __setitem__
>     value[value_selection])
>   File "/home/mrocklin/workspace/zarr/zarr/core.py", line 573, in _chunk_setitem
>     cdata = self._compressor.compress(chunk)
>   File "/home/mrocklin/workspace/zarr/zarr/compressors.py", line 152, in compress
>     return blosc.compress(array, self.cname, self.clevel, self.shuffle)
>   File "zarr/blosc.pyx", line 118, in zarr.blosc.compress (zarr/blosc.c:2027)
>     # perform decompressionTypeError: compress() takes exactly 5 positional arguments (4 given)
>
> I got past this by just using compression=None, which was convenient.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/alimanfoo/zarr/pull/22#issuecomment-218873415>
>


-- 
Alistair Miles
The Wellcome Trust Centre for Human Genetics
Roosevelt Drive
Oxford
OX3 7BN
United Kingdom
Email: alimanfoo@googlemail.com <alimanfoo@gmail.com>
Web: http://purl.org/net/aliman
Tel: +44 (0)1865 287721

mrocklin:
I was just creating an empty zarr array in Python 3.  My environment clearly has issue though, so maybe that's all

I figured the compress issue was a version problem.  I nuked everything I could and reinstalled but to no avail.  Seems like this is more likely an issue with my environment than with this PR though, so it can likely be ignored.

alimanfoo:
Ok, going for release on Monday if no objections.

alimanfoo:
@mrocklin can you help, if I want to make the new release of Zarr also available via conda, what is the best way to do that? Create a conda recipe within the Zarr GitHub repo? If so, what's the best way to then build and distribute?

mrocklin:
I'd try out [conda forge](https://conda-forge.github.io/) which both Continuum and the community seem to be converging on as a nice community solution.

I've never done this before (I rarely make conda packages even for the default channel) but would be happy to help if necessary.  

Thoughts on vendoring blosc vs depending on it?

alimanfoo:
Thanks a lot, I'll try conda forge.

Re blosc, you mean it would be an option to create a separate conda package
for it then make it a build dependency, rather than bundling it?

On Fri, May 13, 2016 at 3:05 PM, Matthew Rocklin <notifications@github.com>
wrote:

> I'd try out conda forge <https://conda-forge.github.io/> which both
> Continuum and the community seem to be converging on as a nice community
> solution.
>
> I've never done this before (I rarely make conda packages even for the
> default channel) but would be happy to help if necessary.
>
> Thoughts on vendoring blosc vs depending on it?
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/alimanfoo/zarr/pull/22#issuecomment-219052539>
>



-- 
Alistair Miles
The Wellcome Trust Centre for Human Genetics
Roosevelt Drive
Oxford
OX3 7BN
United Kingdom
Email: alimanfoo@googlemail.com <alimanfoo@gmail.com>
Web: http://purl.org/net/aliman
Tel: +44 (0)1865 287721

mrocklin:
Yes, that was my question.  Why do you bundle it now?

alimanfoo:
It saves the user having to download blosc and install separately, then
figure out how to point the zarr install to the right location so it can
build against it. It also means we can be sure to fix a specific c-blosc
version. I was copying what bcolz does, except they actually copy c-blosc
into the bcolz source tree whereas zarr uses git submodules.

I'd like to keep a 1-line install for zarr. I guess this could be done with
conda with a separate blosc dependency. But for installing zarr via pip, I
think bundling blosc is the only option?



On Fri, May 13, 2016 at 3:40 PM, Matthew Rocklin <notifications@github.com>
wrote:

> Yes, that was my question. Why do you bundle it now?
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/alimanfoo/zarr/pull/22#issuecomment-219062663>
>



-- 
Alistair Miles
The Wellcome Trust Centre for Human Genetics
Roosevelt Drive
Oxford
OX3 7BN
United Kingdom
Email: alimanfoo@googlemail.com <alimanfoo@gmail.com>
Web: http://purl.org/net/aliman
Tel: +44 (0)1865 287721

FrancescAlted:
Hey, pretty impressive work.  Well done!  BTW, why not enabling Blosc on Windows?  In 1.8.0 I reverted some incompatibilities with MSVC 2008, so everything should be good there.  In case you don't have access to a Windows box, you can adapt the [AppVeyor config files from bcolz](https://github.com/Blosc/bcolz/blob/master/appveyor.yml) to test that everything is fine.

alimanfoo:
Thanks Francesc, it would be great to have blosc compile on Windows. Thanks
for the pointer to the appveyor setup, I'll see if I can get that working.

On Friday, 13 May 2016, FrancescAlted <notifications@github.com> wrote:

> Hey, pretty impressive work. Well done! BTW, why not enabling Blosc on
> Windows? In 1.8.0 I reverted some incompatibilities with MSVC 2008, so
> everything should be good there. In case you don't have access to a Windows
> box, you can adapt the AppVeyor config files from bcolz
> <https://github.com/Blosc/bcolz/blob/master/appveyor.yml> to test that
> everything is fine.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/alimanfoo/zarr/pull/22#issuecomment-219096323>
>


-- 
Alistair Miles
The Wellcome Trust Centre for Human Genetics
Roosevelt Drive
Oxford
OX3 7BN
United Kingdom
Email: alimanfoo@googlemail.com <alimanfoo@gmail.com>
Web: http://purl.org/net/aliman
Tel: +44 (0)1865 287721

alimanfoo:
I've modified setup.py so Zarr will now attempt to build the blosc
extension on Windows.

I've verified manually that the blosc extension will build on Windows for
Python 3.5, and all tests pass. In the process I found a Windows-specific
bug in the DirectoryStore class which was fairly important, so it was well
worth it just to find that.

AppVeyor is set up and all tests are passing for 32-bit builds. However the
Python 2.7 64-bit build is failing. The error message suggests syntax
errors in shuffle.c?
https://ci.appveyor.com/project/alimanfoo/zarr/build/1.0.15/job/skeweaiuuekga03w#L151

@FrancescAlted do you know what's happening with the 64-bit build?

On Fri, May 13, 2016 at 7:40 PM, Alistair Miles <alimanfoo@googlemail.com>
wrote:

> Thanks Francesc, it would be great to have blosc compile on Windows. Thanks
> for the pointer to the appveyor setup, I'll see if I can get that working.
>
>
> On Friday, 13 May 2016, FrancescAlted <notifications@github.com> wrote:
>
>> Hey, pretty impressive work. Well done! BTW, why not enabling Blosc on
>> Windows? In 1.8.0 I reverted some incompatibilities with MSVC 2008, so
>> everything should be good there. In case you don't have access to a Windows
>> box, you can adapt the AppVeyor config files from bcolz
>> <https://github.com/Blosc/bcolz/blob/master/appveyor.yml> to test that
>> everything is fine.
>>
>> —
>> You are receiving this because you were mentioned.
>> Reply to this email directly or view it on GitHub
>> <https://github.com/alimanfoo/zarr/pull/22#issuecomment-219096323>
>>
>
>
> --
> Alistair Miles
> Head of Epidemiological Informatics
> Centre for Genomics and Global Health <http://cggh.org>
> The Wellcome Trust Centre for Human Genetics
> Roosevelt Drive
> Oxford
> OX3 7BN
> United Kingdom
> Email: alimanfoo@googlemail.com <alimanfoo@gmail.com>
> Web: http://purl.org/net/aliman
> Twitter: https://twitter.com/alimanfoo
> Tel: +44 (0)1865 287721
>
>


-- 
Alistair Miles
The Wellcome Trust Centre for Human Genetics
Roosevelt Drive
Oxford
OX3 7BN
United Kingdom
Email: alimanfoo@googlemail.com <alimanfoo@gmail.com>
Web: http://purl.org/net/aliman
Tel: +44 (0)1865 287721

FrancescAlted:
Hmm, yes, could be an issue in C-Blosc 1.8.1.  Could you try with C-Blosc 1.9.0?  It should work well (at least it worked fine with [previous builds of  bcolz](https://ci.appveyor.com/project/FrancescAlted/bcolz/build/1.0.189/job/e9gp0q43d76c1fjn)).

alimanfoo:
@FrancescAlted [retrying with a simpler appveyor config](https://ci.appveyor.com/project/alimanfoo/zarr/build/1.0.21) that doesn't use conda shows it is [just Python27-x64](https://ci.appveyor.com/project/alimanfoo/zarr/build/1.0.21/job/sxfvm862wgr1tujq) that is problematic. With this appveyor config the blosc extension now builds without any errors, however the tests that test the extension fail.

I haven't tried upgrading to c-blosc 1.9.0 yet, I'll try that now and see if it changes anything.

alimanfoo:
[Confirmed](https://ci.appveyor.com/project/alimanfoo/zarr/build/1.0.23) behaviour is the same with c-blosc 1.9.0, blosc extension builds but tests fail (segfault?).

The problem could be in my cython code rather than in c-blosc. On Linux everything is fine with PY27 64-bit so must be Windows-specific. 

I'll try to find some time to investigate but may have to release without this and fix later.

FrancescAlted:
Hmm, that's interesting because bcolz has no problems specifically with Python 2.7 on Win64 and conda build.  I would stick with your simplified appveyor configuration as it compiles well.  And yes, I would say that the segfault (or whatever is this) may indicate a problem with your code (but one cannot be sure without further debugging).

FrancescAlted:
For what is worth, I cloned the same strategy than @alimanfoo in order to use contextual Blosc calls when bcolz is called from a multithreaded application.  You can see some [preliminary results here](https://github.com/Blosc/bcolz/blob/release-gil/bench/multithreaded_copy.ipynb).  There I discovered that zarr is more efficient when not using compression.  My hunch is that bcolz still uses a Blosc container even when not using compression (this is for simplicity), and that has some cost (which, albeit small, it is non-negligible).

alimanfoo:
OK, after some thrashing about I've fixed the appveyor build. 

For the record, the example given [here](http://python-packaging-user-guide.readthedocs.io/en/latest/appveyor/) for appveyor setup is wrong (I think) for Python 2.7 64-bit, it ends up using the wrong compiler version.

Also I had to add ``extra_compile_args.append('-Dinline=__inline')`` when compiling for Py27-64bit because the "inline" keyword is apparently not supported in the compiler version used (and so you get a syntax error reported against c-blosc/shuffle.c). How bcolz builds on appveyor without having to apply this workaround I do not understand.

alimanfoo:
Thanks @FrancescAlted, interesting benchmarks. Btw how many cores do you have?

I haven't done any systematic testing but when using blosc in multi-threaded (global locking) mode I don't usually see much if any speed improvement beyond 4 blosc internal threads. So using dask with blosc in contextual mode should scale better as the number of available cores increases past 4.

alimanfoo:
Oh @FrancescAlted I've been meaning to ask, what would you expect to happen if I compile blosc on a CPU with AVX2 support, but then try to run those binaries on an older CPU without AVX2?

FrancescAlted:
OMG, the wonderful world of Microsoft compilers is a bit like black magic for Unix nerds :P  Uh, no idea why bcolz is compiling C-Blosc correctly whereas zarr needs `-Dinline=__inline`, but it will be interesting for me to do an exercise and try to reproduce your results.

FrancescAlted:
Regarding the C-Blosc AVX2 support, all the internally recognized SIMD (SSE2, AVX2) are detected in *run-time*, so your best bet would be to create wheels (or any other binary) in a machine with AVX2, and the internal Blosc machinery will dispatch the best SIMD found in the executing machine.

FrancescAlted:
And my benchmarking machine has 8 physical cores (Xeon E3-1240 v3 @ 3.40GHz).  But I did not put too much effort in the benchmark above, so it well may be that using less threads internally in Blosc may give better results.  Although now that I think when running dask with your patch, the internal number of threads in Blosc is set to 1, so I'd say the current setup should be pretty good.

alimanfoo:
Thanks Francesc.

Btw I meant that when using zarr by itself (i.e., not with dask) but
varying the number of internal threads blosc is allowed to use, I generally
see performance max out with 4 internal threads. I.e., allowing blosc to
use more than 4 internal threads doesn't seem to improve things. But I
should double-check that...

On Mon, May 16, 2016 at 3:53 PM, FrancescAlted <notifications@github.com>
wrote:

> And my benchmarking machine has 8 physical cores (Xeon E3-1240 v3 @
> 3.40GHz). But I did not put too much effort in the benchmark above, so it
> well may be that using less threads internally in Blosc may give better
> results. Although now that I think when running dask with your patch, the
> internal number of threads in Blosc is set to 1, so I'd say the current
> setup should be pretty good.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/alimanfoo/zarr/pull/22#issuecomment-219445789>
>



-- 
Alistair Miles
The Wellcome Trust Centre for Human Genetics
Roosevelt Drive
Oxford
OX3 7BN
United Kingdom
Email: alimanfoo@googlemail.com <alimanfoo@gmail.com>
Web: http://purl.org/net/aliman
Tel: +44 (0)1865 287721

FrancescAlted:
I did a quick try, and 8 threads still has an advantage over 4 for this machine and the random walk example:

- 1 threads: ~ 1240 ms
- 2 threads: ~ 800 ms
- 4 threads: ~ 470 ms
- 6 threads: ~ 410 ms
- 8 threads: ~ 370 ms

FrancescAlted:
BTW, I realized that in [my benchmark](https://github.com/Blosc/bcolz/blob/release-gil/bench/multithreaded_copy.ipynb) the Bokeh plots are not stored.  Any hint?

mrocklin:
GitHub won't render the Bokeh plots by default (they run JS code).  NBViewer works:

http://nbviewer.jupyter.org/github/Blosc/bcolz/blob/release-gil/bench/multithreaded_copy.ipynb

FrancescAlted:
Nice.  Thanks @mrocklin

alimanfoo:
Thanks Francesc, interesting. Looks like there is a fixed overhead
associated with using multiple threads? ...

In [16]: x = np.array([1, 2, 4, 6, 8])  # no. threads used

In [17]: y1 = np.array([1240, 800, 470, 410, 370])  # actual time

In [18]: y2 = (y1[0] / x).astype(int)  # expected time (theoretical minimum)

In [19]: y2
Out[19]: array([1240,  620,  310,  206,  155])

In [20]: y1 - y2
Out[20]: array([  0, 180, 160, 204, 215])

On Monday, May 16, 2016, FrancescAlted <notifications@github.com> wrote:

> I did a quick try, and 8 threads still has an advantage over 4 for this
> machine and the random walk example:
>
>    - 1 threads: ~ 1240 ms
>    - 2 threads: ~ 800 ms
>    - 4 threads: ~ 470 ms
>    - 6 threads: ~ 410 ms
>    - 8 threads: ~ 370 ms
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/alimanfoo/zarr/pull/22#issuecomment-219459151>
>


-- 
Alistair Miles
The Wellcome Trust Centre for Human Genetics
Roosevelt Drive
Oxford
OX3 7BN
United Kingdom
Email: alimanfoo@googlemail.com <alimanfoo@gmail.com>
Web: http://purl.org/net/aliman
Tel: +44 (0)1865 287721

FrancescAlted:
You mean that you are trying to pinpoint where the overhead is?  If so, heh, that's difficult to assess.  CPUs are complex beasts, specially cache behavior, and understanding where the overhead lies is difficult, to say the least.  Despite that I have spent several years (not continuously obviously :)  thinking about this subject, I don't dare to give even a guess.

alimanfoo:
Fair enough :-) Just thought it was interesting that speed looked like s/n
+ o where s is time taken with single thread, n is number of threads and c
is a constant. If generally true this would be good scalability, as long as
o is small compared to s/n.

On Monday, 16 May 2016, FrancescAlted <notifications@github.com> wrote:

> You mean that you are trying to pinpoint where the overhead is? If so,
> heh, that's difficult to assess. CPUs are complex beasts, specially cache
> behavior, and understanding where the overhead lies is difficult, to say
> the least. Despite that I have spent several years (not continuously
> obviously :) thinking about this subject, I don't dare to give even a guess.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/alimanfoo/zarr/pull/22#issuecomment-219482160>
>


-- 
Alistair Miles
The Wellcome Trust Centre for Human Genetics
Roosevelt Drive
Oxford
OX3 7BN
United Kingdom
Email: alimanfoo@googlemail.com <alimanfoo@gmail.com>
Web: http://purl.org/net/aliman
Tel: +44 (0)1865 287721

alimanfoo:
Merging and release tomorrow if no comments.

FrancescAlted:
Right, this kind of scalability is quite typical in concurrent systems; the goal is of course minimize the 'o'.

mrocklin:
Very glad to see this merged :)

alimanfoo:
Thanks Matt. Thanks to everyone for the support.

On Tuesday, May 17, 2016, Matthew Rocklin <notifications@github.com> wrote:

> Very glad to see this merged :)
>
> On Tue, May 17, 2016 at 1:30 AM, FrancescAlted <notifications@github.com
> <javascript:_e(%7B%7D,'cvml','notifications@github.com');>>
> wrote:
>
> > Right, this kind of scalability is quite typical in concurrent systems;
> > the goal is of course minimize the 'o'.
> >
> > —
> > You are receiving this because you were mentioned.
> > Reply to this email directly or view it on GitHub
> > <https://github.com/alimanfoo/zarr/pull/22#issuecomment-219653079>
> >
>
> —
> You are receiving this because you modified the open/close state.
> Reply to this email directly or view it on GitHub
> <https://github.com/alimanfoo/zarr/pull/22#issuecomment-219729465>
>


-- 
Alistair Miles
The Wellcome Trust Centre for Human Genetics
Roosevelt Drive
Oxford
OX3 7BN
United Kingdom
Email: alimanfoo@googlemail.com <alimanfoo@gmail.com>
Web: http://purl.org/net/aliman
Tel: +44 (0)1865 287721

shoyer:
I'm also really stoked to see this! I think zarr could have a lot of impact
-- personally I'm much more excited about it than ASDF, for example.

alimanfoo:
Thanks Stephan. I'm really glad you and Matt had the idea of refactoring
the storage layer, I think it's led to something many times better than
what I started out with.

On Tuesday, 17 May 2016, Stephan Hoyer <notifications@github.com> wrote:

> I'm also really stoked to see this! I think zarr could have a lot of impact
> -- personally I'm much more excited about it than ASDF, for example.
>
> On Tue, May 17, 2016 at 9:58 AM, Alistair Miles <notifications@github.com
> <javascript:_e(%7B%7D,'cvml','notifications@github.com');>>
> wrote:
>
> > Thanks Matt. Thanks to everyone for the support.
> >
> > On Tuesday, May 17, 2016, Matthew Rocklin <notifications@github.com
> <javascript:_e(%7B%7D,'cvml','notifications@github.com');>>
> > wrote:
> >
> > > Very glad to see this merged :)
> > >
> > > On Tue, May 17, 2016 at 1:30 AM, FrancescAlted <
> notifications@github.com
> <javascript:_e(%7B%7D,'cvml','notifications@github.com');>
> > > <javascript:_e(%7B%7D,'cvml','notifications@github.com
> <javascript:_e(%7B%7D,'cvml','notifications@github.com');>');>>
> > > wrote:
> > >
> > > > Right, this kind of scalability is quite typical in concurrent
> systems;
> > > > the goal is of course minimize the 'o'.
> > > >
> > > > —
> > > > You are receiving this because you were mentioned.
> > > > Reply to this email directly or view it on GitHub
> > > > <https://github.com/alimanfoo/zarr/pull/22#issuecomment-219653079>
> > > >
> > >
> > > —
> > > You are receiving this because you modified the open/close state.
> > > Reply to this email directly or view it on GitHub
> > > <https://github.com/alimanfoo/zarr/pull/22#issuecomment-219729465>
> > >
> >
> >
> > --
> > Alistair Miles
> > Head of Epidemiological Informatics
> > Centre for Genomics and Global Health <http://cggh.org>
> > The Wellcome Trust Centre for Human Genetics
> > Roosevelt Drive
> > Oxford
> > OX3 7BN
> > United Kingdom
> > Email: alimanfoo@googlemail.com
> <javascript:_e(%7B%7D,'cvml','alimanfoo@googlemail.com');> <
> alimanfoo@gmail.com <javascript:_e(%7B%7D,'cvml','alimanfoo@gmail.com');>>
> > Web: http://purl.org/net/aliman
> > Twitter: https://twitter.com/alimanfoo
> > Tel: +44 (0)1865 287721
> >
> > —
> > You are receiving this because you were mentioned.
> > Reply to this email directly or view it on GitHub
> > <https://github.com/alimanfoo/zarr/pull/22#issuecomment-219783152>
> >
>
> —
> You are receiving this because you modified the open/close state.
> Reply to this email directly or view it on GitHub
> <https://github.com/alimanfoo/zarr/pull/22#issuecomment-219796881>
>


-- 
Alistair Miles
The Wellcome Trust Centre for Human Genetics
Roosevelt Drive
Oxford
OX3 7BN
United Kingdom
Email: alimanfoo@googlemail.com <alimanfoo@gmail.com>
Web: http://purl.org/net/aliman
Tel: +44 (0)1865 287721

PR #24, opened 2016-04-22
Refactor 3

mrocklin:
Flush and maintain metadata and attrs in data dict

### DIFF
--- a/zarr/core.py
+++ b/zarr/core.py
@@ -1,5 +1,7 @@
 # -*- coding: utf-8 -*-
 from __future__ import absolute_import, print_function, division
+
+from collections import MutableMapping
 from functools import reduce  # TODO PY2 compatibility
 import operator
 import itertools
@@ -54,7 +56,7 @@ _repr_shuffle = [
 
 class Array(object):
 
-    def __init__(self, store):
+    def __init__(self, store, **kwargs):
         """Instantiate an array.
 
         Parameters
@@ -63,6 +65,9 @@ class Array(object):
             Array store.
 
         """
+        if isinstance(store, MutableMapping):
+            from .store import ArrayStore
+            store = ArrayStore(data=store, **kwargs)
 
         self.store = store
         self.shape = store.meta['shape']
--- a/zarr/meta.py
+++ b/zarr/meta.py
@@ -13,26 +13,32 @@ PY2 = sys.version_info[0] == 2
 
 def loads(b):
     meta = json.loads(b)
-
-    # decode some values
-    meta['shape'] = tuple(meta['shape'])
-    meta['chunks'] = tuple(meta['chunks'])
-    meta['cname'] = meta['cname'].encode('ascii')
-    meta['dtype'] = decode_dtype(meta['dtype'])
+    if meta['compression'] not in ('blosc', None):
+        raise NotImplementedError("Only Blosc and no-compression are supported")
+    meta = dict(
+        shape=tuple(meta['shape']),
+        chunks=tuple(meta['chunks']),
+        dtype=decode_dtype(meta['dtype']),
+        fill_value=meta['fill_value'],
+        cname=meta['compression_opts']['cname'].encode('ascii'),
+        clevel=meta['compression_opts']['clevel'],
+        shuffle=meta['compression_opts']['shuffle']
+    )
 
     return meta
 
 
 def dumps(meta):
-    # construct metadata dictionary
     meta = dict(
         shape=meta['shape'],
         chunks=meta['chunks'],
         dtype=encode_dtype(meta['dtype']),
-        cname=meta['cname'] if PY2 else str(meta['cname'], 'ascii'),
-        clevel=meta['clevel'],
-        shuffle=meta['shuffle'],
         fill_value=meta['fill_value'],
+        compression='blosc',
+        compression_opts=dict(
+            cname=meta['cname'] if PY2 else str(meta['cname'], 'ascii'),
+            clevel=meta['clevel'],
+            shuffle=meta['shuffle'])
     )
 
     return json.dumps(meta, indent=4, sort_keys=True)
--- a/zarr/meta.py
+++ b/zarr/meta.py
@@ -1,0 +1,62 @@
+# -*- coding: utf-8 -*-
+from __future__ import absolute_import, print_function, division
+
+import json
+import sys
+import numpy as np
+
+from zarr import defaults as _defaults
+
+
+PY2 = sys.version_info[0] == 2
+
+
+def loads(b):
+    meta = json.loads(b)
+
+    # decode some values
+    meta['shape'] = tuple(meta['shape'])
+    meta['chunks'] = tuple(meta['chunks'])
+    meta['cname'] = meta['cname'].encode('ascii')
+    meta['dtype'] = decode_dtype(meta['dtype'])
+
+    return meta
+
+
+def dumps(meta):
+    # construct metadata dictionary
+    meta = dict(
+        shape=meta['shape'],
+        chunks=meta['chunks'],
+        dtype=encode_dtype(meta['dtype']),
+        cname=meta['cname'] if PY2 else str(meta['cname'], 'ascii'),
+        clevel=meta['clevel'],
+        shuffle=meta['shuffle'],
+        fill_value=meta['fill_value'],
+    )
+
+    return json.dumps(meta, indent=4, sort_keys=True)
+
+
+def encode_dtype(d):
+    if d.fields is None:
+        return d.str
+    else:
+        return d.descr
+
+
+def _decode_dtype_descr(d):
+    # need to convert list of lists to list of tuples
+    if isinstance(d, list):
+        # recurse to handle nested structures
+        if PY2:
+            # under PY2 numpy rejects unicode field names
+            d = [(f.encode('ascii'), _decode_dtype_descr(v)) for f, v in d]
+        else:
+            d = [(f, _decode_dtype_descr(v)) for f, v in d]
+    return d
+
+
+def decode_dtype(d):
+    d = _decode_dtype_descr(d)
+    return np.dtype(d)
--- a/zarr/store.py
+++ b/zarr/store.py
@@ -1,8 +1,9 @@
 # -*- coding: utf-8 -*-
 from __future__ import absolute_import, print_function, division
+
+import json
 import os
 import shutil
-
 
 import numpy as np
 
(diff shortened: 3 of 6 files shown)


alimanfoo:
Thanks Matt. It is tempting to store ``meta`` and ``attrs`` encoded as ``json`` within ``data``, the simplicity of initialising an array store with a single ``MutableMapping`` is appealing.

On the other hand you lose some flexibility in how the ``meta``, ``attrs`` and ``data`` are stored. Also for in-memory arrays there is no point in encoding ``attrs`` and ``meta`` as JSON and storing in ``data`` so this seems like an unnecessary contorsion.

Let's discuss tomorrow.

mrocklin:
My objective here was to be able to hand someone a zip file, a directory, or an s3 bucket and say "here's my zarr array" and have everything be self describing from there.  

The concerns you raise seem valid to me.  Here are options that might address them:

1.  Meta seems fairly nailed down in terms of what gets stored.  I suspect that JSON + custom handling of dtypes would do ok for most applications (though I don't actually have much on the ground experience with zarr use cases, so I'm not sure.)
2.  For attrs we could store the encoding in meta, including an option for `None` which would just be normal in-memory unencoded data.

Alternatively it would be useful to think about other ways to make the array self-describing when given just a location / mutablemapping.  There are probably several good ways to address this issue.

alimanfoo:
I'm not saying this is the best design, but just for information, under the current architecture the ``DirectoryStore`` class takes in the path of a directory, and then has baked in the convention that there will be a JSON file at relative path ``__zmeta__`` holding config metadata and another JSON file at relative path ``__zattr__`` holding user attributes, and all chunk data will be held in a subdirectory at relative path ``__zdata__``. It then sets up three mutable mappings based on this knowledge, relevant code is here: https://github.com/alimanfoo/zarr/blob/refactor/zarr/store.py#L208

By analogy, store classes for ZipFile or S3 could be written in a similar way. E.g., there could be an ``S3Store`` class which takes in an S3 bucket location, and has baked in a convention where metadata is stored in a JSON resource named ``__zmeta__`` within the bucket, attributes are in another JSON resource named ``__zattr__``, and all chunk data are in resources named by chunk indices like ``0.0`` etc. within the same bucket. It could then set up three mutable mappings, with meta and attrs each managed via a separate ``JSONS3ResourceMap`` (analogous to ``JSONFileMap``), and data managed via an ``S3BucketMap`` (analogous to ``DirectoryMap``).

The superclass ``ArrayStore`` currently [accepts the three mutable mappings (meta, data, attrs)](https://github.com/alimanfoo/zarr/blob/refactor/zarr/store.py#L16) and doesn't care what these are or how (if at all) they are persisted. So there is full flexibility to setup stores that follow the above conventions or do something completely different. E.g., you can have a store where the meta is in a Python dict, the attrs are in a JSON file on the local file system, and the data is in an S3 bucket. I'm not sure you'd ever want to do that, but the current architecture has that flexibility.

mrocklin:
What is presented here still allows for that same flexibility if desired, but it checks a standard location in data if only data is provided.

At the end of the day I suspect that you're not going to want to maintain separate `ZipStore`, `S3Store`, `HDFSStore` objects within the zarr codebase.  The testing infrastructure for those stores alone would be larger than the rest of the zarr codebase.  I think that if we can figure out a *default* convention that allows for general support of all data in a single MutableMapping then we remove all pressure on zarr  developers to support exotic storage systems (if we're successful on S3 then someone will want Google Cloud Storage or Ceph).  I'm not suggesting that single-dict be the only option, just that it be available to people who don't build their own `Store` classes.

Making such a convention with a bit less flexibility probably also makes it more attractive to library developers in other languages (though this might be jumping the gun.)

alimanfoo:
Thanks for the explanation, apologies I hadn't looked carefully enough at
your PR. I like the approach in general and also like the idea of not
including any more store implementations within the zarr code base.

On Monday, 25 April 2016, Matthew Rocklin <notifications@github.com> wrote:

> What is presented here still allows for that same flexibility if desired,
> but it checks a standard location in data if only data is provided.
>
> At the end of the day I suspect that you're not going to want to maintain
> separate ZipStore, S3Store, HDFSStore objects within the zarr codebase.
> The testing infrastructure for those stores alone would be larger than the
> rest of the zarr codebase. I think that if we can figure out a *default*
> convention that allows for general support of all data in a single
> MutableMapping then we remove all pressure on zarr developers to support
> exotic storage systems (if we're successful on S3 then someone will want
> Google Cloud Storage or Ceph). I'm not suggesting that single-dict be the
> only option, just that it be available to people who don't build their own
> Store classes.
>
> Making such a convention with a bit less flexibility probably also makes
> it more attractive to library developers in other languages (though this
> might be jumping the gun.)
>
> —
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub
> <https://github.com/alimanfoo/zarr/pull/24#issuecomment-214410358>
>


-- 
Alistair Miles
The Wellcome Trust Centre for Human Genetics
Roosevelt Drive
Oxford
OX3 7BN
United Kingdom
Email: alimanfoo@googlemail.com <alimanfoo@gmail.com>
Web: http://purl.org/net/aliman
Tel: +44 (0)1865 287721

mrocklin:
OK, I've rebased, updated a bit, and pushed.  Feedback welcome.

Regarding compression metadata do you have thoughts/preferences between either of the following layouts for encoding?  Other thoughts?

```python
{'shape': (1000, 1000),
 'chunks': (100, 100),
 'dtype': 'f8',
 'compression': 'blosclz',
 'compression_opts': {'shuffle': True,
                      'level': 5},
 'fill_value': 0}

{'shape': (1000, 1000),
 'chunks': (100, 100),
 'dtype': 'f8',
 'compression': {'name': 'blosclz',
                 'shuffle': True,
                 'level': 5},
 'fill_value': 0}
```

mrocklin:
Other questions:

1.  Thoughts on implementing `DirectoryStore` as mostly just a `DirectoryMap` on top of `ArrayStore`?  The two approaches appear to be similar except in where they keep meta/attr files.
2.  Thoughts on using strings rather than bytestrings for compression names?  E.g. `'blosclz'` rather than `b'blosclz'`.  
3.  Thoughts on Pure Python compatibility if Blosc isn't around?

mrocklin:
I've gone ahead with the second option, pulling all of compression out to a separate option for now.  Happy to do either way.

alimanfoo:
Thanks Matt. Regarding generalising the compression metadata, I was thinking of this for blosc:

```
{'shape': (1000, 1000),
 'chunks': (100, 100),
 'dtype': 'f8',
 'compression': 'blosc',
 'compression_opts': {
     'cname': 'blosclz',
     'clevel': 5,
     'shuffle': 1,
 }
 'fill_value': 0}
```

Another example, using blosc as the primary compressor with zlib as the secondary compressor within blosc, and using the bitshuffle filter:

```
{'shape': (1000, 1000),
 'chunks': (100, 100),
 'dtype': 'f8',
 'compression': 'blosc',
 'compression_opts': {
     'cname': 'zlib',
     'clevel': 5,
     'shuffle': 2,
 }
 'fill_value': 0}
```

If in future someone wanted to use a different compressor, e.g., use zlib directly, then they would do:

```
{'shape': (1000, 1000),
 'chunks': (100, 100),
 'dtype': 'f8',
 'compression': 'zlib',
 'compression_opts': 1,
 'fill_value': 0}
```

...which says use zlib directly to compress chunks, with compression level 1.

One way to think about this would be that the value of the 'compression' key identifies the primary compression method, and the value of the 'compression_opts' key has the argument or arguments that should get passed into the constructor for some class implementing that compression method.

For the next release, zarr would only implement blosc compression. So it would check any existing array metadata and raise an exception if the value of the 'compression' key was anything other than 'blosc'. In a future release we might implement other forms of compression, in which case there would be a registry mapping names like 'blosc', 'zlib', etc., onto classes providing the implementations.

How does that sound?

alimanfoo:
Thinking about the storage architecture, I think I'd like to simplify a bit and create a clear separation between two situations.

In the first situation, a user is "opening" an existing zarr array. I put "opening" in quotes because it's not like opening a file, but the point is they want to work with an array created previously. When doing so, I'm thinking all they should provide is the ``data`` mapping. The implementation would then check for the presence of a 'meta' key within the ``data`` mapping, and raise an exception if not present. I don't think we should allow users to provide any extra metadata when "opening" an existing array, because I think all array metadata should be fixed at the time of creation, other than 'shape'. The implementation would decode the value of the 'meta' key, check for all essential metadata fields and raise an exception if any were missing or invalid.

In the second situation, a user is creating a new zarr array. When creating a new zarr array I'm thinking the user should provide the ``data`` mapping and essential metadata (shape, chunks, dtype, compression, compression_opts, fill_value). There probably also should be an option like 'overwrite' which would determine how any pre-existing keys in ``data`` are handled. If 'overwrite' is True then all keys in ``data`` would be deleted before setting new values for 'meta' and 'attrs'. If 'overwrite' was False and there were any existing keys at all in ``data`` (including 'meta' or any chunks or anything else, i.e., ``data`` was not totally empty) then an exception would be raised. The implementation would normalize all values provided by the user for essential metadata fields, providing defaults if any were not specified, encode as JSON and store in ``data`` via the 'meta' key.

Also, whatever we do with attrs, we need to make sure that any modifications to attrs get flushed back into ``data``.

I realise this is going back on what I said previously about keeping the flexibility to have ``data``, ``meta`` and ``attrs`` in independent mappings, but I think we need to reduce the complexity.

This also might mean we could do away with the store classes completely, and just pass a single ``MutableMapping`` into the ``Array`` constructor, as was your original proposal. Array creation could be handled via a ``create()`` static method on the ``Array`` class, so the constructor was not overloaded to handle both array opening and array creation.

It's hard to see how this would all pan out without writing the code, but any thoughts welcome.

mrocklin:
Will change compression.  I had forgotten that blosc had a cname within it.

mrocklin:
I hadn't thought about the side effects of opening up metadata on opening an existing file.  The changes in this PR could definitely be used to produce some bad effects.

Doing away with `Store` sounds good to me.  I'm always happy whenever one can reduce the number of abstractions within a project.  I think that we will lose a couple things though:

1.  We won't be able to overload `cbytes` for different situations.  I think that this is a small price to pay, cbytes is mostly of interest to us developers and not super useful to me in practice.
2.  We won't be able to overload moving data between keys on a resize event.  `data[x] = data[y]` will always pass bytes through the local process.  

Personally I'm fine losing these.  We could always subclass Array if necessary.

mrocklin:
Can I suggest that we go with the following plan:

1.  I update metadata format in this PR
2.  We merge
3.  One of us (whoever has time first) tries to refactor out `Store`

I suspect that you'll be able to do the Store refactor more easily than me (I haven't actually touched any zarr code outside of store) but I'm also happy to do it if you're occupied.

One this PR gets in, and if we're happy with the overall data layout then I can start creating datasets and benchmarking distributed computations.

alimanfoo:
Sounds good. I'll have some time tomorrow (long train journey to/from Cambridge) so if you let me know when you're done with the metadata format changes I can merge the PR and do some work on storage refactoring and see where it takes us.

mrocklin:
OK, I've pushed in the following:

```python
def dumps(meta):
    meta = dict(
        shape=meta['shape'],
        chunks=meta['chunks'],
        dtype=encode_dtype(meta['dtype']),
        fill_value=meta['fill_value'],
        compression='blosc',
        compression_opts=dict(
            cname=meta['cname'] if PY2 else str(meta['cname'], 'ascii'),
            clevel=meta['clevel'],
            shuffle=meta['shuffle'])
    )

    return json.dumps(meta, indent=4, sort_keys=True)
```

mrocklin:
Let me know if there is anything else I can do

alimanfoo:
Thanks.

PR #28, opened 2016-05-12
ensure bytes for blosc name

mrocklin:
This failed for me on Python 3.  There might be more cases where this is necessary


alimanfoo:
Thanks.

alimanfoo:
On a deeper look these changes shouldn't be necessary, blosc.compname_to_compcode already has an internal check to convert text to bytes. The error was probably also due to Zarr versions being out of date on some workers. I'm going to back out these changes but let me know if you still hit problems..

PR #116, opened 2017-02-20
WIP: Use kenjutsu

jakirkham:
Fixes https://github.com/alimanfoo/zarr/issues/93
Fixes https://github.com/alimanfoo/zarr/issues/78

Use [kenjutsu]( https://github.com/jakirkham/kenjutsu ) to handle slice normalization and shape determination.

Also have extend further to handle a sequence of indices in the slice, but may require some discussion about whether this is ok.

### DIFF
--- a/zarr/core.py
+++ b/zarr/core.py
@@ -6,6 +6,7 @@ import itertools
 
 import numpy as np
 
+from kenjutsu.measure import len_slices
 
 from zarr.util import is_total_slice, normalize_array_selection, \
     get_chunk_range, human_readable_size, normalize_resize_args, \
@@ -449,8 +450,7 @@ class Array(object):
         selection = normalize_array_selection(item, self._shape)
 
         # determine output array shape
-        out_shape = tuple(s.stop - s.start for s in selection
-                          if isinstance(s, slice))
+        out_shape = len_slices(selection)
 
         # setup output array
         out = np.empty(out_shape, dtype=self._dtype, order=self._order)
@@ -571,10 +571,8 @@ class Array(object):
         selection = normalize_array_selection(item, self._shape)
 
         # check value shape
-        expected_shape = tuple(
-            s.stop - s.start for s in selection
-            if isinstance(s, slice)
-        )
+        expected_shape = len_slices(selection)
+
         if np.isscalar(value):
             pass
         elif expected_shape != value.shape:
--- a/zarr/core.py
+++ b/zarr/core.py
@@ -608,6 +608,26 @@ class Array(object):
                              % (str(expected_shape),
                                 str(value.shape)))
 
+        # Find where sequences of indices are.
+        seqs_locs = imap(lambda v: isinstance(v, collections.Sequence), selection)
+        seqs_locs = itertools.compress(irange(len(selection)), seqs_locs)
+        seqs_locs = list(seqs_locs)
+
+        # Set each index individually and return the result.
+        if seqs_locs:
+            assert len(seqs_locs) == 1
+            seq_loc = seqs_locs[0]
+            if not np.isscalar(value):
+                value = value.swapaxes(0, seq_loc)
+            for i, each_selection in enumerate(split_indices(selection)):
+                each_value = value
+                if not np.isscalar(value):
+                    each_value = value[i][None]
+                    each_value = each_value.swapaxes(0, seq_loc)
+                self.__setitem__(each_selection, each_value)
+
+            return
+
         # determine indices of chunks overlapping the selection
         chunk_range = get_chunk_range(selection, self._chunks)
 
--- a/zarr/core.py
+++ b/zarr/core.py
@@ -1,11 +1,23 @@
 # -*- coding: utf-8 -*-
 from __future__ import absolute_import, print_function, division
+import collections
 import operator
 import itertools
+
+try:
+    irange = xrange
+except NameError:
+    irange = range
+
+try:
+    from itertools import map as imap
+except ImportError:
+    imap = map
 
 
 import numpy as np
 
+from kenjutsu.format import split_indices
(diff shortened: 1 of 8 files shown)


zarr/util.py
```diff
@@ -170,29 +166,11 @@ def normalize_array_selection(item, shape):
...
-    elif isinstance(item, slice):
-        item = (item,)
-    elif item == Ellipsis:
-        item = (slice(None),)
-
-    # handle tuple of indices/slices
-    if isinstance(item, tuple):
+    rf_item = reformat_slices(item)
+    if Ellipsis not in rf_item:
+        rf_item += (Ellipsis,)
```
jakirkham: This shouldn't be necessary, but is really a bug on my part. Will fix in `kenjutsu` with PR ( https://github.com/jakirkham/kenjutsu/pull/59 ). In the interim, this shouldn't cause any issues and should be fine even when we fix it.

jakirkham:
Appears there is not enough coverage, but only in the Python 3.6 case. 😕 Not really sure what to do about that though or whether that is something for me to improve or whether it lies outside these changes.

zarr/core.py
```diff
@@ -449,12 +462,29 @@ def __getitem__(self, item):
...
         # setup output array
         out = np.empty(out_shape, dtype=self._dtype, order=self._order)
 
+        # Find where sequences of indices are.
+        seqs_locs = imap(lambda v: isinstance(v, collections.Sequence), selection)
+        seqs_locs = itertools.compress(irange(len(selection)), seqs_locs)
+        seqs_locs = list(seqs_locs)
+
+        if len(seqs_locs) > 1:
+            raise ValueError("Only one sequence of integers is permitted.")
```
jakirkham: Should say "indices" not "integers".

zarr/util.py
jakirkham: Solved in 0.4.1.

alimanfoo:
Hi @jakirkham, thank you for this.

Would mind breaking out just the changes that address #93 into a separate PR? That would help a lot.

Regarding fancy indexing, if I've understood this correctly, I think there may be performance issues. In the worst case, chunks could get decompressed many times over. Also, again in the worst case, if there are lots of indices, then there could be a lot of Python looping. I think we need a solution that ensures each chunk of the array is decompressed at most once, and also which minimises looping. I have some ideas, happy to discuss further.

jakirkham:
> Would mind breaking out just the changes that address #93 into a separate PR? That would help a lot.

So just move the multiple indices into a separate PR. No problem.

> Regarding fancy indexing, if I've understood this correctly, I think there may be performance issues.

Indeed. Was interested first to see if we could get it to work. Plus it would provide some demonstration as to how one might use this functionality to handle these slices. Figured optimization would require some discussion or passing of the baton. Not sure I know enough about Zarr's internals to do this alone.

jakirkham:
I'll probably borrow from this PR and tweak it as it makes for a nice template. However, I have broken out the slice normalization fixes into PR ( https://github.com/alimanfoo/zarr/pull/119 ).

alimanfoo:
Thanks for breaking out #119. 

For fancy indexing, the critical section of code is within ``Array.__getitem__``, particularly the [loop over chunk indices](https://github.com/alimanfoo/zarr/blob/master/zarr/core.py#L458-L490). This is the crux of Zarr, it's the only bit of code that has any real complexity, and took me some time to get right for the simple case of slicing out a contiguous region currently supported. Before entering this loop, the ``get_chunk_range`` function figures out which chunks overlap the selected region. Then the loop iterates over only the chunks in range (important optimisation - iterating over all chunks would be very slow). For each chunk, the code within the loop then figures out which part of the chunk needs to be loaded, and where that should end up in the output array.

For fancy indexing I think it is potentially simpler to handle the fancy part of the index item as a boolean array, rather than a sequence of integer indices. The code within the loop could then be modified to slice this boolean array, to obtain a sub-array of the boolean array which would be appropriate for indexing just that chunk. I.e., the boolean array would get divided up into pieces, with one piece for each chunk along the dimension the boolean array is being applied to. Figuring out the pieces is easy because they are equal sized, determined by the size of the chunks along that dimension. Within the loop there is already the current offset for each dimension, which tells you where you are in the context of the whole array. So this offset and the chunk shape is all that is needed to slice the boolean indexing array for the current chunk. Then the whole chunk can be loaded into a numpy array and the selection applied via numpy indexing. The trick is then to figure out where the data extracted from the current chunk should end up in the output array. I think probably we'd have to keep track of the offsets into the output array too, which means counting how many True values from the boolean array have been handled so far. All of this could I think be done with an integer index array instead of a boolean array, but would be a bit harder as it would require searching the array to find the set of indices that apply to the current chunk, then mapping those indices into the chunk's coordinate system, which I think will be more complex to code and slower.

There may also need to be some modification to the get_chunk_range function to handle a Boolean array, which in the first instance could be dumb and just return all chunks for the indexed dimension as overlapping the selection, but potentially could be optimised so only the chunks where there is some data to extract are returned as in-range.

This is probably all making very little sense, but thought I'd share ideas. This is the basic route I was planning to explore at some point, although have some other priorities I have to work through at the moment.

samiur:
@alimanfoo any update on this? Would love to use this to update an Array using an array of indices.

alimanfoo:
Hi @samiur, apologies I'm snowed under with other work at the moment. TBH I think support for fancy indexing to update the contents of a zarr array may be some way off. I need to think through how to implement limited support for fancy indexing in ``__getitem__`` first which is not entirely trivial.

samiur:
@alimanfoo no worries! Thanks for the help on this.

alimanfoo:
Closing as indexing code is substantially changed via #172. Happy to revisit using kenjutsu in future if it looks like there's a lot of common code.

PR #118, opened 2017-02-20
Use Coveralls

jakirkham:
Fixes https://github.com/alimanfoo/zarr/issues/117

Displays line numbers missed by the test coverage. Reports coverage to Coveralls. Needs to have Coveralls enabled on the repo to work.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -11,8 +11,8 @@ setenv =
     PYTHONHASHSEED = 42
 commands =
     python setup.py build_ext --inplace
-    py27,py34,py35: nosetests -v zarr
-    py36: nosetests -v --with-coverage --cover-erase --cover-min-percentage=100 --cover-package=zarr --with-doctest --doctest-options=+NORMALIZE_WHITESPACE zarr
+    py27,py34,py35: nosetests -v --with-coverage --cover-erase --cover-package=zarr zarr
+    py36: nosetests -v --with-coverage --cover-erase --cover-package=zarr --with-doctest --doctest-options=+NORMALIZE_WHITESPACE zarr
     py36: python -m doctest -o NORMALIZE_WHITESPACE -o ELLIPSIS docs/tutorial.rst docs/spec/v2.rst
     py36: flake8 zarr
     python setup.py bdist_wheel
--- a/tox.ini
+++ b/tox.ini
@@ -16,6 +16,7 @@ commands =
     py36: python -m doctest -o NORMALIZE_WHITESPACE -o ELLIPSIS docs/tutorial.rst docs/spec/v2.rst
     py36: flake8 zarr
     python setup.py bdist_wheel
+    coverage report -m
 deps =
     -rrequirements_dev.txt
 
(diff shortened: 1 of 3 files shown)


jakirkham:
An example of the coverage report from Coveralls can be seen in this [report]( https://coveralls.io/builds/10230984 ).

tox.ini
```diff
@@ -11,11 +11,12 @@ setenv =
     PYTHONHASHSEED = 42
 commands =
     python setup.py build_ext --inplace
-    py27,py34,py35: nosetests -v zarr
-    py36: nosetests -v --with-coverage --cover-erase --cover-min-percentage=100 --cover-package=zarr --with-doctest --doctest-options=+NORMALIZE_WHITESPACE zarr
+    py27: nosetests -v --with-coverage --cover-erase --cover-package=zarr zarr
+    py34,py35,py36: nosetests -v --with-coverage --cover-erase --cover-package=zarr --with-doctest --doctest-options=+NORMALIZE_WHITESPACE zarr
```
alimanfoo: Could you modify this so that the doctests are only run under PY36? There are some dict ordering issues which mean getting consistent doctest runs across all PY3.X versions is hard, I'm happy if doctests are only run for PY36.

jakirkham: No problem.

jakirkham:
Alright, I think this is ready on my end. Is [Coveralls]( http://coveralls.io/ ) already enabled on this repo?

tox.ini
jakirkham: Done.

alimanfoo:
Thanks. Coveralls is enabled, could you add the badge to the README:

```
.. image:: https://coveralls.io/repos/github/alimanfoo/zarr/badge.svg?branch=master
:target: https://coveralls.io/github/alimanfoo/zarr?branch=master
```

jakirkham:
Of course. Added.

tox.ini
jakirkham: Are the `dict` ordering issues related to Python's salting of hashes? If so, I would have thought setting `PYTHONHASHSEED` to a fixed value would solve this issue. Though a more robust solution would probably be to use `OrderedDict`s. That said, given you have set `PYTHONHASHSEED` and it was working before, I'm a bit curious to know if there is more to the story.

jakirkham:
Should add that Coveralls by default is enabled to leave comments and/or statuses on PRs/commits. My recommendation would be to turn off comments, but leave on the status.

Also one can add a coverage threshold. So this would be a good way to enforce 100%. Plus it is very visible as it can show up in the PR status, which makes it easy to see that coverage dropped for a PR and to go back and figure out why coverage dropped.

alimanfoo:
Awesome, thank you!

tox.ini
alimanfoo: Yeah, I don't understand. It's something new in PY36, works fine in both PY34 and PY35. I know dict implementation has changed in PY36, guessing that has something to do with it.

PR #119, opened 2017-02-20
Fix slice normalization and length determination

jakirkham:
Fixes https://github.com/alimanfoo/zarr/issues/93

Uses kenjutsu to handle a variety of tricky cases with Ellipsis, negative step sizes, negative indices, and standard cases as well. This simplifies the code, avoids a variety of bugs. Also this should avoid the need for normalizing each axis, but have not yet cut that code out.

### DIFF
--- a/zarr/util.py
+++ b/zarr/util.py
@@ -114,23 +114,9 @@ def is_total_slice(item, shape):
     given `shape`. Used to optimize __setitem__ operations on the Chunk
     class."""
 
-    # N.B., assume shape is normalized
+    rf_item = normalize_array_selection(item, shape)
 
-    if item == Ellipsis:
-        return True
-    if item == slice(None):
-        return True
-    if isinstance(item, slice):
-        item = item,
-    if isinstance(item, tuple):
-        return all(
-            (isinstance(s, slice) and
-                ((s == slice(None)) or
-                 ((s.stop - s.start == l) and (s.step in [1, None]))))
-            for s, l in zip(item, shape)
-        )
-    else:
-        raise TypeError('expected slice or tuple of slices, found %r' % item)
+    return len_slices(rf_item) == shape
 
 
 def normalize_axis_selection(item, l):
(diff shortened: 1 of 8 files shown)


zarr/util.py
```diff
@@ -111,23 +113,9 @@ def is_total_slice(item, shape):
...
-            for s, l in zip(item, shape)
-        )
-    else:
-        raise TypeError('expected slice or tuple of slices, found %r' % item)
+    rf_item = normalize_array_selection(item, shape)
+
+    return len_slices(rf_item) == shape
 
 
 def normalize_axis_selection(item, l):
```
jakirkham: This function is no longer needed. That said, I can leave it, change it to use `kenjutsu` as well, or just remove it. Please let me know what you think.

jakirkham:
Should add that I saw a lot of constraints on what kind of `slice`s are permissible (e.g. step size). Am happy to add them back in.

Though some of them like an overly large `slice` are technically permissible both in vanilla Python and NumPy. Not to mention when provided with a `slice` and a shape, `kenjutsu` cuts these `slice`s down to a proper size. So it shouldn't cost anything extra to just use them after they are reformatted/normalized.

zarr/util.py
jakirkham: Went ahead and tried to restructure things to use this function anyways for validation purposes. That way we can leverage all of the current tests in the first pass.

jakirkham:
Alright, I think this is ready for review.

alimanfoo:
Thank you, very helpful, and very nice that the new ellipsis tests are passing. 

I'd like to mull this over for a bit. I've spent some time looking at the `reformat_slices` and `reformat_slice` functions in kenjutsu where the goodness happens. Mostly I can see what's happening but there are some bits I need to spend more time on.

I think I may bring simplified versions of these functions into Zarr as an interim solution. That way I can go through every line of code and get acquainted. This is such a critical piece of code I think I need to understand what every line is doing. Also I think it will help to have this functionality inside Zarr when working on fancy indexing, as fancy indexing will require some changes to these functions. But then ultimately when the dust settles from that, maybe consider factoring back out to kenjutsu.

alimanfoo:
Closing as superseded by indexing work #172.

PR #120, opened 2017-02-21
Fix representation of AsType's dtypes

jakirkham:
Use string representations of `AsType`'s `dtype`s instead of allowing the `dtype`s to be converted automatically. This matches more closely with what `Delta` is doing in its representation.

### DIFF
--- a/zarr/codecs.py
+++ b/zarr/codecs.py
@@ -557,10 +557,10 @@ class AsType(Codec):
 
     def __repr__(self):
         return (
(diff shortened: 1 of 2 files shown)


jakirkham:
I know we are working on moving this to Numcodecs, but I'd still very much like to see this merged if that is ok.

jakirkham:
Thanks @alimanfoo. Much appreciated.

PR #122, opened 2017-02-22
Add visitor pattern methods

jakirkham:
Fixes https://github.com/alimanfoo/zarr/issues/92

Provides `visit` and `visititem` that behave identically to their h5py counterparts. This should make it easier to traverse the full hierarchy of a Zarr group. Also should make it easier for users with h5py code to try out using Zarr with fewer changes.

### DIFF
--- a/zarr/hierarchy.py
+++ b/zarr/hierarchy.py
@@ -57,6 +57,7 @@ class Group(MutableMapping):
     array_keys
     arrays
     visit
+    visitkeys
     visitvalues
     visititems
     create_group
@@ -496,6 +497,12 @@ class Group(MutableMapping):
 
         base_len = len(self.name)
         return self.visitvalues(lambda o: func(o.name[base_len:].lstrip("/")))
+
+    def visitkeys(self, func):
+        """An alias for :py:meth:`~Group.visit`.
+        """
+
+        return self.visit(func)
 
     def visititems(self, func):
         """Run ``func`` on each object's path and the object itself.
--- a/zarr/hierarchy.py
+++ b/zarr/hierarchy.py
@@ -57,6 +57,7 @@ class Group(MutableMapping):
     array_keys
     arrays
     visit
+    visitvalues
     visititems
     create_group
     require_group
@@ -417,6 +418,54 @@ class Group(MutableMapping):
                                  chunk_store=self._chunk_store,
                                  synchronizer=self._synchronizer)
 
+    def visitvalues(self, func):
+        """Run ``func`` on each object.
+
+        Note: If ``func`` returns ``None`` (or doesn't return),
+              iteration continues. However, if ``func`` returns
+              anything else, it ceases and returns that value.
+
+        Examples
+        --------
+        >>> import zarr
+        >>> g1 = zarr.group()
+        >>> g2 = g1.create_group('foo')
+        >>> g3 = g1.create_group('bar')
+        >>> g4 = g3.create_group('baz')
+        >>> g5 = g3.create_group('quux')
+        >>> def print_visitor(obj):
+        ...     print(obj)
+        >>> g1.visitvalues(print_visitor)
+        Group(/bar, 2)
+          groups: 2; baz, quux
+          store: DictStore
+        Group(/bar/baz, 0)
+          store: DictStore
+        Group(/bar/quux, 0)
+          store: DictStore
+        Group(/foo, 0)
+          store: DictStore
+        >>> g3.visitvalues(print_visitor)
+        Group(/bar/baz, 0)
+          store: DictStore
+        Group(/bar/quux, 0)
+          store: DictStore
+
+        """
+
+        def _visit(obj):
+            yield obj
+
+            keys = sorted(getattr(obj, "keys", lambda : [])())
+            for each_key in keys:
+                for each_obj in _visit(obj[each_key]):
+                    yield each_obj
+
+        for each_obj in islice(_visit(self), 1, None):
+            value = func(each_obj)
+            if value is not None:
+                return value
+
     def visit(self, func):
         """Run ``func`` on each object's path.
 
@@ -445,19 +494,8 @@ class Group(MutableMapping):
 
         """
 
-        def _visit(obj):
-            yield obj
-
-            keys = sorted(getattr(obj, "keys", lambda : [])())
-            for each_key in keys:
-                for each_obj in _visit(obj[each_key]):
-                    yield each_obj
-
         base_len = len(self.name)
-        for each_obj in islice(_visit(self), 1, None):
-            value = func(each_obj.name[base_len:].lstrip("/"))
-            if value is not None:
-                return value
+        return self.visitvalues(lambda o: func(o.name[base_len:].lstrip("/")))
 
     def visititems(self, func):
         """Run ``func`` on each object's path and the object itself.
@@ -494,19 +532,8 @@ class Group(MutableMapping):
 
         """
 
-        def _visit(obj):
-            yield obj
-
-            keys = sorted(getattr(obj, "keys", lambda : [])())
-            for each_key in keys:
-                for each_obj in _visit(obj[each_key]):
-                    yield each_obj
-
         base_len = len(self.name)
-        for each_obj in islice(_visit(self), 1, None):
-            value = func(each_obj.name[base_len:].lstrip("/"), each_obj)
-            if value is not None:
-                return value
+        return self.visitvalues(lambda o: func(o.name[base_len:].lstrip("/"), o))
 
     def _write_op(self, f, *args, **kwargs):
 
--- a/zarr/hierarchy.py
+++ b/zarr/hierarchy.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 from __future__ import absolute_import, print_function, division
 from collections import MutableMapping
+from itertools import islice
 
 
 import numpy as np
@@ -55,6 +56,8 @@ class Group(MutableMapping):
     groups
     array_keys
     arrays
+    visit
+    visititems
     create_group
     require_group
     create_groups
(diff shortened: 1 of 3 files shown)


jakirkham:
Only has doctests ATM. Could use some unit tests as well. That said, please let me know what you think.

zarr/hierarchy.py
```diff
@@ -414,6 +416,78 @@ def arrays(self):
                                  chunk_store=self._chunk_store,
                                  synchronizer=self._synchronizer)
 
+    def visit(self, func):
+        """Run callable on each object name.
+
+        Examples
+        --------
+        >>> from __future__ import print_function, unicode_literals
```
alimanfoo: I don't think this is needed in the example. For Zarr docstrings are standardised to PY3(6).

zarr/hierarchy.py
```diff
@@ -414,6 +416,78 @@ def arrays(self):
...
+        >>> import zarr
+        >>> g1 = zarr.group()
+        >>> g2 = g1.create_group('foo')
+        >>> g3 = g1.create_group('bar')
+        >>> d1 = g1.create_dataset('baz', shape=100, chunks=10, fill_value=0)
+        >>> d2 = g1.create_dataset('quux', shape=200, chunks=20, fill_value=0)
```
alimanfoo: Maybe use 'foo/quux' as the dataset name or something like that, to show what happens when you have a hierarchy?

zarr/hierarchy.py
```diff
@@ -414,6 +416,78 @@ def arrays(self):
...
```
alimanfoo: Again suggest to have an example showing what happens with a hierarchy, e.g., 'foo/quux'.

zarr/hierarchy.py
```diff
@@ -414,6 +416,78 @@ def arrays(self):
                                  chunk_store=self._chunk_store,
                                  synchronizer=self._synchronizer)
 
+    def visit(self, func):
+        """Run callable on each object name.
```
alimanfoo: Suggest to replace ``callable`` with `` `func` ``. Also maybe worth adding a note about how the return value is handled, e.g., from h5py source: "Returning None continues iteration; returning anything else aborts iteration and returns that value."

zarr/hierarchy.py
```diff
@@ -414,6 +416,78 @@ def arrays(self):
...
+            for each_key in keys:
+                for value in _visit(obj[each_key], func=func):
+                    yield value
+
+        for value in _visit(self):
+            if value is not None:
+                return value
+
+    def visititems(self, func):
+        """Run callable on each object name and object.
```
alimanfoo: As for `visit()`, suggest to replace `callable` with `` `func` `` and adding a note about how the return value is handled.

zarr/hierarchy.py
```diff
@@ -414,6 +416,78 @@ def arrays(self):
...
+
+        Examples
+        --------
+        >>> from __future__ import print_function, unicode_literals
```
alimanfoo: Not needed I don't think.

zarr/hierarchy.py
```diff
@@ -414,6 +416,78 @@ def arrays(self):
...
+        baz
+        foo
+        quux
+
+        """
+
+        def _visit(obj, func=func):
+            keys = sorted(getattr(obj, "keys", lambda : [])())
+            for each_key in keys:
+                for value in _visit(obj[each_key], func=func):
```
alimanfoo: Think there's a call to ``func(each_key)`` missing here?

zarr/hierarchy.py
```diff
@@ -414,6 +416,78 @@ def arrays(self):
...
+        ('quux', Array(/quux, (200,), float64, chunks=(20,), order=C)
+          nbytes: 1.6K; nbytes_stored: 290; ratio: 5.5; initialized: 0/10
+          compressor: Blosc(cname='lz4', clevel=5, shuffle=1)
+          store: DictStore)
+
+        """
+
+        def _visit(obj, func=func):
+            keys = sorted(getattr(obj, "keys", lambda : [])())
+            for each_key in keys:
```
alimanfoo: Missing call to ``func(each_key, obj[each_key])``?

zarr/hierarchy.py
jakirkham: Yeah, I figured it couldn't hurt.

zarr/hierarchy.py
jakirkham: We can do that. Was thinking about doing something similar. Though the tests explore this as well.

May also drop datasets (or at least disable compression) as the compression levels seem to vary, which would make the doctest annoying to keep.

zarr/hierarchy.py
jakirkham: Note to self: https://github.com/alimanfoo/zarr/pull/122#discussion_r102596944

zarr/hierarchy.py
jakirkham: Same as above.

zarr/hierarchy.py
jakirkham: xref: https://github.com/alimanfoo/zarr/pull/122#discussion_r102634014

zarr/hierarchy.py
jakirkham: Yeah, I was playing with stuff and pushed a bad commit. Unfortunately have been AFK until now. Should be fixed.

zarr/hierarchy.py
jakirkham: Same as above.

jakirkham:
Ok, addressed comments and fixed bugs. This is ok on my end.

alimanfoo:
Something weird happening with CI, maybe related to install of appdirs? Try upgrading to ``appdirs==1.4.1`` in ``requirements_dev.txt``? Or maybe remove ``appdirs`` altogether from ``requirements_dev.txt`` as it's probably not important to pin it, and let latest get brought in by whichever package requires it?

jakirkham:
Yeah, this cropped up with the release of 1.4.1. Have raised upstream to see if they have any idea.

xref: https://github.com/ActiveState/appdirs/issues/89

jakirkham:
Put together PR ( https://github.com/alimanfoo/zarr/pull/124 ) to try and fix the CI issues.

zarr/hierarchy.py
```diff
@@ -414,6 +417,89 @@ def arrays(self):
...
+    def visititems(self, func):
+        """Run ``func`` on each object's path and the object itself.
+
+        Note: If ``func`` returns ``None`` (or doesn't return),
+              iteration continues. However, if ``func`` returns
+              anything else, it ceases and returns that value.
+
+        Examples
+        --------
+        >>> from __future__ import print_function, unicode_literals
```
alimanfoo: Same as above.

zarr/hierarchy.py
```diff
@@ -414,6 +417,89 @@ def arrays(self):
...
+        ('bar/baz', Group(/bar/baz, 0)
+          store: DictStore)
+        ('bar/quux', Group(/bar/quux, 0)
+          store: DictStore)
+        ('foo', Group(/foo, 0)
+          store: DictStore)
+
+        """
+
+        def _visit(obj):
```
alimanfoo: This is now a duplicate of _visit above. Suggest to factor out into a single private function.

jakirkham: Yep, was actually thinking of refactoring it differently though. I think we could have a `visitvalues` function that both `visit` and `visititems` use. Have a WIP branch for this, but it isn't quite done yet.

zarr/hierarchy.py
```diff
@@ -414,6 +417,89 @@ def arrays(self):
...
+    def visit(self, func):
+        """Run ``func`` on each object's path.
+
+        Note: If ``func`` returns ``None`` (or doesn't return),
+              iteration continues. However, if ``func`` returns
+              anything else, it ceases and returns that value.
+
+        Examples
+        --------
+        >>> from __future__ import print_function, unicode_literals
```
jakirkham: Sure, done.

zarr/hierarchy.py
jakirkham: Also done.

zarr/hierarchy.py
jakirkham: Have pushed a refactoring using `visitvalues`, which seems to work nicely.

jakirkham:
Also made changes so that all of these will show up in the docs. Built the docs locally to make sure.

alimanfoo:
Thanks for the changes, looks good.

So I'm just reading the [h5py docs](http://docs.h5py.org/en/latest/high/group.html#Group.visit), and apparently the name passed into the visiting function "will be the name of the object relative to the current group". As currently coded, this PR will pass the object path, which is the name relative to the root group. I guess if we are aiming for h5py compatibility where possible then we need to follow the same behaviour? Unless you have a reason to do it differently?

jakirkham:
Just to make sure we are on the same page, by current group, I believe they mean the group we called `visit*` on. At least that is what the h5py's `visit*` methods seem to do.

That's a good point. Thanks for raising it. This was definitely an oversight on my part. Not intentional.

Do you have any suggestions for how we should do this? If not, I'm thinking we could just strip out that first `x` characters that correspond to the parent group path or replace the first match.

alimanfoo:
No problem, yes that's how I read the docs, name is supposed to be relative to the group we called ``visit*`` on. I think it would be fine to do something like slice the ``.path`` property of the object being visited, using the length of the ``.path`` property on the group we called ``visit*`` on as the slice start coordinate.

jakirkham:
Alright, I have now updated the `visit` and `visititems` functions to provide relative paths. Added an example of this to the documentation. Also included a bunch of related unit tests. Hopefully that addresses it.

alimanfoo:
Great, thank you!

jakirkham:
Thanks for the good review.

Issue #74, opened 2016-09-21
Use numcodecs


alimanfoo:
The codecs have been factored out into a new package https://github.com/alimanfoo/numcodecs. This means the zarr.codecs module could be removed and replaced by adding numcodecs as a runtime dependency.

alimanfoo:
Resolved via #139

PR #123, opened 2017-02-23
Support LZMA on Python 2 via backports.lzma

jakirkham:
Backport(?) of PR ( https://github.com/alimanfoo/numcodecs/pull/13 ).

Tries to use `backports.lzma` as a fallback if `lzma` is not found (e.g. Python 2). Adds testing for it on Travis CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,12 @@
 language: python
 
+sudo: false
+
+addons:
+  apt:
+    packages:
+      - liblzma-dev
+
 python:
   - 2.7
   - 3.4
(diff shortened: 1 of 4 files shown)


jakirkham:
Appears to be [working]( https://travis-ci.org/alimanfoo/zarr/jobs/204482130#L529-L530 ).

alimanfoo:
Happy to include if would be helpful to you, but I will try to find some time to do #74, which I definitely want to do before the next Zarr release.

jakirkham:
Yeah, I've been doing development builds of Zarr for my uses. So having this included already instead of having to patch it in would be preferable.

jakirkham:
Thanks.

PR #126, opened 2017-02-24
Unpin setuptools requirements

jakirkham:
Fixes https://github.com/alimanfoo/zarr/issues/125

As all of these are now requirements of `setuptools` instead of being vendored as they were before, pinning them to anything other than the latest version has adverse consequences. So this drops all of these pins. More details about the problem and possible fixes in issue ( https://github.com/pypa/pip/issues/4264 ).

### DIFF
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -1,5 +1,5 @@
 # pinned dependencies for the development and CI environment
-appdirs==1.4.1
+appdirs
 args==0.1.0
 clint==0.5.1
 coverage==4.3.4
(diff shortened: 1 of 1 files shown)


alimanfoo:
FWIW if these are all requirements of setuptools then we could remove them completely from requirements_dev.txt. Happy either way.

jakirkham:
We may want to pin them in the future so we can have some reproducibility in our lives again. Leaving them present, but unpinned can serve as a reminder.

That said, if you feel strongly about dropping them, I'm willing to drop them entirely.

alimanfoo:
Good point, merged as-is.

PR #127, opened 2017-02-24
Remove .ipynb_checkpoints

jakirkham:
This directory only exists for Jupyter Notebooks' to store manual saves separately from automatic saves. There shouldn't be any need to track this in git.


alimanfoo:
Thanks for this, nice to clean up.

PR #148, opened 2017-04-26
repr changes

### LINKED ISSUE #83: Bring back "zarr" into repr
In a mixed context it's not clear that a zarr array is from zarr, put "zarr" back into array and group reprs somehow.

### LINKED ISSUE #115: ENH: read_only in repr
Consider adding read_only status to repr as a reminder if forget to open read_only.

### LINKED ISSUE #132: Group and Array repr performance
Generating a repr for a group can currently be slow, for the quirky reason that the repr needs to decide if the ``chunk_store`` is different to the ``store`` in order to generate an extra line showing the chunk store if different, and this triggers a complete comparison of store contents if the store is a shelve. E.g.:

```
In [37]: %lprun -f zarr.hierarchy.Group.__repr__ repr(g)
Timer unit: 1e-06 s

Total time: 5.60861 s
File: /home/aliman/miniconda3/envs/biipy_20170126_py35/lib/python3.5/site-packages/zarr/hierarchy.py
Function: __repr__ at line 194

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
   194                                               def __repr__(self):
   195                                           
   196                                                   # main line
   197         1           11     11.0      0.0          r = '%s(' % type(self).__name__
   198         1           28     28.0      0.0          r += self.name + ', '
   199         1       276958 276958.0      4.9          r += str(len(self))
   200         1            2      2.0      0.0          r += ')'
   201                                           
   202                                                   # members
   203         1       244629 244629.0      4.4          array_keys = list(self.array_keys())
   204         1            2      2.0      0.0          if array_keys:
   205         1            1      1.0      0.0              arrays_line = '\n  arrays: %s; %s' % \
   206         1            8      8.0      0.0                  (len(array_keys), ', '.join(array_keys))
   207         1            1      1.0      0.0              if len(arrays_line) > 80:
   208         1            1      1.0      0.0                  arrays_line = arrays_line[:77] + '...'
   209         1            1      1.0      0.0              r += arrays_line
   210         1       240623 240623.0      4.3          group_keys = list(self.group_keys())
   211         1            3      3.0      0.0          if group_keys:
   212                                                       groups_line = '\n  groups: %s; %s' % \
   213                                                           (len(group_keys), ', '.join(group_keys))
   214                                                       if len(groups_line) > 80:
   215                                                           groups_line = groups_line[:77] + '...'
   216                                                       r += groups_line
   217                                           
   218                                                   # storage and synchronizer classes
   219         1           10     10.0      0.0          r += '\n  store: %s' % type(self.store).__name__
   220         1      4846320 4846320.0     86.4          if self.store != self.chunk_store:
   221                                                       r += '; chunk_store: %s' % type(self.chunk_store).__name__
   222         1           13     13.0      0.0          if self.synchronizer is not None:
   223                                                       r += '; synchronizer: %s' % type(self.synchronizer).__name__
   224                                           
   225         1            1      1.0      0.0          return r
```

There should be a way to avoid this comparison, at least in the case where ``chunk_store`` is None when the Group is instantiated, because then it is known there is no separate chunk store.

alimanfoo:
This PR has some work on array and group repr, resolves #83, resolves #115, resolves #132.


alimanfoo:
I've simplified the array and group repr to a minimal output consistent with h5py. To still get diagnostic information on an array when you want it there is a new info() method on the array class. E.g.:

```python
>>> import zarr
>>> g = zarr.group()
>>> g
<zarr group '/' (0 arrays, 0 groups)>
>>> g.create_group('foo/bar')
<zarr group '/foo/bar' (0 arrays, 0 groups)>
>>> g
<zarr group '/' (0 arrays, 1 groups)>
>>> g.create_dataset('qux', shape=1000, chunks=100, dtype='i4', filters=[zarr.Delta('i4')], compression='gzip')
<zarr array '/qux': shape (1000,), type '<i4'>
>>> g
<zarr group '/' (1 arrays, 1 groups)>
>>> z = g['qux']
>>> z
<zarr array '/qux': shape (1000,), type '<i4'>
>>> print(z.info())
Name                   : /qux
Type                   : zarr.core.Array
Data type              : int32
Shape                  : (1000,)
Chunk shape            : (100,)
Order                  : C
Filter [0]             : Delta(dtype='<i4')
Compressor             : Zlib(level=1)
Store type             : zarr.storage.DictStore
No. bytes              : 4000 (3.9K)
No. bytes stored       : 350
Storage ratio          : 11.43
No. chunks initialized : 0/10

>>> z[:] = 42
>>> print(z.info())
Name                   : /qux
Type                   : zarr.core.Array
Data type              : int32
Shape                  : (1000,)
Chunk shape            : (100,)
Order                  : C
Filter [0]             : Delta(dtype='<i4')
Compressor             : Zlib(level=1)
Store type             : zarr.storage.DictStore
No. bytes              : 4000 (3.9K)
No. bytes stored       : 490
Storage ratio          : 8.16
No. chunks initialized : 10/10

```

cc @jakirkham, comments welcome.

alimanfoo:
I've made some further changes to also simplify the group repr and avoid any potentially expensive computations when generating reprs. New examples:

```python
    >>> import zarr
    >>> root = zarr.group()
    >>> foo = root.create_group('foo')
    >>> bar = foo.zeros('bar', shape=1000000, chunks=100000)
    >>> bar[:] = 42
    >>> root
    <zarr.hierarchy.Group '/'>
    >>> root.info
    Name        : /
    Type        : zarr.hierarchy.Group
    Read-only   : False
    Store type  : zarr.storage.DictStore
    No. members : 1
    No. arrays  : 0
    No. groups  : 1
    Groups      : foo

    >>> foo
    <zarr.hierarchy.Group '/foo'>
    >>> foo.info
    Name        : /foo
    Type        : zarr.hierarchy.Group
    Read-only   : False
    Store type  : zarr.storage.DictStore
    No. members : 1
    No. arrays  : 1
    No. groups  : 0
    Arrays      : bar

    >>> bar
    <zarr.core.Array '/foo/bar' (1000000,) float64>
    >>> bar.info
    Name               : /foo/bar
    Type               : zarr.core.Array
    Data type          : float64
    Shape              : (1000000,)
    Chunk shape        : (100000,)
    Order              : C
    Read-only          : False
    Compressor         : Blosc(cname='lz4', clevel=5, shuffle=SHUFFLE, blocksize=0)
    Store type         : zarr.storage.DictStore
    No. bytes          : 8000000 (7.6M)
    No. bytes stored   : 38482 (37.6K)
    Storage ratio      : 207.9
    Chunks initialized : 10/10

```

alimanfoo:
Test coverage should be back up. Merging soon.

jakirkham:
Sorry coming to this late and have only skimmed the changes. Generally I like the new repr. It's compact, simple, and clean. Adding everything else to `info` is nice. Also like that `info` is much more orderly.

The only comment I might have is whether we want to be listing groups through `info` or whether we should be using the tree-work to handle that somehow.

alimanfoo:
No problem, thanks for comments. FWIW I think it's worth including a listing of groups and arrays in group info, it's potentially more compact than viewing as a tree, and so provides a convenient way to get an overview of first level members. The tree view is then useful to browse beyond immediate children.

tox.ini
```diff
@@ -10,13 +10,11 @@ envlist = py27, py34, py35, py36, docs
 setenv =
     PYTHONHASHSEED = 42
 commands =
-    python setup.py build_ext --inplace
-    py27,py34,py35: nosetests -v --with-coverage --cover-erase --cover-package=zarr zarr
+    py27,py34,py35: nosetests -v zarr
```
jakirkham: So I missed this change the other day. Part of the problem with dropping coverage is we may miss Python 2 specific branches. Am suggesting we revert this one line in PR ( https://github.com/alimanfoo/zarr/pull/169 ). Though it would be good to know the motivation for this change in case there is a better solution.

PR #153, opened 2017-09-05
Updated cpuinfo.py so that it can use the saved dmesg.boot information.

newt0311:
In long running systems, the initial dmesg entries with the CPU info often get wiped out. This information is still saved in the dmesg.boot file (this is just dmesg shortly after boot). This change will get the cpu info script to check the dmesg.boot file as well.


alimanfoo:
Thanks for contributing this. Since codecs were factored out to numcodecs package, cpuinfo is not needed here, but is needed in numcodecs, so will merge these changes there.

PR #160, opened 2017-10-24
Zero length dimensions

### LINKED ISSUE #150: Zarr can not handle 0 dimension like numpy does
In certain cases, numpy can return a dimension that is 0.  This is different from reducing the dimension so the dimension is missing.  Therefore, in numpy you can have shape like (1000, 0, 0), which is actually empty data-wise but it is internally consistent.  In such case zarr seems to fail:

```
In [68]: import zarr
In [69]: arr = zarr.empty((1000, 0, 0), dtype='float64')
/.../site-packages/zarr/util.py:48: RuntimeWarning: divide by zero encountered in log10
  target_size = CHUNK_BASE * (2**np.log10(dset_size/(1024.*1024)))
```

where it would work in numpy:

```
In [70]: import numpy
In [71]: arr = numpy.empty((1000, 0, 0), dtype='float64')
In [72]: print arr
[]
In [73]: print arr.shape
(1000, 0, 0)
```

alimanfoo:
This PR focuses on support for arrays with one or more zero-length dimensions. Some changes from #154 have been manually merged. Also some further tests to verify slicing behaviour is consistent with numpy. Resolves #150.

### DIFF
--- a/zarr/tests/test_core.py
+++ b/zarr/tests/test_core.py
@@ -514,6 +514,74 @@ class TestArray(unittest.TestCase):
         assert_array_equal(np.take(a, indices, axis=1),
                            np.take(a, zi, axis=1))
 
+    def test_0len_dim_1d(self):
+        # Test behavioud for 1D array with zero-length dimension.
+
+        z = self.create_array(shape=0, fill_value=0)
+        a = np.zeros(0)
+        eq(a.ndim, z.ndim)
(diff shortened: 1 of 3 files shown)


alimanfoo:
I think this is good to go. Will merge soon if no comments.

alimanfoo:
Thanks @jakirkham.

PR #161, opened 2017-10-24
0-dimensional arrays

alimanfoo:
This PR provides tests and implementation of support for 0-dimensional arrays. This builds off work done in #154, however I needed to add special case implementations of ``__getitem__`` and ``__setitem__`` to the ``Array`` class to get indexing working with both empty tuple and ellipsis.

### DIFF
--- a/zarr/core.py
+++ b/zarr/core.py
@@ -613,11 +613,15 @@ class Array(object):
         if item not in ((), Ellipsis):
             raise IndexError('too many indices for array')
 
-        # obtain key for chunk storage
-        ckey = self._chunk_key((0,))
-
         # setup data to store
         arr = np.asarray(value, dtype=self._dtype)
+
+        # check value
+        if arr.shape != ():
+            raise ValueError('bad value; expected scalar, found %r' % value)
+
+        # obtain key for chunk storage
+        ckey = self._chunk_key((0,))
 
         # encode and store
         cdata = self._encode_chunk(arr)
@@ -637,7 +641,7 @@ class Array(object):
         if np.isscalar(value):
             pass
         elif expected_shape != value.shape:
-            raise ValueError('value has wrong shape, expecting %s, found %s'
+            raise ValueError('value has wrong shape; expected %s, found %s'
                              % (str(expected_shape),
                                 str(value.shape)))
 
--- a/zarr/core.py
+++ b/zarr/core.py
@@ -268,7 +268,7 @@ class Array(object):
 
     @property
     def _size(self):
-        return reduce(operator.mul, self._shape)
+        return reduce(operator.mul, self._shape, 1)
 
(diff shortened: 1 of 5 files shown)


alimanfoo:
OK, I think this is good to go.

zarr/core.py
```diff
@@ -443,6 +446,44 @@ def __getitem__(self, item):
...
+        except KeyError:
+
+            # chunk not initialized
+            out = np.empty((), dtype=self._dtype)
+            if self._fill_value is not None:
+                out.fill(self._fill_value)
+
+        else:
+
+            out = self._decode_chunk(cdata)
```
jakirkham: Minor point. This seems a bit spaced out. Not sure if that was intentional or not.

alimanfoo:
Behaviour should be consistent with h5py and numpy, e.g.:

```python
>>> import zarr
>>> z = zarr.zeros(())
>>> z
<zarr.core.Array () float64>
>>> z[...]
array(0.0)
>>> z[()]
0.0
>>> z[...] = 42
>>> z[...]
array(42.0)
>>> z[()] = 43
>>> z[()]
43.0
>>> z[:]
  File "/home/aliman/src/github/alimanfoo/zarr/zarr/core.py", line 451, in __getitem__
    return self._getitem_zd(item)
  File "/home/aliman/src/github/alimanfoo/zarr/zarr/core.py", line 460, in _getitem_zd
    raise IndexError('too many indices for array')
IndexError: too many indices for array
```

zarr/tests/test_core.py
```diff
@@ -587,6 +589,50 @@ def test_0len_dim_2d(self):
...
+        eq(a.ndim, z.ndim)
+        eq(a.shape, z.shape)
+        eq(a.size, z.size)
+        eq(a.dtype, z.dtype)
+        with assert_raises(TypeError):
+            len(z)
+        eq((), z.chunks)
+        eq(1, z.nchunks)
+        # compressor always None - no point in compressing a scalar value
+        assert_is_none(z.compressor)
```
jakirkham: Would it be worth checking `nbytes` as well or is that too far afield.

jakirkham:
Does `_cdata_shape` need to be updated?

alimanfoo:
Also can now set scalar on group:

```python
>>> import zarr
>>> g = zarr.group()
>>> g['foo'] = 42
>>> g['foo']
<zarr.core.Array '/foo' () int64>
>>> g['foo'][()]
42
>>> g['bar/baz/qux'] = 123.4
>>> g['bar/baz/qux']
<zarr.core.Array '/bar/baz/qux' () float64>
>>> g['bar/baz/qux'][()]
123.40000000000001
```

jakirkham:
LGTM

PR #163, opened 2017-10-25
Add unicode_type to compat

jakirkham:
Basically provide something that always points to the unicode string type on both Python 2 and Python 3.


jakirkham:
Is this one alright, @alimanfoo?

alimanfoo:
You could just use ``text_type`` which is already there?

jakirkham:
Yep. Sorry for the noise.

PR #169, opened 2017-10-26
Collect coverage on all Pythons

jakirkham:
Seems that Python 2 was getting left out of the mix. However sometimes code needs to be special cased for Python 2. As these branches were not being counted if tested on Python 2, it caused the coverage to dip superficially. Hence this changes tox to track coverage on all Python versions and report the results.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -10,11 +10,11 @@ envlist = py27, py34, py35, py36, docs
 setenv =
     PYTHONHASHSEED = 42
 commands =
-    py27,py34,py35: nosetests -v zarr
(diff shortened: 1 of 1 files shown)


alimanfoo:
Thanks, I've made a PR #171 which re-instates coverage for all pythons, and also adds a pragma so coverage can be excluded based on which major python version. Previously I had just focused on PY3 and put "pragma no cover" for all PY2 code lines, hence I thought there was no point in including PY2 in coveralls, however this made me look for a way to handle exclusions better.

jakirkham:
Cool, your PR seems fine to me. Adding pragmas for Python versions is a nice touch as well.

PR #171, opened 2017-10-26
Coverage and Python versions

alimanfoo:
This PR re-instates coverage for all Python versions (#169). Also a mechanism is added for excluding lines from coverage analysis based on which Python major version.


jakirkham:
Thanks. LGTM.

PR #184, opened 2017-11-15
Use jstree

alimanfoo:
This PR modifies the tree HTML representation to use jstree. Also adds a "level" parameter to tree() to allow limiting depth to which tree is printed.

### DIFF
--- a/notebooks/repr_tree.ipynb
+++ b/notebooks/repr_tree.ipynb
@@ -6,13 +6,9 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "import zarr\n",
-    "\n",
-    "# Python 2/3 trick\n",
-    "try:\n",
-    "    unicode\n",
-    "except NameError:\n",
-    "    unicode = str"
+    "import sys\n",
+    "sys.path.insert(0, '..')\n",
+    "import zarr"
    ]
   },
   {
@@ -22,10 +18,14 @@
    "outputs": [],
    "source": [
     "g1 = zarr.group()\n",
+    "g2 = g1.create_group('foo')\n",
     "g3 = g1.create_group('bar')\n",
     "g3.create_group('baz')\n",
+    "g3.create_dataset('xxx', shape=100)\n",
+    "g3.create_dataset('yyy', shape=(100, 100), dtype='i4')\n",
     "g5 = g3.create_group('quux')\n",
-    "g5.create_dataset('baz', shape=100, chunks=10)\n",
+    "g5.create_dataset('aaa', shape=100)\n",
+    "g5.create_dataset('bbb', shape=(100, 100), dtype='i4')\n",
     "g7 = g3.create_group('zoo')"
    ]
   },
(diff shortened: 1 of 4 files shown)


alimanfoo:
[example notebook](http://nbviewer.jupyter.org/github/alimanfoo/zarr/blob/5cc474a3ff21afc7b0173aacb13aa7aae1b4c6c8/notebooks/repr_tree.ipynb) is working in nbviewer

alimanfoo:
cc @jakirkham

notebooks/repr_tree.ipynb
```diff
@@ -6,13 +6,9 @@
...
    "source": [
-    "import zarr\n",
-    "\n",
-    "# Python 2/3 trick\n",
-    "try:\n",
-    "    unicode\n",
-    "except NameError:\n",
-    "    unicode = str"
+    "import sys\n",
+    "sys.path.insert(0, '..')\n",
```
jakirkham: Why is this needed?

zarr/util.py
```diff
@@ -312,117 +313,126 @@ def _repr_html_(self):
...
-    """)
-    result += "</style>\n\n"
-
-    # Insert the HTML list
-    result += """<div class="zarr-tree">\n"""
-    result += "<ul>\n"
-    result += custom_html_sublist(group, indent=indent)
-    result += "</ul>\n"
-    result += "</div>\n"
+def tree_html(group, expand, level):
```
jakirkham: This is probably my error in the first place, but should we just make this a method of `TreeViewer`?

zarr/util.py
```diff
@@ -312,117 +313,126 @@ def _repr_html_(self):
...
+    """ % (node_id, tree_group_icon, tree_array_icon))
 
     return result
 
 
+tree_group_icon = 'fa fa-folder'
+tree_array_icon = 'fa fa-table'
+# alternatives...
+# tree_group_icon: 'jstree-folder'
+# tree_array_icon: 'jstree-file'
```
jakirkham: Maybe these should be variables in `TreeViewer`.

notebooks/repr_tree.ipynb
alimanfoo: So zarr is on the path. How did you do that, did you set PYTHONPATH?

zarr/util.py
```diff
@@ -312,117 +313,126 @@ def _repr_html_(self):
...
-    result += custom_html_sublist(group, indent=indent)
-    result += "</ul>\n"
-    result += "</div>\n"
+def tree_html(group, expand, level):
+
+    result = ''
+
+    # include CSS for jstree default theme
+    css_url = '//cdnjs.cloudflare.com/ajax/libs/jstree/3.3.3/themes/default/style.min.css'
+    result += '<link rel="stylesheet" href="{}"/>'.format(css_url)
```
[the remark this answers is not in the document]
alimanfoo: I think that just gets handled automatically. I've just been copying examples there.

jakirkham:
Looks pretty good to me. Made a few comments above, but nothing essential. Would be nice to get rid of the `sys.path` hack if possible, but not crucial if there is some issue requiring it as a workaround.

zarr/util.py
alimanfoo: Yes, could do.

zarr/util.py
alimanfoo: Yes could be. I was trying to think of best way to make it so a really enthusiastic user could change the default icons if they want to use their favourite emojis instead :-)

alimanfoo:
Thanks for review, would be good to know what you do instead of sys.path hack.

zarr/util.py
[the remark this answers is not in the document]
jakirkham: Guess so. Interesting. Also this [SO question]( https://stackoverflow.com/q/9374203 ) seems relevant.

zarr/util.py
jakirkham: Was thinking about having static variables of the class as an option as well. Maybe that would work better?

jakirkham:
Normally I've been doing `pip install -e .` when working locally, which ensures the development version is on the path. If that's not working, would be curious to know more details about the issue that is causing the need for the hack.

alimanfoo:
Thanks for the tip re pip install -e, I've updated the demo notebook. Also added support for ``expand`` being an int, so can control to what depth tree is expanded. I think I'll leave the rest as-is, having some of these methods broken out as functions is convenient, especially where long lines are needed for embedded html/javascript.

jakirkham:
Thanks @alimanfoo. Sounds reasonable. We can always go back and tweak later if needed.

PR #192, opened 2017-11-21
LMDB store

### LINKED ISSUE #194: ZipStore is not thread-safe
I believe that the standard library zipfile module is not thread-safe, and so attempts to write to a ZipStore from multiple threads may break. Should verify this, and if so, suggest adding a thread lock to ZipStore `__setitem__`.

alimanfoo:
This PR adds a new ``LMDBStore`` class.

Also some other improvements to the storage module, including making ZipStore thread safe (resolves #194).

### DIFF
--- a/docs/api/storage.rst
+++ b/docs/api/storage.rst
@@ -16,6 +16,11 @@ Storage (``zarr.storage``)
     .. automethod:: close
     .. automethod:: sync
 
+.. autoclass:: LMDBStore
+
+    .. automethod:: close
+    .. automethod:: sync
+
 .. autofunction:: init_array
 .. autofunction:: init_group
 .. autofunction:: migrate_1to2
--- a/docs/api/storage.rst
+++ b/docs/api/storage.rst
@@ -14,12 +14,12 @@ Storage (``zarr.storage``)
 .. autoclass:: DBMStore
 
     .. automethod:: close
-    .. automethod:: sync
+    .. automethod:: flush
 
 .. autoclass:: LMDBStore
 
     .. automethod:: close
-    .. automethod:: sync
+    .. automethod:: flush
 
 .. autofunction:: init_array
 .. autofunction:: init_group
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -30,14 +30,19 @@ Enhancements
   :issue:`123`, :issue:`139`.
 
 * **New storage class for DBM-style databases**. The
-  :class:`zarr.storage.DBMStore` class enables any DBM-style database to be used
-  as the backing store for an array or group. See the tutorial section on
-  :ref:`tutorial_storage` for some examples. :issue:`133`, :issue:`186`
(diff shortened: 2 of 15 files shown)


alimanfoo:
Some [very simple benchmarks](https://github.com/alimanfoo/zarr/blob/2a6731a0d080bf9548c6c1e1b61b346090dc7250/notebooks/store_benchmark.ipynb) suggest this is worth adding, performance is better than DirectoryStore or DBMStore with BerkeleyDB, and close if not better than in-memory store (dict).

alimanfoo:
[Updated benchmarks](https://github.com/alimanfoo/zarr/blob/5190d4f7f8d11a02c7403f5067406cd626be0c4f/notebooks/store_benchmark.ipynb) with larger arrays and some simple dask workflows. Generally confirm that lmdb performs better than bdb.

alimanfoo:
I've updated the [benchmark notebook](https://github.com/alimanfoo/zarr/blob/55a8e16f4aa2d4080cb68e1c172ec2d1d83658da/notebooks/store_benchmark.ipynb) again to include zip store and dask examples. Also demonstrates that zip store is thread safe.

alimanfoo:
Another [benchmark update](https://github.com/alimanfoo/zarr/blob/0601cd129c31db1322b3f4eeca73f04c56d07995/notebooks/store_benchmark.ipynb) including ndbm and bdb btree versus hash.

alimanfoo:
cc @jeromekelleher, if you're already getting good performance with Berkeley DB then this probably won't make much difference, but FWIW it looks like LMDB is [quite a bit faster](https://github.com/alimanfoo/zarr/blob/0601cd129c31db1322b3f4eeca73f04c56d07995/notebooks/store_benchmark.ipynb), almost as fast as memory, probably because the whole database is memory-mapped.

alimanfoo:
FTR I've added some locks to DictStore and DBMStore as well. These don't seem to impact on performance at all even under parallel workloads. They may not be strictly necessary for DictStore because of the GIL, and may not be necessary for DBMStore because at least some of the DBM implementations (GDBM, Berkeley) claim to do their own locking, but seemed worth the extra caution pending some deeper analysis.

alimanfoo:
I think this is ready to go.

jeromekelleher:
Read/write performance really isn't important to me as the vast majority of the time is spend in decompression. LMDB sounds good though. Definitely nice to hear about the extra locking, I was a little worried about letting it all happen at the DB layer.

jakirkham:
Have you played with the number of threads Zarr’s Blosc uses for decompression? Also there may be other tricks to play with like filtering before compression.

jeromekelleher:
I've played around with it a bit all right, but it's not a bottleneck for me. I'm doing compression/decompression in worker threads while the rest of the program is doing other stuff, so I'm not particularly sensitive to the performance.

PR #193, opened 2017-11-21
Add Store rename method

jakirkham:
Lays the ground work for resolving issue ( https://github.com/alimanfoo/zarr/issues/191 ). Namely adds a `rename` function on some `Store`s like `Dict` and `DirectoryStore`. This can later be used by `move`. Since this requires a deletion step it won't work for things like `ZipStore`.

### DIFF
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -554,6 +554,15 @@ class DictStore(MutableMapping):
         else:
             return []
 
+    def rename(self, src_path, dst_path):
+        src_path = normalize_storage_path(src_path)
+        dst_path = normalize_storage_path(dst_path)
+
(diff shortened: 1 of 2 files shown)


alimanfoo:
Looks great to me. Only thing is that rename from keys is not covered by the tests, this is my fault as you were following the pattern for listdir and rmdir where the tests only happen if the store has these methods. I'd be happy to merge then figure out a way to get this covered, may need to split up StoreTests.test_hierarchy into separate functions for each of the "special" methods, so stores like ZipStore can easily override tests for rmdir and rename which are not supported.

jakirkham:
Yeah. Saw that too. Think that was fixed with PR ( https://github.com/alimanfoo/zarr/pull/196 ).

PR #203, opened 2017-11-27
Method to find Zarr Array checksum

jakirkham:
Fixes https://github.com/alimanfoo/zarr/issues/98

Provides a method for Zarr Array's to compute their checksum. Uses the `sha1` checksum by default thanks to its speed and low probability of collisions. Though the user can change the checksum to any supported by their version of `openssl` (per `hashlib` docs) should they desire to. The hashing is done on the data in the chunk store (so encoded) and the metadata (shape, dtype, order, filtering, and attributes).

### DIFF
--- a/zarr/core.py
+++ b/zarr/core.py
@@ -1840,6 +1840,19 @@ class Array(object):
     def hexdigest(self, hashname="sha1"):
         """
         Compute a checksum for the data. Default uses sha1 for speed.
+
+        Examples
+        --------
+        >>> import zarr
+        >>> z = zarr.empty(shape=(10000, 10000), chunks=(1000, 1000))
+        >>> z.hexdigest()
+        '041f90bc7a571452af4f850a8ca2c6cddfa8a1ac'
+        >>> z = zarr.zeros(shape=(10000, 10000), chunks=(1000, 1000))
+        >>> z.hexdigest()
+        '7162d416d26a68063b66ed1f30e0a866e4abed60'
+        >>> z = zarr.zeros(shape=(10000, 10000), dtype="u1", chunks=(1000, 1000))
+        >>> z.hexdigest()
+        'cb387af37410ae5a3222e893cf3373e4e4f22816'
         """
 
         h = hashlib.new(hashname)
--- a/zarr/core.py
+++ b/zarr/core.py
@@ -2,6 +2,7 @@
 from __future__ import absolute_import, print_function, division
 import operator
 import itertools
+import hashlib
 import re
 
 
@@ -87,6 +88,7 @@ class Array(object):
     set_mask_selection
     get_coordinate_selection
     set_coordinate_selection
+    hexdigest
     resize
     append
     view
@@ -1834,6 +1836,24 @@ class Array(object):
         ]
 
         return items
+
+    def hexdigest(self, hashname="sha1"):
+        """
+        Compute a checksum for the data. Default uses sha1 for speed.
+        """
+
+        h = hashlib.new(hashname)
+
+        for i in itertools.product(*[range(s) for s in self.cdata_shape]):
+            h.update(self.chunk_store.get(self._chunk_key(i), b""))
+
+        h.update(self.store.get(self._key_prefix + array_meta_key, b""))
+
+        h.update(self.store.get(self.attrs.key, b""))
+
+        checksum = h.hexdigest()
+
+        return checksum
 
     def __getstate__(self):
         return (self._store, self._path, self._read_only, self._chunk_store,
--- a/docs/api/core.rst
+++ b/docs/api/core.rst
@@ -14,6 +14,7 @@ The Array class (``zarr.core``)
     .. automethod:: set_coordinate_selection
     .. automethod:: get_orthogonal_selection
     .. automethod:: set_orthogonal_selection
+    .. automethod:: hexdigest
     .. automethod:: resize
     .. automethod:: append
     .. automethod:: view
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -104,6 +104,9 @@ Enhancements
 
 * **ZipStore is now thread-safe**; :issue:`194`, :issue:`192`.
 
+* **New Array.hexdigest() method** computes an ``Array``'s hash with ``hashlib``.
+  By :user:`John Kirkham <jakirkham>`, :issue:`98`, :issue:`203`.
+
 
 Bug fixes
 ~~~~~~~~~
--- a/zarr/attrs.py
+++ b/zarr/attrs.py
@@ -24,7 +24,7 @@ class Attributes(MutableMapping):
         return self.asdict()[item]
 
     def _put(self, d):
-        s = json.dumps(d, indent=4, sort_keys=True, ensure_ascii=True)
+        s = json.dumps(d, indent=4, sort_keys=True, ensure_ascii=True, separators=(',', ': '))
         self.store[self.key] = s.encode('ascii')
 
     def _write_op(self, f, *args, **kwargs):
--- a/zarr/meta.py
+++ b/zarr/meta.py
@@ -64,7 +64,7 @@ def encode_array_metadata(meta):
         order=meta['order'],
         filters=meta['filters'],
     )
-    s = json.dumps(meta, indent=4, sort_keys=True, ensure_ascii=True)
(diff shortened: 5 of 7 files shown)


zarr/core.py
```diff
@@ -1835,6 +1837,48 @@ def bytestr(n):
...
+            except KeyError:
+                pass
+
+        filters = self.filters if self.filters else []
+        for meta in [self.fill_value, self.dtype, self.shape, self.order] + filters:
+            r = repr(meta)
+            try:
+                h.update(r)
+            except TypeError:
+                h.update(r.encode())
```
alimanfoo: Instead of this you could just pull out '.zarray' from the store and hash that.

zarr/core.py
```diff
@@ -1835,6 +1837,48 @@ def bytestr(n):
...
+
+        for k, v in self.attrs.items():
+            h.update((k, v))
```
alimanfoo: You could just pull out '.zattrs' from the store and hash that.

jakirkham: However that is not always there right? Though I suppose we could handle this case.

This said, was trying to workaround issues where `.zattrs` may be too large to fit into memory all at once. In this case, the key-value pairs should still fit in memory comfortably.

zarr/core.py
jakirkham: Sounds like a much better idea. Can we rely on `.zarray` always being there or should we add some logic for when it is missing?

alimanfoo: It will always be there, an array isn't an array if it doesn't have a .zarray.

jakirkham:
Thanks for taking a look. Happy to make tweaks. Added a couple comments inline.

As a side note, would like to use this to implement `__dask_tokenize__` as well.

alimanfoo:
Re .zattrs, FWIW I think you can rely on that being small enough to pull into memory, any attribute key update requires loading the whole JSON document into memory anyway.

alimanfoo:
Re ``__dask_tokenize__`` I was just looking at the [deterministic hashing](http://dask.pydata.org/en/latest/custom-collections.html#deterministic-hashing) docs, but doesn't actually say what the hashing is used for. Is this to do with caching intermediate values in a dask computation?

jakirkham:
It's used for things like constructing the `name` of the Dask Array created by `from_array`. Though I agree that should be more explicit in the docs (particularly `from_array`).

Really any time something calls `tokenize` on an object it ends up searching for `__dask_tokenize__` or a function registered with `normalize_token`.

jakirkham:
@jcrist can probably clarify this better than I though. 😉

jakirkham:
Hmm...the checksums I get locally do not match with those the CIs seem to be getting. 😟

Also I'm a little surprised that the checksums of `TestArrayWithPath` and `TestArrayWithChunkStore` cases do not match the `TestArray` case for me locally. Wouldn't expect storage location or format to have an effect, but it seems to. 😕

Any thoughts on either of these would be appreciated.

jakirkham:
Looks like the checksum mismatch issues are restricted to Python 2 only. Not sure of the cause yet, but could be some subtlety with how the data is represented in Python 2 as opposed to Python 3. The concerns were determined to be non-issues.

jakirkham:
Turns out that an extra space is added right before newlines (after commas) in the JSON for things like `.zarray` on Python 2, but this does not occur with Python 3. Verified this is not a CR character, but actually a space (at least on my Mac).

alimanfoo:
Hm, the PY2/PY3 JSON issue is annoying. It would be ideal to get the same hexdigest on all platforms. Not sure what to suggest, will give it some thought.

Re ``__dask_tokenize__`` I wonder how much it matters if the value is potentially expensive to compute. E.g., if it gets called every time you do ``da.from_array(z)`` and ``z`` is a large array and/or an array with relatively slow storage (e.g., S3), the delay to compute the hexdigest could make it unusable.

alimanfoo:
If you set ``indent=None`` for all calls to ``json.dump``, does this give consistent JSON in PY2 and PY3? It's supposed to be the most compact representation. I'd be happy to sacrifice readability for consistency.

jakirkham:
Found the magic sauce as far as Python 2/3 JSON goes. Turns out there was a subtle change in Python 3.4 to the default `separator` behavior (referenced below). If we just override `separator` to match the Python 3.4+ behavior, the issue goes away for Python 2. Pushed a commit to do exactly that. Happy to make the commit message more expressive given this subtlety.

> Changed in version 3.4: Use (',', ': ') as default if indent is not None.

ref: https://docs.python.org/3/library/json.html#json.dump

jakirkham:
Should add IDK if we want something like `digest` provided by `hashlib` algorithms, but it is probably worth considering at this point. Basically it provides the binary string equivalent to the more verbose `hexdigest`.

jakirkham:
> Re \_\_dask_tokenize\_\_ I wonder how much it matters if the value is potentially expensive to compute. E.g., if it gets called every time you do da.from_array(z) and z is a large array and/or an array with relatively slow storage (e.g., S3), the delay to compute the hexdigest could make it unusable.

So I haven't done any benchmarking yet. It would be good to do this for sure. Have access to slow storage (e.g. NFS) to do some experimenting on. Can explore that more tomorrow.

For more context, am wanting to explore caching long running computational results to disk. As part of this, I'd like the `.name` attribute of Dask Arrays to be consistent between runs (if no parameters or inputs have changed). So using some sort of hashing for `__dask_tokenize__` would be ideal. Since Dask chose SHA1 as a default for NumPy Arrays, figured it was a safe bet here. They also try using things like CityHash and MurmurHash, which could be worth looking at if we don't mind another dependency. Though am open to other stable hash options if you have any in mind.

alimanfoo:
Looks good.

alimanfoo:
FWIW SHA1 seems fine to me. 

If you'd like to add ``digest()`` for completeness that's OK by me, fine either way (I'll probably only use ``hexdigest()``).

Re ``__dask_tokenize__`` I think in the general case we have to assume that computing a full hash may be slow simply due to I/O, either because of very large data or slow storage or both. By slow I mean too slow for interactive work. So I'm all for supporting caching, as long as we don't interrupt the flow for the average dask user.

jakirkham:
Ok. I'm not planning to use `digest` either. So let's leave it out. After all people can just call [`binascii.unhexlify`]( https://docs.python.org/3/library/binascii.html#binascii.unhexlify ) to convert anyways.

That may be true. Though was hoping since we avoided decompression, this might be less of an issue. Planning on playing with this more tomorrow. In any event, `__dask_tokenize__` is not pressing as one can still use `normalize_token.register` to define their own way to do this with Zarr or anything else for this matter.

jakirkham:
Added in a changelog entry as well. This is good to go on my end.

jakirkham:
FWIW tried running `hexdigest` with the default `sha1` on some real data to get an idea of how long this might take. Information about the dataset and the time taken to compute the hash are included below. This was run on a Scientific Linux 7.3 cluster using NFSv3.

```python
>>> imgs.info_items()
[('Name', '/images'),
 ('Type', 'zarr.core.Array'),
 ('Data type', 'uint16'),
 ('Shape', '(1636, 1016, 1262)'),
 ('Chunk shape', '(103, 127, 158)'),
 ('Order', 'C'),
 ('Read-only', 'True'),
 ('Compressor', "Blosc(cname='lz4', clevel=5, shuffle=SHUFFLE, blocksize=0)"),
 ('Store type', 'zarr.storage.DirectoryStore'),
 ('No. bytes', '4195332224 (3.9G)'),
 ('No. bytes stored', '2569038619 (2.4G)'),
 ('Storage ratio', '1.6'),
 ('Chunks initialized', '1024/1024')]
>>> %timeit -n 10 imgs.hexdigest()
4.66 s ± 23.6 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

alimanfoo:
Many thanks, nice feature.

jakirkham:
Thanks. We might want to take a look at BLAKE2 down the road as that would provide the option for tree hashing, which seems very nice to use with something like Zarr.

jakirkham:
Found some value in having a `digest` method as well. So put together an implementation in PR ( https://github.com/alimanfoo/zarr/pull/213 ).

PR #207, opened 2017-11-28
Clean up dbm files properly

### LINKED ISSUE #205: FileNotFoundError during test cleanup
Not sure why this is cropping up, but am seeing `FileNotFoundError`s cropping up when trying to delete `ndbm` data. An easy fix might be to change these `atexit` functions to check before trying to delete. Though it doesn't provide any clarity on why the error is occurring in the first place. Sample error below.

```
Error in atexit._run_exitfuncs:
FileNotFoundError: [Errno 2] No such file or directory: '/var/folders/jj/pp1xwt1n4zdgy9zxbxc22t8w66tr7q/T/tmpp98w47w3.ndbm'
```

alimanfoo:
Resolves #205

### DIFF
--- a/zarr/tests/test_storage.py
+++ b/zarr/tests/test_storage.py
@@ -19,7 +19,8 @@ from nose.tools import assert_raises, eq_ as eq, assert_is_none
 from zarr.storage import (init_array, array_meta_key, attrs_key, DictStore,
                           DirectoryStore, ZipStore, init_group, group_meta_key,
                           getsize, migrate_1to2, TempStore, atexit_rmtree,
-                          NestedDirectoryStore, default_compressor, DBMStore, LMDBStore)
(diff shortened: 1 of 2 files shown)


zarr/tests/test_storage.py
```diff
@@ -798,7 +799,7 @@ class TestDBMStoreNDBM(TestDBMStore):
 
             def create_store(self):
                 path = tempfile.mktemp(suffix='.ndbm')
-                atexit.register(os.remove, path)
+                atexit_rmglob(path + '*')
```
jakirkham: These are still suppose to be passed to `atexit.register`, correct?

alimanfoo: Yes, that happens within atexit_rmglob.

alimanfoo: Doh no it doesn't.

Issue #178, opened 2017-11-07
Further tests for structured arrays


alimanfoo:
More tests are needed for basic functionality for arrays with structured dtype, including fill value, 0d array.

jakirkham:
Out of curiosity, are there existing tests for structured arrays somewhere?

jakirkham:
Looks like there were some structured array tests including with fill value already. However there wasn't a test for 0-D structured arrays. So tweaked the existing tests to include the 0-D case. Put this together as PR ( https://github.com/alimanfoo/zarr/pull/210 ).

PR #210, opened 2017-11-28
Test 0-length structured array

jakirkham:
Fixes https://github.com/alimanfoo/zarr/issues/178

Make sure that we are able to store 0-length structured arrays in Zarr.


jakirkham:
Thoughts on this @alimanfoo?

alimanfoo:
I think this is great. I reckon we have enough coverage for structured arrays with this, happy to merge and call #178 done.

jakirkham:
Thanks.

PR #211, opened 2017-11-29
Drops vendored copy of py-cpuinfo

jakirkham:
py-cpuinfo was needed to determine the architecture to build Cython extensions against. However those extensions have migrated to Numcodecs. Hence Zarr has no need for this functionality. That said, one of the notebooks still makes use of py-cpuinfo. Thus it is added as a development requirement.


jakirkham:
FWIW `py-cpuinfo` was only used in `notebooks/genotype_benchmark_compressors.ipynb`, which runs fine for me with this change in the dev environment.

alimanfoo:
Great, thank you.

PR #223, opened 2017-12-31
LRU store cache

alimanfoo:
This PR adds a new ``zarr.storage.LRUStoreCache`` class which can be used as a cache layer for slow stores (e.g., stores accessed via network).

Also in developing this PR some inconsistencies in the way that ``zarr.storage.getsize()`` was implemented for different store types came to light and have been fixed.

TODO:
* [x] tests with max_size not None
* [x] test key caching
* [x] test coverage to 100%
* [x] API docs
* [x] document in tutorial
* [x] document changes

### DIFF
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -1696,6 +1696,7 @@ class LRUStoreCache(MutableMapping):
         self._max_size = max_size
         self._current_size = 0
         self._keys_cache = None
+        self._contains_cache = None
         self._listdir_cache = dict()
         self._values_cache = OrderedDict()
         self._mutex = Lock()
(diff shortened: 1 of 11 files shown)


zarr/storage.py
```diff
@@ -1676,3 +1687,125 @@ def __iter__(self):
...
+        return len(self._store)
+
+    def __iter__(self):
+        return self.keys()
+
+    def keys(self):
+        with self._mutex:
+            if self._keys_cache is None:
+                self._keys_cache = list(self._store.keys())
+        return iter(self._keys_cache)
```
alimanfoo: Should probably move this under the mutex lock.

alimanfoo:
cc @jhamman, @mrocklin, this PR provides another strategy for improving performance with cloud stores, adding a ``zarr.LRUStoreCache`` class which implements a local in-memory cache layer for any backing store. This won't help with write performance but should help with read performance for any computations that revisit the same data and/or metadata more than once. Simple usage example [here](https://github.com/alimanfoo/zarr/pull/223/files#diff-4d9b827ea9acd34bb46c7b7354630920R773).

PR #237, opened 2018-01-29
Require numcodecs 0.5.3 and drop nose

jakirkham:
Bumps the `numcodecs` requirement to `0.5.3` as this version uses `pytest` only (without `nose`). As `zarr` uses some testing utilities from `numcodecs`, the version bump is needed to remove the indirect dependency on `nose`. With this requirement change, drop usage of `nose` throughout `zarr`. Also update the testing instructions to use `pytest` (instead of `nose`).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,7 +39,6 @@ htmlcov/
 .coverage
 .coverage.*
 .cache
-nosetests.xml
 coverage.xml
 *,cover
 
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -48,8 +48,8 @@ To work with Zarr source code in development, install from GitHub::
 
 To verify that Zarr has been fully installed, run the test suite::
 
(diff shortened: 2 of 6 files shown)


jakirkham:
Looks like there are some changes to the doctest results, which are causing a failure. Probably due to using the new blosc in numcodecs. How would you like me to proceed?

alimanfoo:
I should've known that would happen. It's tedious I know but would you mind updating the doctests? These looks like all just changes in size of compressed data, probably due to some internal blosc tweaks. We've minimised the number of doctests containing this type of information but there are still a few illustrative examples in the tutorial where it is relevant to show the compressed size.

alimanfoo:
Thought I'd save you a job, have fixed the doctests.

jakirkham:
Thanks @alimanfoo. LGTM.

PR #246, opened 2018-03-22
WIP: Rename DirectoryStore's file/directory to delete

jakirkham:
Before deleting a file or directory in `DirectoryStore`, rename it beforehand so that others are not tempted to touch it while it is being deleted.

### DIFF
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -781,12 +781,15 @@ class DirectoryStore(MutableMapping):
 
     def __delitem__(self, key):
         path = os.path.join(self.path, key)
+        temp_path = path + '.partial'
         if os.path.isfile(path):
-            os.remove(path)
+            os.rename(path, temp_path)
+            os.remove(temp_path)
         elif os.path.isdir(path):
             # include support for deleting directories, even though strictly
             # speaking these do not exist as keys in the store
-            shutil.rmtree(path)
+            os.rename(path, temp_path)
+            shutil.rmtree(temp_path)
         else:
             raise KeyError(key)
 


zarr/storage.py
```diff
@@ -781,12 +781,15 @@ def __setitem__(self, key, value):
 
     def __delitem__(self, key):
         path = os.path.join(self.path, key)
+        temp_path = path + '.partial'
         if os.path.isfile(path):
-            os.remove(path)
+            os.rename(path, temp_path)
```
kalvdans: This rename will fail if the `.partial` file already exist.

zarr/storage.py
```diff
@@ -781,12 +781,15 @@ def __setitem__(self, key, value):
...
+            os.remove(temp_path)
         elif os.path.isdir(path):
             # include support for deleting directories, even though strictly
             # speaking these do not exist as keys in the store
-            shutil.rmtree(path)
+            os.rename(path, temp_path)
```
kalvdans: This rename will fail if the `.partial` file already exist.

zarr/storage.py
jakirkham: Actually that's only true on Windows. Unix it is deleted silently. Please see [this note in the docs]( https://docs.python.org/3/library/os.html#os.rename ).

That said, I think this is an unlike scenario and even if it occurred would expect we want it to fail (see below).

We only create a `*.partial` file currently if we are writing a value, but have not completed writing yet. Also we make sure to delete the corresponding file `*` before creating `*.partial`. So both should not exist at the same time. If we do ever see a `*.partial` file, that's an indication of an incomplete write (IOW data corruption), in which case we should fail somehow. Admittedly we may want to adjust the failure to make it more informative and less reliant on internal spec.

So agree that behavior should be adjusted, but think it should be in the direction of always failing if a `*.partial` file exists.

zarr/storage.py
jakirkham: Same as above.

jakirkham: Actually would add we don't create a directory named `.partial` at all currently.

jakirkham:
The goal is to make sure that if an Array is in the process of being written to, this will block users from touching it in an incomplete state. This should be thought of in conjunction with PR ( https://github.com/zarr-developers/zarr/pull/248 ).

I'm not sure that I would want it merged in its current state (even with a test) as it is helpful for some use cases, but harmful for others. Will need to give it some thought.

kalvdans:
Thanks for the explanation! Detection of unsupported operations are always better than silently corrupting data. I have no more objections right now.

jakirkham:
Of course. Glad that helped.

Thanks for asking about it. Helped me mull over this idea a bit more. Will try to refine it a bit more and we can discuss it again once it's ready for another look (if you are interested).

jakirkham:
That just about sums it up, @alimanfoo. This change alone isn’t really enough, but it’s a starting point.

alimanfoo:
Re synchronization, to be clear, you are concerned about a situation where one parallel worker deletes an array and another worker tries to write to the array at the same time? If so, is this theoretical or a real use case for you? 

Re NFS performance, I am surprised

alimanfoo:
...that rename is slower than remove. We have NFS v4 on our cluster so I can try it on Monday but just wondering if you have any timing data?

PR #248, opened 2018-03-22
WIP: Ignore DirectoryStore dir/file ending in .partial

jakirkham:
If a directory name or file name ends in `.partial`, exclude them from `keys`. These are effectively internal spec that no one should be touching or interacting with.

### DIFF
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -805,7 +805,11 @@ class DirectoryStore(MutableMapping):
             directories = [(self.path, '')]
             while directories:
                 dir_name, prefix = directories.pop()
+                if dir_name.endswith('.partial'):
+                    continue
                 for name in os.listdir(dir_name):
+                    if name.endswith('.partial'):
+                        continue
                     path = os.path.join(dir_name, name)
                     if os.path.isfile(path):
                         yield prefix + name


jakirkham:
Not sure if there are other places we would need to do this (assuming we are ok with it).

alimanfoo:
Ignoring files ending in ".partial" is probably reasonable.

Don't think we'd want to ignore directories, a user might reasonably want to create a group or array called "something.partial". 

Is there a use case motivating this?

jakirkham:
Yeah, wanted to make sure that data wouldn't accidentally be accessed in an incomplete state.

Also wanted to have a way of knowing when an Array was completely written to and thus use this to block anything from touching it beforehand. In this case supporting something like `.partial` for directories would be useful.

However being able to read directories in an incomplete state (e.g. when some chunks are already written) is very helpful for other use cases. So I don't think this is the solution, but it is an attempt to start that conversation.

alimanfoo:
Fwiw synchronisation issues may be better dealt with at a higher level in
the architecture. Zarr has a synchronisation module and supports locking of
a chunk, an array or a group. If you could specify the operations that you
want to synchronize, we could look at how to achieve via locking.

PR #249, opened 2018-03-22
Use `os.renames` in `DirectoryStore`'s `rename`

jakirkham:
Instead of using `os.rename` and creating missing directories before when renaming content in a `DirectoryStore`, just use `os.rename` to handle this move for us. This will automatically create missing directories for us in the destination path so we don't have to. Also it prunes away any empty directories in the old path, which we were not doing before. IOW this fixes a bug as well.


alimanfoo:
Thanks, this looks tidy. What bug does this fix out of interest?

jakirkham:
If renaming left an empty directory behind, this will clean it up. That said, I'm not actually sure that would ever happen.

PR #273, opened 2018-07-06
fix pickle tests: avoid multiple simultaneous active writers

ryan-williams:
Several `test_pickle` cases fail for me on OSX (10.13.5) and in a CentOS (7.5) VM; [here's a full example](https://gist.github.com/ryan-williams/b4093db6a978c1c9066da5c15efd9d9d).

A minimal repro:

```python
import tempfile
from zarr.storage import DBMStore
import pickle
path = tempfile.mktemp()
store = DBMStore(path, flag='n')
store['foo'] = b'bar'
pickled = pickle.dumps(store)
store2 = pickle.loads(pickled)
```

Running this gives:

```
  File "test.py", line 8, in <module>
    store2 = pickle.loads(pickled)
  File "/Users/ryan/c/hdf5-experiments/zarr/zarr/storage.py", line 1449, in __setstate__
    write_lock=write_lock, **open_kws)
  File "/Users/ryan/c/hdf5-experiments/zarr/zarr/storage.py", line 1425, in __init__
    self.db = open(path, flag, mode, **open_kwargs)
  File "/Users/ryan/.pyenv/versions/3.6.5/lib/python3.6/dbm/__init__.py", line 94, in open
    return mod.open(file, flag, mode)
_gdbm.error: [Errno 35] Resource temporarily unavailable
```

It seems to be related to having multiple file-descriptors open with "write" capabilities to a given path.

This fixes them for me locally; I don't know why they don't appear on Travis.

### DIFF
--- a/zarr/tests/test_core.py
+++ b/zarr/tests/test_core.py
@@ -659,8 +659,16 @@ class TestArray(unittest.TestCase):
         z = self.create_array(shape=1000, chunks=100, dtype=int, cache_metadata=False,
                               cache_attrs=False)
         z[:] = np.random.randint(0, 1000, 1000)
-        z2 = pickle.loads(pickle.dumps(z))
-        assert z.shape == z2.shape
+        pickled = pickle.dumps(z)
+
+        shape = z.shape
+        array = z[:]
+
+        if hasattr(z.store, 'close'):
+            z.store.close()
+
+        z2 = pickle.loads(pickled)
+        assert shape == z2.shape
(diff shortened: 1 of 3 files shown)


ryan-williams:
The coverage decrease seems spurious to me; please advise?

alimanfoo:
Thanks again @ryan-williams for this. I ended up manually merging some changes based on this PR into another PR (#308), so this should be resolved in master now.

Issue #267, opened 2018-06-07
negative chunk indexes, offset chunk origin


chairmank:
This is a feature request. If you think that it is worth doing, I volunteer to update the zarr specification and implementation.

The v2 specification states

> For example, given an array with shape (10000, 10000) and chunk shape (1000, 1000) there will be 100 chunks laid out in a 10 by 10 grid. The chunk with indices (0, 0) provides data for rows 0-1000 and columns 0-1000 and is stored under the key “0.0”; the chunk with indices (2, 4) provides data for rows 2000-3000 and columns 4000-5000 and is stored under the key “2.4”; etc.

> Note that all chunks in an array have the same shape. If the length of any array dimension is not exactly divisible by the length of the corresponding chunk dimension then some chunks will overhang the edge of the array. The contents of any chunk region falling outside the array are undefined.

Together, these statements imply the following restrictions:

- the `[0, 0, ..., 0]` origin "corner" of a N-dimensional array must coincide with the origin corner of the `[0, 0, ..., 0]`th chunk
- overhanging chunks may only appear at the edges of the array that are far from the origin

These restrictions make it convenient to grow/shrink an N-dimensional array along the edges that are far from zero but inconvenient to grow/shrink along the zero-index edges. For example, consider this chunked array with shape `[3, 3]`:

```
[[ 0, 1, 2],
 [ 3, 4, 5],
 [ 6, 7, 8]]
```

If this array is split into chunks of `[2, 2]`, the chunks are

- chunk `[0, 0]` is `[[0, 1], [3, 4]]`
- chunk `[0, 1]` is `[[2, undefined], [5, undefined]]`
- chunk `[1, 0]` is `[[6, 7], [undefined, undefined]]`
- chunk `[1, 1]` is `[[8, undefined], [undefined, undefined]]`

To concatenate an array of zeroes `[[0, 0, 0]]` on the non-zero edge of the 0th dimension, I only need to change the shape of the array to `[4, 3]` and update chunks `[1, 0]` and `[1, 1]`:

```
[[ 0, 1, 2],
 [ 3, 4, 5],
 [ 6, 7, 8],
 [ 0, 0, 0]]
```

- chunk `[0, 0]` is unchanged
- chunk `[0, 1]` is unchanged
- chunk `[1, 0]` becomes `[[6, 7], [0, 0]]`
- chunk `[1, 1]` becomes `[[8, undefined], [0, undefined]]`

[...]

Issue #244, opened 2018-03-07
Complex dtype error  


alimanfoo:
#### Minimal, reproducible code sample, a copy-pastable example if possible

```python
zarr.array(1j*np.arange(10))
```
#### Problem description

Above code snippet generates an exception:

```
TypeError: Object of type 'complex128' is not JSON serializable
```

This is due to the default fill value of zero being cast as complex before attempting to store as fill value in JSON. Workaround is to set ``fill_value=None`` when creating the array.

#### Version and installation information

Please provide the following:

* Value of ``zarr.__version__``: 2.2.0
* Value of ``numcodecs.__version__``: 0.5.4
* Version of Python interpreter: 3.6
* Operating system (Linux/Windows/Mac): Linux
* How Zarr was installed (e.g., "using pip into virtual environment", or "using conda"): pip into venv

chairmank:
To fix this problem, what do you think about evolving the specification

https://zarr.readthedocs.io/en/stable/spec/v2.html#fill-value-encoding

so that the fill value in the metadata JSON is a base64-encoded *binary representation* of a value with the specified data type? For a given NumPy-style type string (even for structs), we can unambiguously decode the binary representation. To distinguish it from legacy `fill_value` specifications, this binary representation could be packaged in a special JSON object. Examples are below.

For backwards compatibility, continue to parse JSON numeric literals:
```
{
    "dtype": "<f8",
    "fill_value": 0.9868011474609375,
 }
```

For backwards compatibility, continue to recognize strings like `"NaN"` as special values:
```
{
    "dtype": "<f8",
    "fill_value": "NaN",
 }
```

New style binary representation of fill value. `"AAAAAAAA+H8="` is the base64 encoding of `'\x00\x00\x00\x00\x00\x00\xf8\x7f` which is NaN as a little-endian float64.

```
{
    "dtype": "<f8",
    "fill_value": {
      "data": "AAAAAAAA+H8="
    }
 }
```

chairmank:
@jakirkham suggested a similar idea on https://github.com/zarr-developers/zarr/issues/216

jakirkham:
As a potentially quicker fix, we could do something similar to what we do with datetime/timedelta types and cast the complex values to something else (maybe a record with floats).

```python
zarr.array((1j*np.arange(10)).view(np.dtype([("real", np.float64), ("imag", np.float64)])))
```

jakirkham:
[...]

PR #276, opened 2018-07-15
DOC: zarr spec v3: adds optional dimensions and the "netZDF" format

shoyer:
xref #167

For ease of review, this is currently written by modifying `docs/spec/v2.rst`, but this would of course actually be submitted as a separate v3 file.

This does not yet include any changes to the zarr reference implementation, which would need to grow at least:

- `Array.dimensions`
- `Group.dimensions`
- `Group.resize` for simultaneously resizing multiple arrays in a group which share the same dimension (conflicting dimension sizes are forbidden by the spec)
- `Group.netzdf` for indicating whether a group satisfies the netzdf spec or not.

Note: I do like "netzdf" but I'm open to less punny alternatives :).

### DIFF
--- a/docs/spec/v2.rst
+++ b/docs/spec/v2.rst
@@ -1,6 +1,6 @@
 .. _spec_v2:
 
-Zarr storage specification version 2
+Zarr storage specification version 3
 ====================================
 
 This document provides a technical specification of the protocol and format
@@ -78,6 +78,14 @@ filters
     filters are to be applied. Each codec configuration object MUST contain a
     ``"id"`` key identifying the codec to be used.
 
+The following keys MAY be present:
+
+dimensions
+    A list of string or ``null`` values providing optional names for each ofthe
+    array's dimensions. If provided, the list MUST have length equal to the
+    number of array dimensions. If omitted, the array MUST be treated
+    equivalently to providing dimensions as a list of all ``null`` values.
+
 Other keys MUST NOT be present within the metadata object.
 
 For example, the JSON object below defines a 2-dimensional array of 64-bit
@@ -98,6 +106,10 @@ using the Blosc compression library prior to storage::
             "clevel": 5,
             "shuffle": 1
         },
+        "dimensions": [
+            "row",
+            "column"
+        ]
         "dtype": "<f8",
         "fill_value": "NaN",
         "filters": [
@@ -108,7 +120,7 @@ using the Blosc compression library prior to storage::
             10000,
             10000
         ],
-        "zarr_format": 2
+        "zarr_format": 3
     }
 
 .. _spec_v2_array_dtype:
@@ -284,6 +296,20 @@ zarr_format
     An integer defining the version of the storage specification to which the
     array store adheres.
 
+The following keys are OPTIONAL:
+
+dimensions
+    A JSON object defining a map from string dimension names to integer sizes.
+    All arrays in a group or its descendents with dimension names MUST have
+    matching size along their named dimensions, unless any of those dimensions
+    are overriden by dimensions in a descendent group.
+netzdf
+    An optional boolean indicating whether arrays within the group and its
+    descendents adhere to the more restrictive "netZDF" file-format (detailed
+    below), in which dimensions are REQUIRED for all arrays. If omitted,
+    software SHOULD NOT make assumptions about whether or not dimensions can be
+    found on all arrays.
+
 Other keys MUST NOT be present within the metadata object.
 
 The members of a group are arrays and groups stored under logical paths that
@@ -311,6 +337,61 @@ For example, the JSON object below encodes three attributes named
         "bar": "apples",
         "baz": [1, 2, 3, 4]
     }
+
+.. _spec_v2_dimensions:
+
+Dimensions
+----------
+
+Groups and arrays can be associated with optional dimension names. This feature
+is intended to facilitate self-described datasets.
+
+Dimensions are required to be consistent. Any dimensions set on an array
+(any non-``null`` value), MUST also be defined on an ancestor group. Dimension
+sizes can be overwritten in descendent groups, but the size of each named
+dimensions on an array MUST match the size of that dimension on the most direct
+ancestor group on which it is defined.
+
+For example, the JSON objects below describe a hierarchy of arrays, where the
+dimension ``x`` has size 1000 on the array ``foo`` and size 2000 on the array
+``nested/bar``::
+
+    .zgroup:
+        {
+            "dimensions": {"x": 1000},
+            ...
+        }
+    foo/.zarray:
+        {
+            "dimensions": ["x"],
+            "shape": [1000]
+            ...
+        }
+    nested/.zgroup:
+        {
+            "dimensions": {"x": 2000},
+            ...
+        }
+    nested/bar/.zarray:
+        {
+            "dimensions": ["x"],
+            "shape": [2000]
+            ...
+        }
+
+.. _spec_v2_netzdf:
+
+NetZDF
+------
+
+NetZDF is a more restricted variant of the Zarr storage format, with the
+following changes:
+
+* The "dimensions" field is REQUIRED for all arrays.
+* All entries in "dimensions" on arrays MUST be strings: ``null`` dimensions are
+  not allowed.
+* The "netzdf" field is REQUIRED, with a value of ``true``, on all groups that
+  that obey the netZDF spec.
 
 .. _spec_v2_examples:
 
@@ -358,7 +439,7 @@ Inspect the array metadata::
             20,
             20
         ],
-        "zarr_format": 2
+        "zarr_format": 3
     }
 
 Chunks are initialized on demand. E.g., set some data::
@@ -433,7 +514,7 @@ Inspect the group metadata::
 
     >>> print(open('data/group.zarr/.zgroup').read())
     {
-        "zarr_format": 2
+        "zarr_format": 3
     }
 
 Create a sub-group::
@@ -494,6 +575,13 @@ What has been stored::
 
 Changes
 -------
+
+Version 3 changes
+~~~~~~~~~~~~~~~~~~~
+
+* Optional support for named dimensions on arrays and groups.
+* Added a description of the more restrictive "netZDF" format, inspired by the
+  `netCDF <https://www.unidata.ucar.edu/netcdf>`_ data model.
 
 Version 2 clarifications
 ~~~~~~~~~~~~~~~~~~~~~~~~
--- a/docs/spec/v2.rst
+++ b/docs/spec/v2.rst
@@ -346,11 +346,11 @@ Dimensions
 Groups and arrays can be associated with optional dimension names. This feature
 is intended to facilitate self-described datasets.
 
-Dimensions are required to be consistent. Any dimensions set on an array
-(any non-``null`` value), MUST also be defined on an ancestor group. Dimension
-sizes can be overwritten in descendent groups, but the size of each named
-dimensions on an array MUST match the size of that dimension on the most direct
-ancestor group on which it is defined.
+Setting dimensions on groups is an OPTIONAL way to indicate that arrays that
+use reuse the same dimension have a consistent size. When a dimension is set on
+a group, the size of each dimensions on arrays inside that group is REQUIRED to
+match. This includes arrays inside descendent groups, unless the dimension is
+explicitly overwritten by dimensions on a descendent group.
 
 For example, the JSON objects below describe a hierarchy of arrays, where the
 dimension ``x`` has size 1000 on the array ``foo`` and size 2000 on the array
@@ -379,6 +379,10 @@ dimension ``x`` has size 1000 on the array ``foo`` and size 2000 on the array
             ...
         }
 
+If dimensions were removed from ``nested/.zarray`` then the array store would
+be invalid, because the array ``nested/bar`` has inconsistent size for
+dimension ``x`` from the size of the dimension in the root group.
+
 .. _spec_v2_netzdf:
 
 NetZDF
@@ -390,6 +394,7 @@ following changes:
 * The "dimensions" field is REQUIRED for all arrays.
 * All entries in "dimensions" on arrays MUST be strings: ``null`` dimensions are
   not allowed.
+* Every dimension on an array MUST be found on dimensions in an ancestor group.
 * The "netzdf" field is REQUIRED, with a value of ``true``, on all groups that
   that obey the netZDF spec.
 


jhamman:
Thanks @shoyer for writing this up. I had been using ZCDF as the acronym for this feature set in zarr but also don't have strong feelings about the name at this point. FWI, @alimanfoo, @rabernat, @mrocklin, and I have had a few offline exchanges on the subject (see https://github.com/jhamman/zcdf_spec for a Zarr subspec that describes what xarray has currently implemented). Without speaking for anyone else, I think there is growing excitement about the concept of a Zarr+NetCDF data model.

alimanfoo:
Great to see this. I like the design, it's simple and intuitive. Couple of questions...

1. Do you need a way to handle coordinate variables, i.e., being able to express the fact that an array contains the coordinates for some dimension?

2. These features could be implemented within .zattrs without requiring any changes to the spec. I'm open to considering a spec change, and there may be other reasons for wanting to update the spec in the not-too-distant future (xref #244). But a spec change will mean some disruption for existing users and some minor complexities about supporting migration etc. Ultimately I'll be happy to follow the consensus but was just wondering what the rationale was for including these changes within the core metadata and not .zattrs.

mrocklin:
> These features could be implemented within .zattrs without requiring any changes to the spec

I find myself agreeing with this.  I think that ideally Zarr would remain low-level and that we would provide extra conventions/subspecs on top of it.  

My understanding is that one reason for HDF's current inertia is that it had a bunch of special features glommed onto it by various user communities.  If we can separate these out that might be nice for long term maintenance.

shoyer:
> Do you need a way to handle coordinate variables, i.e., being able to express the fact that an array contains the coordinates for some dimension?

No, for more sophisticated metadata needs we can simply use a subset of [CF Conventions](http://cfconventions.org/). These are pretty standard for applications that handle netCDF files, like xarray.

> Ultimately I'll be happy to follow the consensus but was just wondering what the rationale was for including these changes within the core metadata and not .zattrs.

This is a good question. Mostly it comes down to having the specs all in one place, so it's obvious where to find this convention for everyone implementing the zarr spec. Dimensions are broadly useful enough for self-described data that I think people in many fields would find them useful. In particular, I would hate to see separate communities develop their own specs for dimensions, just because they didn't think to search for `zarr netcdf`.

I also think there are probably use cases for defining named dimensions on some but not all arrays and/or axes. This wouldn't make sense as part of the "netzdf" spec which xarray would require.

Finally, incorporating `dimensions` directly in the data model follows precedence from the [netCDF format itself](https://www.unidata.ucar.edu/software/netcdf/docs/file_format_specifications.html), which is actually pretty simple. I agree that we don't want to make zar as complex as HDF5 (which is part of why these aren't full fledged dimension scales) but adding a couple of optional metadata keys is only a small step in that direction.

jakirkham:
Have not thought about this too deeply yet. So this is just a very rough idea that we can discard if it doesn't fit, but what if we added ZCDF as a library within the org that built off Zarr? This would address some of the discoverability, and feature creep concerns raised thus far. It would also eliminate the need for things like checks as to whether the NetCDF spec is implemented by specific objects.

alimanfoo:
> This is a good question. Mostly it comes down to having the specs all in one place, so it's obvious where to find this convention for everyone implementing the zarr spec. Dimensions are broadly useful enough for self-described data that I think people in many fields would find them useful. In particular, I would hate to see separate communities develop their own specs for dimensions, just because they didn't think to search for zarr netcdf.

FWIW we could add this as a "NetZDF spec" (or whatever name) alongside the existing storage specs in the [specs section of the Zarr docs](http://zarr.readthedocs.io/en/stable/spec.html), should be pretty visible (in fact might be more visible as it would get its own heading in the toc tree).

I would be keen to minimise disruption for existing users and implementers if possible. A spec version change would imply some inconvenience, even if relatively small, as existing data would need to be migrated.

mrocklin:
My understanding is that this proposal is entirely compatible (both
backwards and forwards) with existing data

shoyer:
> My understanding is that this proposal is entirely compatible (both
backwards and forwards) with existing data

Indeed. I considered naming this "v2.1" based on semantic versioning until I saw that the zarr spec only uses integer versions.

The only backwards incompatibility it introduces is the addition of new optional metadata fields. I would hope that any existing software would simply ignore these, rather than assume that no fields could ever be introduced in the future.

shoyer:
> Have not thought about this too deeply yet. So this is just a very rough idea that we can discard if it doesn't fit, but what if we added ZCDF as a library within the org that built off Zarr? This would address some of the discoverability, and feature creep concerns raised thus far. It would also eliminate the need for things like checks as to whether the NetCDF spec is implemented by specific objects.

Yes, this makes some amount of sense.

The main downside of incorporating these changes into zarr proper is that for netCDF compatibility we really want the guarantee of consistent dimension sizes between arrays. This would require a small amount of refactoring and additional complexity to achieve within the Zarr library.

docs/spec/v2.rst
```diff
@@ -312,6 +338,61 @@ For example, the JSON object below encodes three attributes named
...
+----------
+
+Groups and arrays can be associated with optional dimension names. This feature
+is intended to facilitate self-described datasets.
+
+Dimensions are required to be consistent. Any dimensions set on an array
+(any non-``null`` value), MUST also be defined on an ancestor group. Dimension
+sizes can be overwritten in descendent groups, but the size of each named
+dimensions on an array MUST match the size of that dimension on the most direct
+ancestor group on which it is defined.
```
shoyer: I think I'm going to change this, to make group dimensions and consistency entirely optional:
```
If dimensions are set in a group, their sizes on all contained arrays
are REQUIRED to be consistent. Dimension sizes can be overwritten
in descendant groups, but the size of each named dimension (any
non-`null` value) on an array MUST match the size of that dimension
on the most direct ancestor group on which it is defined.
```

jakirkham:
From a neuroscience data perspective, this gets pretty complicated pretty fast if one wants to be general. Please see [NWB as an example]( https://nwb-schema.readthedocs.io/en/latest/ ). Personally wouldn't want Zarr to take any of this on. It would be better handled in a library on top of Zarr. Note NWB currently is built on top of HDF5, but it would be reasonable to consider an NWB spec on top of Zarr.

Can't speak to the Earth Sciences or what people in this field want out of Zarr. If dimensionality is the one thing desired, maybe this is ok. If there are 5 or 6 more things needed in the pipe, maybe having a library built on top of Zarr would be better. Would be good if some people could answer these sorts of questions.

jakirkham:
Sorry for the multiple posts. GitHub is having some sort of issue that is affecting me.

jakirkham:
Some miscellaneous thoughts about dimensionality in our field since Matt asked.

Naming dimensions has certainly come up before. Here is [one example]( https://ukoethe.github.io/vigra/doc-release/vigranumpy/index.html#axistag-reference ) and [another]( https://ukoethe.github.io/vigra/doc-release/vigranumpy/index.html#vigra.VigraArray.withAxes ). Also some discussion about [axes names in this comment and below]( https://github.com/imageio/imageio/issues/263#issuecomment-306718628 ). Scientist definitely like having this sort of feature as it helps them keep track of what something means and is useful if the order ever needs to change for an operation. So this sort of use case benefits from the proposal.

The other thing that typically comes to mind when discussing dimensions, which I don't think has come up thus far is units. It's pretty useful to know something is in `ms`, `mV`, or other relevant units. Libraries like [quantities]( http://python-quantities.readthedocs.io/en/latest/user/tutorial.html ) or [pint]( http://pint.readthedocs.io/en/latest/ ) are useful for tracking units and combining them sensible. This could be an addition to the proposal or perhaps something to add to a downstream data format library.

For tracking time in some cases we have timestamps. This supplants the need for dimension or units and often parallels other information (e.g. snapshots of other data at a particular time). This could use existing things like structured arrays.

However when looking applying some basic machine learning, dimensions pretty quickly become a mess. Especially if various different kinds of dimensions get mixed together. For example PCA is a pretty common technique to perform in a variety of cases to find the biggest contributors to some distribution. The units of this sort of thing are frequently strange and difficult to think about. This case probably either needs a different proposal or users have to work with existing metadata information to make this work for their use case.

mrocklin:
Also cc'ing @ambrosejcarr and @cadair to add some domain breadth to this
discussion

DennisHeimbigner:
From the point of view of the Unidata netcdf group, named dimensions (shared dimensions in netcdf parlance) is essential for managing coordinate variables. So the netcdf extension to Zarr (or possibly TIleDB) will include only named dimensions and anonymous dimensions will probably be suppressed. We went around about this with the HDF5 group long ago.
One of the sticking points was multi-dimensional coordinate variables.

alimanfoo:
Speaking as a user in the genomics domain, I certainly would find this feature useful, it is common to have multiple arrays sharing dimensions. I don’t have a broad experience in other domains but expect this feature to be generally very useful. So I am very supportive and would like to give this as much prominence as possible. 

My reasons for leaning towards use of .zattrs is not meant in any way to diminish the importance or broad applicability of this feature, it is based purely on technical considerations, basically on what is easiest to implement and provides the least disruption for existing users and implementers.

> My understanding is that this proposal is entirely compatible (both backwards and forwards) with existing data

Yes in theory, although unfortunately it’s not quite that simple in practice. I’ll try to unpack some details about versioning and change management in Zarr. Btw I’m not suggesting this ideal or the best solution, thinking ahead about possible changes and managing compatibility is quite hard.

This proposal adds a new feature (dimensions) to the Zarr storage spec. This feature is optional in two senses. First it is optional in that it specifies elements that do not need to be present in the array or group metadata. Second it is optional for the implementation, i.e., an implementation can ignore these elements if present in the metadata and still be conformant with the storage spec. 

When I wrote the v2 storage spec and was thinking about mechanisms for managing change, for better or worse, I did not allow any mechanisms for adding optional features to the spec. There is no concept of minor spec versions, only major versions (single version number). The only way the spec can change is via a major version change, which implies a break in compatibility. If the current implementation finds anything other than “2” as the spec version number in array metadata, it raises an exception. The spec does not define any concept of optional features or leave open the possibility of introducing them (other than via a major version change).

If I had been farsighted, I might have seen this coming, and I might have defined a notion of optional features, which could be introduced via a minor version increment to the spec, and implementations could include some flexibility in matching the format minor version number when deciding if they can read some data or not. To be fair I did give this some thought, although I couldn’t have articulated it very well at the time. In the end I decided on a simple versioning policy I think partly because it was simple to articulate and implement, and also because I thought that the user attributes (.zattrs) always provided a means for optional functionality to be layered on. Also the separation between .zattrs and core metadata (.zarray, .zgroup) is nice in a way because it makes it very clear where is the line between optional and required features. I.e., to be conformant, a minimal implementation has to understand everything in .zarray, and can ignore everything in .zattrs.

So given all this, there are at least three options for how to introduce this feature. In below, by “old code” I mean the current version of the zarr package (which does not implement this feature), by “old data” I mean data created using old code, by “new code” I mean the next version of the zarr package (which does implement this feature), by “new data” I mean data created using new code.

Option 1: Use .zattrs, write this as a separate spec. Full compatibility, old code will be able to read new data, and new code will be able to read old data.

Option 2: Use .zarray/.zgroup, incorporate into the storage spec, major version bump (v3). Old code will not be able to read new data. New code can read old data if data is migrated (which just requires replacing the version number in metadata files) or if new code is allowed to read both v2 and v3.

Option 3: Use .zarray/.zgroup, incorporate into the storage spec but leave spec version unchanged (v2). Full compatibility, old code will be able to read new data, and new code will be able to read old data. However, this is potentially confusing because the spec has changed but the spec version number hasn’t. 

Hence I lean towards option 1 because it has maximum compatibility and seems simplest/least disruptive. But very happy to discuss. And I’m sure there are other options too that I haven’t thought of.

rabernat:
> The other thing that typically comes to mind when discussing dimensions, which I don't think has come up thus far is units.

> For tracking time in some cases we have timestamps.

@jakirkham - both of these issues arise in geoscience use cases. We handle them by providing metadata that follows [CF conventions](http://cfconventions.org/) and then using xarray to decode the metadata into appropriate types (like a `numpy.datetime64'). This works today with zarr + xarray and doesn't require any changes to the spec.

shoyer:
> From the point of view of the Unidata netcdf group, named dimensions (shared dimensions in netcdf parlance) is essential for managing coordinate variables. So the netcdf extension to Zarr (or possibly TIleDB) will include only named dimensions and anonymous dimensions will probably be suppressed. We went around about this with the HDF5 group long ago. One of the sticking points was multi-dimensional coordinate variables.

Yes, this is why I want named dimensions.

I don't think we need explicit support for multi-dimensional coordinate variables in Zarr. NetCDF doesn't have explicit support for coordinates *at all*, and we get along just fine using CF conventions.

HDF5's dimension scales include coordinate values as well as dimension names. But in my opinion this is unnecessarily complex. Simple conventions like treating variables with the same name as a dimension as supplying coordinate values are sufficient.

DennisHeimbigner:
WRT to things like units, you need to be very careful about embedding
domain specific semantics into the data model. Our experience is that this
is best left to metadata conventions.

DennisHeimbigner:
Remember that the same dimension may be used in multiple variables, so it is
probably not a good idea to attach dimension information (other than the name)
to a variable.

mrocklin:
Just wanted to briefly chime in that I'm very happy to see NetCDF folks active in this discussion.

DennisHeimbigner:
BTW, one common example of multidimensional coordinate variables is when
defining a trajectory data set.

shoyer:
@alimanfoo I suspect this will not be the last change we will want in the zarr spec (e.g., to support complex numbers), so it might make sense to "bite the bullet" now with a major version number increase, and at the same time establish a clear policy on *forwards compatibility* for Zarr. I am confident that Zarr will only become more and more popular in the future!

I would suggest looking at the forwards compatibility policies from [HDF5](https://support.hdfgroup.org/HDF5/faq/bkfwd-compat.html) and [protocol buffers](https://docs.google.com/document/d/1KMRX-G91Aa-Y2FkEaHeeviLRRNblgIahbsk4wA14gRk/edit#) for inspiration:

- HDF5 files write the *minimum* possible version number that supplies all needed features, to ensure that old clients can read files written with newer versions of the HDF5 library.
- [Protocol buffers](https://en.wikipedia.org/wiki/Protocol_Buffers) are essentially a domain specific language for writing custom file formats. It is widely used at Google and elsewhere (e.g., Apache Arrow uses a protocol buffer successor called flatbuffers). Hard experience has taught us that the right way to handle forward compatibility concerns is to ensure that protocol buffer implementations *ignore but preserve* unknown fields. Protocol buffers are designed to evolve by adding new fields, but changing the meaning of existing fields is strongly discouraged (this would correspond to a major version bump).

Going forward, I would suggest the following forward and backwards compatibility policies, which we can add to the spec:
- **Backwards compatibility:** As much as practical, new versions of the Zarr library should support reading files generated with old versions of the spec (e.g., the zarr library version 3 should still support reading version 2 stores).
- **Forwards compatibility**: New versions of the Zarr library should write the minimum possible version number.
- **Minor version numbers**: The Zarr spec should specify versions with a string (e.g., "2.1") instead of a number. Minor version numbers indicate forwards compatible changes (e.g., the use of new optional features, such as dimension names). Older versions of the Zarr library should support reading newer files and simply ignore/preserve unknown fields. Issuing a warning would be appropriate.

Doing a little more searching, it appears that such a convention is actually widely used. E.g., see ["Versioning Conventions" for ASDF](http://asdf-standard.readthedocs.io/en/latest/versioning.html) and [this page on "Designing File Formats"](https://fadden.com/tech/file-formats.html) (it calls what I've described "major/minor" versioning).

shoyer:
@DennisHeimbigner

>> NetCDF doesn't have explicit support for coordinates at all
> I do not believe this is completely correct; there is no syntactic
support, but if you look at the netcdf 3 and 4 specifications,
it is part of the netcdf semantics.

In the [netCDF spec](https://www.unidata.ucar.edu/software/netcdf/docs/file_format_specifications.html) I find "coordinates" only mentioned for netCDF4, specifically for the `_Netcdf4Coordinates`. That said, I don't really understand why this exists: netCDF's public APIs (e.g., `nc_def_var`) don't reference coordinates at all.

I see that internally, netCDF4 maintains its own notion of "dimension scales" that support more than 1 dimension (beyond what HDF5 supports), which it appears to use for variables if their first dimension matches the name of the variable:
https://github.com/Unidata/netcdf-c/blob/7196dfd6064d778a9973797200d8e64c999d63c5/libsrc4/nc4var.c#L594-L598

Note that this definition of a multi-dimensional coordinate does not even match the typical interpretation of "coordinates" by software that reads netCDF files (e.g., per CF Conventions, "coordinates" are defined merely by being referenced by a "coordinates" attribute, without any requirements on dimension names).

I'm getting a little off track here, but I think the conclusions we can draw from the netCDF4 experience for Zarr are:

1. Dimensions scales as implemented in HDF5 aren't even a particularly good fit for the netCDF data model, given the lengths to which netCDF4 had to go to adapt its data model to HDF5.
2. We don't need to expose an explicit notion of coordinates (as understood by CF Conventions and xarray) for a netCDF-like API. This can be handled by downstream conventions.

DennisHeimbigner:
I stand corrected. One discussion of coordinate variables is here as a convention.
https://www.unidata.ucar.edu/software/netcdf/docs/netcdf_data_set_components.html#coordinate_variables
That reference says that it is a convention, but the source code does take them
into account.
You are correct. We have long recignized that using dimension scales in the netcdf-4
code was probably a mistake and it contorts the code in a number of places.

The multi-dimensional issue is complex because it is most often used with what amounts
to point data (like trajectories) and others have noted that indexed array storage is not very efficient at handling point data: relational systems work much better.
So the multi-dim coordinate variable may be a red-herring wrt this discussion.

DennisHeimbigner:
A couple of other things.
1. I think Stephan two point are correct as guiding principles.
2. I think the simplest solution is to just implement simple named dimensions
    and allow variable definitions to reference them. Anonymous dimensions
    would continue to exist.

p.s. there is one other thing about named dimensions. If they can be defined
in any group and referenced from a variable in some other group, then some kind
of fully qualified name probably needs to be defined so that it is possible to
unambiguously reference a dimension.

sbalmer:
To add my perspective: We're dealing with NetCDF files a lot and having Zarr support NetCDF could simplify things. We already use `zattrs` named after the NetCDF format.

I do like the simplicity in Zarr and there is value in preserving that. As it looks like the functionality can easily be implemented as a wrapper on top of Zarr I'd opt for that. This way only people interested in NetCDF get exposed to the added complexity.

alimanfoo:
@shoyer thanks for the links and comments re compatibility, very helpful, I'll do some reading. Yes if there are other changes needed to the spec then it may be worth biting the bullet. For reference here are all the other issues I'm aware of that would require some kind of spec change: #267, #244, #216, #111. I'd like to mull this over for a bit. Allowing some mechanism of forwards compatibility seems very sensible in principle, but I'd like to work that all through to understand all the consequences for implementation. And this is not a counterargument, but just noting that .zattrs effectively provides a "preserve but ignore" mechanism already. And regardless of how dimensions ends up getting implemented, we may want some mechanism for standardising metadata conventions like units etc., i.e., things that are effectively optional modules that people may want to mix-and-match.

If we do end up with consensus for a spec change, it would be good to consult the z5py developers, I think they are close to a 1.0 release which implements the zarr v2 spec. And anyone else who is working on or considering an implementation.

On a side point, I would be happy for zarr *not* to implement HDF5 dimension scales. The complexities associated with keeping the bidirectional references in a consistent state seem hard to manage.

DennisHeimbigner:
I second the point about dimension scales. They cause a lot of problems in the
netcdf-4 code. Best to work with simpler concepts.

rsignell-usgs:
So one idea is to define the existing conventions used by xarray as something like "NetZDF-1.0 Conventions for Zarr",  just as other software in the community is enabled by "CF-1.6 Conventions for NetCDF", right?

alimanfoo:
@rsignell-usgs yes I think that's definitely an option.

jakirkham:
Oddly we do have some cases for dimension scales, but I agree this is too complicated for Zarr.

jhamman:
Thanks all for the really productive conversation so far. I really like where this is headed. @alimanfoo [suggested](https://github.com/zarr-developers/zarr/pull/276#issuecomment-405512140) we reach out to the Z5 devs (@constantinpape, @clbarnes) to get their take on the subject. Are we missing input from others that we should be soliciting?

@lmadaus and @kmpaul are two others that have voiced interest in this area.

lmadaus:
Thanks for the tag @jhamman. I've been following this thread with interest and very much support this effort. Almost  our entire usage of Zarr at Jupiter has been through xarray for read/writes, and we've appreciated the xarray-enforced spec that lets us easily go from netCDF-zarr or geoTIF-zarr. We'd like to work towards more direct compatibility between Zarr and netCDF to build toward legacy geophysical models able to output in zarr-like formats, and this seems like a good step in that direction.

ambrosejcarr:
Thanks @mrocklin for ccing me onto this thread. By way of introduction, I'm contributing to a package to process image-based transcriptomics (biological microscopy) data, which is easiest to think about as a bunch of two dimensional tiles that sample the z-dimension of the tissue being imaged for multiple color channels and time points. The tiles can be stuck together as a 5-d tensor. 

We're defining a simple file format for this data which right now is a JSON schema that specifies how to piece the tensor together from a collection of TIFF files stored on a disk or file system, and stores relevant metadata about each tile. This looks a lot like zarr (and Z5), and we're excited by the prospect of contributing to an existing ecosystem instead of rolling our own. 

One concern we had was that the zarr format could be harder for someone to pick up and use _outside_ the zarr/python ecosystem (we expect to have many R users, for example). The addition of column names is a really nice step towards greater expressivity and self-description, so we like this change quite a lot. 

Off-topic, we have some more general questions about the zarr format. Is there someone who would be a good contact? 

tagging @dganguli @freeman-lab @ttung

alimanfoo:
Thanks @ambrosejcarr for getting in touch and sharing your use case, very interesting.

Regarding usage from R, I've raised a separate issue (#279), would be great to explore ways of making that possible. 

If you have general questions please feel free to ask via the issue tracker, no problem if an issue is a bag of different questions.

shoyer:
My current thought for revising this proposal is at the very least the "netZDF" or "Zarr-netCDF format" (no new jargon) should be separated into another doc.

However, I still think that *optional* dimension names as described here (on arrays and groups) could have a place in the Zarr format itself -- assuming we figure out backwards/forwards compatibility concerns.

As for additional conventions themselves (referenced in my draft netZDF spec), I'm thinking that it could make sense to define a group-level `conventions` attribute (either private or public) that should be a mapping from convention names to versions, e.g., `{'conventions': {'netZDF': 1.0, 'CF': 1.6}}`. That's definitely a better choice than mere boolean values, and is nicely extensible by other domain specific conventions without needing to hard wire "netZDF" into the Zarr spec.

shoyer:
> The addition of column names is a really nice step towards greater expressivity and self-description, so we like this change quite a lot.

@ambrosejcarr  could you clarify what you mean by "column names" here? These changes currently only refer to *dimension* names. In principle, *column* names could either be represented by different arrays (with different names) or "column" dimension with an additional array providing coordinate labels. In netCDF files, the typical convention is to use a variable with the same name as its sole dimension: the values of that 1D array provide labels for each point.

ambrosejcarr:
@shoyer Sorry, I was imprecise in my language. If I'm interpreting dimension names properly, they are a vector of values equal in length to the number of dimensions in the array. For our data, that would be `channel`, `imaging_round`, `z`, `y`, and `x`.

clbarnes:
I can't speak for @constantinpape , who is really the architect of z5py while I've just been tinkering round the edges, but I think it's unlikely that a new zarr spec which created a significant divergence from [N5](https://github.com/saalfeldlab/n5) would make it into z5py any time soon - at v1.0, it only supports a subset of the fairly simple (and flexible) N5 spec. @axtimwalde would need to confirm but I would guess that N5 is likely to stay simple, with individual applications defining their own schema for handling more complicated attributes.

This shouldn't discourage you from making these changes, of course - dimension labelling is certainly something which would be helpful to us in the neuroimaging field where the tool set uncomfortably straddles the fortran-ordered java/imageJ and C-ordered python realms, and enforcing dimension parity between raw volumes, labels and so on might be nice too.

The combination of these factors makes me personally lean towards 

> Option 1: Use .zattrs, write this as a separate spec. Full compatibility, old code will be able to read new data, and new code will be able to read old data.

where there is a well-defined way (and probably library) to represent a netCDF-like schema in zarr but it purely uses zarr as a backend rather than being built into the format itself. Naming collisions with user-defined attributes should obviously be avoided; in N5 this is done simply by keeping all of the application-specific optional fields in a dict within the attributes JSON with a name like `"_netZDF"`.

alimanfoo:
Thanks @clbarnes for the comments, much appreciated. 

Thanks also for raising the point about naming collisions, I was just thinking about that. It would be good to figure out some best practices for how metadata conventions are defined and used. I'm more than happy for that conversation to continue here, but I've also surfaced that as a separate issue to give it some prominence: #280.

docs/spec/v2.rst
```diff
@@ -78,6 +78,14 @@ filters
     filters are to be applied. Each codec configuration object MUST contain a
     ``"id"`` key identifying the codec to be used.
 
+The following keys MAY be present:
+
+dimensions
+    A list of string or ``null`` values providing optional names for each ofthe
```
joshmoore: s/ofthe/of the/

rabernat:
Do we want to make a decision about what to do here?

Maybe @wardf and the Unidata team might want to weigh in with their thoughts about the future direction of netCDF / zarr integration?

WardF:
Absolutely. Reviewing now for comment :)

shoyer:
It sounds like the consensus was learning towards making this as a separate "Zarr Dimensions Spec", which we could feature in the Zarr docs.

I'd be happy with that, with my main concern being that we should have an unambiguous way to avoid name collisions being spec attributes and arbitrary user-defined attributes.

WardF:
I would agree with making this as a separate spec as part of the Zarr documentation.  

We are working towards adding a new format in the core C library; the underlying file storage adheres to the Zarr spec (stored either in object storage or on local disk), with a corresponding data-model represented by the intersection of the NetCDF enhanced data model and the Zarr data model.  Zarr already supports the "big" features that are primarily used in the Enhanced file format; compression and chunking.  (As an aside, I was in a meeting yesterday and the topic of alternatives to the enhanced file format in an HPC setting was raised.  I have no idea if Zarr is a potential solution for HPC in a posix environment, yet, but if it is, there seems to be a lot of interest).  

In regards to Zarr/NetCDF integration in the future: I think it would be great to have a defined spec or convention that would be adhered to by those writing netCDF model data using Zarr, as well as those writing model data using the core C library/NetCDF API.  Our thinking has been that by adhering to the functionality provided by Zarr, it would be a matter of documenting this new format so that anybody writing the model data via Zarr would have the information they need to ensure compatibility.  

From a technical standpoint, we need to suss out the architecture of the netCDF-C dispatch layer to map into whatever API we end up using or, as appears more and more likely, implement ourselves.  While we'd love to use `z5`, we need to stick to C-compatible implementations for the time being.  

So in a nutshell, we are very interested in the potential integration of Zarr/NetCDF, and will participate as best we can on the Zarr side of things in addition to focusing on the NetCDF library development.  I'll also try to broadcast our plans a little more clearly; I shared them at a Pangeo workshop last month(?), and at a couple other meetings since then.  I'll see about writing a blog post or finding some other signal-boosting platform on our side.  

Thanks to everybody who tagged me, and then emailed me, and then emailed me again to make sure I stayed focused on weighing in on this, it shouldn't be necessary now that the 2018 workshop is over :).

jakirkham:
Thanks for the update @WardF.

A pure C library that implements the Zarr spec would be a huge win. If this were open sourced with a sufficiently permissive license, I think the community would take to that very quickly; adding various language bindings on top of it. Some discussion along those lines came up in issue ( https://github.com/zarr-developers/zarr/issues/285 ).

FWIW most of the people I know (myself included) use Zarr/N5 for large image data processing in an HPC setting. So this is not only a reasonable path for others using HPC, but one that is being actively exercised.

Sticking with a pure C base implementation makes perfect sense. The z5 route was mainly interesting from the standpoint of getting something simple up and running quickly and then evolving based on community feedback. Starting with a pure C implementation is definitely the more principled way of doing this and has all the benefits that a pure C implementation comes with. (On a side note: The [libxnd]( https://xnd.readthedocs.io/en/latest/libxnd/index.html ) container library seems very useful for this effort.)

Glad to hear we will be seeing you around more. Looking forward to reading the blog post. Thanks again for the update. :)

WardF:
With the caveat that I haven’t spoken with @dennisheimbigner about it, I wonder if having that stand-alone Zarr C library would make the most sense, as opposed to baked-in to the rest of the C library. It feels like a good idea insofar as it would be more likely to enjoy broader adoption.

clbarnes:
>  I wonder if having that stand-alone Zarr C library would make the most sense, as opposed to baked-in to the rest of the C library

+1_000_000

Especially if the specification strategy is going to be a base zarr spec with different projects defining their own attribute schemas on top of it, having a base C implementation that everyone can use with whatever language they choose to build the reference implementation of their schema in, would be of immense value compared to every schema having to work from the ground up.

So long as the base library could be distributed in such a way that this modular approach is easy, that is! I haven't worked with C myself but I understand the packaging and dependency management ecosystems are non-trivial.

I have worked a little with rust, where such concerns are very much trivial. It also seems that generating C bindings at build time is not difficult, and much of the logic could probably be lifted straight from https://github.com/aschampion/rust-n5 .

rabernat:
It's great to see so much support for the C library. I am all in on this.

Who would be a good candidate to actually develop such a library? This is important enough to enough [funded] project that we are not necessarily talking about a volunteer effort.

jhamman:
Also +1 on a stand-alone Zarr C library. I'll advocate that such a library should live in the zarr-developers org and that UNIDATA devs should take an active role in helping build/maintain it. I suspect we'll find other interested groups that can help with the development/maintenance burden. I recognize that is somewhat orthogonal to the business as usual development strategy but I think the long term benefits to UNIDATA and other Zarr communities will pay off.

WardF:
The ownership and maintenance of any project certainly merits discussion.  It would be fantastic if there were an externally-owned project that we could contribute to; there's no inherent reason it should be a Unidata product.  There would be questions about how much responsibility we can take for non-Unidata products in the long term, but in the immediate future it is work that needs to be done regardless of who owns the product.  It is certainly something we would be interested in collaborating on.  Any thoughts, @DennisHeimbigner ?

WardF:
Some stream-of-consciousness thoughts about a stand-alone library, and what we need from external libraries adopted for use in netCDF.  

1. The licensing is a big thing; BSD 3-Clause is what we're standardizing on here for many of our products (stay tuned for a related announcement with the next netCDF release). 
2. Can be built in a cross-platform manner, for at least the big 3 (Linux, OSX, Windows+Visual Studio).   This means cmake-based build configuration to support the latter case, although not necessarily exclusively; autoconf-based tools still work well enough on Linux & OSX.
3. Longevity concerns; to be honest, if this is work we participate in (and we are interested in collaborating), and the owning organization makes changes either to the licensing or goes away, Unidata will need to be in a position to take ownership of the library in the absence of a good alternative.  This doesn't seem likely of course, but it is on our minds given recent developments with other libraries we depend on. 

No point to these other than they are considerations from the netCDF team, moving forward with any solution.

mrocklin:
From @jhamman 
> I'll advocate that such a library should live in the zarr-developers org and that UNIDATA devs should take an active role in helping build/maintain it

From @WardF 
> The ownership and maintenance of any project certainly merits discussion. It would be fantastic if there were an externally-owned project that we could contribute to; there's no inherent reason it should be a Unidata product. There would be questions about how much responsibility we can take for non-Unidata products in the long term, but in the immediate future it is work that needs to be done regardless of who owns the product

From @WardF 
> Longevity concerns; to be honest, if this is work we participate in (and we are interested in collaborating), and the owning organization makes changes either to the licensing or goes away, Unidata will need to be in a position to take ownership of the library in the absence of a good alternative. This doesn't seem likely of course, but it is on our minds given recent developments with other libraries we depend on.

From my perspective if this project is likely to be mostly used by the earth science community then I have no concern with Unidata owning the code.  They've proven to be effective stewards in the past and have a longer time horizon than typical OSS groups.  My only concern here would be that people outside of Unidata would need an easy to way to also maintain some level of control and activity.  [This is hard.](http://matthewrocklin.com/blog/work/2018/08/21/institutional-open-source)

However, if this project is likely to be used by groups outside of the earth science community like @jakirkham 's imaging groups or @ambrosejcarr 's genomics/bio groups then I would suggest that the project be legally owned by a stewardship organization like NumFOCUS, but provide permissive commit/repo-owernship rights to at least one representative of each of the scientific domains so that no one gets blocked by the inaction of others should they choose to take a different path in the future.


Regarding @WardF 's other points, I don't anticpate permissive OSS licenses or cross-platform build objectives to be contentious issues in this crowd.

jakirkham:
This discussion is great! Thanks all for sharing your thoughts.

Am thinking we should probably segue into a different thread to discuss the pure C implementation of Zarr. Have raised issue ( https://github.com/zarr-developers/zarr/issues/317 ) for this purpose. Sorry for not doing this sooner. Look forward to discussing with you over there.

jakirkham:
Should have added in regards to @mrocklin's point about organizational structure. We have been discussing this broadly w.r.t. Zarr in issue ( https://github.com/zarr-developers/zarr/issues/291 ) and more detailed discussions on specific points have been broken out from there.

ambrosejcarr:
> However, if this project is likely to be used by groups outside of the earth science community like @jakirkham 's imaging groups or @ambrosejcarr 's genomics/bio groups

Checking back in here -- I'm prototyping a zarr-based library for our imaging project now, and we've already swapped our sequencing-based output data to zarr. It seems very likely that we will use zarr in production for at least one aspect of our project. I'll chime in on #291 and will likely have a few feature requests in the next few months. 👍

jhamman:
I suggest we close this as it is now quite stale. The v3 spec conversation, along with ZEP 4 and NCZarr have surpassed this design. Feel free to reopen if there is more to do here.

PR #306, opened 2018-10-11
LRU cache for decoded chunks 

shikharsg:
Issue: [#278](https://github.com/zarr-developers/zarr/issues/278)

The `ChunkCache` implemented here is an almost identical copy of the `LRUStoreCache` class, but with the keys cache functionality removed(same for its tests).

I was hesitant to implement it as a super class of `LRUStoreCache` as the two are supposed to be used in different places. While `LRUStoreCache` is like any other storage class of zarr and can be wrapped around any storage class of zarr, `ChunkCache` is specifically to be used for storage of decoded chunks, and to be passed to an `Array`.

I have implemented both read-write cache, i.e. in addition to caching when one reads from the array, if one writes to a certain chunk, instead of invalidating that chunk in the `ChunkCache`(which would happen if one wanted only read cache), that chunk being written to is cached in the ChunkCache object. There is a problem with this approach as the below 3 tests are failing. If we first write data to zarr array, and then we read it, and when not using `ChunkCache`(as implemented above) it always goes through the encode phase, and then the decode phase when we write and subsequently read from the array. If `ChunkCache` is used, with write cache as implemented here, it's possible that it does not go through the encode-decode phase when we write and subsequently read the data. Now for the following three tests to pass, it is imperative that it go through the encode-decode phase when we write and then subsequently read from the array:
- [TestArray.test_dtypes](https://github.com/zarr-developers/zarr/blob/master/zarr/tests/test_core.py#L858)
- [TestArray.test_object_arrays](https://github.com/zarr-developers/zarr/blob/master/zarr/tests/test_core.py#L893)
- [TestArraytest_object_arrays_vlen_array](https://github.com/zarr-developers/zarr/blob/master/zarr/tests/test_core.py#L1017)

Tracebacks can be seen in Travis CI

I wonder if there is a way to get around this while keeping write cache implemented, which makes me think if I should implement write cache at all? The above 3 tests will pass if we implement only read cache.

Finally, the logic of chunk cache as implemented in the `_chunk_getitem` function is making that function much more hard to read than it already was. I could refactor for better readability if it is so desired.

TODO:
* [x] Unit tests and doctests pass locally under Python 3.6 (e.g., run ``tox -e py36`` or 
  ``pytest -v --doctest-modules zarr``)
* [x] Unit tests pass locally under Python 2.7 (e.g., run ``tox -e py27`` or
  ``pytest -v zarr``)
* [x] PEP8 checks pass (e.g., run ``tox -e py36`` or ``flake8 --max-line-length=100 zarr``)
* [x] Doctests in tutorial pass (e.g., run ``tox -e py36`` or ``python -m doctest -o NORMALIZE_WHITESPACE -o ELLIPSIS docs/tutorial.rst``)
* [x] Test coverage to 100% (Coveralls passes)


zarr/core.py
```diff
@@ -103,7 +103,8 @@ class Array(object):
     """
 
     def __init__(self, store, path=None, read_only=False, chunk_store=None,
-                 synchronizer=None, cache_metadata=True, cache_attrs=True):
+                 synchronizer=None, cache_metadata=True, cache_attrs=True,
+                 chunk_cache=None):
```
alimanfoo: Need to add chunk_cache to the parameters section in the class docstring above.

zarr/storage.py
```diff
@@ -1883,3 +1883,121 @@ def __delitem__(self, key):
         with self._mutex:
             self._invalidate_keys()
             self._invalidate_value(key)
+
+
+class ChunkCache(MutableMapping):
```
alimanfoo: Maybe should rename this to "LRUChunkCache" just to be explicit about caching strategy.

zarr/storage.py
```diff
@@ -1883,3 +1883,121 @@ def __delitem__(self, key):
...
+    """Storage class that implements a least-recently-used (LRU) cache for array chunks.
```
alimanfoo: Edit "Storage class that implements ..." to "Class that implements ..." or even just "A least-recently-used (LRU) cache ..." (as this isn't a storage class).

zarr/storage.py
```diff
@@ -1883,3 +1883,121 @@ def __delitem__(self, key):
...
+    Intended primarily for use with stores that can be slow to access, e.g., remote stores that
+    require network communication to store and retrieve data.
```
alimanfoo: Maybe add on the end of this sentence ", and/or arrays where decompression of data is computationally expensive.".

zarr/storage.py
```diff
@@ -1883,3 +1883,121 @@ def __delitem__(self, key):
...
+        _, v = self._values_cache.popitem(last=False)
+        return v
+
+    def _accommodate_value(self, value_size):
+        if self._max_size is None:
+            return
+        # ensure there is enough space in the cache for a new value
+        while self._current_size + value_size > self._max_size:
+            v = self._pop_value()
+            self._current_size -= buffer_size(v)
```
alimanfoo: Maybe this could be ``v.nbytes`` instead of ``buffer_size(v)`` to work around the test failure regarding 'M' dtypes. Or even push that down into the ``buffer_size()`` function, i.e., a special case to detect numpy arrays and return ``.nbytes``.

zarr/storage.py
```diff
@@ -1883,3 +1883,121 @@ def __delitem__(self, key):
...
+
+    def _cache_value(self, key, value):
+        # cache a value
+        value_size = buffer_size(value)
```
alimanfoo: See comment above regarding ``.nbytes`` instead of ``buffer_size()``.

zarr/storage.py
```diff
@@ -1883,3 +1883,121 @@ def __delitem__(self, key):
...
+
+    def invalidate_values(self):
+        """Clear the values cache."""
+        with self._mutex:
+            self._values_cache.clear()
+
+    def _invalidate_value(self, key):
+        if key in self._values_cache:
+            value = self._values_cache.pop(key)
+            self._current_size -= buffer_size(value)
```
alimanfoo: See comment above regarding ``.nbytes`` instead of ``buffer_size()``.

alimanfoo:
Regarding the msgpack failure, maybe that test needs to be skipped for the array tests with chunk cache, the behaviour with msgpack is broken and in fact correct when using a chunk cache.

Regarding the test failure about mismatching dtypes, I don't know what's causing that. Maybe worth revising the implementation of _chunk_getitem() as suggested above and seeing if it persists.

Seems worth pursuing the support for write cache implementation, should be a way to deal with these test issues.

zarr/core.py
```diff
@@ -1714,6 +1740,8 @@ def _chunk_setitem_nosync(self, chunk_coords, chunk_selection, value, fields=Non
             else:
                 chunk[chunk_selection] = value
 
+        if self._chunk_cache is not None:
+            self._chunk_cache[ckey] = np.copy(chunk)
```
shikharsg: To solve the test_object_arrays_vlen_array failure instead of the following two lines:
```python
        if self._chunk_cache is not None:
            self._chunk_cache[ckey] = np.copy(chunk) 
```

I could do this(have tried and test passes):
```python
        import numcodecs  # will put this import up of course
        if self._chunk_cache is not None:
            if self._filters:
                for f in self._filters:
                    if type(f) is numcodecs.VLenArray:
                        chunk = f.decode(f.encode(chunk))
            self._chunk_cache[ckey] = np.copy(chunk) 
```
This does look quite messy though.

We could do the above or we could expect the user to provide correct dtypes of arrays when using write cache and VLenArray codec, and skip the test in the build. I'm not sure how else to resolve this test failure.

alimanfoo:
Thanks @shikharsg. The test_object_arrays_vlen_array situation is tricky. It exposes a potential difficulty in general with object arrays and caching on write. I can see a few possible options at the moment...

Option 1a. Don't support caching on write at all. 

Option 1b. Don't support caching on write for object arrays.

Option 2. Relax the test. 

Option 3. Round-trip the chunk through encode/decode before caching.

FWIW I would be fine with option 1a or 1b if you want to get the read caching in now and leave write caching for later. The primary use cases Zarr targets are write-once-read-many so read caching is the priority. 

I think I'd also be OK with option 2. However, there is another issue that I just realised would need to be addressed, and that is that during write, saving the chunk to the cache occurs too early, and there is a possibility that the chunk could get cached but a failure could subsequently occur either during encoding or storing. This could arise, for example, if an object array is passed but that cannot be encoded (e.g., as a vlen array). This could be addressed by moving the call to set the chunk in the cache to be the very last statement inside the method. I.e., the end of Array._chunk_setitem_nosync() becomes:

```python
        # encode chunk
        cdata = self._encode_chunk(chunk)

        # store
        self.chunk_store[ckey] = cdata

        # cache the chunk
        if self._chunk_cache is not None:
            self._chunk_cache[ckey] = np.copy(chunk)
```

Option 3 also seems OK for object arrays. There is a way this could be done that would handle a wider range of scenarios for object arrays, if the end of Array._chunk_setitem_nosync() becomes:

```python
        # encode chunk
        cdata = self._encode_chunk(chunk)

        # store
        self.chunk_store[ckey] = cdata

        # cache the chunk
        if self._chunk_cache is not None:
            if self._dtype == object:
                # ensure cached chunk has been round-tripped through encode/decode
                chunk = self._decode_chunk(cdata)
            self._chunk_cache[ckey] = np.copy(chunk)
```

Note that this still would mean that using a chunk cache would slow down writing of object arrays, because it would incur an extra decode as chunks are written.

shikharsg:
I have moved write cache to the very last statement inside the method. I have also implemented caching for 0-d arrays

About the test_object_arrays_vlen_array problem, option 3(but just for dtype=object arrays) seems best to me and which I have also implemented as you have suggested. We might want to document the slowdown for write-cache object arrays somewhere. Which part of the docs should I put this in?

zarr/core.py
```diff
@@ -692,19 +698,36 @@ def _get_basic_selection_zd(self, selection, out=None, fields=None):
...
+                chunk = np.zeros((), dtype=self._dtype)
+                if self._fill_value is not None:
+                    chunk.fill(self._fill_value)
+
+            else:
+                chunk = self._decode_chunk(cdata)
+
+        # cache decoded chunk
+        if self._chunk_cache is not None:
+            self._chunk_cache[ckey] = chunk
```
alimanfoo: I think this block should go under ``if chunk is None:``, i.e., only need to cache the chunk if it wasn't already cached.

shikharsg: Don't we need to reset the value to preserve the least-recently-used property of the cache?

shikharsg:
After getting a few bugs in LRUChunkCache and seeing that it didn't have enough tests I figured that some of the tests from the base class StoreTests could be used for LRUChunkCache, but not all tests. So I separated the methods of StoreTests into two classes MutableMappingStoreTests and StoreTests, the latter inheriting the former. StoreTests, in addition to the methods of MutableMappingStoreTests , has test_hierarchy and the test_init* methods. Also LRUChunkCache inherits MutableMappingStoreTests. Doing this also let me easily fix a few bugs in LRUChunkCache. I hope this is not too big of a change and if you would like to shift any methods from StoreTests to MutableMappingStoreTests or vice versa, or even entirely revert this change do let me know.

I would like to work the the _chunk_getitem() simplification for a little more time, if that's okay. Will get back on this soon.

I would certainly go with documenting this feature as experimental which allow us to make API changes if needed. Do let me know where I could document this.

alimanfoo:
Thanks for pushing forward on this. I'm out of radio contact for two weeks
now, but I think this is moving in a good direction, please feel free to
continue as you see fit. Will be great if we can try to keep code as simple
and easy to understand as possible, with good separation of concerns and
minimal code duplication.

alimanfoo:
Maybe we should be looking to pull out common code from the two LRU classes into a common base class e.g. LRUMappingCache. Might also help with some refactoring of tests.

shikharsg:
Apologies for the late reply. I have pulled out some common code from the two classes into LRUMappingCache. You can see in a PR here: https://github.com/shikharsg/zarr/pull/3/files. I haven't committed it to the PR branch yet, as I'm not sure if that's what you're asking for. So the LRUMappingCache will be like an Abstract Base Class which is not supposed to be used anywhere but as a super class? If so, is whatever I have done in the link above correct?

About refactoring of the tests (https://github.com/zarr-developers/zarr/pull/306/commits/6fac2adc4ab20289e84c254a3cb259a8717b4a14), so that some of the StoreTests can be used for LRUChunkCache, is that ok? Or should I revert that?

shikharsg:
@alimanfoo, I have slightly simplified the _chunk_getitem function and excluded the optimization block from dealing with the chunk cache case.

Apart from refactoring of the store tests, is there anything left to do? Otherwise I think this would be good to go in?

shikharsg:
Hi all. Sorry for the massive delays on this.

I have refactored the LRUStoreCache and LRUChunkCache to pull out common code between them. I have also done the same for the tests of LRUStoreCache and LRUChunkCache.

@alimanfoo could you give a review to see if your concerns have been addressed?

Other thoughs @zarr-developers/core-devs ?

mcherkassky:
Wondering when this will get released? Saw that it almost made it in to 2.3.2 - maybe next one?

jobh:
It looks like the `write_direct` path in `_chunk_getitem` prevents caching in many cases. Maybe disable `write_direct` if both a cache and a compressor is attached.

shikharsg:
> It looks like the `write_direct` path in `_chunk_getitem` prevents caching in many cases. Maybe disable `write_direct` if both a cache and a compressor is attached.

At the moment, this is the behavior yes. I guess it's not being picked up by tests, because there are no tests for this case. Will try adding some tests for this soon

jakirkham:
Am curious what the state of things are here. Is more work needed? Does it need a review? 🙂

shikharsg:
@jakirkham most of the code in this PR has been reviewed by @alimanfoo.

There were some concerns about refactoring the `LRUStoreCache` and the `LRUChunkCache` code(and the related tests) to pull out the common logic from both. I have done that in the last few commits but this hasn't been reviewed yet. Would be great if this can be reviewed.

jakirkham:
Do we happen to know why the tests are failing? Or is that expected at this stage?

shikharsg:
> 
> 
> Do we happen to know why the tests are failing? Or is that expected at this stage?

If

shikharsg:
> 
> 
> Do we happen to know why the tests are failing? Or is that expected at this stage?

Sorry, that was just because of one failing doctest which I have now fixed.

joshmoore:
Hi @shikharsg, I tried my hand at merging origin/master into your branch. The diff looks sensible but the new code path at https://github.com/zarr-developers/zarr-python/pull/306/commits/8c8a69f002e20c6a0af35afaae83cb48e7286d53#diff-f640a0c570765203632323d71a9d1145ec6121c90db73f56e1271ab0efd5b8d8R1632 might need looking into. All the best, ~Josh

shikharsg:
Thanks @joshmoore. Turned out be an easy fix actually.

zarr/core.py
```diff
@@ -739,19 +750,36 @@ def _get_basic_selection_zd(self, selection, out=None, fields=None):
...
+        chunk = None
 
-        else:
-            chunk = self._decode_chunk(cdata)
+        # check for cached chunk
+        if self._chunk_cache is not None:
+            try:
+                chunk = self._chunk_cache[ckey]
+            except KeyError:
+                pass
```
jakirkham, line 764: This might also be combined with the code below somehow

```suggestion
            chunk = self._chunk_cache.get(ckey)
```

shikharsg: yes certainly looks simpler

joshmoore:
What's the next step here?

shikharsg:
In terms of functionality, it's all working. But there were some concerns regarding refactoring the code. So this PR needs a detailed review. Otherwise it's good to go in.

martindurant:
Quick question: does this layer respect get/setitems for concurrent access with fsspec? 

I ask because fsspec, of course, has its own idea of caching. That is file-oriented rather than chunk-wise, but since ReferenceFileSystem, chunks of other remote URLs can be regarded as standalone files too. It would be nice to have fsspec's caching layer also respect LRU (i.e., number of entries, or size of cache). That is orthogonal and shouldn't delay this PR any further.

niowniow:
Any chance this can be reviewed? After quite some profiling I realized that LRUStoreCache does not offer the performance benefits I expected, because of this decoding issue.

docs/tutorial.rst
```diff
@@ -849,10 +849,34 @@ store. E.g.::
     b'Hello from the cloud!'
     0.0009490990014455747
 
-If you are still experiencing poor performance with distributed/cloud storage,
-please raise an issue on the GitHub issue tracker with any profiling data you
-can provide, as there may be opportunities to optimise further either within
-Zarr or within the mapping interface to the storage.
+The above :class:`zarr.storage.LRUStoreCache` wraps any Zarr storage class, and stores
+encoded chunks. So every time cache is accessed, the chunk has to be decoded. For cases
```
joshmoore, line 853:
```suggestion
encoded chunks. Every time the cache is accessed, the chunk must be decoded. For cases
```

zarr/core.py
```diff
@@ -76,6 +76,15 @@ class Array:
         read and decompressed when possible.
 
         .. versionadded:: 2.7
+    chunk_cache: MutableMapping, optional
+        Mapping to store decoded chunks for caching. Can be used in repeated
+        chunk access scenarios when decoding of data is computationally
+        expensive.
+        NOTE: When using the write cache feature with object arrays(i.e.
```
joshmoore, line 83:
```suggestion
        NOTE: When using the write cache feature with object arrays (i.e.
```

zarr/creation.py
```diff
@@ -51,6 +51,15 @@ def create(shape, chunks=True, dtype=None, compressor='default',
     chunk_store : MutableMapping, optional
         Separate storage for chunks. If not provided, `store` will be used
         for storage of both chunks and metadata.
+    chunk_cache: MutableMapping, optional
+        Mapping to store decoded chunks for caching. Can be used in repeated
+        chunk access scenarios when decoding of data is computationally
+        expensive.
+        NOTE: When using the write cache feature with object arrays(i.e.
```
joshmoore, line 58:
```suggestion
        NOTE: When using the write cache feature with object arrays (i.e.
```

zarr/creation.py
```diff
@@ -436,6 +447,15 @@ def open_array(
         read and decompressed when possible.
 
         .. versionadded:: 2.7
+    chunk_cache: MutableMapping, optional
+        Mapping to store decoded chunks for caching. Can be used in repeated
+        chunk access scenarios when decoding of data is computationally
+        expensive.
+        NOTE: When using the write cache feature with object arrays(i.e.
```
joshmoore, line 454:
```suggestion
        NOTE: When using the write cache feature with object arrays (i.e.
```

joshmoore (commented):
Few minor wording suggestions on a quick read. There are also now a few conflicts with the mainline.

At a very high-level, my biggest question would be whether or not this could have been a concern of the LRUStoreCache **IF** it had been possible to more reliably detect which key represents a chunk.

niowniow:
I think the issue is mainly that the `Array` class is responsible for decompressing. Is there a reason the de/compressor is implemented this way? I haven't checked through the code in detail yet. Please bear with me if the following is nonsense:

My idea basically is to move all functionality of the compressor out of the `Array` class.
As alternative one could  implement a compressor that can be wrapped around a store, similar to how the  `LRUStoreCache` is wrapped around it. then we could e.g.  `store =LRUStoreCache(CompressorWrapper(DirectoryStore)))` and the `LRUChunkCache` wouldn't be necessary. Or is there another benefit from using the `LRUChunkCache`? 

In the `Array` class we could remove `__compressor` and instead wrap the store with `CompressorWrapper` if there is no `CompressorWrapper` in the store chain. I guess the default would be to wrap the store unless the user explicitly disables it because they have wrapped it themselves. All parameters given to `Array` can just be propagated through to its `CompressorWrapper` instance

niowniow:
I had another look at this PR. There is one corner-case not covered. If `write_direct` ([here](https://github.com/shikharsg/zarr/blob/8b6a6999bd3972f545b3f719cdf64e32c3fa8454/zarr/core.py#L1745)) is true, i.e. we request one whole chunk. It will not be cached. I don't know if that was intended.

joshmoore:
@ericpre: from https://github.com/hyperspy/hyperspy/pull/2798#issuecomment-916060062, do I understand correctly that you've tested this and it's working for you?

ericpre:
> @ericpre: from [hyperspy/hyperspy#2798 (comment)](https://github.com/hyperspy/hyperspy/pull/2798#issuecomment-916060062), do I understand correctly that you've tested this and it's working for you?

Yes, I have tested this branch and I can confirm the speed improvement. If you would like to have more details, I would need to check again but I would not have time to do before a week or so.

jakirkham:
Probably fixing merge conflicts is the first thing. Agree this is something we'd like to see included.

claesenm:
I’m also quite interested in this feature but it seems the PR has gone stale. Are there any remaining open issues that are to be addressed? Having gone through the thread, it looks ready for merging.

joshmoore:
Hi @claesenm. The primary issue will be the merge conflicts:

```
zarr/__init__.py
zarr/core.py
zarr/creation.py
zarr/hierarchy.py
zarr/storage.py
zarr/tests/test_storage.py
```

If you or anyone else could try to create a merge commit, that would certainly help. (Thanks!)

jhamman:
For those following this issue, I think the most important thing right now is to look at #1583 and help us think about how this functionality should be supported in zarr-python's 3.0 api. There's a lot of surface area in this PR I want to make sure we cover the functionality going forward if its important.

jhamman:
Closing this out as stale. We'll keep the feature request open.

Issue #272, opened 2018-07-06
Race condition in storage.py for NestedDirectoryStore writes


jmswaney:
I've encountered a `FileExistsError` when writing in parallel to a `NestedDirectoryStore` that says:

`FileExistsError: [Errno 17] File exists: '/media/jswaney/Drive/Justin/coregistration/whole_brain_tde/fixed/zarr/4/10'`

The error points to the `os.makedirs(dir_path)` line in `__setitem__` for the `DirectoryStore` class:

```python
dir_path, file_name = os.path.split(file_path)
if os.path.isfile(dir_path):
    raise KeyError(key)
if not os.path.exists(dir_path):
    try:
        os.makedirs(dir_path)  # Error raised here
    except Exception:
        raise KeyError(key)
```

I was doing the chunk compression with 12 workers, so I think one of them created the subdirectory while another had already checked that it didn't exist. When it got to `os.makedirs`, it gave the expected error. A quick fix would just be to catch this error, but a better fix might be to just make all the directories for a `NestedDirectoryStore` ahead of time.

alimanfoo:
Thanks Justin. On Python 3 I think we could just do os.makedirs(...,
exist_ok=True). Not available in Python 2 unfortunately.

There's also another race condition issue in DirectoryStore, would be good
to get both resolved, could be addressed in same PR:
https://github.com/zarr-developers/zarr/issues/263

jmswaney:
It looks like there are a couple options for the Python 2 vs 3 problem listed [here](https://stackoverflow.com/questions/273192/how-can-i-safely-create-a-nested-directory-in-python?rq=1). They mention the `exist_ok=True` option in the context of the `pathlib` module which has been backported to Python 2 in the `pathlib2` module, but it might be weird to use this OO-style path handling just for this though. Another option is to catch the general `OSError` and just throw an error if the directory still doesn't exist.

jakirkham:
It sounds like we are already interested in using `pathlib`. ( https://github.com/zarr-developers/zarr/issues/261 ) Maybe using the [`pathlib`-based solution]( https://stackoverflow.com/a/45547008 ) will work for us and we can add [`pathlib2`]( https://pypi.org/project/pathlib2/ ) as a requirement.

jakirkham:
Note: Work on this is currently being pursued in PR ( https://github.com/zarr-developers/zarr/pull/318 ).

PR #318, opened 2018-10-25
#272 Fix race condition in DirectoryStore

jmswaney:
Multiple workers may attempt to create subfolders when using `NestedDirectoryStore`, leading to `FileExistsError` raised. This fix catches `OSError` (for python 2 support) and checks if the directory already exists--if it does, then no exception is raised.

Fixes https://github.com/zarr-developers/zarr/issues/272


zarr/storage.py
```diff
@@ -758,6 +758,9 @@ def __setitem__(self, key, value):
         if not os.path.exists(dir_path):
             try:
                 os.makedirs(dir_path)
+            except OSError:
+                if not os.path.isdir(dir_path):
+                    raise KeyError(key)
             except Exception:
                 raise KeyError(key)
```
jakirkham: Hmm...so this `except` case doesn't catch it?

jmswaney: The `except Exception` block was catching it before, but it was just throwing `KeyError` instead--so my code would still error out. The block that I added only raises `KeyError` if the folder doesn't exist already. This way, workers can go on with their business if the folder already exists.

jakirkham: Should we then drop the `except Exception` entirely and only use `except OSError`? It seems bad to mask other errors by raising `KeyError` instead.

jmswaney: I think that's a good idea. I wasn't sure about that part so I left it, but I'm not sure what other exceptions you can even expect from `os.makedirs`. If you got `OSError` because of permissions or something else that prevented making the folder, then the block that I added would raise `KeyError` anyway.

jakirkham: Good point. We can improve this a bit then by checking the `errno` of the `OSError`. Namely if we find `e.errno == errno.EEXIST`, then we know this is a `FileExistsError`. If it's any other `errno`, we can just re-raise the exception as is.

ref: https://docs.python.org/3/library/errno.html#errno.EEXIST

jakirkham: Have taken the liberty of pushing a commit to your branch using the `errno` bit. Please let me know what you think.

jakirkham: > I wasn't sure about that part so I left it...

That makes sense. Honestly this is an orthogonal issue. So we need not address it here. Will raise an issue to follow-up on this after.

jakirkham:
@alimanfoo, when you have a chance, could you please give this a quick look? Shouldn't take too long I hope. :)

zarr/storage.py
jakirkham: Went ahead and dropped the generic exception case as it didn't seem to be covered. Happy to revisit if we figure out it was needed for some reason.

jakirkham:
Thanks @jmswaney!

jakirkham:
Have added the release note to PR ( https://github.com/zarr-developers/zarr/pull/409 ). If there are any other issues, please let us know and we can follow-up on them.

PR #327, opened 2018-11-08
avoid race condition during chunk write

### LINKED ISSUE #263: remove() before os.rename() breaks thread-safety
This section in [storage.py](https://github.com/zarr-developers/zarr/blob/master/zarr/storage.py#L773) breaks thread/process-safety:

```python
# move temporary file into place
if os.path.exists(file_path):
    os.remove(file_path)
os.rename(temp_path, file_path)
```
Problem is, a "reader" process may fail because the file ``file_path`` is non-existent from time to time.

The ``os.path.exists()`` check was introduced in order to fix tests on Windows:
https://github.com/zarr-developers/zarr/blame/fe1c120930ebd9c529d3286bfe88500f8edd6620/zarr/storage.py#L209

Possible solutions:
* Execute ``os.remove(file_path)`` on Windows only.
* Use [os.replace()](https://docs.python.org/3.5/library/os.html#os.replace) instead of ``os.rename``. This breaks Python 2.7 compatibility, though.

Both solutions probably do not guarantee atomic renaming on Windows.

sbalmer:
When the chunk file is first removed before the new version
is moved into place, racing reads may encounter a missing chunk.

Using rename() or replace() without remove() avoids the issue
on Posix-Systems as the methods are atomic. The fallback of
remove() -> rename() is included for Windows pre Python 3.3.

Fixes #263

TODO:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 from __future__ import absolute_import, print_function, division
 from setuptools import setup
+import sys
 
 
 DESCRIPTION = 'An implementation of chunked, compressed, ' \
@@ -8,6 +9,16 @@ DESCRIPTION = 'An implementation of chunked, compressed, ' \
 
 with open('README.rst') as f:
     LONG_DESCRIPTION = f.read()
+
+dependencies = [
+    'asciitree',
+    'numpy>=1.7',
+    'fasteners',
+    'numcodecs>=0.5.3',
+]
+
+if sys.version_info < (3, 3) and sys.platform == "win32":
+    dependencies.append('pyosreplace')
 
 setup(
     name='zarr',
@@ -22,13 +33,7 @@ setup(
         'setuptools>18.0',
         'setuptools-scm>1.5.4'
     ],
-    install_requires=[
-        'asciitree',
-        'numpy>=1.7',
-        'fasteners',
-        'numcodecs>=0.5.3',
-        "pyosreplace : python_version < '3.3' and sys.platform == 'win32'",
-    ],
+    install_requires=dependencies,
     package_dir={'': '.'},
     packages=['zarr', 'zarr.tests'],
     classifiers=[
--- a/setup.py
+++ b/setup.py
@@ -27,6 +27,7 @@ setup(
         'numpy>=1.7',
         'fasteners',
         'numcodecs>=0.5.3',
+        "pyosreplace : python_version < '3.3' and sys.platform == 'win32'",
     ],
     package_dir={'': '.'},
     packages=['zarr', 'zarr.tests'],
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -772,9 +772,18 @@ class DirectoryStore(MutableMapping):
                 f.write(value)
 
             # move temporary file into place
-            if os.path.exists(file_path):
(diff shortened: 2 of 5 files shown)


zarr/storage.py
```diff
@@ -54,6 +53,24 @@
...
+    replace = os.replace
+else:
+    # In Python < 3.3 use rename().
+    if os.name == 'nt':
+        def replace(old, new):
+            if os.path.exists(new):
+                # On windows, rename() can't overwrite files. So
+                # the file is removed first.
+                os.remove(new)
+            os.rename(old, new)
```
jakirkham: So this is the case where we would want the `pyosreplace` package. It's a Windows only backport for `os.replace` on Python 2.

alimanfoo:
Just to mention that there are some potential consequences from implementing this PR that we should be aware of, described in #328. I don't think that should stop us from moving ahead with this PR, using os.replace() is clearly a better thing to do under any circumstances, but just something to be aware of.

sbalmer:
@jakirkham I've made a guess at the `setup.py` changes needed for `pyosreplace`. I couldn't get environmental markers to wrok. See previous commit. Could you please have a look at it? And is something else needed?

setup.py
```diff
@@ -27,6 +27,7 @@
         'numpy>=1.7',
         'fasteners',
         'numcodecs>=0.5.3',
+        "pyosreplace : python_version < '3.3' and sys.platform == 'win32'",
```
jakirkham: Some examples show this as being a semicolon instead.

jakirkham:
Think you had it basically right. Unfortunately there is some lack of clarity about when a colon or semicolon should be used. IIUC it should be a semicolon.

setup.py
sbalmer: Ah, that explains why the env markers didn't work. If nobody complains I'll leave the later commit.

alimanfoo:
Thanks @sbalmer. Could you add pyosreplace to requirements.txt and requirements_dev.txt, pinned to the latest version in the latter. I guess environment markers are needed in both.

sbalmer:
@alimanfoo `pyosreplace` is now in the requirement files too. I've made a guess at how the env markers should look. If I invert the predicates then `pip install -r requirements.txt` tries to install the package. Which means I have some confidence it should work.

jakirkham:
I wonder if we can just drop the `try...finally` now.

alimanfoo:
> I wonder if we can just drop the `try...finally` now.

FWIW I imagine it could still be possible that some failure occurs during the attempt to write to the temporary file (e.g., device gets full or something like that), in which case an exception occurs before the value has been fully written, in which case we'd still want to clean up the temporary file.

jakirkham:
Good point.

alimanfoo:
I've resolved conflicts after merging #352 and added a release note. Will merge if CI passes.

alimanfoo:
@sbalmer apologies I didn't know your full name to include in the release note, you're credited currently just as "sbalmer", happy to leave it that way or add your full name, whichever you prefer.

jakirkham:
Thanks @sbalmer 😄

sbalmer:
Thanks @alimanfoo, @jakirkham!

Now we can use mainline again :-)

Issue #324, opened 2018-11-06
MsgPack codec is broken for array of bytes objects


alimanfoo:
Currently the tests involving the msgpack codec expect broken behaviour where the codec converts an array of bytes objects to an array of str (unicode) objects. Relevant line of code is here: 

https://github.com/zarr-developers/zarr/blob/a4d5656e3042216308b03ab7514567139d4004fc/zarr/tests/test_core.py#L985

This is causing problems elsewhere, e.g., https://github.com/zarr-developers/zarr/issues/278#issuecomment-425006955.

jakirkham:
Went ahead and added this to the v2.3 milestone so we can track it. Though if we decide to punt on it, we can always re-milestone it.

alimanfoo:
Thank you, yes forgot I had meant to include this in 2.3.

PR #347, opened 2018-11-30
Bump Numcodecs requirement to 0.6.1

jakirkham:
As there are some critical fixes in the latest Numcodecs, this bumps our lower bound to the latest version.

TODO:

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -176,7 +176,7 @@ print some diagnostics, e.g.::
     Read-only          : False
     Compressor         : Blosc(cname='zstd', clevel=3, shuffle=BITSHUFFLE,
                        : blocksize=0)
-    Store type         : builtins.dict
+    Store type         : zarr.storage.DictStore
     No. bytes          : 400000000 (381.5M)
     No. bytes stored   : 3242241 (3.1M)
     Storage ratio      : 123.4
@@ -268,7 +268,7 @@ Here is an example using a delta filter with the Blosc compressor::
     Read-only          : False
     Filter [0]         : Delta(dtype='<i4')
     Compressor         : Blosc(cname='zstd', clevel=1, shuffle=SHUFFLE, blocksize=0)
-    Store type         : builtins.dict
+    Store type         : zarr.storage.DictStore
     No. bytes          : 400000000 (381.5M)
     No. bytes stored   : 648605 (633.4K)
     Storage ratio      : 616.7
@@ -1181,7 +1181,7 @@ ratios, depending on the correlation structure within the data. E.g.::
     Order              : C
     Read-only          : False
     Compressor         : Blosc(cname='lz4', clevel=5, shuffle=SHUFFLE, blocksize=0)
-    Store type         : builtins.dict
+    Store type         : zarr.storage.DictStore
     No. bytes          : 400000000 (381.5M)
     No. bytes stored   : 6696010 (6.4M)
     Storage ratio      : 59.7
@@ -1195,7 +1195,7 @@ ratios, depending on the correlation structure within the data. E.g.::
     Order              : F
     Read-only          : False
     Compressor         : Blosc(cname='lz4', clevel=5, shuffle=SHUFFLE, blocksize=0)
-    Store type         : builtins.dict
+    Store type         : zarr.storage.DictStore
     No. bytes          : 400000000 (381.5M)
     No. bytes stored   : 4684636 (4.5M)
     Storage ratio      : 85.4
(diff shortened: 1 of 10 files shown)


alimanfoo:
Ah, yes, sorry, forgot about that. I expect some test failures due to #324. May be other consequences of the upgrade too. Suggest dealing with them in this PR.

jakirkham:
SGTM. How about this one?

```python
        # create an object array using pickle
        z = self.create_array(shape=10, chunks=3, dtype=object, object_codec=Pickle())
        z[0] = 'foo'
>       assert z[0] == 'foo'
```

ref: https://travis-ci.org/zarr-developers/zarr/jobs/461883537#L2717-L2720

...or this one?

```python
            with pytest.raises(RuntimeError):
                # noinspection PyStatementEffect
>               v[:]
E               Failed: DID NOT RAISE <type 'exceptions.RuntimeError'>
```

ref: https://travis-ci.org/zarr-developers/zarr/jobs/461883537#L2621-L2624

jakirkham:
Also there is another failure. Appears `Pickle` wasn't able to handle general buffer protocol conforming types in `decode`. Though every other codec worked fine with these (which seems to be a consequence of them using the new utility functions 😉). Apparently `Pickle` was not using these. Have put together PR ( https://github.com/zarr-developers/numcodecs/pull/143 ) with a fix for this. Would be good to do another patch release to get that out (and anything else these failures might need).

jakirkham:
Based on testing locally, it looks like we are down to the `RuntimeError` not being raised by an `object` array after removing the filters.

```python
        for compressor in Zlib(1), Blosc():
            z = self.create_array(shape=len(data), chunks=30, dtype=object,
                                  object_codec=Categorize(greetings,
                                                          dtype=object),
                                  compressor=compressor)
            z[:] = data
            v = z.view(filters=[])
            with pytest.raises(RuntimeError):
                # noinspection PyStatementEffect
>               v[:]
E               Failed: DID NOT RAISE <type 'exceptions.RuntimeError'>
```

alimanfoo:
Hi @jakirkham, took the liberty to push a commit that should resolve the test failure due to RuntimeError not being raised for object arrays.

Other issues should be resolved with the fixes you've merged into numcodecs, so I'll cut a numcodecs 0.6.2 release then come back and update the version requirement here.

jakirkham:
Thanks @alimanfoo!

jakirkham:
Did a little bit of refactoring on your change. Hope that is ok. Please let me know your thoughts on that.

jakirkham:
On a different point, it looks like `__eq__` in `DictStore` is assuming all items in it are comparable. This assumption runs into problems however when anything that is not easily comparable is stored in the `DictStore`. For instance, `ndarray` is comparable, but does not reduce to a `bool` as expected, which causes [the tutorial's comparison]( https://github.com/zarr-developers/zarr/blob/master/docs/tutorial.rst#pickle-support ) to [fail CI]( https://travis-ci.org/zarr-developers/zarr/jobs/462274519#L2191-L2202 ). There are likely other ill-posed situations we can imagine.

Looking at this example before the recent Numcodecs release, it seems the contents of `DictStore` were always being filled with `bytes` likely as an indirect consequence of applying filters to the data first. Have added a commit, which enforces this behavior intentionally using `ensure_bytes` from Numcodecs. Thus making this behavior something we can rely upon for things like `__eq__` and such.

However there is a wrinkle with this approach. It appears we had a test that was assuming that `object`s could be placed in `DictStore`s. Though this behavior doesn't really work for any of the other stores and we usually require an `object_codec` to be specified to even serialize an `object` type (specifically with `Array`). Given this, I've dropped that test and am contending it doesn't conform to our expectations about how stores typically should work with `object` types.

Also it appears that the `Array`'s `default` store was a `dict`, which we cannot apply these fixes to. So have update the default to `DictStore`. I'm not sure why this wasn't already the case, which means I'm probably missing something.

Happy to hear thoughts or receive pushback on this approach as I may very well be missing something. If you know of a better approach, would also be interested in hearing about that as well.

jakirkham:
Seems the tutorial test is now hung up on some whitespace issue. Not totally sure how that got introduced.

jakirkham:
Simplifies some of the code by making use of Numcodecs utility functions. This avoids a few copies when working with some of the stores. Also generally makes the rest easier to understand.

jakirkham:
Replacing with PR ( https://github.com/zarr-developers/zarr/pull/352 ), which goes to Numcodecs 0.6.2. It also consolidates the changes from here and drops a few that have been pulled into other PRs. Namely ensuring `DictStore` contains only `bytes` for data ( https://github.com/zarr-developers/zarr/pull/350 ) and changing the default backend of `Array` to `DictStore` ( https://github.com/zarr-developers/zarr/pull/351 ).

PR #350, opened 2018-12-03
Ensure `DictStore` contains only `bytes`

jakirkham:
Partially addresses issue ( https://github.com/zarr-developers/zarr/issues/348 ) and issue ( https://github.com/zarr-developers/zarr/issues/349 ).

As the spec notes stores values must be an "arbitrary sequence of bytes", this change ensures that values in `DictStore` meet that constraint. Of course nesting of `DictStore`s are still allowed per usual. However these really just map to a variety of keys, which is fine.

Since the `DictStore`'s values are just `bytes`, there shouldn't be any cases where the size of these values cannot be determined. So drop handling for unknown sizes in `buffer_size`. Also drop the associated test for `DictStore` as this cannot occur.

Add a test case for `getsize` with a non-conforming `dict`-based store where sizes are unknown to make sure that case is tested and handled appropriately.

TODO:

### DIFF
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -544,6 +544,8 @@ class DictStore(MutableMapping):
     def __setitem__(self, item, value):
         with self.write_mutex:
             parent, key = self._require_parent(item)
+            if not isinstance(value, self.cls):
+                value = ensure_bytes(value)
             parent[key] = value
 
     def __delitem__(self, item):
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -544,8 +544,7 @@ class DictStore(MutableMapping):
     def __setitem__(self, item, value):
         with self.write_mutex:
             parent, key = self._require_parent(item)
-            if not isinstance(value, self.cls):
-                value = ensure_bytes(value)
+            value = ensure_bytes(value)
             parent[key] = value
 
     def __delitem__(self, item):
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -644,17 +644,11 @@ class DictStore(MutableMapping):
             size = 0
             for v in value.values():
                 if not isinstance(v, self.cls):
-                    try:
-                        size += buffer_size(v)
-                    except TypeError:
-                        return -1
+                    size += buffer_size(v)
             return size
 
         else:
-            try:
-                return buffer_size(value)
-            except TypeError:
-                return -1
+            return buffer_size(value)
 
     def clear(self):
         with self.write_mutex:
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -35,6 +35,9 @@ Bug fixes
 * Failing tests related to pickling/unpickling have been fixed. By :user:`Ryan Williams <ryan-williams>`,
   :issue:`273`, :issue:`308`.
 
+* Ensure ``DictStore`` contains only ``bytes`` to facilitate comparisons and protect against writes.
+  By :user:`John Kirkham <jakirkham>`, :issue:`350`
+
 Maintenance
 ~~~~~~~~~~~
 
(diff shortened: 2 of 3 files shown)


jakirkham:
Just for clarity, would be perfectly happy with any option that enforces that `bytes`-like data is in `DictStore`. Simply view this as a step in that direction. We can of course change to using `memoryview`s or anything else internally that makes sense and we are comfortable with.

alimanfoo:
> As the spec notes stores values must be an "arbitrary sequence of bytes", this change ensures that values in `DictStore` meet that constraint. 

Just to note, the spec really talks at an abstract level, so "arbitrary sequence of bytes" should be interpreted as an abstract concept. In a given programming language there may be many different types of object that encapsulate an arbitrary sequence of bytes, and the spec does not constrain which should be used. E.g., Python has the buffer protocol, which is a standard way of objects declaring that they encapsulate a sequence of bytes stored in memory. So in a Python implementation, it would be appropriate to require that stores can accept any value that exports the buffer protocol, but that does not have to be a bytes object.

That said, there are separate reasons for thinking that the ``DictStore`` should internally convert values to ``bytes`` objects, e.g., to ensure they are immutable. Sorry for being pedantic, but just wanted to clarify it is not the spec that requires that.

alimanfoo:
Thanks @jakirkham. FWIW I'm happy to go ahead with this, but if we do I think we should back out the changes in numcodecs that modified encode() methods to return ndarray when previously they returned bytes. Otherwise there will be a performance hit for in-memory zarr usage due to the extra memory copy.

jakirkham:
No worries. Completely agree.

By "meet that constraint" wasn't trying to suggest that the spec meant `bytes` specifically. Was more meaning that by using `ensure_bytes` we would reject anything that was not spec conforming from being stored. Tried to clarify this with the follow-up comment above. Sorry if that was still unclear.

Hope this makes more sense. Please let me know if I'm still missing something.

alimanfoo:
>
> By "meet that constraint" wasn't trying to suggest that the spec meant
> bytes specifically. Was more meaning that by using ensure_bytes we would
> reject anything that was not spec conforming from being stored. Tried to
> clarify this with the follow-up comment above. Sorry if that was still
> unclear.
>
Ah, perfect, I get it now, thanks!

jakirkham:
Please let me know if there is anything else needed for this?

jakirkham:
IMHO opinion renaming the store is a separate topic. Have raised issue ( https://github.com/zarr-developers/zarr/issues/356 ) to track it. Though I agree it is a good idea.

Adding some tests seems reasonable for this PR. Have included handling of the non-buffer case (disabled for `dict`). Also have added a test to ensure that `DictStore` coerces data stored to `bytes`.

zarr/storage.py
```diff
@@ -544,6 +544,8 @@ def __getitem__(self, item):
     def __setitem__(self, item, value):
         with self.write_mutex:
             parent, key = self._require_parent(item)
+            if not isinstance(value, self.cls):
+                value = ensure_bytes(value)
```
alimanfoo: I wonder, should this just be:

```python
    def __setitem__(self, item, value):
        with self.write_mutex:
            parent, key = self._require_parent(item)
            value = ensure_bytes(value)
            parent[key] = value
```

I.e., we only allow setting of bytes values, we don't allow any fiddling with the internal hierarchy of dicts?

alimanfoo:
> IMHO opinion renaming the store is a separate topic. Have raised issue ( #356 ) to track it. 

Great, thanks.

> Adding some tests seems reasonable for this PR. Have included handling of the non-buffer case (disabled for `dict`). Also have added a test to ensure that `DictStore` coerces data stored to `bytes`.

Looks good.

Just had one further comment re implementation of ``DictStore.__setitem__()``.

alimanfoo:
Thanks @jakirkham.

PR #351, opened 2018-12-03
WIP: Make `DictStore` the default `Array` store

jakirkham:
Based on discussion in issue ( https://github.com/zarr-developers/zarr/issues/349 ), it sounds like we are ok moving to `DictStore` to back `Array`s. This makes that change.

As the change depends somewhat on PR ( https://github.com/zarr-developers/zarr/pull/350 ), this has been marked as WIP until we resolve that one. Can revisit afterwards.

TODO:

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -176,7 +176,7 @@ print some diagnostics, e.g.::
     Read-only          : False
     Compressor         : Blosc(cname='zstd', clevel=3, shuffle=BITSHUFFLE,
                        : blocksize=0)
-    Store type         : builtins.dict
+    Store type         : zarr.storage.DictStore
     No. bytes          : 400000000 (381.5M)
     No. bytes stored   : 3379344 (3.2M)
     Storage ratio      : 118.4
@@ -268,7 +268,7 @@ Here is an example using a delta filter with the Blosc compressor::
     Read-only          : False
     Filter [0]         : Delta(dtype='<i4')
     Compressor         : Blosc(cname='zstd', clevel=1, shuffle=SHUFFLE, blocksize=0)
-    Store type         : builtins.dict
+    Store type         : zarr.storage.DictStore
     No. bytes          : 400000000 (381.5M)
     No. bytes stored   : 648605 (633.4K)
     Storage ratio      : 616.7
@@ -1198,7 +1198,7 @@ ratios, depending on the correlation structure within the data. E.g.::
     Order              : C
     Read-only          : False
     Compressor         : Blosc(cname='lz4', clevel=5, shuffle=SHUFFLE, blocksize=0)
-    Store type         : builtins.dict
+    Store type         : zarr.storage.DictStore
     No. bytes          : 400000000 (381.5M)
     No. bytes stored   : 6696010 (6.4M)
     Storage ratio      : 59.7
@@ -1212,7 +1212,7 @@ ratios, depending on the correlation structure within the data. E.g.::
     Order              : F
     Read-only          : False
     Compressor         : Blosc(cname='lz4', clevel=5, shuffle=SHUFFLE, blocksize=0)
-    Store type         : builtins.dict
+    Store type         : zarr.storage.DictStore
     No. bytes          : 400000000 (381.5M)
     No. bytes stored   : 4684636 (4.5M)
     Storage ratio      : 85.4
--- a/zarr/core.py
+++ b/zarr/core.py
@@ -108,6 +108,9 @@ class Array(object):
         # N.B., expect at this point store is fully initialized with all
         # configuration metadata fully specified and normalized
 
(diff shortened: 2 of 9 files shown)


jakirkham:
Something to consider here is what we do when a different in-memory `MutableMapping` is provided (e.g. `dict` or `OrderedDict`). One thought is that we construct a new `DictStore` and copy the data over. This should catch any non-conforming cases and ensure that everything conforms to the spec afterwards.

Another thing to consider is a copy was introduced in commit ( https://github.com/zarr-developers/zarr/commit/3c00d52356fb59968d05b8764b05a66c50457850 ) to ensure the underlying data of an in-memory store is not mutated. However if we always use a `DictStore` for in-memory data and that store always guarantees to contain immutable blobs of data ( e.g. `bytes` as in PR https://github.com/zarr-developers/zarr/pull/350 ), then we can drop this copy as it will be taken care of for us.

This latter thought re-raises [the point]( https://github.com/zarr-developers/zarr/issues/348#issuecomment-443659248 ) that we should allow `bytes` passed through by filters/compressors to remain untouched so as to avoid a copy later.

alimanfoo:
>
> Something to consider here is what we do when a different in-memory
> MutableMapping is provided (e.g. dict or OrderedDict). One thought is
> that we construct a new DictStore and copy the data over. This should
> catch any non-conforming cases and ensure that everything conforms to the
> spec afterwards.
>
Not sure what to do in that case, need to mull it over. Might be a bit odd
for the user if we ended up copying data across and using a different
object as the store, i.e., user would be expecting to see data in the
object they provided as store, but we ended up using something else.

The alternative would be to just try and use whatever object is provided as
a store. If it doesn't fully conform to the spec (e.g., plain dict), so be
it, the user may get unexpected behaviour.

> Another thing to consider is a copy was introduced in commit ( 3c00d52
> <https://github.com/zarr-developers/zarr/commit/3c00d52356fb59968d05b8764b05a66c50457850>
> ) to ensure the underlying data of an in-memory store is not mutated.
> However if we always use a DictStore for in-memory data and that store
> always guarantees to contain immutable blobs of data ( e.g. bytes as in
> PR #350 <https://github.com/zarr-developers/zarr/pull/350> ), then we can
> drop this copy as it will be taken care of for us.
>
Agree we could probably drop that copy if we know that DictStore will be
ensuring bytes.

> This latter thought re-raises the point
> <https://github.com/zarr-developers/zarr/issues/348#issuecomment-443659248>
> that we should allow bytes passed through by filters/compressors to
> remain untouched so as to avoid a copy later.
>
+1 to restore bytes output from compression codecs.

jakirkham:
Good point. Some random thoughts.

1. What if we had a generic wrapper class for `MutableMapping`s that exposed a `MutableMapping` API overriding a couple function to ensure they behaved safely (e.g. `__setitem__`)?
2. What if we merely enforce that `Array` provides `bytes` to the store?
3. Should we validate input stores meet the spec somehow?

alimanfoo:
My 2c...

> 1. What if we had a generic wrapper class for `MutableMapping`s that exposed a `MutableMapping` API overriding a couple function to ensure they behaved safely (e.g. `__setitem__`)?

I'm not sure it's necessary. E.g., we already normalise storage paths within the core and hierarchy modules before we ever interact with the storage layer, so we can be sure we'll never send an invalid key to a store.

> 2. What if we merely enforce that `Array` provides `bytes` to the store?

I'm not sure it's necessary to be so strict. Although I suppose that we could validate that whatever value comes out the end of the chunk encoding pipeline is an object that supports the buffer protocol, and raise some kind of encoding error if not, prior to sending the value to the storage layer.

> 3. Should we validate input stores meet the spec somehow?

Personally I think it would be good to provide some support for developers to test their storage class complies with the spec. But at runtime, let people provide whatever type of object they choose to provide as a store, and hope it quacks like the right kind of duck.

alimanfoo:
One other thought, I wondered if it might be worth renaming ``DictStore`` to ``MemoryStore`` so it was more obvious to new users what kind of storage this was (leaving ``DictStore`` as an alias for backwards-compatibility).

jakirkham:
There are still some test cases that use `dict`-based chunk storage for Zarr Arrays. So have added a workaround at the Zarr Array level to special case `dict` handling.

Though I'm not sure if we shouldn't just go for those two changes alone to fix this issue. Have pulled them out in PR ( https://github.com/zarr-developers/zarr/pull/359 ) just in case.

jakirkham:
Have pushed some changes that may be considered breaking. So have removed it from the milestone for now.

joshmoore:
@jakirkham @grlee77 : is it fair to say this has been superceded?

jakirkham:
Yeah I think so. We can always revisit if needed (likely we would need a new PR at this point)

Issue #348, opened 2018-12-02
Comparison of non-trivial, uncompressed, in-memory Zarr Arrays fails


jakirkham:
Comparing two in-memory Zarr arrays that do not use compressors errors out for anything other than the trivial case.

The issue being that NumPy arrays are being placed in the store (`dict` in this case). Even though NumPy arrays can be compared, the comparison is vectorized, which does not return a single `bool` that can be used in conditional expressions.

This usually works fine for other compressors as they often return `bytes`, which are placed in the store and can be compared to return a single `bool`. However this is not the case when no compressors are used or when some combination of filters and/or compressors are used that do return a NumPy array.

#### Minimal, reproducible code sample, a copy-pastable example if possible

```python
In [1]: import zarr                                                             

In [2]: z1 = zarr.zeros((100,), chunks=(10,), dtype='i4', compressor=None)      

In [3]: z2 = zarr.zeros((100,), chunks=(10,), dtype='i4', compressor=None)      

In [4]: z1 == z2                                                                
Out[4]: True

In [5]: z1[0] = 5                                                               

In [6]: z2[0] = 5                                                               

In [7]: z1 == z2                                                                
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-7-4176b4e1c607> in <module>
----> 1 z1 == z2

/zopt/conda3/envs/test_zarr/lib/python3.6/site-packages/zarr/core.py in __eq__(self, other)
    410         return (
    411             isinstance(other, Array) and
--> 412             self.store == other.store and
    413             self.read_only == other.read_only and
    414             self.path == other.path and

ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

#### Problem description

Ideally comparison here would return `True` or `False` instead of erroring out.

#### Version and installation information

Please provide the following:

* Value of ``zarr.__version__``: 2.2.0
* Value of ``numcodecs.__version__``: 0.5.5
* Version of Python interpreter: 3.6.7
* Operating system (Linux/Windows/Mac): macOS
* How Zarr was installed (e.g., "using pip into virtual environment", or "using conda"): Using Conda

Also, if you think it might be relevant, please provide the output from ``pip freeze`` or
``conda env export`` depending on which was used to install Zarr.

<details>

```yaml
name: test_zarr
channels:
  - conda-forge
  - defaults
dependencies:
  - appnope=0.1.0=py36_1000
  - asciitree=0.3.3=py_2
[...]

PR #357, opened 2018-12-06
WIP: Include some tests of store equality

jakirkham:
Provide a few tests that check to see if different stores compare equal or not.

Note: Currently this fails for a few stores. Hopefully this will be a good point to debug and fix those.

TODO:

### DIFF
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -781,10 +781,12 @@ class DirectoryStore(MutableMapping):
         return os.path.isfile(file_path)
 
     def __eq__(self, other):
-        return (
-            isinstance(other, DirectoryStore) and
-            self.path == other.path
-        )
+        if not isinstance(other, DirectoryStore):
+            return False
+        elif self.path == other.path:
+            return True
+        else:
+            return super(DirectoryStore, self).__eq__(other)
 
     def keys(self):
         if os.path.exists(self.path):
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -569,8 +569,8 @@ class DictStore(MutableMapping):
     def __eq__(self, other):
         return (
             isinstance(other, DictStore) and
-            self.root == other.root and
-            self.cls == other.cls
+            self.cls == other.cls and
+            self.root == other.root
         )
 
     def keys(self):
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -1485,13 +1485,15 @@ class DBMStore(MutableMapping):
             del self.db[key]
 
     def __eq__(self, other):
-        return (
-            isinstance(other, DBMStore) and
-            self.path == other.path and
+        if not isinstance(other, DBMStore):
+            return False
+        elif not (self.open == other.open and self.open_kwargs == other.open_kwargs):
             # allow flag and mode to differ
-            self.open == other.open and
-            self.open_kwargs == other.open_kwargs
-        )
+            return False
+        elif self.path == other.path:
+            return True
+        else:
+            return super(DBMStore, self).__eq__(other)
 
     def keys(self):
         return (_dbm_decode_key(k) for k in iter(self.db.keys()))
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -1968,6 +1968,14 @@ class SQLiteStore(MutableMapping):
         self.cursor.close()
         self.db.close()
 
+    def __eq__(self, other):
+        return (
+            isinstance(other, SQLiteStore) and
+            self.path == other.path and
+            self.kwargs == other.kwargs and
+            self.db == other.db
+        )
+
     def __getitem__(self, key):
         value = self.cursor.execute('SELECT v FROM zarr WHERE (k = ?)', (key,))
         for v, in value:
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -570,7 +570,7 @@ class DictStore(MutableMapping):
         return (
             isinstance(other, DictStore) and
             self.cls == other.cls and
-            self.root == other.root
+            self.root is other.root
         )
 
     def keys(self):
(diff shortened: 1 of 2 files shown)


alimanfoo:
Thanks @jakirkham. FWIW we might need to think a bit about what it should mean for two stores to compare equal. Currently, for example, two ``DirectoryStore`` objects will [compare equal](https://github.com/zarr-developers/zarr/blob/master/zarr/storage.py#L781) if they both use the same directory path on the file system. Consequently, two ``DirectoryStore`` objects that use different directory paths will always compare unequal, regardless of what data are stored there. Also, a ``DirectoryStore`` object and a store object of a different class (e.g., ``ZipStore``) will always compare unequal, regardless of what data are stored. This seemed like a simple and sensible thing to do at the time I wrote those storage classes, but I'll need to go back and try to remember why. Looking now to find what tests cover the current store ``__eq__()`` implementations...

alimanfoo:
So I'm pretty sure that the ``__eq__()`` implementations on the store classes are not currently tested directly, however they will be used inside the [``Array.__eq__()`` implementation](https://github.com/zarr-developers/zarr/blob/master/zarr/core.py#L416) and in the [``Group.__eq__()`` implementation](https://github.com/zarr-developers/zarr/blob/master/zarr/hierarchy.py#L184). Now trying to find where those are tested...

alimanfoo:
OK, ``Group`` equality is used in several tests in the test_hierarchy module, e.g., [test_require_group](https://github.com/zarr-developers/zarr/blob/master/zarr/tests/test_hierarchy.py#L168), [test_getitem_contains_iterator](https://github.com/zarr-developers/zarr/blob/master/zarr/tests/test_hierarchy.py#L406), and [test_paths](https://github.com/zarr-developers/zarr/blob/master/zarr/tests/test_hierarchy.py#L791). 

Struggling to find where ``Array.__eq__()`` is tested...

alimanfoo:
OK, got it. We currently rely on ``Array.__eq__()`` in several tests within the test_hierarchy module, e.g., [here](https://github.com/zarr-developers/zarr/blob/master/zarr/tests/test_hierarchy.py#L648), [here](https://github.com/zarr-developers/zarr/blob/master/zarr/tests/test_hierarchy.py#L423), [here](https://github.com/zarr-developers/zarr/blob/master/zarr/tests/test_hierarchy.py#L308). We don't currently test it within the test_core module.

alimanfoo:
Reviewing the current tests, I think the current implementations of ``__eq__()`` in the ``Array``, ``Group`` and store classes do serve the tests well. 

Is there a reason for considering a different behaviour for the store ``__eq__()`` implementations?

jakirkham:
Based on our discussions in issue ( https://github.com/zarr-developers/zarr/issues/349 ) and the problem seen in issue ( https://github.com/zarr-developers/zarr/issues/348 ), it does seem like we need to start testing whether stores compare equal and if this is as expected to catch and remove bugs.

alimanfoo:
> Based on our discussions in issue ( #349 ) and the problem seen in issue ( #348 ), it does seem like we need to start testing whether stores compare equal and if this is as expected to catch and remove bugs.

FWIW if we switch to using DictStore as the default store, then #348 should go away. I.e., we will get consistent behaviour for array equals comparison across all stores, which is essentially that two arrays will compare equal iff they share the same storage.

jakirkham:
So I've made some changes inline with the intent. There will still be a failure for the `dict` case that we can discuss.

jakirkham:
> So I'm pretty sure that the `__eq__()` implementations on the store classes are not currently tested directly, however they will be used inside the [`Array.__eq__()` implementation](https://github.com/zarr-developers/zarr/blob/master/zarr/core.py#L416) and in the [`Group.__eq__()` implementation](https://github.com/zarr-developers/zarr/blob/master/zarr/hierarchy.py#L184).

I'm not quite following this. Wouldn't we want to ensure that two Arrays have the same content regardless of where they are stored? Would also be curious why wouldn't we want to compare Groups' contents as well. Thoughts?

alimanfoo:
On Wed, 13 Feb 2019 at 08:47, jakirkham <notifications@github.com> wrote:

> So I'm pretty sure that the __eq__() implementations on the store classes
> are not currently tested directly, however they will be used inside the
> Array.__eq__() implementation
> <https://github.com/zarr-developers/zarr/blob/master/zarr/core.py#L416>
> and in the Group.__eq__() implementation
> <https://github.com/zarr-developers/zarr/blob/master/zarr/hierarchy.py#L184>
> .
>
> I'm not quite following this. Wouldn't we want to ensure that two Arrays
> have the same content regardless of where they are stored?
>
The tests I linked to currently use __eq__ to verify that two arrays draw
their data from the same store (or two groups draw their data from the same
store). They are not trying to compare the actual data. Part of that
behaviour involves testing that two store objects are "equal" in the sense
that they both draw their data from the same source (e.g., the same
directory on a file system). Again that doesn't involve comparing any
actual data. Those tests all need some way of making this kind of
comparison, although perhaps __eq__ is not the right method for it.

I could imagine in other situations you might want to compare the actual
data contents of two arrays (or two groups or two stores) and return true
if the data are the same, regardless if drawn from the same data source or
not. That's a separate requirement. Maybe __eq__ should behave like that.
In this case, should zarr arrays behave like numpy arrays, and return an
array of boolean values with broadcasting etc.?

Bottom line, I think there are two separate requirements for arrays: (1)
compare two arrays and return True if they are "the same" in the sense that
they both draw their data from the same source; and (2) compare the data
contents of two arrays and return something.

Similarly, I think there are two separate requirements for groups: (1)
return True if two groups draw their data from the same source; (2) compare
the contents of two groups.

Similarly for stores: (1) return True if two store objects draw their data
from the same underlying source; (2) compare the contents of two stores.

Now the question is, what should __eq__ do? And what other methods do we
need to cover any remaining requirements not covered by __eq__?

PR #359, opened 2018-12-07
Special case chunk encoding for `dict` chunk store

jakirkham:
This removes the special case copying of chunks in commit ( https://github.com/zarr-developers/zarr/commit/3c00d52356fb59968d05b8764b05a66c50457850 ). Instead this checks to see if a `dict` instance is being used to back chunk storage. If it is, this coerces the chunk data to `bytes` (if it is not already `bytes`). This avoids a copy for well-defined backends that already protect their data against unwanted tampering. Also continues to avoid copies when compressors and/or filters already return the data as `bytes`. However this will copy in any case where a non-`bytes` object (e.g. an `ndarray`) is placed in the chunk store (e.g. no filters/compressors are used or some specific filter is used at the end). This behavior for `dict`-based chunk stores is analogous to what `DictStore` does currently for stored values.

Note: This also extends to subclasses of `dict`, which include `defaultdict` and `OrderedDict`. It does not include `ChainMap` or `UserDict`; however, it's unclear whether we should be special casing those two as `ChainMap` is meant to provide a unified interface onto many `MutableMapping`s of various kinds and `UserDict` is intended as a clean and simple alternative for subclassing `dict`.

TODO:

### DIFF
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -56,6 +56,11 @@ Bug fixes
 * Always use a ``tuple`` when indexing a NumPy ``ndarray``.
   By :user:`John Kirkham <jakirkham>`, :issue:`376`
 
+* Ensure when ``Array`` uses a ``dict``-based chunk store that it only contains
+  ``bytes`` to facilitate comparisons and protect against writes. Drop the copy
+  for the no filter/compressor case as this handles that case.
+  By :user:`John Kirkham <jakirkham>`, :issue:`359`
+
 Maintenance
 ~~~~~~~~~~~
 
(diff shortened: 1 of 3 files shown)


jakirkham:
This is ready for review. Would be curious to hear thoughts on this. Basically this is meant to make working with `dict` objects as an `Array` chunk store a bit better behaved.

jakirkham:
Would be great to hear your thoughts on this when you have a chance, @alimanfoo.

jakirkham:
Adding to the [v2.3 milestone]( https://github.com/zarr-developers/zarr/milestone/7 ) so we can keep track of it. Happy to change that as needed.

jakirkham:
Thoughts @alimanfoo?

jakirkham:
Thoughts @zarr-developers/core-devs?

PR #363, opened 2018-12-15
Encode/Decode complex fill values

jakirkham:
Fixes https://github.com/zarr-developers/zarr/issues/244

Provides a simple implementation of encoding/decoding complex fill values and tests that it works in a basic case.

TODO:

### DIFF
--- a/zarr/meta.py
+++ b/zarr/meta.py
@@ -163,6 +163,11 @@ def decode_fill_value(v, dtype):
             return np.NINF
         else:
             return np.array(v, dtype=dtype)[()]
+    elif dtype.kind in 'c':
+        v = (decode_fill_value(v[0], dtype.type().real.dtype),
+             decode_fill_value(v[1], dtype.type().imag.dtype))
+        v = v[0] + 1j * v[1]
+        return np.array(v, dtype=dtype)[()]
     elif dtype.kind == 'S':
         # noinspection PyBroadException
         try:
@@ -201,6 +206,10 @@ def encode_fill_value(v, dtype):
         return int(v)
     elif dtype.kind == 'b':
         return bool(v)
+    elif dtype.kind in 'c':
+        v = (encode_fill_value(v.real, dtype.type().real.dtype),
+             encode_fill_value(v.imag, dtype.type().imag.dtype))
+        return v
     elif dtype.kind in 'SV':
         v = base64.standard_b64encode(v)
         if not PY2:  # pragma: py2 no cover
--- a/zarr/tests/test_core.py
+++ b/zarr/tests/test_core.py
@@ -949,6 +949,15 @@ class TestArray(unittest.TestCase):
             z[:] = a
             assert_array_almost_equal(a, z[:])
 
+        # complex
+        for dtype in 'c8', 'c16':
+            z = self.create_array(shape=10, chunks=3, dtype=dtype)
+            assert z.dtype == np.dtype(dtype)
+            a = np.linspace(0, 1, z.shape[0], dtype=dtype)
+            a -= 1j * a
+            z[:] = a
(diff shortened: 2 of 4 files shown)


jakirkham:
Adding to the [v2.3 milestone]( https://github.com/zarr-developers/zarr/milestone/7 ) to make it easier to keep track of this. However if we opt for a different approach or decide to pass on complex numbers for now, we can always remove it from the milestone.

zarr/meta.py
```diff
@@ -201,6 +206,10 @@ def encode_fill_value(v, dtype):
         return int(v)
     elif dtype.kind == 'b':
         return bool(v)
+    elif dtype.kind in 'c':
+        v = (encode_fill_value(v.real, dtype.type().real.dtype),
+             encode_fill_value(v.imag, dtype.type().imag.dtype))
+        return v
```
rabernat: I take it this encoding / decoding of fill values is necessary is necessary because json does not support serialization of complex data types?

zarr/tests/test_core.py
```diff
@@ -949,6 +949,15 @@ def test_dtypes(self):
...
             assert_array_almost_equal(a, z[:])
 
+        # complex
+        for dtype in 'c8', 'c16':
+            z = self.create_array(shape=10, chunks=3, dtype=dtype)
+            assert z.dtype == np.dtype(dtype)
+            a = np.linspace(0, 1, z.shape[0], dtype=dtype)
+            a -= 1j * a
+            z[:] = a
+            assert_array_almost_equal(a, z[:])
```
rabernat: Given the special treatment required for fill values, would it make sense to add an explicit check that complex data with missing values is handled correctly? I see that the metadata encoding is checked below, but not the actual filling procedure (as far as I can tell).

zarr/meta.py
jakirkham: That's right.

This will encode to a JSON list, which is nice for readability. Though we could certainly imagine other ways of handling this (e.g. base64 encoded string).

Any preferences?

zarr/tests/test_core.py
jakirkham: Could you please elaborate a bit on what kind of representation of missing data we are discussing? We handle `NaN` currently, but maybe this is not what you are thinking about.

jakirkham:
Thanks for the feedback @rabernat. Responded inline above. Happy to work with you to address these concerns.

zarr/meta.py
rabernat: That seems totally reasonable.

zarr/tests/test_core.py
rabernat: What I mean is to make sure that missing data (the kind that triggers the `fill_value` machinery to be activated) is present in the array. For example, by adding
```python
a[0] = np.nan
```

However, I might be understanding what `fill_value` does in zarr. Is `fill_value` only used to represent uninitialized portions of the array? Or are all nans explicitly translated to `fill_value` before serialization?

jakirkham: AFAIK `fill_value` is only used to handle uninitialized chunks. Am not aware of any special behavior around `NaN`s in Zarr. Does that help?

jakirkham:
Any other thoughts on this? Currently thinking it would be good to merge this by end of week.

jakirkham:
Please let me know if anything else occurs to you, @rabernat. Happy to address in a follow-up.

jakirkham:
As this followed PR ( https://github.com/zarr-developers/zarr/pull/309 ), it effectively started testing `N5Store` with complex data, which doesn't work as evidenced by [the CI failure]( https://travis-ci.org/zarr-developers/zarr/builds/498601006 ). Have added PR ( https://github.com/zarr-developers/zarr/pull/411 ) to correct this.

PR #367, opened 2018-12-16
WIP, RFC: Optimize setitem with chunk equal to fill_value

jakirkham:
Fixes https://github.com/zarr-developers/zarr/issues/366

If a chunk only contains the `fill_value`, delete the key-value pair for that chunk instead of storing the array. Should cutdown on storage space require for the `Array`. Also should improve the performance of copying one `Array` to another.

TODO:


rabernat:
I have a very large, relatively sparse (lots of empty chunks) array. I would really like to avoid writing all those empty chunks to disk.

So :+1: to this!

zarr/core.py
```diff
@@ -1476,6 +1476,17 @@ def _set_basic_selection_zd(self, selection, value, fields=None):
         else:
             chunk[selection] = value
 
+        # clear chunk if it only contains the fill value
+        if np.all(np.equal(chunk, self._fill_value)):
```
rabernat: This seems like it could sometimes be an expensive operation. So it should be optional.

jakirkham:
Thanks! Yeah that was the idea behind it.

There was funkiness with `ZipStore` as deletion is not supported there.

Can give this another look 🙂

joshmoore:
I actually assumed recently that this was the default.

zarr/core.py
jakirkham: How large are typical chunk sizes for you?

jakirkham:
Nope, but I think we discussed recently how we might get this to work for `ZipStore` and other non-deletable stores by just making the value `b''`. So I think we may be able to generalize this a bit

d-v-b:
I was also bitten by thinking that was the default! Would love to see this merged.

zarr/core.py
d-v-b: regardless of chunk size, in many cases the user knows in advance whether this check will ever be useful. For dense imaging data, this conditional will almost never evaluate to true; for sparse data, this conditional will often evaluate to true. So the user should have some discretion over whether this chunk occurs.

d-v-b:
@jakirkham I would be happy to move this PR forward, but I don't think I have push access to your fork. I made some changes (added a kwarg to the `Array` constructor, and added tests) here: https://github.com/d-v-b/zarr-python/tree/opt_setitem_fill_value

PR #372, opened 2018-12-27
MongoDB and Redis stores

### LINKED ISSUE #299: Add MongoDB storage backend?
I have been using MongoDB lately for logging the results of my computational experiments, and I have found that it is a pretty fun piece of software to use. Since MongoDB is designed to store large amounts of data across a distributed network, It seems like it could be a good match for zarr. I went ahead and made a very simple MutableMapping wrapper to MongoDB: https://gist.github.com/nbren12/9842cf22f173a864a7c8377c01ad06c5, and I am wondering if this is something that might be worth cleaning up and adding to `zarr`. 

It would be interesting to benchmark it's performance compared to S3 and GCS backends. FWIW, a package which seems to use mongoDB in a similar way to store large binary datasets is https://github.com/manahl/arctic.

jhamman:
This adds two new stores to zarr, MongoDB and Redis. This is basically the same implementation that @nbren12 proposed in [this gist](https://gist.github.com/nbren12/9842cf22f173a864a7c8377c01ad06c5). The PR is far from complete but I wanted to put it up for others to see, should there be interest.

This is also building on #368 so that should be merged first. 

Finally, I should say that I've been meaning to learn a bit about how zarr stores are implemented so I'm using this as a learning experience. 

closes #299 

TODO:

### DIFF
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -29,7 +29,7 @@ Bug fixes
 
 * The required version of the `numcodecs <http://numcodecs.rtfd.io>`_ package has been upgraded 
   to 0.6.2, which has enabled some code simplification and fixes a failing test involving
-  msgpack encoding. By :user:`John Kirkham <jakirkham>`, :issue:`352`, :issue:`355`, 
+  msgpack encoding. By :user:`John Kirkham <jakirkham>`, :issue:`360`, :issue:`352`, :issue:`355`,
   :issue:`324`.
 
 * Failing tests related to pickling/unpickling have been fixed. By :user:`Ryan Williams <ryan-williams>`,
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -21,7 +21,7 @@ Enhancements
 
 * Adds the SQLite-backed :class:`zarr.storage.SQLiteStore` class enabling an
   SQLite database to be used as the backing store for an array or group.
-  :issue:`368`.
+   By :user:`John Kirkham <jakirkham>`, :issue:`368`, :issue:`365`.
 
 Bug fixes
 ~~~~~~~~~
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -21,7 +21,12 @@ Enhancements
 
 * Adds the SQLite-backed :class:`zarr.storage.SQLiteStore` class enabling an
   SQLite database to be used as the backing store for an array or group.
-   By :user:`John Kirkham <jakirkham>`, :issue:`368`, :issue:`365`.
+  By :user:`John Kirkham <jakirkham>`, :issue:`368`, :issue:`365`.
+
+* Adds the :class:`zarr.storage.RedisStore` and :class:`zarr.storage.MongoDBStore` classes
+  enabling an Redis and MongoDB databases to be used as the backing store for an array
+  or group.
+  By :user:`Joe Hamman <jhamman>`, :issue:`299`, :issue:`372`.
 
 Bug fixes
 ~~~~~~~~~
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -23,6 +23,9 @@ Enhancements
   SQLite database to be used as the backing store for an array or group.
   By :user:`John Kirkham <jakirkham>`, :issue:`368`, :issue:`365`.
 
+* Efficient iteration over arrays by decompressing chunkwise.
+  By :user:`Jerome Kelleher <jeromekelleher>`, :issue:`398`, :issue:`399`.
+
 * Adds the Redis-backed :class:`zarr.storage.RedisStore` class enabling an
   Redis database to be used as the backing store for an array or group.
   By :user:`Joe Hamman <jhamman>`, :issue:`299`, :issue:`372`.
@@ -34,12 +37,12 @@ Enhancements
 Bug fixes
 ~~~~~~~~~
 
-* The implementation of the :class:`zarr.storage.DirectoryStore` class has been modified to 
-  ensure that writes are atomic and there are no race conditions where a chunk might appear 
-  transiently missing during a write operation. By :user:`sbalmer <sbalmer>`, :issue:`327`, 
+* The implementation of the :class:`zarr.storage.DirectoryStore` class has been modified to
+  ensure that writes are atomic and there are no race conditions where a chunk might appear
+  transiently missing during a write operation. By :user:`sbalmer <sbalmer>`, :issue:`327`,
   :issue:`263`.
 
-* The required version of the `numcodecs <http://numcodecs.rtfd.io>`_ package has been upgraded 
+* The required version of the `numcodecs <http://numcodecs.rtfd.io>`_ package has been upgraded
   to 0.6.2, which has enabled some code simplification and fixes a failing test involving
   msgpack encoding. By :user:`John Kirkham <jakirkham>`, :issue:`361`, :issue:`360`, :issue:`352`,
   :issue:`355`, :issue:`324`.
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -23,9 +23,12 @@ Enhancements
   SQLite database to be used as the backing store for an array or group.
   By :user:`John Kirkham <jakirkham>`, :issue:`368`, :issue:`365`.
 
-* Adds the :class:`zarr.storage.RedisStore` and :class:`zarr.storage.MongoDBStore` classes
-  enabling an Redis and MongoDB databases to be used as the backing store for an array
-  or group.
+* Adds the Redis-backed :class:`zarr.storage.RedisStore` class enabling an
+  Redis database to be used as the backing store for an array or group.
+  By :user:`Joe Hamman <jhamman>`, :issue:`299`, :issue:`372`.
+
+* Adds the MongoDB-backed :class:`zarr.storage.MongoDBStore` class enabling an
+  MongoDB database to be used as the backing store for an array or group.
   By :user:`Joe Hamman <jhamman>`, :issue:`299`, :issue:`372`.
 
 Bug fixes
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -26,11 +26,11 @@ Enhancements
 * Efficient iteration over arrays by decompressing chunkwise.
   By :user:`Jerome Kelleher <jeromekelleher>`, :issue:`398`, :issue:`399`.
 
-* Adds the Redis-backed :class:`zarr.storage.RedisStore` class enabling an
+* Adds the Redis-backed :class:`zarr.storage.RedisStore` class enabling a
   Redis database to be used as the backing store for an array or group.
   By :user:`Joe Hamman <jhamman>`, :issue:`299`, :issue:`372`.
 
-* Adds the MongoDB-backed :class:`zarr.storage.MongoDBStore` class enabling an
+* Adds the MongoDB-backed :class:`zarr.storage.MongoDBStore` class enabling a
   MongoDB database to be used as the backing store for an array or group.
   By :user:`Joe Hamman <jhamman>`, :issue:`299`, :issue:`372`.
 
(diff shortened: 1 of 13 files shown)


nbren12:
Thanks for running with this @jhamman. Let me know if you want any help from my end.

jhamman:
@nbren12 - I think I have this to a place that would be useful for you to give a review. Current test failures include:

- Coveralls: Missing test for `KeyError` in `RedisStore.__getitem__` ([link](https://coveralls.io/builds/20847231/source?filename=zarr%2Fstorage.py#L2205))
- Travis Python 2: `MongoDBStore` json decoder error ([link](https://travis-ci.org/zarr-developers/zarr/jobs/473753393#L2148))

zarr/storage.py
```diff
@@ -1877,6 +1879,368 @@ def __delitem__(self, key):
...
+        self._prefix = prefix
+        self._kwargs = kwargs
+
+        self.client = redis.Redis(**kwargs)
+
+    def _key(self, key):
+        return '{prefix}:{key}'.format(prefix=self._prefix, key=key)
+
+    def __getitem__(self, key):
+        value = self.client[self._key(key)]
```
alimanfoo: I think the client will raise KeyError here for you if the key doesn't exist, so could simplify implementation to just:

```python
return self.client[self._key(key)]
```

alimanfoo:
Not completely sure why the PY2 test failures, looks like the values for the metadata keys are getting mangled, possibly due to use of `ensure_bytes()` within `__setitem__`. This is a guess, but you could try something like this:

```python
from zarr.compat import text_type

def __setitem__(key, value):
    if isinstance(value, text_type):
        value = value.encode('ascii')
    else:
        value = ensure_bytes(value)
    # ... carry on as before
```

zarr/storage.py
```diff
@@ -1877,6 +1879,368 @@ def __delitem__(self, key):
...
+    def __getitem__(self, key):
+        doc = self.collection.find_one({self._key: key})
+
+        if doc is None:
+            raise KeyError(key)
+        else:
+            return doc[self._value]
+
+    def __setitem__(self, key, value):
+        value = ensure_bytes(value)
```
nbren12: Should we impose some kind of size limit here since MongoDB documents must be smaller than 16MB? Or just leave it MongoClient to raise its own errors?

jhamman: If MongoClient will raise an error, I'd prefer to let them do it. We can add some documentation for this point but I want to be careful to not artificially introduce a limit that MongoDB could raise in future versions.

alimanfoo:
Sorry, looks like my random guess about how to fix the PY2 failures was wrong.

Looking at this test failure output:

```
_______________________ TestMongoDBStore.test_init_array _______________________
self = <zarr.tests.test_storage.TestMongoDBStore testMethod=test_init_array>
    def test_init_array(self):
        store = self.create_store()
        init_array(store, shape=1000, chunks=100)
    
        # check metadata
        assert array_meta_key in store
>       meta = decode_array_metadata(store[array_meta_key])
zarr/tests/test_storage.py:322: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
zarr/meta.py:52: in decode_array_metadata
    meta = parse_metadata(s)
zarr/meta.py:46: in parse_metadata
    meta = json.loads(s)
/opt/python/2.7.14/lib/python2.7/json/__init__.py:339: in loads
    return _default_decoder.decode(s)
/opt/python/2.7.14/lib/python2.7/json/decoder.py:364: in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
self = <json.decoder.JSONDecoder object at 0x7fd319083b10>
s = '{\x00\x00\x00\n\x00\x00\x00 \x00\x00\x00 \x00\x00\x00 \x00\x00\x00 \x00\x00\x00"\x00\x00\x00c\x00\x00\x00h\x00\x00\x0...m\x00\x00\x00a\x00\x00\x00t\x00\x00\x00"\x00\x00\x00:\x00\x00\x00 \x00\x00\x002\x00\x00\x00\n\x00\x00\x00}\x00\x00\x00'
```

...I still think it is something to do with how the JSON metadata is getting mangled when it is put into the store. This looks like JSON but where 3 null bytes have been inserted after every character. E.g., the last part is probably the end of the `"zarr_format": 2`.

zarr/storage.py
```diff
@@ -1877,6 +1879,365 @@ def __delitem__(self, key):
...
+        self.db = self.client.get_database(self._database)
+        self.collection = self.db.get_collection(self._collection)
+
+    def __getitem__(self, key):
+        doc = self.collection.find_one({self._key: key})
+
+        if doc is None:
+            raise KeyError(key)
+        else:
+            return binary_type(doc[self._value])
```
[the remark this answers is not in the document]
jakirkham: Interesting. Wonder if `ensure_bytes` would work here as well. 🤔

jhamman:
@jakirkham and @alimanfoo - I think this is basically done. I don't know why the coveralls report is failing. My reading of the coveralls report show that we have 100% coverage on the new bits in this PR but please let me know if I'm missing something.

alimanfoo:
@jhamman I think coveralls is sometimes broken, I have seen that before where coverage is 100% on the new/modified files but somehow coveralls reports total coverage down. One way to check is to make a trivial commit and see if the problem persists. I'll also do some digging.

zarr/storage.py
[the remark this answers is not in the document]
alimanfoo: Out of interest, what type of object is `doc[self._value]`?

zarr/storage.py
```diff
@@ -2084,6 +2084,184 @@ def clear(self):
...
+    def __getitem__(self, key):
+        return self.client[self._key(key)]
+
+    def __setitem__(self, key, value):
+        value = ensure_bytes(value)
+        self.client[self._key(key)] = value
+
+    def __delitem__(self, key):
+        if key not in self.keylist():
+            raise KeyError(key)
```
alimanfoo: Just wondering if there's a way to do this without checking the keylist first. I.e., is it possible just to call `del self.client[self._key(key)]` and either propagate or transform the error raised by redis if no key exists?

zarr/tests/test_storage.py
```diff
@@ -930,6 +940,31 @@ def test_pickle(self):
             pickle.dumps(store)
 
 
+@unittest.skipIf(pymongo is None, 'test requires pymongo')
+class TestMongoDBStore(StoreTests, unittest.TestCase):
+
+    def create_store(self):
+        # TODO: this is the default host for MongoDB on Travis,
+        # we probably want to generalize this though
```
alimanfoo: Is this TODO something you want to address before merge? Would we ever need to do something different?

zarr/tests/test_storage.py
```diff
@@ -930,6 +940,31 @@ def test_pickle(self):
...
+        # start with an empty store
+        store.clear()
+        return store
+
+
+@unittest.skipIf(redis is None, 'test requires redis')
+class TestRedisStore(StoreTests, unittest.TestCase):
+
+    def create_store(self):
+        # TODO: this is the default host for Redis on Travis,
```
alimanfoo: As above, is this TODO something that needs to be resolved before merge, or can we live with as-is?

requirements_dev_optional.txt
```diff
@@ -1,3 +1,5 @@
 # These packages are currently not available on Windows.
 bsddb3==6.2.6
 lmdb==0.94
+redis
+pymongo
```
alimanfoo: Can we pin these?

alimanfoo:
I think the coveralls report is broken, all files have 100% coverage.

zarr/tests/test_storage.py
jhamman: Right. I think the more important question is how do you want to handle these tests when a mongodb or redis server is not available. How should we decide when to skip these tests?

zarr/storage.py
jhamman: As far as I can tell, we need this if we want a KeyError to be raised when the key does not exist. I tried commenting this out and the __delitem__ tests fail when we don't explicitly raise this error.

jhamman: Actually, it looks like we can use the `delete` and check for the number of deleted keys. I've done this in e2988be so we should be good here.

docs/release.rst
```diff
@@ -26,6 +26,14 @@ Enhancements
 * Efficient iteration over arrays by decompressing chunkwise.
   By :user:`Jerome Kelleher <jeromekelleher>`, :issue:`398`, :issue:`399`.
 
+* Adds the Redis-backed :class:`zarr.storage.RedisStore` class enabling an
```
alimanfoo:
```suggestion
* Adds the Redis-backed :class:`zarr.storage.RedisStore` class enabling a
```

docs/release.rst
```diff
@@ -26,6 +26,14 @@ Enhancements
...
+  Redis database to be used as the backing store for an array or group.
+  By :user:`Joe Hamman <jhamman>`, :issue:`299`, :issue:`372`.
+
+* Adds the MongoDB-backed :class:`zarr.storage.MongoDBStore` class enabling an
```
alimanfoo:
```suggestion
* Adds the MongoDB-backed :class:`zarr.storage.MongoDBStore` class enabling a
```

zarr/tests/test_storage.py
```diff
@@ -900,6 +900,16 @@ def test_context_manager(self):
 except ImportError:  # pragma: no cover
     sqlite3 = None
 
+try:
+    import pymongo
+except ImportError:  # pragma: no cover
+    pymongo = None
```
alimanfoo: What about adding an additional check here for the presence of a local server against which to run tests? E.g.:

```python
try:
    import pymongo
    # check for presence of local server to run tests against
    pymongo.MongoClient(host='127.0.0.1', database='zarr_tests', collection='zarr_tests')
except ImportError:  # pragma: no cover
    pymongo = None
except WhateverExceptionMongoRaisesOnFailedConnection:
    pymongo = None
```

This would ensure tests are skipped if server is not available.

zarr/tests/test_storage.py
```diff
@@ -900,6 +900,16 @@ def test_context_manager(self):
...
+
+try:
+    import redis
+except ImportError:  # pragma: no cover
+    redis = None
```
alimanfoo: Similar to comment above for mongo, what about adding a check here for a local server, as a mechanism for skipping redis tests if server not available?

zarr/tests/test_storage.py
jhamman: done.

zarr/tests/test_storage.py
jhamman: done.

alimanfoo:
Travis fails because of flake8:

```
py37 runtests: commands[4] | flake8 zarr
zarr/tests/test_storage.py:908:37: E128 continuation line under-indented for visual indent
```

(Aside: I'm going to enable the pep8speaks bot so we don't have to go digging through the travis logs.)

alimanfoo:
I've fixed the pep8 problem, I think this is good to go, will merge if CI passes.

alimanfoo:
OK, this is going in, thanks all.

zarr/storage.py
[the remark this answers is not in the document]
jakirkham: AFAICT it's a `bytes` object. Am curious to see if we may have already fixed this with other changes and this can be dropped. Have added PR ( https://github.com/zarr-developers/zarr/pull/401 ) to give this a try.

PR #376, opened 2019-01-02
Always use `tuple`s for multidimensional indexing

jakirkham:
NumPy 1.15.0 deprecates the use of sequences in `ndarray` selection for sequences other than `tuple`s. We have a few cases where we are using `list`s instead as they are mutable. Thus are easier to build up and change. However passing `list`s into `ndarray` selection will cause this `FutureWarning` to be emitted (and will eventually result in an error).

To fix this simply, just convert these `list`s to `tuple`s after we are done changing their contents, but before we use them with `ndarray`s. This works just as well on older versions of NumPy. Avoids the `FutureWarning` in NumPy 1.15.0+. Not to mention this will continue to work on future versions of NumPy, which may change this warning into an error.

xref: https://github.com/numpy/numpy/pull/9686

TODO:

### DIFF
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -38,6 +38,9 @@ Bug fixes
 * Ensure ``DictStore`` contains only ``bytes`` to facilitate comparisons and protect against writes.
   By :user:`John Kirkham <jakirkham>`, :issue:`350`
 
+* Always use a ``tuple`` when indexing a NumPy ``ndarray``.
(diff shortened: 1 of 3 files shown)


jakirkham:
Thoughts @zarr-developers/core-devs?

shoyer:
Yes, I think this is definitely the right way to do things.

jakirkham:
Thanks all :)

Went ahead and merged since it was a small change and we seemed to be in agreement.

PR #377, opened 2019-01-04
RFC: Optionally support memory-mapping DirectoryStore values

jakirkham:
Fixes https://github.com/zarr-developers/zarr/issues/265
Depends on ~~https://github.com/zarr-developers/zarr/pull/378~~ & ~~https://github.com/zarr-developers/zarr/pull/379~~ & ~~https://github.com/zarr-developers/zarr/pull/380~~

Adds the `memmap` option to `DirectoryStore`, which defaults to `False` (current behavior). If set to `True`, this will memory-map each value returned instead of loading it all into memory. This behavior is extended to `NestedDirectoryStore` and `TempStore`.

Given the recent work in PR ( https://github.com/zarr-developers/numcodecs/pull/128 ) and PR ( https://github.com/zarr-developers/zarr/pull/352 ), this just works out-of-the-box. The tests needed some adjustment to be a bit more flexible as they were strictly requiring `bytes` before. Also had to implement `pop` and `popitem` to ensure data was copied into memory before files were removed.

xref: https://github.com/zarr-developers/zarr/issues/320
xref: https://github.com/zarr-developers/zarr/issues/321
xref: https://github.com/zarr-developers/zarr/issues/334

TODO:

### DIFF
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -1035,6 +1035,10 @@ class NestedDirectoryStore(DirectoryStore):
             self.path == other.path
         )
 
+    def pop(self, key, *args, **kwargs):
+        key = _nested_map_ckey(key)
+        return super(NestedDirectoryStore, self).pop(key, *args, **kwargs)
+
     def listdir(self, path=None):
         children = super(NestedDirectoryStore, self).listdir(path=path)
         if array_meta_key in children:
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -901,10 +901,10 @@ class TempStore(DirectoryStore):
     """
 
     # noinspection PyShadowingBuiltins
-    def __init__(self, suffix='', prefix='zarr', dir=None):
+    def __init__(self, suffix='', prefix='zarr', dir=None, memmap=False):
         path = tempfile.mkdtemp(suffix=suffix, prefix=prefix, dir=dir)
         atexit.register(atexit_rmtree, path)
-        super(TempStore, self).__init__(path)
+        super(TempStore, self).__init__(path, memmap=memmap)
 
 
 _prog_ckey = re.compile(r'^(\d+)(\.\d+)+$')
@@ -987,8 +987,8 @@ class NestedDirectoryStore(DirectoryStore):
 
     """
 
-    def __init__(self, path):
-        super(NestedDirectoryStore, self).__init__(path)
+    def __init__(self, path, memmap=False):
+        super(NestedDirectoryStore, self).__init__(path, memmap=memmap)
 
     def __getitem__(self, key):
         key = _nested_map_ckey(key)
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -810,6 +810,30 @@ class DirectoryStore(MutableMapping):
     def __len__(self):
         return sum(1 for _ in self.keys())
 
+    __marker = object()
+
+    def pop(self, key, default=__marker):
+        try:
+            value = ensure_bytes(self[key])
+        except KeyError:
+            if default is self.__marker:
+                raise
+            else:
+                return default
+        else:
+            del self[key]
+            return value
+
+    def popitem(self):
+        try:
+            key = next(self.keys())
+        except StopIteration:
+            raise KeyError("Store empty")
+        else:
+            value = ensure_bytes(self[key])
+            del self[key]
+            return (key, value)
+
     def dir_path(self, path=None):
         store_path = normalize_storage_path(path)
         dir_path = self.path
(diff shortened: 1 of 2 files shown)


alimanfoo:
Thanks, very helpful to see this.

I'm realising that I don't really understand memory mapped files very well at all, so these questions might not make sense. But these come to mind:

* What do we gain by using `numpy.memmap` rather than `mmap` directly? 
* With memory-mapped files, the underlying file needs to be kept open (is that true?). If lots of chunks are being read, could this cause any problems due to lots of open files?
* What about writing? Would it ever make sense to also write data via a memory map?

One thing I'm really not clear about is the division of responsibility between operating system and Python in terms of who manages the memory mapping and owns the memory. I believe it is the operating system, but I'm not clear. 

If this is the case, what happens when two threads (or processes) open a memory map onto the same file? Do they access the same memory map, or are two separate memory maps created? 

What about if a single thread (or process) sequentially opens a memory map onto the same file multiple times? Will the same memory map be re-used, or will it be created again from scratch each time?

jakirkham:
As a side note, would be good to get the 3 testing PRs this depends on (listed in the OP) reviewed and hopefully merged in the interim. Those seems like useful additions independent of this PR.

alimanfoo:
Thanks @jakirkham for the info, much appreciated - learning a lot. Funny that you can interact with the mmap after the original file is closed, I did see that in the numpy memmap source code but wasn't sure I was seeing it right. I.e., this works:

```
In [4]: with open('test-mmap.dat', 'r+b') as f: 
   ...:     mm = mmap.mmap(f.fileno(), 0) 
   ...:                                                                                                                                                       

In [5]: mm[:]                                                                                                                                                 
Out[5]: b'asdlkjdaslkjdasljkdasjlkdas\n'

In [6]: mm[:5] = b'aaaaa'                                                                                                                                     
```

I'll try to look at the upstream PRs ASAP.

jakirkham:
Right that’s because `mmap` duplicates the file descriptor. It’s both useful and odd.

alimanfoo:
Just a short note to say that in reading around memory-mapped files, I happened again on information about LMDB ([home page](http://www.lmdb.tech/doc/), [wikipedia page](https://en.wikipedia.org/wiki/Lightning_Memory-Mapped_Database)). I don't understand the details but get the sense that if we want to explore the potential benefits of memory-mapping then we may do well to build on existing key-value store implementations where a lot of thought may already have gone into managing memory maps while also providing other useful properties like support for concurrent access. Zarr already has an LMDBStore class, which I have found extremely fast in some casual benchmarks (e.g., substantially faster then berkeley DB). LevelDB also comes up in discussions of memory-mapping, Zarr doesn't have a store class for this yet but would probably be straightforward via [plyvel](https://plyvel.readthedocs.io/en/latest/index.html).

This is not to stifle exploration of possible uses of Python's mmap module, but just thought I'd make this connection.

jhamman:
Btw, it also sounds like MongoDB has a [memory mapping storage backend](https://docs.mongodb.com/manual/faq/storage/#mmapv1-storage-engine). This will also be available once #372 is completed.

jbaayen:
Is there a plan to merge this PR at some point?

jakirkham:
I don't personally have time to work on this. Though if this is of interest to you and you are willing to take it up, would encourage you to resubmit and will close this out

jbaayen:
This can be implemented in recent Zarr versions using a simple subclass:

```
class MemoryMappedDirectoryStore(DirectoryStore):
    def _fromfile(self, fn):
        return memoryview(np.memmap(fn, mode="r"))
```

joshmoore:
Interesting, so @jakirkham fixed it himself in #503? :smile:

joshmoore:
Shall we close then @jakirkham or is there anything you want to capture? Maybe just @jbaayen's example in the tutorial?

jakirkham:
Possibly. Though we may want this supported in `NestedDirectoryStore` as well, which would make this still desirable

joshmoore:
Hmmm.... I thought that would have worked through inheritance. Ok. Happy to leave this open until someone gets a chance to look at it more closely.

jakirkham:
Ah meaning one should inherit from `NestedDirectoryStore` and then override this method? Sure I suppose that would work

jakirkham:
FWIW a variant using [the Python builtin `mmap`]( https://docs.python.org/3/library/mmap.html ) instead of `np.memmap` could be

```python
class MemoryMappedDirectoryStore(DirectoryStore):
    def _fromfile(self, fn):
        with open(fn, "rb") as fh:
            return memoryview(mmap.mmap(fh.fileno(), 0, prot=mmap.PROT_READ))
```

jakirkham:
Filed a doc issue ( https://github.com/zarr-developers/zarr-python/issues/1245 )

PR #401, opened 2019-02-12
Use binary_type in MongoDBStore.__getitem__ for Python 2 only

jakirkham:
Previously there were some [test failures on Python 2]( https://travis-ci.org/zarr-developers/zarr/jobs/473753393#L2148 ) and referenced in [this comment (and later discussion)]( https://github.com/zarr-developers/zarr/pull/372#issuecomment-450667294 ). Turns out on Python 2 a `bson.Binary` instance is returned, which subclasses `bytes`, as [explained in the docs]( http://api.mongodb.com/python/current/python3.html#id3 ). So the coercion to `bytes` is only needed on Python 2 to convert `bson.Binary` to `bytes`. As Python 3 already returns a `bytes` instance, there is nothing we need to do there. Thus we restrict the `binary_type` call to Python 2. Should avoid a copy on Python 3.

xref: https://github.com/zarr-developers/zarr/pull/372

TODO:

### DIFF
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -2147,7 +2147,7 @@ class MongoDBStore(MutableMapping):
         if doc is None:
             raise KeyError(key)
         else:
-            return binary_type(doc[self._value])
+            return doc[self._value]
 
     def __setitem__(self, key, value):
         value = ensure_bytes(value)
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -2147,7 +2147,15 @@ class MongoDBStore(MutableMapping):
         if doc is None:
             raise KeyError(key)
         else:
-            return doc[self._value]
+            value = doc[self._value]
+
+            # Coerce `bson.Binary` to `bytes` type on Python 2.
+            # PyMongo handles this conversion for us on Python 3.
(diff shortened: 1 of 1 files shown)


jhamman:
Still seems to be failing. This by the way was why I opened #393.

jakirkham:
That's right. Just wanted to see if that was still the case after the dust settled.

After doing a little research, it looks like we can safely constrain this coercion to Python 2 only. Have updated the PR to do that. Should make it easier for us to remember to remove when we do drop Python 2.

Could you please give it another look? :)

jhamman:
yep, this looks right. I agree, this will make it easier to undo when PY2 goes away. Thanks.

jakirkham:
Thanks @jhamman :)

alimanfoo:
Thanks @jakirkham for this, nice to know exactly what's going on :+1:

jakirkham:
Of course. 😄

FWIW I think we can get a better handle on this behavior by adding a flag to the `ensure_*` functions to handle subclasses of the expected type (e.g. `bytes` in this case). Opened PR ( https://github.com/zarr-developers/numcodecs/pull/173 ) along these lines.

PR #402, opened 2019-02-12
Fix pep8speaks

jakirkham:
Tries to fix some false positives from pep8speaks using [this configuration doc]( https://github.com/OrkoHunter/pep8speaks#configuration ) to guide the way.

TODO:
* [ ] Add unit tests and/or doctests in docstrings
* [ ] Add docstrings and API docs for any new/modified user-facing classes and functions
* [ ] New/modified features documented in docs/tutorial.rst
* [ ] Changes documented in docs/release.rst
* [ ] Docs build locally (e.g., run ``tox -e docs``)
* [ ] AppVeyor and Travis CI passes
* [ ] Test coverage is 100% (Coveralls passes)

PR #403, opened 2019-02-12
Set pycodestyle's line length limit to flake8's

jakirkham:
This should make sure that pycodestyle doesn't raise false positives due to long lines in our code. Hopefully should also help with pep8speaks, which uses pycodestyle.

TODO:

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -52,3 +52,6 @@ commands =
 
 [flake8]
 max-line-length = 100
+
+[pycodestyle]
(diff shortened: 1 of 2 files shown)


alimanfoo:
Hi @jakirkham, this and #402 look the same? Should we be looking at one rather than the other?

jakirkham:
Yeah, sorry, was trying to figure out what would work, but it seems like neither do. 😕 Am probably missing something.

PR #417, opened 2019-03-19
Fixtureize and parametrize h5py copy tests.

QuLogic:
The main reason for this is that closing h5py files using `atexit` seems to be a bit buggy. If I run with `pytest`, it's fine, but if I use `python setup.py test`, it will crash on exit.

Using pytest fixtures, we can avoid `atexit` entirely and files will be closed at the end of the test. Unfortunately, this means `unittest.TestCase` cannot be used, so switch to a plain class works. With the fixture, we can parametrize zarr vs. h5py and drop the extra classes.

TODO:
* [N/A] Add docstrings and API docs for any new/modified user-facing classes and functions
* [N/A] New/modified features documented in docs/tutorial.rst
* [N/A] Changes documented in docs/release.rst
* [N/A] Docs build locally (e.g., run ``tox -e docs``)

### DIFF
--- a/zarr/tests/test_convenience.py
+++ b/zarr/tests/test_convenience.py
@@ -1,5 +1,4 @@
 import atexit
-import os
 import tempfile
 import unittest
 from numbers import Integral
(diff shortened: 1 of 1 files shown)


QuLogic:
Note: most of the beginning change is just an indent-level increase so I could put the h5 file in a context manager.

joshmoore:
@QuLogic : are you still interested in taking this forward? If so, there are a few conflicts in your test changes.

joshmoore (approved):
Before & After without `h5py` installed:
```
=========================== 29 passed, 45 skipped in 0.37s ===========================
```
Before & After with `h5py` installed:
```
================================= 74 passed in 4.89s =================================
```
:+1:

PR #432, opened 2019-04-21
Fix `collections`'s `DeprecationWarning`s

jakirkham:
When importing abstract base classes from `collections` (e.g. `MutableMapping`), a `DeprecationWarning` was issued stating that these would no longer be available in Python 3.8 and that we should use `collections.abc`s for this purpose instead. Here we handle importing `MutableMapping` in the `compat` module and have other modules import it from there. This allows us to have the right Python 2/3 behavior while avoiding these `DeprecationWarning`s.

TODO:


jakirkham:
Planning on merging Monday evening if no comments.

PR #445, opened 2019-06-26
Set file permissions in DirectoryStorage according to umask

### LINKED ISSUE #320: Discussion with Darr developers
Ran across @gbeckers's Darr library recently, which seems very similar in intent and behavior to Zarr. Would be great to compare and contrast Darr and Zarr to see where the two implementations can learn from each other and possibly work together.

### LINKED ISSUE #325: Set file permissions accoding to umask
Zarr uses temporary files that are moved in place to write meta-files and chunks. As a side effect, the created containers are only readable by the user who created them (as is default for temporary files). A better approach might be to set the file permissions after the move according to the users `umask`.

eddienko:
Fixes #325

This is my take on having DirectoryStorage play nice with the local defined umask.

### DIFF
--- a/zarr/tests/test_storage.py
+++ b/zarr/tests/test_storage.py
@@ -721,6 +721,13 @@ class TestDirectoryStore(StoreTests, unittest.TestCase):
         store['foo'] = b'bar'
         assert os.path.isdir(path)
 
+        # check correct permissions
+        stat = os.stat(path)
+        mode = stat.st_mode & 0o666
+        umask = os.umask(0)
+        os.umask(umask)
+        assert mode == (0o666 & ~umask)
+
         # test behaviour with file path
         with tempfile.NamedTemporaryFile() as f:
             with pytest.raises(ValueError):
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -74,6 +74,14 @@ def _path_to_prefix(path):
     return prefix
 
 
+def _NamedTemporaryFileUmask(*args, **kargs):
+    fdesc = tempfile.NamedTemporaryFile(*args, **kargs)
+    umask = os.umask(0)
+    os.umask(umask)
+    os.chmod(fdesc.name, 0o666 & ~umask)
+    return fdesc
+
+
 def contains_array(store, path=None):
     """Return True if the store contains an array at the given logical path."""
     path = normalize_storage_path(path)
@@ -752,9 +760,9 @@ class DirectoryStore(MutableMapping):
         # write to temporary file
         temp_path = None
         try:
-            with tempfile.NamedTemporaryFile(mode='wb', delete=False, dir=dir_path,
-                                             prefix=file_name + '.',
-                                             suffix='.partial') as f:
+            with _NamedTemporaryFileUmask(mode='wb', delete=False, dir=dir_path,
+                                          prefix=file_name + '.',
+                                          suffix='.partial') as f:
                 temp_path = f.name
                 f.write(value)
 


zarr/tests/test_storage.py
```diff
@@ -721,6 +721,14 @@ def test_filesystem_path(self):
...
         assert os.path.isdir(path)
 
+        # check correct permissions
+        stat = os.stat(path)
+        mode = stat.st_mode & 0o666
+        umask = os.umask(0)
+        os.umask(umask)
+        assert mode == (0o666 & ~umask)
+
+
```
jrbourbeau: This extra blank line is causing a `flake8` [linting error](https://travis-ci.org/zarr-developers/zarr-python/jobs/553953573#L2486). Removing the extra line should fix the failing Travis CI build:

```suggestion
```

eddienko: > This extra blank line is causing a `flake8` [linting error](https://travis-ci.org/zarr-developers/zarr-python/jobs/553953573#L2486). Removing the extra line should fix the failing Travis CI build:

Thanks for the heads up. Fixed now.

eddienko:
@manolinux I cannot reproduce the error that you are seeing - my understanding is that if you are the owner of a file you can chmod it. Could you add more info about the permissions that you have in those directories or a simple script that replicates the problem?

manolinux:
No, the problem is that I'm not the owner, but I belong to the owner's
group. This is quite normal when using some shared storage (mounted CIFS
volume). That's why I cannot chmod (need the owner user, but python is
spawned with a group user). To overcome this, the Zarr file  should be
created with umask-derived  permissions at once.

tempfile.NamedTemporaryFile creates a file with 600 (security concerns)
permissions. Then, proposed commit changes permissions to umask-based. This
is going to work perfectly if you are the real owner, but not in the case
of group-based permissions

Buy anyway, this may be my particular case, don't pay much attention if you
consider this a security risk.

Kind regards, and go on with the project, seems great!
Manuel.

eddienko:
Sorry but I tried lots of combinations with different group permissions and users and still cannot reproduce. If anyone can provide a test where this fails I am happy to have a look at it.

manolinux:
Hi, don't worry anymore about this.

The problem is my configuration of shared CIFS filesystem that maps users to a uid,gid distinct than current user, that's why tempfile.NamedTemporaryFile is unable to chmod.

I have to keep my tempfile._mkstemp_inner patching for going on with Zarr, but your code has then nothing to do with the issue.

Sorry for the inconvenience, and thanks a lot for your support!
Manuel.

d-v-b:
@jrbourbeau Is there anything blocking this PR? Our group would really love to see this merged, we've had a lot of permissions-related headaches stemming from this issue.

manolinux:
Hi, It seems there is not a scheduled 2.3.3 for the moment that could fix
some of these issues that have been corrected by community.
My guess (read something about it) is that efforts are put now in
redefining underlying protocols and unification with N5.py.

As early adopters of Zarr, I think we're paying a high price; I have had to
use some of the PRs by downloading repository from git and applying
manually:

https://gist.github.com/adam-p/15413fabef6cffecd897

PRs installed this way seem to install ok with python setup install

I'm opening an issue requesting a 2.3.3 version, and see if people vote for
it.

Kind regards!

jrbourbeau:
Apologies for the delayed response @d-v-b. I'm not a Zarr maintainer, so I don't control which PRs are merged. Others like @jakirkham and @alimanfoo are more equipped to review your changes in this PR (although please keep in mind they may have very limited bandwidth right now)

That said, it looks like in the original issue, it was proposed that we could move away from using temporary files here to resolve permissions issues while avoiding the need to query the current umask and then chmod (xref https://github.com/zarr-developers/zarr-python/issues/325#issuecomment-437078563 and https://github.com/zarr-developers/zarr-python/issues/325#issuecomment-437332397). Not sure if that approach that should be taken here, or if the current changes in this PR are generally okay.

d-v-b:
For me, this PR does not fix permissions problems for `attributes.json` files when using `N5Store`. I'll see where that comes from.

eddienko:
Hi,

I generally agree that this should be considered a bug and should be fixed prior to having a more general solution if there is any. Only thing however is to check that it does not break anything else, though I guess tests should do that.

@d-v-b for me this works fine as well with N5Store:

```
import zarr
store = zarr.N5Store('/tmp/array.n5')
z = zarr.zeros((10, 10), chunks=(5, 5), store=store, overwrite=True)
z[:] = 42
```

and

```
$ umask
022
$ ls -la array.n5
total 8
drwxr-xr-x   5 user  wheel  160  5 Oct 07:35 .
drwxrwxrwt  17 root   wheel  544  5 Oct 07:35 ..
drwxr-xr-x   4 user  wheel  128  5 Oct 07:35 0
drwxr-xr-x   4 user  wheel  128  5 Oct 07:35 1
-rw-r--r--   1 user  wheel  244  5 Oct 07:35 attributes.json
```

jakirkham:
It sounds like this is addressed with PR ( https://github.com/zarr-developers/zarr-python/pull/493 ). Would you agree @eddienko? If so, happy to close?

Issue #295, opened 2018-08-23
Optimisation for the __contains__ method of storage.LRUStoreCache


shikharsg:
I have a few TB large dataset with 11 parameters and about 100000 chunks, and am storing it in azure blob using the `ABSStore` mutable mapping. When I do `zarr.open_group(store=store, mode='r')` with store as `zarr.LRUStoreCache(max_size=2**33, store=zarr.storage.ABSStore('testcontainer', 'mydataset', BLOB_ACCOUNT_NAME, BLOB_ACCOUNT_KEY))`, it takes about 45 seconds to open the group. Without the `LRU` wrapper the `open_group` operation is instantaneous. I traced the problem to the `__contains__` method in `LRUStoreCache` mutable mapping wrapper(`open_group` calls the `contains_array` method). The `__contains__` method[(here)](https://github.com/zarr-developers/zarr/blob/master/zarr/storage.py#L1771) in `LRUStoreCache`  is implemented by listing all the keys in the mutable mapping of the underlying store, and therefore, all 100000 chunks are listed before checking for existence. In the context of cloud storage this can cause significant overhead.

This is the method of `LRUStoreCache` now:
```python
    def __contains__(self, key):
        with self._mutex:
            if self._contains_cache is None:
                self._contains_cache = set(self._keys())
            return key in self._contains_cache
```

when I changed it to this:
```python
    def __contains__(self, key):
        return key in self._store
```
the `open_group` operation is almost instantaneous as the `__contains__` method of the underlying `ABSStore` class uses the `exists` option on azure blob and so doesn't have to list all keys.

martindurant:
You may be interested in the [consolidated metadata](https://github.com/zarr-developers/zarr/pull/268) PR, which would grab the metadata in a single call, and thereafter access in memory (including for `__contains__`).

shikharsg:
Thanks for that link @martindurant. I've been following that PR. I'm a little confused though as I'm new to zarr, so sorry if my questions sound stupidwill . How the consolidated metadata feature help in the `__contains__` call? Will it be possible to store all the keys of the mutable mapping in the `.zmetadata` key?

Also, I think the way the `__contains__` method is implemented in `LRUStoreCache` is not efficient as it lists all the keys of the underlying store(and caches them), before checking for existence of the key. In my opinion it would be better to let the underlying store implement the `__contains__` call

martindurant:
> Will it be possible to store all the keys of the mutable mapping in the .zmetadata key?

[...]

PR #447, opened 2019-06-30
optmizaton for LRUStoreCache.__contains__

shikharsg:
Closes https://github.com/zarr-developers/zarr/issues/295.

As mentioned in https://github.com/zarr-developers/zarr/issues/295, `LRUStoreCache.__contains__` made a call to `self._keys`, which can take long, if for example we have an array with terabytes of data and millions of keys stored on cloud, the call to `self._keys` will first list all those keys and cache them. This PR removes the `self._keys` call from `__contains__`. The call to `self._store.keys()` still exists in the `self.keys` method and keys will still be cached, but just not when the `__contains__` method is called.

TODO:

### DIFF
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -1763,8 +1763,22 @@ class LRUStoreCache(MutableMapping):
 
     def __contains__(self, key):
         with self._mutex:
-            if self._contains_cache is None:
-                self._contains_cache = set(self._keys())
+            # first check the keys cache since it has ALL the keys,
+            # then check the contains cache, else create a new
+            # contains cache
+            if self._keys_cache:
+                self._contains_cache = self._keys_cache
+            elif self._contains_cache:
+                # the below 2 if statements are not done together
+                # as an 'and' because we can possibly avoid one expensive
+                # __contains__ call to the underlying store
+                if not key in self._contains_cache:
+                    if key in self._store:
+                        self._contains_cache.add(key)
+            else:
+                self._contains_cache = set()
+                if key in self._store:
+                    self._contains_cache.add(key)
             return key in self._contains_cache
 
     def clear(self):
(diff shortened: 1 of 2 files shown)


martindurant:
My immediate thought here: what happens if the underlying store has changed since the last caching of the keys?

shikharsg:
> My immediate thought here: what happens if the underlying store has changed since the last caching of the keys?

If you see the `__delitem__` and the `__setitem__` methods of `LRUStoreCache`, if they are called, the cache is invalidated. Are you talking about the case when some other user of the underlying store, changes the store since the last caching of the keys? In which case I guess that was a problem even before this PR right?

martindurant:
OK, OK, I got what this is doing - sorry I didn't immediately click in - this is for calls on `__contains__` of the target store, where you haven't explicitly listed keys. 

Is there a case for having the contains_cache be size-limited, since we are thinking there may be a very large number of keys?

I am OK with the implementation, but I haven't looked at the tests yet.

(the point about the target storage changing from some other process does stand, but is general to this caching mechanism, so not relevant to this particular change)

shikharsg:
I initially did think of limiting the size of contains_cache, but then if you see LRUStoreCache, the keys method is still being cached, just not through the contains methods, but through the keys method, i.e. if a call to _keys is made, ALL the keys will be cached. So I was wondering if it makes sense to put a size limit on the contains cache, when there is no size limit for keys cache. I guess we could:
1. put a size limit on both keys cache and contains cache(the default size for both being unlimited to maintain backwards compatibility)
2. put a size limit on contains cache and completely remove keys caching

I like the 2nd one, but it would break backwards compatibility so actually I'm not sure which one of the two options we should do.

joshmoore:
Hi @shikharsg. Just going through open issues for a review and found this linked from #295. Any thoughts from your side?

jhamman:
Closing this out as stale. We'll keep the feature request open.

Issue #259, opened 2018-05-07
HTML jstree doesn't work in jupyter lab


alimanfoo:
#### Problem description

When calling ``group.tree()`` in a jupyter notebook environment, an expandable tree should be rendered via jstree. This works in classical jupyter notebook but doesn't work in jupyter lab.

#### Version and installation information

* Value of ``zarr.__version__``: 2.2.0
* Jupyter Lab version: 0.32.1
* Version of Python interpreter: 3.6.5
* Operating system (Linux/Windows/Mac): Linux
* How Zarr was installed (e.g., "using pip into virtual environment", or "using conda"): conda

jakirkham:
Can't find the exact issue, but JupyterLab apparently disallows inlining the `script` tag. So this is likely causing us issues. I'm not sure what the easiest solution is, but one option would be to rewrite this using ipywidgets. Maybe someone else knows a simpler option.

ref: https://github.com/jupyterlab/jupyterlab/issues/2710#issuecomment-316476443

jakirkham:
Looks like PR ( https://github.com/jupyterlab/jupyterlab/pull/2595 ) is where inlined `script` tags got sanitized. A fair bit of discussion is happening in issue ( https://github.com/jupyterlab/jupyterlab/issues/3118 ) about whether this is a good idea and how best to address user expectations. Maybe this is one of those wait and see problems.

alimanfoo:
Thanks @jakirkham, I [added a comment](https://github.com/jupyterlab/jupyterlab/issues/3118#issuecomment-387373986) re zarr.

jakirkham:
Thanks for doing that @alimanfoo. :)

jakirkham:
cc-ing @jasongrout (as we talked about this one in person ;), for advice on how we should fix this.

jakirkham:
Maybe [this doc page]( https://jupyterlab.readthedocs.io/en/stable/developer/extension_dev.html ) is useful? Not sure if those are the most up-to-date ones though.

jakirkham:
Probably also worth looking at how XArray is approaching this. ( https://github.com/pydata/xarray/pull/1820 )

jakirkham:
So one short-term option might be to add `_repr_json_`. This is something that JupyterLab knows how to display. That said, it does leave some things to be desired.

[...]

PR #450, opened 2019-07-04
Use ipytree in `.tree` representation

jakirkham:
Fixes https://github.com/zarr-developers/zarr-python/issues/259

Reimplements the `.tree` representation in the Jupyter Notebook using [ipytree]( https://github.com/QuantStack/ipytree ), a Jupyter-widget wrapped jsTree implementation. This is similar to what we had before in that it still uses jsTree under-the-hood. However this is significantly easier for us to maintain and test. Not to mention it works seamlessly in both the legacy notebook and JupyterLab. This does require an extra installation step on the part of the user though. FWIW this was surprisingly quick to implement. 🙂

~~It would be good to reference how users [should install ipytree]( https://github.com/QuantStack/ipytree#installation ). Also still needs a little work to get testing on CI going.~~ That said, this could be handy for a slide or a blogpost while we iron these things out.

TODO:


alimanfoo:
This certainly cleans things up **a lot**! Very nice.

jakirkham:
Yeah it's also much easier to test. We can inspect nodes and see how they relate to each other. Haven't dug into that, but it is definitely an option. 🙂

jakirkham:
Have pushed the notebook with this change for comparison. Can [see with nbviewer]( https://nbviewer.jupyter.org/github/zarr-developers/zarr-python/blob/e7a1a9a2bb1042cf2972965516467aaa32100eba/notebooks/repr_tree.ipynb ).

Different icons don't work as well. So just dropped those examples.

alimanfoo:
Looks very nice :-)

 I'm getting an odd behaviour viewing on my phone, tapping on a node makes
it disappear. Then tapping another node makes the first node reappear, but
the tapped node disappear. Maybe something to do with styling for selected
node. Haven't checked on my laptop yet, don't know if specific to phone.

jakirkham:
Thanks for the feedback. This seems to be caused by selecting the icons in nbviewer. Doesn't happen in the actual Jupyter notebook.

jakirkham:
Have pushed a change that disables all nodes so they cannot be selected. This seems to workaround the issue.

jakirkham:
Here's an [updated link]( https://nbviewer.jupyter.org/github/zarr-developers/zarr-python/blob/11cbdb1e262167a8905c0cea1d879a6ec706a007/notebooks/repr_tree.ipynb ).

alimanfoo:
Cool, thanks John, looks good.

jakirkham:
@alimanfoo, would you have a chance to do a code review?

alimanfoo:
Hi @jakirkham, code looks good to my eye, however looking at the [example notebook](https://nbviewer.jupyter.org/github/zarr-developers/zarr-python/blob/11cbdb1e262167a8905c0cea1d879a6ec706a007/notebooks/repr_tree.ipynb) I don't think the `expand` argument is working correctly, because the trees with `expand=True`, `expand=1`, `expand=2` and `expand=3` all look the same.

jakirkham:
Good catch @alimanfoo! Sorry for the long delay. As with most bugs this was a simple thing 😉

Have rerun on the notebook. Results can be seen [on nbviewer]( https://nbviewer.jupyter.org/github/zarr-developers/zarr-python/blob/45b3ff21c0d12c3913e12899086ab28327ebcf95/notebooks/repr_tree.ipynb ).

jakirkham:
Not sure what is up with Coveralls. Tried messing with it last night and this morning, but it still doesn't seem to be working. Otherwise CI passes.

jakirkham:
Thanks for the review @jrbourbeau! Nothing more from me. Merge away 😀

PR #452, opened 2019-07-08
Drop `decode` from `ConsolidatedMetadataStore`

jakirkham:
Fixes the issue described in [this SO question]( https://stackoverflow.com/q/56904331/3877089 ).

This call to `decode` has been made unnecessary thanks to the `json_loads` utility function used here. Unfortunately this code was left over when that was added in. Drop this now unneeded code so as to better handle `bytes`-like data.

cc @martindurant

TODO:

### DIFF
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -2451,11 +2451,7 @@ class ConsolidatedMetadataStore(MutableMapping):
         self.store = store
 
         # retrieve consolidated metadata
-        if sys.version_info.major == 3 and sys.version_info.minor < 6:
-            d = store[metadata_key].decode()  # pragma: no cover
-        else:  # pragma: no cover
-            d = store[metadata_key]
-        meta = json_loads(d)
+        meta = json_loads(store[metadata_key])
 
         # check format of consolidated metadata
         consolidated_format = meta.get('zarr_consolidated_format', None)


martindurant:
You are too fast for me.
I seem to remember that there was a reason for doing it like this, but I can't now remember it. Is a similar change required in consolidate_metadata? Perhaps there should be a roundtrip test for non-bytes stores.

jakirkham:
I'm just familiar with this stuff since I did a bunch of work on it not that long ago.

Yeah there is some subtlety between Python 2.7, 3.5, and 3.6 with what types of data `json.loads` can use. However all of them are able to use `unicode` typed data in `json.loads`. Under the hood, `json_loads` converts everything to `unicode` (if it isn't already) and then runs `json.loads` on it. So this should already handle the corner cases I believe you are referring to.

Nope, that was [already done]( https://github.com/zarr-developers/zarr-python/blob/v2.3.2/zarr/convenience.py#L1123 ). We made the same change here too. Just forgot to sweep out the unneeded code.

That's probably not a bad idea, but I won't tackle that in this PR. If you'd like to raise an issue so we can keep track of it, that would be helpful.

jakirkham:
FWIW Travis CI already passed here. It just hasn't updated the status for some reason.

martindurant:
OK, so this is fine with me then

jakirkham:
Thanks for the reviews! 😄

PR #456, opened 2019-07-17
Use uniform chunking when specifying chunks as an integer

### LINKED ISSUE #364: Specifying chunks with an integer

```python
In [1]: import zarr                                                             

In [2]: z = zarr.zeros(shape=(4, 4), chunks=1)                                  

In [3]: z.chunks                                                                
Out[3]: (1, 4)
```

#### Problem description

When specifying chunks with an integer, initially thought this would behave the same way that Dask does (please see example below). Namely Zarr would apply this chunking for all dimensions. However it seems Zarr only applies this chunking to the first dimension and lets the rest of the dimensions have a chunk size that matches the shape for that dimension.

```python
In [1]: import dask.array as da                                                 

In [2]: a = da.zeros(shape=(4, 4), chunks=1)                                    

In [3]: a.chunksize                                                             
Out[3]: (1, 1)
```

Please provide the following:

* Value of ``zarr.__version__``: 2.2.0
* Value of ``numcodecs.__version__``: 0.5.5
* Version of Python interpreter: 3.6.6
* Operating system (Linux/Windows/Mac): macOS
* How Zarr was installed (e.g., "using pip into virtual environment", or "using conda"): Using Conda

jrbourbeau:
This PR updates `normalize_chunks` to apply the same chunking to all dimensions when `chunks` is specified as an integer. Closes #364.

TODO:

### DIFF
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -4,6 +4,9 @@ Release notes
 Upcoming Release
 ----------------
 
+* Use uniform chunking for all dimensions when specifying ``chunks`` as an integer.
+  By :user:`James Bourbeau <jrbourbeau>`; :issue:`456`
+
 * Rename ``DictStore`` to ``MemoryStore``.
   By :user:`James Bourbeau <jrbourbeau>`; :issue:`455`
 
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -5,6 +5,7 @@ Upcoming Release
 ----------------
 
 * Use uniform chunking for all dimensions when specifying ``chunks`` as an integer.
+  Also adds support for specifying ``-1`` to chunk across an entire dimension.
   By :user:`James Bourbeau <jrbourbeau>`; :issue:`456`
 
 * Rename ``DictStore`` to ``MemoryStore``.
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -1170,7 +1170,7 @@ better performance, at least when using the Blosc compression library.
 The optimal chunk shape will depend on how you want to access the data. E.g.,
 for a 2-dimensional array, if you only ever take slices along the first
 dimension, then chunk across the second dimenson. If you know you want to chunk
-across an entire dimension you can use ``None`` within the ``chunks`` argument,
+across an entire dimension you can use ``None`` or ``-1``within the ``chunks`` argument,
 e.g.::
 
     >>> z1 = zarr.zeros((10000, 10000), chunks=(100, None), dtype='i4')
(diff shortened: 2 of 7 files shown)


jrbourbeau:
Currently, some tests which rely on hard coded values for the output of `.hexdigest()` (e.g. `zarr/tests/test_core.py::TestArray.test_hexdigest`) are failing due to the change in chunking behavior.

I suspect the hardcoded hashes can just be updated to their new values. However, I wanted to check to see if this will cause any backwards compatibility issues.

alimanfoo:
I think it should be OK to update the hashes in this case.

jrbourbeau:
Hmmm the Python 3.7 failure on Travis seems unrelated to the changes in this PR (in fact, the same failure seems to be [occurring on `master`](https://travis-ci.org/zarr-developers/zarr-python/builds/560181617)).

It looks like the failure correlates with a new release of `s3fs` earlier today. After inspecting http://zarr-demo.s3-eu-west-2.amazonaws.com/, it appears there is still a `foo/bar/baz` key in the Zarr test data. cc @martindurant

alimanfoo:
Confirm this breaks for me locally after upgrading to s3fs 0.3.0. It's the example in the [LRUStoreCache docstring](https://zarr.readthedocs.io/en/stable/api/storage.html#zarr.storage.LRUStoreCache) that breaks. Strange as I can read fine from the s3fs store directly (i.e., without the cache).

alimanfoo:
Narrowed this down to s3fs missing some keys. E.g.:

```
In [17]: store = s3fs.S3Map(root='zarr-demo/store', s3=s3, check=False) 

In [18]: list(store.keys())                                                           
Out[18]: 
['.zattrs',
 '.zgroup',
 'foo/.zattrs',
 'foo/.zgroup',
 'foo/bar/.zattrs',
 'foo/bar/.zgroup',
 'spam/.zgroup',
 'spam/eggs',
 'spam/eggs/.zarray',
 'spam/eggs/0.0',
 'spam/eggs/0.1',
 'spam/eggs/1.0',
 'spam/eggs/1.1']

In [19]: 'foo/bar/baz/.zarray' in store                                               
Out[19]: True

In [20]: 'foo/bar/baz/.zarray' in list(store.keys())                                  
Out[20]: False
```

cc @martindurant

alimanfoo:
Raise upstream: https://github.com/dask/s3fs/issues/199

jrbourbeau:
I can confirm that `pytest -v --doctest-modules zarr/storage.py::zarr.storage.LRUStoreCache` passes for me locally after upgrading `fsspec`.

Thanks @alimanfoo for investigating / raising the issue upstream and @martindurant for the fix and quick release!

jrbourbeau:
+1 for adding -1 chunk support. Just pushed a couple commits that should handle the -1 case

alimanfoo:
Awesome, thanks @jrbourbeau and @martindurant!

PR #492, opened 2019-10-25
Use pytest.importorskip in tests

### LINKED ISSUE #229: Consider using pytest.importorskip
https://github.com/zarr-developers/zarr/pull/227#pullrequestreview-86556798

jrbourbeau:
This PR replaces (almost all) occurrences of `try`/`except` blocks to determine if a library is available with [`pytest.importorskip`](http://doc.pytest.org/en/latest/skipping.html#skipping-on-a-missing-import-dependency). I wasn't able to replace this `LZMA` import

https://github.com/zarr-developers/zarr-python/blob/55ae9eb95d2c7c59582fed02f1dba88298f10fda/zarr/tests/test_core.py#L1944-L1947

because it's used in other conditional statements in `test_core.py`. 

Also added a `skip_test_env_var` utility function to `zarr/tests/util.py` to replace some of the boilerplate code related to skipping tests based on environment variables (e.g. `ZARR_TEST_ABS`)

Closes #229 

TODO:

### DIFF
--- a/zarr/tests/util.py
+++ b/zarr/tests/util.py
@@ -1,6 +1,9 @@
 # -*- coding: utf-8 -*-
 import collections
 from collections.abc import MutableMapping
+import os
+
+import pytest
 
 
 class CountingDict(MutableMapping):
@@ -36,3 +39,10 @@ class CountingDict(MutableMapping):
     def __delitem__(self, key):
         self.counter['__delitem__', key] += 1
         del self.wrapped[key]
+
+
+def skip_test_env_var(name):
+    """ Checks for environment variables indicating whether tests requiring services should be run
+    """
+    value = os.environ.get(name, '0')
+    return pytest.mark.skipif(value == '0', reason='Tests not enabled via environment variable')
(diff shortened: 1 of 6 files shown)


zarr/tests/util.py
```diff
@@ -36,3 +39,10 @@ def __setitem__(self, key, value):
     def __delitem__(self, key):
         self.counter['__delitem__', key] += 1
         del self.wrapped[key]
+
+
+def skip_test_env_var(name):
+    """ Checks for environment variables indicating whether tests requiring services should be run
+    """
+    value = os.environ.get(name, '0')
```
jakirkham: Should we handle empty strings? If so, this might help.

```suggestion
    value = os.environ.get(name, '') or '0'
```

jakirkham:
Thanks for working on this James! This looks much better.

As CIs install some sets of these, we may want to set the environment variables on CI.

zarr/tests/util.py
jrbourbeau: Hmm I'm not sure. Is an empty string a common pattern for disabling things with an env var? If so, then it's probably worth supporting here. No strong opinion from me one way or the other

jakirkham: It's more so that it is not a common pattern for enabling things. So the concern would be someone accidentally enabling these.

That said, let's leave it since it was already this way and we can follow-up in another issue/PR.

jakirkham:
Thanks @jrbourbeau! 😄

PR #485, opened 2019-10-17
undo parameter renaming for compatibility with n5-blosc

axtimwalde:
n5-blosc expects the original blosc parameters including 'blocksize'

TODO:
* [ ] Add unit tests and/or doctests in docstrings
* [x] Changes documented in docs/release.rst
* [x] Docs build locally (e.g., run ``tox -e docs``)
* [x] AppVeyor and Travis CI passes
* [x] Test coverage is 100% (Coveralls passes)

PR #496, opened 2019-11-06
Also include required `nthreads` in compression metadata

chris-allan:
Feed on from #485. The N5 reference library blosc compression adapter `saalfeldlab/n5-blosc` expects `nthreads` to be present in the metadata:

 * https://github.com/saalfeldlab/n5-blosc/blob/master/src/main/java/org/janelia/saalfeldlab/n5/blosc/BloscCompression.java#L68

Without `nthreads` reading a blosc compressed N5 dataset, which is the default if you do not specify a specific compression strategy, using the N5 reference library causes exceptions to be thrown of the form:

```
java.lang.IllegalArgumentException: Can not set int field org.janelia.saalfeldlab.n5.blosc.BloscCompression.nthreads to null value
        at sun.reflect.UnsafeFieldAccessorImpl.throwSetIllegalArgumentException(UnsafeFieldAccessorImpl.java:167)
        at sun.reflect.UnsafeFieldAccessorImpl.throwSetIllegalArgumentException(UnsafeFieldAccessorImpl.java:171)
        at sun.reflect.UnsafeIntegerFieldAccessorImpl.set(UnsafeIntegerFieldAccessorImpl.java:80)
        at java.lang.reflect.Field.set(Field.java:764)
        at org.janelia.saalfeldlab.n5.CompressionAdapter.deserialize(CompressionAdapter.java:160)
        at org.janelia.saalfeldlab.n5.CompressionAdapter.deserialize(CompressionAdapter.java:57)
        at com.google.gson.internal.bind.TreeTypeAdapter.read(TreeTypeAdapter.java:69)
        at com.google.gson.Gson.fromJson(Gson.java:927)
        at com.google.gson.Gson.fromJson(Gson.java:994)
        at com.google.gson.Gson.fromJson(Gson.java:967)
        at org.janelia.saalfeldlab.n5.GsonAttributesParser.parseAttribute(GsonAttributesParser.java:78)
        at org.janelia.saalfeldlab.n5.AbstractGsonReader.getDatasetAttributes(AbstractGsonReader.java:95)
...
```

This change does feel a little hacky but I'm throwing it out there for discussion. I'm not 100% on why N5 expects the `nthreads` value in the first place as this seems to be a consumer concern rather than a file format one. It is not present in the equivalent Zarr metadata.

There also seems to be a bit of a disconnect between the default blosc threading configuration here in Zarr vs. N5 but this is probably not the place for that discussion.

/cc @axtimwalde, @joshmoore 

TODO:

### DIFF
--- a/zarr/n5.py
+++ b/zarr/n5.py
@@ -8,6 +8,7 @@ import warnings
 
 import numpy as np
 from numcodecs.abc import Codec
+from numcodecs.blosc import get_nthreads
 from numcodecs.compat import ndarray_copy
 from numcodecs.registry import get_codec, register_codec
 
@@ -418,6 +419,7 @@ def compressor_config_to_n5(compressor_config):
         n5_config['clevel'] = compressor_config['clevel']
         n5_config['shuffle'] = compressor_config['shuffle']
         n5_config['blocksize'] = compressor_config['blocksize']
+        n5_config['nthreads'] = get_nthreads()
 
     elif codec_id == 'lzma':
 


alimanfoo:
FWIW I don't think nthreads should go in the metadata, it is a parameter that the user should control depending on the system.

jakirkham:
Yeah I agree with @alimanfoo. Though it would be good to understand other points of view. Perhaps there is another agreeable solution.

chris-allan:
:+1: from me as well in agreement though I expect it doesn't mean much. I'm happy to try to open a PR making `n5-blosc` more amenable to that data being missing if that is what's decided. I'll wait for @axtimwalde to weigh in before doing anything else.

Incidentally, I've seen no changes to `n5-blosc` since #485 was opened and merged so I'm not sure why `nthreads` wasn't mentioned there. The errors would have been very similar as those for `cname` and `clevel`.

jakirkham:
Thanks for raising this. Agreed let’s hear Stephan’s thoughts 🙂

axtimwalde:
Thanks for raising this, I will fix this with a default value of 1 in n5-blosc.  Background why it's there: JBlosc enables me to set this parameter for each `Compression` instance (https://github.com/Blosc/JBlosc/blob/master/jblosc/src/main/java/org/blosc/JBlosc.java#L45) and I blindly exposed all available parameters to the Compression interface.  This makes as much sense as specifying other compression parameters (none for reading and as a hint for writing) in a Java API where you have more fine grained control over multi-threading.

Let's not introduce this parameter in zarr as it affects only the zarr.n5 when used in combination with n5-zarr.  It will be easier and more meaningful for me to specify defaults in n5-zarr.

chris-allan:
Definitely makes sense to me to adjust the Gson parsing in n5-blosc.

However, I'm not sure I completely understand the `Compression` instance comment. `JBlosc.set_nthreads()` is going to manipulate the c-blosc threadpool via `blosc_set_nthreads()` in C. Semantically equivalent to `numcodec.blosc.set_nthreads()` in Zarr on the Python side. That thread pool is global.

Under the hood of `BloscCompression` it seems like `JBlosc.decompressCtx()` is being used. If we delve down into context based compression in c-blosc each context execution of `parallel_blosc()` just ends up running `blosc_set_nthreads()` anyway. Thus, it doesn't seem like it would make much sense to be on each `Compression` instance. At worst there's the possibility of data loss; Blosc is re-initialized if the thread counts change. I'd have to read more of c-blosc to be completely certain of the semantics.

A static API of `BloscCompression.setNumThreads(threadCount)` which calls `JBlosc.set_nthreads()` and not having it in the `BloscCompression` instance would make more sense to me but maybe I'm missing something.

All of that doesn't really matter much to Zarr though. Happy to take this up on whatever gets opened on the n5-blosc side to reduce noise to everyone else.

axtimwalde:
Done:
https://github.com/saalfeldlab/n5/commit/1d3057835abb7536b15e44d33a37dc5f5310e692
The releases [n5-2.1.3](https://github.com/saalfeldlab/n5/commits/2.1.3) and [n5-blosc-1.0.1](https://github.com/saalfeldlab/n5-blosc/commits/1.0.1) fix the issue.  There is no need currently for a new n5-zarr release because it will work with the updated dependencies.
This update fixes all future default value handling for primitive compression parameters.  Objects could already be null.
I think this means that this pull request can be closed?
Thanks for pointing this out.  We had indeed run into this before together with the renaming issue, fixed only the zarr.n5 that part of it and forgot about the n5 side...

jakirkham:
Thanks Stephan! 😄

chris-allan:
Thanks @axtimwalde! We still need a Zarr release which includes #485 for N5 files written with Zarr to be compatible with the reference library, correct?

Happy to take up the compression conversation on the n5-blosc repository if that's of use.

jakirkham:
We are working on it.

chris-allan:
Thanks for everyone's input! Closing.

PR #501, opened 2019-11-08
WIP: Provide a meta array to change array type returned

jakirkham:
Adds a user provided meta-array to change the array types allocated.

cc @mrocklin (thanks for the tip! 😄)

TODO:


zarr/core.py
```diff
@@ -711,7 +719,7 @@ def _get_basic_selection_zd(self, selection, out=None, fields=None):
 
         except KeyError:
             # chunk not initialized
-            chunk = np.zeros((), dtype=self._dtype)
+            chunk = np.zeros_like(self._meta, shape=(), dtype=self._dtype)
```
mrocklin: Beware, this may restrict you to a somewhat recent version of Numpy

jakirkham: (was just writing that)

mrocklin:
I'm glad to have provided a helpful tip.  Also cc'ing @pentschev, who I'm sure will appreciate seeing others go through the same process that he pioneered.

jakirkham:
So this relies on the `shape` keyword in the NumPy `*_like` functions, which was introduced in NumPy 1.17. IDK if we are comfortable bumping to 1.17.0 as a minimum requirement or if need to add some fallback logic. Feedback on this point would be welcome. 🙂

jakirkham:
This may need some tweaks as `flag.writeable` doesn't seem to always be a thing.

zarr/core.py
pentschev: You can always have a safe wrapper for this, like we do in dask: https://github.com/dask/dask/blob/master/dask/array/utils.py#L288-L334

pentschev:
That's cool @jakirkham that you're adding the meta array here as well. Good luck with that! :)

zarr/core.py
jakirkham: Yeah this is tricky if the storage layer is already producing an array of a different type, this will then cause a conversion. With CuPy for example, we know this is not an option ( https://github.com/cupy/cupy/issues/589 ). I'm not sure if we shouldn't just error out earlier.

pentschev: Not sure I understand, why won't CuPy work? CuPy supports the `*_like` functions in here since https://github.com/cupy/cupy/pull/2171.

jakirkham: Meaning that when we can't use `*_like` functions we will create a NumPy array here, but we may be trying to store CuPy arrays into it. So this would fail because the CuPy arrays would be coerced to NumPy arrays. This is just an example.

jakirkham:
Maybe @alimanfoo or @jrbourbeau can provide some feedback to help determine what we do for older NumPy versions.

zarr/core.py
pentschev: Yeah, I guess that goes back to the array creation issue that we're trying to address in NumPy. :)

jakirkham: Yep :)

jakirkham:
Should add we will need to update [the array creation functions]( https://zarr.readthedocs.io/en/stable/api/creation.html ) accordingly.

alimanfoo:
Hi @jakirkham, this is interesting :-)

Would be cool to understand a bit more about specific goals here. 

I'm guessing one goal is that when slicing a zarr array (``__getitem__``) a user wants the result returned as something other than a numpy array, e.g., a cupy array? If so, is there a potential performance gain, or is it just about convenience to the user?

What's the goal when storing data into a zarr array (``__setitem__``)? 

Re numpy version compatibility, we've only recently upgraded to numpy 1.17 ourselves, I suspect we'll need to be friendly to users who haven't done so yet.

PR #503, opened 2019-11-08
Refactor out `_tofile`/`_fromfile` from `DirectoryStore`

jakirkham:
To make it easier for users to implement their own methods for reading and writing from disk ( like memory-mapping https://github.com/zarr-developers/zarr-python/pull/377 ), refactor out `_tofile` and `_fromfile` methods that are called to handle the reading and writing of the value. Should avoid duplicating a lot of other logic when trying to extend `DirectoryStore`s for these cases.

TODO:

### DIFF
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -779,7 +779,7 @@ class DirectoryStore(MutableMapping):
         file writing logic.
         """
         with open(fn, mode='wb') as f:
-            return f.write(a)
+            f.write(a)
 
     def __getitem__(self, key):
         key = self._normalize_key(key)
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -747,10 +747,37 @@ class DirectoryStore(MutableMapping):
         return key.lower() if self.normalize_keys else key
 
     def _fromfile(self, fn):
+        """ Read data from a file
+
+        Parameters
+        ----------
+        fn: str
+            Filepath to open and read from.
+
+        Notes
+        -----
+        Subclasses should overload this method to specify any custom
+        file reading logic.
+        """
         with open(fn, 'rb') as f:
             return f.read()
 
     def _tofile(self, a, fn):
+        """ Write data to a file
+
+        Parameters
+        ----------
+        a: array-like
+            Data to write into the file.
+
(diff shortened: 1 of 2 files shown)


zarr/storage.py
```diff
@@ -746,12 +746,19 @@ def __init__(self, path, normalize_keys=False):
     def _normalize_key(self, key):
         return key.lower() if self.normalize_keys else key
 
+    def _fromfile(self, fn):
+        with open(filepath, 'rb') as f:
```
jrbourbeau:
```suggestion
        with open(fn, 'rb') as f:
```

jakirkham: Good catch! Thanks 😄

jakirkham:
The idea would be subclasses could override `_tofile` and `_fromfile`. So they could change modes or even the way the file is written/read altogether.

zarr/storage.py
```diff
@@ -746,12 +746,19 @@ def __init__(self, path, normalize_keys=False):
     def _normalize_key(self, key):
         return key.lower() if self.normalize_keys else key
 
+    def _fromfile(self, fn):
```
jrbourbeau: It may be worth adding a docstring that mentions some motivation for why we're pulling this functionality into it's own method. Maybe something like:

```python
""" Read data from a file

Parameters
----------
fn: str
    Filepath to open and read from.

Notes
-----
Subclasses should overload this method to specify any custom
file reading logic.
"""
```

zarr/storage.py
```diff
@@ -746,12 +746,19 @@ def __init__(self, path, normalize_keys=False):
...
+        with open(fn, 'rb') as f:
+            return f.read()
+
+    def _tofile(self, a, fn):
```
jrbourbeau: Similar comment about adding a docstring with `Notes` section

jakirkham:
Thanks for the feedback. Have added two docstrings based on what you mentioned. Please let me know if there is anything else. 🙂

jakirkham:
Planning on merging EOD tomorrow if no comments.

jakirkham:
Thanks James! 😄

PR #506, opened 2019-11-11
Use `ensure_ndarray` in a few more places

jakirkham:
As we can coerce any object that is array-like to an `ndarray`, there should be no need for these explicit `ndarray` checks. So eliminate them by using `ensure_ndarray` to get an `ndarray` object instead.

TODO:

### DIFF
--- a/zarr/core.py
+++ b/zarr/core.py
@@ -1769,7 +1769,7 @@ class Array(object):
                 chunk = f.encode(chunk)
 
         # check object encoding
-        if isinstance(chunk, np.ndarray) and chunk.dtype == object:
+        if ensure_ndarray(chunk).dtype == object:
             raise RuntimeError('cannot write object array without object codec')
 
         # compress
--- a/zarr/core.py
+++ b/zarr/core.py
@@ -1570,8 +1570,13 @@ class Array(object):
 
         """
 
-        out = ensure_ndarray(out)
         assert len(chunk_coords) == len(self._cdata_shape)
+
+        out_is_ndarray = True
+        try:
+            out = ensure_ndarray(out)
+        except TypeError:
+            out_is_ndarray = False
 
         # obtain key for chunk
         ckey = self._chunk_key(chunk_coords)
@@ -1591,7 +1596,8 @@ class Array(object):
 
         else:
 
-            if (not fields and
+            if (out_is_ndarray and
+                    not fields and
                     is_contiguous_selection(out_selection) and
                     is_total_slice(chunk_selection, self._chunks) and
                     not self._filters and
--- a/zarr/core.py
+++ b/zarr/core.py
@@ -1570,6 +1570,7 @@ class Array(object):
 
         """
 
+        out = ensure_ndarray(out)
         assert len(chunk_coords) == len(self._cdata_shape)
 
         # obtain key for chunk
@@ -1590,8 +1591,7 @@ class Array(object):
 
         else:
 
-            if (isinstance(out, np.ndarray) and
-                    not fields and
+            if (not fields and
                     is_contiguous_selection(out_selection) and
                     is_total_slice(chunk_selection, self._chunks) and
                     not self._filters and
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -40,6 +40,9 @@ Upcoming Release
 * Use ``math.ceil`` for scalars.
   By :user:`John Kirkham <jakirkham>`; :issue:`500`.
 
+* Use ``ensure_ndarray`` in a few more places.
+  By :user:`John Kirkham <jakirkham>`; :issue:`506`.
+
 .. _release_2.3.2:
 
 2.3.2


jakirkham:
@jrbourbeau, if you are around, this could use a review 🙂

zarr/core.py
```diff
@@ -1572,6 +1572,12 @@ def _chunk_getitem(self, chunk_coords, chunk_selection, out, out_selection,
 
         assert len(chunk_coords) == len(self._cdata_shape)
 
+        out_is_ndarray = True
+        try:
+            out = ensure_ndarray(out)
```
jakirkham: Note that this is already checked by `test_get_selection_out`. Without this `try`/`except`, we get the following `TypeError` because `out` is a Zarr `Array`, which cannot be coerced to an `ndarray`. This is ok and intentional. So we just carry on without coercing that case and note that we do not have an `ndarray` when checking later.

<details>

```python
_________________________ test_get_selection_out ____________________________

    def test_get_selection_out():
    
        # basic selections
        a = np.arange(1050)
        z = zarr.create(shape=1050, chunks=100, dtype=a.dtype)
        z[:] = a
        selections = [
            slice(50, 150),
            slice(0, 1050),
            slice(1, 2),
        ]
        for selection in selections:
            expect = a[selection]
            out = zarr.create(shape=expect.shape, chunks=10, dtype=expect.dtype, fill_value=0)
>           z.get_basic_selection(selection, out=out)

zarr/tests/test_indexing.py:1036: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
zarr/core.py:698: in get_basic_selection
    fields=fields)
zarr/core.py:740: in _get_basic_selection_nd
zarr/core.py:1028: in _get_selection
    drop_axes=indexer.drop_axes, fields=fields)
zarr/core.py:1573: in _chunk_getitem
    out = ensure_ndarray(out)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

buf = <zarr.core.Array (100,) int64>

    def ensure_ndarray(buf):
        """Convenience function to coerce `buf` to a numpy array, if it is not already a
        numpy array.
    
        Parameters
        ----------
        buf : array-like or bytes-like
            A numpy array or any object exporting a buffer interface.
    
        Returns
        -------
        arr : ndarray
            A numpy array, sharing memory with `buf`.
    
        Notes
        -----
        This function will not create a copy under any circumstances, it is guaranteed to
        return a view on memory exported by `buf`.
    
        """
    
        if isinstance(buf, np.ndarray):
            # already a numpy array
            arr = buf
    
        elif isinstance(buf, array.array) and buf.typecode in 'cu':
            # Guard condition, do not support array.array with unicode type, this is
            # problematic because numpy does not support it on all platforms. Also do not
            # support char as it was removed in Python 3.
            raise TypeError('array.array with char or unicode type is not supported')
    
        else:
    
            # N.B., first take a memoryview to make sure that we subsequently create a
            # numpy array from a memory buffer with no copy
    
            if PY2:  # pragma: py3 no cover
                try:
                    mem = memoryview(buf)
                except TypeError:
                    # on PY2 also check if object exports old-style buffer interface
                    mem = np.getbuffer(buf)
    
            else:  # pragma: py2 no cover
>               mem = memoryview(buf)
E               TypeError: memoryview: a bytes-like object is required, not 'Array'

.tox/py36/lib/python3.6/site-packages/numcodecs/compat.py:74: TypeError
```
</details>

ref: https://travis-ci.org/zarr-developers/zarr-python/jobs/610533967#L2316

jakirkham:
Planning on merging EOD tomorrow if no comments.

zarr/core.py
```diff
@@ -1769,7 +1775,7 @@ def _encode_chunk(self, chunk):
                 chunk = f.encode(chunk)
 
         # check object encoding
-        if isinstance(chunk, np.ndarray) and chunk.dtype == object:
+        if ensure_ndarray(chunk).dtype == object:
```
jrbourbeau: Is it possible that some non-`ndarray`s would previously skip over this `if` block but now fail if `ensure_ndarray` raises an error? For example:

```python
In [28]: from numcodecs.compat import ensure_ndarray

In [29]: import array

In [30]: chunk = array.array('u', 'hello \u2641')

In [31]: if isinstance(chunk, np.ndarray) and chunk.dtype == object:
    ...:     pass
    ...:

In [32]: if ensure_ndarray(chunk).dtype == object:
    ...:     pass
    ...:
<ipython-input-32-e1e15cbf81c2> in <module>
----> 1 if ensure_ndarray(chunk).dtype == object:
      2     pass
      3

~/miniconda/envs/zarr-python-dev/lib/python3.7/site-packages/numcodecs/compat.py in ensure_ndarray(buf)
     63         # problematic because numpy does not support it on all platforms. Also do not
     64         # support char as it was removed in Python 3.
---> 65         raise TypeError('array.array with char or unicode type is not supported')
     66
     67     else:

TypeError: array.array with char or unicode type is not supported
```

To be clear, I'm not sure how likely this is (or if it's even possible) to come up in practice. Do you have a sense for this?

jakirkham: That exception would be expected as we have decided not to work with Python builtin arrays that use character or unicode types. There is some more detailed discussion in [this thread]( https://github.com/zarr-developers/numcodecs/pull/128#discussion_r236476705 ).

jakirkham: Basically the only exception that we need to worry about is [the one coming from `memoryview`]( https://github.com/zarr-developers/numcodecs/blob/8de24ba51492c08bfe6898e209a0a05727f928bc/numcodecs/compat.py#L68 ), which will be a `TypeError` if the object doesn't support [the buffer protocol]( https://docs.python.org/3/c-api/buffer.html ). After that point we have a `memoryview`. So we know the rest of the code will work (unless NumPy gets a bug ;).

jakirkham: For context this function is trying to turn some data into bytes that can be serialized (say to a file on disk).

The filters before this step should be returning something that either is an `ndarray` or could be coerced to one. This is required by the compressors that follow and the storage layer afterwards. So if `ensure_ndarray` fails here, then we have invalid data and raising an exception to the user would be appropriate.

jakirkham:
Thanks @jrbourbeau for the review and fix! 😄

Issue #509, opened 2019-11-14
Feature request: context manager methods for Group


clbarnes:
A few libraries out there try to abstract over a number of different hierarchical data store groups. Some such stores (e.g. h5py's File) are best opened using a context manager (`with` statement). Some libraries (z5py, elf's knossos implementation, h5py-like/ pyn5) aim for h5py compatibility as a design goal; zarr does not, in the main, but for basic access of Groups and Arrays it can be convenient to treat them that way.

The main impediment to doing this using an abstraction layer like `elf.io` is that if you try to use a non-context-manager object like a context manager, you get attribute errors (see my workaround here: https://github.com/constantinpape/elf/blob/master/elf/io/extensions.py#L83 ).

It doesn't do any harm to throw a

```python
def __enter__(self):
    return self

def __exit__(self, *args):
    pass
```

on (could even slot a class into the inheritance hierarchy underneath MutableMapping which just adds these trivial methods, and have Stores and Groups inherit from that instead).

This doesn't actually benefit zarr-python as a library, though, and is potentially extra code noise; it's just convenient for downstream libraries trying to abstract over zarr-python as one of several backends.

alimanfoo:
Some store classes currently do implement the context manager protocol, e.g., [ZipStore](https://github.com/zarr-developers/zarr-python/blob/58b1786fd03414d1c920cf1e19956a34bf5cfc90/zarr/storage.py#L1241), but others don't. Personally I'd be happy for all store classes to implement context manager protocol, doing this via a `Store` ABC that inherits from MutableMapping would be a nice way to go IMO.

It would probably also be fine for the `Group` class to implement the context manager protocol, calling `store.close()` on exit (which may be a no-op depending on the store).

So, e.g., add:

```python

class Store(MutableMapping):

    def close(self):
        pass

    def __enter__(self):
        return self

    def __exit__(self, *args):
        self.close() 
```

...and modify `Group` to add:

```python
    def __enter__(self):
        return self

    def __exit__(self, *args):
        self.store.close() 
```

clbarnes:
I definitely think a Store superclass would be a good idea, as that could also implement the slow fallback behaviour for listdir and rmdir. Using core and 3rd-party Stores as interchangeably as possible would be great.

I also thought about the Group calling its store's close, but are there any potential cases where a store might want to stay open and have multiple groups dipped in and out of? Probably not, given that Groups currently don't need the context manager anyway.

alimanfoo:
[...]

PR #510, opened 2019-11-14
Group as context manager

clbarnes:
For compatibility with other group-like things (e.g. `h5py.File`),
`Groups` now have enter/exit methods for use as a context manager.
If the underlying store has a `close` method, that will be called on
exit; otherwise this does nothing.

See issue #509.

It would have been nice to extend the context manager stuff to a Store superclass (an ABC which also inherits from MutableMapping), but as described in the issue, that makes guarantees that existing code may already break (e.g. using a raw `dict` as a store).

TODO:

### DIFF
--- a/zarr/tests/test_hierarchy.py
+++ b/zarr/tests/test_hierarchy.py
@@ -3,6 +3,7 @@ import atexit
 import os
 import pickle
 import shutil
+import sys
 import tempfile
 import textwrap
 import unittest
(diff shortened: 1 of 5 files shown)


clbarnes:
`tox -e docs` fails due to numcodecs not being installed. Is this a known issue?

zarr/tests/test_hierarchy.py
```diff
@@ -912,6 +919,17 @@ def create_store():
...
+    def test_context_manager(self):
+
+        with self.create_group() as g:
+            store = g.store
+            d = g.create_dataset('foo/bar', shape=100, chunks=10)
+            d[:] = np.arange(100)
+
+        error = ValueError if sys.version_info >= (3, 6) else RuntimeError
+        with pytest.raises(error):
+            store.zf.extractall()
```
alimanfoo: Just checking, is this testing that the underlying ZipStore has been closed and hence the zip file is closed? If so, might be useful to add a comment that this is what this is doing.

alimanfoo:
> `tox -e docs` fails due to numcodecs not being installed. Is this a known issue?

Hm, no, not sure what's happening there.

zarr/tests/test_hierarchy.py
clbarnes: Exactly. I'll do that.

Issue #525, opened 2019-12-11
Azure blob storage prepends a '/' to the blob names when the prefix is None


shikharsg:
Azure blob storage prepends a '/' to the blob names, even when the prefix is None. Code sample:

#### Minimal, reproducible code sample
This code needs to connect to Azure Blob (the sample below connects to a local emulator)
```python
>>> import zarr
>>> store = zarr.ABSStore('test', prefix=None, blob_service_kwargs={'is_emulated':True})
>>> g = zarr.open_group(store)
>>> ## list all blobs in the container
>>> print([blob.name for blob in store.client.list_blobs('test')])
['/.zgroup']
```
which shows up in the storage explorer like this:
![image](https://user-images.githubusercontent.com/8643775/70611383-03a45a80-1bfd-11ea-92ea-cc6c94330ca0.png)

There should ideally be no '/' before the `.zgroup` and it should look like this:
```python
>>> import zarr
>>> store = zarr.ABSStore('test', prefix=None, blob_service_kwargs={'is_emulated':True})
>>> g = zarr.open_group(store)
>>> ## list all blobs in the container
>>> print([blob.name for blob in store.client.list_blobs('test')])
['.zgroup']
```
![image](https://user-images.githubusercontent.com/8643775/70611581-541bb800-1bfd-11ea-97dd-c3eb9f62440f.png)

Will submit a PR for this in a bit

tjcrone:
It's not clear to me why you would want .zgroup to be directly under a blob container. The blob container is not a Zarr group.

shikharsg:
Fixed in PR #526

PR #526, opened 2019-12-11
Fixed '/' prepend bug in ABSStore

shikharsg:
Fixes '/' prepend bug in `ABSStore`, issue: https://github.com/zarr-developers/zarr-python/issues/525

TODO:

### DIFF
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -1964,7 +1964,7 @@ class ABSStore(MutableMapping):
     In order to use this store, you must install the Microsoft Azure Storage SDK for Python.
     """
 
-    def __init__(self, container, prefix=None, account_name=None, account_key=None,
+    def __init__(self, container, prefix='', account_name=None, account_key=None,
                  blob_service_kwargs=None):
         from azure.storage.blob import BlockBlobService
         self.container = container
@@ -1994,8 +1994,7 @@ class ABSStore(MutableMapping):
         if self.prefix == '':
             return normalize_storage_path(path)
         else:
-            return '/'.join([normalize_storage_path(self.prefix),
-                             normalize_storage_path(path)])
+            return '/'.join([self.prefix, normalize_storage_path(path)])
 
     @staticmethod
     def _strip_prefix_from_path(path, prefix):
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -1964,7 +1964,7 @@ class ABSStore(MutableMapping):
     In order to use this store, you must install the Microsoft Azure Storage SDK for Python.
     """
 
-    def __init__(self, container, prefix, account_name=None, account_key=None,
+    def __init__(self, container, prefix=None, account_name=None, account_key=None,
                  blob_service_kwargs=None):
         from azure.storage.blob import BlockBlobService
         self.container = container
@@ -1990,21 +1990,26 @@ class ABSStore(MutableMapping):
         self.client = BlockBlobService(self.account_name, self.account_key,
                                        **self.blob_service_kwargs)
 
-    @staticmethod
-    def _append_path_to_prefix(path, prefix):
-        return '/'.join([normalize_storage_path(prefix),
-                         normalize_storage_path(path)])
+    def _append_path_to_prefix(self, path):
+        if self.prefix == '':
+            return normalize_storage_path(path)
+        else:
+            return '/'.join([normalize_storage_path(self.prefix),
+                             normalize_storage_path(path)])
 
     @staticmethod
     def _strip_prefix_from_path(path, prefix):
         # normalized things will not have any leading or trailing slashes
         path_norm = normalize_storage_path(path)
         prefix_norm = normalize_storage_path(prefix)
-        return path_norm[(len(prefix_norm)+1):]
+        if prefix:
+            return path_norm[(len(prefix_norm)+1):]
+        else:
+            return path_norm
 
     def __getitem__(self, key):
         from azure.common import AzureMissingResourceHttpError
-        blob_name = '/'.join([self.prefix, key])
+        blob_name = self._append_path_to_prefix(key)
         try:
             blob = self.client.get_blob_to_bytes(self.container, blob_name)
             return blob.content
@@ -2013,13 +2018,13 @@ class ABSStore(MutableMapping):
 
     def __setitem__(self, key, value):
         value = ensure_bytes(value)
-        blob_name = '/'.join([self.prefix, key])
+        blob_name = self._append_path_to_prefix(key)
         self.client.create_blob_from_bytes(self.container, blob_name, value)
 
     def __delitem__(self, key):
         from azure.common import AzureMissingResourceHttpError
         try:
-            self.client.delete_blob(self.container, '/'.join([self.prefix, key]))
+            self.client.delete_blob(self.container, self._append_path_to_prefix(key))
         except AzureMissingResourceHttpError:
             raise KeyError('Blob %s not found' % key)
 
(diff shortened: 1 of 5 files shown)


zarr/storage.py
```diff
@@ -1990,21 +1990,26 @@ def __setstate__(self, state):
         self.client = BlockBlobService(self.account_name, self.account_key,
                                        **self.blob_service_kwargs)
 
-    @staticmethod
-    def _append_path_to_prefix(path, prefix):
-        return '/'.join([normalize_storage_path(prefix),
-                         normalize_storage_path(path)])
+    def _append_path_to_prefix(self, path):
```
alimanfoo: Just wondering, what happens here if `self.prefix` is `None` (default)?

zarr/tests/test_storage.py
```diff
@@ -1473,16 +1473,40 @@ def test_format_compatibility():
...
         blob_client.delete_container('test')
         blob_client.create_container('test')
-        store = ABSStore(container='test', prefix='zarrtesting/', account_name='foo',
+        store = ABSStore(container='test', prefix=prefix, account_name='foo',
                          account_key='bar', blob_service_kwargs={'is_emulated': True})
         store.rmdir()
         return store
 
+    def test_iterators_with_prefix(self):
+        store = self.create_store(prefix='test_prefix')
```
alimanfoo: Not sure this makes sense as haven't looked super closely at the implementation, but maybe worth testing multiple prefixes here, with and without leading or trailing slashes, just to confirm everything behaves as you expect? E.g.:

```
for prefix in 'test_prefix', '/test_prefix', 'test_prefix/', 'test/prefix', '', None:
    store = self.create_store(prefix=prefix)
    # ... tests
```

zarr/storage.py
shikharsg: During instantiation in the `__init__` function  any `prefix` supplied is passed to the `zarr.util.normalize_storage_path` function before storing it in `self.prefix`, so if prefix is `None` it will be stored as `''` in `self.prefix`

zarr/tests/test_storage.py
shikharsg: Since `prefix` is passed to `normalize_storage_path` before storing in `self.prefix` any leading or trailing slashes will be removed.  

Putting in some print statements in the tests:
```python
    def test_iterators_with_prefix(self):
        for prefix in ['test_prefix', '/test_prefix', 'test_prefix/', 'test/prefix', '', None]:
            print(prefix)
            store = self.create_store(prefix=prefix)

            # test iterator methods on empty store
            assert 0 == len(store)
            assert set() == set(store)
            assert set() == set(store.keys())
            assert set() == set(store.values())
            assert set() == set(store.items())

            # setup some values
            store['a'] = b'aaa'
            store['b'] = b'bbb'
            store['c/d'] = b'ddd'
            store['c/e/f'] = b'fff'
            print([blob for blob in store.client.list_blob_names('test')])

            # test iterators on store with data
            assert 4 == len(store)
            assert {'a', 'b', 'c/d', 'c/e/f'} == set(store)
            assert {'a', 'b', 'c/d', 'c/e/f'} == set(store.keys())
            assert {b'aaa', b'bbb', b'ddd', b'fff'} == set(store.values())
            assert ({('a', b'aaa'), ('b', b'bbb'), ('c/d', b'ddd'), ('c/e/f', b'fff')} ==
                    set(store.items()))
 
```
Output is this:
```
test_prefix
['test_prefix/a', 'test_prefix/b', 'test_prefix/c/d', 'test_prefix/c/e/f']
/test_prefix
['test_prefix/a', 'test_prefix/b', 'test_prefix/c/d', 'test_prefix/c/e/f']
test_prefix/
['test_prefix/a', 'test_prefix/b', 'test_prefix/c/d', 'test_prefix/c/e/f']
test/prefix
['test/prefix/a', 'test/prefix/b', 'test/prefix/c/d', 'test/prefix/c/e/f']

['a', 'b', 'c/d', 'c/e/f']
None
['a', 'b', 'c/d', 'c/e/f']
```

Which is what you would expect

tjcrone:
Sorry if I am a little confused by this. But are you trying to put a Zarr group inside of a blob storage account not inside a blob container?

shikharsg:
@tjcrone sorry if that wasn't clear. All metadata (.zgroup, .zarray...) and data(chunks) will be blobs inside a particular container. Nothing is at the storage account level.

The issue was that ABSStore prepends an extra '/' to all blob names. So if you opened a group in the root of the container the blobs would be named as ['/.zgroup', '/foo/.zgroup'], instead of ['.zgroup', 'foo/.zgroup']. The former is also awkward to look at in the storage explorer.

tjcrone:
I don't think we ever contemplated the idea of turning a blob container into a Zarr group, which is what you are doing here by passing prefix=None. I guess we always thought that Zarr groups (sometimes referred to as Zarr "files") would have names, and live inside blob containers, which seems like the right thing to do. Maybe it is fine to do this, but at the moment I remain skeptical. @alimanfoo, do you have an opinion here? Would you recommend treating a blob container like a Zarr group? More info on Azure blob containers here: https://docs.microsoft.com/en-us/azure/storage/blobs/storage-blobs-introduction.

shikharsg:
Ah i understand now. Yes, this PR will allow a zarr container to become a group(also array for that matter).

Before this PR we would probably have a structure like this:
`['group1/.zgroup', 'group1/group2/.zgroup', 'group1/array1/.zarray']`
where everything is inside of `group1`. I'm not sure how that would be "conceptually different" from making the container itself a group where we have in addition to the above three also a `.zgroup` in the container: 
`['.zgroup', 'group1/.zgroup', 'group1/group2/.zgroup', 'group1/array1/.zarray']`

shikharsg:
@martindurant does this functionality exist in `s3fs.S3Map`, i.e. are we allowed to treat an S3 bucket as a zarr group?

zarr/storage.py
alimanfoo: Ah cool, thanks. In which case maybe the default value for the prefix parameter could the empty string rather than None, might be a little more obvious. Also if prefix is normalised in the constructor then probably could remove calls to normalize self.prefix in the methods.

zarr/tests/test_storage.py
alimanfoo: Great, thanks.

tjcrone:
In your example group1 is the name of the Zarr "file", which is typically given a .zarr extension, but that's not required, and it is inside of a blob container. I think the idea of treating a blob container as a Zarr "file" might be a bad idea, primarily because a container is not a blob. They are created and destroyed differently. I think if you tried to create a Zarr file that was a container rather than a blob inside a container, it would not work. Have you tried doing this?

alimanfoo:
FWIW I don't see any problem in putting a zarr group (or array) at the root of a bucket, if that's what you want to do, i.e., you don't want to use the bucket for anything else. Should work fine on S3 and GCS.

martindurant:
@shikharsg , I think so? Should try!

tjcrone:
I think we should maintain as much parity with the S3/GCS functionality as possible. So if it is possible to treat an S3 bucket as a Zarr group or array, and it has the functionality of creating a bucket just like it would create a file when the Zarr object needs creating, I'm fine with this.

zarr/storage.py
```diff
@@ -1994,8 +1994,7 @@ def _append_path_to_prefix(self, path):
         if self.prefix == '':
             return normalize_storage_path(path)
         else:
-            return '/'.join([normalize_storage_path(self.prefix),
-                             normalize_storage_path(path)])
+            return '/'.join([self.prefix, normalize_storage_path(path)])
```
tjcrone: So what happens here if prefix='/'?

shikharsg: Because the `prefix` is normalized in the constructor, if you pass in `'/'` for the `prefix` it will still be stored as `''` in `self.prefix`

tjcrone: Okay great. Thanks.

tjcrone:
Have we confirmed that the S3 and GCS will allow placing a Zarr group or array at the root of a bucket? If they do, we should merge this. I really think that treating an entire bucket or container as a Zarr file is an odd thing to do, perhaps because it removes a layer of abstraction that I think is useful. But maybe it's me that's odd. I will obviously defer to @alimanfoo, on this.

tjcrone:
I guess it is sortof like putting the contents of a file at the root of a file system (not  as a file in /, but actually at the root, because you don't plan to use the file system anyway. I don't think most file systems will let you do this, because the root of the file system is a container for storing files and other containers, so doing this would be weird. But honestly if y'all think that storing Zarr collections at containers not in them is a good idea, then let's try it and see how it goes.

shikharsg:
With S3 it works:
```python
>>> s3 = s3fs.S3FileSystem(secret='wJalrXUtnFEMI/K7MDENG/bPxRfiCYEXAMPLEKEY', key='AKIAIOSFODNN7EXAMPLE', anon=False, client_kwargs={'endpoint_url':'http://localhost:9000'})
>>> store = s3fs.S3Map(root='test', s3=s3, check=False)
>>> root = zarr.group(store=store)
```

image


Couldn't try GCS as I don't have an account and there doesn't seem to be an emulator.

shikharsg:
Thanks for the helpful comments @alimanfoo @tjcrone @martindurant . If there are no further reviews/objections can I go ahead and merge?

PR #470, opened 2019-08-13
Drop Python 2 Support

jhamman:
Drops support for Python 2. Closes #393.

TODO:
* [ ] Add unit tests and/or doctests in docstrings
* [ ] Add docstrings and API docs for any new/modified user-facing classes and functions
* [ ] New/modified features documented in docs/tutorial.rst
* [x] Changes documented in docs/release.rst
* [x] Docs build locally (e.g., run ``tox -e docs``)
* [x] AppVeyor and Travis CI passes
* [x] Test coverage is 100% (Coveralls passes)

PR #542, opened 2020-03-03
Maximize coverage

jakirkham:
These were uncovered lines in [this Coveralls report]( https://coveralls.io/builds/29088639 ). As we already dropped Python 2 in PR ( https://github.com/zarr-developers/zarr-python/pull/470 ), go ahead and drop this no longer need Python 2 code.

Also smooth over differences in `unravel_index`'s signature across NumPy versions.

### DIFF
--- a/zarr/indexing.py
+++ b/zarr/indexing.py
@@ -691,11 +691,7 @@ class CoordinateIndexer(object):
         self.chunk_rixs = np.nonzero(self.chunk_nitems)[0]
 
         # unravel chunk indices
-        if tuple(map(int, np.__version__.split('.')[:2])) < (1, 16):
-            self.chunk_mixs = np.unravel_index(self.chunk_rixs, dims=array._cdata_shape)
-        else:
-            # deal with change dims->shape in arguments as of numpy 1.16
-            self.chunk_mixs = np.unravel_index(self.chunk_rixs, shape=array._cdata_shape)
+        self.chunk_mixs = np.unravel_index(self.chunk_rixs, array._cdata_shape)
 
     def __iter__(self):
 
--- a/zarr/tests/test_meta.py
+++ b/zarr/tests/test_meta.py
@@ -13,8 +13,6 @@ from zarr.meta import (ZARR_FORMAT, decode_array_metadata, decode_dtype,
 
 
 def assert_json_equal(expect, actual):
-    if isinstance(expect, bytes):  # pragma: py3 no cover
-        expect = str(expect, 'ascii')
     if isinstance(actual, bytes):
         actual = str(actual, 'ascii')
     ej = json.loads(expect)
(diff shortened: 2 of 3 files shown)


zarr/indexing.py
```diff
@@ -691,11 +691,7 @@ def __init__(self, selection, array):
         self.chunk_rixs = np.nonzero(self.chunk_nitems)[0]
 
         # unravel chunk indices
-        if tuple(map(int, np.__version__.split('.')[:2])) < (1, 16):
-            self.chunk_mixs = np.unravel_index(self.chunk_rixs, dims=array._cdata_shape)
-        else:
-            # deal with change dims->shape in arguments as of numpy 1.16
-            self.chunk_mixs = np.unravel_index(self.chunk_rixs, shape=array._cdata_shape)
+        self.chunk_mixs = np.unravel_index(self.chunk_rixs, array._cdata_shape)
```
jakirkham: Do you know if the keyword argument was needed for some reason @alimanfoo? Noticed this change came from [here]( https://github.com/zarr-developers/zarr-python/pull/468/files#diff-927f13103811bc87516c2e8e88776ce5 ). Hence the question 🙂

mzjp2: In general, I think using kwargs is better when it's defined as a kwarg rather than relying on the order of positional args. The previous block looked like it was here to deal with numpy <1.16 and numpy >1.16 but in our `setup.py` we specify `numpy>=1.17`, so I think it's safe to drop handling old `numpy` versions and use `shape=` but not fussed too much either way.

jrbourbeau: IIUC our `setup.py` has NumPy 1.7 as the minimum supported version

https://github.com/zarr-developers/zarr-python/blob/3ae943914c5241b7e2e4c2ee730a23344d5299be/setup.py#L12-L17

mzjp2: Yep. :)

jrbourbeau: So I guess we do need to consider older versions of NumPy where the signature keywords may have been different

mzjp2: Ah, I kept reading 1.7 as 1.17... 🤦‍♂

jakirkham: Seems to be largely unchanged over time. 🙂 Here are [the docs from NumPy 1.7]( https://docs.scipy.org/doc/numpy-1.7.0/reference/generated/numpy.unravel_index.html ).

mzjp2: Cool, cool. Happy for this to be merged on my side :)

jakirkham:
Going to go ahead and merge. Guessing Alistair is busy with other things and the risk here is low. Though if that's not true, please raise an issue and we can work to address it. Thanks all! 😄

Issue #410, opened 2019-02-25
Implement N5Store as a transformation layer over other stores


alimanfoo:
The current work on an N5Store in #309 implements a storage class that does two things. First, it implements a transformation from the keys and values used in the zarr protocol into keys and values used in the N5 protocol. It then also implements storage of keys and values as files on a file system.

It would be possible to separate these concerns, such that the N5Store just implemented the protocol transformation, and then delegated the actual storage or retrieval of N5 keys and values from an inner store. This would mean that the N5Store class could be used with file systems or cloud stores or any form of key/value storage.

E.g., what is currently now:

```
store = zarr.N5Store('/path/to/store.n5')
```

...could become a shorthand for:

```
store = zarr.N5Store(zarr.DirectoryStore('/path/to/store.n5'))
```

...and other inner stores could be used, e.g.:

```
from gcsfs import GCSMap
store = zarr.N5Store(GCSMap('my-bucket', ...))
```

Internally, to make this possible, the N5Store class would replace all the logic for opening, reading or writing files with calls to `inner.__getitem__(key)` and `inner.__setitem__(key, value)` where `inner` is the inner store wrapped by the N5Store class.

jakirkham:
Thanks for writing this up.

Would add that this is also true of `NestedDirectoryStore`, which `N5Store` currently relies on. If we change `N5Store` into a transformational layer, we may want to do something similar with `NestedDirectoryStore`.

alimanfoo:
Good point, yes using "/" instead of "." as the separator within chunk keys
could also be implemented as a protocol transformation over any other
store.

Would probably need a new class with an appropriate name for that, e.g.,
NestedChunkStore. So then NestedDirectoryStore(path) would become a
shorthand for NestedChunkStore(DirectoryStore(path)).

jakirkham:
As far as naming these classes that are not quite stores, what should we call them?

1. `...Store`
2. `...Layer`
3. `...Transformer`
4. ?

alimanfoo:
> As far as naming these classes that are not quite stores, what should we call them?
> 
> 1. `...Store`
> 2. `...Layer`
> 3. `...Transformer`
> 4. ?

Good question. I guess `...StoreTransformer` would be the most descriptive, but also the most verbose. I'd be happy with just `...Store` too. 

xref https://github.com/zarr-developers/zarr-specs/issues/8

PR #546, opened 2020-03-17
Add FSStore

### LINKED ISSUE #540: N5Store support of cloud buckets
The N5Store works great for storing local files.  What are the technical challenges in moving to support something like GCS for storage?   Other pieces of the zarr library seamlessly work if I move to gs:// prefixes (with the right packages installed), but N5Store ends up writing to the local filesystem when using that prefix.

Is it a matter of going through n5.py and storage.py and changing uses of the os library to uses of fsspec?  Or are there operations that fundamentally won't work on cloud objects?

martindurant:
Fixes #540 
Ref https://github.com/zarr-developers/zarr-python/pull/373#issuecomment-592722584 (@rabernat )

Introduces a short Store implementation for generic fsspec url+options. Allows both lowercasing of keys and choice between '.' and '/'-based ("nested") keys.

For testing, have hijacked the TestNestedDirectoryStore tests just as an example - this is not how things will remain.

### DIFF
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -949,7 +949,7 @@ def atexit_rmglob(path,
 class FSStore(MutableMapping):
 
     def __init__(self, url, normalize_keys=True, key_separator='.',
-                 mode='r', **storage_options):
+                 mode='w', **storage_options):
         import fsspec
         self.path = url
         self.normalize_keys = normalize_keys
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -21,7 +21,6 @@ import glob
 import multiprocessing
 import operator
 import os
-import json
 import re
 import shutil
 import sys
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -21,6 +21,7 @@ import glob
 import multiprocessing
 import operator
 import os
+import json
 import re
 import shutil
 import sys
@@ -949,7 +950,7 @@ def atexit_rmglob(path,
 class FSStore(MutableMapping):
 
     def __init__(self, url, normalize_keys=True, key_separator='.',
-                 mode='w', **storage_options):
+                 mode='w', consolidated=False, metadata_key='.zmetadata', **storage_options):
         import fsspec
         self.path = url
         self.normalize_keys = normalize_keys
@@ -957,6 +958,24 @@ class FSStore(MutableMapping):
         self.map = fsspec.get_mapper(url, **storage_options)
         self.fs = self.map.fs  # for direct operations
         self.mode = mode
+        # TODO: should warn if consolidated and write mode?
+        if self.fs.exists(url) and not self.fs.isdir(url):
+            err_fspath_exists_notdir(url)
+        self.consolidated = consolidated
+        self.metadata_key = metadata_key
+        if consolidated:
+            self.meta = json.loads(self.map.get(metadata_key, b"{}").decode())
+            if mode == 'r' or 'zarr_consolidated_format' in self.meta:
+                consolidated_format = self.meta.get('zarr_consolidated_format', None)
+                if consolidated_format != 1:
+                    raise MetadataError('unsupported zarr consolidated metadata format: %s' %
+                                        consolidated_format)
+            else:
+                self.meta['zarr_consolidated_format'] = 1
+
+    @staticmethod
+    def _is_meta(key):
+        return key.split('/')[-1] in [attrs_key, group_meta_key, array_meta_key]
 
     def _normalize_key(self, key):
         key = normalize_storage_path(key).lstrip('/')
@@ -966,12 +985,17 @@ class FSStore(MutableMapping):
         return key.lower() if self.normalize_keys else key
 
     def __getitem__(self, key):
+        if self.consolidated and self._is_meta(key):
+            return self.meta[key]
         key = self._normalize_key(key)
         return self.map[key]
 
     def __setitem__(self, key, value):
         if self.mode == 'r':
-            raise PermissionError
+            err_read_only()
+        if self.consolidated and self._is_meta(key):
+            self.meta[key] = value.decode()
+            self.map[self.metadata_key] = json.dumps(self.meta).encode()
         key = self._normalize_key(key)
         path = self.dir_path(key)
         value = ensure_contiguous_ndarray(value)
@@ -984,7 +1008,10 @@ class FSStore(MutableMapping):
 
     def __delitem__(self, key):
         if self.mode == 'r':
-            raise PermissionError
+            err_read_only()
+        if self.consolidated and self._is_meta(key):
+            del self.meta[key]
+            self.map[self.metadata_key] = json.dumps(self.meta).encode()
         key = self._normalize_key(key)
         path = self.dir_path(key)
         if self.fs.isdir(path):
@@ -993,6 +1020,8 @@ class FSStore(MutableMapping):
             del self.map[key]
 
     def __contains__(self, key):
+        if self.consolidated and self._is_meta(key):
+            return key in self.meta
         key = self._normalize_key(key)
         return key in self.map
 
@@ -1022,6 +1051,8 @@ class FSStore(MutableMapping):
             return []
 
     def rmdir(self, path=None):
+        if self.mode == 'r':
+            err_read_only()
         store_path = self.dir_path(path)
         if self.fs.isdir(store_path):
             self.fs.rm(store_path, recursive=True)
@@ -1031,6 +1062,10 @@ class FSStore(MutableMapping):
         return self.fs.du(store_path, True, True)
 
     def clear(self):
+        if self.mode == 'r':
+            err_read_only()
+        if self.consolidated:
+            self.meta = {}
         self.map.clear()
 
 
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -946,6 +946,87 @@ def atexit_rmglob(path,
             rmtree(p)
 
 
+class FSStore(MutableMapping):
+
+    def __init__(self, url, normalize_keys=True, key_separator='.',
+                 **storage_options):
+        import fsspec
+        self.path = url
+        self.normalize_keys = normalize_keys
+        self.key_separator = key_separator
+        self.map = fsspec.get_mapper(url, **storage_options)
+        self.fs = self.map.fs  # for direct operations
+
+    def _normalize_key(self, key):
+        key = normalize_storage_path(key)
+        if key:
+            *bits, end = key.split('/')
+            key = '/'.join(bits + [end.replace('.', self.key_separator)])
+        return key.lower() if self.normalize_keys else key
+
+    def __getitem__(self, key):
+        key = self._normalize_key(key)
+        return self.map[key]
+
+    def __setitem__(self, key, value):
+        key = self._normalize_key(key)
+        path = self.dir_path(key)
+        value = ensure_contiguous_ndarray(value)
+        try:
+            if self.fs.isdir(path):
+                self.fs.rm(path, recursive=True)
+            self.map[key] = value
+        except IOError:
+            raise KeyError(key)
+
+    def __delitem__(self, key):
+        key = self._normalize_key(key)
+        path = self.dir_path(key)
+        if self.fs.isdir(path):
+            self.fs.rm(path, recursive=True)
+        else:
+            del self.map[key]
+
+    def __contains__(self, key):
+        key = self._normalize_key(key)
+        return key in self.map
+
+    def __eq__(self, other):
+        return type(self) == type(other) and self.map == other.map
+
+    def keys(self):
+        return iter(self.map)
+
+    def __iter__(self):
+        return self.keys()
+
+    def __len__(self):
+        return len(list(self.keys()))
+
+    def dir_path(self, path=None):
+        store_path = normalize_storage_path(path)
+        return self.map._key_to_str(store_path)
+
+    def listdir(self, path=None):
+        dir_path = self.dir_path(path)
+        try:
+            return sorted(p.rsplit('/', 1)[-1] for p in self.fs.ls(dir_path))
+        except IOError:
+            return []
+
+    def rmdir(self, path=None):
+        store_path = self.dir_path(path)
+        if self.fs.isdir(store_path):
+            self.fs.rm(store_path, recursive=True)
+
+    def getsize(self, path=None):
+        store_path = self.dir_path(path)
+        return self.fs.du(store_path, True, True)
+
+    def clear(self):
+        self.map.clear()
+
+
 class TempStore(DirectoryStore):
     """Directory store using a temporary directory for storage.
 
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -950,7 +950,9 @@ def atexit_rmglob(path,
 class FSStore(MutableMapping):
 
     def __init__(self, url, normalize_keys=True, key_separator='.',
-                 mode='w', consolidated=False, metadata_key='.zmetadata', **storage_options):
+                 mode='w', consolidated=False, metadata_key='.zmetadata',
+                 exceptions=(KeyError, PermissionError, IOError),
+                 **storage_options):
         import fsspec
         self.path = url
         self.normalize_keys = normalize_keys
@@ -958,6 +960,7 @@ class FSStore(MutableMapping):
         self.map = fsspec.get_mapper(url, **storage_options)
         self.fs = self.map.fs  # for direct operations
         self.mode = mode
+        self.exceptions = exceptions
         # TODO: should warn if consolidated and write mode?
         if self.fs.exists(url) and not self.fs.isdir(url):
             err_fspath_exists_notdir(url)
@@ -988,7 +991,10 @@ class FSStore(MutableMapping):
         if self.consolidated and self._is_meta(key):
             return self.meta[key]
         key = self._normalize_key(key)
-        return self.map[key]
+        try:
+            return self.map[key]
+        except self.exceptions  as e:
+            raise KeyError(key) from e
 
     def __setitem__(self, key, value):
         if self.mode == 'r':
@@ -1003,8 +1009,8 @@ class FSStore(MutableMapping):
             if self.fs.isdir(path):
                 self.fs.rm(path, recursive=True)
             self.map[key] = value
-        except IOError:
-            raise KeyError(key)
+        except self.exceptions as e:
+            raise KeyError(key) from e
 
     def __delitem__(self, key):
         if self.mode == 'r':
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -948,9 +948,30 @@ def atexit_rmglob(path,
 
 
 class FSStore(MutableMapping):
+    """Wraps an fsspec.FSMap to give access to arbitrary filesystems
+
+    Requires that ``fsspec`` is installed, as well as any additional
+    requirements for the protocol chosen.
+
+    Parameters
+    ----------
+    url : str
+        The destination to map. Should include protocol and path,
+        like "s3://bucket/root"
+    normalize_keys : bool
+    key_separator : str
+        Character to use when constructing the target path strings
+        for data keys
+    mode : str
+        "w" for writable, "r" for read-only
+    exceptions : list of Exception subclasses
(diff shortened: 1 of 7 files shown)


alimanfoo:
Thanks Martin, great to see this.

rabernat:
Thanks a lot @martindurant for getting this started! Let me know when you're ready for a review or some feedback.

It would be really nice to implement `getsize` and `rename` as well.

martindurant:
`getsize` is already there ( https://github.com/zarr-developers/zarr-python/pull/546/files#diff-31d15042dbeedbf2942ace2ad4b9b2e2R1022 )
Do you have the signature for `rename`?

chrisroat:
@martindurant I was just reloading the underlying issue into my brain's cache this morning, and excited to see this.  What is the overall plan, with regard to how this would interact with things like N5Store and NestedDirectoryStore?  How can I help?

martindurant:
The idea is, that nested and consolidated, at least, would be options to this store, no need for extra classes. I don't know enough about the N5 case to know if that is as simple.

As far as I know, the little code that is here "works", but needs a good deal of testing and doubtless some massage to get totally right. At the moment, the PR just hijacks some existing nested tests, which all then pass locally, and almost all on CI.

chrisroat:
If this is to fix 540, then I would like to see how I could help update N5Store to use this.  The N5Store is a NestedDirectoryStore (also a MutableMapping like FSStore) which lays out the data according the N5 spec.   

I do not think N5Store can just inherit from FSStore, since it will lose the Directory/NestedDirectory functionality.   It's possible that N5Store could take a FSStore in it's c'tor, but that may not be the cleanest way to combine the N5, FS, and NestedDirectory functionality.

Is FSStore a possible replacement for a DirectoryStore?

Regarding the comment above about the `rename` method, [here is the DirectoryStore](https://github.com/zarr-developers/zarr-python/blob/e7d5f39b4488a2d230a7ef5df7e69ef848f6ae78/zarr/storage.py#L887)
implementation.

martindurant:
> I do not think N5Store can just inherit from FSStore, since it will lose the Directory/NestedDirectory functionality

In this POC, FSStore is tested as if it were a nested store, but it can do both, depending on the arguments passed. That's what I meant bu not needing to have separate classes.

>  here is the DirectoryStore implementation.

OK, so simple enough

alimanfoo:
Just to say FWIW I think the best way forward for N5Store is to modify it so it becomes a transformation layer over another store, as suggested [here](https://github.com/zarr-developers/zarr-python/issues/540#issuecomment-593563942) and in #410. That way N5Store can be used with any type of underlying storage. I.e., work on N5Store can be orthogonal to this PR.

chrisroat:
@martindurant I did a quick swap on N5Store to have it inherit from FSStore (and use the FSStore listdir within N5Store.listdir), and it does seem to correctly read/write a single array to GCS.  I didn't test other functionality, like groups or anything related to listdir explicitly.

@alimanfoo OK.  I'll pick up the discussion about the correct interaction of N5Store and FSStore on 540.

martindurant:
Right, there will be problems - so far I only ran the nestedstore tests, and see that they pass. You must of course pass `key_separator='/'` to the init, which for a subclass would be
```
    super().__init__(key_separator='/', other-kwargs)
```

I haven't got very far yet, but the code here should show that it isn't too complicated.

zarr/storage.py
```diff
@@ -946,6 +946,87 @@ def atexit_rmglob(path,
...
+        return len(list(self.keys()))
+
+    def dir_path(self, path=None):
+        store_path = normalize_storage_path(path)
+        return self.map._key_to_str(store_path)
+
+    def listdir(self, path=None):
+        dir_path = self.dir_path(path)
+        try:
+            return sorted(p.rsplit('/', 1)[-1] for p in self.fs.ls(dir_path))
```
joshmoore, line 1013: This line throws an AttributeError for `group.arrays()` when serving a directory created with:

```
import zarr
store = zarr.DirectoryStore('dir.zarr')
group = zarr.group(store=store, overwrite=True)
z = group.zeros("test", shape=(4, 4))
z[:] = 4
```

via `python -m http.server`:

```
In [3]: from zarr.storage import FSStore

In [4]: store = FSStore("http://localhost:8000")

In [5]: group = zarr.group(store=store)

In [6]: group.arrays()
Out[6]: <generator object Group._array_iter at 0x7fd2e0bff258>

In [7]: list(group.arrays())
AttributeError                            Traceback (most recent call last)
<ipython-input-7-b56aa2c546df> in <module>
----> 1 list(group.arrays())

~/opt/zarr/zarr/hierarchy.py in _array_iter(self, keys_only, method, recurse)
    477
    478     def _array_iter(self, keys_only, method, recurse):
--> 479         for key in sorted(listdir(self._store, self._path)):
    480             path = self._key_prefix + key
    481             if contains_array(self._store, path):

~/opt/zarr/zarr/storage.py in listdir(store, path)
    165     if hasattr(store, 'listdir'):
    166         # pass through
--> 167         return store.listdir(path)
    168     else:
    169         # slow version, iterate through all keys

~/opt/zarr/zarr/storage.py in listdir(self, path)
   1011         dir_path = self.dir_path(path)
   1012         try:
-> 1013             return sorted(p.rsplit('/', 1)[-1] for p in self.fs.ls(dir_path))
   1014         except IOError:
   1015             return []

~/opt/zarr/zarr/storage.py in <genexpr>(.0)
   1011         dir_path = self.dir_path(path)
   1012         try:
-> 1013             return sorted(p.rsplit('/', 1)[-1] for p in self.fs.ls(dir_path))
   1014         except IOError:
   1015             return []

AttributeError: 'dict' object has no attribute 'rsplit'
```

while the current direct use of `from fsspec.implementations.http import HTTPFileSystem` shows:

```
In [33]: list(zg.arrays())
Out[33]: [('test', <zarr.core.Array '/test' (4, 4) float64 read-only>)]
```

martindurant: fixed

zarr/storage.py
```diff
@@ -949,16 +949,17 @@ def atexit_rmglob(path,
 class FSStore(MutableMapping):
 
     def __init__(self, url, normalize_keys=True, key_separator='.',
-                 **storage_options):
+                 mode='r', **storage_options):
```
joshmoore, line 952: With this change, the TestNestedDirectoryStore failure becomes:

```
_____________________________________________ TestNestedDirectoryStore.test_chunk_nesting _____________________________________________

self = <zarr.tests.test_storage.TestNestedDirectoryStore testMethod=test_chunk_nesting>

    def test_chunk_nesting(self):
        store = self.create_store()
        # any path where last segment looks like a chunk key gets special handling
>       store['0.0'] = b'xxx'

zarr/tests/test_storage.py:773:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <zarr.storage.FSStore object at 0x7fb7501c2c18>, key = '0.0', value = b'xxx'

    def __setitem__(self, key, value):
        if self.mode == 'r':
>           raise PermissionError
E           PermissionError

zarr/storage.py:974: PermissionError
```

We realized recently that the default mode for a regular DirectoryStore is `w` meaning that directories are automatically created:

```
In [18]: z=zarr.open('this/file/does/not/exist')
In [19]: list(z.groups()), list(z.arrays())
Out[19]: ([], [])
```

I was debating suggesting a change to "r" on a separate issue, but we'll probably need to keep it in mind for this PR as well.

If I force a "w" mode:

```
@@ -764,7 +764,7 @@ class TestNestedDirectoryStore(TestDirectoryStore, unittest.TestCase):
         path = tempfile.mkdtemp()
         atexit.register(atexit_rmtree, path)
         store = NestedDirectoryStore(path, normalize_keys=normalize_keys,
-                                     key_separator='/', auto_mkdir=True)
+                                     key_separator='/', auto_mkdir=True, mode="w")
         return store
```

then I again see the previous error:

```
zarr/tests/test_storage.py:230:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
zarr/storage.py:970: in __getitem__
    return self.map[key]
/opt/anaconda/envs/py36/lib/python3.6/site-packages/fsspec/mapping.py:76: in __getitem__
    result = self.fs.cat(k)
/opt/anaconda/envs/py36/lib/python3.6/site-packages/fsspec/spec.py:587: in cat
    return self.open(path, "rb").read()
/opt/anaconda/envs/py36/lib/python3.6/site-packages/fsspec/spec.py:775: in open
    **kwargs
/opt/anaconda/envs/py36/lib/python3.6/site-packages/fsspec/implementations/local.py:108: in _open
    return LocalFileOpener(path, mode, fs=self, **kwargs)
/opt/anaconda/envs/py36/lib/python3.6/site-packages/fsspec/implementations/local.py:175: in __init__
    self._open()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <fsspec.implementations.local.LocalFileOpener object at 0x7faa10221908>

    def _open(self):
        if self.f is None or self.f.closed:
            if self.autocommit or "w" not in self.mode:
>               self.f = open(self.path, mode=self.mode)
E               IsADirectoryError: [Errno 21] Is a directory: '/var/folders/z5/txc_jj6x5l5cm81r56ck1n9c0000gn/T/tmpwebfv41i/c'

/opt/anaconda/envs/py36/lib/python3.6/site-packages/fsspec/implementations/local.py:180: IsADirectoryError
```

which seems related, but I'm unclear how to get started with it, since it seems to be saying that `fsspec` expects one of the intermediate keys to be a file rather than a directory. I'll keep poking around but suggestions welcome.

martindurant: I probably should have started with the non-nested stores for testing...
Exactly when directory trees should be removed and when files should be removed to make way for directories, I tries to reverse-engineer from the existing tests. 

The actual error here, is that zarr expects a `KeyError` when trying to read "c", but `fsspec` as of https://github.com/intake/filesystem_spec/pull/259 only converts `FileNotFound` into a `KeyError`. I suppose `IsADirectory` should be caught too, or the code should call `fs.isfile()`?

zarr/tests/test_storage.py
```diff
@@ -762,7 +763,8 @@ class TestNestedDirectoryStore(TestDirectoryStore, unittest.TestCase):
     def create_store(self, normalize_keys=False):
         path = tempfile.mkdtemp()
         atexit.register(atexit_rmtree, path)
-        store = NestedDirectoryStore(path, normalize_keys=normalize_keys)
+        store = NestedDirectoryStore(path, normalize_keys=normalize_keys,
+                                     key_separator='/', auto_mkdir=True)
```
joshmoore, line 767: @martindurant : when I try `key_separator='/'`, I'm seeing:
```
DEBUG:s3fs:CALL: head_object - ({},) - {'Bucket': 'idr', 'Key': 'zarr/v0.1-extra/6001240.zarr.v0.1n/0/0//zarray'}
```

martindurant: What are you expecting to see?
(or are you wondering why there is logging at all?)

joshmoore: I'd expect the HEAD/GET to be for `.../.zarray` rather than `...//zarray`

martindurant: aha, OK - the "." got replaced - I suppose we have to special-case for the `".x*"` keys as opposed to leaf data files

martindurant:
This appears to pass all tests with fsspec master, where I have substituted FSStore for DirectoryStore and NestedDirectoryStore in test_storage (not the final solution!).

Note that FSStore now directly implements consolidation, also in write mode (which is controversial, since it makes each metadata write take two remote calls, and introduces possible sync problems).

zarr/tests/test_storage.py
joshmoore: I opened https://github.com/martindurant/zarr/pull/1 with the special casing.

martindurant:
Would it be useful to have fsspec master in this POC to show it will pass?

rabernat:
I appreciate all the work on this from @martindurant. At the same time, I do find it confusing to hear things like "FSStore now directly implements consolidation". We should have a clear separation between a storage layer and the zarr layer. It makes me nervous to see more features seeping into the storage layer.

martindurant:
> We should have a clear separation between a storage layer and the zarr layer

In my mind, fsspec is the storage layer, and everything in zarr.storage can be specific/specialised (such as the original [conslolidated](https://github.com/zarr-developers/zarr-python/blob/master/zarr/storage.py#L2462)). You *could* use the previous consolidated with FSStore if you wanted, but this new version seems to me to remove duplication, as well as adding write-mode.

On of the issues I am trying to solve is "which store do I subclass": anyone coming to zarr and wanting to implement something new now would have to accommodate regular keys, nested keys *or* consolidated, not any combination.

rabernat:
Thanks for the clarification Martin. Now it makes sense to me.

martindurant:
cc @joshmoore . I would need to reread what is contained here to remind myself where things stood. I know that I tried to do too much in one class. This does *not* touch any async issue.

joshmoore:
@martindurant : understood. Thanks. I still see value in the simplicity. Happy to test.

martindurant:
(actually passing tests, just failing on flake, because I stole the DirectoryStore and NestedDirectoryStore tests rather than write new ones)

martindurant:
This PR is ready for comments, as it shows that things pass for normal directories and nested (not that N5 depends upon fsspec - but this shows that it could). storage_options has appeared in open_array and open_group and implied in the various create functions; but the arg is not useful for anything other than fsspec. You *can* specify consolidated as part of the store, which is interesting, but perhaps not necessary (und so far, untested).

martindurant:
(windows test is failing because PermissionError is not being converted into KeyError - another point of discussion)

joshmoore:
Two quick feedback points that I haven't been able to pinpoint yet:
 - Testing this required an update of `fsspec` which required manually install `aiohttp`
 - For one of the implementations I'm testing, `isdir()` seems to have changed because `self.info(path)` is now type file.

martindurant:
> which required manually install aiohttp

fsspec only needs aiohttp when using the HTTP backend, so it's not strictly a dependency (see https://github.com/intake/filesystem_spec/pull/371 )

> isdir() seems to have changed

Please raise an issue

joshmoore:
> so it's not strictly a dependency (see intake/filesystem_spec#371 )

:+1: I'll add a dependency on `fsspec[http]` one that's ready.

> Please raise an issue

Will do.

joshmoore:
@martindurant : where did the commits from https://github.com/martindurant/zarr/pull/1/files go? Do I need to resurrect them?

martindurant:
The "select commits" button on that page still seems to allow you to pick. 
Note that for s3fs based on aiobotocore, moto mocking has some problems, and tests should be run against moto server (which runs as a separate process and you need to specify the local endpoint). This is what is now done in dask and pandas.

zarr/tests/util.py
```diff
@@ -1,7 +1,9 @@
 # -*- coding: utf-8 -*-
 import collections
+from distutils.version import LooseVersion
```
Carreau, line 3: distutils will be removed in Python 3.10 which is planned for end of 2021

zarr/storage.py
```diff
@@ -946,6 +947,135 @@ def atexit_rmglob(path,
...
+                 **storage_options):
+        import fsspec
+        self.path = url
+        self.normalize_keys = normalize_keys
+        self.key_separator = key_separator
+        self.map = fsspec.get_mapper(url, **storage_options)
+        self.fs = self.map.fs  # for direct operations
+        self.mode = mode
+        self.exceptions = exceptions
+        # TODO: should warn if consolidated and write mode?
```
Carreau, line 964: TODO ?

zarr/storage.py
```diff
@@ -946,6 +947,135 @@ def atexit_rmglob(path,
...
+        if self.fs.exists(url) and not self.fs.isdir(url):
+            err_fspath_exists_notdir(url)
+        self.consolidated = consolidated
+        self.metadata_key = metadata_key
+        if consolidated:
+            self.meta = json.loads(self.map.get(metadata_key, b"{}").decode())
```
Carreau, line 970: There is a `json_loads` utils function used in other place of this file:

```
def json_loads(s):
    """Read JSON in a consistent way."""
    return json.loads(ensure_text(s, 'ascii'))
```

Do you want to use it ?

alimanfoo:
Dropping in comments from today's discussion on the community call. I have no objection to this PR, would be great to see it go in. Two points that come up are:

* Should this store support consolidated metadata internally? Or can it let that be handled in the layer above? I would have a mild preference to let this be handled in the layer above, but happy to go either way.

* Should this store support a chunk key separator argument? Or can it let that be handled in a layer above via composable stores? Ultimately composable stores would seem like a good idea, but given they don't exist yet it could be useful for @joshmoore and others to have FSStore support this for now.

@martindurant happy for you to make the call on how to round this one off.

martindurant:
Has anyone any idea why numcodecs failed to build during the docs stage [here](https://travis-ci.com/github/zarr-developers/zarr-python/jobs/383280395) ?

joshmoore:
> Has anyone any idea why numcodecs failed to build during the docs stage [here](https://travis-ci.com/github/zarr-developers/zarr-python/jobs/383280395) ?

Due to the now yanked https://pypi.org/project/numcodecs/0.7.0/ as we try to get the wheels deployed. (Sorry 'bout that)

martindurant:
(never mind on the latter, coverage just updated itself)

joshmoore:
> Should the problem go away then?

It should already be gone! Oddly travis is still finding it, perhaps because of some caching??

```
https://files.pythonhosted.org/packages/a1/b2/9c4fc0e4bc10a59442ced40dafc474f31bdc49699479da2e8912714e88af/numcodecs-0.7.0.tar.gz (3.0MB) ...
```

martindurant:
@joshmoore : is that [S3 test](https://github.com/zarr-developers/zarr-python/pull/546/files#diff-9e9d305bfa7d91cb050c0936c12ee1adR914) OK for you?

martindurant:
(I plan to merge this as is, unless there are further comments)

joshmoore (approved):
Tests pass, thanks @martindurant. I also tried adding a ```@pytest.mark.skipif(have_fsspec is False, reason="needs fsspec")
class TestNestedFSStore2(TestFSStore):

    def create_store(self, normalize_keys=False):
        path = tempfile.mkdtemp()
        atexit.register(atexit_rmtree, path)
        store = FSStore(path, normalize_keys=normalize_keys,
                        key_separator='/', auto_mkdir=True)
        return store
```
locally which equally passed. :+1:

zarr/tests/test_storage.py
joshmoore: Note: this issue still exists. I've opened https://github.com/zarr-developers/zarr-python/pull/699 against the mainline.

shoyer:
Quick question -- is there a good reason why FSStore defaults to `normalize_keys=True`, unlike the other Zarr stores? We were pretty surprised to find different behavior using Zarr directly and GCSFS.

joshmoore:
I had the same question. I've need to hard-code the argument in all of my uses. cc: @martindurant

martindurant:
I certainly don't mind the default changing, if that is the consensus. I would like to claim there was a good reason for the current , but ...

PR #561, opened 2020-05-06
modify top level description of zarr

rabernat:
This PR changes the very first sentence of our documentation from

> Zarr is a Python package providing an implementation of chunked, compressed, N-dimensional arrays.

to

> Zarr is format for the storage of chunked, compressed, N-dimensional arrays. These documents describe the Zarr format and its Python implementation.

I think it's important at this stage to make this clarification.

Should we also consider linking to the implementations in other languages?

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -4,7 +4,7 @@
 Zarr
 ====
 
-Zarr is format for the storage of chunked, compressed, N-dimensional arrays.
+Zarr is a format for the storage of chunked, compressed, N-dimensional arrays.
 These documents describe the Zarr format and its Python implementation.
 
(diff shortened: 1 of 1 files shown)


docs/index.rst
```diff
@@ -4,8 +4,8 @@
 Zarr
 ====
 
-Zarr is a Python package providing an implementation of chunked,
-compressed, N-dimensional arrays.
+Zarr is format for the storage of chunked, compressed, N-dimensional arrays.
```
joshmoore, line 7:
```suggestion
Zarr is a format for the storage of chunked, compressed, N-dimensional arrays.
```

PR #564, opened 2020-05-11
Always display the 10 slowest test when running pytest.

Carreau:
This is useful when inspecting CI builds to have an idea of what takes
the longest time – especially since CI may have some slow resources
(disk, network). And may help to speedup test suite, which is always
good for fast development.


Carreau:
If you run pytest locally you will also love `--ff` and `--lf`, respectively run last failing test first, and only last failing tests.

Carreau:
Thanks !

PR #565, opened 2020-05-14
Spec v2 update link to stable numpy docs

Carreau:
Numpy docs have move, update links to newer version.


Carreau:
Ok, i've done it using intersphinx in order for the link to stay valid even if they move inside the numpy docs.

Issue #569, opened 2020-06-15
Bugfix: json_dump can not encode numpy data types


pziarsolo:
#### Minimal, reproducible code sample, a copy-pastable example if possible

```python

import numpy
import dask.array
import zarr


def main():
    np_array = numpy.random.randint(1, 10, size=1000000)
    array = dask.array.from_array(np_array)

    array = array[array > 5]

    with TemporaryDirectory() as tmpdir:
        array.compute_chunk_sizes()
        delayed = dask.array.to_zarr(array, url=tmpdir, compute=False,
                                                      component='/data')
        dask.compute(delayed)

        z_object = zarr.open_group(tmpdir, mode='r')

        assert numpy.all(np_array[np_array > 5] == z_object.data[:])


if __name__ == '__main__':
    main()


```

#### Problem description
dask array.compute_chunk_size() generates chunks with numpy.int64 integers.

When zarr tries to encode dask array metadata into json format, it  fails because default json encoder does not understand numpy data types.

Giving an numpy  data type encoder to json.dump solves the problem

#### Version and installation information

Please provide the following:
zarr.__version__ 2.4
numcodecs      0.6.4
python 3.8
Operating system: Linux
How Zarr was installed: pip into virtual environment

pziarsolo:
pull request #570 solves the problem

jakirkham:
> dask array.compute_chunk_size() generates chunks with numpy.int64 integers.

In that case, I think the right thing to do would be to fix this in Dask. I'm pretty sure this is not intentional behavior. Could you please raise an issue on [the Dask issue tracker]( https://github.com/dask/dask/issues ) or better yet add a PR over there? Thanks for letting us know about this issue 🙂

pziarsolo:
> > dask array.compute_chunk_size() generates chunks with numpy.int64 integers.
> 
> In that case, I think the right thing to do would be to fix this in Dask. I'm pretty sure this is not intentional behavior. Could you please raise an issue on [the Dask issue tracker](https://github.com/dask/dask/issues) or better yet add a PR over there? Thanks for letting us know about this issue slightly_smiling_face

Ok.
I will create and issue in dask
thanks

alimanfoo:
Just to second @jakirkham, probably better to get dask to convert this back into plain Python ints.

jakirkham:
[...]

PR #570, opened 2020-06-15
json dump can encode numpy data types

pziarsolo:
It adds a numpy data type encoder to json_dumps in zarr/utils

All tests passed. No tests added

### DIFF
--- a/zarr/util.py
+++ b/zarr/util.py
@@ -19,10 +19,24 @@ object_codecs = {
 }
 
 
+# from
+# https://stackoverflow.com/questions/50916422/python-typeerror-object-of-type-int64-is-not-json-serializable/50916741
+class NpEncoder(json.JSONEncoder):
+    def default(self, obj):
+        if isinstance(obj, np.integer):
+            return int(obj)
+        elif isinstance(obj, np.floating):
+            return float(obj)
+        elif isinstance(obj, np.ndarray):
+            return obj.tolist()
+        else:
+            return super(NpEncoder, self).default(obj)
+
+
 def json_dumps(o):
     """Write JSON in a consistent, human-readable way."""
     return json.dumps(o, indent=4, sort_keys=True, ensure_ascii=True,
-                      separators=(',', ': ')).encode('ascii')
+                      separators=(',', ': '), cls=NpEncoder).encode('ascii')
 
 
 def json_loads(s):


zarr/util.py
```diff
@@ -19,10 +19,24 @@
 }
 
 
+# from
+# https://stackoverflow.com/questions/50916422/python-typeerror-object-of-type-int64-is-not-json-serializable/50916741
```
jakirkham, line 23: I'm not sure we can take code from StackOverflow due to [licensing]( https://stackoverflow.com/help/licensing ), which would be incompatible with MIT (Zarr's licensing). Is there a way to write this where it doesn't depend on StackOverflow code?

zarr/util.py
```diff
@@ -19,10 +19,24 @@
...
+            return obj.tolist()
+        else:
+            return super(NpEncoder, self).default(obj)
+
+
 def json_dumps(o):
     """Write JSON in a consistent, human-readable way."""
     return json.dumps(o, indent=4, sort_keys=True, ensure_ascii=True,
-                      separators=(',', ': ')).encode('ascii')
+                      separators=(',', ': '), cls=NpEncoder).encode('ascii')
```
jakirkham, line 39: Wouldn't we need a way to decode NumPy values too? How would we do that?

zarr/util.py
pziarsolo: It's just a class. Maybe renaming it and removing all references?

The reference is just to give credit where it is due.

zarr/util.py
pziarsolo: I don't understand this. 

What do you mean with numpy values? The encoder is just what it does. It adds encoders for three numpy data types: integers, floats and arrays.

alimanfoo: I think @jakirkham means, when reading the JSON back, do you want to convert the array back into a numpy array, or are you happy to see it as a bare list? 

I.e., this encoder will not round-trip, so some users might get a surprise when the store an array and then read back and get a list.

zarr/util.py
```diff
@@ -19,10 +19,24 @@
...
 
 
+# from
+# https://stackoverflow.com/questions/50916422/python-typeerror-object-of-type-int64-is-not-json-serializable/50916741
+class NpEncoder(json.JSONEncoder):
+    def default(self, obj):
+        if isinstance(obj, np.integer):
+            return int(obj)
+        elif isinstance(obj, np.floating):
+            return float(obj)
```
shoyer, line 29: Rather than only supporting int/float, you could support arbitrary NumPy scalars with:
```suggestion
        if isinstance(obj, np.generic):
            return obj.item()
```

zarr/util.py
shoyer: I think it's definitely OK to return Python builtin scalars instead of NumPy scalars. Half the time users don't even realize they're using NumPy scalars because NumPy uses the same repr 😒 

I'm less sure about whether or not it's a good idea to automatically convert arrays into lists.

zarr/util.py
jakirkham: Yeah another way we might do this is...

```suggestion
        if isinstance(obj, numbers.Integral):
            return int(obj)
        elif isinstance(obj, numbers.Real):
            return float(obj)
```

shoyer: The other numpy scalar types that come to mind (missed by the current checks) are `bool_`, `bytes_` and `str_`.

jakirkham:
As noted in issue ( https://github.com/zarr-developers/zarr-python/issues/569 ), the underlying issue appeared to be with Dask and has since been fixed in Dask 2.19.0 with PR ( https://github.com/dask/dask/pull/6326 ). Given that, am going to go ahead and close this out. Though please feel free to reopen if there are other cases where this functionality is needed. Thanks!

PR #572, opened 2020-06-25
Remove redundant Conditional imports.

Carreau:
Zarr is only Python 3.5+ so this check is not necessary anymore.

### DIFF
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -29,7 +29,7 @@ import warnings
 import zipfile
 from collections import OrderedDict
 from collections.abc import MutableMapping
-from os import scandir
+from os import scandir, replace
 from pickle import PicklingError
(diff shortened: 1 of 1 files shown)


zarr/storage.py
```diff
@@ -65,14 +65,7 @@
...
 
-# Find which function to use for atomic replace
-if sys.version_info >= (3, 3):
-    from os import replace
-elif sys.platform == "win32":  # pragma: no cover
-    from osreplace import replace
-else:  # pragma: no cover
-    # POSIX rename() is always atomic
-    from os import rename as replace
+from os import replace
```
jakirkham, line 68: Can we combine this with the `from os import ...` above?

Carreau: done.

PR #581, opened 2020-08-04
Move metadata handling in a class.

Carreau:
This will make it a tiny bit easier to make a transition to V3 by having
groups/array use an instance to encode/decode instead of refering to the
function/module.

While for v2 it does not make much difference; I'm expecting instances
in V3 to likely be (some of) the pieces that handle extensions, and be
passed around in a hierarchy.

I currently re-expose the methods at the module level to not change the
API/tests, though I expect to wrap them and add deprecation once I've
move the Groups/Arrays/Attributes to use those.

### DIFF
--- a/zarr/meta.py
+++ b/zarr/meta.py
@@ -9,185 +9,194 @@ from zarr.util import json_dumps, json_loads
 
 ZARR_FORMAT = 2
 
-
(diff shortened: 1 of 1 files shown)


joshmoore:
Handled in #839.

PR #584, opened 2020-08-11
adds partial_decompress capabilites

andrewfulton9:
Draft for adding capability to partial decompress chunks when they have been compressed for zarr. Not finished yet. Have added an indexer for getting the within chunk start and nitem pairs. Still need to finish adding logic to core.Array._chunk_getitem to use the indexer with and add tests. Also need to add test of example I found where I can get the PartialChunkIterator to fail. Test isn't added because I need to figure out the expected value. It fails with the below example array and slice tuple because it calculates the start value wrong if a middle dimension is 1. Also need to add tests that check if the index works for 'F' order

```
arr = np.arange(2, 100002).reshape((10, 1, 10000))
buff = codec.encode(arr)

slices = (slice(5, 8, 1), slice(2, 4, 1), slice(0, 5, 1))
```

TODO:


zarr/indexing.py
```diff
@@ -822,3 +822,62 @@ def pop_fields(selection):
...
+        selection_slice = selection[i] if i < len(selection) else None
+        if selection_slice:
+            selection_slice_size = len(range(*selection_slice.indices(len(arr))))
+            selection_shape.append(selection_slice_size)
+        else:
+            selection_shape.append(size)
+    return tuple(selection_shape)
+
+
+class PartialChunkIterator(object):
```
Carreau, line 841: Python 3 only, we can drop objects now :-P
```suggestion
class PartialChunkIterator:
```

zarr/indexing.py
```diff
@@ -822,3 +822,62 @@ def pop_fields(selection):
...
+
+    def __init__(self, selection, arr):
+        self.arr = arr
+        self.selection = list(selection)
+
+        for i, dim_shape in enumerate(self.arr.shape[slice(None, None, -1)]):
```
Carreau, line 847: A tiny bit shorter ?

```suggestion
        for i, dim_shape in enumerate(self.arr.shape[::-1]):
```

zarr/tests/test_indexing.py
```diff
@@ -1289,3 +1289,29 @@ def test_set_selections_with_fields():
             a[key][ix] = v[key][ix]
             z.set_mask_selection(ix, v[key][ix], fields=fields)
             assert_array_equal(a, z[:])
+
+
+@pytest.mark.parametrize('selection, expected', [
+    ((slice(5, 8, 1), slice(2, 4, 1), slice(0, 100, 1)),
+     [(5200, 200, (slice(0, 1, 1), slice(0, 2, 1))),
+      (6200, 200, (slice(1, 2, 1), slice(0, 2, 1))),
+      (7200, 200, (slice(2, 3, 1), slice(0, 2, 1)))]),
```
Carreau, line 1298: So trying to understand this. 

```
arr[5:8, 2:4, 0:100]
```

mean we need to read 200 items at position 5200, 6200 and 7200 ?

zarr/tests/test_indexing.py
```diff
@@ -1289,3 +1289,29 @@ def test_set_selections_with_fields():
...
+    ((slice(5, 8, 1), slice(2, 4, 1), slice(0, 5, 1)),
+     [(5200.0, 5.0, (slice(0, 1, 1), slice(0, 1, 1), slice(0, 5, 1))),
+      (5300.0, 5.0, (slice(0, 1, 1), slice(1, 2, 1), slice(0, 5, 1))),
+      (6200.0, 5.0, (slice(1, 2, 1), slice(0, 1, 1), slice(0, 5, 1))),
+      (6300.0, 5.0, (slice(1, 2, 1), slice(1, 2, 1), slice(0, 5, 1))),
+      (7200.0, 5.0, (slice(2, 3, 1), slice(0, 1, 1), slice(0, 5, 1))),
+      (7300.0, 5.0, (slice(2, 3, 1), slice(1, 2, 1), slice(0, 5, 1)))]),
```
Carreau, line 1305: floats are weird, only int in output no ?

zarr/indexing.py
```diff
@@ -822,3 +822,62 @@ def pop_fields(selection):
...
+            index = len(self.arr.shape) - (i+1)
+            if index <= len(selection)-1:
+                slice_nitems = len(range(*selection[index].indices(len(self.arr))))
+                if slice_nitems == dim_shape:
+                    self.selection.pop()
+                else:
+                    break
```
Carreau, line 854: If I understand this block of code correctly:

here we seem to be looking for dimensions where we select the whole thing.
we start from the end of selection, and shape, and while there is no selections
or the length is 100%, we pop.

Is that correct ?

zarr/tests/test_indexing.py
andrewfulton9: Yeah, really its taking 1 item from the first dimension and 2 items of 100 from the second dimension. So I wrote the code so that it takes 200 items since they are consecutive in the compressed buffer anyway. The data just has to be reshaped before it is put into the chunk output array

zarr/tests/test_indexing.py
andrewfulton9: Yeah, I think the np.prod call used to generate the nitems is making that a float. I'll update that

zarr/indexing.py
```diff
@@ -822,3 +822,62 @@ def pop_fields(selection):
...
+        out_slices = []
+        chunk_loc_slices = []
+
+        last_dim_slice = None if self.selection[-1].step > 1 else self.selection.pop()
+        for sl in self.selection:
+            dim_out_slices = []
+            dim_chunk_loc_slices = []
+            for i, x in enumerate(range(*sl.indices(len(self.arr)))):
+                dim_out_slices.append(slice(i, i+1, 1))
+                dim_chunk_loc_slices.append(slice(x, x+1, 1))
```
Carreau, line 865: IIUC here you are computing each individual 1 element wide selection across each axis both in the origin chunk coordinate system, and in the output version right ?

Would/could this be simpler if the step of the `sl` ic 1 ? because in that case you have a continuous sl right ? or did I misunderstood.

No a request to change, it can be an optimisation for later.

zarr/indexing.py
andrewfulton9: That's right. I do this to maximize nitems/minimize the number of partial decompresions called. This logic helps get to the 200 nitems in the test example above.

zarr/indexing.py
```diff
@@ -822,3 +822,61 @@ def pop_fields(selection):
...
+            out_slices.append(dim_out_slices)
+            chunk_loc_slices.append(dim_chunk_loc_slices)
+        if last_dim_slice:
+            out_slices.append(
+                [slice(0, last_dim_slice.stop - last_dim_slice.start, 1)])
+            chunk_loc_slices.append([last_dim_slice])
+
+        self.out_slices = itertools.product(*out_slices)
+        self.chunk_loc_slices = itertools.product(*chunk_loc_slices)
```
Carreau, line 874: and for myself, `out_slices` are the slices to assign elements in the output array, `chunk_loc_slices` the indices of where to read the the original chunk.

Carreau:
You may want to also test with negative indexing, and negative steps just in case.

zarr/indexing.py
```diff
@@ -822,3 +822,61 @@ def pop_fields(selection):
...
+
+    def __iter__(self):
+        for out_selection, chunk_selection in zip(self.out_slices, self.chunk_loc_slices):
+            start = 0
+            for i, sl in enumerate(chunk_selection):
+                start += sl.start * np.prod(self.arr.shape[i+1:])
+            nitems = (chunk_selection[-1].stop - chunk_selection[-1].start) * np.prod(self.arr.shape[len(chunk_selection):])
```
Carreau, line 881: is `len(chunk_selection)` ever changing ? it comes from `itertools.product(*chunk_loc_slices)`, so each item is likely the same length ? So likely `np.prod(self.arr.shape[len(chunk_selection):])` is a constant right ? Which sorta I think make sens as you are basically measuring the jump between two stride in an N-dimensional lattice, but it's hard to imagine.

zarr/indexing.py
andrewfulton9: Thats right, except for the slice of the last dimension, if the that slice has a step of 1. if that slice has a step of more than one, then it is included here, though if thats the last dimension of the chunk, then it would be really inefficient since `nitems` would be 1. If the slice of the last dimension has a step of one, then the slice will cover more than one slice and the output slice is calculated by stop - start

zarr/indexing.py
andrewfulton9: Thats right

zarr/tests/test_indexing.py
```diff
@@ -1289,3 +1289,61 @@ def test_set_selections_with_fields():
             a[key][ix] = v[key][ix]
             z.set_mask_selection(ix, v[key][ix], fields=fields)
             assert_array_equal(a, z[:])
+
+
+@pytest.mark.parametrize('selection, arr, expected', [
+    ((slice(5, 8, 1), slice(2, 4, 1), slice(0, 100, 1)),
+     np.arange(2, 100_002).reshape((100, 10, 100)),
```
Carreau, line 1296: Python 3.5 won't like those... we can argue to drop Python 3.5 I think

zarr/core.py
```diff
@@ -1731,11 +1750,16 @@ def _chunk_setitem_nosync(self, chunk_coords, chunk_selection, value, fields=Non
...
 
-    def _decode_chunk(self, cdata):
+    def _decode_chunk(self, cdata, start=None, nitems=None, expected_shape=None):    
 
         # decompress
         if self._compressor:
-            chunk = self._compressor.decode(cdata)
+            # only decode requested items
+            if (all([x is not None for x in [start, nitems]])
+                and self._compressor.codec_id == 'blosc'):
```
Carreau, line 1759:
```suggestion
                and self._compressor.codec_id == 'blosc') and hasattr(self._compressor, 'decode_partial'):
```

And/or you may want to modify `.travis.yml` to install numcodec from master.

joshmoore:
Hi @andrewfulton9. Master is now conflicting with this branch. Do you have a moment to take a look at the conflicts?

andrewfulton9:
Hey @joshmoore. I think this PR should actually be closed. I used the work here for the partial read capabilities that was merged into master in PR #667

PR #590, opened 2020-08-27
Use Error subclasses instead of functions.

Carreau:
I believe this has several advantage:

  1) reading the code it is obvious we are directly raising instead of
  doing more stuff.

  2) in traceback the last frame will be correct, so it's easier to find
  where the error occurs.

  3) All of these are now subclasses which can be caught independently
  by libraries while still being subclass of ValueError, so backward
  compatible. Can be caught without catching other ValueErrors.

  4) the name of the class is descriptive

I'm not too sure about some naming. Is ContainsXxxxError sufficient to
understand that it contain Xxxxx but Xxxx was not expected ?

TODO:

I'm not too sure if I should remove the err_ function, as they may be public APIs.
Thus the coverage will decrease; 

if you are +1 on this approach I'll do the few remaining and update the docs.

### DIFF
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -15,6 +15,10 @@ Next release
 * Explicitly close stores during testing.
   By :user:`Elliott Sales de Andrade <QuLogic>`; :issue:`442`
 
+* Many of the convenience functions to emit errors (``err_*`` from
+  ``zarr.errors``  have been replaced by ``ValueError`` subclasses. The
+  functions are deprecated and will be removed in the future. :issue:`590` )
+
 * Improve consistency of terminology regarding arrays and datasets in the 
   documentation.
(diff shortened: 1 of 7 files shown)


jakirkham:
LGTM

@alimanfoo thoughts? 🙂

jakirkham:
This still makes sense to me. Would be great if @alimanfoo is able to look. I'm guessing there was something this function was solving, but I don't have any details unfortunately. So hopefully Alistair can shed some light on that 🙂

Carreau:
> I'm guessing there was something this function was solving

My guess is/was custom error messages, it is pretty uncommon to subclass Errors and tweak the `__init__`, or at least it is not a practice that is often seen.

jakirkham (approved):

Issue #602, opened 2020-09-11
End support for python 3.5


clbarnes:
[EOL 2020-09-13](https://devguide.python.org/#status-of-python-branches) (see also [PEP478](https://www.python.org/dev/peps/pep-0478/#id4)) so no need to merge this in the next couple of days.

This will make pathlib use easier (see #601), among other things.

joshmoore:
see also: https://github.com/zarr-developers/governance/issues/11

joshmoore:
Closing along with https://github.com/zarr-developers/zarr-python/pull/603 (in order to open the Python 3.6 version)

PR #603, opened 2020-09-11
End support for EOL python 3.5

clbarnes:
Remove references to 3.5 from CI and test configuration, setup.py, and tox, as per #602

Note that the ipython notebooks seem to have been run against 3.5; we should consider running them against 3.8 and check them in again to future-proof us a little longer.

Strictly, 3.5 only becomes EOL 2020-09-13, so take the weekend ;) As stated in https://github.com/zarr-developers/governance/issues/11 it may be worth putting out a final patch version supporting 3.5 before this goes in.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,6 @@ services:
 
 matrix:
   include:
-    - python: 3.5
     - python: 3.6
       env: BUILD_DOCS='true'
     - python: 3.7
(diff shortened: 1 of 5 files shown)


Carreau:
We've done the last release with compatibility with 3.5, though this now have conflicts, would you mind rebasing ?

jakirkham:
Indeed happy to +1 once conflicts are resolved :)

clbarnes:
Rebased!

Carreau (approved):

PR #606, opened 2020-09-16
Fs multiget

martindurant:
For async IO. 

Read-only for now, and needs fsspec master

TODO:


martindurant:
Thoughts?

requirements_dev_optional.txt
```diff
@@ -18,6 +18,7 @@ pytest-cov==2.7.1
 pytest-doctestplus==0.4.0
 pytest-remotedata==0.3.2
 h5py==2.10.0
-s3fs==0.5.0; python_version > '3.6'
+s3fs==0.5.1; python_version > '3.6'
+git+https://github.com/intake/filesystem_spec; python_version > '3.6'
```
jakirkham, line 22: Is there a package on PyPI or conda-forge that we could install?

martindurant: Not yet, no.

jakirkham: Do we have a rough sense when these might be created?

martindurant: When everything is good here, I can be sure to make a release before merging, so that this line can be reverted

alimanfoo:
We're  just reviewing this now in the zarr v3 dev call. Looks good to me :-)

requirements_dev_optional.txt
alimanfoo: Sounds good @martindurant.

martindurant:
@alimanfoo , can I get your +1 on this?

martindurant:
@jakirkham @joshmoore , any more thoughts?
I'd like to merge. Since this only affects fsspec storage, which is not used by anyone in the wold yet, I think this is safe.

rabernat:
Just catching up on this PR. Fantastic work @martindurant!

Are we planning to eventually do `multiset` as well?

martindurant:
Yes, of course. I have local code that is partway there.

martindurant:
Yes, the (test) requirement that went in with this PR is for the latest release fsspec. I don't know if we want to say anything about this functionality's dependency in the release notes.

jakirkham:
Good point. If there is some wording you'd like to see in the release notes, would suggest proposing it in PR ( https://github.com/zarr-developers/zarr-python/pull/616 )

martindurant:
Hm, not sure whether to say use fsspec>=0.8.0 or to fallback to sequential

cgohlke:
Is the `getitems` method documented?  Sorry if I missed something obvious, but I could not find it. I try to implement `getitems` for a TIFF store.

martindurant:
https://filesystem-spec.readthedocs.io/en/latest/api.html#fsspec.FSMap.getitems

cgohlke:
Thank you. I don't use fsspec but try to implement `getitems` independently. My understanding is that the following should work? Somehow it returns garbage
```
    def getitems(self, keys):
        return {key: self.__getitem__(key) for key in keys}
```

martindurant:
That looks OK to me, but that's what the current implementation is effectively doing anyway, with the loop outside of the storage class.

jakirkham:
I think we would still be open to a PR if someone wants to implement that directly in Zarr.

martindurant:
It would may be sensible to make a fallback implementation of getitems in the storage class, doing the same loop currently higher up; but that would require a strict class hierarchy and not actually make anything simpler overall.

cgohlke:
False alarm, `return {key: self.__getitem__(key) for key in keys}` works as expected. I had forgotten to comment out a `yield` statement in the function that was left from previous tries to implement `getitems` as a generator...

jakirkham:
> It would may be sensible to make a fallback implementation of getitems in the storage class, doing the same loop currently higher up; but that would require a strict class hierarchy and not actually make anything simpler overall.

I don't think that is strictly true. We already implement functions that have fallbacks (like [`rmdir`]( https://github.com/zarr-developers/zarr-python/blob/e3cdd1a50e1047304be2c91a017da6362f8df533/zarr/storage.py#L110 )) for storage classes that don't have them. So we could presumably do the same thing with `getitems`. The main benefit would be other functions in `Group` and `Array` could rely on `getitems` instead of `getitem` more generally.

That said, I think having a class hierarchy would be a nice improvement as long as we still have a way to coerce generic `MutableMapping`-based objects into something this hierarchy can work with (perhaps through a wrapper class). This mostly just reiterates what has come up in a few meeting now though.

PR #611, opened 2020-09-18
Reduce number of store queries in open(mode='a').

Carreau:
In case where 'shape' is passed as argument, this will avoid calling
"contains_array" which does store queries, as well as avoid testign if
the array contains a group as anyway this is what we were going to do in
the else case.

This does change behavior in edge case:
 - if shape is passed and the key is a group, we'll try to open as array
 instead of groups.
 - if contains_array or contains_group would have raise we might skip
 over those checks.

Though both of the above case should not be found in user code and
are bugs AFAICT, and would just fail differently.

--- 

Let's see if this triggers any test failure. 

TODO:

### DIFF
--- a/zarr/convenience.py
+++ b/zarr/convenience.py
@@ -83,12 +83,8 @@ def open(store=None, mode='a', **kwargs):
         else:
             return open_group(store, mode=mode, **kwargs)
 
-    elif mode == 'a':
-        if contains_array(store, path):
-            return open_array(store, mode=mode, **kwargs)
-        elif contains_group(store, path):
-            return open_group(store, mode=mode, **kwargs)
-        elif 'shape' in kwargs:
+    elif mode == "a":
+        if "shape" in kwargs or contains_array(store, path):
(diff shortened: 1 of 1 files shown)


zarr/convenience.py
```diff
@@ -83,12 +83,8 @@ def open(store=None, mode='a', **kwargs):
...
             return open_group(store, mode=mode, **kwargs)
 
-    elif mode == 'a':
-        if contains_array(store, path):
-            return open_array(store, mode=mode, **kwargs)
-        elif contains_group(store, path):
-            return open_group(store, mode=mode, **kwargs)
-        elif 'shape' in kwargs:
+    elif mode == "a":
+        if "shape" in kwargs or contains_array(store, path):
```
Carreau, line 87:
```suggestion
        if ("shape" in kwargs) or contains_array(store, path):
```

alimanfoo (approved):

jakirkham (approved):

PR #616, opened 2020-09-23
Misc release notes

Carreau:
Misc addition to release notes.
I've also tagged all the PRs merged since 2.4 with the 2.5 milestone.

TODO:

### DIFF
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -17,8 +17,6 @@ This release will be the last to support Python 3.5, next version of Zarr will b
 * Fix minor bug in `N5Store`. 
   By :user:`gsakkis`, :issue:`550`.
 
-* add new `FSStore` to open any ``fsspec`` supported protocol :issue:`546`
-
 * Improve error message in Jupyter when trying to use the ``ipytree`` widget
   without ``ipytree`` installed.
   By :user:`Zain Patel <mzjp2>`; :issue:`537`
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -2,8 +2,8 @@ Release notes
 =============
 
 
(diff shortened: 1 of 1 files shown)


docs/release.rst
```diff
@@ -14,6 +17,8 @@ Next release
 * Fix minor bug in `N5Store`. 
   By :user:`gsakkis`, :issue:`550`.
 
+* add new `FSStore` to open any ``fsspec`` supported protocol :issue:`546`
+
```
jakirkham, line 21: Looks like this actually shows up below. So maybe we can drop it?

jakirkham (approved):
After looking at the updated notes, it appears we have everything here. Multiget appears listed below and so does the removal of error functions. Maybe I was looking at an outdated diff before? Anyways this looks good on my end. Thanks Matthias! 😄

docs/release.rst
Carreau:
```suggestion
```

Carreau:
Travis CI appear twice (one green, one orange) and the orange one seem to be a bug as te URL is the same as the passing one. 
I'll override and merge.

jakirkham:
Interesting was wondering if this was the leftover `travis-ci.org` project, but going to that link it shows the project as migrated to `travis-ci.com`. Not sure why the other build showed up.

Visiting the associated pending Travis CI detail link shows it as passing as well. So it may just be the GitHub status API didn't work when updating the build as passed.

Anyways all this to say ignoring it seems fine and I also don't have any clue why it showed up 😅

Carreau:
it's not the first time I see that. So 🤷

Issue #618, opened 2020-09-24
ABSStore ImportError


jhamman:
I'm finding that current versions of `azure-storage-blob` (12.5.0) are not importable when using the `ABSStore`. This was actually warned in this comment: https://github.com/zarr-developers/zarr-python/issues/528#issuecomment-657507825. Is anyone using this store anymore? Presumably, fsspec's mapper interface should work

#### Minimal, reproducible code sample, a copy-pastable example if possible

```python
import zarr

store = zarr.storage.ABSStore('carbonplan-scratch',
                              prefix=f'cmip-raw-conus/',
                              account_name="carbonplan",
                              account_key=os.environ["BLOB_ACCOUNT_KEY"])
```

#### Problem description

Simply creating the `ABSStore` raises an `ImportError`:

```python-traceback
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
<ipython-input-7-ad89fee562aa> in <module>
      4                               prefix=f'cmip-raw-conus/',
      5                               account_name="carbonplan",
----> 6                               account_key=os.environ["BLOB_ACCOUNT_KEY"])
      7 store

/srv/conda/envs/notebook/lib/python3.7/site-packages/zarr/storage.py in __init__(self, container, prefix, account_name, account_key, blob_service_kwargs)
   1980     def __init__(self, container, prefix='', account_name=None, account_key=None,
   1981                  blob_service_kwargs=None):
-> 1982         from azure.storage.blob import BlockBlobService
   1983         self.container = container
   1984         self.prefix = normalize_storage_path(prefix)

ImportError: cannot import name 'BlockBlobService' from 'azure.storage.blob' (/srv/conda/envs/notebook/lib/python3.7/site-packages/azure/storage/blob/__init__.py)
```

ABSStore is unusable in its current state.

#### Version and installation information

* Value of ``zarr.__version__``: '2.4.0'
* Value of ``numcodecs.__version__``: '0.7.2'
* Version of Python interpreter: 3.7.8
* Operating system (Linux/Windows/Mac): Linux
* How Zarr was installed (e.g., "using pip into virtual environment", or "using conda"): conda-forge

``conda env export`` :

<details>
<pre>
name: notebook
channels:
  - conda-forge
  - defaults
dependencies:
  - _libgcc_mutex=0.1=conda_forge
  - _openmp_mutex=4.5=1_gnu
  - abseil-cpp=20200225.2=he1b5a44_2
[...]

PR #620, opened 2020-09-27
Update ABSStore to current Azure Storage API version

jhamman:
This PR addresses #618 and updates the minimum supported version of `azure-storage-blob>=12`. 

You'll note the tests will fail for now. I haven't figured out where the emulation feature of the azure api is in the new version. I'll keep digging but tests on actual data are indicating this is mostly functional now. 

cc @tjcrone 

TODO:


zarr/storage.py
```diff
@@ -2255,29 +2253,32 @@ def rmdir(self, path=None):
...
-                if type(blob) == Blob:
-                    size += blob.properties.content_length
-            return size
+        for blob in self.client.list_blobs(name_starts_with=dir_path):
+            self.client.delete_blob(blob)
+
+    # It is possible azure.store.blob doesn't provide the content_length attribute on
+    # the blob propoeries object anymore. Something to look into.
+    #
+    # def getsize(self, path=None):
```
jhamman, line 2262: Flagging this as one major thing I haven't figured out.

zarr/storage.py
```diff
@@ -2224,27 +2224,25 @@ def __iter__(self):
...
             dir_path += '/'
         items = list()
-        for blob in self.client.list_blobs(self.container, prefix=dir_path, delimiter='/'):
-            if type(blob) == Blob:
+        for blob in self.client.list_blobs(name_starts_with=dir_path):
+            # items.append(self._strip_prefix_from_path(blob.name, dir_path))
+            if '/' not in blob.name:  # what is this doing?
                 items.append(self._strip_prefix_from_path(blob.name, dir_path))
             else:
                 items.append(self._strip_prefix_from_path(
```
jhamman, line 2248: @tjcrone - I'm pretty confused here. Any points on the `listdir` method would be much appreciated.

zarr/storage.py
tjcrone: https://azuresdkdocs.blob.core.windows.net/$web/python/azure-storage-blob/12.5.0/azure.storage.blob.html#azure.storage.blob.BlobProperties

zarr/storage.py
tjcrone: https://github.com/zarr-developers/zarr-python/blob/e3cdd1a50e1047304be2c91a017da6362f8df533/zarr/storage.py#L8

zarr/storage.py
```diff
@@ -2255,29 +2253,32 @@ def rmdir(self, path=None):
...
+    #
+    # def getsize(self, path=None):
+    #     from azure.storage.blob import Blob
+    #     store_path = normalize_storage_path(path)
+    #     fs_path = self.prefix
+    #     if store_path:
+    #         fs_path = self._append_path_to_prefix(store_path)
+    #     if self.client.get_blob_client(fs_path).exists():
+    #         return self.client.get_blob_properties(self.container,
+    #                                                fs_path).properties.content_length
```
shikharsg, line 2270: Thanks for this PR @jhamman 

I think the way to go here would be like so
```python
        blob_client = self.client.get_blob_client(fs_path)  # blob may or may not exist
        if blob_client.exists():
            return blob_client.get_blob_properties().size
```

zarr/storage.py
shikharsg: The `listdir` method which was implemented using the old version of the azure storage library, used the `list_blobs` method in that old library. That method in the old library had a [delimiter](https://docs.microsoft.com/en-us/rest/api/storageservices/list-blobs#uri-parameters) parameter, which helps in listing directories much easier. For example if we have two directories in the container, say `foo/` and `bar/` each of which could have hundreds of thousands of blobs, we would not want to list all of them in the `list_blobs` operations. Instead if you specify `delimiter='/'`, you would only be returned `['foo/', 'bar/']` in the `list_blobs` operation. The new version of the azure storage library seems not to have the `delimiter` option in the `list_blobs` function. Instead there is a function [walk_blobs](https://azuresdkdocs.blob.core.windows.net/$web/python/azure-storage-blob/12.5.0/azure.storage.blob.html#azure.storage.blob.ContainerClient.walk_blobs) which has this parameter. I think that function would be more appropriate here instead of `list_blobs`. I think just replacing `self.client.list_blobs(name_starts_with=dir_path)` with `self.client.walk_blobs(name_starts_with=dir_path, delimiter='/')`, should work.

zarr/storage.py
```diff
@@ -2224,27 +2224,25 @@ def __iter__(self):
...
+        # from azure.storage.blob import Blob
         dir_path = normalize_storage_path(self._append_path_to_prefix(path))
         if dir_path:
             dir_path += '/'
         items = list()
-        for blob in self.client.list_blobs(self.container, prefix=dir_path, delimiter='/'):
-            if type(blob) == Blob:
+        for blob in self.client.list_blobs(name_starts_with=dir_path):
+            # items.append(self._strip_prefix_from_path(blob.name, dir_path))
+            if '/' not in blob.name:  # what is this doing?
```
shikharsg, line 2245: When the list_blobs will be replaced with walk_blobs with a `delimiter='/'` parameter, the return items will be the contents of the directory. So if the directory structure is like so:
```
.zgroup
foo/.zgroup
foo/bar/.zarray
```
then `walk_blobs(self.container, name_starts_with='foo/', delimiter='/')` will return `['foo/.zgroup', 'foo/bar/']`
The `.zgroup` will be a case of the first condition `if '/' not in blob.name`, which if true, means that we have a blob, otherwise it's a directory. And the code in the else block removes the '/' from the directory.
So return value of `listdir('foo')` should be `['.zgroup', 'bar']`.

I admit, it is not immediately obvious. Any refactoring would be greatly appreciated.

jhamman:
@shikharsg - do you happen to understand how we can do the blob emulation with the new version of the api? The current test uses: 

https://github.com/zarr-developers/zarr-python/blob/610db340bf4f2430dbea4f9cab0d2d9c76726967/zarr/tests/test_storage.py#L1775-L1783

As far as I can tell, this option is no longer available in the `azure.storage.blob`.

shikharsg:
> 
> 
> @shikharsg - do you happen to understand how we can do the blob emulation with the new version of the api? The current test uses:
> 
> https://github.com/zarr-developers/zarr-python/blob/610db340bf4f2430dbea4f9cab0d2d9c76726967/zarr/tests/test_storage.py#L1775-L1783
> 
> As far as I can tell, this option is no longer available in the `azure.storage.blob`.

Can't find it either. So the azure storage emulators ([azurite](https://docs.microsoft.com/en-us/azure/storage/common/storage-use-azurite?toc=/azure/storage/blobs/toc.json) or [storage emulator](https://docs.microsoft.com/en-us/azure/storage/common/storage-use-emulator?toc=/azure/storage/blobs/toc.json )) run on a default host and port and have default credentials. The connection string is:
```
DefaultEndpointsProtocol=https;AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;BlobEndpoint=https://127.0.0.1:10000/devstoreaccount1;

Should be something like this I think:

```python
```
```

tjcrone:
I have some time to help move this forward. @jhamman and @shikharsg, what is the primary blocker on this PR? Thanks!

tjcrone:
@alimanfoo, what is your preferred method for me to work on this PR? Should we merge into a fix/absstore branch in this repo 
 that we can continue working on or would you like me to grab this branch from @jhamman and start a new PR? Thanks!

shikharsg:
> I have some time to help move this forward. @jhamman and @shikharsg, what is the primary blocker on this PR? Thanks!

I think main blocker is getting the tests to work with the emulator. Once the tests work it should most likely tell you if something else was breaking as well. Since the new storage library does not have a convenient way to connect to the emulator, we will have to use the default credentials that the emulator runs on, to connect to it. See [this](https://github.com/zarr-developers/zarr-python/pull/620#issuecomment-700300696) comment.

joshmoore:
Merged in master and pushed to kick off the tests again.

joshmoore:
`E       AttributeError: module 'azure.storage.blob' has no attribute 'BlockBlobService'`

jhamman:
Thanks @joshmoore!

I've actually been thinking this we should discuss closing this PR and deprecating the `ABSStore` all together. Lately, I've been using the `adlfs` mapper implementation with increasing success. Given the complexities with maintaining the test suite and interface API compatibility, maybe it would be wise to consolidate  effort elsewhere? Certainly open to other opinions (@tjcrone, @shikharsg) but I'm thinking I'm unlikely to bring this one across the finish line.

joshmoore:
@jhamman: this can now be closed with #759 merged, right? That way #781 (which implements #764) doesn't conflict. But if you are still thinking about refactoring ABSStore further, I can hold off.

jhamman:
Yes! Closing.

Issue #615, opened 2020-09-23
Expand efficient row iteration with possibility to start iterating anywhere


yetyetanotherusername:
As sort of a follow up to https://github.com/zarr-developers/zarr-python/issues/398, would it be possible to expand this useful feature with a way to set the start point for the iterator? This way, when one wants to efficiently iterate over only a subset of the array with a generator, you don't have to walk over the unneeded parts of the array.

yetyetanotherusername:
If i'm not mistaken, this small modification to https://github.com/zarr-developers/zarr-python/issues/398 should already do the trick:

```python
    def islice(self, start=None, end=None):
        if len(self.shape) == 0:
            # Same error as numpy
            raise TypeError("iteration over a 0-d array")
        # Avoid repeatedly decompressing chunks by iterating over the chunks
        if start is None:
            start = 0
        if end is None:
            end = self.shape[0]
        # in the first dimension.
        chunk_size = self.chunks[0]
        for j in range(self.shape[0], start, end):
            if j % chunk_size == 0:
                chunk = self[j: j + chunk_size]
            yield chunk[j % chunk_size]
```

yetyetanotherusername:
I've gone ahead and implemented this in a fork together with a test. I can still make a pull request, but would like some confirmation that it has a chance of getting accepted before I proceed.

jeromekelleher:
It looks good to me @yetyetanotherusername, I'm happy to review the PR. Seems like a fairly uncontroversial addition, but I'm not one of the core devs (who may disagree).

yetyetanotherusername:
Thanks @jeromekelleher! I've opened a [pull request](https://github.com/zarr-developers/zarr-python/pull/621). I still have some stuff to do regarding documentation and I'm a bit unhappy about the tests for `__iter__` and `islice` partially duplicating each other. For starters I left the original test unchanged since I wanted to make extra sure I don't break anything. 

I'm not intimately familiar with the rest of the `zarr` interface and its error handling, so if you feel that something should be caught that isn't right now, I'd be very happy about any input. For example, I don't support reverse iteration right now, so probably trying it should raise a `NotImplementedError`.

yetyetanotherusername:
Closed by https://github.com/zarr-developers/zarr-python/pull/621

PR #621, opened 2020-09-28
Start stop for iterator

yetyetanotherusername:
Expand the array iteration block caching with the option to set a start and end index for the generator. This speeds up use cases where one only wants to iterate over part of an array, as previously the generator would have to iterate over the unused parts of the array as well.

This adds a new function `islice` to the array class.

TODO:


Carreau:
Thanks ! It seem the email you use is not associated with your github username, I'm not sure if you care about his or not.

yetyetanotherusername:
Fixed, thanks!

zarr/core.py
```diff
@@ -423,18 +423,31 @@ def __array__(self, *args):
             a = a.astype(args[0])
         return a
 
-    def __iter__(self):
+    def islice(self, start=None, end=None):
```
jeromekelleher, line 426: There should probably be some range checking on ``start`` and ``end`` here. I guess we should follow the interface of ``itertools.islice`` as closely as possible and return the same errors they do on different out-of-range conditions.

jeromekelleher (commented):
Looks good to me. I think we just need some documentation and clarity about what happens in edge cases, following the behaviour of itertools.islice.

zarr/core.py
yetyetanotherusername: I played around a little with `islice`. Writing this down here so i don't forget:

- Setting a bound exceeding the array bound silently returns the array without warning or error.
- Bounds must be non negative integers (or None), otherwise a ValueError is raised.
- end > start returns [], reverse iteration is not supported at all.

I think that should cover it, or am I forgetting something?

yetyetanotherusername:
I added a docstring to `islice` and added a missing edge case to the tests. Regarding documentation, since this is a minor change, are any changes in the `.rst` files needed? Does anything else from the TODO-list apply here?

Otherwise I think this is ready for review.

zarr/core.py
```diff
@@ -424,6 +424,36 @@ def __array__(self, *args):
         return a
 
     def islice(self, start=None, end=None):
+        """
+        Yield a generator for iterating over the entire or parts of the
+        array.
```
jeromekelleher, line 429: It's probably worthwhile saying why this is useful and more efficient than iterating directly over a slice.

yetyetanotherusername:
Since `islice` is a public function, I also added an item to the API docs and a note to release.rst.

yetyetanotherusername:
Just to be sure, is there anything missing on my part?

jeromekelleher (commented):
LGTM, thanks @yetyetanotherusername!

jeromekelleher:
> Just to be sure, is there anything missing on my part?

It all looks good to me, I think we just need a review from one of the maintainers. @Carreau, any chance you could take a look please, or maybe recommend an appropriate reviewer?

yetyetanotherusername:
> LGTM, thanks @yetyetanotherusername!

Thanks for your input!

yetyetanotherusername:
Interesting, i can't for the life of me reproduce the mypy error that got the python3.7 pipeline to fail.

joshmoore:
Merged master where tests are currently green. Let's see how this goes. Thanks for your patience!

joshmoore:
The PR is green again! Any @zarr-developers/core-devs have thoughts?

jeromekelleher:
+1 from me - seems like a useful feature.

joshmoore:
Treating @jeromekelleher's comment as a review.

yetyetanotherusername:
Thanks! This closes https://github.com/zarr-developers/zarr-python/issues/615

PR #627, opened 2020-10-06
Remove uncovered lines in test.

Carreau:
this is part of the goal to go back to 100% coverage.

### DIFF
--- a/zarr/tests/test_core.py
+++ b/zarr/tests/test_core.py
@@ -139,9 +139,6 @@ class TestArray(unittest.TestCase):
         except TypeError:
             pass
 
-        if hasattr(z.store, 'close'):
-            z.store.close()
-
     # noinspection PyStatementEffect
     def test_array_1d(self):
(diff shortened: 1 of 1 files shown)


jakirkham:
This also shows up in some lines above. Should we remove those as well?

I wonder if we should just be using a context manager instead of this open/close code. Not sure how easy that will be to do though.

Carreau:
That's, as far as I can tell the only line uncovered here. 
Hey I think a context manager would make sens, but my guess is making the context manager make more sens as part of the PR that try to make a base store.

jakirkham:
Yeah it's just odd we don't see the same coverage issue with say `test_store_has_bytes_values`, which also does this same thing.

Carreau:
> Yeah it's just odd we don't see the same coverage issue with say `test_store_has_bytes_values`, which also does this same thing.

No, that's because this one is not overwritten in subclasses, so it _is_ tested with stores having `close` method.

PR #636, opened 2020-10-22
Checking that the store is an instance of dict seem incorrect.

Carreau:
Indeed MemoryStore is an instance of mutable mapping that have the same
property as dict, but would not be seen as as instance of dict.

With some caveat inlines.

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -797,12 +797,14 @@ Here is an example using S3Map to read an array created previously::
     Compressor         : Blosc(cname='lz4', clevel=5, shuffle=SHUFFLE, blocksize=0)
     Store type         : fsspec.mapping.FSMap
     No. bytes          : 21
+    No. bytes stored   : 382
+    Storage ratio      : 0.1
     Chunks initialized : 3/3
     >>> z[:]
     array([b'H', b'e', b'l', b'l', b'o', b' ', b'f', b'r', b'o', b'm', b' ',
            b't', b'h', b'e', b' ', b'c', b'l', b'o', b'u', b'd', b'!'],
           dtype='|S1')
-    >>> z[:].tostring()
+    >>> z[:].tobytes()
     b'Hello from the cloud!'
 
 Zarr now also has a builtin storage backend for Azure Blob Storage.
@@ -841,11 +843,11 @@ store. E.g.::
     >>> z = root['foo/bar/baz']
     >>> from timeit import timeit
     >>> # first data access is relatively slow, retrieved from store
-    ... timeit('print(z[:].tostring())', number=1, globals=globals())  # doctest: +SKIP
+    ... timeit('print(z[:].tobytes())', number=1, globals=globals())  # doctest: +SKIP
     b'Hello from the cloud!'
     0.1081731989979744
     >>> # second data access is faster, uses cache
-    ... timeit('print(z[:].tostring())', number=1, globals=globals())  # doctest: +SKIP
+    ... timeit('print(z[:].tobytes())', number=1, globals=globals())  # doctest: +SKIP
     b'Hello from the cloud!'
     0.0009490990014455747
 
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -797,8 +797,6 @@ Here is an example using S3Map to read an array created previously::
     Compressor         : Blosc(cname='lz4', clevel=5, shuffle=SHUFFLE, blocksize=0)
     Store type         : fsspec.mapping.FSMap
     No. bytes          : 21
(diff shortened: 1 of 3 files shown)


zarr/core.py
```diff
@@ -1821,7 +1821,7 @@ def _encode_chunk(self, chunk):
             cdata = chunk
 
         # ensure in-memory data is immutable and easy to compare
-        if isinstance(self.chunk_store, dict):
+        if isinstance(self.chunk_store, MutableMapping):
             cdata = ensure_bytes(cdata)
```
Carreau, line 1825: I'm not sure why we don't want to always do that; there is the question of the Consolidated metadata one where we might not want to serialize all, and this might also trigger ensure_bytes  for more stores. I came across this why trying to make a base class for all stores and have store for example always have a `close()` method, so I wrapped `dict()` in a memorystore-like object and test were failing.

zarr/storage.py
```diff
@@ -176,7 +176,7 @@ def getsize(store, path: Path = None) -> int:
     if hasattr(store, 'getsize'):
         # pass through
         return store.getsize(path)
-    elif isinstance(store, dict):
```
Carreau, line 179: I don't see why we should limit ourselves to dict here, it shoudl work with any mutable mappings.

zarr/core.py
jakirkham: This is only intended for in-memory stores. In that case we coerce to `bytes` because they are immutable.

We don't want to do this generally as this incurs a copy and if we are just going to write to disk or send over the wire, there is no need.

zarr/storage.py
jakirkham: Yeah this seems fine.

zarr/core.py
Carreau: I understand that, the problem is  tell stores implementations they may not receive bytes, and have to deal with it, which is weird as now the stores `__getitem__` might not return the same as `__setitem__`. But fine let's assume we are ok with this. 

The following (existing) test fails for MutableMapping that are not dict, and do not call something  like `ensure_bytes`.

```
    def test_store_has_bytes_values(self):
        # Test that many stores do hold bytes values.
        # Though this is not a strict requirement.
        # Should be disabled by any stores that fail this as needed.

        # Initialize array
        np.random.seed(42)
        z = self.create_array(shape=(1050,), chunks=100, dtype='f8', compressor=[])
        z[:] = np.random.random(z.shape)

        # Check in-memory array only contains `bytes`
        assert all([isinstance(v, bytes) for v in z.chunk_store.values()])

        if hasattr(z.store, 'close'):
            z.store.close()
```

And note that the name of the test is xxxx_store_xxxx, not xxx_dict_xxxx

So IMHO either we:
 - stick to stores have bytes interfaces, 
 - relax stores interface and tell them "you may get things that are not bytes", deal with it.
 - have a proper way to have a store tell "I'm ok not receiving bytes". 

Otherwise many of the tests rely on all stores having all the same behavior but testing only with `dict()` which is special cased.

Regardless I'll revert that and we can move to to another issue.

docs/tutorial.rst
```diff
@@ -797,12 +797,14 @@ Here is an example using S3Map to read an array created previously::
     Compressor         : Blosc(cname='lz4', clevel=5, shuffle=SHUFFLE, blocksize=0)
     Store type         : fsspec.mapping.FSMap
     No. bytes          : 21
+    No. bytes stored   : 382
+    Storage ratio      : 0.1
```
Carreau, line 801:
```suggestion
```

zarr/core.py
```diff
@@ -1955,7 +1956,7 @@ def _encode_chunk(self, chunk):
             cdata = chunk
 
         # ensure in-memory data is immutable and easy to compare
-        if isinstance(self.chunk_store, dict):
+        if isinstance(self.chunk_store, MutableMapping):
```
jakirkham, line 1959: As mentioned upthread, this change is too broad and may force unnecessary copies. The goal of this check was to make sure there would not be unexpected mutation of in-memory data so would copy to `bytes` before storing. If we are writing to disk or sending to a cloud store, this is unneeded. Perhaps the check needs to be updated, but maybe that can be handled as part of Greg's work?

jakirkham:
Have cherry-picked the doc fixes from here into PR ( https://github.com/zarr-developers/zarr-python/pull/907 )

joshmoore:
@grlee77 : any thoughts on what to do here?

jakirkham:
Guessing Greg has likely handled this along with other things in his PRs. Though could be wrong about that.

joshmoore:
I think the same is true. Closing in favor of #898 and friends.

PR #637, opened 2020-10-22
Test on Python 3.9-dev, also remove sudo

Carreau:
3.9 have been release a few weeks ago, and sudo seem unnecessay.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,5 @@
 language: python
 
-sudo: false
-
 addons:
   apt:
     packages:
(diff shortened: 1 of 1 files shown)


Carreau:
Seem to fail because of h5py.

Carreau:
libh5py does not seem to work yet on 3.9

jakirkham:
Would xfailing some tests on Python 3.9 help?

Carreau:
> Would xfailing some tests on Python 3.9 help?

Yeah, likely or maker them skipif if we can't import h5py. Right now it's more a question on conditionally trying to install h5py as CI fails before starting the tests.

jakirkham:
Guessing this will need an update to move to GH Actions

joshmoore:
Yeah, actually, looking at it, I'm not sure it's worth keeping it around. Closing. If anything is salvageable, let's do that in a new PR.

PR #639, opened 2020-10-22
Turn UserWarnings into errors

Carreau:
And turn then back into only warnings into relevant tests.
This ensure that we are not using deprecated functionalities, except
when testing for it.


joshmoore:
```
  D:\a\zarr-python\zarr-python\zarr\n5.py:417: RuntimeWarning: Not all N5 implementations support blosc compression (yet). You might not be able to open the dataset with another N5 library.
    warnings.warn(
```

```
c:\miniconda\envs\zarr-env\lib\site-packages\ipytree\tree.py:43
  c:\miniconda\envs\zarr-env\lib\site-packages\ipytree\tree.py:43: DeprecationWarning: metadata {'trait': <traitlets.traitlets.Instance object at 0x000001ED3D3C4D60>} was set from the constructor. With traitlets 4.1, metadata should be set using the .tag() method, e.g., Int().tag(key1='value1', key2='value2')
    nodes = Tuple(trait=Instance(Widget)).tag(
```

joshmoore:
Looks like this was already rolled in elsewhere.

PR #641, opened 2020-10-28
Remove 3.5 configuration from tox, run pyflakes on ghaction.

Carreau:
This will run way faster than travis and should allow to distinguish
actual test failure from formatting failures in PRs if we deactivate
flake8 from tox.


joshmoore:
Re-opening to trigger GHA.

jakirkham (approved):

PR #643, opened 2020-11-02
Mute a bit codecov report.

Carreau:
Codecov report are a bit noisy, this mute them a bit y removing the
large graph and postigin only if there is a coverage change.


mzjp2 (approved):
<3

Issue #644, opened 2020-11-03
FSStore getitems fails with arrays that have a 0 length shape dimension


andrewfulton9:
```python
from tempfile import mkdtemp
from zarr.core import Array
from zarr.storage import FSStore, init_array

path = mkdtemp()
store = FSStore(path)
init_array(store, shape=(1, 0), chunks=(1, 1), dtype=int))
arr = Array(store=store)
arr[:]
```
fails with:
```
ValueError                                Traceback (most recent call last)
<ipython-input-11-47ed00660ce3> in <module>
----> 1 arr[:]

~/Dropbox/quansight/zarr/zarr-python/zarr/core.py in __getitem__(self, selection)
    569 
    570         fields, selection = pop_fields(selection)
--> 571         return self.get_basic_selection(selection, fields=fields)
    572 
    573     def get_basic_selection(self, selection=Ellipsis, out=None, fields=None):

~/Dropbox/quansight/zarr/zarr-python/zarr/core.py in get_basic_selection(self, selection, out, fields)
    695         else:
    696             return self._get_basic_selection_nd(selection=selection, out=out,
--> 697                                                 fields=fields)
    698 
    699     def _get_basic_selection_zd(self, selection, out=None, fields=None):

~/Dropbox/quansight/zarr/zarr-python/zarr/core.py in _get_basic_selection_nd(self, selection, out, fields)
    737         indexer = BasicIndexer(selection, self)
    738 
--> 739         return self._get_selection(indexer=indexer, out=out, fields=fields)
    740 
    741     def get_orthogonal_selection(self, selection, out=None, fields=None):

~/Dropbox/quansight/zarr/zarr-python/zarr/core.py in _get_selection(self, indexer, out, fields)
   1030         else:
   1031             # allow storage to get multiple items at once
-> 1032             lchunk_coords, lchunk_selection, lout_selection = zip(*indexer)
   1033             self._chunk_getitems(lchunk_coords, lchunk_selection, out, lout_selection,
   1034                                  drop_axes=indexer.drop_axes, fields=fields)

ValueError: not enough values to unpack (expected 3, got 0
```

#### Problem description

If an array using FSStore has a 0 length dimension, Value error is raised because `zip(*indexer)` [here](https://github.com/zarr-developers/zarr-python/blob/480d21adc74db4d5d15abe69d64b3ba871301b97/zarr/core.py#L1032) is empty 

#### Version and installation information

[...]

PR #645, opened 2020-11-03
FSStore indexer error fix

andrewfulton9:
[Description of PR]

Adds tests for `Array` set up with `FSStore` and fixes the bug in  #644 

TODO:

### DIFF
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -1113,7 +1113,9 @@ class FSStore(MutableMapping):
             raise ReadOnlyError()
         store_path = self.dir_path(path)
         if self.fs.isdir(store_path):
-            self.fs.rm(store_path, recursive=True)
+            # sometimes doesn't delete on first try with S3
+            while self.fs.exists(store_path):
+                self.fs.rm(store_path, recursive=True)
 
     def getsize(self, path=None):
         store_path = self.dir_path(path)
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -10,6 +10,9 @@ This release will be the first release of Zarr not supporting Python 3.5.
 * End Python 3.5 support.
   By :user:`Chris Barnes <clbarnes>`; :issue:`602`.
 
+* Add `Array` tests for FSStore.
+  By :user:`Andrew Fulton <andrewfulton9>`; :issue: ``.
+
 2.5.0
 -----
 
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -11,7 +11,7 @@ This release will be the first release of Zarr not supporting Python 3.5.
   By :user:`Chris Barnes <clbarnes>`; :issue:`602`.
 
 * Add `Array` tests for FSStore.
-  By :user:`Andrew Fulton <andrewfulton9>`; :issue: ``.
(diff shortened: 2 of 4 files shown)


andrewfulton9:
Looks like `_set_selection` has the same bug as `_get_selection`. I'll go ahead and fix that as well and add that to the PR.

andrewfulton9:
I am also having trouble understanding why `test_pickle` is failing. The raised error is that `FSMap` object has not attribute 'missing_exceptions', but I can't figure out why it wouldn't. I tried upgrading fsspec to 0.8.4 and that allows that test fo pass, but causes `test_storage.py::TestFSStore::test_s3` and `test_storage.py::TestFSStore::test_s3_complex` to fail. I assume to failure is due to the fixture, but I am not positive about that. Would it be prefered to upgrade fsspec and try to fix the s3 tests or try to figure out a way to fix `FSMap` issue? in the currently pinned version?

Carreau:
From discussion with Andrew, the failure is/was dues to FSStore 0.8.3 not having self.missing_exceptions in the reduced; which was fixed in 0.8.4.

zarr/storage.py
```diff
@@ -1113,7 +1113,9 @@ def rmdir(self, path=None):
             raise ReadOnlyError()
         store_path = self.dir_path(path)
         if self.fs.isdir(store_path):
-            self.fs.rm(store_path, recursive=True)
+            # sometimes doesn't delete on first try with S3
+            while self.fs.exists(store_path):
+                self.fs.rm(store_path, recursive=True)
```
Carreau, line 1118: We may want to have exponential backoff with a timeout. But that can be done later in a subsequent PR.

Carreau (approved):

zarr/storage.py
martindurant: This shouldn't be required - did you figure out what it going on? Perhaps just time for the metadata to sync?

andrewfulton9: @martindurant, I tried a `time.sleep` for various times (5, 10, 30 seconds) and that didn't seem to work. This is the only way I could figure out how to get it to pass without changing [test_s3_complex](https://github.com/zarr-developers/zarr-python/blob/06450d038aa4bd5502e1a72eee824c1d5f409494/zarr/tests/test_storage.py#L958-L999). Another way that I could get it to pass was by running :
```python
a = g.create_dataset("data", shape=(8, 8, 8), mode='w',
                             fill_value=-1, chunks=(1, 1, 1), overwrite=True)
a = g.create_dataset("data", shape=(8, 8, 8), mode='w',
                             fill_value=-1, chunks=(1, 1, 1), overwrite=True)
```
in place of the code [here](https://github.com/zarr-developers/zarr-python/blob/06450d038aa4bd5502e1a72eee824c1d5f409494/zarr/tests/test_storage.py#L997) and that also seemed to resolve the problem. For what it's worth, this test does't fail if fsspec is set to `0.8.3`, ony when its set to `0.8.4`, although at `0.8.3`, [test_pickle](https://github.com/zarr-developers/zarr-python/blob/06450d038aa4bd5502e1a72eee824c1d5f409494/zarr/tests/test_core.py#L759-L796) will fail.

PR #650, opened 2020-11-11
Fixes #649

### LINKED ISSUE #649: FileNotFoundError when using zarr with dask/s3fs version >= 0.5.0
When using a zarr store  from an s3 bucket with not storing physically chunks which are in uninitialized state, as described in the specification of zarr (https://zarr.readthedocs.io/en/stable/spec/v2.html#chunks), a `FileNotFoundError` occurs. This is new since `s3fs version >= 0.5.0`. 

```python
import s3fs
import zarr

# AWS S3 path
s3_path = 's3://dcs4cop/bc-cmems-spm-1997-2018_1x704x640.zarr' 
# Initilize the S3 file system
s3 = s3fs.S3FileSystem(anon=True)
store = s3fs.S3Map(root=s3_path, s3=s3, check=False)
# Read Zarr file
ds = zarr.open(store=store, mode='r')

# Get some values:
ds.SPM[5]
```
This results in `FileNotFoundError: The specified key does not exist.`

Please provide the following:

* Value of ``zarr.__version__`` 2.5.0
* Value of ``s3fs.__version__`` 0.5.1
* Value of ``fsspec.__version__`` 0.8.0
* Version of Python interpreter 3.7.6
* Operating system Linux
* Zarr was installed "using conda"

### Working as expected with the following versions: 
* Value of ``zarr.__version__`` 2.5.0
* Value of ``s3fs.__version__`` 0.4.0.
* Value of ``fsspec.__version__`` 0.6.2

image

martindurant:
[Description of PR]

TODO:
* [ ] ~Changes documented in docs/release.rst~

### DIFF
--- a/zarr/convenience.py
+++ b/zarr/convenience.py
@@ -73,6 +73,8 @@ def open(store=None, mode='a', **kwargs):
     path = kwargs.get('path', None)
     # handle polymorphic store arg
     clobber = mode == 'w'
+    # we pass storage options explicitly, since normalize_store_arg might construct
+    # a store if the input is a fsspec-compatible URL
     store = normalize_store_arg(store, clobber=clobber,
                                 storage_options=kwargs.pop("storage_options", {}))
     path = normalize_storage_path(path)
--- a/zarr/convenience.py
+++ b/zarr/convenience.py
@@ -73,7 +73,8 @@ def open(store=None, mode='a', **kwargs):
     path = kwargs.get('path', None)
     # handle polymorphic store arg
     clobber = mode == 'w'
-    store = normalize_store_arg(store, clobber=clobber)
+    store = normalize_store_arg(store, clobber=clobber,
+                                storage_options=kwargs.pop("storage_options", {}))
     path = normalize_storage_path(path)
 
     if mode in {'w', 'w-', 'x'}:
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -1038,7 +1038,7 @@ class FSStore(MutableMapping):
             key = '/'.join(bits + [end.replace('.', self.key_separator)])
         return key.lower() if self.normalize_keys else key
 
-    def getitems(self, keys):
+    def getitems(self, keys, **kwargs):
         keys = [self._normalize_key(key) for key in keys]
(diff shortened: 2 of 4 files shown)


zarr/convenience.py
```diff
@@ -73,7 +73,8 @@ def open(store=None, mode='a', **kwargs):
     path = kwargs.get('path', None)
     # handle polymorphic store arg
     clobber = mode == 'w'
-    store = normalize_store_arg(store, clobber=clobber)
+    store = normalize_store_arg(store, clobber=clobber,
+                                storage_options=kwargs.pop("storage_options", {}))
```
rabernat, line 77: I would appreciate a comment here so that the other devs understand why were are treating this argument specially.

zarr/storage.py
```diff
@@ -1038,7 +1038,7 @@ def _normalize_key(self, key):
             key = '/'.join(bits + [end.replace('.', self.key_separator)])
         return key.lower() if self.normalize_keys else key
 
-    def getitems(self, keys):
+    def getitems(self, keys, **kwargs):
```
rabernat, line 1041: To me it looks like these `**kwargs` are never used. The `on_error="omit"` passed here by `_chunk_getitems` will not be propagated, and instead `on_error="omit"` is hard coded below in the call to `self.map.getitems`.

rabernat (commented):
I'm finding the call sequence a bit hard to follow here.

zarr/storage.py
martindurant: Correct, but the caller now passes `on_error="omit"` to getitems on any mapping. It would make sense to pass kwargs through to the mapping method, but I would argue that anything other than "omit" doesn't make sense here.

zarr/convenience.py
martindurant: done

Carreau (approved):

jakirkham:
@rabernat, would you have time to take another look? No worries if not 🙂

rabernat (approved):
Fantastic, thanks @martindurant!

martindurant:
Merging as this already passed CI earlier, but didn't start a new build for some reason (there were no code changes).

PR #654, opened 2020-11-18
Mode is not one of the kwargs create_dataset takes

Carreau:
It triggers a warnings that make the test-suite fails on strict mode.


joshmoore:
Not sure why appveyor is still triggering. Happy to ignore it. I'm not sure if the assert is the most informative, but I've run into this issue before as well.

If there are no objections, I'd tend to get this in.

joshmoore:
Closing. If the warnings appear again, I'll resurrect elsewhere.

PR #666, opened 2020-12-01
Add guard as to version number;

Carreau:
It seem like setuptools scm can screw up and this should ensure we catch
that in CI before making the wheel.

### DIFF
--- a/.github/workflows/releases.yml
+++ b/.github/workflows/releases.yml
@@ -21,11 +21,19 @@ jobs:
 
       - name: Install PyBuild
         run: |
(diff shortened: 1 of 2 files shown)


jakirkham (approved):

Carreau:
Found https://stackoverflow.com/questions/62968271/python-automatic-versioning-not-happening-while-running-in-github-actions as well;

Carreau:
Ok, downgrading build to 0.4 seem to work for now, I'm going to merge and tag a 2.6.1, and work on a proper fix later.

Issue #59, opened 2016-09-06
TempStore


alimanfoo:
Would be useful to have a TempStore extending DirectoryStore but using a temporary directory.

PR #667, opened 2020-12-01
Partial read

andrewfulton9:
This PR adds the capability to partially read and decompress chunks in an arrays that are initialized with blosc decompression and an fsspec based backend. Its related to [zarr-specs #59](https://github.com/zarr-developers/zarr-specs/issues/59)

I did some initial benchmarking showing processing time for indexing with various array and chunk shapes filled with random numbers stored on s3:

array_shape=(1000, 1000), chunk_shape=(100, 500)
-------------------------------------------------------------------------

image

array_shape=(10000, 10000), chunk_shape=(1000, 1000)
-------------------------------------------------------------------------------------
image

array_shape=(10000, 10000), chunk_shape=(2000, 5000)
----------------------------------------------------------------------------------
image

array_shape=(10000, 10000), chunk_shape=(5000, 5000)
-------------------------------------------------------------------------------
image

TODO:

### DIFF
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -30,6 +30,18 @@ This release od Zarr Python is is the first release of Zarr to not supporting Py
 See `this link <https://github.com/zarr-developers/zarr-python/milestone/11?closed=1>` for the full list of closed and
 merged PR tagged with the 2.6 milestone.
 
+* Add ability to partially read and decompress arrays, see :issue:`667`. It is
+  only available to chunks stored using fs-spec and using bloc as a compressor.
+
+  For certain analysis case when only a small portion of chunks is needed it can
+  be advantageous to only access and decompress part of the chunks. Doing
+  partial read and decompression add high latency to many of the operation so
+  should be used only when the subset of the data is small compared to the full
+  chunks and is stored contiguously (that is to say either last dimensions for C
+  layout, firsts for F). Pass ``partial_decompress=True`` as argument when
+  creating an ``Array``, or when using ``open_array``. No option exists yet to
+  apply partial read and decompress on a per-operation basis.
+
 2.5.0
 -----
 
--- a/zarr/core.py
+++ b/zarr/core.py
@@ -1147,7 +1147,7 @@ class Array:
 
         See Also
         --------
-        basic_selection, set_basic_selection, get_mask_selection, set_mask_selection,
+        get_basic_selection, set_basic_selection, get_mask_selection, set_mask_selection,
         get_coordinate_selection, set_coordinate_selection, get_orthogonal_selection,
         set_orthogonal_selection, vindex, oindex, __getitem__
 
--- a/zarr/core.py
+++ b/zarr/core.py
@@ -11,22 +11,41 @@ from numcodecs.compat import ensure_bytes, ensure_ndarray
 
 from zarr.attrs import Attributes
 from zarr.codecs import AsType, get_codec
-from zarr.errors import ArrayNotFoundError, ReadOnlyError
-from zarr.indexing import (BasicIndexer, CoordinateIndexer, MaskIndexer,
-                           OIndex, OrthogonalIndexer, VIndex, check_fields,
-                           check_no_multi_fields, ensure_tuple,
-                           err_too_many_indices, is_contiguous_selection,
-                           is_scalar, pop_fields)
+from zarr.errors import ArrayNotFoundError, ReadOnlyError, ArrayIndexError
+from zarr.indexing import (
+    BasicIndexer,
+    CoordinateIndexer,
+    MaskIndexer,
+    OIndex,
+    OrthogonalIndexer,
+    VIndex,
+    PartialChunkIterator,
+    check_fields,
+    check_no_multi_fields,
+    ensure_tuple,
+    err_too_many_indices,
+    is_contiguous_selection,
+    is_scalar,
+    pop_fields,
+)
 from zarr.meta import decode_array_metadata, encode_array_metadata
 from zarr.storage import array_meta_key, attrs_key, getsize, listdir
-from zarr.util import (InfoReporter, check_array_shape, human_readable_size,
-                       is_total_slice, nolock, normalize_chunks,
-                       normalize_resize_args, normalize_shape,
-                       normalize_storage_path)
+from zarr.util import (
+    InfoReporter,
+    check_array_shape,
+    human_readable_size,
+    is_total_slice,
+    nolock,
+    normalize_chunks,
+    normalize_resize_args,
+    normalize_shape,
+    normalize_storage_path,
+    PartialReadBuffer,
+)
 
 
 # noinspection PyUnresolvedReferences
-class Array(object):
+class Array:
     """Instantiate an array from an initialized store.
 
     Parameters
@@ -51,6 +70,12 @@ class Array(object):
         If True (default), user attributes will be cached for attribute read
         operations. If False, user attributes are reloaded from the store prior
         to all attribute read operations.
+    partial_decompress : bool, optional
+        If True and while the chunk_store is a FSStore and the compresion used
+        is Blosc, when getting data from the array chunks will be partially
+        read and decompressed when possible.
+
+        .. versionadded:: 2.7
 
     Attributes
     ----------
@@ -102,8 +127,17 @@ class Array(object):
 
     """
 
-    def __init__(self, store, path=None, read_only=False, chunk_store=None,
-                 synchronizer=None, cache_metadata=True, cache_attrs=True):
+    def __init__(
+        self,
+        store,
+        path=None,
+        read_only=False,
+        chunk_store=None,
+        synchronizer=None,
+        cache_metadata=True,
+        cache_attrs=True,
+        partial_decompress=False,
+    ):
         # N.B., expect at this point store is fully initialized with all
         # configuration metadata fully specified and normalized
 
@@ -118,6 +152,7 @@ class Array(object):
         self._synchronizer = synchronizer
         self._cache_metadata = cache_metadata
         self._is_view = False
+        self._partial_decompress = partial_decompress
 
         # initialize metadata
         self._load_metadata()
@@ -1112,7 +1147,7 @@ class Array(object):
 
         See Also
         --------
-        get_basic_selection, set_basic_selection, get_mask_selection, set_mask_selection,
+        basic_selection, set_basic_selection, get_mask_selection, set_mask_selection,
         get_coordinate_selection, set_coordinate_selection, get_orthogonal_selection,
         set_orthogonal_selection, vindex, oindex, __getitem__
 
@@ -1580,8 +1615,17 @@ class Array(object):
             self._chunk_setitems(lchunk_coords, lchunk_selection, chunk_values,
                                  fields=fields)
 
-    def _process_chunk(self, out, cdata, chunk_selection, drop_axes,
-                       out_is_ndarray, fields, out_selection):
+    def _process_chunk(
+        self,
+        out,
+        cdata,
+        chunk_selection,
+        drop_axes,
+        out_is_ndarray,
+        fields,
+        out_selection,
+        partial_read_decode=False,
+    ):
         """Take binary data from storage and fill output array"""
         if (out_is_ndarray and
                 not fields and
@@ -1604,8 +1648,9 @@ class Array(object):
                 # optimization: we want the whole chunk, and the destination is
                 # contiguous, so we can decompress directly from the chunk
                 # into the destination array
-
                 if self._compressor:
+                    if isinstance(cdata, PartialReadBuffer):
+                        cdata = cdata.read_full()
                     self._compressor.decode(cdata, dest)
                 else:
                     chunk = ensure_ndarray(cdata).view(self._dtype)
@@ -1614,6 +1659,33 @@ class Array(object):
                 return
 
         # decode chunk
+        try:
+            if partial_read_decode:
+                cdata.prepare_chunk()
+                # size of chunk
+                tmp = np.empty(self._chunks, dtype=self.dtype)
+                index_selection = PartialChunkIterator(chunk_selection, self.chunks)
+                for start, nitems, partial_out_selection in index_selection:
+                    expected_shape = [
+                        len(
+                            range(*partial_out_selection[i].indices(self.chunks[0] + 1))
+                        )
+                        if i < len(partial_out_selection)
+                        else dim
+                        for i, dim in enumerate(self.chunks)
+                    ]
+                    cdata.read_part(start, nitems)
+                    chunk_partial = self._decode_chunk(
+                        cdata.buff,
+                        start=start,
+                        nitems=nitems,
+                        expected_shape=expected_shape,
+                    )
+                    tmp[partial_out_selection] = chunk_partial
+                out[out_selection] = tmp[chunk_selection]
+                return
+        except ArrayIndexError:
+            cdata = cdata.read_full()
         chunk = self._decode_chunk(cdata)
 
         # select data from chunk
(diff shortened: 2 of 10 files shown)


Carreau:
I added some release notes (that are conflicting), removed some ded code (see 745248281fced5cfb96012d42d338eab136fe1ec), and try to find some test code to increase coverage.


```
z = self.create_array(shape=(500,500, 500), chunks=(50, 50, 50), dtype='<i4')
z[:,:,:] = 0
z[0,:,0].any() # fails
```

Fails with 

```
zarr/core.py:578: in __getitem__
    return self.get_basic_selection(selection, fields=fields)
zarr/core.py:703: in get_basic_selection
    return self._get_basic_selection_nd(selection=selection, out=out,
zarr/core.py:746: in _get_basic_selection_nd
zarr/core.py:1041: in _get_selection
    self._chunk_getitems(lchunk_coords, lchunk_selection, out, lout_selection,
zarr/core.py:1731: in _chunk_getitems
    self._process_chunk(out, cdatas[ckey], chunk_select, drop_axes,
zarr/core.py:1638: in _process_chunk
    chunk_partial = self._decode_chunk(
zarr/core.py:1851: in _decode_chunk
    chunk = self._compressor.decode(cdata)
numcodecs/blosc.pyx:495: in numcodecs.blosc.Blosc.decode
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

>   ???
E   RuntimeError: error during blosc decompression: -1

numcodecs/blosc.pyx:399: RuntimeError
```

I'll try to dive into this.

I'm wondering if it's not `start_points_buffer = self.fs.read_block(self.key_path, 16, int(self.nblocks*4))` where int rounds down instead of up. Running in a meeting will look at that later.

Carreau:
I'm going to rebase because of conflicts, I may end up squashing commits for simplicity of rebasing. current commit is/was 3ae237f0b292d4855be0a15306ef7ad7b2e40d8e

zarr/indexing.py
```diff
@@ -825,3 +826,94 @@ def pop_fields(selection):
...
+        # number of selection dimensions can't be greater than the number of chunk dimensions
+        if len(self.selection) > len(self.arr_shape):
+            raise ValueError(
+                "Selection has more dimensions then the array:\n"
+                "selection dimensions = {len(self.selection)\n"
+                "array dimensions = {len(self.arr_shape)}"
+            )
+
+        # any selection can not be out of the range of the chunk
+        self.selection_shape = np.empty(self.arr_shape)[self.selection].shape
```
Carreau, line 879: `self.selection_shape` size is determined by `self.selection`, which later on is muted with `self.selection.pop()`; which might be a source of bug; though it might be ok as self.selection_shape seem to be unused....

Carreau:
This raises another error

```
from zarr.storage import (
    FSStore,
)

from zarr.storage import (
    init_array,
)
from zarr.core import Array
import atexit
import shutil
from tempfile import mkdtemp, mktemp
from numcodecs import Blosc

path = './tmp-test'
path = mkdtemp()
atexit.register(shutil.rmtree, path)
store = FSStore(path)
init_array(store, compressor=Blosc(), shape=(50, 50, 50), chunks=(5, 5, 5), dtype="<i4")
z = Array(
    store,
    cache_metadata=True,
    cache_attrs=True,
    partial_decompress=True,
)

z[:, :, :] = 0
assert z[:, 0, :].any()
```


```
  File "foo.py", line 31, in <module>
    assert z[:, 0, :].any()
  File "/Users/bussonniermatthias/dev/zarr-python/zarr/core.py", line 604, in __getitem__
    return self.get_basic_selection(selection, fields=fields)
  File "/Users/bussonniermatthias/dev/zarr-python/zarr/core.py", line 729, in get_basic_selection
    return self._get_basic_selection_nd(selection=selection, out=out,
  File "/Users/bussonniermatthias/dev/zarr-python/zarr/core.py", line 772, in _get_basic_selection_nd
  File "/Users/bussonniermatthias/dev/zarr-python/zarr/core.py", line 1067, in _get_selection
    self._chunk_getitems(lchunk_coords, lchunk_selection, out, lout_selection,
  File "/Users/bussonniermatthias/dev/zarr-python/zarr/core.py", line 1781, in _chunk_getitems
    self._process_chunk(
  File "/Users/bussonniermatthias/dev/zarr-python/zarr/core.py", line 1677, in _process_chunk
    chunk_partial = self._decode_chunk(
  File "/Users/bussonniermatthias/dev/zarr-python/zarr/core.py", line 1933, in _decode_chunk
    chunk = chunk.reshape(expected_shape or self._chunks, order=self._order)
ValueError: cannot reshape array of size 125 into shape (1,1,5)
```

I've also found some simplification of instance variables that can be local variable I'll push that.

Carreau:
this second failure seem to be due to me having older numcodec that does not have decode partial for blosc;but still assuming it only partially decoded.

Carreau:
So both failure are due to me having an old version numcodec, I've added a check that completely kip the partial reading logic if the compressor does not have a decode_partial method.

Carreau:
I've also restored the `.coveragerc` file that was nuked in one of your merge with master; so that may help with the coverall. I'm going to wait to see where CI brings us.

zarr/indexing.py
```diff
@@ -825,3 +826,116 @@ def pop_fields(selection):
...
+                if slice_size == dim_size and index > 0:
+                    self.selection.pop()
+                else:
+                    break
+
+        chunk_loc_slices = []
+        last_dim_slice = None if self.selection[-1].step > 1 else self.selection.pop()
+        for i, sl in enumerate(self.selection):
+            dim_chunk_loc_slices = []
+            for i, x in enumerate(slice_to_range(sl, arr_shape[i])):
```
Carreau, line 925: you use `i` in both loop and the argument of the second depends on I; that seem weird.

Carreau: inner i seem unused as well.

Carreau:
I have a couple more cleanup in https://github.com/Carreau/zarr-python/tree/partial_read (d8464d4), but won't push now on this branch as I want to see what the coverage status is.

Carreau:
We ran out of time with travi CI. I'm working on migrating to GH actions;  Test won't complete without this. Apologies for the delay.

jakirkham:
cc @rabernat @jhamman (as I think we discussed this last year at SciPy 2019 so may be of interest)

Carreau:
Rebased on master, now that master have all the CI migrated to gh action. 
Added some extra cleanup, will see what the total coverage is and iterate on this.

docs/release.rst
```diff
@@ -30,6 +30,16 @@ This release od Zarr Python is is the first release of Zarr to not supporting Py
 See `this link <https://github.com/zarr-developers/zarr-python/milestone/11?closed=1>` for the full list of closed and
 merged PR tagged with the 2.6 milestone.
 
+* Add ability to partially read and decompress arrays, see :issue:`667`. IT is
```
Carreau, line 33:
```suggestion
* Add ability to partially read and decompress arrays, see :issue:`667`. It is
```

Carreau:
@andrewfulton9 

I've turned some of your conditional logics into asserts in d098528; so that there is no uncovered branch. Let me know if that's wrong. I've also use `np.prod(..., dtype=int)` to avoid polluting intermediate results with floating point numbers when the inner iterable was empty and it was returning `1.0`  instead of 1. 

I've turned some instance variables into local ones when not used outside the function and sprinkled some docs and and release note. (make Partial read opt-in, and added the option to open_array) .

I've left some `?` in the docstring of the partial reader as I'm not always sure if the ints are number of bytes or number of elements. Not sure if it matters much. 

Before I pushed last commits, all test seemed to be passing, windows was still pending... though coverage was missing in afew spot. You know the code better than I, so it will likely be easier to know how to cover the missing lines..

zarr/indexing.py
```diff
@@ -825,3 +826,113 @@ def pop_fields(selection):
...
+    Returns
+    -------
+    Tuple with 3 elements:
+
+    start: int
+        ? bytes offset in the chunk to read from 
+    nitems: int
+        ? number of (bytes? elements?) to read in the chunk from start
+    partial_out_selection: list of slices
+        ? indices of the final selection to assign to after the partial read.
```
Carreau, line 874: Not sure if those explanation are right.

zarr/indexing.py
```diff
@@ -825,3 +826,113 @@ def pop_fields(selection):
...
+
+    def __iter__(self):
+        chunk1 = self.chunk_loc_slices[0]
+        nitems = (chunk1[-1].stop - chunk1[-1].start) * np.prod(
+            self.arr_shape[len(chunk1) :], dtype=int
+        )
+        for chunk_selection in self.chunk_loc_slices:
+            start = 0
+            for i, sl in enumerate(chunk_selection):
+                start += sl.start * np.prod(self.arr_shape[i + 1 :], dtype=int)
```
Carreau, line 937: used dtype here and above

zarr/util.py
```diff
@@ -532,3 +534,78 @@ def __exit__(self, *args):
...
+            if blocks_to_decompress == int(blocks_to_decompress)
+            else int(blocks_to_decompress + 1)
+        )
+        start_block = int(start / self.n_per_block)
+        wanted_decompressed = 0
+        while wanted_decompressed < nitems:
+            if start_block not in self.read_blocks:
+                start_byte = self.start_points[start_block]
+                if start_byte == self.start_points_max:
+                    stop_byte = self.cbytes
```
Carreau, line 597: This branch seem like it is never encountered.

zarr/util.py
```diff
@@ -532,3 +534,78 @@ def __exit__(self, *args):
...
+                else:
+                    stop_byte = self.start_points[self.start_points > start_byte].min()
+                length = stop_byte - start_byte
+                data_buff = self.fs.read_block(self.key_path, start_byte, length)
+                self.buff[start_byte:stop_byte] = data_buff
+                self.read_blocks.add(start_block)
+            if wanted_decompressed == 0:
+                wanted_decompressed += ((start_block + 1) * self.n_per_block) - start
+            else:
+                wanted_decompressed += self.n_per_block
```
Carreau, line 607: neither is one of these two.

zarr/util.py
andrewfulton9: This is encountered if you are trying to get items out of the last block of an array like `Array[:, -1]`. It's there because if I just use `stop_byte = self.start_points[self.start_points > start_byte].min()`, the `min()` method raises an error on an empty array in this case.

zarr/util.py
andrewfulton9: I added this logic for errors that are raised in the case that you have nitems that are less than the size of a block, but span multiple blocks. The `if` clause is encountered on the first block you decompress. It's setting `wanted_decompressed` to the number of items from that first block that are part of nitems, otherwise on following iterations the wanted items for decompression that are read are just set to the number of items in a block since the start point of that block is the first item

For example:
```
z = zarr.zeros((100), chunks=(10))
z[8:12]
```
would failing before I added this logic because the chunk size is 10, and so it would only read the first block and then end because the number read was greater then `nitems` even though the actual `nitems` wanted weren't all read yet. So in this case, line 604 would evaluate to True on the first block and set `wanted_decompressed = 2` which would then allow `while wanted_decompressed<nitems` in line 591 to evaluate to `True` on the next iteration of the loop so it can read then next block to get the 2 items for `nitems` from the following block as well.

Does that make sense?

andrewfulton9:
@Carreau, thanks for the review and the contributions you added. I'll look over the coverage try to get that back up to 100, and fix the doc strings you added the ? to.

zarr/util.py
andrewfulton9: I just added tests to hit this and the above case. I also realized that the example case above won't cause the failure because the chunk isn't big enough for Blosc to split into different compression blocks. The test case I added works for it though

zarr/tests/test_core.py
```diff
@@ -276,6 +285,7 @@ def test_array_1d_selections(self):
         ix = [99, 100, 101]
         bix = np.zeros_like(a, dtype=bool)
         bix[ix] = True
+        print(a[ix])
```
Carreau, line 288:
```suggestion
```

zarr/util.py
Carreau: +1

zarr/util.py
Carreau: +1

Carreau:
Do you think you can rebase ? There seem to be merge commits that make changes hard to follow.
Other than than +1 to get that in.

joshmoore:
Note: discussion (plus notes) during the community call yesterday. General call for feedback so this can get merged to move forward the BaseStore work followed by the V3 updates.

Carreau:
If there is no feedback by this Wed, i'll go ahead and merge, we can revert if necessary.

zarr/core.py
```diff
@@ -1112,7 +1147,7 @@ def __setitem__(self, selection, value):
 
         See Also
         --------
-        get_basic_selection, set_basic_selection, get_mask_selection, set_mask_selection,
+        basic_selection, set_basic_selection, get_mask_selection, set_mask_selection,
```
jakirkham, line 1150: Is there some reason this needed to change? If not, may make sense to revert.

```suggestion
        get_basic_selection, set_basic_selection, get_mask_selection, set_mask_selection,
```

zarr/indexing.py
```diff
@@ -825,3 +827,120 @@ def pop_fields(selection):
         selection = tuple(s for s in selection if not isinstance(s, str))
         selection = selection[0] if len(selection) == 1 else selection
     return fields, selection
+
+
+def int_to_slice(dim_selection):
+    return slice(dim_selection, dim_selection + 1, 1)
```
jakirkham, line 833: Maybe we can just inline this since it is simple enough and used in only a couple places

jakirkham:
Thanks Andrew! 😄 Had a couple suggestions above

andrewfulton9:
Just pushed up those changes. Thanks @jakirkham !

Carreau (approved):

Carreau:
Thanks, with @jakirkham feedback and the last two commits I think we can move forward ! I'm approving as well and merging.

jakirkham:
Thanks all! 😀

martindurant:
Great to see this! I think it's important enough that we may want to put a blog somewhere. What do the final benchmarks look like? Am I right in thinking that blosc is still the default compressor and will continue to be?

jakirkham:
We do have a blog on the webpage repo. So writing something up there makes sense.

More generally we may want to do this to summarize the work done for the CZI grant. (cc @davidbrochart as well)

davidbrochart:
Thanks for pinging me @jakirkham, I would be happy to summarize the work done on the C++ side in this blog. We could also have a meeting and present to each other what we have done, what do you think?

rabernat:
Pangeo is organizing a new webinar series of short tech talks. It would be great to have a presentation on the latest developments from Zarr. Presentations will be recorded, publicized and assigned a DOI.

If anyone wants to sign up to present, just fill out this short form: https://forms.gle/zuU8XcQHHKS6DBvv8

jakirkham:
> Thanks for pinging me @jakirkham, I would be happy to summarize the work done on the C++ side in this blog. We could also have a meeting and present to each other what we have done, what do you think?

Yeah, @davidbrochart, I think that makes sense. Do you want to go ahead and set up a poll? I think we should try to schedule this in mid-February if that works

davidbrochart:
Actually I submitted a proposal for a presentation on xtensor-zarr in the Pangeo webinar series. If that is accepted, that could be a better option, since people could see the recording. What do you think?

jakirkham:
I think both are useful. The blog is also likely not too different from other write ups we are doing. So hopefully doesn’t take too much effort to produce. Plus it gives us an opportunity to share this work with a broader audience and publicly recognize CZI for their support, which I’m sure they will appreciate 🙂

andrewfulton9:
I can do a blog on the partial read work as well.

martindurant:
Very late question here: has partial read been implemented for *uncompressed* data? It should be even easier.

rabernat:
No

naught101:
Would it be possible to rename this "Partial read of blosc-compressed chunks" or similar?

jakirkham:
We could do that. Though this likely still comes in the commit history (and potentially other search paths)

That said, as there is a whole new sharding codec planned (please see [ZEP 2]( https://zarr.dev/zeps/accepted/ZEP0002.html )), this is probably best handled via that pathway, which is planned as part of the v3 work ( https://github.com/zarr-developers/zarr-python/pull/1583 )

Separately there has been discussion of moving from Blosc 1 to 2 ( https://github.com/zarr-developers/numcodecs/issues/413 ), which would likely entail other changes as well

So it is probably best to direct attention to more recent issues/PRs at this stage

Issue #669, opened 2020-12-03
Failures on Python 3.10


vstinner:
Hi,

39 tests fails when run on Python 3.10. See logs below. There are 4 different errors:

* _lib.pymdb_put(): TypeError: initializer for ctype 'char *' must be a cdata pointer, not bytearray
* _lib.pymdb_put(): TypeError: initializer for ctype 'char *' must be a cdata pointer, not numpy.ndarray
* _lib.pymdb_put(): TypeError: len() of unsized object
* _lmdb_decode_key_buffer(): AttributeError: '_cffi_backend.buffer' object has no attribute 'tobytes'

I tested zarr-python 2.5.0. I get similar the same errors with zarr-python 2.6.1.

Downstream Fedora issue: https://bugzilla.redhat.com/show_bug.cgi?id=1903075

Full tests log: [tests.log](https://github.com/zarr-developers/zarr-python/files/5636515/tests.log)

#### Version and installation information

Please provide the following:

* Value of ``zarr.__version__``: 2.5.0
* Value of ``numcodecs.__version__``: 0.7.2
* Version of Python interpreter: Python 3.10.0a2
* Operating system (Linux/Windows/Mac): Linux
* How Zarr was installed (e.g., "using pip into virtual environment", or "using conda"): build the Fedora package, see https://src.fedoraproject.org/rpms/python-zarr/tree/master

vstinner:
> _lib.pymdb_put(): TypeError: len() of unsized object

This one comes from: zarr/tests/test_core.py::TestArrayWithLMDBStore::test_array_0d

The error comes from `len(value)` where value type is: `<class 'numpy.ndarray'>`.

On the tested numpy 1.19.4, 0d arrays have no length. Example:

```
$ python3
Python 3.9.0 (default, Oct  6 2020, 00:00:00) 
>>> import numpy
>>> x = numpy.array(1.0, numpy.float64)
>>> x
array(1.)
>>> numpy.ndim(x)
0
>>> len(x)
TypeError: len() of unsized object
```

vstinner:
> _lmdb_decode_key_buffer(): AttributeError: '_cffi_backend.buffer' object has no attribute 'tobytes'

_cffi_backend.buffer never had a tobytes() method. Maybe previously, the key type was not _cffi_backend.buffer but another type?

Maybe `key[:]` or `bytes(buffer)` can be used instead.

[...]

PR #670, opened 2020-12-03
Simplify text handling in DB `Store`s

jakirkham:
Fixes https://github.com/zarr-developers/zarr-python/issues/669

`LMDBStore` had its own way of handling text before `ensure_text` was added to Numcodecs. With that utility function now available, we can simplify this a bit and avoid some subtleties of the LMDB implementation.

Similarly `DBMStore` had some of its own text handling issues that can be solved the same way.

Finally inline all `str.encode` calls to make it clearer what is happening in both  `LMDBStore` and `DBMStore`.

TODO:

### DIFF
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -35,7 +35,10 @@ from typing import Optional, Union, List, Tuple, Dict
 import uuid
 import time
 
-from numcodecs.compat import ensure_bytes, ensure_contiguous_ndarray
+from numcodecs.compat import (
+    ensure_bytes,
+    ensure_contiguous_ndarray
+)
 from numcodecs.registry import codec_registry
 
 from zarr.errors import (
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -37,6 +37,7 @@ import time
 
 from numcodecs.compat import (
     ensure_bytes,
+    ensure_text,
     ensure_contiguous_ndarray
 )
 from numcodecs.registry import codec_registry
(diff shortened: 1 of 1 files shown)


Carreau (approved):

Carreau:
Would that bump the minimum version of lmdb ?

jakirkham:
> Would that bump the minimum version of lmdb ?

Shouldn't affect it.

Basically [`ensure_text`]( https://github.com/zarr-developers/numcodecs/blob/1734bee75f113dde5a59ab38046f9ef158a5c296/numcodecs/compat.py#L126-L130 ) leverages the Python Buffer Protocol to coerce the data (`bytes` or other object) to a NumPy array without copying. Effectively this normalizes the data we are working with so we don't need to know about LMDB Buffer objects or whatever other object might come up ( MongoDB had a similar issue https://github.com/zarr-developers/zarr-python/pull/401 ). Then we perform [the `decode` step]( https://github.com/zarr-developers/numcodecs/blob/1734bee75f113dde5a59ab38046f9ef158a5c296/numcodecs/compat.py#L129 ) on the NumPy array, which we [already know to work]( https://github.com/zarr-developers/numcodecs/blob/1734bee75f113dde5a59ab38046f9ef158a5c296/numcodecs/tests/test_compat.py#L12 ).

Carreau:
Thanks for the explanation; let's get that in.

jakirkham:
Thanks Matthias! 😄

PR #671, opened 2020-12-03
Finish removing coverall

Carreau:
There are regular issue where it shows decrease of coverage while only
non-executable lines are added (e.g. comments), this might be dues to
out of order reception of reports from travis-ci jobs.

We did activate codecov.io, which does not appear to have this issue.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,7 @@ before_script:
     - mongo mydb_test --eval 'db.createUser({user:"travis",pwd:"test",roles:["readWrite"]});'
 
 install:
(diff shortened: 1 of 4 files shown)


jakirkham (approved):

Carreau:
I also went into the repo setting s and modify the master branch restriction so that coverall is not required, but codecov/project is. The former was necessary as now coverall does not receive results and therefore would stay yellow forever.

jakirkham:
Ah sorry about that. Thanks for fixing

PR #675, opened 2020-12-16
Clear temporary test files after test.

Carreau:
This is crawled by codecov on github to find coverage report;
Probably want to make that a temporary directory; but that will be for
another time

### DIFF
--- a/zarr/tests/test_storage.py
+++ b/zarr/tests/test_storage.py
@@ -1794,6 +1794,7 @@ def test_format_compatibility():
             else:
                 assert compressor.codec_id == z.compressor.codec_id
                 assert compressor.get_config() == z.compressor.get_config()
+    fixture.clear()
 
 
 @skip_test_env_var("ZARR_TEST_ABS")


jakirkham:
Will this clear the non-temporary data too? If so, that might prove tricky when contributors run the tests locally and commit changes (accidentally delete other files here in the process).

Carreau:
> Will this clear the non-temporary data too? If so, that might prove tricky when contributors run the tests locally and commit changes (accidentally delete other files here in the process).

The `/fixture/` folder does not exist in the zarr hierarchy; the test create it on line 1719.

I'll try to properly turn this into a `mktemp()` context manager.

jakirkham:
It’s using [this directory]( https://github.com/zarr-developers/zarr-python/tree/master/fixture ) right ? Or am I missing something?

Carreau:
Hum...locally I have nested fixture folder:

```
$ git status
On branch grow-matrix
Your branch is up to date with 'github/grow-matrix'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)
	A Notebook has No Name - .ipynb
	A Notebook has No Name - 1.ipynb
	dask-worker-space/
	fixture/fixture/     <- See here
	foo.py
	rasm.nc
	testzarr.zip
	tmp/
	v3.zarr/

nothing added to commit but untracked files present (use "git add" to track)
```

That I had to empty manually, I thought that was the one; but you are right that my above fix would be wrong. will need to find another workaround for codecove to not look at all those files.

jakirkham:
Is there some kind of configure file where we can tell codecov to not look at this directory?

PR #679, opened 2020-12-21
Try to test on windows using github action

Carreau:


Carreau:
This works and does not change coverage; I'll merge as is, and might start to disable some test on AppVeyor to speed up CI testing and not wait hours.

PR #680, opened 2020-12-21
try to install azurite on github action

Carreau:
Again to get back to our 100% coverage.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -47,7 +47,9 @@ jobs:
     - name: Create Conda environment with the rights deps
       shell: "bash -l {0}"
       run: |
-        conda create -n zarr-env python==${{matrix.python-version}} bsddb3 numcodecs==0.6.4 lmdb pip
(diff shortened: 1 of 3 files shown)


jhamman:
@Carreau - per your request, here's a smallish-public Zarr store on Azure: 

```python
store = zarr.storage.ABSStore('carbonplan', prefix='carbonplan-share/zarr-demo', account_name='carbonplan')
zarr.open_consolidated(store)
```

Unfortunately, as I've just learned while preparing this snippet for you, you also need to provide the `account_key` kwarg in Zarr's interface. So I'm not sure how useful this will be for you. FWIW, adlfs doesn't require the `account_key` for public containers.

Carreau:
Thanks super useful, it seem that querying `''` on this instance works, but not on azurite. 
So likely an azurite problem, I'll try to get an older impl of the package.

jhamman:
I wonder if #620 would fix the problem you are seeing with the empty string query.

Carreau:
> I wonder if #620 would fix the problem you are seeing with the empty string query.

Possibly. I can have a look once we are back on track with testing.

Carreau:
This increases coverage; and I've open 681 to track reverting the xfail commits.

PR #682, opened 2021-01-01
try to skip ndbm in coverage.

Carreau:
NDBM is not package by conda on non-mac; so it's hard to test on github action. 
Marking as skip to try to get coverage back to 100%


Carreau:
This increase coverage; merging.

PR #683, opened 2021-01-01
Try to install gdbm to increase coverage.

Carreau:
All in title


Carreau:
GDBM does not work either on conda forge... 

https://github.com/conda-forge/python-feedstock/issues/125 might have to skip coverage there.

PR #686, opened 2021-01-02
Update azure code and skip gdbm coverage.

Carreau:
Newer version of azure and azurite are stricter on allowing queries on
empty blob names; here we add checks that blob names are non empty, as
well as take the slow path (which is the correct one) when trying to get
a store size.


Carreau:
And coverage is at 100% !

PR #687, opened 2021-01-02
Try to install azurite on Windows + GitHub Actions.

Carreau:
If we manage to get azurite on gh action + windows, we might be able to deactivate azzurite on appveyor and get the windows CI to much faster.


Carreau:
That works, good, we might be able to turn off azuritin on appveyor finally, and get CI be fast.

PR #695, opened 2021-01-25
Reenable mypy and flake8 linting on CI.

Carreau:
Those were ran by tox, and got lost in the migration.
Maybe we want to make them a separate  job at some point for speed and not run them on all matrix elements.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -83,7 +83,7 @@ jobs:
       shell: "bash -l {0}"
       run: |
         conda activate zarr-env
-        flake8
+        flake8 zarr
         mypy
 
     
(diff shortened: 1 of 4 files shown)


.github/workflows/python-package.yml
```diff
@@ -79,3 +79,11 @@ jobs:
           #name: codecov-umbrella # optional
           #fail_ci_if_error: true # optional (default = false)
         verbose: true # optional (default = false)
+    - name: Linting
+      shell: "bash -l {0}"
+      run: |
+        conda activate zarr-env
+        flake8
```
joshmoore, line 86:
```suggestion
        flake8 zarr
```

.github/workflows/python-package.yml
```diff
@@ -79,3 +79,11 @@ jobs:
           #name: codecov-umbrella # optional
           #fail_ci_if_error: true # optional (default = false)
         verbose: true # optional (default = false)
+    - name: Linting
+      shell: "bash -l {0}"
+      run: |
+        conda activate zarr-env
+        flake8 zarr
+        mypy
```
joshmoore, line 87:
```suggestion
        mypy zarr
```

joshmoore (approved):
:+1: Green!

Issue #597, opened 2020-09-03
PermissionError when writing to larger zarr array


JoeWise:
#### Code to reproduce
```python
points = zarr.open('data/points.zarr', mode='w', shape=(10000, 20000, 3), dtype=np.uint8, chunks=(10000, 1, 3))

for col in range(0, 20000):
    for row in range(0, 10000):
        points[row, col] = 50, 100, 150
```

#### Problem description

When I run the above code snippet I get a PermissionError, even when running as administrator. This doesn't happen for smaller zarr arrays e.g. shape=(100, 200, 3)

`Traceback (most recent call last):
  File "<stdin>", line 3, in <module>
  File "C:\UserPath\AppData\Roaming\Python\Python38\site-packages\zarr\core.py", line 1115, in __setitem__
    self.set_basic_selection(selection, value, fields=fields)
  File "C:\UserPath\AppData\Roaming\Python\Python38\site-packages\zarr\core.py", line 1210, in set_basic_selection
    return self._set_basic_selection_nd(selection, value, fields=fields)
  File "C:\UserPath\AppData\Roaming\Python\Python38\site-packages\zarr\core.py", line 1501, in _set_basic_selection_nd
    self._set_selection(indexer, value, fields=fields)
  File "C:\UserPath\AppData\Roaming\Python\Python38\site-packages\zarr\core.py", line 1550, in _set_selection
    self._chunk_setitem(chunk_coords, chunk_selection, chunk_value, fields=fields)
  File "C:\UserPath\AppData\Roaming\Python\Python38\site-packages\zarr\core.py", line 1664, in _chunk_setitem
    self._chunk_setitem_nosync(chunk_coords, chunk_selection, value,
  File "C:\UserPath\AppData\Roaming\Python\Python38\site-packages\zarr\core.py", line 1729, in _chunk_setitem_nosync
    self.chunk_store[ckey] = cdata
  File "C:\UserPath\AppData\Roaming\Python\Python38\site-packages\zarr\storage.py", line 825, in __setitem__
    replace(temp_path, file_path)
PermissionError: [WinError 5] Access is denied: 'C:\\UserPath\\DataPath\\data\\points.zarr\\0.0.0.fca098bcac1b4eb6a62733dc4f2d6966.partial' -> 'C:\\UserPath\\DataPath\\data\\points.zarr\\0.0.0'`

#### Version and installation information
* Value of ``zarr.__version__`` == 2.4.0
[...]

PR #698, opened 2021-02-10
Make DirectoryStore __setitem__ resilient against antivirus file locking

ericgyounkin:
See https://github.com/zarr-developers/zarr-python/issues/597

Enterprise and government computers generally have always-on realtime file scanning antivirus software that causes problems with any system that relies on creation/deletion of lots of small files.  We have seen this with SQLite as well, with the temp files it uses.  The fix I'm proposing here is to simply try a few times and wait inbetween tries for the antivirus software to release the file lock.  I've been running with this fix for a while, and I can now use Zarr on government machines.  Hugely helpful.  I've seen PermissionErrors floating around the gitter chat and in stackoverflow, so I think this fix is needed for others as well.

The 0.1 sec delay was empirically determined, I've found that the vast majority of writes that generate a PermissionError on the first attempt will work on the second.

Not sure how to build a test for this.

TODO:


ericgyounkin:
Not sure about these meta.py Linting error messages in Linux Testing / build (3.8).  Am I missing something?

joshmoore:
regarding adding a test, perhaps something like:

https://github.com/joshmoore/zarr-python/commit/3ecee99d3d4733be239416e46e1c8af526506cac

joshmoore:
These errors others have seen:

```
zarr/meta.py:82: error: Incompatible return value type (got "List[Union[Tuple[str, str], Tuple[str, str, Tuple[int, ...]]]]", expected "str")
zarr/meta.py:183: error: "generic" has no attribute "real"
zarr/meta.py:184: error: "generic" has no attribute "imag"
```
so not related to your PR. @Carreau have you seen these elsewhere?

ericgyounkin:
@joshmoore Thats great, thanks!  I hadn't seen pytest raises used before, so that makes a lot of sense.  How would I use your commit in my PR?

ericgyounkin:
Ignore last comment, pulled from @joshmoore commit and updated PR

ericgyounkin:
@joshmoore I'm a little confused about the status of this pull request, is there anything I should do further?  I see that these checks now pass after your changes.  Thanks for your work on this.

joshmoore:
Hey @ericgyounkin. Apologies. I think I pushed the last commit with my fingers crossed and then moved on to something else while waiting on the tests to go green. Let me merge in master and make sure everything is green, but I assume there's nothing else from your side. Thanks for hanging in there!

ericgyounkin:
Thanks @joshmoore!  Do you know how the release schedule works?  When the next release is planned?  Just planning for the future of my app.

joshmoore:
Trying _again_ to get things green after #705.

zarr/storage.py
```diff
@@ -860,8 +860,10 @@ def __setitem__(self, key, value):
         try:
             self._tofile(value, temp_path)
 
-            # move temporary file into place
-            os.replace(temp_path, file_path)
+            # move temporary file into place;
+            # make several attempts at writing the temporary file to get past
+            # potential antivirus file locking issues
+            retry_call(os.replace, (temp_path, file_path))
```
joshmoore, line 866: @ericgyounkin : just realizing (since I have never been able to test this in anger) doesn't `PermissionError` need listing here? Otherwise, no exceptions are going to be caught.

ericgyounkin: @joshmoore Yes, my mistake.  Just added it.

joshmoore (approved):
It's a bit disingenuous for me to review this considering I submitted the function that's being used, but at least there have been two sets of eyes on it. If anyone has issues with the method signature, there's still some time to clean it up before the release of 2.7.0. Otherwise, we'll see how this fairs in the wild.

joshmoore:
Thanks for hanging in there, @ericgyounkin!

ericgyounkin:
@joshmoore Thank you for all your work, big Zarr fan!

PR #699, opened 2021-02-15
FSStore: key_separator fix

joshmoore:
key_separator: fix handling of meta keys

    This rebases and cleans a fix from Martin's original PR.
    Two new test methods fail without it and pass with since
    keys for meta-files like `.zarray` were being converted
    into `/zarray`.

 see: https://github.com/martindurant/zarr/pull/1/commits/5795bc7368eb0b86aaeb1956ffed514e672db96e

TODO:
* [x] Add unit tests and/or doctests in docstrings
* [] Add docstrings and API docs for any new/modified user-facing classes and functions
* [ ] New/modified features documented in docs/tutorial.rst
* [ ] Changes documented in docs/release.rst
* [ ] AppVeyor and Travis CI passes
* [ ] Test coverage is 100% (Coveralls passes)

PR #700, opened 2021-02-15
Update dtype usage in zarr/meta.py

joshmoore:
Changes in zarr's annotations, numpy, and/or mypy have led to failures on `mypy zarr/`across a large number of PRs. This tries to take the most direct route to passing GitHub actions so that PRs can be unblocked.

Thanks to @rgommers and @seberg for jumping in.

TODO:

### DIFF
--- a/zarr/meta.py
+++ b/zarr/meta.py
@@ -6,7 +6,7 @@ import numpy as np
 from zarr.errors import MetadataError
 from zarr.util import json_dumps, json_loads
 
-from typing import Union, Any, List, Mapping as MappingType
+from typing import cast, Union, Any, List, Mapping as MappingType
 
 ZARR_FORMAT = 2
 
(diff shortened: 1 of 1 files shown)


joshmoore:
Failures can be seen in https://github.com/zarr-developers/zarr-python/pull/699/checks?check_run_id=1904975788

```
arr/meta.py:82: error: Incompatible return value type (got "List[Union[Tuple[str, str], Tuple[str, str, Tuple[int, ...]]]]", expected "str")
zarr/meta.py:183: error: "generic" has no attribute "real"
zarr/meta.py:184: error: "generic" has no attribute "imag"
Found 3 errors in 1 file (checked 30 source files)
```

rgommers:
Hmm, the `dtype.descr` change surprises me. Good chance it's on purpose, but I don't really want to comb through multiple NEPs - so let's just ask @seberg.

seberg:
Do you know what changed?  This has nothing to do with the dtype NEPs, but `np.dtype(...).descr` has _never_ returned a string as far as I can tell. The only actual change I can think of (and find!) is that we might be returning unicode strings instead of bytes somewhere around here.

joshmoore:
Thanks, both. I was assuming it was a numpy bump that led to this, but I'll need to try to find what the change was, unless @Carreau knows.

Another possibility is that that mypy has gotten stricter, in which case I could update the annotation rather than the logic.

seberg:
I would guess the typing is due to NumPy 1.20 now including typing.  The `.descr` change I am confused about though, since I am not aware of a change.

joshmoore:
Asked for vetoes in gitter. Since all PRs are currently failing, I'm going to merge and start re-running the builds. If there are any issues with the changes here, I'll be happy to open a follow-up PR.

PR #705, opened 2021-03-01
2.7.0: update release section

joshmoore:
Prepares the release notes for a version before merging more v3 specific code, during which `master` will be in a pre-release state (`.devX`, `.betaY`, etc.)

Also fixes two build issues which are making recent PRs fail:
 * duplicate block in `releases.yml` likely introduced during conflict resolution
 * moto recently went to 2.0 and now requires `moto[server]` for installation

### DIFF
--- a/.github/workflows/releases.yml
+++ b/.github/workflows/releases.yml
@@ -3,6 +3,7 @@ name: Wheels
 on: [push, pull_request]
 
 jobs:
+
   build_artifacts:
     name: Build wheel on ubuntu-latest
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


joshmoore:
First draft. Open to suggestions (especially around whether or not explanation of all the intermediate GHA PRs)

joshmoore:
I've changed the wording to make it clearer that this is the proposed version. If there are no vetoes, I'm going to get this in today since the test fixes are necessary for other PRs.

joshmoore:
Preemptively merging. This PR is in _preparation_ for a release, but it'd be good to get more (some) feedback on what's in / what's out for 2.7.

PR #709, opened 2021-03-11
[WIP] Expand (and pass) nested FSStore tests

d-v-b:
This PR adds tests to `FSStore(..., key_separator="/")`. The first test checks that writing to a zarr array backed by a nested FSStore succeeds. This test fails without changes to `FSStore.getitems` -- the original `getitems` implementation normalizes keys, calls `self.map.getitems(normalized_keys)`, and returns a dict with those normalized keys. It seems that the code calling `FSStore.getitems()` may not recognize these keys and thus discard all the values, producing an array with only fill value. I modified `FSStore.getitems` to return a dict with the input keys.   

This PR also expands on a second test (reading the same array with a separate store instance) which currently fails. I will try to get this second test passing, but any pointers would be appreciated. cc @martindurant 

TODO:

### DIFF
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -1117,9 +1117,26 @@ class FSStore(MutableMapping):
     def listdir(self, path=None):
         dir_path = self.dir_path(path)
         try:
-            out = sorted(p.rstrip('/').rsplit('/', 1)[-1]
-                         for p in self.fs.ls(dir_path, detail=False))
-            return out
+            children = sorted(p.rstrip('/').rsplit('/', 1)[-1]
+                              for p in self.fs.ls(dir_path, detail=False))
+            if self.key_separator == '/':
+                if array_meta_key in children:
+                    # special handling of directories containing an array to map nested chunk
+                    # keys back to standard chunk keys
+                    new_children = []
+                    root_path = self.dir_path(path)
+                    for entry in children:
+                        entry_path = os.path.join(root_path, entry)
+                        if _prog_number.match(entry) and self.fs.isdir(entry_path):
+                            for dir_path, _, file_names in self.fs.walk(entry_path):
+                                for file_name in file_names:
+                                    file_path = os.path.join(dir_path, file_name)
+                                    rel_path = file_path.split(root_path)[1]
+                                    new_children.append(rel_path.replace(os.path.sep, '.'))
+                        else:
+                            new_children.append(entry)
+                    return sorted(new_children)
+            return children
         except IOError:
             return []
 
@@ -2592,6 +2609,7 @@ class RedisStore(MutableMapping):
         Keyword arguments passed through to the `redis.Redis` function.
 
     """
+
     def __init__(self, prefix='zarr', **kwargs):
         import redis
         self._prefix = prefix
@@ -2677,6 +2695,7 @@ class ConsolidatedMetadataStore(MutableMapping):
     zarr.convenience.consolidate_metadata, zarr.convenience.open_consolidated
 
     """
+
     def __init__(self, store, metadata_key='.zmetadata'):
         self.store = store
 
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -1051,11 +1051,8 @@ class FSStore(MutableMapping):
         return key.lower() if self.normalize_keys else key
 
     def getitems(self, keys, **kwargs):
-        keys_transformed = [self._normalize_key(key) for key in keys]
-        results = self.map.getitems(keys_transformed, on_error="omit")
-        # The function calling this method may not recognize the transformed keys
-        # So we send the values returned by self.map.getitems back with the input keys.
-        return {keys[keys_transformed.index(rk)]: rv for rk, rv in results.items()}
+        keys = [self._normalize_key(key) for key in keys]
+        return self.map.getitems(keys, on_error="omit")
 
     def __getitem__(self, key):
         key = self._normalize_key(key)
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -1119,7 +1119,9 @@ class FSStore(MutableMapping):
         try:
             children = sorted(p.rstrip('/').rsplit('/', 1)[-1]
                               for p in self.fs.ls(dir_path, detail=False))
-            if self.key_separator == '/':
+            if self.key_separator != "/":
+                return children
+            else:
                 if array_meta_key in children:
                     # special handling of directories containing an array to map nested chunk
                     # keys back to standard chunk keys
@@ -1136,7 +1138,8 @@ class FSStore(MutableMapping):
                         else:
                             new_children.append(entry)
                     return sorted(new_children)
-            return children
+                else:
+                    return children
         except IOError:
             return []
 
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -1130,11 +1130,10 @@ class FSStore(MutableMapping):
                     for entry in children:
                         entry_path = os.path.join(root_path, entry)
                         if _prog_number.match(entry) and self.fs.isdir(entry_path):
-                            for dir_path, _, file_names in self.fs.walk(entry_path):
-                                for file_name in file_names:
-                                    file_path = os.path.join(dir_path, file_name)
-                                    rel_path = file_path.split(root_path)[1]
-                                    new_children.append(rel_path.replace(os.path.sep, '.'))
+                            for file_name in self.fs.find(entry_path):
+                                file_path = os.path.join(dir_path, file_name)
+                                rel_path = file_path.split(root_path)[1]
+                                new_children.append(rel_path.replace(os.path.sep, '.'))
                         else:
                             new_children.append(entry)
                     return sorted(new_children)
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -1051,8 +1051,11 @@ class FSStore(MutableMapping):
         return key.lower() if self.normalize_keys else key
 
     def getitems(self, keys, **kwargs):
-        keys = [self._normalize_key(key) for key in keys]
-        return self.map.getitems(keys, on_error="omit")
+        keys_transformed = [self._normalize_key(key) for key in keys]
+        results = self.map.getitems(keys_transformed, on_error="omit")
+        # The function calling this method may not recognize the transformed keys
+        # So we send the values returned by self.map.getitems back with the input keys.
+        return {keys[keys_transformed.index(rk)]: rv for rk, rv in results.items()}
 
     def __getitem__(self, key):
         key = self._normalize_key(key)
--- a/zarr/tests/test_core.py
+++ b/zarr/tests/test_core.py
@@ -35,8 +35,9 @@ from zarr.storage import (
 from zarr.util import buffer_size
 from zarr.tests.util import skip_test_env_var, have_fsspec
 
-
 # noinspection PyMethodMayBeStatic
+
+
 class TestArray(unittest.TestCase):
 
     def test_array_init(self):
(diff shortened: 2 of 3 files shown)


martindurant:
Ping me when I should have a look

d-v-b:
Don't look too closely at the git history, it reflects my confusion as I try to understand what is getting tested where :)

I would like some input on how best to parameterize `test_core.TestArrayWithFSStore` over `key_separators`, since there are multiple test classes that would be affected by this and I would rather not duplicate a ton of code. It seems that classes that inherit from `unitttest.testcase`, e.g. `TestArray`, only have methods that take no arguments, so the `key_separator` cannot be passed as an argument to methods like `test_hexdigest`. I don't know pytest well enough to see a simple solution that doesn't involve a lot of copy + paste.

cc @martindurant

joshmoore:
I thought the fixture-based style in https://github.com/zarr-developers/zarr-python/pull/417 was pretty intuitive.

d-v-b:
I think this is ready to look at. Instead of making any big changes to the testing architecture I just went with the flow and added two more subclasses of `TestArray` in `test_core.py` -- `TestArrayWithFSStoreNested` and `TestArrayWithFSStoreNestedPartialRead`. Getting tests to pass for `TestArrayWithFSStoreNested` required changing `FSStore.getitems` and `FSStore.listdir`, and both changes bridge the "."-separated chunk keys used in the core zarr api to the "/" separated keys of the storage implementation. 

Additionally, there are a bunch of failing tests for `TestArrayWithFSStoreNestedPartialRead`, which are due to the same issue -- code is looking for "."-separated keys where there are none. But I don't have a fix for those failing tests at the moment, since the partial read codebase is more opaque to me.

cc @martindurant

martindurant:
Is https://github.com/pydata/xarray/issues/5028 related, perhaps?

d-v-b:
I think that xarray issue is caused by the default use of `fsstore._normalize_keys`, which only changes  

> Is [pydata/xarray#5028](https://github.com/pydata/xarray/issues/5028) related, perhaps?

I think that issue is only mildly related, it's due to this line in https://github.com/zarr-developers/zarr-python/blob/master/zarr/storage.py#L1051 in conjunction with `normalize_keys=True` in the `fsstore` constructor.

martindurant:
I'm not sure what the argument for normalize to be True versus False by default. Are you saying that it ought to be an obvious way to pass the argument down?

d-v-b:
> I'm not sure what the argument for normalize to be True versus False by default. Are you saying that it ought to be an obvious way to pass the argument down?

I don't really have an opinion on that... I think the upper case vs lower case thing is orthogonal to this PR -- the issues I tried to fix in this PR turn on "0.0.0" vs "0/0/0" for chunk keys.

zarr/storage.py
```diff
@@ -1114,9 +1117,26 @@ def dir_path(self, path=None):
...
+                              for p in self.fs.ls(dir_path, detail=False))
+            if self.key_separator == '/':
+                if array_meta_key in children:
+                    # special handling of directories containing an array to map nested chunk
+                    # keys back to standard chunk keys
+                    new_children = []
+                    root_path = self.dir_path(path)
+                    for entry in children:
+                        entry_path = os.path.join(root_path, entry)
+                        if _prog_number.match(entry) and self.fs.isdir(entry_path):
```
martindurant, line 1130: Perhaps `ls.find` would be simpler here?

zarr/storage.py
```diff
@@ -1114,9 +1117,26 @@ def dir_path(self, path=None):
     def listdir(self, path=None):
         dir_path = self.dir_path(path)
         try:
-            out = sorted(p.rstrip('/').rsplit('/', 1)[-1]
-                         for p in self.fs.ls(dir_path, detail=False))
-            return out
+            children = sorted(p.rstrip('/').rsplit('/', 1)[-1]
+                              for p in self.fs.ls(dir_path, detail=False))
+            if self.key_separator == '/':
```
martindurant, line 1122: Maybe simpler to invert the test and return `children` here, being closer to the original and the more common path

martindurant (commented):
I am OK with the changes to FSStore.
I note many "added line not covered" notes in the diff.

zarr/storage.py
d-v-b: can you elaborate? (and do you mean `list.find`)? 

Also, this code largely copied from the same method on `NestedDirectoryStore`: https://github.com/zarr-developers/zarr-python/blob/17728e86ecbcb38778cdd7b0f2e8d430a878de5a/zarr/storage.py#L1276, so if there's a cleaner way to do this then maybe I could put together a PR that cleans up `FSStore` and `NestedDirectoryStore` together.

martindurant: Sorry, I meant `fs.find`. We happen to know for fsspec implementations that this will exist, and sometimes it is more performant that walking the directory tree. `NestedDirectoryStore` probably doesn't have such a function, and walking the tree is the best you can do for local files anyway.

zarr/storage.py
d-v-b: good idea! implemented in babd76a78b53277c98917051a0cfbbc8728f0452

zarr/storage.py
d-v-b: using `fs.find` does clean things up a bit: see b679c08cc637e369a73625feddb42c7a956307e6

d-v-b:
I'm not sure what to make of the test coverage warnings, since I'm not changing the API at all, and some of the lines of code being flagged are inside the tests themselves. I do note that `store.getitems` appears untested, should I add tests for that in `test_storage`?

martindurant:
> I do note that store.getitems appears untested, should I add tests for that in test_storage

If this is the case, then yes; the intent was definitely to test it, and previously we required ==100% coverage.

joshmoore:
Final commit doesn't appear to have triggered. Trying a re-open.

jakirkham:
cc @grlee77

joshmoore:
I've rebased and added the tests from #718. Looks like there are test failures from partial reads (#667)

cc: @andrewfulton9

grlee77:
The one test failure in the non-nested FSStore `test_hexdigest` ([see here](https://github.com/zarr-developers/zarr-python/runs/2473284874?check_suite_focus=true#step:7:74)) seems to be due to the addition of a `dimension_separator` attribute in `.zarray`.

The being tested hex would match for the .zattr below if the line containin `dimension_separator` was removed. (For `DirectoryStore`, there is not `dimension_separator` entry):
```
{
    "chunks": [
        100
    ],
    "compressor": {
        "blocksize": 0,
        "clevel": 5,
        "cname": "lz4",
        "id": "blosc",
        "shuffle": 1
    },
    "dimension_separator": ".",
    "dtype": "<i4",
    "fill_value": null,
    "filters": null,
    "order": "C",
    "shape": [
        1050
    ],
    "zarr_format": 2
}
```

grlee77:
Note that if the `FSStore` is initialized without a `key_separator` kwarg being passed in (as is the case in `TestArrayWithFSStorePartialRead`), then there is no `dimension_separator` entry in the `.zarray`. Is this the expected behavior, or should a `dimension_separator` entry always be present?

joshmoore:
@grlee77 : the initial commits tried deprecating key_separator and always storing dimension_separator. Always storing dimension_separator, however, led to more changes in the `hexdigest()` value for existing datasets. I wasn't sure of the overall impact of that. Since `dimension_separator` could then be `None`, I needed something that could be guaranteed non-`None`, and that's `key_separator`. That's a reversible chain that could likely use some more discussion.

grlee77:
The remaining test nested FSStore failures not related to hex digests are resolved if the `key_path` attribute of `zarr.util.PartialReadBuffer` is updated to take into account '.' -> '/' in the nested case:

For example, switching the current:
```Python
        self.store_key = self.map._key_to_str(store_key)
```
to
```Python
        _key_path = self.map._key_to_str(store_key)
        _key_path = _key_path.split('/')
        _key_path = '/'.join(_key_path[:-1] + [self.chunk_store._normalize_key(_key_path[-1])])
        self.key_path = _key_path
```

works, but that is just the quick hack I came up with during debugging and I haven't looked to closely to see where this fix should actually live. recommendations?

joshmoore:
@grlee77, it doesn't look insane 🙂 and if all current tests (incl. yours) are happy, I'd say it's an improvement. Based on @d-v-b's [earlier evaluation](https://github.com/zarr-developers/zarr-python/pull/718#issuecomment-829515181), it seems like we want a clearer delineation between keys as used in storage.py and core.py, where the former _can_ use "/" and the latter _never_ users "/" (?). `PartialReadBuffer` is only used in core.py, but I don't know where partial read perhaps is _also_ a leaky abstraction. Lots of moving parts.

joshmoore:
@d-v-b : any thoughts on @grlee77's proposal? I imagine a 2.9.0 could go out [soon](https://github.com/zarr-developers/zarr-python/pull/725) and it'd be nice to include this (or have it beforehand in a 2.8.2).

d-v-b:
Disclaimer: I don't know anything at all about the partial read functionality. That being said, it looks like the function in question (`zarr.util.PartialReadBuffer`) is operating at the storage level, i.e. it's working with real keys to objects on a file system. In that case the proposed change doesn't raise any red flags from me.

joshmoore:
Ok. Pushed @grlee77's fix as well as hexdigest fixes. Let's see if we can get it all green!

joshmoore:
_Whew!_

andrewfulton9:
I think what's happening with the `PartialReadBuffer` is that its using the mapper class to rebuild the full path to the file in the storage system so that it can use `fs.read_block`, to only read a part of the file, instead of using `FSMap` to get the file. I think that maybe when I tested that PR, I didn't account for store_keys having a leading '/' which is why @grlee77 had to make the above changes to make it pass. 

The code below would also work for fixing the issue, but I haven't tested it, and the above fix works fine too I think.
``` python
self.key_path = self.map._key_to_str(self.chunk_store._normalize_key(store_key))
```

joshmoore:
Can confirm that the tests pass with @andrewfulton9's version as well.

zarr/storage.py
```diff
@@ -1144,9 +1147,28 @@ def dir_path(self, path=None):
...
-            out = sorted(p.rstrip('/').rsplit('/', 1)[-1]
-                         for p in self.fs.ls(dir_path, detail=False))
-            return out
+            children = sorted(p.rstrip('/').rsplit('/', 1)[-1]
+                              for p in self.fs.ls(dir_path, detail=False))
+            if self.key_separator != "/":
+                return children
+            else:
+                if array_meta_key in children:
+                    # special handling of directories containing an array to map nested chunk
```
joshmoore, line 1156: @martindurant: can you comment if this matches your expectations?

joshmoore:
@martindurant: this is perhaps a bit out of scope for this PR since it's specifically focused on `FSStore` but in case we want to re-evaluate at which level `dimension_separator` needs to be applied, I just realized that this doesn't work:

```
import zarr as zr
import dask.array as da

z = zr.storage.FSStore(
    "foo.zarr",
    dimension_separator="/",
    auto_mkdir=True,
    mode="w",
)
a = zr.creation.array(store=z, data=[range(4) for x in range(4)])

# Passes
base = da.from_zarr(z, "")
m = base.max().compute()
assert m == 3, m

# Fails, since `FSStore` is not used but only FSMap
base = da.from_zarr("foo.zarr")
m = base.max().compute()
assert m == 3, m
```

cc: @will-moore

zarr/storage.py
joshmoore: @martindurant : any thoughts?

martindurant: OK, now I have reminded myself  what it's doing, and I suppose this is reasonable. As to it's accuracy in all cases, I can only guess it looks right.

It makes you wonder how useful `listdir` actually if for the case of `fsspec`. I suppose it remains true that a user might want to start exploring a dataset at some high-level group and descend to a specific array, yet make sure they never list the entire set of files (which could be expensive).

joshmoore: Thanks, @martindurant. I'm going to take that as a :thumbsup: and get this into a 2.8.1 release. It seems that this may be something that needs a re-evaluation down the line.

joshmoore (approved):

joshmoore:
I've migrated the `da.from_zarr` issue to https://github.com/dask/dask/issues/7673

Issue #707, opened 2021-03-05
Nested storage detection in Zarr V2


joshmoore:
see: https://github.com/ome/ngff/issues/29 and https://github.com/bcdev/jzarr/pull/17

In order to better handle Zarr arrays created with `NestedDirectoryStorage` or `FSStore(key_separator="/")`, SabineEmbacher  and I have been working on a "protocol heuristic" that can be used by V2 implementations to detect nested chunking rather than requiring the user to specify it correctly.

**tl;dr: This proposes a new key for `.zarray` which it would be good to have feedback on.**

----

### Proposal 

When creating a zarr array:

 * add to .zarray json: `{"dimension_separator": "/"}`
 * always write a 0-position chunk

When opening an array:

 * try to read the separator character from the .zarray json
 * if not available, try to find a 0-position chunk
 * if not available, at every read action try to find chunks with both standard variants until the situation is clarified. (Standard separator list `["/", "."]`)

----

### Points for discussion:
 * The name `dimension_separator` differs from the code implementation `key_separator` to reduce confusion about whether _every_ separator in the key name is effected.
 * There has been some discussion (community call, gitter) about whether or not "/" could become the default.
 * Is an addition to the .zarray metadata sufficiently low impact to be rolled into v2?

SabineEmbacher:
Thoughts on the name of the separator:
What the separator separates is the respective chunk index per dimension.
See https://github.com/bcdev/jzarr/discussions/19 and https://github.com/bcdev/jzarr/pull/17#issuecomment-788730222 and

Therefore the following suggestions:

- chunk_index_separator
- chunk_index_sep
- chunk_idx_separator
- chunk_idx_sep

jbms:
One issue with the detection protocol is that it does not work when writing.  While auto-detection is nice it might be better to just assume "." unless the metadata key is present --- any existing arrays could be fixed up, and hopefully all future arrays would have the metadata key.  That avoids the added complexity (and inefficiency) of auto-detection.

joshmoore:
Hi @jbms. You mean a `mode="a"`-style writing? If so, I see the problem. I had initially intended the auto-detection only as a means of not needing to "standardize" a new key. Then I got greedy and went for both. Anyone else have thoughts?

In the case of a `mode="w"` situation I would assume the writer simply *asserts* where it will write.

> any existing arrays could be fixed up, and hopefully all future arrays would have the metadata key.

[...]

PR #715, opened 2021-04-08
v2 spec: add optional dimension_separator (see #707)

joshmoore:
Various implementations allow for defining the separator between the
dimension indexes when writing chunks:

 * n5-zarr defines a `dimensionSeparator` parameter;
 * zarr-python's NestedDirectoryStore does so by default
 * and FSStore provides a `key_separator` parameter;
 * tensorstore has a `key_encoding` parameter; and
 * jzarr is looking to add the same functionality.

When writing an array, it is straight-forward to set this separator
and have arrays properly configured. Consumers of such arrays,
however, must either know *a priori* if their arrays use a
non-default separator or must loop through all possible chunks keys
searching for the right one.

By defining adding an optional metadata key to the .zarray, we:

 * preserve the efficient configuration of arrays
 * while keeping the v2 spec backwards compatible.

The primary downsides are that this will be the first optional metadata
value in the v2 spec and therefore we don't have a strong understanding of how that will play out, and datasets which were previously written with non-default separators will need updating in order to enable the detection though that is no worse than the current situation.

cc: @alimanfoo @chris-allan @SabineEmbacher @jbms @axtimwalde @manzt @gzuidhof @haileyajohnson @DennisHeimbigner @zarr-developers/core-devs 

(*If I've missed any Zarr v2 implementors, please add them*)

### DIFF
--- a/docs/spec/v2.rst
+++ b/docs/spec/v2.rst
@@ -81,11 +81,12 @@ filters
 The following keys MAY be present within the object:
 
 dimension_separator
-    A string, optimally of length 1, which defines the separator placed
+    If present, either the string ``"."`` or ``"/""`` definining the separator placed
     between the dimensions of a chunk. If the value is not set, then the
-    default MUST be assumed to ``"."``, leading to chunk keys of the form "0.0".
-    The most common non-default value is ``"/"``, which leads to nested keys of
-    the form "0/0" that in most implementations produces a directory structure.
+    default MUST be assumed to be ``"."``, leading to chunk keys of the form "0.0".
+    Arrays defined with ``"/"`` as the dimension separator can be considered to have
+    nested, or hierarchical, keys of the form "0/0" that SHOULD where possible
+    produce a directory-like structure.
 
 Other keys MUST NOT be present within the metadata object.
 
--- a/docs/spec/v2.rst
+++ b/docs/spec/v2.rst
@@ -78,6 +78,15 @@ filters
     filters are to be applied. Each codec configuration object MUST contain a
     ``"id"`` key identifying the codec to be used.
 
+The following keys MAY be present within the object:
+
+dimension_separator
+    A string, optimally of length 1, which defines the separator placed
+    between the dimensions of a chunk. If the value is not set, then the
+    default MUST be assumed to ``"."``, leading to chunk keys of the form "0.0".
+    The most common non-default value is ``"/"``, which leads to nested keys of
+    the form "0/0" that in most implementations produces a directory structure.
+
 Other keys MUST NOT be present within the metadata object.
 
 For example, the JSON object below defines a 2-dimensional array of 64-bit
--- a/docs/spec/v2.rst
+++ b/docs/spec/v2.rst
@@ -88,7 +88,8 @@ dimension_separator
     nested, or hierarchical, keys of the form "0/0" that SHOULD where possible
     produce a directory-like structure.
 
-Other keys MUST NOT be present within the metadata object.
+Other keys SHOULD NOT be present within the metadata object and SHOULD be
+ignored by implementations.
 
 For example, the JSON object below defines a 2-dimensional array of 64-bit
 little-endian floating point numbers with 10000 rows and 10000 columns, divided


martindurant:
So presumably you would be able to still pass the separator argument OR look for the metadata key. The first would take priority and would remain the recommendation for the current and older versions of zarr and for datasets that were written by them.

joshmoore:
@martindurant: exactly. Wasn't sure if that was too much for this paragraph, but I can look for a spot to go into more details. I imagine if someone passes a separator that doesn't match the expected one we could warn in mode="r". For mode="a|w", I'd personally favor an exception but an alternative would be to update the metadata, effectively hiding the existing chunks.

martindurant:
It would be reasonable to supply a helper function somewhere to convert an existing array from one style of metadata to another.

jbms:
I'm happy with this and can add support to tensorstore and neuroglancer.  Is this the confirmed metadata format or are we waiting on additional discussion?  I'd rather not add this only to then have to also support something else in the future.

martindurant:
I think we are hoping for more +1 from the other maintainers.
I think I'd be OK with merging if there is no opposition within, say, a week.

DennisHeimbigner:
The change in the spec says this:
>The most common non-default value is ``"/"``, which leads to nested keys of the form "0/0" that in most implementations produces a directory structure.

I think the last part "that in most implementations produces a directory structure" should be removed since it is inserting implementation claims into what should be an implementation free document.

martindurant:
>  "that in most implementations produces a directory structure" should be removed 

 Would different wording help, like "in some implementation is a special character used to mark directories" or "the directory delimiter in posix-like storage", or "commonly used as a directory delimiter" ?

DennisHeimbigner:
Why is it needed at all?

shoyer:
I like including an example of a non-default value because provides a succinct motivation for why this feature exists. The Zarr spec already has a number of such examples for explaining other features.

DennisHeimbigner:
Reasonable. Then ignore this suggested change. Thanks.

joshmoore:
Note: 5 days have passed. https://github.com/zarr-developers/governance/blob/master/GOVERNANCE.md states

> Changes to the specification require a dedicated issue on our issue tracker and even more time than an API change. The appropriate length of time is currently under discussion.

Considering this is an optional addition, I'd propose that we leave this call for objections open until the end of the week and then start moving forward? (Any objections to _that plan_ today please. :wink: )

DennisHeimbigner:
Can the separator be a UTF-8 character => multi-byte?

axtimwalde:
I think that's a valid question. I vote for UTF-8 arbitrary length string. ASCII encouraged.

jbms:
For simplicity it might be best to limit it to one of a small set of characters for now, e.g. {"." and "/"}, and possibly expand that later if there is a need for it.

On the implementation side it is not terribly difficult to just concatenate whatever arbitrary string is specified, but there are a few ways things can go wrong:

- If the string starts or ends with a digit, then it is possible that a given key could correspond to more than one chunk.
- If the string is length 0 obviously that leads to ambiguity.
- Currently the allowed characters are `[0-9./]`, which is pretty widely supported by various underlying storage systems.  If you allow arbitrary unicode or even arbitrary ascii you can quickly run into characters not supported by the storage system (e.g. local filesystem), or which may have special behavior when opened --- e.g. on MS Windows there are a bunch of reserved names that cause problems when opened, and they aply to every directory, not just the root.
- Some filesystems do various forms of normalization to unicode characters.

While in general the user should be free to make a bad choice and suffer the consequences, in some cases software that processes zarr may be intended to handle an untrusted zarr volume and then care must be taken to handle these issues.

Unless there is currently a use case for a separator other than {"." and "/"} it seems like it would be reasonable to restrict it to those options for now.

DennisHeimbigner:
Also by putting it into .zarray, it implies that the separator can differ for each
array. I would think it better if the dimension separator was constrained
to be the same for all arrays in the dataset.

jbms:
Putting the separator into `.zarray` seems helpful since currently `.zarray` stands on its own.  If you have to look in parent directory `.zattr` files that makes it more complicated.

As I imagine implementing this, and generally in what seems like the most natural implementation, there wouldn't be any problem with different separators within the same hierarchy.  However I see that having different separators within the same hierarchy would not work with the existing implementation of NestedDirectoryStore.  However I would suppose that once this feature is implemented NestedDirectoryStore would no longer be needed except for API compatibility.

axtimwalde:
I like that it can be different for each dataset, it's simpler to parse and may be useful.

DennisHeimbigner:
I would suggest for version 3 that you introduce a single "superblock" object, ".zarr" say,
to hold metadata that applies to the whole dataset. This could also hold
extension information and the dimension separator could be put there as well.
p.s. I use the term "dataset" to refer to the whole collection of arrays and metadata.
Is there a preferred term for that collection?

DennisHeimbigner:
This is somewhat off-topic, but it sounds like the real unit here is the .zarray+.zattr+chunks.
There really is no notion of a "dataset" as a collection of arrays and groups have no
semantics at all; they are purely for namespaces.
Under that model, then indeed, everything has to be attached to .zarray, including dimension separators and extensions.

axtimwalde:
I used the term dataset for 'array' as in HDF5 and N5, sorry for the confusion. I like the term 'container' for the entirety of groups and arrays that belong to the same root. That would be the equivalent of an HDF5 file.

SabineEmbacher:
> Also by putting it into .zarray, it implies that the separator can differ for each
> array. I would think it better if the dimension separator was constrained
> to be the same for all arrays in the dataset.

The separator should be individually adjustable for each array. A zarr data set can thus also be compiled manually from already existing data.
For example, if already existing data from different processing years are to be merged into a time series data set. Or if subsets of already existing different satellite sensors are merged to one data set, because these extracts from different sensors are needed for certain climate models or other calculations.

chris-allan:
Having some experience dealing with arbitrary unicode both from an implementation and security perspective I would largely parrot @jbms. Many of us are using, or have the desire to use, Zarr in a trusted environments and in particular on large shared filesystems. Allowing arbitrary unicode dimension separators sets up a whole series of cross-language programming safety, path normalization, and subpath containment scenarios that must be dealt with to be safe in such an environment. Having a defined set of allowed ASCII separator characters is both easier for implementers and safer.

DennisHeimbigner:
I agree that limiting the set of separators is a good idea.
I gather that the current proposal is the set '/' + '.'

DennisHeimbigner:
I have another "meta" question. The V2 spec says:
````Other keys MUST NOT be present within the metadata object.````
So how can you add dimension_separator as a new key to .zarray without
potentially breaking all existing implementations that try to read the new format?
Assuming that the separator is '.' for the new file.

**Comment by joshmoore** [comment]
Thanks for the feedback, everyone. Summarizing:

 * There have been no objections voiced to `dimension_separator` as the metadata key.
 * There is concern over having too many possible values, and general agreement that restricting to `/` and `.` for the first pass covers (most known) current use cases.
 * The definition is a _bit_ implementation specific, but not more so than other text in this version. (I've tried to improve it somewhat in the last commit)
 * There's a request for being able to specify the (default?) metadata at the top-level, but that is not possible without v3. (@DennisHeimbigner, would you like to open a v3 issue for that? [See the current state for details](https://github.com/zarr-developers/zarr-specs/tree/core-protocol-v3.0-dev))
 * Finally, @DennisHeimbigner raises a valid point "how can you add dimension_separator as a new key to .zarray without
potentially breaking all existing implementations that try to read the new format?" -- you can't. Technically this is a breaking change. However, as far as I've seen, no implementation is enforcing this contract. 

I've pushed a commit to capture the above. Considering this is a contraction of ... the contract I assume there will be fewer rather than more objections to the key itself. If anyone would like to raise objections concerning the breakingness, now is the time.

I've started work on the Python implementation, but could likely use some help covering the breadth of stores. zarr.js' implementation is in https://github.com/gzuidhof/zarr.js/pull/89. If anyone else has in-progress versions they'd like to list, please feel free.

Other comments:

> https://github.com/zarr-developers/zarr-python/pull/715#issuecomment-819189072 I would suggest for version 3 that you introduce a single "superblock" object,

[Entry point metadata](https://zarr-specs.readthedocs.io/en/core-protocol-v3.0-dev/protocol/core/v3.0.html#entry-point-metadata) is in the new spec.

 > https://github.com/zarr-developers/zarr-python/pull/715#issuecomment-819189072 p.s. I use the term "dataset" to refer to the whole collection of arrays and metadata.

I personally use "fileset" but can get behind standardizing the term. "Dataset" I find problematic because of the HDF5 use.

**Comment by jbms** [comment]
Currently tensorstore does fail if `.zarray` contains extra fields.  However I will remove that check at the same time that I implement this feature.

**Comment by DennisHeimbigner** [comment]
That restriction should probably be removed from the spec and replaced
with one that says that unrecognized keys should be ignored.

**Comment by joshmoore** [comment]
Thanks for the suggestion, @DennisHeimbigner. I've pushed:

```diff
-Other keys MUST NOT be present within the metadata object.
+Other keys SHOULD NOT be present within the metadata object and SHOULD be
+ignored by implementations.
```

joshmoore:
The zarr-python implementation (https://github.com/zarr-developers/zarr-python/pull/716) is finally green. I'd propose we try to get these both merged this week, unless anyone else who is working on an implementation needs more time.

Carreau (approved):

joshmoore:
Thanks for the review, @Carreau. Considering that there are now several implementations based on this (nczarr, tensorstore, and zarr.js minimally) I'm going to merge. Adding comment to #716 and then the big question will be when and as what to release. (My assumption is 2.8.0)

joshmoore:
Some additional transcribed notes from @alimanfoo:
 - We discussed whether or not this should actually lead to a version number bump. However v3 is already defined and there's no process for a version in between.
 - The primary concern would be that existing filesets would become corrupted by older implementations that do not understand the new array key.
 - However, since that's already the case today _without_ the metadata key, this should be considered a "fix rather than a [breaking] change"

Issue #530, opened 2019-12-18
zarr.open_array does not properly recognize NestedDirectoryStore


JimBiardCics:
```python
import numpy
import zarr

a = zarr.create((10, 100, 100), chunks = (1, 100, 100), dtype = 'f4',
                store = zarr.NestedDirectoryStore('foo.zarr'), overwrite = True)

a[...] = 1.0

print('shape =', a.shape)
print('a has non-zero values =', numpy.any(a))

del a

a = zarr.open_array('foo.zarr', mode = 'r')

print('shape =', a.shape)
print('a has non-zero values =', numpy.any(a))
```

#### Problem description

A zarr 'file' created using NestedDirectoryStore will not load correctly when later opened using the path alone. The shape (and other metadata) is correct, but the values are not loaded. In the code sample above, the shape of the original and reloaded array is the same, but the contents don't match. The second array contains only zeroes.

#### Version and installation information

* Value of ``zarr.__version__`` - 2.3.2
* Value of ``numcodecs.__version__`` - 0.6.4
* Version of Python interpreter - 3.6.7 (anaconda)
* Operating system (Linux/Windows/Mac) - Mac OS Mojave (10.14.6)
* How Zarr was installed - using conda

jrbourbeau:
Thanks for opening an issue and the nice example @JimBiardCics! I'm able to reproduce the same behavior.

When using a string path with `zarr.open_array`, zarr will use the following logic to guess what store should be used

https://github.com/zarr-developers/zarr-python/blob/e7708c948d2c0ff91863d675c1072b0dfe9ce2a6/zarr/creation.py#L130-L142

For a path like `'foo.zarr'`, zarr will use a `DirectoryStore`. Since the array was saved using a `NestedDirectoryStore`, you'll want to open the array with the same store. 

In your example, replacing

```python
a = zarr.open_array('foo.zarr', mode = 'r')
```

with

```python
a = zarr.open_array(zarr.NestedDirectoryStore('foo.zarr'), mode = 'r')
```

should fix things

JimBiardCics:
The problem, as I see it, is that the application using the file has no way to determine which type of store to use. It is not particularly reasonable to expect all applications to know ahead of time which store type to use when opening a zarr file.

stuarteberg:
+1 for this issue!

Instead of guessing the store from the file extension, can the `store` class name simply be added to the metadata in `.zarray`?

alimanfoo:
[...]

PR #612, opened 2020-09-19
Create a  Base store class for Zarr Store.

Carreau:
In progress, 

All existing stores in zarr-python now inherit from this; and thus all the test stop testing for a `close()` method and call it unconditionally.

The base store is a bit more strict in what it accepts than subclasses (only allow strings), as it is generally safer for the superclass to be stricter as anything that works with superclass will work with subclass, but the opposite is untrue.

TODO:
* [ ] Add unit tests and/or doctests in docstrings
* [ ] Add docstrings and API docs for any new/modified user-facing classes and functions
* [ ] New/modified features documented in docs/tutorial.rst
* [ ] Changes documented in docs/release.rst
* [ ] AppVeyor and Travis CI passes
* [ ] Test coverage is 100% (Coveralls passes)

PR #716, opened 2021-04-15
Implement `dimension_separator` for Python storage classes (See #715)

joshmoore:
Implements the new optional array metadata key, `dimension_separator`, from #715 for the Python library.

close: #530

## Details

* All top-level storage classes now take an optional `dimension_separator`
  parameter which defaults to `None`, but can also be `.` or `/`.

* A ValueError is raised at normalization time if this is not the case.

* `None`s are normalized to the default of `.` in all except the
  NestedDirectoryStore case.

* The value is stored as `self._dimension_separator` on participating classes
  so that array creation can lookup the value.

* This value deprecates the `key_separator` value from FSStore.

* Wrapper classes like LRUCacheStore and ConsolidatedMetadataStore *do not*
  follow this pattern and instead rely on the value in the underlying store.

## Discussion

* [ ] The major issues I see with this is the propagation of "Yet Another Constructor Argument". A more complete proposal might take all array initialization parameters similar to those which might be stored in the entry-level metadata in v3.

* [ ] Additionally, this only stores the flat separator (`.`) in `.zarray` if the user actively requests it. Since the value is optional, implementations will default to that value anyway.

## TODO:

### DIFF
--- a/zarr/creation.py
+++ b/zarr/creation.py
@@ -69,7 +69,7 @@ def create(shape, chunks=True, dtype=None, compressor='default',
         A codec to encode object arrays, only needed if dtype=object.
     dimension_separator : {'.', '/'}, optional
         Separator placed between the dimensions of a chunk.
-      .. versionadded:: 2.8
+        .. versionadded:: 2.8
 
     Returns
     -------
--- a/zarr/creation.py
+++ b/zarr/creation.py
@@ -123,7 +123,7 @@ def create(shape, chunks=True, dtype=None, compressor='default',
 
     # optional array metadata
     if dimension_separator is None:
-        dimension_separator = getattr(store, "_dimension_separator", ".")
+        dimension_separator = getattr(store, "_dimension_separator", None)
     dimension_separator = normalize_dimension_separator(dimension_separator)
 
     # initialize array metadata
--- a/zarr/creation.py
+++ b/zarr/creation.py
@@ -13,13 +13,14 @@ from zarr.n5 import N5Store
 from zarr.storage import (DirectoryStore, ZipStore, contains_array,
                           contains_group, default_compressor, init_array,
                           normalize_storage_path, FSStore)
+from zarr.util import normalize_dimension_separator
 
 
 def create(shape, chunks=True, dtype=None, compressor='default',
            fill_value=0, order='C', store=None, synchronizer=None,
            overwrite=False, path=None, chunk_store=None, filters=None,
            cache_metadata=True, cache_attrs=True, read_only=False,
-           object_codec=None, **kwargs):
+           object_codec=None, dimension_separator=None, **kwargs):
     """Create an array.
 
     Parameters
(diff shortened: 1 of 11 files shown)


joshmoore:
Latest commit reverts the proactive saving of `dimension_separator` metadata in order to keep `hexdigest()` as stable as possible.

zarr/storage.py
```diff
@@ -1012,30 +1028,44 @@ class FSStore(MutableMapping):
         like "s3://bucket/root"
     normalize_keys : bool
     key_separator : str
-        Character to use when constructing the target path strings
-        for data keys
+        public API for accessing dimension_separator. Never `None`
+        See dimension_separator for more informatino.
```
joshmoore, line 1032:
```suggestion
        See dimension_separator for more information.
```

zarr/creation.py
```diff
@@ -66,6 +67,9 @@ def create(shape, chunks=True, dtype=None, compressor='default',
         True if array should be protected against modification.
     object_codec : Codec, optional
         A codec to encode object arrays, only needed if dtype=object.
+    dimension_separator : {'.', '/'}, optional
+        Separator placed between the dimensions of a chunk.
+      .. versionadded:: 2.8
```
Carreau, line 72:
```suggestion
        .. versionadded:: 2.8
```

Not sure about indentation of the directive.

zarr/util.py
```diff
@@ -243,6 +243,16 @@ def normalize_order(order: str) -> str:
...
 
 
+def normalize_dimension_separator(sep: Optional[str]) -> Optional[str]:
+    if sep is None:
+        return None
+    elif sep not in (".", "/"):
+        raise ValueError(
+            "dimension_separator must be either '.' or '/', found: %r" % sep)
+    else:
+        return sep
```
Carreau, line 253: The  `if sep is None` can be folded into the else.

Not found of the "normalize", normalize confer to me that it might change it (like / to \ on windows); 
maybe just a check ? 

```
assert sep in (".", "/", None)
``` 
might seem enough as it is only used in 2 places ?

Carreau: And once we have a base store we can turned it into a validated property.

joshmoore: The original commit turned `None` into `.` so it was much more normalize-y. I'm not completely sure that logic won't come back though. (And this way I could write a unit test for it!)

joshmoore:
Thanks for the suggestions, @Carreau. Changes pushed.

From the call discussion, there were no objections to this work moving forward so please consider this a last call (incl. if adjustments to the spec wording from #715 are needed) The plan will be to get this merged and started preparing for a v2-specific 2.8.0 on top of which @Carreau can rebase his base store work (#612)

Again: @zarr-developers/core-devs please speak up if there are any concerns.

joshmoore:
Moving forward with zarr-python 2.8.0 later today if there are no further comments.

joshmoore:
Note: one outstanding issue has been filed separately: https://github.com/zarr-developers/zarr-python/issues/722

PR #718, opened 2021-04-15
Update Array to respect FSStore's key_separator 

### LINKED ISSUE #717: Array._chunk_key does not respect FSStore's key_separator attribute

```Python
data = np.arange(1000).reshape(10, 10, 10)
name = 'raw'

store = FSStore('/tmp/test.zr', key_separator='/', auto_mkdir=True)
f = open_group(store, mode='w')
f.create_dataset(name, data=data, chunks=(5, 5, 5),
                 compressor=None)
h = open_group(store, mode='r')
# check below fails because h[name][:] will be all zeros
np.testing.assert_array_equal(h[name][:], data)  
```

#### Problem description

If an nested Zarr array is read in using an FSStore object initialized with `key_separator='/'`, all array entries are reported as zeros.

This seems to be because the `_chunk_key` attribute of `zarr.core.Array` assumes that '.' is the separator

'2.7.1.dev2' (current master branch)

------

I will make a PR addressing this now

grlee77:
[Description of PR]

closes #717

I added a couple of FSStore test classes to test_hierarchy.py. The "NestedFSStore" class tests the fix on the specific case raised in #717.

TODO:
~* [ ] Add docstrings and API docs for any new/modified user-facing classes and functions~
~* [ ] New/modified features documented in docs/tutorial.rst~

### DIFF
--- a/zarr/core.py
+++ b/zarr/core.py
@@ -1950,7 +1950,11 @@ class Array:
         return self._encode_chunk(chunk)
 
     def _chunk_key(self, chunk_coords):
-        return self._key_prefix + '.'.join(map(str, chunk_coords))
+        if hasattr(self._store, 'key_separator'):
+            separator = self._store.key_separator
+        else:
+            separator = '.'
+        return self._key_prefix + separator.join(map(str, chunk_coords))
 
     def _decode_chunk(self, cdata, start=None, nitems=None, expected_shape=None):
         # decompress
--- a/zarr/tests/test_hierarchy.py
+++ b/zarr/tests/test_hierarchy.py
@@ -17,7 +17,6 @@ except ImportError:  # pragma: no cover
 from numcodecs import Zlib
 from numpy.testing import assert_array_equal
 
-import zarr
 from zarr.attrs import Attributes
 from zarr.core import Array
 from zarr.creation import open_array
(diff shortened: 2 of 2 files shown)


zarr/core.py
```diff
@@ -1950,7 +1950,11 @@ def _process_for_setitem(self, ckey, chunk_selection, value, fields=None):
         return self._encode_chunk(chunk)
 
     def _chunk_key(self, chunk_coords):
-        return self._key_prefix + '.'.join(map(str, chunk_coords))
+        if hasattr(self._store, 'key_separator'):
```
joshmoore, line 1953: A heads up that #716 deprecates `key_separator` in favor `store._dimension_separator`. Looking at this usage, I imagine wrapping stores like `LRUCacheStore` and `ConsolidatedMetadataStore` are going to need to set `_dimension_separator as well.

grlee77: Okay, thanks for letting me know. Should we hold off on this PR until #716 is in and then revisit?

joshmoore: Nah, let's get this out as a point release. Just a warning that this "API" will change. It's all kismet.

joshmoore (approved):

zarr/core.py
d-v-b: The behavior of `_chunk_key` should not be changed here. Instead, the behavior of `FSStore._getitems_` should be changed to properly remap the keys. See https://github.com/zarr-developers/zarr-python/pull/709/files#diff-565e487a2f60258b6baa2e4db8ef175cc16b8a949651834bd43d0a9f21e07358R1054

d-v-b:
This PR needs to be reverted, since the solution in this PR breaks the abstraction boundary between array and chunk store, and the problem this PR solves is addressed in https://github.com/zarr-developers/zarr-python/pull/709 by changes to `FSStore`, which is the correct level of abstraction for this fix.

joshmoore:
Thanks, @d-v-b. And sorry for the troubles. If you haven't already, can you capture the test from here in #709?

joshmoore:
see https://github.com/zarr-developers/zarr-python/pull/729

PR #734, opened 2021-05-03
Activate dependabot

### LINKED ISSUE #732: Activate dependabot?
https://github.com/zarr-developers/zarr-python/network/updates offers to generate a dependabot file. Any objections?

joshmoore:
Close #732

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+---
+version: 2
+updates:
+  - package-ecosystem: "pip"
+    directory: "/"
+    schedule:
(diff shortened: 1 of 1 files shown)


joshmoore:
I haven't heard any objections. Merging and if dependabot becomes annoying we can always disable.

joshmoore:
All bumps generated from this activation are green. Merging them all which will cause all PRs to start being tested with them. If necessary, we can either rollback and/or have a more conservative version of the requirements while which _also_ gets tested. Thoughts welcome.

PR #738, opened 2021-05-11
WIP, RFC: Optimize setitem with chunk equal to fill_value, round 2

### LINKED ISSUE #366: Optimize storing fill_value to Array
Would be nice if when storing a scalar or an array that contains only the `fill_value` of the `Array`, the chunk is merely deleted (if it exists) instead of writing out a chunk that contains only the `fill_value` for all elements. The `Array` should behave identically in either case. The only benefit is that storage is cutdown automatically in this case. A situation where this may come up is when copying one `Array` to another.

d-v-b:
Continuation of https://github.com/zarr-developers/zarr-python/pull/367
From the original description:

> Fixes #366
> 
> If a chunk only contains the fill_value, delete the key-value pair for that chunk instead of storing the array. Should  cutdown on storage space require for the Array. Also should improve the performance of copying one Array to another.

While working with sparse datasets, it would great if we could avoid creating objects in storage for empty chunks. This PR implements this functionality, which is controlled via the boolean property `Array._write_empty_chunks`. 

In this PR, the logic for detecting "empty" chunks is implemented in [`Array._process_for_setitem()`](https://github.com/d-v-b/zarr-python/blob/opt_setitem_fill_value/zarr/core.py#L1918), a function that prepares chunk data for writing. I added two levels of conditional execution to this function, the first checking if `self._write_empty_chunks` is `False`, and the second dispatching to a suitable routine for checking if the elements of a chunk are all equal to `fill_value`. Object arrays required a special handling here. 

If `_write_empty_chunks` is `False` and the chunk is empty, `_process_for_setitem()` returns `None` instead of returning compressed bytes, and I added a corresponding check for `None`-ness in the downstream code.

The routines I added have performance implications. Every chunk write now requires evaluating two new conditionals -- the check for `_write_empty_chunks`, and the null check for the output of `_process_for_setitem`. More conditionals are evaluated for arrays with `_write_empty_chunks` set to `False`, and there is added computation required for checking each chunk for emptiness. I would love to avoid these conditionals, especially for the default case when we are not checking chunks for emptiness, so if anyone sees a way to do that, chime in :) 

TODO:


d-v-b:
oh and I should add -- I considered a different logic for the "is the chunk empty" check. instead of comparing python objects (the chunk and the fill value), we could compress the chunk and compare the resulting bytes to the (cached) result of compressing a chunk-sized array of `fill_value`s. The latter approach would use more memory but would avoid the special casing for object arrays.

d-v-b:
ping @jakirkham , would love to get your thoughts on this

I've been using this feature for my workplace and it's been extremely useful for managing large, sparse volumes. I think the broader community would benefit from it.

zarr/core.py
```diff
@@ -1948,6 +1969,25 @@ def _process_for_setitem(self, ckey, chunk_selection, value, fields=None):
...
+        # clear chunk if it only contains the fill value
+        if not self._write_empty_chunks:
+            if self.dtype == 'object':
+                # we have to flatten the result of np.equal to handle outputs like
+                # [np.array([True,True]), True, True]
+                is_empty = all(flatten(np.equal(chunk, self.fill_value, dtype='object')))
+            else:
+                is_empty = np.all(chunk == self._fill_value)
+
+            if is_empty:
```
jni, line 1981: Since these lines are repeated lots of times, @d-v-b, do you want to make a function or two? e.g. `def clear_if_empty(check_obj_dtype=False) -> bool:`, returns True if the block was found to be empty and successfully cleared.

jni:
@zarr-developers/core-devs I'm 👍 on this change, and I would go even further and make the new default behaviour `write_empty_chunks` to be **False**. It will indeed change the output of the library but imho it is an implementation detail and conforms to the spec (afaik! 😅), so I think it's worth pulling the band-aid off. My own personal expectation was that zarr *already* did this!

@d-v-b I've made a minor comment on the code that should be an easy refactor. It would also be good if you could ensure that all the lines are covered, though I don't know what could exercise the `except Exception` clause.

Also, could you update this to match the latest master? Thanks!

zarr/core.py
d-v-b: Noted! In https://github.com/zarr-developers/zarr-python/pull/738/commits/044a9b8c64a37d3ba9bfe247ad23114bd3411611 I functionalize approximately as per your suggestion

d-v-b:
Some new changes inspired by @jni's [comments](https://github.com/zarr-developers/zarr-python/pull/738#discussion_r638377145)

- `_process_for_setitem` no longer returns an encoded chunk. chunk encoding happens in the functions that perform storage.
- new function `_chunk_isempty`, which checks if a chunk is empty. returns `True` if a chunk is empty, `False` otherwise
- new function `_chunk_delitem`, which attempts to delete a chunk key from the store. returns `True` if the delete was successful or on `KeyError`, returns `False` on any other exception.

Still need to improve test coverage..

meggart:
FWIW, I can just report from my experiences in the Julia Zarr implementation. In Zarr.jl, *not* writing empty chunks is already the default. Regarding performance implications, from some small tests that I have done it looked like the time spent for performing the `isempty` check was negligible compared to time spent in compression or even writing to disk, so I would be in favor of changing the default behavior in python as well. 

Having this controlled by a keyword argument would have the disadvantage that users of downstream packages like xarray would not directly benefit from this change but probably some option propagation through xarray would be necessary.

zarr/core.py
```diff
@@ -1859,9 +1879,38 @@ def _chunk_setitems(self, lchunk_coords, lchunk_selection, values, fields=None):
...
+            values = dict(zip(ckeys, map(self._encode_chunk, cdatas)))
         self.chunk_store.setitems(values)
 
+    def _chunk_isempty(self, chunk):
+        if self.dtype == 'object':
+            # we have to flatten the result of np.equal to handle outputs like
+            # [np.array([True,True]), True, True]
+            is_empty = all(flatten(np.equal(chunk, self.fill_value, dtype='object')))
+        else:
+            is_empty = np.all(chunk == self._fill_value)
```
shoyer, line 1897: does this work properly if `fill_value` is NaN?

shoyer:
I agree with @jni and @meggart -- it is not clear to me why you would _not_ want to do this. I think this could be the new default behavior.

zarr/core.py
d-v-b: Nope! Do you have any advice for making that work?

d-v-b:
I found a potential performance issue at the key deletion step. The deletion routine in this PR does not use bulk delete APIs (e.g., via the `delitems` method on `FSSpec.FSMap`) for removing keys. I think this could be very inefficient.

shoyer:
That does sound potentially problematic!

jakirkham:
cc @martindurant (in case you have thoughts on fsspec usage here 🙂)

martindurant:
I am also surprised that this is not the default, and would agree that I can't see people wanting to use the current behaviour ever. I've been wrong before...

I have no problem from the fsspec point of view. Indeed, it would be nice to delitems, if possible (this will be 100% latency on remote filesystems like s3).

d-v-b:
I added a branch for using `chunk_store.map.delitems` conditioned on an attribute check, and I found that this breaks with the FSSpec local file system implementation, which [doesn't handle a list of paths as an input](https://github.com/intake/filesystem_spec/blob/801021bab0139657b716c98efbd9734ed3cebce4/fsspec/implementations/local.py#L107). This looks like a bug in fsspec, given the function signature of [`FSMap.delitems`](https://github.com/intake/filesystem_spec/blob/801021bab0139657b716c98efbd9734ed3cebce4/fsspec/mapping.py#L113) @martindurant

zarr/core.py
jakirkham: Maybe something like this?

```suggestion
            is_empty = np.all(chunk == self._fill_value if np.isnan(self._fill_value) else np.isnan(chunk))
```

d-v-b:
By using an unreleased version of fsspec this passes local tests. NaNs should now work as a fill value, and I added tests for that. 

I have a few lingering questions that I'd like to get some feedback on:

- I don't like the parameter name `write_empty_chunks`. I would be happy to rename it to something shorter / more expressive, e.g. `sparse_write`. But there are probably better ideas out there.
- Should `write_empty_chunks` (or whatever it's called) be a public attribute of `zarr.core.Array`? 
- Should I run some tests detailing the performance impact of the chunk checking?

martindurant:
What change in fsspec does this rely on?

d-v-b:
> What change in fsspec does this rely on?

https://github.com/intake/filesystem_spec/commit/576008ff849392c1c801b0722d595a2369e11797

d-v-b:
Any idea why these FSSpec + s3 tests are failing @martindurant ?

martindurant:
aiobotocore 1.4.0 landed a short time ago and broke things. It may be that s3fs was installed here before the fix appeared in 2021.08.0 .

joshmoore:
I (temporarily) excluded 1.4.0 on the mainline.

joshmoore:
Testing s3fs 2021.8.0 in https://github.com/zarr-developers/zarr-python/pull/815

zarr/storage.py
```diff
@@ -1188,7 +1197,7 @@ def listdir(self, path=None):
                         if _prog_number.match(entry) and self.fs.isdir(entry_path):
                             for file_name in self.fs.find(entry_path):
                                 file_path = os.path.join(dir_path, file_name)
-                                rel_path = file_path.split(root_path)[1]
+                                rel_path = file_path.split(root_path)[1].strip(os.path.sep)
```
joshmoore, line 1200: The mainline is using `lstrip()`. Is both needed?

zarr/storage.py
```diff
@@ -1149,6 +1149,15 @@ def __delitem__(self, key):
         else:
             del self.map[key]
 
+    def delitems(self, keys):
```
joshmoore, line 1152: What's the implication that this is defined on FSStore?

zarr/core.py
```diff
@@ -74,6 +75,15 @@ class Array:
         If True and while the chunk_store is a FSStore and the compresion used
         is Blosc, when getting data from the array chunks will be partially
         read and decompressed when possible.
+    write_empty_chunks : bool, optional
```
joshmoore, line 78: Add default to docstring (once we've decided...)

zarr/core.py
```diff
@@ -1856,11 +1879,59 @@ def _chunk_getitems(self, lchunk_coords, lchunk_selection, out, lout_selection,
...
+            else:
+                for ckey in empty_chunks.keys():
+                    self._chunk_delitem(ckey)
+            nonempty_keys = cdatas.keys() - empty_chunks.keys()
+            to_store = {k: self._encode_chunk(cdatas[k]) for k in nonempty_keys}
+        else:
+            to_store = {k: self._encode_chunk(v) for k, v in cdatas.items()}
+        self.chunk_store.setitems(to_store)
+
+    def _chunk_is_empty(self, chunk) -> bool:
```
d-v-b, line 1899: consider adding a code path where we check if the fill value is Falsey, and use np.any, which has some optimizations for this operation.

cc @jakirkham

joshmoore:
Updated to the working 2021.08.1 fsspec version.

jakirkham:
IDK if this is still an issue (ran into it previously), but `ZipStore` lacks a `__delitem__` implementation atm due to the lack of this functionality on zip files. One workaround would be to write empty strings over them. Raised issue ( https://github.com/zarr-developers/zarr-python/issues/828 ) about this.

jni:
@d-v-b my 2c on your questions:


> I don't like the parameter name write_empty_chunks. I would be happy to rename it to something shorter / more expressive, e.g. sparse_write. But there are probably better ideas out there.

I like `sparse_write` a lot.

> Should write_empty_chunks (or whatever it's called) be a public attribute of zarr.core.Array?

imho, yes, keyword-only.

> Should I run some tests detailing the performance impact of the chunk checking?

For my own peace of mind this is unnecessary — the benefit of this is extremely clear to me.

d-v-b:
I think this is ready to go. I decided to keep the name `write_empty_chunks` because I felt that `sparse_write` ran the risk of making people think about sparse matrices or sparse array storage, which is another topic entirely. I also decided to set `write_empty_chunks=True` by default, because this minimizes any surprise people might experience with the new release. Maybe it's worth thinking about how to gracefully change this default to `False` if people find that desirable.

jakirkham:
Yeah it might be worth creating an RC for these newer features to get them in the hands of users and iterate on anything that needs improvement

jni:
> I chickened out and pushed v2.11.0a1

:chicken: 

> is out for testing

I can confirm that I can paint labels into zarr arrays in napari with 2.11.0a1! :tada: So much fun. :smiley: 

I'll let @d-v-b confirm that `write_empty_chunks=False` works well for him so you can merge #853 and release .a2, @joshmoore! :stuck_out_tongue_closed_eyes: :joy:

PR #759, opened 2021-05-25
Update ABSStore for compatibility with newer `azure.storage.blob`.

TomAugspurger:
This continues https://github.com/zarr-developers/zarr-python/pull/620 to update the Azure `ABSStore` to be compatible with newer versions of `azure.storage.blob`.

In addition to @jhamman's compatibility changes, I made a few changes to make this easier to test. The previous API had users pass various pieces of information like the container name, storage account name, account key, etc. Now we just accept an instance of a `BlobServiceClient`. I've deprecated the old way.

```python
>>> ABSStore("my_container", account_name="account_name", account_key="account_key")  # old
>>> ABSStore(client=azure.storage.blob.ContainerClient(...)  # new
```

I figure the less we care about the details of `azure.storage.blob` the better.

I haven't attempted a compatibility layer with older versions of azure.storage.blob. IMO, it's not worth the effort. I think the common case of users passing a storage account name, a container name, and an account key will continue to work fine (with a warning).

Closes https://github.com/zarr-developers/zarr-python/issues/618.
 
TODO:


TomAugspurger:
cc @tjcrone.

zarr/storage.py
```diff
@@ -2211,50 +2211,84 @@ class ABSStore(MutableMapping):
...
+                 ):
         self._dimension_separator = dimension_separator
-        if blob_service_kwargs is not None:
-            self.blob_service_kwargs = blob_service_kwargs
-        else:  # pragma: no cover
-            self.blob_service_kwargs = dict()
-        self.client = BlockBlobService(self.account_name, self.account_key,
-                                       **self.blob_service_kwargs)
-
-    # needed for pickling
```
TomAugspurger, line 2247: I believe that `ContainerClient` instances are typically pickleable.

```python
In [4]: cc = azure.storage.blob.ContainerClient("https://daymeteuwest.blob.core.windows.net", container_name="daymet-zarr")

In [6]: pickle.loads(pickle.dumps(cc))
Out[6]: <azure.storage.blob._container_client.ContainerClient at 0x7f5ce149ae80>
```

zarr/storage.py
```diff
@@ -2308,63 +2341,52 @@ def __iter__(self):
...
             dir_path += '/'
-        items = list()
-        for blob in self.client.list_blobs(self.container, prefix=dir_path, delimiter='/'):
-            if type(blob) == Blob:
-                items.append(self._strip_prefix_from_path(blob.name, dir_path))
-            else:
-                items.append(self._strip_prefix_from_path(
-                    blob.name[:blob.name.find('/', len(dir_path))], dir_path))
-        return items
+        items = set()
```
TomAugspurger, line 2358: @jhamman is there a reason you used a `set()` here? Are we relying on `items.add` below to maybe deduplicate something?

If not, I assume that just using a list from the beginning (as required by the `listdir` return value) will be faster.

zarr/tests/test_core.py
```diff
@@ -1623,16 +1623,13 @@ def test_nbytes_stored(self):
 
 
 @skip_test_env_var("ZARR_TEST_ABS")
+@pytest.mark.usefixtures("azurite")
 class TestArrayWithABSStore(TestArray):
 
     @staticmethod
     def absstore():
-        asb = pytest.importorskip("azure.storage.blob")
```
TomAugspurger, line 1630: There's a slight change here: previously, setting `ZARR_TEST_ABS` but not having the dependencies would silently pass thanks to the `importorskip`. Now if you set `ZARR_TEST_ABS` we assume you want those tests to run, and fail if the dependencies aren't there.

zarr/tests/util.py
```diff
@@ -53,3 +53,24 @@ def skip_test_env_var(name):
...
     have_fsspec = False
+
+
+def abs_container():
+    from azure.core.exceptions import ResourceExistsError
+    import azure.storage.blob as asb
+
+    URL = "http://127.0.0.1:10000"
+    ACCOUNT_NAME = "devstoreaccount1"
+    KEY = "Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw=="
```
TomAugspurger, line 64: These are the defaults for azurite, the azure storage emulater.

jakirkham:
cc @shikharsg @jhamman

jakirkham:
Welcome Tom 😄 Thanks for working on this!

cc'd a couple of people who may be more familiar with this part of the code (and hopefully able to review)

zarr/storage.py
shikharsg: if `delimiter` is specified to `walk_blobs` there is no need for deduplication (although I can see how the Azure Blob API docs seem ambigous), so should be fine with list instead of set IMO.

shikharsg:
Why isn't CI running for this PR? Can I click on "Approve and run"?

TomAugspurger:
GitHub requires maintainer approval before running CI on PRs from first-time contributors (https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks), to prevent abuses like mining cryptocurrencies.

So you can click it as long as you trust I haven't inserted some crypto-mining code into the CI workflow :)

zarr/tests/conftest.py
```diff
@@ -0,0 +1,19 @@
...
+    azurite = client.containers.run(
+        "mcr.microsoft.com/azure-storage/azurite",
+        "azurite-blob --loose --blobHost 0.0.0.0",
+        detach=True,
+        ports={"10000": "10000"},
+    )
+    print("Successfully created azurite container...")
+    yield azurite
+    print("Teardown azurite docker container")
+    azurite.stop()
```
shikharsg, line 19: Currently this file is what is breaking the tests.

I think a azure blob emulator is already running on localhost port 10000 if you see the Github actions workflow [file](https://github.com/zarr-developers/zarr-python/blob/master/.github/workflows/python-package.yml). So not sure this is needed, you can try and remove it I guess?

TomAugspurger: Thanks. I've updated the things to remove the fixture and just rely on azurite being running. I added the docker command to the contributing docs.

zarr/storage.py
shikharsg: If you see the tests [here](https://github.com/zarr-developers/zarr-python/actions/runs/875115873) (even the ones that were cancelled) you'll see that the pickling has worked, but only for python >=3.7. For python <3.7 the pickling did not work. `TypeError: can't pickle _thread.RLock objects`. A little googling around seems to suggest that it will not work for python 3.6. So not sure what's the best way around this.

TomAugspurger: Ah, thanks for that. We might be able to work around this by pickling the pieces of information needed to recreate the client, but I don't know if that's worth the maintenance effort since it's fixed in 3.7 and newer.

zarr/storage.py
jhamman: I think I agree with @shikharsg. It has been quite a while and I honestly don't recall my rational on this one.

zarr/storage.py
TomAugspurger: https://github.com/zarr-developers/zarr-python/pull/759/commits/5a3307cac910eb981b99d3ca5f8c12ea38b03f77 and https://github.com/zarr-developers/zarr-python/pull/759/commits/a1e0d0c67d2d716255cdb5c07d852f27e4570f4d just skip the tests for 3.6 and earlier. If desired, I can implement the compatibility code for py36, with the caveat that it won't be 100% robust (but likely not worse than before).

shikharsg:
odd error on this build: https://github.com/zarr-developers/zarr-python/pull/759/checks?check_run_id=2674084907
never seen this before, any idea @jakirkham @Carreau  ?

TomAugspurger:
I merged master, in case it was just a transient issue.

TomAugspurger:
Ugh, sorry, flake8 is passing now.

TomAugspurger:
The coverage issue should be fixed by https://github.com/zarr-developers/zarr-python/pull/759/commits/382d45340f1da9faa54e778ea9236ebe59b85e0a (which turned up a bug in the deprecation code)

jakirkham:
Interesting. Had to approve CI to run again. Wonder if GH forgot this was ok test

TomAugspurger:
> Interesting. Had to approve CI to run again. Wonder if GH forgot this was ok test

I think someone (@shikharsg maybe) has been approving the runs on each commit. Sorry about all the failing commits, but things are green now.

TomAugspurger:
Thanks for the reviews so far. Anything else I can do to help move this along?

joshmoore:
Yeah, I've been approving the running of tests as well. I've not yet understood when it sticks and when it doesn't. (Perhaps because several of us are doing it, it's had less rhyme or reason). I'll look into a repo setting.

As for moving it along, I think it's just waiting for a review, no? Anyone in particular come to mind?

shikharsg (commented):
I think this is good to go in now @TomAugspurger  @joshmoore

docs/contributing.rst
```diff
@@ -164,7 +164,11 @@ optional dependencies to be installed), run::
 Note that some tests also require storage services to be running
 locally. To run the Azure Blob Service storage tests, run an Azure
 storage emulator (e.g., azurite) and set the environment variable
-``ZARR_TEST_ABS=1``. To run the Mongo DB storage tests, run a Mongo
+``ZARR_TEST_ABS=1``. If you're the Docker to run azurite, start the service with::
```
joshmoore, line 167:
```suggestion
``ZARR_TEST_ABS=1``. If you're using Docker to run azurite, start the service with::
```

docs/release.rst
```diff
@@ -11,6 +11,9 @@ Bug fixes
 
 * FSStore: default to normalize_keys=False
   By :user:`Josh Moore <joshmoore>`; :issue:`755`.
+* ABSStore: compatibility with ``azure.storage.python>=12``
```
joshmoore, line 14: Thank you for adding this!

joshmoore (changes requested):
I saw one minor typo in the docs, but I'll commit myself and merge. Thanks all!

jakirkham:
> Yeah, I've been approving the running of tests as well. I've not yet understood when it sticks and when it doesn't. (Perhaps because several of us are doing it, it's had less rhyme or reason). I'll look into a repo setting.

I think it is just because Tom wasn't yet a contributor here. That should be a non-issue going forward

Issue #601, opened 2020-09-11
zarr.open and friends should handle Path objects


clbarnes:
Strictly, anything which doesn't cause `os.fspath` to raise a TypeError.

#### Minimal, reproducible code sample, a copy-pastable example if possible

```python
import zarr
from pathlib import Path

path_str = "path/to/array.zarr"

zarr.open(path_str)  # works fine
zarr.open(Path(path_str))  # fails with TypeError because it's not a str and not iterable
```

#### Problem description

pathlib is great and everyone should use it whenever appropriate. Personally, I even subclass PurePosixPath for manipulations of object names inside hdf5/ zarr/ n5 containers.

People with zarr objects on the file system should want to use `Path`s to refer to them, so zarr should support people doing so.

This should just be a case of trying `os.fspath` on the store/path argument(s) for most of the functions in `convenience`.

#### Version and installation information

Please provide the following:

* Value of ``zarr.__version__``: 2.4.0
* Value of ``numcodecs.__version__``: 0.6.4
* Version of Python interpreter: 3.8.5
* Operating system (Linux/Windows/Mac): Linux
* How Zarr was installed (e.g., "using pip into virtual environment", or "using conda"): pip in virtual environment

clbarnes:
Ah, I see zarr still supports 3.5, which has pathlib but doesn't support it in much of the standard library, and doesn't have fspath. This can hold off a couple of days until 3.5 is EOL, then.

clbarnes:
Released in 2.9.0: https://zarr.readthedocs.io/en/stable/release.html#enhancements

Issue #261, opened 2018-05-10
open...() functions should accept pathlib.Path


alimanfoo:
#### Minimal, reproducible code sample, a copy-pastable example if possible

```python
from pathlib import Path
store = Path('/path/to/store')
zarr.open_group(store)
```
...gives:
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-8-4a4272aec66c> in <module>()
----> 1 codon_position = zarr.open_group(output_dir / 'codon_position.zarr.zip')
      2 codon_position.tree()

~/github/wtsi-team112/vector-ops/deps/conda/envs/vector-ops/lib/python3.6/site-packages/zarr/hierarchy.py in open_group(store, mode, cache_attrs, synchronizer, path)
   1117 
   1118     elif mode == 'a':
-> 1119         if contains_array(store, path=path):
   1120             err_contains_array(path)
   1121         if not contains_group(store, path=path):

~/github/wtsi-team112/vector-ops/deps/conda/envs/vector-ops/lib/python3.6/site-packages/zarr/storage.py in contains_array(store, path)
     70     prefix = _path_to_prefix(path)
     71     key = prefix + array_meta_key
---> 72     return key in store
     73 
     74 

TypeError: argument of type 'PosixPath' is not iterable
```

#### Problem description

Functions that accept a string path for the store argument should also accept Path object.

#### Version and installation information

Please provide the following:

* Value of ``zarr.__version__``: 2.2.0
* Version of Python interpreter: 3.6

jakirkham:
Imagine @martindurant has thoughts on this one. 😉

martindurant:
Yeah...
Having filenames as a possibility is a fairly natural convenience, but the real input for zarr functionality is a key-value mapper over some file-system. 

Maybe I'm just too old for pathlib, but wouldn't it be easier for the user to coerce the object to str in the line passing it to zarr rather than zarr having to check for another condition?

nehz:
Passing pathlib paths around is becoming more common, at least the error message should be something more sane I think (e.g TypeError)

martindurant:
Note that fsspec does support Path objects now, except they don't really make sense for anything non-local. From zarr's point of view, a Path should probably be assumed to be local, and so invoke the directory-store directly rather than going via fsspec.

LunarLanding:
Most software I've been using accepts path objects. Would a minimal pull request be accepted?

[...]

PR #768, opened 2021-06-04
Pathlib support

clbarnes:
Minor changes to utility functions to allow convenience functions to work with `os.PathLike`s; see #261 and #601. These now just normalise to `str`, which the DirectoryStore was doing explicitly before anyway. It might be nice to avoid that and for the `path` member of things containing a path to store (/return) it as a `Path`, but that was a more invasive change; the main value here is avoiding the errors.

One module defined a `Path` newtype for typing purposes; I renamed that to `PathType` for clarity.

I found it a little confusing to see "path" used to describe zarr's internal keys; looking through the code, you have to get whether it's talking about key or e.g. file system path from context. Would changing that verbiage be considered in an upcoming major version increment? Another "going forward" thought: it's a bit of a pain/ source of errors to manipulate path-like things as strings, which is why pathlib exists. Would you consider something like

```python
from pathlib import PurePosixPath

class ZarrKey(PurePosixPath):
    pass
```

for internal use to begin with? On the flip side, we'd need to be more careful about `os.PathLike` checks, but there are only a couple in there.

Lastly, where in `release.rst` do new features go? There doesn't seem to be an entry for "current release" features.

TODO:
    - Nothing to document as I would assume this behaviour from anything interacting with the file system

### DIFF
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -20,7 +20,6 @@ import glob
 import multiprocessing
 import operator
 import os
-from pathlib import Path
 import re
 import shutil
 import sys
@@ -78,7 +77,7 @@ except ImportError:  # pragma: no cover
     default_compressor = Zlib()
 
 
-Path = Union[str, bytes, None]
+PathType = Union[str, bytes, None]
 
 
 def _path_to_prefix(path: Optional[str]) -> str:
@@ -90,7 +89,7 @@ def _path_to_prefix(path: Optional[str]) -> str:
     return prefix
 
 
-def contains_array(store: MutableMapping, path: Path = None) -> bool:
+def contains_array(store: MutableMapping, path: PathType = None) -> bool:
     """Return True if the store contains an array at the given logical path."""
     path = normalize_storage_path(path)
     prefix = _path_to_prefix(path)
@@ -98,7 +97,7 @@ def contains_array(store: MutableMapping, path: Path = None) -> bool:
     return key in store
 
 
-def contains_group(store: MutableMapping, path: Path = None) -> bool:
+def contains_group(store: MutableMapping, path: PathType = None) -> bool:
     """Return True if the store contains a group at the given logical path."""
     path = normalize_storage_path(path)
     prefix = _path_to_prefix(path)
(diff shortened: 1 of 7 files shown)


joshmoore:
> Would you consider something like ... `class ZarrKey(PurePosixPath)` for internal use to begin with?

Interesting. I'd be especially happy if I could attach extra metadata to that ;) see https://github.com/zarr-developers/zarr-python/issues/769#issuecomment-861719216

> Lastly, where in `release.rst` do new features go?

Just add to the top. (The mainline may now have an "Unreleased" section at the top)

clbarnes:
Added documentation to release.rst. Build failures are (I hope) unrelated; a segfault coming from numpy.

clbarnes:
Rebased and green!

zarr/storage.py
```diff
@@ -76,7 +76,7 @@
     default_compressor = Zlib()
 
 
-Path = Union[str, bytes, None]
+PathType = Union[str, bytes, None]
```
joshmoore, line 79: This is probably the only change that I don't feel competent to individually review.

Does anyone know the possible impact of this? In my experience, this would count as a breaking change.

joshmoore (commented):
In general, looks good from my side. I've noted one renaming issue on which I'd appreciate thoughts from others.

zarr/storage.py
clbarnes: This could be reverted. From its usage it looked pretty internal (and only in the one module) but I suppose someone downstream could be using it.

joshmoore: > but I suppose someone downstream could be using it.

*nods* That was my thought. I have _zero_ evidence that someone is actually doing that, but just in terms of the semvers'ioning

clbarnes:
Now rebased on master; I've also reverted the `Path` rename and have used `import pathlib; ... pathlib.Path` everywhere it's used to be maximally explicit.

joshmoore (approved):
Thanks for the rebase, @clbarnes! To my eye, this is now a very _non_-invasive widening of what the open methods will accept.

PR #773, opened 2021-06-14
Fix DirectoryStore

### LINKED ISSUE #769: Confusion about the dimension_separator keyword
I don't really understand how to use the new `dimension_separator` keyword, in particular:
1. Creating a `DirectoryStore(dimension_separator="/")` does not have the effect I would expect (see code and problem description below).
2. Why does zarr still have the `NestedDirectoryStore`? Shouldn't it be the same as `DirectoryStore(dimension_separator="/")`? Hence I would assume that `NestedDirectoryStore` could either be removed or (if to be kept for legacy purposes) should just map to `DirectoryStore(dimension_seperator="/")`.

```python
import zarr
store = zarr.DirectoryStore("test.zarr", dimension_separator="/")
g = zarr.open(store, mode="a")
ds = g.create_dataset("test", shape=(10, 10, 10))
ds[:] = 1
```

#### Problem description

Now, I would assume that the chunks are nested, but I get:
```
$ ls test.zarr/test
0.0.0
```
but to, to my confusion, also this:
```
$ cat test.zarr/test/.zarray
...
"dimension_separator": "/",
...
```
If I use `NestedDirectoryStore` instead, the chunks are nested as expected.

Please provide the following:

* Value of ``zarr.__version__``:  2.8.3
* Value of ``numcodecs.__version__``: 0.7.3
* Version of Python interpreter: 3.8.6
* Operating system: Linux
* How Zarr was installed: using conda

joshmoore:
fix #769

Permit DirectoryStore to open Zarr filesets which have been saved with `{"dimension_separator": "/"}`. Currently they silently fallback to using the `fill_value`.

TODO:


joshmoore:
Build failures are unrelated. See https://github.com/zarr-developers/zarr-python/pull/774

zarr/tests/test_storage.py
```diff
@@ -1144,10 +1165,10 @@ def test_chunk_nesting(self):
         # any path where last segment looks like a chunk key gets special handling
         store['0.0'] = b'xxx'
         assert b'xxx' == store['0.0']
-        assert b'xxx' == store['0/0']
+        # assert b'xxx' == store['0/0']
```
martindurant, line 1168: Why did we loose there? Doesn't the store normalise "." -> "/" anyway?

joshmoore: I tried to describe in the commit message on https://github.com/zarr-developers/zarr-python/pull/773/commits/e1835667116b3f86fe002183a6b527565743795c, at least to the best of my understanding.

jakirkham:
> Build failures are unrelated. See #774

That's merged. Would merge `master` back into this PR (as I don't think GH Actions does that itself)

joshmoore:
Pushed more tests based on the blurb from @martindurant in #769. Note that without the fix in this PR these are the failures:

```
zarr/tests/test_dim_separator.py::test_nested[static_flat] FAILED
zarr/tests/test_dim_separator.py::test_nested[directory_flat] FAILED
zarr/tests/test_dim_separator.py::test_nested[directory_default] FAILED
zarr/tests/test_dim_separator.py::test_nested[fs_flat] FAILED
zarr/tests/test_dim_separator.py::test_nested[fs_default] FAILED
```

zarr/core.py
```diff
@@ -1952,7 +1952,7 @@ def _process_for_setitem(self, ckey, chunk_selection, value, fields=None):
         return self._encode_chunk(chunk)
 
     def _chunk_key(self, chunk_coords):
-        return self._key_prefix + '.'.join(map(str, chunk_coords))
+        return self._key_prefix + self._dimension_separator.join(map(str, chunk_coords))
```
d-v-b, line 1955: Can you explain how this change is compatible with `FSStore._normalize_key`? https://github.com/zarr-developers/zarr-python/blob/master/zarr/storage.py#L1076

`FSStore._normalize_key` assumes that all chunk keys are formatted `foo/bar/0.0.0` -- this assumption is the basis of splitting the chunk key into a prefix and a chunk ID via `key.split('/')`. As I understand it, this change breaks this assumption.

joshmoore: Reading the `flat` and `nested` fixtures from this repo (`zarr.open(f"file:///tmp/{x}")[:]`) with some sloppy debugging in place shows:

```
|               | Array._chunk_key | FSStore._normalize_keys |
|---------------|------------------|-------------------------|
| master:nested |  (0, 0) --> 0.0  |       0.0 --> 0.0       |
| master:flat   |  (0, 0) --> 0.0  |       0.0 --> 0.0       |
| PR:nested     |  (0, 0) --> 0/0  |       0/0 --> 0/0       |
| PR:flat       |  (0, 0) --> 0.0  |       0.0 --> 0.0       |
```

which likely points to some logic in FSStore being ripe for removal since the Store is basically just accepting what what the Array has detected. Now, how it is that that's working with _your_ PR, I still haven't figured out.

d-v-b: Right, as your test shows this is fine for FSStore (and maybe we don't need this code in the store at all if the chunk keys come pre-normalized). But this situation is dire for N5Stores, which need to be able to re-order the chunk keys before writing to storage.

joshmoore:
See the extended discussion with @d-v-b at https://gitter.im/zarr-developers/community?at=611e5f31a1ffab59400448cc. Current proposal is to have N5 stores _act_ like they are still using `"."` internally in order to allow them to differentiate between Zarr-like keys and N5-like keys and then convert to `"/"`-based keys just before writing. A result of this is that you can _hack_ access to the true N5 locations by using `"/"`-based keys as questioned by @martindurant in https://github.com/zarr-developers/zarr-python/pull/773#discussion_r651904517

joshmoore:
Please see the additional evidence of https://github.com/zarr-developers/zarr_implementations/pull/47 (and please ignore the code coverage failure behind the curtain... 🧙🏽 )

joshmoore:
Taking @d-v-b's :thumbsup: on https://github.com/zarr-developers/zarr-python/pull/773#issuecomment-901931941 as a sign-off.  Merging and starting to prep a release.

PR #778, opened 2021-06-17
Drop the additional numpy 1.16 test

joshmoore:
Builds that are pinned to the 1.16.x numpy series are starting to fail. Dropping those in favor of 1.17.x numpy to keep as much backwards compatibility as possible.

see:
 * #774 which drops Python 3.6
 * #749 which bumps h5py to 3.2

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       matrix:
         python-version: [3.7, 3.8]
-        numpy_version: ['']
+        numpy_version: ['', '==1.17.*']
     services:
       redis:
         image: redis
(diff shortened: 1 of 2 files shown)


joshmoore:
All tests are green except for the one 1.16.x build which was removed here (and therefore is blocked)

Merging to try to get other PRs green. If there are any objections to dropping this particular test, we can resurrect without impacting end-users.

jakirkham:
NumPy 1.20 is the current version (and has already seen a few patch releases). Plus NumPy 1.21 already had RC 2 come out over a week ago (so is probably coming out soon). Given the pace of NumPy releases, where NumPy is at version-wise, and the general ease of installing NumPy these days, bumping to 1.17 seems pretty reasonable.

PR #789, opened 2021-06-23
Create a Base store class for Zarr Store (update)

grlee77:
# Use a base Store class instead of just MutableMapping

I am resuming the work @Carreau started in #612 here. This is mostly just a rebase of that PR with a small new commit in 8136713. Matthias and I met earlier today and we decided that starting with this PR should make it easier to add zarr-v3 support without trying to implement an entirely independent set of classes. The goal is to reduce code duplication between v2 and v3 where possible (None of the changes in this PR are specific to v3 yet)

Note that I updated the version string text to `2.9.0` in a couple of places, but that may need to be updated again depending on when this is merged.

TODO:

### DIFF
--- a/zarr/tests/test_core.py
+++ b/zarr/tests/test_core.py
@@ -23,11 +23,12 @@ from zarr.storage import (
     ABSStore,
     DBMStore,
     DirectoryStore,
+    FSStore,
+    KVStore,
     LMDBStore,
     LRUStoreCache,
     NestedDirectoryStore,
     SQLiteStore,
-    FSStore,
     atexit_rmglob,
     atexit_rmtree,
     init_array,
@@ -44,8 +45,8 @@ class TestArray(unittest.TestCase):
     def test_array_init(self):
 
         # normal initialization
-        store = dict()
-        init_array(store, shape=100, chunks=10, dtype='<f8')
+        store = KVStore(dict())
+        init_array(store, shape=100, chunks=10, dtype="<f8")
         a = Array(store)
         assert isinstance(a, Array)
         assert (100,) == a.shape
@@ -55,9 +56,10 @@ class TestArray(unittest.TestCase):
         assert a.basename is None
         assert store is a.store
         assert "8fecb7a17ea1493d9c1430d04437b4f5b0b34985" == a.hexdigest()
+        store.close()
 
         # initialize at path
-        store = dict()
+        store = KVStore(dict())
         init_array(store, shape=100, chunks=10, path='foo/bar', dtype='<f8')
         a = Array(store, path='foo/bar')
         assert isinstance(a, Array)
@@ -101,8 +103,7 @@ class TestArray(unittest.TestCase):
             if not isinstance(k, expected_type):  # pragma: no cover
                 pytest.fail("Non-text key: %s" % repr(k))
 
-        if hasattr(z.store, 'close'):
-            z.store.close()
+        z.store.close()
 
     def test_store_has_binary_values(self):
         # Initialize array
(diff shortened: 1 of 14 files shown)


joshmoore:
The `Fatal Python error: Segmentation fault` was also seen in https://github.com/zarr-developers/zarr-python/pull/788#issuecomment-866595788

grlee77:
Yes, I just started looking into it. I can reproduce it locally after upgrading to NumPy 1.21. I does not occur for NumPy 1.20.3

grlee77:
I opened numpy/numpy#19325 regarding this segfault

jakirkham:
cc @martindurant @rabernat @jhamman @shoyer @shikharsg (in case any of you have thoughts here 🙂 this will affect how other store implementations work 😉)

jakirkham:
Also cc @TomAugspurger 🙂

grlee77:
Looking ahead to adding v3 spec support:

In terms of class members for a `Store` interface, [the v3 spec indicates the following should be present](https://zarr-specs.readthedocs.io/en/core-protocol-v3.0-dev/protocol/core/v3.0.html#abstract-store-interface):

- readable stores must have: `get`
- writeable stores must have : `set`, `erase`, `erase_prefix`
- listable stores should have one or more of: `list`, `list_dir`, `list_prefix`:

The `Store` class here matches the existing store implementations in `zarr-python` in that it has a `listdir` method, but this differs from the `list_dir` spelling in the v3 spec which is unfortunate.  `erase` seems equivalent to `__delitem__ ` while `erase_prefix` seems equivalent to `rmdir`

joshmoore:
Green now except for the removed build (==1.16.4) which can be ignored.

martindurant:
Clarification: if we open an array or group, and pass a dict-like as before, it gets wrapped to become a Store, right?

grlee77:
> Clarification: if we open an array or group, and pass a dict-like as before, it gets wrapped to become a Store, right?

This is the intention (via the `Store._ensure_store` method). We may need some additional tests here to make sure this is being done properly everywhere.

zarr/tests/test_core.py
```diff
@@ -102,8 +104,7 @@ def test_store_has_text_keys(self):
             if not isinstance(k, expected_type):  # pragma: no cover
                 pytest.fail("Non-text key: %s" % repr(k))
 
-        if hasattr(z.store, 'close'):
```
joshmoore, line 105: A sidenote that there's a issue somewhere for having `close()` on all stores.

joshmoore (commented):
As mentioned on call, this is looking good. Thanks, @grlee77. Probably the most important remaining step is to get the release blurb so that we can all agree on the degree of breakage and have a comprehensive list of what (if anything) will no longer take `dict`s. Additionally some documentation may be necessary around the "pass by reference" semantics.

Propose then to target this for a 2.11

joshmoore:
Solved the (minor) merge conflict, @grlee77. From my side, I'd say let's try to get this into a RC as early as tomorrow. (Or someone else can do it this evening. I'm going to 🛏️ ) Up to you if you think we should also try to get metadata handling in.

(Release blurb would still be appreciated.)

grlee77:
> Solved the (minor) merge conflict, @grlee77. From my side, I'd say let's try to get this into a RC as early as tomorrow. (Or someone else can do it this evening. I'm going to bed ) Up to you if you think we should also try to get metadata handling in.

I think we may as well also merge the metadata handling for the RC. Let's discuss in the community call. 

One remaining doubt about the approach here is the presence of `listdir` and `rmdir` methods in this class which seems specific to our current v2 store implementations. They are not among the [abstract store methods listed in the v3 spec](https://zarr-specs.readthedocs.io/en/core-protocol-v3.0-dev/protocol/core/v3.0.html#abstract-store-interface) (there is a `list_dir`, not `listdir` but behavior is not exactly the same. `erase_prefix` in the spec is like `rmdir`). I am wondering if we should instead hava a `BaseStore` that does not have these methods and then use a StoreV2 subclass that would add these? 

A future StoreV3 class could then inherit from BaseStore instead of StoreV2, so that it would not have to inherit additional methods that are not part of the spec. That said, at the moment many high-level functions use `listdir` and `rmdir`, so we may want to keep them around, at least temporarily.

PR #793, opened 2021-07-06
N5FSStore

d-v-b:
An fsstore-backed implementation of n5

There's a lot of ugly duplicated logic in `n5.py`. It might be worth some time cleaning that up.

cc @joshmoore, in case you want to test dimension_separator stuff against this branch 

TODO:
* [] Add unit tests and/or doctests in docstrings


joshmoore:
Note that #773 is still passing against this branch. Any thoughts on moving forward here, @d-v-b, re: https://github.com/zarr-developers/zarr-python/issues/769#issuecomment-897112420?

d-v-b:
@joshmoore please try again, I discovered that this branch was missing N5FSStore tests in `test_core`, which were failing until I made some changes. I'm _pretty sure_ anything that changes the chunk keys on the array side is going to break what I have here.

joshmoore:
Good morning, @d-v-b. The merged branches passed for me locally. I've re-activated #794 for fuller testing.

zarr/tests/test_storage.py
```diff
@@ -1207,7 +1207,7 @@ def test_chunk_nesting(self):
         assert 'foo/10.20.30' in store
         assert b'yyy' == store['foo/10.20.30']
         # N5 reverses axis order
-        # assert b'yyy' == store['foo/30/20/10']
+        assert b'yyy' == store['foo/30/20/10']
```
joshmoore: Per https://github.com/zarr-developers/zarr-python/pull/773#pullrequestreview-684161254, I have a heard time understanding how it's possible to support the requirement that multiple keys be supported for the same item. 

This may be part and parcel of the conversation we're having in gitter: https://gitter.im/zarr-developers/community?at=611d562e933ee46b7564044c

joshmoore: Update: ah, now I see that they are re-ordered not just `.`/`/` swapped, but I think my same concern holds.

d-v-b: I agree with the concern -- it looks like this behavior is expected from the NestedDirectoryStore: https://github.com/zarr-developers/zarr-python/blob/master/zarr/tests/test_storage.py#L1148

joshmoore: On my PR I resurrected this. I at least understand know _why_ it's supported, though I think it's a hackdoor of sorts.

d-v-b:
@joshmoore the remaining test failures are a) hexdigest related, (which I think you mentioned are due to something going on with numcodecs) and b) a lingering nested directory store test which I can comment out if needed / whenever I merge with master after your directory store fix goes in.

joshmoore:
Sounds good. If you want to test a reversion of the hexdigests, you can pin numcodecs to `<0.9`.  I'm unsure of whether or not to try fixing the NestedDirectoryStore issue. Nothing is failing at the moment so I'm tempted to leave it as in #773.

joshmoore:
@d-v-b: I pushed a commit trying to get codecov to 100%. I'm still having difficulty getting the likes of:

```
elif is_chunk_key(key):
    key = self._swap_separator(key)
``
to trigger if you have any ideas.

**Comment by d-v-b** [comment]
@joshmoore I copied a `TestN5Store` method over into `TestN5FSStore`, which seems to have done the trick. An alternative to copying code would be to make `TestN5FSStore` inherit from `TestFSStore` _and_ `TestN5Store`, but I'm nervous about using multiple inheritance (and when I ran tests with multiple inheritance, I didn't uncover any surprising failures).

**Comment by joshmoore** [comment]
Sounds good. Three Friday evening points:

 * @zarr-developers/core-devs : any comments?
 * I'm still a _tiny_ bit nervous about having two versions but I guess deprecation might eventually be our friend.
 * What version do you propose? patch or minor?

**Comment by d-v-b** [comment]
> I'm still a tiny bit nervous about having two versions but I guess deprecation might eventually be our friend.

By "two versions" do you mean the N5Store and N5FSStore? Given that a) n5 is relatively niche, and this was flagged as experimental, and b) fsspec isn't a super heavy dependency, I agree that a quick deprecation path looks nice. 

Version-wise I could think this is minor? since we are adding to the public API.

File: zarr/storage.py [hunk 2]
```diff
@@ -1065,22 +1065,28 @@ class FSStore(MutableMapping):
         Separator placed between the dimensions of a chunk.
     storage_options : passed to the fsspec implementation
     """
+    array_meta_key = array_meta_key
```
joshmoore, line 1068: Sorry, I missed these. Would it make sense to have these prefixed with `_`. Do you really mean these to be public API?

d-v-b: good question! I'm fine with making these private.

zarr/__init__.py
```diff
@@ -9,7 +9,7 @@
                            zeros_like)
 from zarr.errors import CopyError, MetadataError
 from zarr.hierarchy import Group, group, open_group
-from zarr.n5 import N5Store
+from zarr.n5 import N5Store, N5FSStore
```
joshmoore, line 12: This concerned me that it would need protecting by try/except block. In testing it, I realized FSStore only throws on `__init__` and therefore N5FSStore could be less conservative. I've pushed:

https://github.com/zarr-developers/zarr-python/pull/793/commits/aa4a723ecb9c776314f03f0a1abf14bacc10c215

zarr/storage.py
```diff
@@ -1065,22 +1065,28 @@ class FSStore(MutableMapping):
...
                  exceptions=(KeyError, PermissionError, IOError),
                  dimension_separator=None,
                  **storage_options):
         import fsspec
         self.normalize_keys = normalize_keys
+
+        protocol, _ = fsspec.core.split_protocol(url)
+        # set auto_mkdir to True for local file system
+        if protocol in (None, "file") and not storage_options.get("auto_mkdir"):
+            storage_options["auto_mkdir"] = True
```
joshmoore, line 1083: :heart:

joshmoore (approved):
Thanks for this, @d-v-b! Pushing forward with a minor release.

d-v-b:
Thanks for your help @joshmoore!

PR #812, opened 2021-08-20
2.9.0 release notes

joshmoore:
With #773 now merged (:tada:), I'd like to get this out (today? Even thought it's a Friday?!) but it's a fairly big one. Any thoughts? Would anyone want an RC to test with?

### DIFF
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -1,10 +1,58 @@
 Release notes
 =============
 
+.. _unreleased:
+
 Unreleased
 ----------
 
+.. _release_2.9.0:
+
+
 This release of Zarr Python is the first release of Zarr to not support Python 3.6.
+
+Enhancements
+~~~~~~~~~~~~
+
+* Update ABSStore for compatibility with newer `azure.storage.blob`.
+  By :user:`Tom Augspurger <TomAugspurger>`; :issue:`759`.
+
+* Pathlib support.
+  By :user:`Chris Barnes <clbarnes>`; :issue:`768`.
+
+Documentation
+~~~~~~~~~~~~~
+
+* Clarify that arbitrary key/value pairs are OK for attributes.
+  By :user:`Stephan Hoyer <shoyer>`; :issue:`751`.
+
+* Clarify how to manually convert a DirectoryStore to a ZipStore.
+  By :user:`pmav99 <pmav99>`; :issue:`763`.
+
+Bug fixes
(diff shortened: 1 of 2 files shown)


bnavigator:
Hi, I just tested to build your 2_9_0 branch for a package for openSUSE Tumbleweed (which has fsspec 2021.7.0) and I still get

```
[   89s] FAILED zarr/tests/test_core.py::TestArrayWithFSStoreNested::test_array_2d - A...
[   89s] FAILED zarr/tests/test_core.py::TestArrayWithFSStoreNestedPartialRead::test_array_2d
[   89s] FAILED zarr/tests/test_storage.py::TestFSStore::test_create - AttributeError:...
```

(#786, #802)

With #802 applied, the test_array tests pass, only test_create still fails.

Could you try to include #802 into the release and fix the last one (here or in fsspec) please?

bnavigator:
I just found out, that https://github.com/intake/filesystem_spec/pull/710 should fix the test_create failure.

joshmoore:
Hey @bnavigator. Thanks for giving this a try!

> I just found out, that [intake/filesystem_spec#710](https://github.com/intake/filesystem_spec/pull/710) should fix the test_create failure.

Agreed. Just waiting on the next release.

> Could you try to include #802 into the release and fix the last one (here or in fsspec) please?

Do you agree #802 is only needed if one bumps fsspec to 2021.7.0? So, if fsspec is pinned to 2021.6 or earlier (or one installs from the mainline) that 2.9.0 currently works?

bnavigator:
> Do you agree #802 is only needed if one bumps fsspec to 2021.7.0? So, if fsspec is pinned to 2021.6 or earlier (or one installs from the mainline) that 2.9.0 currently works?

Yes, but pinning to an earlier version is not an option for openSUSE Tumblewee. Other packages already triggered the update of fsspec to 2021.7 causing the python-zarr to fail. If you are not going to merge #802 into 2.9.0, I will have to apply the patch to our package manually.

joshmoore:
Agreed. Let's not include #802 just yet. I imagine an fsspec release will be along soon.

Merging and moving forward with the release.

bnavigator:
For the record: `test_core::*::test_array_2d` still fail on 2.9.0 with fsspec-2021.7 and https://github.com/intake/filesystem_spec/pull/710 applied to it. You still need #802 with a new fsspec release.

Issue #817, opened 2021-08-23
fsspec tests are not auto-skipped in test_dim_separator


QuLogic:
#### Problem description

Unlike other test files, the fsspec tests are not skipped in `test_dim_separator.py` automatically if it's not available.

#### Version and installation information

Please provide the following:

* Value of ``zarr.__version__`` 2.9.0
* Value of ``numcodecs.__version__`` 0.9.0
* Version of Python interpreter 3.10.0rc1
* Operating system (Linux/Windows/Mac) Fedora Rawhide
* How Zarr was installed (e.g., "using pip into virtual environment", or "using conda") from source

joshmoore:
Thanks, @QuLogic. Good catch. The danger of _always_ relying on `fsspec` in my local environment. :wink:

joshmoore:
Looks like this is also the cause of the failure in https://github.com/conda-forge/zarr-feedstock/pull/48

```
E       ModuleNotFoundError: No module named 'fsspec'
```

PR #819, opened 2021-08-24
Zarr 2.9.1: fix conda release

joshmoore:
see:
 * https://github.com/conda-forge/zarr-feedstock/pull/48
 * #817

The conda-forge release failed since fsspec is not installed by default. In order to get Zarr deployed ASAP (re: https://github.com/DiamondLightSource/python-tristan/pull/62) this simply properly handles the test issue that @QuLogic brought up in #817.

It might be that moving forward, fsspec should be installed & tested for conda-forge releases. If not, then a minimal test workflow should be added here to show that the conda-forge requirements (below) suffice to run or skip all tests:

```
channels:
  - conda-forge
  - defaults
dependencies:
  - python < 3.9.0
  - wheel
  - numcodecs
  - numpy
  - pip
  - pip:
    - asciitree
    - fasteners
    - pytest
    - setuptools_scm
```

### DIFF
--- a/zarr/tests/test_dim_separator.py
+++ b/zarr/tests/test_dim_separator.py
@@ -63,6 +63,7 @@ def test_open(dataset):
     verify(zarr.open(dataset))
 
 
+@pytest.mark.skipif(have_fsspec is False, reason="needs fsspec")
 def test_fsstore(dataset):
     verify(Array(store=FSStore(dataset)))
 
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -6,6 +6,14 @@ Release notes
 Unreleased
 ----------
 
+.. _release_2.9.1:
+
+Maintenance
+~~~~~~~~~~~
+
+* Correct conda-forge deployment of Zarr.
+  By :user:`Josh Moore <joshmoore>`; :issue:`XXX`.
+
 .. _release_2.9.0:
 
 


joshmoore:
Still ignoring codecov (opinions welcome) and moving forward with this to unblock conda.

zarr/tests/test_dim_separator.py
```diff
@@ -63,6 +63,7 @@ def test_open(dataset):
     verify(zarr.open(dataset))
```
benjaminhwilliams, line 63: I think there's [an additional problem](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=366681&view=logs&j=656edd35-690f-5c53-9ba3-09c10d0bea97&t=e5c8ab1d-8ff9-5cae-b332-e15ae582ed2d&l=2822) with `test_open[static_flat]` and `test_open[static_nested]`, which is also blocking the Conda release.  The Pytest working directory does not contain the expected paths `fixture/flat` and `fixture/nested`, which becomes clearer if you're stricter about the persistence mode you use in `test_open`:
```suggestion
    verify(zarr.open(dataset, "r"))
```

benjaminhwilliams (changes requested):
I'm not sure whether you welcome reviews from outsiders to the project, but I think there's slightly more to the Conda release failure than just catching absent `fsspec`.  See my suggestion.

zarr/tests/test_dim_separator.py
benjaminhwilliams: This has the added benefit of guarding against accidentally leaving any little droppings in the Pytest working directory.

joshmoore: Hi @benjaminhwilliams, I at least welcome it! ;) Would you be up for opening this as a PR?

benjaminhwilliams: Sure.  This small suggestion only highlights the issue, it doesn't fix the problem that the Pytest working directory is not the project root, as the test appears to assume.  But we can fix that by discussion in the PR.  I'll open it and request your review.

PR #835, opened 2021-09-19
Actions: add minimal.yml (close #820)

### LINKED ISSUE #820: Include fsspec in conda-forge testing?
See:
 * https://github.com/conda-forge/zarr-feedstock/pull/48
 * #819 

The 2.9.0 tag of zarr-python failed to build on conda-forge since [fsspec tests weren't skipped everywhere](https://github.com/zarr-developers/zarr-python/issues/817) and conda-forge installs only a minimal environment.

Either we should test the same environment in this repo or update the conda-forge metadata to install more requirements.

joshmoore:
Rather than adding fsspec as a dependency to conda-forge,
this tests a conda-forge style minimal build for each PR.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -3,8 +3,8 @@ channels:
   - defaults
 dependencies:
   - wheel
-  - numcodecs
-  - numpy
+  - numcodecs >= 0.6.4
+  - numpy >= 1.7
   - pip
   - pip:
     - asciitree
--- a/environment.yml
+++ b/environment.yml
@@ -1,0 +1,13 @@
+channels:
+  - conda-forge
+  - defaults
+dependencies:
+  - wheel
+  - numcodecs
+  - numpy
+  - pip
+  - pip:
+    - asciitree
+    - fasteners
+    - pytest
+    - setuptools_scm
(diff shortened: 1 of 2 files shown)


joshmoore:
Verified via https://github.com/zarr-developers/zarr-python/actions/runs/1250936738

Likely the largest point for discussion is whether or not we want a top-level `environment.yml`.

environment.yml
```diff
@@ -0,0 +1,13 @@
...
+dependencies:
+  - wheel
+  - numcodecs
+  - numpy
+  - pip
+  - pip:
+    - asciitree
+    - fasteners
+    - pytest
+    - setuptools_scm
```
jrbourbeau, line 13: We might also consider pinning to the minimum supported version of packages

https://github.com/zarr-developers/zarr-python/blob/5c0ef7ff56e957033bde3d6c57a485efb0490b52/setup.py#L9-L14

to make sure that Zarr works both when certain optional dependencies aren't installed, but also when using, for example, `numpy=1.7`. 

(not sure if there's another CI build which already tests this -- feel free to ignore me if there is 🙂 )

joshmoore: Pushed.

joshmoore:
Assuming this is good to go in. Either way, easy to update the actions as desired.

PR #839, opened 2021-09-24
Move metadata handling to a class (updated)

grlee77:
## Description

This PR is an update of #581 and builds upon #789 (only the last few commits, starting from 0e1fbe8 are new). This PR is not changing existing metadata function behavior, only moving the existing functions to class methods.

The `Store` from #789 gets a new `_metadata_class` attribute that other higher level functions can use to call functions that encode/decode metadata. To transition to v3 support, we would just make a new v3 version of this metadata class that overrides encode/decode methods to account for the changes in metadata format.

TODO:


joshmoore:
No vetoes during the community call yesterday. Rolling into `v2.11.0a2`.

Issue #841, opened 2021-09-28
normalize_keys regression in version 2.10.0


d70-t:
In dca87fcb (#793), the `normalize_keys` parameter default of `FSStore` has been changed back to from `False` to `True` ([see here](https://github.com/zarr-developers/zarr-python/blob/dca87fcbe96ff3be5d74e662815ecb5e67ce9b28/zarr/storage.py#L1072)). This reverts #755 and recreates the issue of #739.

I didn't find a comment on this in the commit message and believe this was done by accident.

joshmoore:
Thanks, @d70-t for all the sleuthing, and sorry if I let something slip through. @d-v-b: can you say if it was intentional or not?

d-v-b:
see #842

PR #842, opened 2021-09-28
set normalize_keys=False in fsstore constructor

d-v-b:
Set `normalize_keys` to `False` by default in `FSStore`. Addresses #841

### DIFF
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -1069,7 +1069,7 @@ class FSStore(MutableMapping):
     _group_meta_key = group_meta_key
     _attrs_key = attrs_key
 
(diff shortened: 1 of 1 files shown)


d-v-b:
I think if we are going to require that constructors take the same default argument, there should be a test for it (such a test would have prevented me from making this change in the first place). This would require changing all the store test classes, so I deemed it out of scope for this patch.

joshmoore (approved):
Thanks for the quick PR, @d-v-b. Sorry for the slow release (I was traveling! :wink:) Going to turn this into 2.10.1 now.

Regarding the constructor arguments, will it help if we define the defaults somewhere as constants?

PR #853, opened 2021-10-20
Set write_empty_chunks to default to False

jni:
In #738, despite general agreement that most people expect write_empty_chunks to default to False, @d-v-b decided to go the safe, uncontroversial route to get the PR merged, and made the default value True. All this PR does is set the default to False.

Convenient links to discussion suggesting write_empty_chunks=False should be the default:
- @meggart [says](https://github.com/zarr-developers/zarr-python/pull/738#issuecomment-848649951): "In Zarr.jl, *not* writing empty chunks is already the default [...] so I would be in favor of changing the default behavior in Python as well.
- @shoyer [says](https://github.com/zarr-developers/zarr-python/pull/738#issuecomment-848950903): "it is not clear to me why you would *not* want to do this. I think this could be the new default behavior."
- @martindurant [says](https://github.com/zarr-developers/zarr-python/pull/738#issuecomment-849697402): "I am also surprised that this is not the default, and would agree that I can't see people wanting to use the current behaviour ever."

There were performance concerns with FSSpec bulk delete items, but it [seems](https://github.com/zarr-developers/zarr-python/pull/738#issuecomment-902828611) that they were addressed upstream.

TODO:

### DIFF
--- a/zarr/util.py
+++ b/zarr/util.py
@@ -670,7 +670,7 @@ def all_equal(value: Any, array: Any):
         # optimized to return on the first truthy value in `array`.
         try:
             return not np.any(array)
-        except TypeError:  # pragma: no cover
+        except (TypeError, ValueError):  # pragma: no cover
             pass
     if np.issubdtype(array.dtype, np.object_):
         # we have to flatten the result of np.equal to handle outputs like
--- a/.github/workflows/minimal.yml
+++ b/.github/workflows/minimal.yml
@@ -25,11 +25,11 @@ jobs:
       run: |
         conda activate minimal
         python -m pip install .
-        pytest -svx
+        pytest -svx --timeout=300
     - name: Fixture generation
       shell: "bash -l {0}"
       run: |
         conda activate minimal
         rm -rf fixture/
-        pytest -svx zarr/tests/test_dim_separator.py zarr/tests/test_storage.py
+        pytest -svx --timeout=300 zarr/tests/test_dim_separator.py zarr/tests/test_storage.py
         # This simulates fixture-less tests in conda and debian packaging
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -73,7 +73,7 @@ jobs:
         conda activate zarr-env
         mkdir ~/blob_emulator
         azurite -l ~/blob_emulator --debug debug.log 2>&1 > stdouterr.log &
-        pytest --cov=zarr --cov-config=.coveragerc --doctest-plus --cov-report xml --cov=./
+        pytest --cov=zarr --cov-config=.coveragerc --doctest-plus --cov-report xml --cov=./ --timeout=300
     - uses: codecov/codecov-action@v1
       with:
         #token: ${{ secrets.CODECOV_TOKEN }} # not required for public repos
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -9,6 +9,9 @@ Unreleased
 Enhancements
 ~~~~~~~~~~~~
 
+* write_empty_chunks defaults to False.
+  By :user:`Juan Nunez-Iglesias <jni>`; :issue:`853`.
+
 * Allow to assign array ``fill_values`` and update metadata accordingly. :issue:`662`
 
 * array indexing with [] (getitem and setitem) now supports fancy indexing.
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -1320,6 +1320,59 @@ each chunk must be compared to the fill value, and these advantages are continge
 If you know that your data will form chunks that are almost always non-empty, then there is no advantage to the optimization described above. 
 In this case, creating an array with ``write_empty_chunks=True`` will instruct Zarr to write every chunk without checking for emptiness.
 
+The following example illustrates the effect of the ``write_empty_chunks`` flag on
+the time required to write an array with different values.::
+
+    >>> import zarr
+    >>> import numpy as np
+    >>> import time
+    >>> from tempfile import TemporaryDirectory
+
+    >>> def timed_write(write_empty_chunks):
+            """
+            Measure the time required and number of objects created when writing 
+            to a Zarr array with random ints or fill value.
+            """
+            chunks = (8192,)
+            shape = (chunks[0] * 1024,)
+            data = np.random.randint(0, 255, shape)
+            dtype = 'uint8'
+
+            with TemporaryDirectory() as store:
+                arr = zarr.open(store,
+                                shape=shape,
+                                chunks=chunks,
+                                dtype=dtype,
+                                write_empty_chunks=write_empty_chunks,
+                                fill_value=0,
+                                mode='w')
+                # initialize all chunks
+                arr[:] = 100
+                result = []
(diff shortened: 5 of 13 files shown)


d-v-b:
Sounds great to me :) Thanks for doing what I didn't have the courage to do.

jakirkham:
@joshmoore wdyt?

rabernat:
:+1: from me. I think this is a much better default.

jakirkham:
There are some lines scattered throughout the tests like these

https://github.com/zarr-developers/zarr-python/blob/edced8eda4930672bfe6365390e9f209af3089eb/zarr/tests/test_core.py#L91

https://github.com/zarr-developers/zarr-python/blob/edced8eda4930672bfe6365390e9f209af3089eb/zarr/tests/test_sync.py#L103

Do we want to update those as well?

jni:
> Do we want to update those as well?

🤷 All this means is that more tests will run with the default True. This might be useful for historical reasons. Both True and False are explicitly tested so I don't think we need to worry.

jni:
This is now rebased on the latest master.

joshmoore:
Per #901, @d-v-b was working on a blurb for blog/release notes/docs so that users are maximally informed of the change.

joshmoore:
Capturing [initial results](https://gitter.im/zarr-developers/community?at=61aaad3d197fa95a1ca55b39) from @d-v-b here: "users with "dense" data will see a ~15% increase in write times"

d-v-b:
N.B. that result doesn't include compression time. I will provide improved benchmarking code later today with a more realistic workflow.

d-v-b:
some interesting benchmarking results. see this repo for code: https://github.com/d-v-b/zarr-benchmarks

Each test writes a 64 MB chunk 128 times with `write_empty_chunks` drawn from `{True, False}` and the write value drawn from `{array of 0s, array of random ints}` , i.e. 4 different conditions. The plots show histrograms with an X axis that is the write duration normalized to the median of the time required under the `(write_empty_chunks=False, write_value=0)` condition.

First, with no compressor:
image

Next, with Blosc compression:
image
 
And last, with GZip compression:
image

Takeaways:
- The overhead associated with the emptiness checking is exposed with Blosc compression and no compressor.
- With blosc, writing empty chunks is actually faster than an emptiness check -> no write. 
- With GZip, handling empty chunks is much faster with `write_empty_chunks=False`, and the overhead associated with emptiness checking is not evident when writing "dense" chunks.

Happy to hear suggestions for changing the benchmark / plotting. E.g., maybe we want to show absolute duration instead of relative durations.

shoyer:
What sort of local disk are you using? I can imagine very different results for SSD vs HDD vs cloud object stores. Extra copies are never desirable, but in most cases I would guess the overhead of checking is acceptable.

d-v-b:
@shoyer this is on an SSD, so pretty ideal in terms of storage latency.

d-v-b:
@joshmoore I added content to the tutorial docs, let me know if you want more or an expanded release note.

joshmoore:
The location looks good. Could I suggest we try to get two or four simple writes with the different flags and inputs. Something like:

```
data = np.random.randint(0, 255, size=shape, dtype=DTYPE)
 arr = zarr.open(zarr.NestedDirectoryStore(store), 
                 write_empty_chunks=write_empty_chunks, ...)
 arr[:] = 255
```
from https://github.com/d-v-b/zarr-benchmarks/blob/main/src/zarr-benchmarks/empty_chunks.py with some form of marker ("15% slower") for someone to see at a glance.

> I'm not sure the benchmarks are that illuminating or surprising, so maybe we don't really need to involve them?

Agreed. Happy to have them in your repo. I find them pretty fascinating. ;)

jakirkham:
@d-v-b, did you have a chance to try the other flags Josh mentioned above? 🙂

d-v-b:
I added a short benchmark example, but I can't build the docs for some reason (`tox -e docs` yields some inscrutable error from a different part of the codebase)

d-v-b:
Oh and there's a new test failure in an unrelated part of the codebase. Not sure what's going on there.

joshmoore:
_Oy vey_. That's a new year for you. The new failures is related to the category filter. I don't remember anything that should impact it. Just starting to look at the docs failure.

(Thanks for the benchmark example!)

joshmoore:
Ok, @d-v-b, that should fix the docs build. Let's see what's failing on this run.

jni:
Hi both, I'm unclear about whether I need to do anything here, please ping me directly if you want me to update something. 😊

joshmoore:
Sorry for dropping the ball here. Don't think there's anything needed from your side, @jni. Just very unsure what's up with this failing test. Other than that, I'd very much like to get 2.11 out ASAP.

joshmoore:
Ok. Pushed a proposed fix related to @d-v-b's `all_equal`.

zarr/util.py
```diff
@@ -670,7 +670,7 @@ def all_equal(value: Any, array: Any):
         # optimized to return on the first truthy value in `array`.
         try:
             return not np.any(array)
-        except TypeError:  # pragma: no cover
+        except (TypeError, ValueError):  # pragma: no cover
```
jakirkham, line 673: What causes the `ValueError`? Is there a particular array type or value that `np.any` is raising on?

joshmoore: It occurs when a category value (`baz`) gets passed to `int()`.

jakirkham: Interesting. Looks like the error message and type changed from NumPy 1.21 to 1.22. Don't think that was intentional, but could be wrong. Raised issue ( https://github.com/numpy/numpy/issues/20898 )

In any event this workaround seems reasonable in the interim

joshmoore: Great, thanks! Once this is green, I'll move forward with 2.11.0.

jakirkham:
> Hi both, I'm unclear about whether I need to do anything here, please ping me directly if you want me to update something. 😊

If we can convince both you, @jni, and @d-v-b to write a few lines for the blogpost ( https://github.com/zarr-developers/zarr-python/issues/901 ), that would be fantastic! 😃

Can just post in that issue if you like

joshmoore:
> If we can convince both you, @jni, and @d-v-b to write a few lines for the blogpost ( #901 )

e.g. was just mentioning to @msankeys963 how cool the use case from https://forum.image.sc/t/data-store-and-library-backend-for-napari-plugin/61779 is ;)

(...._still_ waiting on actions...)

joshmoore:
Trying to reopen to get tests to complete.

joshmoore:
I've not been able to get the 3.8 and 3.9 tests to run on this PR yet. (3-4 retries)

jakirkham:
Looks like some of the past commits in this PR (though not all) had the same issue

joshmoore:
Oddly py 3.7 is still spinning here compared to #951

joshmoore:
Power cycling for CI.

jakirkham (approved):

joshmoore:
Holy moly that was an adventure! Looks like we're good to move forward with 2.11. (🍺 🍸 or similar all around this weekend)

Issue #858, opened 2021-10-22
test_nested[static_flat] fails on Debian sid Python 3.9


avalentino:
#### Minimal, reproducible code sample, a copy-pastable example if possible

```sh
$ python3.9 -m pytest

============================= test session starts ==============================
platform linux -- Python 3.9.7, pytest-6.2.5, py-1.10.0, pluggy-0.13.0
rootdir: /build/zarr-2.10.2+ds, configfile: pytest.ini
collected 2598 items

zarr/tests/test_attrs.py ........                                        [  0%]
zarr/tests/test_convenience.py ......................................... [  1%]
......................................                                   [  3%]
zarr/tests/test_core.py ................................................ [  5%]
........................................................................ [  7%]
....................................................................ssss [ 10%]
sssssssssssssssssssssssssssssssssssssssssss............................. [ 13%]
..................................................................ssssss [ 16%]
ssssssssssssssssssssssssssssssssssssssssss.............................. [ 19%]
.................sssssssssssssss.sssssssss.sssssssssssssssssssssssssssss [ 21%]
sssssss.sssssssss.ssssssssssssss.sssssssssssssssssssss.sssssssss.sssssss [ 24%]
ssssssssssssss.......................................................... [ 27%]
........................................................................ [ 30%]
........................................................................ [ 32%]
........................................................................ [ 35%]
........................................................................ [ 38%]
........................................................................ [ 41%]
.........sssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssss [ 43%]
ssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssss [ 46%]
sssssssssssssssssssssssssssssssssssssssssssssssssssssssssss              [ 49%]
zarr/tests/test_creation.py ..............                               [ 49%]
zarr/tests/test_dim_separator.py .........sssssssssssssss.........sssFx. [ 51%]
...x..sss                                                                [ 51%]
zarr/tests/test_filters.py .......                                       [ 51%]
zarr/tests/test_hierarchy.py ........................................... [ 53%]
....................sssssssssssssssssssss.....................ssssssssss [ 56%]
sssssssssssssssssssssssssssssssssss..................................... [ 58%]
.....ssssssssssssssssssssssssssssssssssssssssss......................... [ 61%]
............................................                             [ 63%]
zarr/tests/test_indexing.py .................................xx          [ 64%]
zarr/tests/test_info.py .                                                [ 64%]
[...]

PR #859, opened 2021-10-22
Activate Python 3.9

### LINKED ISSUE #629: Test on Python 3.9
Python 3.9 has been released, it is now time to start enabling it in CI once available.

joshmoore:
In an attempt to track down https://github.com/zarr-developers/zarr-python/issues/858
Closes #629

TODO:

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -47,7 +47,7 @@ jobs:
     - name: Create Conda environment with the rights deps
       shell: "bash -l {0}"
       run: |
-        conda create -n zarr-env python==${{matrix.python-version}} bsddb3 numcodecs==0.6.4 lmdb pip nodejs flake8 mypy
+        conda create -n zarr-env python==${{matrix.python-version}} bsddb3 numcodecs lmdb pip nodejs flake8 mypy
         conda activate zarr-env
         npm install -g azurite
     - name: Install dependencies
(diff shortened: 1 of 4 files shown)


joshmoore:
Ok. Now down to just the codecov issue. The next clue in debugging is that:

```
        nbytes, self.cbytes, blocksize = cbuffer_sizes(header)
```
appears to now return matching values for `nbytes` and `blocksize` with the numcodecs update. That feels significant. Not sure if related to https://github.com/zarr-developers/numcodecs/pull/283

andrewfulton9:
@joshmoore, sorry I missed my mention earlier. It looks like something in has changed to cause `nbytes` to always be the same as `blocksize`. It definitely could be [zarr-developers/numcodecs#283](https://github.com/zarr-developers/numcodecs/pull/283). I'll have to look closer to be sure though.

jakirkham:
Oh interesting, meaning that Blosc may have changed something, which is affecting us here?

andrewfulton9:
It looks like maybe Blosc has changed the default to just use 1 block if blocksize isn't defined so nbytes will equal blocksize by default, where previously it used to use some default (or tried to guess an optimum) blocksize. In `TestArrayWithFSStoreNested` I just used the [default block object](https://github.com/zarr-developers/zarr-python/blob/a92550b9ff9f4931353b4066486aaf9755895ce3/zarr/tests/test_core.py#L2711) so that is probably why nbytes always equals blocksize now. I can make a PR defining a blocksize there to see if that fixed it.

jakirkham:
Think that would be helpful. We'd like to keep the partial read functionality in good working order 🙂

joshmoore:
With #867 merged, this is now green. Merging (but #858/debian-failure) is still a mystery.

jakirkham:
Thanks Josh! 😀

Issue #757, opened 2021-05-23
behavior of `is_total_slice` for boundary chunks


d-v-b:
The `is_total_slice` function is used to determine whether a selection corresponds to an entire chunk (`is_total_slice` -> `True`) or not (`is_total_slice` -> `False`). 

I noticed that `is_total_slice` is always `False` for "partial chunks", i.e. chunks that are not filled by the array: 

```python
import zarr
from zarr.util import is_total_slice
# create an array with 1 full chunk and 1 partial chunk
a = zarr.open('test.zarr', path='test', shape=(10,), chunks=(9,), dtype='uint8', mode='w')
for x in BasicIndexer(slice(None), a):
    print(x.chunk_selection, is_total_slice(x.chunk_selection, a._chunks))
```
Which prints this:
```python
(slice(0, 9, 1),) True
(slice(0, 1, 1),) False
```
Although the last selection is not the size of a full chunk, it is "total" with respect to the output of that selection in the array.
A direct consequence of this behavior is unnecessary chunk loading when performing array assignments -- zarr uses the result of `is_total_slice` to decide whether to load existing chunk data or not. Because `is_total_slice` is always `False` for partial chunks, zarr _always_ loads boundary chunks during assignment.

A solution to this would be to augment the `is_total_slice` function to account for partial chunks. I'm not sure at the moment how to do this exactly, but it shouldn't be hard. Happy to bring forth a PR if people agree that this is an issue worth addressing.

jakirkham:
Yeah that seems reasonable

fAndreuzzi:
@d-v-b Thank you very much for the hints!

dcherian:
This also affects reads. For e.g. when dask chunks exactly overlap with Zarr chunks, it should be possible to decode directly to a memory buffer.

Instead because `is_total_slice` fails for boundary chunks, all boundary chunks are always uselessly copied one extra time.

It should be easy to tell `_process_chunk` the expected chunk shape for that particular chunk and pass that to `is_total_slice` instead of `self._chunks` (the loop that calls `_process_chunk` knows the chunk key)

An example profile in which 18 chunks take the fast path, and 9 chunks are uselessly copied in the last line.
```
Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
  2013                                               def _process_chunk(
  2014                                                   self,
  2015                                                   out,
  2016                                                   cdata,
  2017                                                   chunk_selection,
  2018                                                   drop_axes,
[...]

PR #863, opened 2021-10-29
[WIP] Boundary chunk optimization and typing

d-v-b:
This PR addresses #757. `is_total_slice` takes a slice-like argument and a shape and returns True if the slice-like argument "fills" the shape, False otherwise. When performing writes, the output of `is_total_slice` determines whether a chunk is completely re-written (fast) or read -> modified -> written (slower). In master, `is_total_slice` is always called with the array's chunk shape as the second argument, and thus this function always returns `False` for chunks at the boundary of an array. This PR adds a new function `zarr.util.trim_chunks` in that essentially computes `effective_chunk_shape = (chunk_shape * num_chunks) - array_shape` and passes this value to `is_total_slice`, which allows writing to boundary chunks without a read operation. 

Although all the core tests are passing, there's a deadlock in the `test_parallel_append` test in the `test_sync` suite. I have no idea what could be causing this.

Additionally, I kind of went crazy with adding type hints. It's possible that I will use this PR to add type hints to the entire library. It's possible that I should split the typing stuff into its own PR, but since none of it affects runtime I figured it couldn't hurt to combine it with some functional changes.

TODO:


joshmoore:
:+1: for the addition of type annotations from my side.

jakirkham:
Davis mentioned there is a process synchronization test failing. We discussed briefly how to debug this a bit further to see what is going on there

jhamman:

PR #563, opened 2020-05-09
Attempt to use os.walk to speedup reading from Directory Store

Carreau:
usage of os.walk for tree (os.scandir for folders) is faster than
listdir as it avoids many stats call.

This Should make the DirectoryStore faster.

See #562 

TODO:
* [x] Add unit tests and/or doctests in docstrings
* [x] Add docstrings and API docs for any new/modified user-facing classes and functions
* [ ] New/modified features documented in docs/tutorial.rst
* [ ] Changes documented in docs/release.rst
* [x] AppVeyor and Travis CI passes
* [x] Test coverage is 100% (Coveralls passes)

PR #889, opened 2021-11-26
Unguarded next inside generator

### LINKED ISSUE #881: LGTM.com warning: Unguarded next in generator

```python
zarr.storage. _keys_fast("")
```

#### Problem description

LGTM.com emits this warning for `zarr.storage. _keys_fast()`:
https://lgtm.com/projects/g/zarr-developers/zarr-python/snapshot/618aae80d21fc61f2b531e875e441d4a5b2c3042/files/zarr/storage.py?sort=name&dir=ASC&mode=heatmap#xb387346b6d81cfd5:1

In practice, I believe that  `os.walk(path)` will never be empty, because  `zarr.storage. _keys_fast()` is only called by `zarr.storage. keys()` on **existing paths**:
https://github.com/zarr-developers/zarr-python/blob/e14b3718d2b781cc87456ce5909795a2fefd75af/zarr/storage.py#L960-L962

In theory, it is not totally impossible that a path is deleted between the moment its existence is tested with `os.path.exists()` and the moment it is actually used - typically by exterior software and perhaps an attacker.

For that reason, it's probably best to explicitly take into account an empty `os.walk(path)` which would raise a `StopIteration` exception in `next(it)`:
https://github.com/zarr-developers/zarr-python/blob/e14b3718d2b781cc87456ce5909795a2fefd75af/zarr/storage.py#L972-L973
Perhaps change into:
```python
        it = iter(walker(path))
        try:
            d0, dirnames, filenames = next(it)
        except StopIteration:
            return  # alternatively explictly raise some other exception  
```

See [PEP 479](https://www.python.org/dev/peps/pep-0479/) for more information.

Please provide the following:

* Value of ``zarr.__version__`` : `master` branch

DimitriPapadopoulos:
Fixes #881.

Fix this DeepSource.io alert:
https://deepsource.io/gh/DimitriPapadopoulos/zarr-python/issue/PTC-W0063/occurrences

> Calls to `next()` should be inside `try-except` block.
> 
> When the iterator is exhausted, `StopIteration` exception is raised. When used inside a generator, this can cause unexpected behavior. If not handled, it will propagate out of the generator causing termination. [PEP-479](https://www.python.org/dev/peps/pep-0479/) has been accepted to fix this problem. It will modify the behavior of `StopIteration` in generators.
> 
> Each call to `next()` should be wrapped in a `try-except` block to explicitly handle `StopIteration` exceptions.

TODO:

### DIFF
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -966,12 +966,14 @@ class DirectoryStore(Store):
     @staticmethod
     def _keys_fast(path, walker=os.walk):
         for dirpath, _, filenames in walker(path):
-            dirpath = os.path.relpath(dirpath, path).replace("\\", "/")
-            for f in filenames:
-                if dirpath == os.curdir:
+            dirpath = os.path.relpath(dirpath, path)
+            if dirpath == os.curdir:
+                for f in filenames:
                     yield f
-                else:
-                    yield dirpath + "/" + f
+            else:
+                dirpath = dirpath.replace("\\", "/")
+                for f in filenames:
+                    yield "/".join((dirpath, f))
 
     def __iter__(self):
         return self.keys()
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -965,23 +965,13 @@ class DirectoryStore(Store):
 
     @staticmethod
     def _keys_fast(path, walker=os.walk):
-        """
-
-        Faster logic on platform where the separator is `/` and using
-        `os.walk()` to decrease the number of stats.call.
-
-        """
-        it = iter(walker(path))
-        d0, dirnames, filenames = next(it)
-        if d0.endswith('/'):
-            root_len = len(d0)
-        else:
-            root_len = len(d0)+1
-        for f in filenames:
-            yield f
-        for dirpath, _, filenames in it:
+        for dirpath, _, filenames in walker(path):
+            dirpath = os.path.relpath(dirpath, path).replace("\\", "/")
             for f in filenames:
-                yield dirpath[root_len:].replace('\\', '/')+'/'+f
+                if dirpath == ".":
+                    yield f
+                else:
+                    yield dirpath + "/" + f
 
     def __iter__(self):
         return self.keys()
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -968,7 +968,7 @@ class DirectoryStore(Store):
         for dirpath, _, filenames in walker(path):
             dirpath = os.path.relpath(dirpath, path).replace("\\", "/")
             for f in filenames:
-                if dirpath == ".":
+                if dirpath == os.curdir:
                     yield f
                 else:
                     yield dirpath + "/" + f


joshmoore:
Hmmm....will need to consider how to get the codecov back up.

DimitriPapadopoulos:
I could call `_keys_fast()` directly from a test.

An alternative would be an "attack" where you call `keys()` on an existing path and delete the path before `keys()` calls `_keys_fast()`. Good luck with that :smiley:

jakirkham:
Should we try to just integrate this into a single `for`-loop?

DimitriPapadopoulos:
As far as I can see, there is no loop. Just a single call to `next(it)`:
https://github.com/zarr-developers/zarr-python/blob/e6483f926f322496660847988f27e326bd54cea1/zarr/storage.py#L974-L975

DimitriPapadopoulos:
Indeed, I don't get why the root directory is extracted from the first iteration of `os.walk()` instead of `path`.
* It seems plain wrong, as the walker is not guaranteed to return the root as the first iterable - think `os.walk(..., topdown=True, ...)`.
* It is cumbersome.

But then I thought the current code might be trying to work around an undocumented issue.

* cumbersome,

jakirkham:
Good point. Looking at `git blame` it seems all of this code goes back to PR ( https://github.com/zarr-developers/zarr-python/pull/563 ). Any underlying issues don't seem to be mentioned there. Given this, it seems reasonable to try cleaning up the code a bit more (since that is already what we are doing)

We recently did an alpha release and may do a couple more depending. So if it does turn out that [there were dragons here]( https://en.wikipedia.org/wiki/Here_be_dragons ), we should have time to come back and fix them

DimitriPapadopoulos:
I fear the corner case is the case of platforms where the path spearator is not `/` - but I don't have one to test.

Also is the current code really fast than a call to `os.path.relpath()`?

I was thinking of going back to something like this, (discarding the `_fast_keys()` fiunction:
```python
    def keys(self):
        for dirpath, _, filenames in os.walk(self.path):
            dirpath = os.path.relpath(dirpath, self.path)
            if os.sep != "/":
                dirpath = dirpath.replace(os.sep, "/")
            for f in filenames:
                yield dirpath + "/" + f
```
Are there systems where the

jakirkham:
What would be a good example of that fail case? Windows? If so, that is covered in CI here. Though if you can think of additional tests, those would also be welcome

Yeah dropping `_keys_fast` is fine. AIUI that just made it simpler to write the iterator

DimitriPapadopoulos:
I don't know whether `os.sep` is `":"` or `"/"` on macOS.

jakirkham:
```bash
$ sw_vers
ProductName:	Mac OS X
ProductVersion:	10.15.7
BuildVersion:	19H1419
$ python -c "import os; print(os.sep)"
/
```

DimitriPapadopoulos:
Ah, the corner case might be when `self.path = "."`.

DimitriPapadopoulos:
I am not removing the `_keys_fast()` for now. Tests directly call `_keys_fast()`, and I don't want to change both the code and the tests at the same time.

zarr/storage.py
```diff
@@ -965,23 +965,13 @@ def keys(self):
...
-        else:
-            root_len = len(d0)+1
-        for f in filenames:
-            yield f
-        for dirpath, _, filenames in it:
+        for dirpath, _, filenames in walker(path):
+            dirpath = os.path.relpath(dirpath, path).replace("\\", "/")
             for f in filenames:
-                yield dirpath[root_len:].replace('\\', '/')+'/'+f
+                if dirpath == ".":
```
jakirkham, line 971:
```suggestion
                if dirpath == os.curdir:
```

jakirkham:
Makes sense

Was looking to see if `pathlib` would help us with the Windows path issues, but maybe that is too much for our needs

DimitriPapadopoulos:
Thinking about it again, the corner case is when the `dirpath` returned by the first `os.walk()` iteration is different from `path`. Could that be the case with paths that are not normalized, such as `/my/././path`, or maybe with symlinks? Hard to tell without specific tests that would convey the original author's intent.

jakirkham:
Could we just call `os.path.abspath` on both and compare?

DimitriPapadopoulos:
We could, but it could differ from the initial implementation, plus it's probably not needed after all:
* iterating over `os.walk("/././tmp")` yields a `dirpath` that starts with "/././tmp",
* `os.path.relpath()` appears to be normalizing `dirpath` well enough:
  ```python
  >>> os.path.relpath("/././tmp", "/././tmp")
  '.'
  >>> os.path.relpath("/././tmp", "/tmp")
  '.'
  >>> os.path.relpath("/tmp", "/././tmp")
  '.'
  ```

jakirkham:
From the Python [docs for `relpath`]( https://docs.python.org/3/library/os.path.html#os.path.relpath ):

> This is a path computation: the filesystem is not accessed to confirm the existence or nature of path or start.

So symlinks shouldn’t do anything special

jakirkham:
Anyways this seems like a good approach to me. Any other thoughts on your end?

DimitriPapadopoulos:
I was rather afraid of symlinks with `os.walk()`, not `os.path.relpath()`. However, I cannot identify a corner case where it wouldn't work as before: `os.walk()` appears to always use  the `path` passed as an argument as a prefix, without ever trying to normalize or being "clever" in any way. Even paths with a trailing `/` are propagated to the first `dirpath`:
```python
>>> next(iter(os.walk("/tmp/foo")))
('/tmp/foo', ['bar'], [])
>>> next(iter(os.walk("/tmp/foo/")))
('/tmp/foo/', ['bar'], [])
>>>
```
Therefore, `path` and the `dirpath` returned by the first iteration of `os.walk()` cannot be but identical.

This tells me we need to remove any trailing `/` from `path`. What do you think?

jakirkham (approved):

Issue #600, opened 2020-09-09
Have a BaseStore class ?


Carreau:
Thi is a suggestion to have a `BaseStore` class or ABC, that store implementations can inherit from to have basic functionality.

#### Problem description

Saying that stores are "just" `MutableMapping` may be insufficient, a number of places call methods on stores that are not available in `MutableMapping`, which:

  - Make static type Checking fails if one uses `MutableMapping` as a type annotation.
  - Would provide a default methods, like `close()` method that is no-op, and would avoid the need to check for method existence.
  - Avoid reimplementation of boilerplate :`def __len__(self): return len(self.keys())`, still allowing overwrite.

jakirkham:
Also issue ( https://github.com/zarr-developers/zarr-python/issues/573 ) and issue ( https://github.com/zarr-developers/zarr-python/issues/349 ) are related.

jakirkham:
One thing that might be nice is to have a store that wraps a minimal `MutableMapping` with the extra bits of this ABC. That way users still have a path for using a minimal `MutableMapping` in this new framework.

joshmoore:
see https://github.com/manzt/napari-dzi-zarr/blob/master/napari_dzi_zarr/store.py from @manzt for a possible consumer of a base class.

Carreau:
there seem to be a number of place where the test assumes that stores can do things quite differently than what I was expecting and where `dict`, pass the test, but `MemoryStore` does not.  

Typically things like

```

def test_getsize():
    ...

    store = # Put dict() or MemoryStore() here
    store['boo'] = None
    assert -1 == getsize(store)
```

Will pass for dict, but `MemoryStore` will refuse to store None. 


or 


```
from zarr.storage import MemoryStore, init_array
from zarr import Array

store = # Put dict() or MemoryStore() here
init_array(store, path="foo/bar", shape=100, chunks=10)
z =  Array(
    store,
    path="foo/bar"
)

z[:] = 42
z.store['foo'] = b'bar'
z[:] = 43
```


Where MemoryStore will be quite unhappy because `z.store['foo'] = b'bar'` will mess-up the hierarchy, but `dict` won't call "requires parent" so works fine. 

So i'm unsure which is the "correct" behavior.

jakirkham:
Done in PR ( https://github.com/zarr-developers/zarr-python/pull/789 ) and preceding work

PR #911, opened 2021-12-14
Create fsstore from filesystem

### LINKED ISSUE #993: Cannot create FSStore for unlistable http URLs

```python
import zarr
import fsspec

# public
url = 'https://g-c70289.0da32.08cc.data.globus.org/pyqg_ensemble/001.zarr'

# this errors
store = zarr.storage.FSStore(url)

# because of this
fs = fsspec.implementations.http.HTTPFileSystem()
fs.isdir(url) # -> False

# however, there is consolidated metadata
data = fs.get_mapper(url)['.zmetadata']
```

#### Problem description

We often store Zarr on HTTP servers that don't support directory listing. This works because Zarr consolidated metadata tells us everything we need to know about the data available, so no listing is required.

However, Zarr won't let me create such a Store, as shown in the example above.

There needs to be a workaround to allow unlistable http stores with consolidated metadata. 

Please provide the following:

* zarr 2.11.0
* fsspec 2022.02.0

This came up in https://github.com/pangeo-forge/pangeo-forge-recipes/issues/222

rabernat:
This PR adds a new argument to `FSStore` which allows one to pass an existing fsspec filesystem object from which to generate the FSStore. The need to initialize an FSStore from an existing filesystem came up in https://github.com/pangeo-forge/pangeo-forge-recipes/issues/254. It seems like this is something people may want to do more generally.

Also fixes #993 by simply removing the check for list-ability when initializing an FSStore.

TODO:

### DIFF
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -22,7 +22,7 @@ Enhancements
   By :user:`Ryan Abernathey <rabernat>`.
 
 
-.. _release_2.11.1:
+.. _release_2.11.3:
 
 2.11.3
 ------
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -9,6 +9,27 @@ Unreleased
 Bug fixes
 ~~~~~~~~~
 
+* Fix bug that made it impossible to create an ``FSStore`` on unlistable filesystems
+  (e.g. some HTTP servers).
+  By :user:`Ryan Abernathey <rabernat>`; :issue:`993`.
+
+Enhancements
+~~~~~~~~~~~~
+
+* **Create FSStore from an existing fsspec filesystem**. If you have created
+  an fsspec filesystem outside of Zarr, you can now pass it as a keyword
+  argument to ``FSStore``.
+  By :user:`Ryan Abernathey <rabernat>`.
+
+
+.. _release_2.11.1:
+
+2.11.3
+------
+
+Bug fixes
+~~~~~~~~~
+
 * Changes the default value of ``write_empty_chunks`` to ``True`` to prevent
   unanticipated data losses when the data types do not have a proper default
   value when empty chunks are read back in.
@@ -319,7 +340,7 @@ Bug fixes
 
 * FSStore: default to normalize_keys=False
   By :user:`Josh Moore <joshmoore>`; :issue:`755`.
-* ABSStore: compatibility with ``azure.storage.python>=12`` 
+* ABSStore: compatibility with ``azure.storage.python>=12``
   By :user:`Tom Augspurger <tomaugspurger>`; :issue:`618`
 
 
@@ -484,7 +505,7 @@ This release will be the last to support Python 3.5, next version of Zarr will b
 
 * `DirectoryStore` now uses `os.scandir`, which should make listing large store
   faster, :issue:`563`
-  
+
 * Remove a few remaining Python 2-isms.
   By :user:`Poruri Sai Rahul <rahulporuri>`; :issue:`393`.
 
@@ -504,7 +525,7 @@ This release will be the last to support Python 3.5, next version of Zarr will b
   ``zarr.errors``  have been replaced by ``ValueError`` subclasses. The corresponding
   ``err_*`` function have been removed. :issue:`590`, :issue:`614`)
 
-* Improve consistency of terminology regarding arrays and datasets in the 
+* Improve consistency of terminology regarding arrays and datasets in the
   documentation.
   By :user:`Josh Moore <joshmoore>`; :issue:`571`.
 
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -1298,6 +1298,15 @@ class FSStore(Store):
         Separator placed between the dimensions of a chunk.
     fs : fsspec.spec.AbstractFileSystem, optional
         An existing filesystem to use for the store.
+    check : bool, optional
+        If True, performs a touch at the root location, to check for write access.
+        Passed to `fsspec.mapping.FSMap` constructor.
+    create : bool, optional
+        If True, performs a mkdir at the rool location.
+        Passed to `fsspec.mapping.FSMap` constructor.
+    missing_exceptions : sequence of Exceptions, optional
+        Exceptions classes to associate with missing files.
+        Passed to `fsspec.mapping.FSMap` constructor.
     storage_options : passed to the fsspec implementation. Cannot be used
         together with fs.
     """
@@ -1310,6 +1319,9 @@ class FSStore(Store):
                  exceptions=(KeyError, PermissionError, IOError),
                  dimension_separator=None,
                  fs=None,
+                 check=False,
+                 create=False,
+                 missing_exceptions=None,
                  **storage_options):
         import fsspec
 
@@ -1318,7 +1330,13 @@ class FSStore(Store):
             # set auto_mkdir to True for local file system
             if protocol in (None, "file") and not storage_options.get("auto_mkdir"):
                 storage_options["auto_mkdir"] = True
-            self.map = fsspec.get_mapper(url, **storage_options)
+            self.map = fsspec.get_mapper(
+                url,
+                check=check,
+                create=create,
+                missing_exceptions=missing_exceptions,
+                **storage_options
+            )
             self.fs = self.map.fs  # for direct operations
             self.path = self.fs._strip_protocol(url)
         else:
@@ -1326,7 +1344,12 @@ class FSStore(Store):
                 raise ValueError("Cannot specify both fs and storage_options")
             self.fs = fs
             self.path = self.fs._strip_protocol(url)
-            self.map = self.fs.get_mapper(self.path)
+            self.map = self.fs.get_mapper(
+                self.path,
+                check=check,
+                create=create,
+                missing_exceptions=missing_exceptions
+            )
 
         self.normalize_keys = normalize_keys
         self.mode = mode
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -1326,9 +1326,9 @@ class FSStore(Store):
         import fsspec
 
         mapper_options = {"check": check, "create": create}
-        # Workaround for https://github.com/zarr-developers/zarr-python/pull/911#discussion_r841926292
-        # Some fsspec implementations don't accept missing_exceptions. This will
-        # avoid passing it in the most common scenarios.
+        # https://github.com/zarr-developers/zarr-python/pull/911#discussion_r841926292
+        # Some fsspec implementations don't accept missing_exceptions.
+        # This is a workaround to avoid passing it in the most common scenarios.
         if missing_exceptions is not None:
             mapper_options["missing_exceptions"] = missing_exceptions
 
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -1326,7 +1326,7 @@ class FSStore(Store):
                 raise ValueError("Cannot specify both fs and storage_options")
             self.fs = fs
             self.path = self.fs._strip_protocol(url)
-            self.map = self.fs.get_mapper(path)
+            self.map = self.fs.get_mapper(self.path)
 
         self.normalize_keys = normalize_keys
         self.mode = mode
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -1329,8 +1329,9 @@ class FSStore(Store):
         # https://github.com/zarr-developers/zarr-python/pull/911#discussion_r841926292
         # Some fsspec implementations don't accept missing_exceptions.
         # This is a workaround to avoid passing it in the most common scenarios.
+        # Remove this and add missing_exceptions to mapper_options when fsspec is released.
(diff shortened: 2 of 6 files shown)


zarr/storage.py
```diff
@@ -1091,8 +1091,9 @@ class FSStore(Store):
     Parameters
     ----------
     url : str
-        The destination to map. Should include protocol and path,
-        like "s3://bucket/root"
+        The destination to map. If no fs is provided, should include protocol
```
martindurant, line 1094: Perhaps an even simpler alternative is if the input is already an FSMap? We could handle both those cases. 

Is the intent that this will be called from zarr.open* or xarray, or that calling code will instantiate this themselves before calling zarr.open(fsstore) ?

zarr/storage.py
```diff
@@ -1104,7 +1105,10 @@ class FSStore(Store):
         as a missing key
     dimension_separator : {'.', '/'}, optional
         Separator placed between the dimensions of a chunk.
-    storage_options : passed to the fsspec implementation
+    fs : fsspec.spec.AbstractFileSystem, optional
+        An existing filesystem to use for the store.
+    storage_options : passed to the fsspec implementation. Cannot be used
```
martindurant, line 1110: There are a couple of extra options that FSMap takes over what a FileSystem takes (check, create, exceptions) that could go here.

zarr/tests/test_storage.py
```diff
@@ -1134,6 +1134,35 @@ def create_store(self, normalize_keys=False, key_separator=".", **kwargs):
             key_separator=key_separator)
 
 
+@pytest.mark.skipif(have_fsspec is False, reason="needs fsspec")
+class TestFSStoreFromFilesystem(StoreTests):
```
martindurant, line 1138: I suppose we should test these in combination with top level zarr objects.

zarr/storage.py
rabernat: Ok so it seems like there are three basic cases we need to cover:
1. The user wants to specify a URL (including protocol)
2. The user already has a filesystem initialized and wants to use that to create the FSStore
3. The user already has an FSMap object and want to use that to create the FSStore

I am already uncomfortable with the overloading of the constructor on this class to handle two different scenarios. If we expand it to cover 3, it will get even more messy. A better pattern would be

```python
class FSStore(Store):

    @classmethod
    def from_url(cls, url):
        # figure out fs and path
        return cls(path, fs)

    @classmethod
    def from_fsmap(cls, fsmap):
        # figure out fs and path
        return cls(path, fs)

    def __init__(self, path, fs):
        ...
```
However, this would change the API and break lots of code.

More generally, looking at this code has made me realize that the interface between zarr.storage and FSSpec is not very clean. `FSStore` reimplements and extends many of the MutableMapping methods that `FSMap` implements. `FSStore` calls both `self.map` methods and `self.fs` methods (some of them private 😬 ).

zarr/storage.py
rabernat: Are these the only possible options - https://github.com/fsspec/filesystem_spec/blob/201555581a9f0c0de810efc372704db952397330/fsspec/mapping.py#L35 - `check=False, create=False, missing_exceptions=None`?

zarr/tests/test_storage.py
rabernat: Where do those tests belong?

zarr/storage.py
martindurant: I agree with all of that. As you have seen, this is a consequence of the evolution of the storage class in zarr itself - originally it was to be a MutableMap alone, so that could go into fsspec and be generally useful, but then stores started to gain additional capabilities. And here we are.

I am *not* uncomfortable with the class methods. I think we can have `__init__` stay the same (and not break code), but add `from_filesystem(fs, path, args)`, `from_mapper(mapper)` class methods. Either the calling code will determine which to call, or we add another helper to decide?

zarr/tests/test_storage.py
martindurant: I would say right here. I haven't looked to see what other tests do for full integration/roundtrips, but fsspec is certainly a bit special.

rabernat:
I would like to revive and move forward this PR. Martin, you gave a few helpful comments and ideas above. Would you mind turning these into a more formal review in terms of what you think is _needed_ vs _interesting possibility_.

What we have here is a tightly scoped and tested PR to unblock a rather serious upstream problem. If there is no major technical debt incurred by this PR, could we consider merging as is (modulo doc updates from checklist) rather than expanding scope? If not, please let us know what else would be required.

Also pinging @grlee77 for potential interaction with ongoing refactor of storage classes (#600).

zarr/storage.py
grlee77: Yeah, if `__init__` stayed the same then the rest of this seems pretty non-disruptive to me. I do think a classmethod approach looks nice.

zarr/tests/test_storage.py
grlee77: There are tests with higher-level zarr `Array` objects in `test_core.py` (e.g. `TestArrayWithFSStore`) and for `Group` in `test_hierarchy.py` (see `TestGroupWithFSStore`). Currently those just create an FSStore using a path string as returned by `tempfile.mkdtemp()`

grlee77:
> Also pinging @grlee77 and @joshmoore for potential interaction with ongoing refactor of storage classes (#789).

Thanks @rabernat, I commented above. If the `__init__` signature does not change then this should be pretty easy to merge into the work there.

zarr/storage.py
rabernat: Thanks a lot for the feedback Greg! Just to be totally explicit, is adding an additional argument  to `__init__` ( as I have have done here with `fs`), while keeping the other arguments backwards compatible, considered a "disruptive change"? Or is this PR already sufficiently non-disruptive.

grlee77: Yeah, what is currently in the PR seems fine to me. It should be easy to just merge this into the v3 branch with minimal conflicts or extra work.

joshmoore:
Thanks, @grlee77. From my side, that was the only concern in terms of impact on existing work, @rabernat. @martindurant's point about the general API "burden" seems to be outstanding. Is it that a concrete alternative that doesn't break the API needs proposing from your side, @rabernat?

A side note that testing is currently blocked due to GHA weirdness. @jakirkham and I will try to get that sorted ASAP.

zarr/storage.py
```diff
@@ -1328,8 +1363,6 @@ def __init__(self, url, normalize_keys=False, key_separator=None,
 
         # Pass attributes to array creation
         self._dimension_separator = dimension_separator
-        if self.fs.exists(self.path) and not self.fs.isdir(self.path):
-            raise FSPathExistNotDir(url)
```
rabernat, line 1332: To fix #993 I have simply removed this check...

zarr/tests/test_storage.py
```diff
@@ -1248,10 +1248,6 @@ def test_read_only(self):
 
         assert store[self.root + 'foo'] == b"bar"
 
-        filepath = os.path.join(path, self.root + "foo")
-        with pytest.raises(ValueError):
-            self.create_store(path=filepath, mode='r')
```
rabernat, line 1253: ...and its corresponding test.

This test did not do anything other than test the error above. `self.create_store` does not actually attempt to write any data or list the store.

With the check gone, users will now see an error from fsspec if they try to do an operation that requires listing on an unlistable FSSpec store, as they would already on a generic MutableMapping-based store.

zarr/storage.py
rabernat: As the current [test failure](https://github.com/zarr-developers/zarr-python/runs/5820026384?check_suite_focus=true#step:7:1634) shows, it is not supported by `LocalFileSystem`.

zarr/storage.py
rabernat: Now that https://github.com/fsspec/filesystem_spec/pull/939 has gone in, it should be possible to create an FSStore from an FSMap. But let's address that in a future PR, since that functionality is still unreleased.

docs/release.rst
```diff
@@ -9,6 +9,27 @@ Unreleased
...
+* **Create FSStore from an existing fsspec filesystem**. If you have created
+  an fsspec filesystem outside of Zarr, you can now pass it as a keyword
+  argument to ``FSStore``.
+  By :user:`Ryan Abernathey <rabernat>`.
+
+
+.. _release_2.11.3:
+
+2.11.3
+------
```
rabernat, line 28: Although 2.11.3 has been released, I noted that the docs had not been updated, so I added it here.

It seems we have skipped release 2.11.2?

joshmoore: Those releases are coming from the `2_11` branch. (See the related https://github.com/zarr-developers/zarr-python/pull/898#issuecomment-1076912502)

I assume you intend this for release as a 2.11.x?

martindurant: Please ping this thread with any decision on which branch to merge into

joshmoore: @martindurant : @rabernat seemed to think it was ok to hold off on @grlee77's upcoming v3 configuration fix, i.e. this can stay on the branch it is and be released as something `>2.11.x`. If it looks like we want it in a quick `~=2.11.x` release, I can attempt the backport.

rabernat: I will do whatever the other devs recommend here (either backport to 2.11.x or put this in 2.12.0)

I am confused about how we will maintain a consistent changelog for these two branches. So I am standing by for advice.

rabernat:
I feel confused about how to move forward with this PR in terms of how to update the changelog (see https://github.com/zarr-developers/zarr-python/pull/911#discussion_r848818092). Please help me 🙏. I would love to see this merged soon as I believe it is finished. It is not starting to accumulate conflicts.

joshmoore:
Sorry for having left you hanging. I think either will be fine @rabernat. Either just add your item to unreleased or copy over from the 2 11 branch. Either way feel free to get this in and I will double check before the next release. (Of whatever branch)

joshmoore (approved):
I've hopefully handled the conflicts with Greg's v3 work. Once this goes green, I'll merge and update the release notes to include previous versions.

joshmoore:
It looks like the semantics of the `update()` method have changed with the refactoring for v3:

```
pytest zarr/tests/test_convenience.py -svx -k2

...snip...

>           store_to_open.update(store)  # copy original store to new unlistable store

zarr/tests/test_convenience.py:305:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <zarr.storage.FSStore object at 0x1ba010970>, other = '/var/folders/z5/txc_jj6x5l5cm81r56ck1n9c0000gn/T/tmp0gy4ji7i', kwds = {}

    def update(self, other=(), /, **kwds):
        ''' D.update([E, ]**F) -> None.  Update D from mapping/iterable E and F.
            If E present and has a .keys() method, does:     for k in E: D[k] = E[k]
            If E present and lacks .keys() method, does:     for (k, v) in E: D[k] = v
            In either case, this is followed by: for k, v in F.items(): D[k] = v
        '''
        if isinstance(other, Mapping):
            for key in other:
                self[key] = other[key]
        elif hasattr(other, "keys"):
            for key in other.keys():
                self[key] = other[key]
        else:
>           for key, value in other:
E           ValueError: not enough values to unpack (expected 2, got 1)
```


That doesn't appear to be related to the type hierarchy:
```
2.11.2: FSStore.__mro__: (<class 'zarr.storage.FSStore'>, <class 'zarr._storage.store.Store'>, <class 'zarr._storage.store.BaseStore'>, <class 'collections.abc.MutableMapping'>, <class 'collections.abc.Mapping'>, <class 'collections.abc.Collection'>, <class 'collections.abc.Sized'>, <class 'collections.abc.Iterable'>, <class 'collections.abc.Container'>, <class 'object'>)
2.12.x: FSStore.__mro__: (<class 'zarr.storage.FSStore'>, <class 'zarr._storage.store.Store'>, <class 'zarr._storage.store.BaseStore'>, <class 'collections.abc.MutableMapping'>, <class 'collections.abc.Mapping'>, <class 'collections.abc.Collection'>, <class 'collections.abc.Sized'>, <class 'collections.abc.Iterable'>, <class 'collections.abc.Container'>, <class 'object'>)
```
Still digging in case anyone has an idea (cc: @grlee77)

rabernat:
Thanks for finishing this up Josh! I really appreciate it!

Issue #809, opened 2021-08-18
Read Only array upon default opening


camFoltz:
For bug reports, please follow the template below. For enhancement proposals, feel free
to use whatever template makes sense (major new features should be discussed in the
Zarr specifications repository https://github.com/zarr-developers/zarr-specs).

#### Minimal, reproducible code sample, a copy-pastable example if possible

```python
src = 'path/to/dir'
z = zarr.open(src)
z.create('array',shape=(1, 1, 1, 1, 2048, 2048),
        chunks=(1, 1, 1, 1, 2048, 2048),
        compressor=Blosc('zstd', clevel=3, shuffle=Blosc.BITSHUFFLE),
        read_only=True)
z['array'][0,4] = 1

## Reload the array after restarting kernel
z2 = zarr.open(src) ## no longer read_only
```

#### Problem description

I am trying to create a zarr store containing an array that is by default read_only.  We are moving our raw microscopy data to zarr format and we do not want the user to be able to easily edit the zarr arrays.  The default `zarr.open(mode='a')` overrides any sort of read_only setting put onto the array during creation.  This is troublesome because some of our users will not remember to specify `mode='r'` when browsing the raw data.

Is there a way to have the zarr store/array be read_only by default upon opening?

#### Version and installation information

Please provide the following:

* Value of ``zarr.__version__`` = 2.8.3
* Value of ``numcodecs.__version__`` = 0.7.3
* python 3.7
* Mac
* pip install zarr

joshmoore:
Hi @camFoltz. I don't know of a zarr-specific mechanism for recording the read-only setting. As a workaround, have you looked at making it read-only at the underlying storage level?

camFoltz:
Hi @joshmoore , thanks for the response.  If zarr has no specific mechanism then we will move to dealing with it at the permission level.

clbarnes:
While there isn't (and probably shouldn't be, for simplicity purposes) a mechanism for storing the default access mode on individual containers/ groups/ arrays, would changing the global default mode to `"r"` be considered in the next major version? h5py did that (somewhat) recently; it would more closely match python's built-in `open`, and in a post-v3 world where different stores may be readable, writable, and listable, the read-only variant is likely to be the most fundamental and therefore be the simplest assumption to make.

joshmoore:
@clbarnes :  *nods* and I seem to remember a store (`FSStore`) recently switching its local default. I'll try to find that commit/PR.

[...]

Issue #712, opened 2021-03-25
Bug: zarr.open behaves different than zarr.open_group with mode `w-` on `gs://` URIs


adrianloy:
The convenient function `zarr.open` does not work to create a zarr group on a gcp bucket with mode `w-`:

#### Minimal, reproducible code sample:

```python
an_empty_gs_uri = "gs://my-bucket/test.zarr"
zarr.open(an_empty_gs_uri, mode=w-)

```

#### Problem description

If we use w- mode above code breaks, even if we dont have a group at the URI given:

```python
---------------------------------------------------------------------------
ReadOnlyError                             Traceback (most recent call last)
<ipython-input-11-d0b46f9c7e3f> in <module>
----> 1 zarr.open(t3, mode='w-')

~/merantix/mxlabs-chameleon/venv/lib/python3.8/site-packages/zarr/convenience.py in open(store, mode, **kwargs)
     84             return open_array(store, mode=mode, **kwargs)
     85         else:
---> 86             return open_group(store, mode=mode, **kwargs)
     87 
     88     elif mode == "a":

~/merantix/mxlabs-chameleon/venv/lib/python3.8/site-packages/zarr/hierarchy.py in open_group(store, mode, cache_attrs, synchronizer, path, chunk_store, storage_options)
   1181             raise ContainsGroupError(path)
   1182         else:
-> 1183             init_group(store, path=path, chunk_store=chunk_store)
   1184 
   1185     # determine read only status

~/merantix/mxlabs-chameleon/venv/lib/python3.8/site-packages/zarr/storage.py in init_group(store, overwrite, path, chunk_store)
    470 
    471     # initialise metadata
--> 472     _init_group_metadata(store=store, overwrite=overwrite, path=path,
    473                          chunk_store=chunk_store)
    474 

~/merantix/mxlabs-chameleon/venv/lib/python3.8/site-packages/zarr/storage.py in _init_group_metadata(store, overwrite, path, chunk_store)
    497     meta = dict()  # type: ignore
    498     key = _path_to_prefix(path) + group_meta_key
--> 499     store[key] = encode_group_metadata(meta)
    500 
    501 

~/merantix/mxlabs-chameleon/venv/lib/python3.8/site-packages/zarr/storage.py in __setitem__(self, key, value)
   1058     def __setitem__(self, key, value):
   1059         if self.mode == 'r':
-> 1060             raise ReadOnlyError()
   1061         key = self._normalize_key(key)
[...]

PR #916, opened 2021-12-20
fix consolidate_metadata with FSStore

### LINKED ISSUE #915: Using consolidate_metadata with FSStore result in a read-only store

```python
import xarray as xr
from zarr.convenience import consolidate_metadata


ds = xr.tutorial.open_dataset('air_temperature')
ds.to_zarr('file://foo.zarr', mode='w', consolidated=False)
consolidate_metadata('file://foo.zarr')
```

raises:

```python
ReadOnlyError                             Traceback (most recent call last)
/tmp/ipykernel_63/1636373769.py in <module>
----> 1 consolidate_metadata('file://foo.zarr')

/srv/conda/envs/notebook/lib/python3.9/site-packages/zarr/convenience.py in consolidate_metadata(store, metadata_key)
   1126         }
   1127     }
-> 1128     store[metadata_key] = json_dumps(out)
   1129     return open_consolidated(store, metadata_key=metadata_key)
   1130 

/srv/conda/envs/notebook/lib/python3.9/site-packages/zarr/storage.py in __setitem__(self, key, value)
   1134     def __setitem__(self, key, value):
   1135         if self.mode == 'r':
-> 1136             raise ReadOnlyError()
   1137         key = self._normalize_key(key)
   1138         path = self.dir_path(key)

ReadOnlyError: object is read-only
```

#### Problem description

When using a fsspec compatible string as the store argument, zarr's `consolidate_metadata` function returns a read-only store, resulting in the error shown above. I believe the expected behavior is to have generate an `FSStore` object with write privileges (so the consolidated meatadata key can be written). 

Please provide the following:

* Value of ``zarr.__version__``: 2.10.3
* Value of ``numcodecs.__version__``: 0.9.1
* Version of Python interpreter: 3.9
* Operating system (Linux/Windows/Mac): Linux and OSX
* How Zarr was installed: conda

jhamman:
This PR fixes a small bug in the `consolidate_metadata` convenience function where, when using fsspec compatible store paths, the `FSStore` was opened as read-only, causing the write of the consolidated metadata key to fail.

Closes #915  

cc @martindurant, @freeman-lab

TODO:

### DIFF
--- a/zarr/convenience.py
+++ b/zarr/convenience.py
@@ -1117,7 +1117,7 @@ def consolidate_metadata(store: StoreLike, metadata_key=".zmetadata"):
     open_consolidated
 
     """
-    store = normalize_store_arg(store)
+    store = normalize_store_arg(store, clobber=True)
 
     def is_zarr_key(key):
         return (key.endswith('.zarray') or key.endswith('.zgroup') or
--- a/zarr/tests/test_storage.py
+++ b/zarr/tests/test_storage.py
@@ -19,6 +19,7 @@ from numcodecs.compat import ensure_bytes
 
 import zarr
 from zarr.codecs import BZ2, AsType, Blosc, Zlib
+from zarr.convenience import consolidate_metadata
 from zarr.errors import MetadataError
 from zarr.hierarchy import group
 from zarr.meta import ZARR_FORMAT, decode_array_metadata
@@ -1008,6 +1009,10 @@ class TestFSStore(StoreTests):
         a[:4] = [0, 1, 2, 3]
         assert "data" in os.listdir(path1)
         assert ".zgroup" in os.listdir(path1)
+
+        # consolidated metadata (GH#915)
+        consolidate_metadata("file://" + path1)
+        assert ".zmetadata" in os.listdir(path1)
 
         g = zarr.open_group("simplecache::file://" + path1, mode='r',
                             storage_options={"cache_storage": path2,


martindurant:
+1 for the idea

I wonder if we should check that `store` is a string before doing this - might we otherwise end up writing to a store that has been opened explicitly as readonly? I'm not sure if that's a side-effect, maybe this comment can be ignored.

jhamman:
@martindurant - 

> I wonder if we should check that store is a string before doing this - might we otherwise end up writing to a store that has been opened explicitly as readonly? I'm not sure if that's a side-effect, maybe this comment can be ignored.

I think this is already take care of in `normalize_store_arg`:

https://github.com/zarr-developers/zarr-python/blob/4fb0f7e7f452fd436062d39ce67bef6ccd336420/zarr/storage.py#L112-L115

joshmoore:
Thanks, @jhamman! This could fix or at least play a role in fixing:
 - https://github.com/zarr-developers/zarr-python/issues/712
 - https://github.com/zarr-developers/zarr-python/issues/809

jhamman:
> Thanks, @jhamman! This could fix or at least play a role in fixing:
> 
> * [Bug: zarr.open behaves different than zarr.open_group with mode `w-` on `gs://` URIs #712](https://github.com/zarr-developers/zarr-python/issues/712)
> * [Read only array default upon opening #809](https://github.com/zarr-developers/zarr-python/issues/809)

It's quite possible these are related, but I think the relation is, at most, limited to how `FSStore`  is instantiated in various places. The issue (and changes here) are confined to `consolidate_metadata` and the `clobber` argument to `normalize_store_arg`. It may be worth taking a deeper look at the call path for `normalize_store_arg` to see if there is some relation.

Issue #917, opened 2021-12-21
DOC: make Creating a development environment doc copy-pastable


raybellwaves:
In the docs (https://zarr.readthedocs.io/en/stable/contributing.html#creating-a-development-environment).
Tried a quick copy paste to set up a dev environment but got 
```
virtualenv --no-site-packages --python=/usr/bin/python3.8 ~/pyenv/zarr-dev
FileNotFoundError: [Errno 2] No such file or directory: '/usr/bin/python3.8'
```

I see 

❯ python --version
Python 3.9.1

Azaya89:
Hi @joshmoore , can I work on this issue?

MSanKeys963:
Hi @Azaya89.
Yeah, sure. Please go ahead.

Azaya89:
Hi @MSanKeys963, when i do` git fetch upstream` it tells me permission denied. How do i get around that?

joshmoore:
What does the following command show?

```
git remote -v
```

Azaya89:
It shows both my local cloned directory and the upstream. Anyway, I have solved the issue. I had to remove and re-add the upstream https and got it to work now.

blackpandan:
Is this issue still open?

joshmoore:
@blackpandan: #1189 is starting on this. Perhaps you want to review it?

d-v-b:
we have hatch and a justfile so the dev environment is much more self-supporting. closing as completed.

PR #918, opened 2021-12-21
DOC: drop 3.8 reference

raybellwaves:
Closes https://github.com/zarr-developers/zarr-python/issues/917

I don't have 3.8 installed (I have 3.9 installed). Dropping the 3.8 reference makes it copy-pastable. I would hope users have a recent version of python installed to work work (i.e. not 2.7 :grimacing:). I Could have chosen to make the version 3.9 but it may require a bump to 3.10 in the future.

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -90,7 +90,7 @@ you have cloned the Zarr source code and your current working directory is the r
 the repository, you can do something like the following::
 
     $ mkdir -p ~/pyenv/zarr-dev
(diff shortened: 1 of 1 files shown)


joshmoore:
Thanks @raybellwaves. I'm unsure of the original motivation for hard-coding `3.8` but I think it's safe to drop. I could also see going a step further and replacing:
```
virtualenv --no-site-packages --python=/usr/bin/python
```
with
```
python -m venv
```

raybellwaves:
closing in favor of https://github.com/zarr-developers/zarr-python/pull/921

PR #920, opened 2021-12-21
Drop shortcut `fsspec[s3]` for dependency

joshmoore:
Use of the short rather than `fsspec` and `s3fs` separately leads
to an old version of s3fs being installed. (The benefit of the
shortcut is that it prevents dependabot PRs from failing.)

see:
* https://github.com/fsspec/s3fs/issues/528
* https://github.com/zarr-developers/zarr-python/pull/914


joshmoore:
Merging. Should minimally fix: #853 and #913

PR #930, opened 2022-01-05
Fix bug where the checksum of zipfiles is wrong

orenwatson:
There is a bug causing incorrect length being written to the zip file, because Zipfile thinks the len() of the passed object is the length in bytes, but it was passing a ndarray, whose len() is the number of rows. The fix is to convert to bytes before passing to zipfile.writestr()

More details on how to reproduce this bug are here:
https://github.com/zarr-developers/zarr-python/issues/931

### DIFF
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -1565,7 +1565,7 @@ class ZipStore(Store):
             else:
                 keyinfo.external_attr = 0o644 << 16     # ?rw-r--r--
 
-            self.zf.writestr(keyinfo, value)
+            self.zf.writestr(keyinfo, value.tobytes())
 
     def __delitem__(self, key):
         raise NotImplementedError
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -1551,7 +1551,7 @@ class ZipStore(Store):
     def __setitem__(self, key, value):
         if self.mode == 'r':
             raise ReadOnlyError()
-        value = ensure_contiguous_ndarray(value)
+        value = ensure_contiguous_ndarray(value).view("u1")
         with self.mutex:
             # writestr(key, value) writes with default permissions from
             # zipfile (600) that are too restrictive, build ZipInfo for
@@ -1565,7 +1565,7 @@ class ZipStore(Store):
             else:
                 keyinfo.external_attr = 0o644 << 16     # ?rw-r--r--
 
-            self.zf.writestr(keyinfo, value.tobytes())
+            self.zf.writestr(keyinfo, value)
 
     def __delitem__(self, key):
         raise NotImplementedError
--- a/zarr/tests/test_storage.py
+++ b/zarr/tests/test_storage.py
@@ -1549,6 +1549,13 @@ class TestZipStore(StoreTests):
             assert perm == '0o40775'
         z.close()
 
+    def test_store_and_retrieve_ndarray(self):
+        store = ZipStore('data/store.zip')
+        x = np.array([[1, 2], [3, 4]])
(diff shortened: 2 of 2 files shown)


joshmoore:
Hi @orenwatson, thanks for this! Do you have an idea what a failing test _before_ your fix would look like?

zarr/storage.py
```diff
@@ -1565,7 +1565,7 @@ def __setitem__(self, key, value):
             else:
                 keyinfo.external_attr = 0o644 << 16     # ?rw-r--r--
 
-            self.zf.writestr(keyinfo, value)
+            self.zf.writestr(keyinfo, value.tobytes())
```
jakirkham, line 1568: This would introduce an additional copy before writing to disk.

Guessing the issue is `ZipFile` assumes data here is `uint8` even if that may not be the case. Also suspect this only happens when a compressor is not used (otherwise this would already be the case)

Would suggest changing line 1554 above to add `.view("u1")` to cast to `uint8`. This should have the same end result (data is represented as `uint8`), but not cause a copy

jakirkham:
Thank Oren! 😄

Had a suggestion above. Could you please also include a test case that expose the original bug (thus confirming this fix)?

orenwatson:
It would I suppose look like writing a numpy array to a ZipStore (with [] = ), closing it, and trying to open it again? I was getting a exception thrown from zipfile library because of a failed cksum

zarr/storage.py
orenwatson: Good point. I'll write a test case as well.

jakirkham:
Looking at the code in issue ( https://github.com/zarr-developers/zarr-python/issues/931 ). Wondering what would happen if the file were closed, reopened, and entry `"foo"` read?

orenwatson:
image
On my machine, it didn't even require reopening the zipstore.

jakirkham:
Sounds like a good test case then 🙂

zarr/storage.py
orenwatson: Test is added now.

jakirkham (approved):
LGTM. Thanks Oren! 😄

zarr/tests/test_storage.py
```diff
@@ -1549,6 +1549,13 @@ def test_permissions(self):
             assert perm == '0o40775'
         z.close()
 
+    def test_store_and_retrieve_ndarray(self):
+        store = ZipStore('data/store.zip')
```
joshmoore, line 1553: I'm not actually sure why some of these tests use `data/store.zip` explicitly here, but I think for your test, @orenwatson, it would be best to use the `self.create_store()` method so that you are testing a fresh zip, independently of other tests.

I can confirm though that this test fails without this PR and passes with it. :+1:

NB: A _second_ run with this PR raises the warning:

```
/usr/local/anaconda3/envs/z/lib/python3.9/zipfile.py:1505: UserWarning: Duplicate name: 'foo'
  return self._open_to_write(zinfo, force_zip64=force_zip64)
```

which is what makes me think that `create_store` would be safer.

jakirkham: Good point Josh 👍

I think with `test_permissions` this is needed because the file needs to be flushed/closed and then reopened for the test (so it doesn't follow the norm). Though agree we should be able to use `create_store` here

joshmoore: Makes sense. If the test is passing, I say we release @orenwatson (and add a grand "Huzzah!" in thanks 🎉) and we can handle the cleanup separately.

PR #934, opened 2022-01-11
[REVIEW] Support of alternative array classes

madsbk:
This PR implements support of alternative array types (other than NumPy arrays) by introducing a new argument, `meta_array`, that specifies the type/class of the underlying array.

The `meta_array` argument can be any class instance that can be used as the `like` argument in NumPy (see [NEP 35](https://numpy.org/neps/nep-0035-array-creation-dispatch-with-array-function.html)).

Also, this PR implements support of CuPy arrays by introducing a CPU compressor:
```python 

class CuPyCPUCompressor(Codec):
    """CPU compressor for CuPy arrays

    This compressor moves CuPy arrays to host memory before compressing
    the arrays using `compressor`.

    Parameters
    ----------
    compressor : numcodecs.abc.Codec
        The codec to use for compression and decompression.
    """
```

#### Example

A common limitation in GPU programming is available GPU memory. In the following, we create a Zarr array of CuPy arrays that are saved to host memory using  the `Zlib` codec for compression/decompression:
```python
zarr.array(a, chunks=10, meta_array=cupy.empty(()), compressor=CuPyCPUCompressor(Zlib()))
```

If we don't want compression but still want to copy to/from host memory, we can do the following:
```python
zarr.array(a, chunks=10, meta_array=cupy.empty(()), compressor=CuPyCPUCompressor())
```

If the store handles CuPy arrays directly, we can disable compression:
```python
zarr.array(a, chunks=10, meta_array=cupy.empty(()), compressor=None)
```

#### Notice
* This feature becomes more relevant when we have stores that support direct read/write of GPU memory. E.g. a store that use 
[GPUDirect Storage (GDS)](https://docs.nvidia.com/gpudirect-storage/overview-guide/index.html) would be able to bypass the host memory when copying GPU memory to/from disk.
* This PR implements a new version of `ensure_ndarray()` and `ensure_contiguous_ndarray()` that accepts CuPy arrays as well as NumPy arrays. Instead, we should properly implement more generic versions of them in numcodecs (see https://github.com/zarr-developers/numcodecs/pull/212).
* This PR depend on NumPy version 1.20 since it uses [NEP 35](https://numpy.org/neps/nep-0035-array-creation-dispatch-with-array-function.html). Is this an issue?

#### TODO

cc. @jakirkham


zarr/core.py
```diff
@@ -175,6 +182,11 @@ def __init__(
         self._is_view = False
         self._partial_decompress = partial_decompress
         self._write_empty_chunks = write_empty_chunks
+        self._meta_array = meta_array
+        if meta_array is not None:
+            self._meta_array = np.empty_like(meta_array)
```
jakirkham, line 187: Should this be the same size as the other case?

```suggestion
            self._meta_array = np.empty_like(meta_array, shape=())
```

zarr/cupy.py
```diff
@@ -0,0 +1,59 @@
...
+from numcodecs.registry import get_codec, register_codec
+
+from .util import ensure_contiguous_ndarray
+
+
+class CuPyCPUCompressor(Codec):
+    """CPU compressor for CuPy arrays
+
+    This compressor converts CuPy arrays host memory before compressing
+    the arrays using `compressor`.
```
jakirkham, line 11: Should we be using host compressors on GPU data? Or should be using different compressors like those from [nvCOMP]( https://github.com/NVIDIA/nvcomp )?

cc @thomcom (in case you have thoughts here 🙂)

zarr/hierarchy.py
```diff
@@ -121,6 +128,11 @@ def __init__(self, store, path=None, read_only=False, chunk_store=None,
             self._key_prefix = ''
         self._read_only = read_only
         self._synchronizer = synchronizer
+        self._meta_array = meta_array
+        if meta_array is not None:
+            self._meta_array = np.empty_like(meta_array)
```
jakirkham, line 133:
```suggestion
            self._meta_array = np.empty_like(meta_array, shape=())
```

zarr/util.py
```diff
@@ -10,13 +11,121 @@
...
+        mem = memoryview(buf)
+
+        # instantiate array from memoryview, ensures no copy
+        arr = np.array(mem, copy=False)
+
+    return arr
+
+
+# TODO: move the function to numcodecs
+def ensure_contiguous_ndarray(buf, max_buffer_size=None):
```
jakirkham, line 73: Just flagging this so we remember to move to Numcodecs

Curious to know what changes were needed here 🙂

zarr/util.py
```diff
@@ -10,13 +11,121 @@
...
+
+    else:
+
+        # TODO: make it possible to register array-like types lazily.
+        try:
+            import cupy
+            if isinstance(buf, cupy.ndarray):
+                return buf
+        except ImportError:
+            pass
```
jakirkham, line 60: Wondering if we can use entrypoints as @martindurant is doing in PR ( https://github.com/zarr-developers/numcodecs/pull/300 )

This would be good to do in conjunction with hooking in other compressors

zarr/util.py
```diff
@@ -10,13 +11,121 @@
...
-from numcodecs.compat import ensure_ndarray, ensure_text
+from numcodecs.compat import ensure_text
 from numcodecs.registry import codec_registry
 from numcodecs.blosc import cbuffer_sizes, cbuffer_metainfo
 
 from typing import Any, Callable, Dict, Optional, Tuple, Union
 
 
+# TODO: move the function to numcodecs
+def ensure_ndarray(buf):
```
jakirkham, line 22: Were there other changes here outside the `try...except...` for CuPy below?

zarr/core.py
```diff
@@ -155,6 +161,7 @@ def __init__(
         cache_attrs=True,
         partial_decompress=False,
         write_empty_chunks=True,
+        meta_array=None,
```
jakirkham, line 164: I wonder if this could be picked up from the `Store` itself. That way a `Store` could specify its return type (NumPy or otherwise). This would save the user from getting involved as much in the process and hopefully make it easier for them to get started. WDYT?

@grlee77 would be good to get your thoughts as well in light of the `BaseStore` work 🙂

jakirkham (commented):
Thanks Mads! 😄

Have a few initial comments below.

Still thinking through if there are ways we can improve the `meta_array` interface from the end user perspective. Curious to get your thoughts on this as well 🙂

jakirkham:
cc @d-v-b @joshmoore (as we discussed this last week it would be good to get your takes as well 🙂)

jakirkham:
cc @d70-t ( in case you have thoughts here as well 🙂 )

joshmoore:
Knowing little about the requirements here, just looking at the example:

```
zarr.array(a, chunks=10, meta_array=cupy.empty(()), compressor=CuPyCPUCompressor(Zlib()))
```

I wonder a couple of things:
 - Could this be done higher-level? As @jakirkham mentions above, could array initialization get this from the store. Nothing against making it an option here, but I imagine you will typically want to ignore the specifics by the time you get to this point in your code. (We may need a higher-level configuration object to not need to pass every flag down to arrays)
 - Similarly, I wonder if a compressor wrapper is the best strategy. It may well be, but it would be good to have an idea of other related use cases. Are there any inspection methods on codecs to inform the wrappers of what they need to know? What other wrappers could we imagine? And as always, could we make it work with the user needing to remember to add the wrapper? (perhaps by making use of @data-apis style methods??)

zarr/core.py
```diff
@@ -1539,9 +1559,13 @@ def set_coordinate_selection(self, selection, value, fields=None):
         # setup indexer
         indexer = CoordinateIndexer(selection, self)
 
-        # handle value - need to flatten
+        # handle value - need ndarray-like flatten value
         if not is_scalar(value, self._dtype):
-            value = np.asanyarray(value)
+            try:
+                value = ensure_ndarray(value)
```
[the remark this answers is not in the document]
madsbk: Before, `bytearrays` are converted to  `uint8` arrays and `bytes` are converted to `|S` arrays.
Whereas now, both `bytes` and `bytearrays` are converted to a NumPy array of type `uint8`.

I don't think this make any difference in practice.

zarr/core.py
```diff
@@ -1772,8 +1796,9 @@ def _process_chunk(
                 self._dtype != object):
 
             dest = out[out_selection]
+            dest_is_writable = getattr(dest, "writeable", True)
```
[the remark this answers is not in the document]
madsbk: CuPy array doesn't have a `writeable` flag: https://github.com/cupy/cupy/issues/2616
Added
```python
# Assume that array-like objects that doesn't have a
# `writeable` flag is writable.
dest_is_writable = getattr(dest, "writeable", True)
```

zarr/core.py
```diff
@@ -2120,7 +2151,7 @@ def _encode_chunk(self, chunk):
             cdata = chunk
 
         # ensure in-memory data is immutable and easy to compare
-        if isinstance(self.chunk_store, MutableMapping):
+        if isinstance(self.chunk_store, KVStore):
```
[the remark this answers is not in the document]
madsbk: Since all store classes derived `BaseStore` also derives from `MutableMapping`, this check was always `True`.
Now by checking against `KVStore`, we actually check if the store is `in-memory`.

zarr/cupy.py
madsbk: Good point, it would make sense to implement a compressor, `CuPyGPUCompressor`,  that do the compression on the GPU.

However, in order to limit the scope of this PR, I think we should use the existing CPU compressors for now. Then in a follow up PR, we can implement compression directly in the GPU using something like nvCOMP.

zarr/util.py
madsbk: No, this is the only change

zarr/core.py
madsbk: I think it would be a good idea to have `Store` define a default `meta_array` but in the general case, I think a `Store` should be able to handle both NumPy and other types like CuPy arrays simultaneously.

d70-t (commented):
I'm with @joshmoore, that this example:
```python
z = zarr.array(a, chunks=10, meta_array=cupy.empty(()), compressor=CuPyCPUCompressor(Zlib()))
```
looks a little odd.

I'm wondering if we should make more of a distinction between 
* the compressor specification (the goal of how bytes should look like in the store) and
* the compressor implementation (the path of how to make and read those bytes specifically)

When thinking of what I'd want to express in this line, I'd say
> I want to put my array `a` onto a store (which is implicitly `dict` in this case), in chunks of 10 and it should be zlib-compressed.

I don't really care about how the data moves from `a` to the store (if a is a `numpy` array, I'd probably like to use `numcodecs` directly, if `a` is a `cupy` array, I'd like to use GDS and a GPU compressor if available and like to fall back to numcodecs if not). In that sense, the line should maybe read like `z = zarr.array(a, chunks=10, compressor=Zlib())`.

Afterwards, I'd probably want to get a selection back from the zarr-array `z`, and I might want that to live on the GPU (which **might be a different choice** from what `a` was...), this creates a bit of a problem. I might want to write something like:

```python
selection: cupy.ndarray = z[3:7]
```
... but Python doesnt't (yet?) work this way, so we need to pass the type-information differently into `z`. It might be
```python
selection = z.get(slice(3,7), meta_array=cupy.empty(()))
```
which is far less beautiful than
```python
selection = z[3:7]
```
For the last option to work, there must be a way to change the default return type of the `[]`-operator, so passing the `meta_array` into `zarr.array` additionally might be a viable option. Still something like the `get` might be useful for some, if the array might be used both on CPU and GPU.

In any case, up to now, the `CuPyCPUCompressor` (which I consider as a description of the path between array and store) would not be visible at any place, and it might also be different for different set / get calls of the same zarr array `z`. It might be given as an optional hint on creation of the array, if a specific path is desired. Probably even with a way of overriding it on set or get. In any case, I believe that having any implementation (path) -specific information stored in zarray metadata would be a bad thing. In particular, I'd argue that `cupy_cpu_compressor` should maybe not appear in `.zarray` metadata.

madsbk:
> * Could this be done higher-level? As @jakirkham mentions above, could array initialization get this from the store. Nothing against making it an option here, but I imagine you will typically want to ignore the specifics by the time you get to this point in your code. (We may need a higher-level configuration object to not need to pass every flag down to arrays)

Agree, it would be a good idea to have `Store` define a default `meta_array`.

> I'm with @joshmoore, that this example:
> 
> ```python
> z = zarr.array(a, chunks=10, meta_array=cupy.empty(()), compressor=CuPyCPUCompressor(Zlib()))
> ```
> 
> looks a little odd.

Sorry I should have been more clear, the `CuPyCPUCompressor` here it is meant as a method to use the existing stores and compressors together with CuPy without having to make them all GPU aware. This ways, we can implement GPU aware stores and compressors iteratively. 
The main goal of this PR is to make it possible to write GPU aware stores without having Zarr force-casting everything to NumPy arrays.

> When thinking of what I'd want to express in this line, I'd say
> 
> > I want to put my array `a` onto a store (which is implicitly `dict` in this case), in chunks of 10 and it should be zlib-compressed.

Agree, when writing arrays it should be easy enough to infer the array-like type but when creating a Zarr array from scratch, we don't have that information. 
Now if I understand you correctly, you are saying that we could delay the array-like type inference until the user reads or writes the Zarr array. And if we have compatible compressors, we can even write using one array-like type and read using another array-like type. 
I think this is a great idea and I think this PR actually gets us some of the way.

d70-t:
> Sorry I should have been more clear, the `CuPyCPUCompressor` here it is meant as a method to use the existing stores and compressors together with CuPy without having to make them all GPU aware. This way, we can implement GPU aware stores and compressors iteratively.

True, a compressor-wrapper seems to be a relatively light way to implement such GPU aware stores. But I'm wondering if that's really the API we'll want to end up with. Probably we'll also want to end up having such compressor-wrappers as a simple way to make more compressor-implementations, but maybe we don't want to write them out at this function call. That's about what I meant with separating compressor-implementations and compressor-specifications. We likely want to create zlib-compressed data on the store (the specification), but might want to use the CuPyCPU(Zlib) implementation for the transformation at that very moment of writing the data.

> Agree, when writing arrays it should be easy enough to infer the array-like type but when creating a Zarr array from scratch, we don't have that information.

Do we need that information, when creating the array from scratch? The array will probably only contain the metadata keys and those should maybe be independent of the compressor-implementation.

> Now if I understand you correctly, you are saying that we could delay the array-like type inference until the user reads or writes the Zarr array. And if we have compatible compressors, we can even write using one array-like type and read using another array-like type.

Yes, that's about what I had in mind. It would be great to have arrays of any kind on one side and stores and compression specifications on the other side and some flexible, auto-optimizing routing in between. I believe that the specific algorithm of how to compress / decompress and move the data in between should neither belong to the store nor to the array-type.

In particular, it seems to be a totally valid approach to use e.g. a CPU to write data to some (e.g. S3-) store and read data back from the same store later on via GPU.
If one might want to go totally crazy, one might even write some parts of an array using CPU and other parts using GPU (I know of some people in C++ communities who work on parallelizing loops such that parts of the same loop run on available CPUs and parts run on available GPUs in order to get the most out of a system).

d70-t:
I just stumbled again across the mentioned entrypoints... It's probably worth a thought if something roughly like:
```python
compressors =[
    {"array": "numpy", "algorithm": "zlib": "compressor": Zlib, "priority": 1},
    {"array": "cupy", "algorithm": "zlib": "compressor": CuPyCPUCompressor(Zlib), "priority": 2},
    {"array": "cupy", "algorithm": "zlib": "compressor": CuPyGDSZlib, "priority": 1},
    ...
]
```
could be established via entrypoints? That list would be dependent on which libraries are present on the system and would help finding the right compressor implementation for a given combination of array and codec.... Still I don't have though yet about how that might interfere with `filters`, which may also play a role...

madsbk:
@d70-t I agree with your vision but I think we should move in small steps.  

I think we all can agree that we need some way to specify the desired array-like type of a `getitem` operation. For backward compatibility and for convenience, I suggest that we specify this through the `meta_array` argument on `zarr.Array` creation. But this is only the default array-like type, I agree with @d70-t, it should be possible to overwrite the default type using something like: 

```python
selection = z.get(slice(3,7), meta_array=cupy.empty(())) 
``` 

And as @joshmoore suggest, we might even make it higher-level configurable (e.g. through a config file and/or context manager). 

However, I think this is outside of the scope of this PR. The goal here is to avoid force-casting all data to NumPy arrays. 

 

> In any case, up to now, the `CuPyCPUCompressor` (which I consider as a description of the path between array and store) would not be visible at any place, and it might also be different for different set / get calls of the same zarr array `z`. It might be given as an optional hint on creation of the array, if a specific path is desired. Probably even with a way of overriding it on set or get. In any case, I believe that having any implementation (path) -specific information stored in zarray metadata would be a bad thing. In particular, I'd argue that `cupy_cpu_compressor` should maybe not appear in `.zarray` metadata.


 

 

Again, I agree. I can move the implementation of `CuPyCPUCompressor` to `test_cupy.py` to make it clear that it is just a way to use CuPy arrays with the current API and is not part of the official API.

d70-t:
:+1: for moving in small steps. And I also agree to the rest of the post @madsbk

The only thing vision-wise which I believe should be accounted for right now, would be that indications of the (in-memory) array type should **not** be part of the zarr-metadata (and I'm still uncertain about if it would be ok to do so as a part of a test. Probably that's ok). But that's more a personal feeling and should maybe be up to discussion and others might have different opinions.

jakirkham:
For more context this is intended to be used with [kvikio]( https://github.com/rapidsai/kvikio ), which handles the data loading from disk to GPU

quasiben:
@joshmoore do you know the minimum version of numpy used in the tests ?  I believe [NEP-35](https://numpy.org/neps/nep-0035-array-creation-dispatch-with-array-function.html) was included in NumPy 1.20

joshmoore:
Ah, ok. There's a job that pins to 1.17 for testing the minimal version.

quasiben:
We use the following in dask to check:

```
_numpy_120 = _np_version >= parse_version("1.20.0")
@pytest.mark.skipif(not _numpy_120, reason="NEP-35 is not available")
def test_foo()
```

Would that be ok here ?

quasiben:
@madsbk can you restrict the test to only run when `numpy.__version__ >= 1.20.0` .  I imagine this is the last piece here

grlee77:
New conflicts due to merge of the v3 branch are resolved in PR https://github.com/madsbk/zarr-python/pull/1

madsbk:
> New conflicts due to merge of the v3 branch are resolved in PR [madsbk#1](https://github.com/madsbk/zarr-python/pull/1)

Thanks @grlee77 !

madsbk:
### The plan going forward

1. Wait for https://github.com/zarr-developers/zarr-python/pull/988 to be merged. 
2. Simplify this PR and reduce the scope to the original goal of avoiding force-casting all data to NumPy arrays.
3. Move the `CuPyCPUCompressor` class to the test suite, it will not be part of the API.
4. Merge this PR.
5. Gain experience using Zarr with CuPy arrays.
6. Introduce new API and semantic to ease the transition between NumPy and CuPy backed Zarr arrays.

madsbk:
This is ready for another round of reviews

cc. @jakirkham, @d70-t

joshmoore:
@madsbk: I'm slowly working my way towards a re-review here, but in the meantime, any thoughts on the codecov front?

madsbk:
> @madsbk: I'm slowly working my way towards a re-review here, but in the meantime, any thoughts on the codecov front?

The codecov errors can be ignored. As far as I can see, all of them are false negatives.

joshmoore:
> The codecov errors can be ignored. As far as I can see, all of them are false negatives.

Or is the issue that the cupy tests aren't being run and therefore the `meta_array` property is not accessed anywhere?

madsbk:
> Or is the issue that the cupy tests aren't being run and therefore the `meta_array` property is not accessed anywhere?

You are absolutely right! No idea how I could miss that :)

Adding tests of `meta_array` when CuPy isn't available.

joshmoore:
Hi @madsbk. No, sorry. This just landed on the vacation :mountain:. I'll start reviewing for a 2.13 release ASAP.

joshmoore:
Rolling into a 2.13 pre-release. If anyone has any lingering concerns about the API, there would still be a (small) window to address them in.

jakirkham:
Thanks Mads for working on this and Josh for reviewing! 🙏

Issue #877, opened 2021-11-18
Add Sharding Support


jstriebel:
Update 2022-08-22:

Please note that a sharding proposal is now formalized as a [**Zarr Enhancement Proposal, ZEP 2**](https://zarr.dev/zeps/draft/ZEP0002.html).

A new efficient implementation of sharding is being discussed as part of https://github.com/zarr-developers/zarr-python/discussions/1569.

----

TLDR: We want to add sharding to zarr. There are some PRs as starting points for further discussions, as well as issue https://github.com/zarr-developers/zarr-specs/issues/127 to update the spec.
* #876,
* a PR in the clean room implementation zarrita in the spirit of the previous PR, but simplified as much as possible
  https://github.com/alimanfoo/zarrita/pull/40, and
* another prototype where the translation logic happens purely in the Array class
  #947

Please see https://github.com/zarr-developers/zarr-python/issues/877#issuecomment-1031475493 for a comparison of the implementation approaches.

---- 

Currently, zarr maps one array chunk to one storage key, e.g. one file for the `DirectoryStore`. It would be great to decouple the concept of chunks (e.g. one compressible unit) and storage keys, since the storage might be optimized for larger data per entry and might have an upper limit for the number of entries, such as the file block size and maximum inode number for disk-storage. This does not necessarily fit the access patterns of the data, so chunks might need to be smaller than one storage key.

For those reasons, we would like to add sharding support to zarr. One shard corresponds to one storage key, but can contain multiple chunks:
![sharding](https://user-images.githubusercontent.com/7216331/142424772-fa0b700a-8176-4663-a136-9807e64f7078.png)

We, this is [scalable minds](https://scalableminds.com/), will provide a PR for initial sharding support in zarr-python. This work is funded by the CZI through the [EOSS program](https://chanzuckerberg.com/eoss).

We see the following **requirements** to implement sharding support:

1. Add shard abstraction (multiple chunks in a single file/storage object)
2. Chunks should continue to be compressible (blosc, lz4, zstd, …)
3. Chunks should be stored in a continuous byte stream within the shard file
4. Shard abstraction should be transparent when reading/writing
5. Support arbitrary array sizes (not necessarily chunk/shard aligned)
6. Support arbitrary number of dimensions
7. Possibly store chunks in Morton-order within a shard file

With this issue and an accompanying prototype PR we want to start a discussion about **possible implementation approaches**. Currently, we see four approaches, which have different pros and cons:

[...]

PR #876, opened 2021-11-18
Sharding Prototype I: implementation as translating Store

jstriebel:
This PR is for an early prototype of sharding support, as described in the corresponding issue #877. It serves mainly to discuss the overall implementation approach for sharding. This PR is not (yet) meant to be merged.

This prototype
* allows to specify shards as the number of chunks that should be contained in a shard (e.g. using `arr.zeros((20, 3), chunks=(3, 3), shards=(2, 2), …)`).
   One shard corresponds to one storage key, but can contain multiple chunks:
  sharding
* ensures that this setting is persisted in the `.zarray` config and loaded when opening an array again, adding two entries:
    * `"shard_format": "indexed"` specifies the binary format of the shards and allows to extend sharding with other formats later
    * `"shards": [2, 2]` specifies how many chunks are contained in a shard,
* adds a `IndexedShardedStore` class that is used to wrap the chunk-store when sharding is enabled. This store handles the grouping of multiple chunks to one shard and transparently reads and writes them via the inner store in a binary format which is specified below. The original store API does not need to be adapted, it just stores shards instead of chunks, which are translated back to chunks by the `IndexedShardedStore`.
* adds a small script `chunking_test.py` for demonstration purposes, this is not meant to be merged but servers to illustrate the changes.

-----------------

The currently implemented file format is still up for discussion. It implements "Format 2" @jbms describes in https://github.com/zarr-developers/zarr-python/pull/876#issuecomment-973462279.

Chunks are written successively in a shard (unused space between them is allowed), followed by an index referencing them.
The index holding an `offset, length` pair of little-endian uint64 per chunk, the chunks-order in the index is row-major (C) order,
e.g. for (2, 2) chunks per shard an index would look like:

```
| chunk (0, 0)    | chunk (0, 1)    | chunk (1, 0)    | chunk (1, 1)    |
| offset | length | offset | length | offset | length | offset | length |
| uint64 | uint64 | uint64 | uint64 | uint64 | uint64 | uint64 | uint64 |
```

Empty chunks are denoted by setting both offset and length to `2^64 - 1`. All the index always has the full shape of all possible chunks per shard, even if they are outside of the array size.

For the default order of the actual chunk-content in a shard I'd propose to use Morton order, but this can easily be changed and customized, since any order can be read.

-----------------

If the overall direction of this PR is pursued, the following steps (and possibly more) are missing:
* Functionality
  * [ ] Use a default write-order (Morton) of chunks in a shard and allow customization
[...]

PR #947, opened 2022-01-19
Sharding Prototype II: implementation on Array

jstriebel:
This PR is for an early prototype of sharding support, as described in the corresponding issue #877, an alternative to it's cousin PR #876. Please also see the discussion there. It serves mainly to discuss the overall implementation approach for sharding. This PR is not (yet) meant to be merged.

This prototype
* allows to specify shards as the number of chunks that should be contained in a shard (e.g. using `arr.zeros((20, 3), chunks=(3, 3), shards=(2, 2), …)`).
   One shard corresponds to one storage key, but can contain multiple chunks:
  sharding
* ensures that this setting is persisted in the `.zarray` config and loaded when opening an array again, adding two entries:
    * `"shard_format": "indexed"` specifies the binary format of the shards and allows to extend sharding with other formats later
    * `"shards": [2, 2]` specifies how many chunks are contained in a shard,
* adds a small script `sharding_test.py` for demonstration purposes, this is not meant to be merged but servers to illustrate the changes.

-----------------

The currently implemented file format is still up for discussion. It implements "Format 2" @jbms describes in https://github.com/zarr-developers/zarr-python/pull/876#issuecomment-973462279.

Chunks are written successively in a shard (unused space between them is allowed), followed by an index referencing them.
The index holding an `offset, length` pair of little-endian uint64 per chunk, the chunks-order in the index is row-major (C) order,
e.g. for (2, 2) chunks per shard an index would look like:

```
| chunk (0, 0)    | chunk (0, 1)    | chunk (1, 0)    | chunk (1, 1)    |
| offset | length | offset | length | offset | length | offset | length |
| uint64 | uint64 | uint64 | uint64 | uint64 | uint64 | uint64 | uint64 |
```

Empty chunks are denoted by setting both offset and length to `2^64 - 1`. All the index always has the full shape of all possible chunks per shard, even if they are outside of the array size.

For the default order of the actual chunk-content in a shard I'd propose to use Morton order, but this can easily be changed and customized, since any order can be read.


mkitti:
Cross posting from https://github.com/zarr-developers/zarr-python/pull/876#issuecomment-1088902921. Perhaps [`nbytes`](https://github.com/zarr-developers/zarr-python/blob/6e80e97a7f3decc4c9576b27d626af3b3824f8f3/zarr/core.py#L452-L454) is a less ambiguous term than `length` to describe the number of bytes in a chunk?

jstriebel:
A similar proposal is now formalized as a [Zarr Enhancement Proposal, ZEP 2](https://zarr.dev/zeps/draft/ZEP0002.html). The basic sharding concept and the binary representation is the same, however it is formalized as a storage transformer for Zarr v3.

Further feedback is very welcome! Either
* on the abstract specification as comments on PR https://github.com/zarr-developers/zarr-specs/pull/152, or
* on the Python implementation PR https://github.com/zarr-developers/zarr-python/pull/1111.

I'll close this PR to consolidate the open sharding threads, please feel free to continue any discussions or add feedback on the PRs mentioned above.

PR #951, opened 2022-01-26
Default write empty chunks false

joshmoore:
Re-opening https://github.com/zarr-developers/zarr-python/pull/853 from another branch to see if that fixes the github action jobs.

### DIFF
--- a/.github/workflows/minimal.yml
+++ b/.github/workflows/minimal.yml
@@ -25,11 +25,11 @@ jobs:
       run: |
         conda activate minimal
         python -m pip install .
-        pytest -svx
+        pytest -svx --timeout=300
     - name: Fixture generation
       shell: "bash -l {0}"
       run: |
         conda activate minimal
         rm -rf fixture/
-        pytest -svx zarr/tests/test_dim_separator.py zarr/tests/test_storage.py
+        pytest -svx --timeout=300 zarr/tests/test_dim_separator.py zarr/tests/test_storage.py
         # This simulates fixture-less tests in conda and debian packaging
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -73,7 +73,7 @@ jobs:
         conda activate zarr-env
         mkdir ~/blob_emulator
         azurite -l ~/blob_emulator --debug debug.log 2>&1 > stdouterr.log &
-        pytest --cov=zarr --cov-config=.coveragerc --doctest-plus --cov-report xml --cov=./
+        pytest --cov=zarr --cov-config=.coveragerc --doctest-plus --cov-report xml --cov=./ --timeout=300
     - uses: codecov/codecov-action@v1
       with:
         #token: ${{ secrets.CODECOV_TOKEN }} # not required for public repos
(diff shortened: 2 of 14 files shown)


jakirkham:
Also trying a no change PR ( https://github.com/zarr-developers/zarr-python/pull/952 ) to see if this is a CI issue

joshmoore:
Fewer are hanging but not nothing. 😵‍💫

jakirkham:
It would be good to get a clearer idea on why that is happening. Did we manage to get a log that we could look at? Waited for a while to see if GH Actions would include output, but nothing showed up.

jni:
@joshmoore @jakirkham this is *very* reminiscent of my own misadventures with `test_sync`, deadlocks, and CI timeouts. @joshmoore want to try your own advice?

https://github.com/zarr-developers/zarr-python/pull/725#issuecomment-853813097

>  Usually slow tests are printed at the end, but if they aren't completing we don't get the list! Can you add pytest -sv to the workflow or add pypi.org/project/pytest-timeout so we can get an idea on what's gone sideways?

This commit adds the timeout:

[`4e7733c` (#725)](https://github.com/zarr-developers/zarr-python/pull/725/commits/4e7733c0f0640c66d223dd769cab6f16ea037961)

joshmoore:
Heh. I think I thought that that was now done everywhere. Pushing a commit with lots of timeouts here now.

jakirkham:
The thing is this hang showed up very recently. It makes me wonder if there is something we changed or a dependency update in the environment that changed, which is causing issues.

For example `fasteners` was updated recently. To save writing this again, please read this comment ( https://github.com/zarr-developers/zarr-python/pull/952#issuecomment-1022793956 ).

joshmoore:
Interesting that not even `timeout` seems to be working here.

joshmoore:
0.16.3 seemes safe, @jakirkham. Trying 0.17.1 here now.

joshmoore:
0.17.1 seems to be hanging on 3.9 != 1.21.0 as well.

joshmoore:
Pushed the commits up to the 0.16.3 pin to #853.

jakirkham:
The fact that timeout doesn't work makes me wonder if the hang is in C. Is it possible this is NumPy related? Is NumPy 1.22 hanging somewhere? What if we pinned to NumPy 1.20 in the jobs currently skipping 1.21?

jakirkham:
Yeah sorry I think you may be right. Going with 0.16 seems like the right idea. There were deadlock issues in previous 0.17 releases it seems

joshmoore:
> Going with 0.16 seems like the right idea.

Yeah, I was feeling pretty good about it until https://github.com/zarr-developers/zarr-python/pull/853/commits/080a854e28bd386fe1bbef9620aba2a83d1a909c timed out again. But perhaps that PR is somehow "tainted".

jakirkham:
Maybe we should run tests with more verbosity? Perhaps that points out which test hangs so we can investigate further

jakirkham:
Power cycling for CI

jakirkham:
Looks like this is going to fail without the NumPy upgrade ( https://github.com/zarr-developers/zarr-python/pull/955 )

joshmoore:
Closing along with https://github.com/zarr-developers/zarr-python/pull/952

Looks like the scare is over.

PR #950, opened 2022-01-26
add release note regarding introduction of BaseStore and KVStore

grlee77:
As requested by @joshmoore, this PR adds a release note regarding the recent addition of `BaseStore` and `KVStore` in Zarr 2.11. Please suggest improvements/clarifications!

I think the changes should be transparent to the typical user of Zarr (especially those using it indirectly via Dask or Xarray). It is mainly important for library authors or those developing their own store classes.

TODO:
* [ ] GitHub Actions have all passed

PR #952, opened 2022-01-26
[NO MRG] Test CI

jakirkham:
Opening just to test the current state of CI

### DIFF
--- a/requirements_dev_minimal.txt
+++ b/requirements_dev_minimal.txt
@@ -1,6 +1,6 @@
 # library requirements
 asciitree==0.3.3
-fasteners<0.17
+fasteners==0.17.3
 numcodecs==0.9.1
 msgpack-python==0.5.6
 setuptools-scm==6.4.2
--- a/requirements_dev_minimal.txt
+++ b/requirements_dev_minimal.txt
@@ -1,6 +1,6 @@
 # library requirements
 asciitree==0.3.3
-fasteners==0.17.2
+fasteners<0.17
 numcodecs==0.9.1
 msgpack-python==0.5.6
 setuptools-scm==6.4.2
--- a/requirements_dev_minimal.txt
+++ b/requirements_dev_minimal.txt
@@ -1,6 +1,6 @@
 # library requirements
 asciitree==0.3.3
-fasteners==0.17.3
+fasteners==0.17.2
 numcodecs==0.9.1
 msgpack-python==0.5.6
 setuptools-scm==6.4.2
--- a/requirements_dev_minimal.txt
+++ b/requirements_dev_minimal.txt
@@ -1,6 +1,6 @@
 # library requirements
 asciitree==0.3.3
-fasteners==0.17.2
+fasteners<0.17.2
 numcodecs==0.9.1
 msgpack-python==0.5.6
 setuptools-scm==6.4.2
--- a/requirements_dev_minimal.txt
+++ b/requirements_dev_minimal.txt
@@ -1,6 +1,6 @@
 # library requirements
 asciitree==0.3.3
-fasteners<0.17.2
+fasteners==0.17.2
 numcodecs==0.9.1
 msgpack-python==0.5.6
 setuptools-scm==6.4.2
--- a/.github/workflows/minimal.yml
+++ b/.github/workflows/minimal.yml
@@ -25,11 +25,11 @@ jobs:
       run: |
         conda activate minimal
         python -m pip install .
-        pytest -svvvx
+        pytest -svx
     - name: Fixture generation
       shell: "bash -l {0}"
       run: |
         conda activate minimal
         rm -rf fixture/
-        pytest -svvvx zarr/tests/test_dim_separator.py zarr/tests/test_storage.py
+        pytest -svx zarr/tests/test_dim_separator.py zarr/tests/test_storage.py
         # This simulates fixture-less tests in conda and debian packaging
--- a/.github/workflows/minimal.yml
+++ b/.github/workflows/minimal.yml
@@ -25,11 +25,11 @@ jobs:
       run: |
         conda activate minimal
         python -m pip install .
-        pytest -svx
+        pytest -svvvx
     - name: Fixture generation
       shell: "bash -l {0}"
       run: |
         conda activate minimal
         rm -rf fixture/
-        pytest -svx zarr/tests/test_dim_separator.py zarr/tests/test_storage.py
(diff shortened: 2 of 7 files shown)


jakirkham:
Am seeing the same failure with commit ( https://github.com/zarr-developers/zarr-python/commit/e15001bbe756520df70a43e025537f1eef6e7228 ) on `master` (the latest change). Going to try reverting that to see what happens

requirements_dev_minimal.txt
```diff
@@ -1,6 +1,6 @@
 # library requirements
 asciitree==0.3.3
-fasteners==0.17.3
+fasteners==0.17.2
```
joshmoore, line 3: Did you mean `<` here as well?

jakirkham:
Here's a diff between CI for commit ( https://github.com/zarr-developers/zarr-python/commit/21739ad12ba807b28534c814cd6b35d05775a575 )'s [run]( https://github.com/zarr-developers/zarr-python/runs/4878605730 ) and commit ( https://github.com/zarr-developers/zarr-python/commit/e15001bbe756520df70a43e025537f1eef6e7228 )'s [run]( https://github.com/zarr-developers/zarr-python/runs/4948772851 ) where the issue first appeared

<details>
<summary>Diff:</summary>

```diff
27c27
< coverage==6.2
---
> coverage==6.3
38c38
< fasteners==0.17.2
---
> fasteners==0.17.3
62c62
< jsondiff==1.3.0
---
> jsondiff==1.3.1
68c68
< jupyter-client==7.1.1
---
> jupyter-client==7.1.2
78c78
< moto==2.3.2
---
> moto==3.0.1
82c82
< multidict==5.2.0
---
> multidict==6.0.2
90c90
< notebook==6.4.7
---
> notebook==6.4.8
103c103
< prometheus-client==0.12.0
---
> prometheus-client==0.13.0
107c107
< pure-eval==0.2.1
---
> pure-eval==0.2.2
115c115
< pyparsing==3.0.6
---
> pyparsing==3.0.7
```

</details>

jakirkham:
Spotted issue ( https://github.com/harlowja/fasteners/issues/86 ), which may also be related.

jakirkham:
Looks like it was still hanging even with 0.16. Going to try running with more verbosity.

jakirkham:
Canceled the tests and looked at the tests run. They were definitely using the `ProcessSynchronizer`. There wasn't any one test that was jamming things up. Each job had a different test running when canceled. This makes me think there are still some locking issues, but when they crop up is sporadic.

jakirkham:
@joshmoore would be good to get your thoughts here on what we should do next? Should we debug further? Should we xfail the `ProcessSynchronizer` tests and go ahead with the release? Or should we do something else?

joshmoore:
Hmmmm..... so we're sure it's just a change in fasteners and not (also) a change in the mainline that's leading to this?

It would _definitely_ be good to get 2.11 out. We could move the synchronizer tests out to their own workflow (which only gets run periodically?)

jakirkham:
TO be completely honest, I'm not sure.

At the moment I'm torn between whether a new copy of `fasteners` is sneaking in somewhere despite the version constraints or whether there is some subtle locking issue introduced by not writing empty chunks (or maybe something else?).

That said, CI on `master` was passing until the most recent commit, which coincides with the recent `fasteners` release.

Davis' change ( https://github.com/zarr-developers/zarr-python/commit/831e687599da80b80ac3e97f1b0cef6f3f09b40f ) went in a couple months back (October) and issues only cropped up recently. Think we would have seen issues sooner if that was the cause.

All of the testing here is without the recent empty chunks change and it still has issues. So doesn't seem like that's the cause.

Finally where the tests slow down is when testing `ProcessSynchronizer`, which is the one place `fasteners` is used.

To summarize, there isn't an obvious cause outside of `fasteners`. Though it is disquieting that pinning doesn't seem to work

joshmoore:
Currently failing to reproduce locally with:

```
pytest -svvvx --count=1000 --random-order zarr/tests/test_sync.py
```

psarka:
Hi, fasteners maintainer here, saw a note in our issue about the deadlock, got interested. I'm puzzled by this thing, looking at the commit log I see that your tests passed 9 days ago with fasteners 17.2, but didn't with 2 days ago with fasteners 17.3:
image

Which is weird, as you are using ProcessLock, which was not touched in 17.3. Maybe it would be helpful to rerun your CI on the commit 21739ad12ba807b28534c814cd6b35d05775a575 and see if it still passes?

joshmoore:
Thanks, @psarka. Happy to give it a try. The previous action run was:

https://github.com/zarr-developers/zarr-python/runs/4878605616?check_suite_focus=true
Screen Shot 2022-01-28 at 15 38 16

I've just re-launched. Let's see what happens.

psarka:
Looks like it is hanging :exploding_head: 

Is there any other explanation, other than that something changed with github actions infrastructure? I'll rerun fasteners CI just in case.

jakirkham:
Thanks for taking a look Paulius! 😄

Yeah the hang seems to be occurring with the `ProcessSynchronizer` tests, which use `ProcessLock`. That was what we were seeing here as well.

That said, we probably need to turn this into a smaller reproducer so it is a bit easier to dig into.

joshmoore:
I got side tracked by trying to reproduce locally with https://github.com/nektos/act (unsuccessfully). I will give up on that and try to find a way forward.

joshmoore:
Plan discussed with @jakirkham:
 - I will try on bare-metal linux & linux/docker to try to reproduce.
 - John will try to tease apart whether or not the version of fasteners that we expect is actually being installed.
 - If we can't find a solution by tomorrow EOB, we will disable the synchronizer test in GHA and move forward with the 2.11 release.

joshmoore:
Linux & Linux/docker don't show a hang locally when following the python-package.yml workflow instructions. Currently running with `--count=1000` but I don't expect it will show anything.

jakirkham:
More strange is we saw PR ( https://github.com/zarr-developers/zarr-python/pull/955 ) pass. Trying to see if we can replicate that elsewhere. If so, maybe it was a GH Action's change?

joshmoore:
Maybe it was a temporary issue in our segment of the GHA cloud?! (if so, reminder for next time: contact GH earlier)

jakirkham:
Checking here as well ( https://github.com/zarr-developers/zarr-python/pull/950 ) just to make sure this is consistently working

jakirkham:
Looks like things are working 🎉

Sorry for the false alarm @psarka and thanks for the help 🙂

Going to go ahead and close

PR #963, opened 2022-02-10
Fix a TOCTOU race in `open_array(mode="a")`

alanhdu:
If you have multiple writers trying todo `zarr.open_array(..., "a")`,
the old code code sometimes raise a `ContainsArrayError`, even thogh
opening an already-existing array should be fine.

The underlying problem is a time-of-check/time-of-use race condition,
where the array can be created (by another writer) between the
`contains_array` check and the `init_array` call.

To fix this, we simply swallow the error from the `init_array` call.

I'm not sure how best to test this (since the bug is inherently non-deterministic based on scheduling), but I'm happy to take suggestoins.


joshmoore:
Thanks, @alanhdu. Do you have any ideas for a test that would trigger the race condition so we can keep the code coverage up?

joshmoore:
@alanhdu: I've added my test proposal but I've not yet been able to reproduce a failure.

p.s. @Alt-Shivam, thanks!
```
flake8...................................................................Failed
- hook id: flake8
- exit code: 1

zarr/tests/test_creation.py:733:5: F841 local variable 'expected_zarr_version' is assigned to but never used

Check Yaml...............................................................Passed
```

jhamman:

PR #966, opened 2022-02-15
Correct default fill_value for creation.create

### LINKED ISSUE #965: Incorrect default fill value causes byte arrays to become numeric when `write_empty_chunks=False`
Minimal example:
```python
import zarr
a = zarr.create((1,), dtype=bytes)
a[0] = b''
assert a[0] == b''

  File "example.py", line 5, in <module>
    assert a[0] == b''
AssertionError
```
The value of `a[0]` is actually `0`, when it should be `b''`.

Found by one of our users at https://github.com/tskit-dev/tsinfer/issues/628 this bug was introduced in the latest release (`v2.11.0`) in this commit: https://github.com/zarr-developers/zarr-python/commit/f461eb78fbb88187582cd9123d6ec7622d9abd26 when the default for `write_empty_chunks` was changed to `False`. The default `fill_value` for arrays created via `zarr.creation.create` is `0`, so when an empty, unwritten chunk is re-created the previous value `b''` becomes `0`. I assume this `fill_value` should be `None`.

benjeffery:
Attempting to fix #965

I have added a small test in what I think is the right place. This bug might have been caught if the tests in `test_core.py` used `creation.create` instead of `self.create_array`. Maybe I am missing something, but I couldn't see any obvious reason why those tests have their own creation routine that uses `storage.init_array` (which has `fill_value=None`). I'm happy to make that change if it is the right thing to do, but it breaks a lot of the nbytes and hexdigest tests, so I didn't propose it in the PR.
Thanks!

TODO:


zarr/creation.py
```diff
@@ -16,7 +16,7 @@
 
 
 def create(shape, chunks=True, dtype=None, compressor='default',
-           fill_value=0, order='C', store=None, synchronizer=None,
+           fill_value=None, order='C', store=None, synchronizer=None,
```
jakirkham, line 19: Are there any other effects from changing this from `0` to `None`?

Should we be updating `fill_value`s elsewhere to match?

jakirkham:
Thanks Ben! 😄

Had a couple questions above 🙂

zarr/creation.py
benjeffery: The only other use of `fill_value=0` I can see is `creation.open_array`. I assume this should probably be `None` too, but I'm not too familiar with all the effects here.

benjeffery:
Looks like I've broken a bunch of tests here too - I was only running a subset for speed locally. WIll fix up tomorrow.

zarr/creation.py
jakirkham: `open_array` just [calls `init_array` if needed]( https://github.com/zarr-developers/zarr-python/blob/f461eb78fbb88187582cd9123d6ec7622d9abd26/zarr/creation.py#L515-L539 ), which already has [`fill_value=None`]( https://github.com/zarr-developers/zarr-python/blob/22ded1d656cb2034180cf8a3f23f82b51f79d8b4/zarr/storage.py#L231 ). So making it `None` in `open_array` seems more consistent with the underlying function.

jakirkham:
The tests that fail seem to expect `0` (as opposed to `None`). Wondering how much difference there is between these two choices.

d-v-b:
Because of [this check](https://github.com/zarr-developers/zarr-python/blob/master/zarr/util.py#L665), arrays with a fill value of `None` cannot have empty chunks from the perspective of the `write_empty_chunks` machinery. So if `None` is the default fill value, then arrays will default to writing empty chunks, regardless of the state of the `write_empty_chunks` flag, which is a bit unfortunate.  However, a sensible default fill value kind of depends on the datatype of the array (e.g., it's not clear that 0 is a good fill value for `int16` data), so I can see this getting rather complicated.

jakirkham:
Interesting we might want to revisit that check then. Most of the code uses `fill_value=None` as the default. As Ben points out, this case where `fill_value=0` is an exception.

Typically anywhere we have a `fill_value`, we pass it into [`normalize_fill_value`]( https://github.com/zarr-developers/zarr-python/blob/22ded1d656cb2034180cf8a3f23f82b51f79d8b4/zarr/util.py#L263 ). However that normally keeps `None`.

If no `fill_value` is set, zarr-python treats that as `0` (for [example]( https://github.com/zarr-developers/zarr-python/blob/f461eb78fbb88187582cd9123d6ec7622d9abd26/zarr/core.py#L864-L866 )).

Generally though the Zarr v2 spec [leaves this ambiguous]( https://zarr.readthedocs.io/en/stable/spec/v2.html?highlight=fill_value#chunks ):

> If the “fill_value” field is null then the contents of the chunk are undefined.

jni:
>Generally though the Zarr v2 spec [leaves this ambiguous](https://zarr.readthedocs.io/en/stable/spec/v2.html?highlight=fill_value#chunks):
>
>  > If the “fill_value” field is null then the contents of the chunk are undefined.

😲

benjeffery:
I've pushed up quick fixes to the tests here.

joshmoore:
Still a number errors of the form:

```
        # compare properties
        for k, v in expect_props.items():
>           assert v == getattr(copied, k)
E           assert None == 0
E            +  where 0 = getattr(<HDF5 dataset "baz": shape (100,), type "<i8">, 'fillvalue')
```

not rolling this into 2.11.1.

joshmoore:
@benjeffery: let me know if help is needed here.

jni:
Does this mean that numeric types will default to None also? I find that suboptimal, and I think 0 is a good default fill value for numeric types. So I would use `normalize_fill_value` to have numeric arrays default to 0.

However, I really don't have strong feelings about the issue here. But if numeric arrays will *need* to specify a fill_value, then some loud documentation needs to exist about that... Again, this is all based on my own default expectations, which were *always* sparse writing and 0 default, but might not match others'.

benjeffery:
I'm not sure my fixes here are the correct way to go about this, and that some of the ideas in #965 are the right way to do this.

PR #974, opened 2022-02-28
Fix indexing for scalar numpy values (#967)

### LINKED ISSUE #967: Change in indexing behaviour with numpy ints
After https://github.com/zarr-developers/zarr-python/commit/7c31f041ca8baab860564d9c6332c52f19777e80 there was a breaking change to indexing behavior when a numpy integer type is passed:

For the following array:
```python
z = zarr.group()
a = z.create_dataset("test",shape=(10,))
a[:]=range(10)
```
Before this commit:
`a[np.int64(1)]` has the value `1.0`
After:
`a[np.int64(1)]` has the value `[1.]`

For both cases `a[1]` stays as `1.0` as expected. Was this change expected? This happens a lot when iterating over one numpy array and indexing with it as the iterable is of the numpy type. The fix in our application was to cast to a python `int` before indexing.

EDIT: Also note that numpy indexing returns the single value as was the case before this commit.

benjeffery:
Fixes #967 by adding a check for scalar numpy types.

TODO:
* [X] New/modified features documented in docs/tutorial.rst

### DIFF
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -6,6 +6,13 @@ Release notes
 Unreleased
 ----------
 
+Bug fixes
+~~~~~~~~~
+
+* Fix bug where indexing with a scalar numpy value returned a single-value array.
+  By :user:`Ben Jeffery <benjeffery>` :issue:`967`.
+
+
 .. _release_2.11.0:
 
 2.11.0
--- a/zarr/indexing.py
+++ b/zarr/indexing.py
@@ -34,7 +34,10 @@ def is_integer_list(x):
 
 
 def is_integer_array(x, ndim=None):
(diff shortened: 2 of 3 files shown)


joshmoore:
(Thanks, @benjeffery. Triggered the tests)

benjeffery:
Sorry I missed the linting in the dev guide. I can recommend https://pre-commit.com/ for these type of checks, we use it heavily over at tskit: https://github.com/tskit-dev/tskit/blob/main/.pre-commit-config.yaml

joshmoore:
@benjeffery : in general, big :+1: for pre-commit from my side (and maybe a :-1: for using the GH editor 🙂 )

joshmoore:
Proposing to get this out ASAP as 2.11.1

PR #976, opened 2022-03-01
remove `clobber` from `normalize_store_arg` to enable writes to consolidated FSStore groups

### LINKED ISSUE #975: Open mode is lost when opening via FSStore
This issue describes failed writes to xarray groups using `FSStore` where I'd expect the write to succeed. My current assumption is, that these issues occure due to mode-flags getting lost when the store is created implicitly (i.e. only an url is passed to zarr functions). I don't yet know how to fix this or if this is intended behaviour, thus I'm creating an issue.

```python
import zarr
import os

here = os.path.abspath(os.path.dirname(__file__))
url = "file://" + here + "/bug.zarr"  # intentionally using fsspec, problem occurs when using any FSStore backend

root = zarr.group(url, overwrite=True)
```

#### Problem description

The code above raises:
```
  File "/Users/tobi/Documents/code/zarrbug/test_write.py", line 7, in <module>
    root = zarr.group(url, overwrite=True)
  File "/usr/local/lib/python3.9/site-packages/zarr/hierarchy.py", line 1112, in group
    init_group(store, overwrite=overwrite, chunk_store=chunk_store,
  File "/usr/local/lib/python3.9/site-packages/zarr/storage.py", line 490, in init_group
    _init_group_metadata(store=store, overwrite=overwrite, path=path,
  File "/usr/local/lib/python3.9/site-packages/zarr/storage.py", line 504, in _init_group_metadata
    rmdir(store, path)
  File "/usr/local/lib/python3.9/site-packages/zarr/storage.py", line 138, in rmdir
    store.rmdir(path)  # type: ignore
  File "/usr/local/lib/python3.9/site-packages/zarr/storage.py", line 1259, in rmdir
    raise ReadOnlyError()
zarr.errors.ReadOnlyError: object is read-only
```

I'd expect that the code works (in particular due to `overwrite=True`)

#### (likely) related variant

```python3
import zarr    
import os    
    
here = os.path.abspath(os.path.dirname(__file__))    
url = "file://" + here + "/bug.zarr"    
store = zarr.storage.FSStore(url)    
    
root = zarr.group(store, overwrite=True)    
root.zeros('baz', shape=(10000, 10000), chunks=(1000, 1000), dtype='i4')    
zarr.consolidate_metadata(store)    
    
    
root = zarr.open_consolidated(url, mode="r+")    
root["baz"][0,0] = 7 
```

raises:

```
  File "/Users/tobi/Documents/code/zarrbug/test.py", line 14, in <module>
    root["baz"][0,0] = 7
  File "/usr/local/lib/python3.9/site-packages/zarr/core.py", line 1285, in __setitem__
    self.set_basic_selection(pure_selection, value, fields=fields)
  File "/usr/local/lib/python3.9/site-packages/zarr/core.py", line 1380, in set_basic_selection
    return self._set_basic_selection_nd(selection, value, fields=fields)
  File "/usr/local/lib/python3.9/site-packages/zarr/core.py", line 1680, in _set_basic_selection_nd
    self._set_selection(indexer, value, fields=fields)
  File "/usr/local/lib/python3.9/site-packages/zarr/core.py", line 1732, in _set_selection
    self._chunk_setitem(chunk_coords, chunk_selection, chunk_value, fields=fields)
  File "/usr/local/lib/python3.9/site-packages/zarr/core.py", line 1994, in _chunk_setitem
    self._chunk_setitem_nosync(chunk_coords, chunk_selection, value,
  File "/usr/local/lib/python3.9/site-packages/zarr/core.py", line 2005, in _chunk_setitem_nosync
    self.chunk_store[ckey] = self._encode_chunk(cdata)
  File "/usr/local/lib/python3.9/site-packages/zarr/storage.py", line 1177, in __setitem__
    raise ReadOnlyError()
zarr.errors.ReadOnlyError: object is read-only
```

I've explicitly instantiated the store when creating the dataset (which circuments the previous error).
But when trying to use the url directly for updating the dataset, I recieve a similar error (note that this error is raised on the line `root["baz"][0,0] = 7`).

#### context

The following xarray-based code raises the same (second) exception:

```python3
import numpy as np
import xarray as xr
import os

here = os.path.abspath(os.path.dirname(__file__))
url = "file://" + here + "/bug.zarr"

ds = xr.Dataset({"baz": (("x", "y"), np.zeros((10000, 10000), dtype="i4"))}).chunk({"x": 1000, "y": 1000})
ds.to_zarr(url, compute=False)

ds.isel(x=slice(1000), y=slice(1000)).to_zarr(url, mode="r+", region={"x": slice(1000), "y": slice(1000)})
```

Thanks @observingClouds wo showed me this example.

* Value of ``zarr.__version__``: '2.11.0'
* Value of ``numcodecs.__version__``: '0.7.3'
* Python 3.9.10
* Operating system: Linux
* zarr was installed using pip globally

d70-t:
This PR removes the `clobber` argument from `normalize_store_arg` and fixes #975 .

~TODO:~

### DIFF
--- a/zarr/tests/test_storage.py
+++ b/zarr/tests/test_storage.py
@@ -1021,7 +1021,8 @@ class TestFSStore(StoreTests):
         with pytest.raises(PermissionError):
             g.data[:] = 1
 
-    def test_modify_consolidated(self):
+    @pytest.mark.parametrize('mode,allowed', [("r", False), ("r+", True)])
+    def test_modify_consolidated(self, mode, allowed):
         import zarr
         url = "file://" + tempfile.mkdtemp()
 
@@ -1031,11 +1032,15 @@ class TestFSStore(StoreTests):
         zarr.consolidate_metadata(url)
 
         # reopen and modify
-        root = zarr.open_consolidated(url, mode="r+")
-        root["baz"][0, 0] = 7
+        root = zarr.open_consolidated(url, mode=mode)
+        if allowed:
+            root["baz"][0, 0] = 7
 
-        root = zarr.open_consolidated(url, mode="r")
-        assert root["baz"][0, 0] == 7
+            root = zarr.open_consolidated(url, mode="r")
+            assert root["baz"][0, 0] == 7
+        else:
+            with pytest.raises(zarr.errors.ReadOnlyError):
+                root["baz"][0, 0] = 7
 
     @pytest.mark.parametrize('mode', ["r", "r+"])
     def test_modify_consolidated_metadata_raises(self, mode):
--- a/zarr/tests/test_storage.py
+++ b/zarr/tests/test_storage.py
@@ -1021,6 +1021,19 @@ class TestFSStore(StoreTests):
         with pytest.raises(PermissionError):
             g.data[:] = 1
 
+    def test_modify_consolidated(self):
+        import zarr
+        url = "file://" + tempfile.mkdtemp()
+
+        # create
+        root = zarr.open_group(url, mode="w")
+        root.zeros('baz', shape=(10000, 10000), chunks=(1000, 1000), dtype='i4')
+        zarr.consolidate_metadata(url)
+
+        # reopen and modify
+        root = zarr.open_consolidated(url, mode="r+")
+        root["baz"][0, 0] = 7
+
     def test_read_only(self):
         path = tempfile.mkdtemp()
         atexit.register(atexit_rmtree, path)
--- a/zarr/tests/test_storage.py
+++ b/zarr/tests/test_storage.py
@@ -1037,6 +1037,21 @@ class TestFSStore(StoreTests):
         root = zarr.open_consolidated(url, mode="r")
         assert root["baz"][0, 0] == 7
 
+    @pytest.mark.parametrize('mode', ["r", "r+"])
+    def test_modify_consolidated_metadata_raises(self, mode):
+        import zarr
+        url = "file://" + tempfile.mkdtemp()
+
+        # create
+        root = zarr.open_group(url, mode="w")
+        root.zeros('baz', shape=(10000, 10000), chunks=(1000, 1000), dtype='i4')
+        zarr.consolidate_metadata(url)
+
+        # reopen and modify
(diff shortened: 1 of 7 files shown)


joshmoore:
@martindurant: would you be up for taking a closer look here?

zarr/tests/test_storage.py
```diff
@@ -1021,6 +1021,19 @@ def test_create(self):
...
+        url = "file://" + tempfile.mkdtemp()
+
+        # create
+        root = zarr.open_group(url, mode="w")
+        root.zeros('baz', shape=(10000, 10000), chunks=(1000, 1000), dtype='i4')
+        zarr.consolidate_metadata(url)
+
+        # reopen and modify
+        root = zarr.open_consolidated(url, mode="r+")
+        root["baz"][0, 0] = 7
```
martindurant, line 1035: Should verify that this did indeed change the data.
Also would add a case for each mode which attempts to change the metadata - should succeed for mode "w" and fail for "r+".

martindurant (commented):
As far as I can tell, things appear to be in order, but I ask to slightly extend the test to make sure.

zarr/tests/test_storage.py
d70-t: `open_consolidated` may only be called with modes ("r" and "r+"). I've added test for the exeption on those and added a test on "w-" for `open_consolidated.

If you prefer, I can also add the "a", "w" and "w-" tests here, but they won't accept changes to the metadata, but will fail on opening (which is likely already covered by the tests in convenience).

martindurant: > I can also add the "a", "w" and "w-" tests here

Not necessary.

Maybe a test showing an exception when you open with "r" and try to write data? I know this was always an exception. If you think it's already covered elsewhere, I'm happy to take your word for it.

martindurant (approved):

zarr/tests/test_storage.py
martindurant: +1

martindurant:
If no one else has any comments, I'll merge this by tomorrow.

PR #988, opened 2022-03-17
Bump minimum NumPy version to 1.20

madsbk:
To make use of [NEP-35](https://numpy.org/neps/nep-0035-array-creation-dispatch-with-array-function.html), features such as https://github.com/zarr-developers/zarr-python/pull/934 need NumPy version 1.20.

TODO:


jakirkham:
@rabernat would be good to get your thoughts on this 🙂 What NumPy version is Xarray using?

Also took a look at [NEP 29]( https://numpy.org/neps/nep-0029-deprecation_policy.html ) as another reference point. It appears they recommend moving to NumPy 1.20+ by June 21 (so in 3 months).

jakirkham:
Would it be ok for Xarray if Zarr bumped its NumPy requirement? Or would that present issues?

Do we have a sense of how many users are on NumPy 1.18 or 1.19? Or NumPy 1.17 for that matter (Zarr's current minimum)?

quasiben:
I think it's possible that work in https://github.com/zarr-developers/zarr-python/pull/934 _could_ be done for older versions of NumPy however, if NumPy 1.20 is going to be adopted in June I don't think there is much value for the effort in supporting older NumPy

joshmoore:
So it sounds like we need to get the order of a few things right:

* Merging this PR
* Releasing Zarr 2.12
* June 21st

jakirkham:
We put [a quick poll on Twitter]( https://twitter.com/zarr_dev/status/1521755830051389441 ) yesterday and shared that in a few places. To summarize we had 58 responses. The majority (60%) were using NumPy 1.20 or newer. NumPy 1.19 still had a fair number of users (22%). Versions prior had users, but less than 10% in each case.

While NumPy 1.20+, is used by a majority of users it is not as large as one might like when making a decision around dropping versions. Setting a minimum of NumPy 1.19 is a bit more tenable in the near term given that covers more than 80% of users. It is also in line with the NEP and [the current conda-forge minimum]( https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/3cb25a68cc5fbb2527674c28b584aa526abcf929/recipe/migrations/pypy38.yaml#L39-L46 ).

Would be curious to know how much it would help us to bump to NumPy 1.19 in the near term? Admittedly there may still be workarounds, but perhaps we could have fewer of them. It may also make it easier to bump to NumPy 1.20 in the future.

Also interested in hearing others thoughts on all of this 🙂

jakirkham:
@joshmoore @rabernat, thoughts on the comment above or this PR generally?

joshmoore:
Guess my biggest 2c is that the June date is quite close. I'd be happy though to commit to getting a minor version out with the older numpy and then drop in the next minor version so as to not prevent anyone from making use of 2.12.

madsbk:
Are we getting ready to require NumPy v1.20+ ?

joshmoore:
Just doing the numcodecs 0.10.0 release now. I'm inclined to get 2.12 out _without_ this PR, and then start prepping a quick 2.13 with this PR as well as @d-v-b's Python 3.8+ typing usage. (Happy to be told I'm being overly cautious though)

jakirkham:
Doing a release beforehand makes sense. Thanks for getting that started Josh 🙏

madsbk:
@joshmoore now that `2.12` is released, I think we should merge this. We don't necessarily have to do a new release, having it in `main` should be enough for me to continue my work on https://github.com/zarr-developers/zarr-python/pull/934

jakirkham:
Thanks all! 🙏

PR #1001, opened 2022-04-01
Revert change to default write_empty_chunks.

### LINKED ISSUE #965: Incorrect default fill value causes byte arrays to become numeric when `write_empty_chunks=False`

vyasr:
This PR resolves #965 by changing the default of `write_empty_chunks` to `True`. I am not sure if we want to document the current limitations of `write_empty_chunks=False`, or if we want to treat those as bugs that will be fixed eventually. In the future we may want to error if a user sets `write_empty_chunks=True` with an unsupported data type, but that change should probably be made in a future PR that implements the solution described in https://github.com/zarr-developers/zarr-python/issues/965#issuecomment-1058723797 and https://github.com/zarr-developers/zarr-python/issues/965#issuecomment-1058725944

TODO:

### DIFF
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -6,6 +6,14 @@ Release notes
 Unreleased
 ----------
 
+Bug fixes
+~~~~~~~~~
+
+* Changes the default value of ``write_empty_chunks`` to ``True`` to prevent
+  unanticipated data losses when the data types do not have a proper default
+  value when empty chunks are read back in.
+  By :user:`Vyas Ramasubramani <vyasr>`; :issue:`965`.
+
 .. _release_2.11.1:
 
 2.11.1
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -1309,7 +1309,7 @@ Empty chunks
  
 As of version 2.11, it is possible to configure how Zarr handles the storage of
 chunks that are "empty" (i.e., every element in the chunk is equal to the array's fill value).
-When creating an array with ``write_empty_chunks=False`` (the default), 
+When creating an array with ``write_empty_chunks=False``, 
 Zarr will check whether a chunk is empty before compression and storage. If a chunk is empty,
 then Zarr does not store it, and instead deletes the chunk from storage 
 if the chunk had been previously stored. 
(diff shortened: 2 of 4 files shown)


jni (approved):

jeromekelleher:
Great to see this,  thanks @vyasr! cc @benjeffery @tomwhite

joshmoore:
I assume the general consensus is to get this out as an immediate 2.11.2?

jni:
That's my understanding!

joshmoore (approved):
Thanks, @vyasr!

Merging and I'll start preparing a `v2` branch for releasing from. (As a side note: slightly surprised that none of the tests needed updating. I assume that means that none are testing the default value, eh?)

jakirkham:
Thanks all! 😄

joshmoore:
2.11.2 is released from the `2_11` branch with this PR backported.

jakirkham:
Should we create a GitHub release for that tag as well?

joshmoore:
Ah, apologies. I forgot that step. Yes, I'll do it now.

jakirkham:
Not at all. Thanks for handling this Josh 🙏

tomwhite:
Thanks for releasing 2.11.2 @joshmoore. Unfortunately, it looks like it doesn't fully revert the change:

```python
>>> import zarr
>>> zarr.__version__
'2.11.2'
>>> a = zarr.create((100, 100), chunks=(100, 50), dtype="i4", store="example.zarr")
>>> a[:] = 0
>>> import os
>>> os.listdir("example.zarr")
['.zarray']
```

Whereas with a previous version of Zarr (and the main branch):

```python
>>> import zarr
>>> zarr.__version__
'2.10.3'
>>> a = zarr.create((100, 100), chunks=(100, 50), dtype="i4", store="example.zarr")
>>> a[:] = 0
>>> import os
>>> os.listdir("example.zarr")
['.zarray', '0.0', '0.1']
```

The problem is in `zarr.creation.create` where the default is still `write_empty_chunks=False`:

https://github.com/zarr-developers/zarr-python/blob/04519bdfa30edf0fcfdc6a17bf5880b7f91eb208/zarr/creation.py#L23

jakirkham:
@tomwhite would you be willing to send a PR? 🙂

tomwhite:
Opened #1005

joshmoore:
@vyasr : not at all. [My fault.](https://github.com/zarr-developers/zarr-python/pull/1005#issuecomment-1091419391)

joshmoore:
@d-v-b, if you are still planning to add an `auto` setting based on dtype, there are perhaps some other interesting comments regarding defaults in #8.

PR #1006, opened 2022-04-11
misc Zarr V3 bug fixes: open_group, open_consolidated and MemoryStoreV3

grlee77:
The first commit here fixes a couple of bugs in the v3 convenience functions that were revealed when working on a branch of Xarray that adds tests with Zarr v3.

The `rename` method on MemoryStoreV3 was also fixed here (GroupV3 tests with this store type had inadvertently been left commented out in the previously merged PR).

Some outdated commments/TODOs were also cleaned up and comments were added explaining some cases where desired v3 behavior is not fully clear (e.g. `test_path`).

TODO:

### DIFF
--- a/zarr/convenience.py
+++ b/zarr/convenience.py
@@ -1277,7 +1277,9 @@ def open_consolidated(store: StoreLike, metadata_key=".zmetadata", mode="r+", **
     """
 
     # normalize parameters
-    store = normalize_store_arg(store, storage_options=kwargs.get("storage_options"), mode=mode)
+    zarr_version = kwargs.get('zarr_version', None)
+    store = normalize_store_arg(store, storage_options=kwargs.get("storage_options"), mode=mode,
+                                zarr_version=zarr_version)
     if mode not in {'r', 'r+'}:
         raise ValueError("invalid mode, expected either 'r' or 'r+'; found {!r}"
                          .format(mode))
--- a/zarr/tests/test_convenience.py
+++ b/zarr/tests/test_convenience.py
@@ -204,19 +204,28 @@ def test_tree(zarr_version):
 
 @pytest.mark.parametrize('zarr_version', [2, 3])
 @pytest.mark.parametrize('with_chunk_store', [False, True], ids=['default', 'with_chunk_store'])
-def test_consolidate_metadata(with_chunk_store, zarr_version):
-
-    if zarr_version == 2:
-        MemoryStoreClass = MemoryStore
-        path = ''
-    else:
-        MemoryStoreClass = MemoryStoreV3
-        path = 'dataset'
-
+@pytest.mark.parametrize('stores_from_path', [False, True])
+def test_consolidate_metadata(with_chunk_store, zarr_version, stores_from_path):
     # setup initial data
-    store = MemoryStoreClass()
-    chunk_store = MemoryStoreClass() if with_chunk_store else None
-    z = group(store, chunk_store=chunk_store, path=path)
+    if stores_from_path:
+        store = tempfile.mkdtemp()
+        atexit.register(atexit_rmtree, store)
+        if with_chunk_store:
+            chunk_store = tempfile.mkdtemp()
+            atexit.register(atexit_rmtree, chunk_store)
+        else:
+            chunk_store = None
+        version_kwarg = {'zarr_version': zarr_version}
+    else:
+        if zarr_version == 2:
+            store = MemoryStore()
+            chunk_store = MemoryStore() if with_chunk_store else None
+        elif zarr_version == 3:
+            store = MemoryStoreV3()
+            chunk_store = MemoryStoreV3() if with_chunk_store else None
+        version_kwarg = {}
+    path = 'dataset' if zarr_version == 3 else None
+    z = group(store, chunk_store=chunk_store, path=path, **version_kwarg)
     z.create_group('g1')
     g2 = z.create_group('g2')
     g2.attrs['hello'] = 'world'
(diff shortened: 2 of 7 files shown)


joshmoore:
Any thoughts on:
Screen Shot 2022-04-12 at 12 57 12

joshmoore:
Green! :smile:

zarr/convenience.py
```diff
@@ -1277,7 +1277,9 @@ def open_consolidated(store: StoreLike, metadata_key=".zmetadata", mode="r+", **
     """
 
     # normalize parameters
-    store = normalize_store_arg(store, storage_options=kwargs.get("storage_options"), mode=mode)
+    zarr_version = kwargs.get('zarr_version', None)
```
joshmoore, line 1280: Perhaps more for a follow-up, but if this is getting extracted does it make sense to list as a parameter?

joshmoore: I notice that's not happening with `storage_options` either, so it probably is more of a higher-level aesthetic decision.

zarr/tests/test_convenience.py
```diff
@@ -202,23 +202,30 @@ def test_tree(zarr_version):
...
-
-    if zarr_version == 2:
-        MemoryStoreClass = MemoryStore
-        path = ''
-    else:
-        MemoryStoreClass = MemoryStoreV3
-        path = 'dataset'
-
+@pytest.mark.parametrize('stores_from_path', [False, True])
+def test_consolidate_metadata(with_chunk_store, zarr_version, stores_from_path):
```
joshmoore, line 208: Outwith this PR, but this reminds me if v3 would make standardizing consolidated metadata easier. See https://github.com/zarr-developers/zarr-specs/issues/113

zarr/tests/test_convenience.py
```diff
@@ -229,64 +236,81 @@ def test_consolidate_metadata(with_chunk_store, zarr_version):
     arr[:] = 1.0
     assert 16 == arr.nchunks_initialized
 
+    if stores_from_path:
```
joshmoore, line 239: Seems fine to have in the test, but this is an interesting block, @grlee77. Can you explain it?

zarr/tests/test_convenience.py
```diff
@@ -229,64 +236,81 @@ def test_consolidate_metadata(with_chunk_store, zarr_version):
...
+            assert '.zmetadata' in store
+            meta_keys = ['.zgroup',
+                         'g1/.zgroup',
+                         'g2/.zgroup',
+                         'g2/.zattrs',
+                         'g2/arr/.zarray',
+                         'g2/arr/.zattrs']
+        else:
+            assert isinstance(out._store, ConsolidatedMetadataStoreV3)
+            assert 'meta/root/consolidated/.zmetadata' in store
```
joshmoore, line 269: Again, outwith this PR but back to https://github.com/zarr-developers/zarr-specs/issues/113, could the v3 spec here simply replace all the individual JSON files _solely_ with the consolidated metadata?

zarr/tests/test_hierarchy.py
```diff
@@ -970,22 +989,44 @@ def test_paths(self):
...
+            assert g2 == g1['foo/bar/']
+            assert g2 == g1['//foo/bar']
+            assert g2 == g1['//foo//bar//']
+            assert g2 == g1['///foo///bar///']
+            assert g2 == g2['/foo/bar']
+        else:
+            # the expected key format gives a match
+            assert g2 == g1['foo/bar']
+
+            # TODO: Should presence of a trailing slash raise KeyError?
```
joshmoore, line 1010: This should likely get captured as an issue for the V3 final roadmap.

zarr/tests/test_hierarchy.py
```diff
@@ -888,12 +892,27 @@ def test_move(self):
...
+            if g._chunk_store:
+                assert 'data/root/bar/c0' in g._chunk_store
+            else:
+                assert 'data/root/bar/c0' in g._store
         assert isinstance(g["foo2"], Group)
         if g2._version == 2:
-            # TODO: how to access element created outside of group.path in v3?
             assert_array_equal(data, g["bar"])
+        else:
+            # TODO: How to access element created outside of group.path in v3?
```
joshmoore, line 910: This should likely get captured as an issue for the V3 final roadmap.

joshmoore (approved):
Looks good, @grlee77. Thanks! I'm merging so that we have this to validate other PRs but I've kicked off a few conversations that are likely of interest as V3 gets finalized.

zarr/convenience.py
grlee77: I am fine doing it either way.

zarr/tests/test_convenience.py
grlee77: also /zarr-developers/zarr-specs#136 related to where this metadata would be stored

zarr/tests/test_convenience.py
grlee77: I think it could be done, but haven't tried it. All store types are already somewhat consolidated for v3 since the entire `meta` folder tree is independent of the `data` one. I suppose if there are many arrays/groups it may still be preferred to further consolidate into one file, though? I would defer to those using the feature on what is the desired behavior.

zarr/tests/test_convenience.py
grlee77: There are tests below that do `del store[key]`, but when the store is just a path string instead of a `BaseStore` class this will not work. So, the test uses `z._store` directly. I think I had originally just tried deleting the keys directly from the consolidated store, `z`, instead of accessing the private `z._store`, but this was not possible because `ConsolidatedMetadataStore` classes are read-only.

Issue #938, opened 2022-01-14
When I repeatedly append data to a zarr array in s3, appending takes longer and longer and longer


christine-e-smit:
#### Minimal, reproducible code sample, a copy-pastable example if possible

```python
#!/usr/bin/env python3

###
#  Please run me in AWS! The command runs as:
# 
#    python3 loop_zarr.py  $YOUR_S3_BUCKET_HERE $YOUR_PREFIX_HERE > loop_zarr.txt
#
# This will take a while to run. The timing issue only becomes really clear after the code runs for a few hours.
###

import sys
import s3fs
import argparse
import datetime as dt
from textwrap import dedent
import zarr
import numpy as np


def main(argv):
    description = dedent(
        """\
        This script replicates the issues we see with repeatedly appending to an array
        in s3.
        """
    )
    parser = argparse.ArgumentParser(
        description=description, formatter_class=argparse.RawTextHelpFormatter
    )
    parser.add_argument("s3_bucket", help="s3 bucket for data")
    parser.add_argument("s3_path", help="path prefix in the bucket for the store")

    args = parser.parse_args(argv)

    # number of 8-byte floats in a chunk
    chunk_size = 2500
    # number of chunks to write each iteration before writing the timing
    num_chunks = 500
    # number of iterations for the loop
    num_iterations = 500

    # initialize the zarr store
    time_str = dt.datetime.now().strftime("%Y-%m-%d__%H-%M-%S.%f__")
    path = f"{args.s3_bucket}/{args.s3_path}{time_str}"
    print(f"Writing zarr data under {path}", flush=True)
    var_name = "var"
    print(f"Chunk size (in floats): {chunk_size}", flush=True)
    print(f"Number of chunks written per iteration: {num_chunks}", flush=True)
    print(f"Number of iterations: {num_iterations}", flush=True)
    # note: 8-byte floats
    total_bytes = chunk_size * 8 * num_chunks * num_iterations
    print(f"Total number of gigabytes to be written: {total_bytes/1e9}", flush=True)

    create_dataset(path, var_name, chunk_size)
    print("initialized zarr", flush=True)

    # Create some data to write out. We don't actually care what it is.
    chunk = np.random.normal(size=chunk_size)

    # Iterate through
    for i in range(num_iterations):
        start_iteration = dt.datetime.now()
        for j in range(num_chunks):
            append_data(path, var_name, chunk)
        iteration_length = (dt.datetime.now() - start_iteration).total_seconds()
[...]

PR #1014, opened 2022-04-25
Appending performance improvement

hailiangzhang:
This PR is trying to address the issue reported in #938:
```
When I repeatedly append data to a zarr array in s3, appending takes longer and longer and longer
```
The issue appears to be coming from [a method in zarr/core.py](https://github.com/zarr-developers/zarr-python/blob/e381bf76261a1cc9c10613465ab39b968a24737b/zarr/core.py#L2414) when trying to [remove any chunks not within range](https://github.com/zarr-developers/zarr-python/blob/e381bf76261a1cc9c10613465ab39b968a24737b/zarr/core.py#L2430).

Basically the existing implementation iterates through **all** the `old` chunks and removes those that don't exist in the `new` chunks. This seems time-consuming and unnecessary when appending new chunks to the end of an existing array with a large number of chunks.

Regarding this, this PR will iterate through each dimension, and **only** find and remove the chunk slices that exist in `old` but not `new` data. It also introduced a mutable list to dynamically adjust the number of chunks along the already-processed dimensions in order to avoid duplicate chunk removal. These details have been documented in [the comments from my changes](https://github.com/zarr-developers/zarr-python/compare/master...hailiangzhang:appending_performance_improvement#diff-f640a0c570765203632323d71a9d1145ec6121c90db73f56e1271ab0efd5b8d8R2431).

This should noticeably improve the performance when appending data to a huge zarr array; for our benchmark, it reduced the processing time by half. Moreover, the time of appending should stay the same as the array size grows.

TODO:

### DIFF
--- a/zarr/core.py
+++ b/zarr/core.py
@@ -2448,18 +2448,6 @@ class Array:
                     # chunk not initialized
                     pass
             old_cdata_shape_working_list[idx_cdata] = min(val_old_cdata, val_new_cdata)
-        '''
-        for cidx in itertools.product(*[range(n) for n in old_cdata_shape]):
-            if all(i < c for i, c in zip(cidx, new_cdata_shape)):
-                pass  # keep the chunk
-            else:
-                key = self._chunk_key(cidx)
-                try:
-                    del chunk_store[key]
-                except KeyError:
-                    # chunk not initialized
-                    pass
-        '''
 
     def append(self, data, axis=0):
         """Append `data` to `axis`.
--- a/zarr/core.py
+++ b/zarr/core.py
@@ -2429,6 +2429,26 @@ class Array:
 
         # remove any chunks not within range
         chunk_store = self.chunk_store
+        old_cdata_shape_working_list = list(old_cdata_shape)
+        for idx_cdata, (val_old_cdata, val_new_cdata) in enumerate(
+            zip(old_cdata_shape_working_list, new_cdata_shape)
+        ):
+            for cidx in itertools.product(
+                *[
+                    range(n_new, n_old) if (idx == idx_cdata) else range(n_old)
+                    for idx, (n_old, n_new) in enumerate(
+                        zip(old_cdata_shape_working_list, new_cdata_shape)
+                    )
+                ]
+            ):
+                key = self._chunk_key(cidx)
+                try:
+                    del chunk_store[key]
+                except KeyError:
+                    # chunk not initialized
+                    pass
+            old_cdata_shape_working_list[idx_cdata] = min(val_old_cdata, val_new_cdata)
+        '''
         for cidx in itertools.product(*[range(n) for n in old_cdata_shape]):
             if all(i < c for i, c in zip(cidx, new_cdata_shape)):
                 pass  # keep the chunk
@@ -2439,6 +2459,7 @@ class Array:
                 except KeyError:
                     # chunk not initialized
                     pass
+        '''
 
     def append(self, data, axis=0):
         """Append `data` to `axis`.
--- a/zarr/core.py
+++ b/zarr/core.py
@@ -2432,7 +2432,7 @@ class Array:
         #     only find and remove the chunk slices that exist in 'old' but not 'new' data.
         #   Note that a mutable list ('old_cdata_shape_working_list') is introduced here
         #     to dynamically adjust the number of chunks along the already-processed dimensions
-        #     in order to avoid duplicate chunk removal. 
+        #     in order to avoid duplicate chunk removal.
         chunk_store = self.chunk_store
         old_cdata_shape_working_list = list(old_cdata_shape)
         for idx_cdata, (val_old_cdata, val_new_cdata) in enumerate(
--- a/zarr/core.py
+++ b/zarr/core.py
@@ -2428,6 +2428,8 @@ class Array:
                                 for s, c in zip(new_shape, chunks))
 
         # remove any chunks not within range
+        #   The idea is that, along each dimension,
+        #   find and remove the chunk slices that only exist in 'old_cdata_shape'.
         chunk_store = self.chunk_store
         old_cdata_shape_working_list = list(old_cdata_shape)
         for idx_cdata, (val_old_cdata, val_new_cdata) in enumerate(
--- a/zarr/core.py
+++ b/zarr/core.py
@@ -2429,7 +2429,10 @@ class Array:
 
         # remove any chunks not within range
         #   The idea is that, along each dimension,
-        #   find and remove the chunk slices that only exist in 'old_cdata_shape'.
+        #     only find and remove the chunk slices that exist in 'old' but not 'new' data.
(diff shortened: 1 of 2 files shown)


joshmoore:
Thanks, @hailiangzhang! I've launched the tests.

joshmoore:
Relaunched after update, thanks, @hailiangzhang.

Reading the comment and the implementation this makes sense & all tests are passing. I do wonder, though, if you have any ideas for edge cases which may need extra tests.

Also: is it possible to share the benchmark you are evaluating with?

rabernat:
I think there is another potential opportunity for optimization here

https://github.com/zarr-developers/zarr-python/blob/539755fdeff5e1b9d6f9d6cb00c454ad805ba3c2/zarr/core.py#L2449-L2451

We are deleting chunks synchronously, in a loop. It would be better to first populate the list of keys and then call `delitems` on them. This would potentially dispatch asynchronously leading to much faster delete speed (especially with object storage). 

Only FSStore implements delitems.

https://github.com/zarr-developers/zarr-python/blob/27cf31555975eaeb34e802bfdd59c11222f91597/zarr/storage.py#L1372

hailiangzhang:
> Relaunched after update, thanks, @hailiangzhang.
> 
> Reading the comment and the implementation this makes sense & all tests are passing. I do wonder, though, if you have any ideas for edge cases which may need extra tests.
> 
> Also: is it possible to share the benchmark you are evaluating with?

Thanks for your comment, @joshmoore !

The benchmark that I used is based on the script that [Christine has posted in her issue report](https://github.com/zarr-developers/zarr-python/issues/938), where I found performance degradation when the number of chunks grew to 113,000, and observed 50% improvement after applying changes from this PR.
Regarding `edge cases which may need extra tests`, actually I think, _in theory_, my implementation should be identical the original code, so I am not sure whether new tests will be necessary to cover the changes or not.

However, I did look into the [related tests](https://github.com/zarr-developers/zarr-python/blob/27cf31555975eaeb34e802bfdd59c11222f91597/zarr/tests/test_core.py#L631), where it seems not covering the case when resizing the array by increasing one dimension while decreasing the other dimension.

`    def test_resize_2d(self):

        z = self.create_array(shape=(105, 105), chunks=(10, 10), dtype='i4',
                              fill_value=0)
        a = np.arange(105*105, dtype='i4').reshape((105, 105))
        z[:] = a
        assert (105, 105) == z.shape
        assert (105, 105) == z[:].shape
        assert np.dtype('i4') == z.dtype
        assert np.dtype('i4') == z[:].dtype
        assert (10, 10) == z.chunks
        assert_array_equal(a, z[:])

        z.resize((205, 205))
        assert (205, 205) == z.shape
        assert (205, 205) == z[:].shape
        assert np.dtype('i4') == z.dtype
        assert np.dtype('i4') == z[:].dtype
        assert (10, 10) == z.chunks
        assert_array_equal(a, z[:105, :105])
        assert_array_equal(np.zeros((100, 205), dtype='i4'), z[105:, :])
        assert_array_equal(np.zeros((205, 100), dtype='i4'), z[:, 105:])

        z.resize((55, 55))
        assert (55, 55) == z.shape
        assert (55, 55) == z[:].shape
        assert np.dtype('i4') == z.dtype
        assert np.dtype('i4') == z[:].dtype
        assert (10, 10) == z.chunks
        assert_array_equal(a[:55, :55], z[:])

        z.resize((55, 1))
        assert (55, 1) == z.shape
        assert (55, 1) == z[:].shape
        assert np.dtype('i4') == z.dtype
        assert np.dtype('i4') == z[:].dtype
        assert (10, 10) == z.chunks
        assert_array_equal(a[:55, :1], z[:])`

Maybe I can add a test case of `z.resize((1, 55))` at the end? Again, I don't think it is related to this PR, but please let me know if you think it's necessary and I will be happy to do that:)

joshmoore:
@hailiangzhang, there's a little time before the next release. If you have a chance to add the extra test(s), that'd be wonderful!

hailiangzhang:
> I think there is another potential opportunity for optimization here
> 
> https://github.com/zarr-developers/zarr-python/blob/539755fdeff5e1b9d6f9d6cb00c454ad805ba3c2/zarr/core.py#L2449-L2451
> 
> We are deleting chunks synchronously, in a loop. It would be better to first populate the list of keys and then call `delitems` on them. This would potentially dispatch asynchronously leading to much faster delete speed (especially with object storage).
> 
> Only FSStore implements delitems.
> 
> https://github.com/zarr-developers/zarr-python/blob/27cf31555975eaeb34e802bfdd59c11222f91597/zarr/storage.py#L1372

Thanks for your comment @rabernat !
Yes, I can first populate the list of keys and then call `delitems` on them at the end, but this still seems a synchronous operation, and I am not sure how this `would potentially dispatch asynchronously`. Sorry that I may have missed something here, but could you explain more on this?

hailiangzhang:
> @hailiangzhang, there's a little time before the next release. If you have a chance to add the extra test(s), that'd be wonderful!

@joshmoore , actually I found something interesting when trying to add [a new test as I suggested above (add a test case of z.resize((1, 55)) at the end)](https://github.com/zarr-developers/zarr-python/pull/1014#issuecomment-1115230278).

Basically I found that by using `resize`, if I reduce the shape of a zarr array to be smaller than the chunk size, and then increase the array shape afterward, the values that were supposed to be removed by the first `resize` was brought back by the second `resize`.

Sorry if I didn't explain it well, but I think this is a separate issue, and I [have just filed an issue report here](https://github.com/zarr-developers/zarr-python/issues/1017).

Again, since this issue is not related to this PR, I think we can probably add this test after this issue is addressed. I will be willing to look into this a later time and see whether it's easy to get it fixed.

hailiangzhang:
> > @hailiangzhang, there's a little time before the next release. If you have a chance to add the extra test(s), that'd be wonderful!
> 
> @joshmoore , actually I found something interesting when trying to add [a new test as I suggested above (add a test case of z.resize((1, 55)) at the end)](https://github.com/zarr-developers/zarr-python/pull/1014#issuecomment-1115230278).
> 
> Basically I found that by using `resize`, if I reduce the shape of a zarr array to be smaller than the chunk size, and then increase the array shape afterward, the values that were supposed to be removed by the first `resize` was brought back by the second `resize`.
> 
> Sorry if I didn't explain it well, but I think this is a separate issue, and I [have just filed an issue report here](https://github.com/zarr-developers/zarr-python/issues/1017).
> 
> Again, since this issue is not related to this PR, I think we can probably add this test after this issue is addressed. I will be willing to look into this a later time and see whether it's easy to get it fixed.

Hi @joshmoore , as mentioned in [my issue report (actually a feature request:)](https://github.com/zarr-developers/zarr-python/issues/1017), I have added an edge case test for resize method as we originally planned. This test can be adapted if my feature request is implemented in the future:)

joshmoore:
Thanks, @hailiangzhang. Merging to get this into a pre-release of 2.12

PR #1015, opened 2022-04-26
Pre-commit configuration

Alt-Shivam:
[Description of PR]
Flake8 is already added.
and the following will be added soon.

With few changes
https://github.com/pre-commit/mirrors-mypy
https://github.com/adrienverge/yamllint.git
https://github.com/asottile/pyupgrade
With moderate changes
https://github.com/pre-commit/pre-commit-hooks
https://github.com/asottile/seed-isort-config
https://github.com/PyCQA/isort
With lots of changes
https://github.com/psf/black

PR #1016, opened 2022-04-30
Added Yaml Check in Pre-commit

Alt-Shivam:
Yaml check for pre-commit.
Related to #1015 and #1012

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -11,3 +11,7 @@ repos:
       ]
       exclude: ^(venv/|docs/)
       types: ['python']
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v2.3.0
+    hooks:
+    - id: check-yaml


joshmoore:
Thanks, @Alt-Shivam.

```
Check Yaml...............................................................Passed
```

Based on your previous comment, isn't it a bit surprising that there are no errors?

Alt-Shivam:
Yes, @joshmoore Basically I used a different YAML checker in this case.
and this is slightly less strict and only checks the validity.

but in the previous case, the YAML linter is kind of strict one which checks line lengths, brackets, extra spaces and extra symbols in YAML

Alt-Shivam:
I have done some tests with this one.
image

Alt-Shivam:
I don't want to make lots of changes in the codebase, so I chose this one.
but If you like the strict one, Then I can also accommodate that one :)

joshmoore:
Ah, fair enough. Let's keep https://github.com/adrienverge/yamllint.git on the long-term list, but :+1: for starting with the easier introductions.

PR #1020, opened 2022-05-04
v3: use dev version number (draft)

joshmoore:
For discussion. Until v3 is finalized, using a "dev" marker for the v3 version string would prevent datasets that are written in a transitional state from becoming normative.

This follows on from the PRs by @grlee77 to add support for writing V3 (#898) and to hide those implementations unless the user explicitly sets an environment variable (#1007).

An unknown here is how many v3 filesets are already in the wild. xtensor-zarr (cc: @davidbrochart) and z5 (@constantinpape) have had writing support for some time.

If we decide to move forward with this change, we should synchronize implementations and make an effort to contact (known) users of these transitional v3 implementations.


joshmoore:
Closing this in favor of the environment variable protection, though I think this is a strategy we should consider the next time we are amidst a major overhaul.

cc: @jstriebel

PR #1021, opened 2022-05-04
Make all unignored zarr warnings errors

joshmoore:
Following on from #654, I wanted to make sure that no warnings are being missed by the tests. This widens those that are considered errors to everything coming from a zarr.* package. (We could consider including closely related packages like fsspec, numcodecs, etc.)

The only ignore statement that was needed, perhaps for review by @d-v-b, was the n5/compression statement.


jakirkham:
Basically just trying to catch these on CI? If so, seems like a reasonable thing to do

PR #1032, opened 2022-05-20
Mark V3 API experimental

joshmoore:
Following discussion at this week's community call, a few additional
precautions are being implemented here:

- [x] rename the var ZARR_V3_API_AVAILABLE to ZARR_V3_EXPERIMENTAL_API
- [x] add an assertion to prevent use of zarr_version=3 w/o the var

All tests should continue passing in GHA since the variable is set. Local runs, however, need some consideration in order to properly `xfail`.

See also #1020

### DIFF
--- a/.github/workflows/minimal.yml
+++ b/.github/workflows/minimal.yml
@@ -22,12 +22,16 @@ jobs:
           activate-environment: minimal
     - name: Tests
       shell: "bash -l {0}"
+      env:
+        ZARR_V3_EXPERIMENTAL_API: 1
       run: |
         conda activate minimal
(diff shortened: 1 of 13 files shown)


jakirkham:
Restarting failed CI jobs

MSanKeys963:
Thanks for leading this @joshmoore.
Are we also doing a `conda` release along with `pip` for a2?

joshmoore:
> Are we also doing a `conda` release along with `pip` for a2?

Deferring to @jakirkham on that one.

jakirkham:
We can

PR #1068, opened 2022-07-06
remove option to return None when the input is None from _ensure_store

grlee77:
As discussed in https://github.com/zarr-developers/zarr-python/pull/1051#issuecomment-1175539995, it does not seem logical that `_ensure_store` should return None (which is not a `BaseStore`). As far as I am aware, the ability to pass through None is not needed, so this PR removes it.

[Description of PR]

TODO:

### DIFF
--- a/zarr/tests/test_storage.py
+++ b/zarr/tests/test_storage.py
@@ -73,8 +73,6 @@ def test_ensure_store():
     with pytest.raises(ValueError):
         Store._ensure_store(KVStoreV3(dict()))
 
-    assert Store._ensure_store(None) is None
-
 
 def test_capabilities():
     s = KVStore(dict())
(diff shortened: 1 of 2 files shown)


zarr/tests/test_storage.py
```diff
@@ -73,8 +73,6 @@ class InvalidStore:
     with pytest.raises(ValueError):
         Store._ensure_store(KVStoreV3(dict()))
 
-    assert Store._ensure_store(None) is None
```
joshmoore, line 76:
```
    with pytest.raises(ValueError):
        Store._ensure_store(None)
```
❓

joshmoore (approved):
Minor one suggestion for adding a test, this seems to match implementation & documentation. :+1:

zarr/tests/test_storage.py
joshmoore: Pushed myself.

PR #1085, opened 2022-07-19
Zarr v3: support root path

### LINKED ISSUE #1039: Why is path required for opening Zarr v3 groups?
With Zarr v2, I can open a group by passing either a valid Zarr store or with a path specified as a string, i.e., like `zarr.open_group(store_or_path)`. As I understand it, paths get normalized into store objects, e.g., to a local filesystem or via fsspec.

With Zarr v3, as currently implemented, the `path` argument is apparently now required, per https://github.com/pydata/xarray/pull/6475. This feels like a small step backwards in terms of usability. I'm wondering if I'm missing some broader context here? Maybe some examples of how users would canonically create a group, add an array and then access the data in the new v3 API would be helpful.

grlee77:
closes #1039

v3 spec states path = '/' for arrays gives an array at /meta/root.array.json
               path = '/' for groups gives a group at /meta/root.array.json

In this implementation `path=None` will also result in a root array. Creation routines default to `path=None`, so this makes it so that the path argument does not have to be manually specified. 

TODO:

### DIFF
--- a/zarr/_storage/store.py
+++ b/zarr/_storage/store.py
@@ -420,11 +420,11 @@ def _listdir_from_keys(store: BaseStore, path: Optional[str] = None) -> List[str
 
 def _prefix_to_array_key(store: StoreLike, prefix: str) -> str:
     if getattr(store, "_store_version", 2) == 3:
+        sfx = _get_metadata_suffix(store)  # type: ignore
         if prefix:
-            sfx = _get_metadata_suffix(store)  # type: ignore
             key = meta_root + prefix.rstrip("/") + ".array" + sfx
         else:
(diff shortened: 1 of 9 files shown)


joshmoore:
@grlee77, thanks for this! Reading over it, it does seem to address the issues that Stephan raised. Do you have anything else for this PR from your side or just need testing and then rolling into a 2.12.x release??

grlee77:
I think test coverage is pretty good now. It should be ready for review.

joshmoore:
Rolling into a 2.13 pre-release for better testing together with xarray.

cc: @shoyer

PR #1089, opened 2022-07-19
use __reduce__ for pickling instead of __setstate__ / __getstate__

d-v-b:
In main, `Array` and `Group` have `__setstate__` methods that directly call `__init__`, which isn't great (mypy complains about this). We can avoid calling `__init__` by defining [`__reduce__`](https://docs.python.org/3/library/pickle.html#object.__reduce__) for these classes instead of `__setstate__` / `__getstate__`. 

Note that there are a lot of store classes that use `__setstate__` / `__getstate__` with direct calls of `__init__`. Clearing that up might be in scope for additional commits to this branch.

TODO:

### DIFF
--- a/zarr/core.py
+++ b/zarr/core.py
@@ -153,7 +153,7 @@ class Array:
 
     def __init__(
         self,
-        store: Any,  # BaseStore not stricly required due to normalize_store_arg
+        store: Any,  # BaseStore not strictly required due to normalize_store_arg
         path=None,
         read_only=False,
         chunk_store=None,
--- a/zarr/core.py
+++ b/zarr/core.py
@@ -2353,13 +2353,18 @@ class Array:
 
         return checksum
 
-    def __getstate__(self):
-        return (self._store, self._path, self._read_only, self._chunk_store,
-                self._synchronizer, self._cache_metadata, self._attrs.cache,
-                self._partial_decompress, self._write_empty_chunks, self._version)
-
-    def __setstate__(self, state):
-        self.__init__(*state)
+    def __reduce__(self):
+        args = (self.store,
+                self.path,
+                self.read_only,
+                self.chunk_store,
+                self.synchronizer,
+                self._cache_metadata,
+                self.attrs.cache,
+                self._partial_decompress,
+                self.write_empty_chunks,
+                self._version)
+        return (self.__class__, args)
 
     def _synchronized_op(self, f, *args, **kwargs):
 
--- a/.deepsource.toml
+++ b/.deepsource.toml
@@ -5,6 +5,3 @@ test_patterns = ["zarr/tests/test_*.py"]
 [[analyzers]]
 name = "python"
 enabled = true
-
-  [analyzers.meta]
-  runtime_version = "3.x.x"
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.7', '3.8', '3.9', '3.10']
+        python-version: ['3.8', '3.9', '3.10']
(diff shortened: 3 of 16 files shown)


zarr/core.py
```diff
@@ -2353,13 +2353,18 @@ def hexdigest(self, hashname="sha1"):
...
         return checksum
 
-    def __getstate__(self):
-        return (self._store, self._path, self._read_only, self._chunk_store,
-                self._synchronizer, self._cache_metadata, self._attrs.cache,
-                self._partial_decompress, self._write_empty_chunks, self._version)
-
-    def __setstate__(self, state):
-        self.__init__(*state)
+    def __reduce__(self):
```
jakirkham, line 2356: Thoughts on implementing `__getnewargs_ex__` instead?

In particular [this Python doc]( https://docs.python.org/3/library/pickle.html#pickling-class-instances ) text seems relevant:

> Although powerful, implementing `__reduce__()` directly in your classes is error prone. For this reason, class designers should use the high-level interface (i.e., `__getnewargs_ex__()`, `__getstate__()` and `__setstate__()`) whenever possible.

Same question with other changes below

zarr/core.py
```diff
@@ -2353,13 +2353,18 @@ def hexdigest(self, hashname="sha1"):
...
+        args = (self.store,
```
jakirkham, line 2357: Is there a reason for using `self.store` instead of `self._store` as used before? Same question with other lines below

joshmoore:
Things are green and reading https://docs.python.org/3/library/pickle.html#object.__reduce__ this seems to follow sensibly, but open to other opinions.

@d-v-b: can you coordinate the entries you want for docs/release.rst for this?

jakirkham (commented):
Thanks Davis! 🙏

Agree this can be better. Have some suggestions on how to approach that below

zarr/core.py
d-v-b: I wasn't sure which to use, but the public properties seemed more intuitive to me. Happy to change to using the private properties if there's a material difference.

jakirkham: I'd stick with what is already here

zarr/core.py
d-v-b: how does implementing `__getnewargs_ex__` avoid the need to call `__init__` directly in `__setstate__`?

jakirkham: Because this only return the arguments that are passed to `__new__` and not a function to call (nor does it call one) itself. Also because `__setstate__` would be dropped then.

jakirkham: Note: We may be able to use `__getnewargs__` instead of `__getnewargs_ex__` if we only use positional arguments, which is preferred

From [`__getnewargs_ex__` docs]( https://docs.python.org/3/library/pickle.html#object.__getnewargs_ex__ ):

> You should implement this method if the [`__new__()`]( https://docs.python.org/3/reference/datamodel.html#object.__new__ ) method of your class requires keyword-only arguments. Otherwise, it is recommended for compatibility to implement [`__getnewargs__()`]( https://docs.python.org/3/library/pickle.html#object.__getnewargs__ ).

jakirkham: `__new__` is run before `__init__`. The arguments would go to `__init__` as well. We just wouldn't have to do that ourselves. Details in [these docs]( https://docs.python.org/3/reference/datamodel.html#object.__new__ ).

Sure this would work

```python
import pickle


class Value:
    def __init__(self, value):
        self.value = value

    def __getnewargs__(self):
        return (self.value,)


v = Value(5)
v2 = pickle.loads(pickle.dumps(v))
assert 5 == v.value == v2.value
```

d-v-b: OK, that's very helpful! I'll look into using this instead of `__reduce__`

jhamman:

PR #1094, opened 2022-07-27
N5 root array fix

### LINKED ISSUE #1092: n5 format output by zarr not compatible with other programs
### Zarr version

2.11.1

### Numcodecs version

0.9.1

### Python Version

3.9

### Operating System

Windows

### Installation

Installed using conda-forge

### Description

As described in [this issue](https://github.com/saalfeldlab/paintera/issues/457) the .n5 format output by zarr produces errors when users try to load it into other software requiring n5 input: 

Error message:
```
A valid N5 container is a directory that contains a file attributes.json with a key "n5"
```

This is happening because the data chunks are being written straight into the top level directory (the one with the `.n5` suffix), eg:
```
test.n5/
  - attributes.json
  - 0/
  - 1/
  - 2/
  - ...
```

To fix it, we can write the chunks into a subfolder (perhaps with the default dataset name "raw", with a keyword argument for the user to control). So instead it'd look like this:
```
test.n5/
  - attributes.json  # a new file, with the only contents  like this `{"n5": "2.5.0"}`
  - raw/
    - .attributes.json  # this is the same atributes.json that was previously written into the topmost directory level
    - 0/
    - 1/
    - 2/
    - ...
```

I've chosen the bug template to describe this issue since I feel that interoperability with other programs accepting `.n5` files is pretty important, and is what makes this function so useful in the first place.

### Steps to reproduce

1. Create some dummy data, and save it to N5 using zarr, following the instructions at https://zarr.readthedocs.io/en/stable/api/n5.html

```python
import dask.array as da
import zarr

# Create some dummy data
data_shape = (1000, 1000)
data_chunks = (100, 100)
dummy_data = da.random.randint(0, 255, data_shape, chunks=data_chunks)

# Save as N5 using the zarr library
store = zarr.N5Store('test.n5')
z = zarr.zeros(data_shape, chunks=data_chunks, store=store, overwrite=True)
z[:] = dummy_data
```

2. Attempt to load the data in to the [Paintera software](https://github.com/saalfeldlab/paintera), and observe the error message.

```
conda install -c conda-forge paintera
```

* From the menu, select: `File` > `Open` >  `Raw/Label Source`
* In the dialog box that pops up, select `Find` > `Find Folder`; and then select the `test.n5` file generated using zarr.
* From the `Dataset` dropdown, select whichever dataset is shown
* Then click "Ok" on the file open dialog.

Observe the error message:
```
A valid N5 container is a directory that contains a file attributes.json with a key "n5"
```

### Additional output

_No response_

d-v-b:
Adds a routine to `__setitem__` for `N5Store` and `N5FSStore` that inserts the `n5` keyword to arrays that happen to exist at the root of the hierarchy, and adds tests for this behavior.

Fixes #1092 

TODO:
* [] New/modified features documented in docs/tutorial.rst


d-v-b:
cc @GenevieveBuckley

d-v-b:
I went kind of crazy and removed a lot of the old `# pragma: no cover` statements and added corresponding tests. Locally I'm at 100% coverage for `n5.py`.

d-v-b:
the automated tests are timing out, @jakirkham any idea why that might be?

joshmoore:
@d-v-b, _lots_ of goodness in this. Would you suggest this appropriate for an immediate 2.12.1?

d-v-b:
@joshmoore i need to double-check that I didn't break something here: https://github.com/zarr-developers/zarr-python/pull/1094/files#diff-427748c623f9d38d332c4b0e7e733c52bad927fee7b7ad80c36789aa85a8e142R928

This code block was excluded from testing before, and I changed it to pass tests (specifically, I call `.tobytes()` on the output instead of returning a numpy array). This makes the routine type-stable, but I need to make sure that it doesn't break stuff on n5 datasets generated by the java n5 library. I will let you know when that testing is finished.

zarr/n5.py
```diff
@@ -889,13 +919,13 @@ def decode(self, chunk, out=None):
...
             # read partial chunk
-            if chunk_shape != self.chunk_shape:  # pragma: no cover
+            if chunk_shape != self.chunk_shape:
                 chunk = np.frombuffer(chunk, dtype=self.dtype)
                 chunk = chunk.reshape(chunk_shape)
                 complete_chunk = np.zeros(self.chunk_shape, dtype=self.dtype)
                 target_slices = tuple(slice(0, s) for s in chunk_shape)
                 complete_chunk[target_slices] = chunk
-                chunk = complete_chunk
+                chunk = complete_chunk.tobytes()
```
jakirkham, line 928: Could you please expand a bit on why this is needed? Mentioning as this effectively doing another allocation/copy here

jakirkham (commented):
Thanks Davis! 🙏

Had a question below

zarr/n5.py
d-v-b: I don't think it's needed, and I removed it

d-v-b:
@joshmoore I reverted the change I was worried about, so I think this is ready to go

GenevieveBuckley:
Thank you for working on this! However, I don't think this quite matches the N5 spec. 

I've just tried making an N5 file with this PR branch, and the output cannot be read in by the Paintera software. I'm no expert on the N5 spec, but Paintera was developed by the same group that developed the N5 format, so I feel like this is a good test. The problem is that there is not a separate dataset subdirectory, instead the dataset is left loose in the top level root directory. 

Caleb [says in a comment here](https://github.com/saalfeldlab/paintera/issues/457#issuecomment-1155816174) that generally it is expected that N5 directories have this format:
```
container.n5
│   attributes.json         <--- this contains the n5 version tag
└───data
     │   attributes.json    <--- this is the file you posted above
     │   .... (actual data) 
```

What we get from this PR is instead:
```
container.n5
│   attributes.json    <--- now with an extra n5 version tag in the dictionary
│   .... (actual data) 
```

GenevieveBuckley:
Wait, no I **can** open it in Paintera! It's just that:
1. I needed to create a 3D dummy dataset, and 
2. It is extremely difficult to position the mouse to click on the single character `/` that is the root dataset. Mostly it just looks like the dataset dropdown menu is blank & if you do try to click it won't select the root dataset unless your mouse is in *just* the right position. So it's easy to assume there is nothing there, or that the file cannot be opened.

GenevieveBuckley:
Both of these examples work now, yay 🎉 

```python
In [1]: import numpy as np
In [2]: import zarr

In [3]: data_shape = (1000, 1000, 1000)
In [4]: data_chunks = (100, 100, 100)
In [5]: dummy_data = np.random.randint(0, 255, data_shape)

# Example 1: Group dataset
In [6]: store = zarr.N5Store('group.n5')
In [7]: root = zarr.group(store=store, overwrite=True)
In [8]: foo = root.create_group('foo')
In [9]: bar = foo.zeros('bar', shape=data_shape, chunks=data_chunks)
In [10]: bar[:] = dummy_data

# Example 2: a single dataset stored in the root
In [11]: store = zarr.N5Store('array.n5')
In [12]: z = zarr.zeros((10,10, 10), chunks=(5, 5, 5), store=store, overwrite=True)
In [13]: z[...] = 42
```

I am able to open both `array.n5` and `group.n5` with Paintera. Thank you @d-v-b

GenevieveBuckley (approved):

GenevieveBuckley:
One thing I did not actually check is whether this works just as well with `N5FSStore` as it does with `N5Store`. I'm guessing it will, since it's the `N5ChunkWrapper` that has these changes. Just letting you know I didn't explicitly test that out myself.

joshmoore:
Thanks, @GenevieveBuckley. Merging. My plan (as I mentioned on [gitter](https://gitter.im/zarr-developers/community)) was to target this for a 2.13 pre-release. If anyone needs a 2.12 patch release, please let me know.

docs/release.rst
```diff
@@ -29,6 +29,13 @@ Maintenance
   By :user:`Saransh Chopra <Saransh-cpp>` :issue:`1079`.
 
 
+Bug fixes
+~~~~~~~~~
+
+* Fix bug in N5 storage that prevented arrays located in the root of the hierarchy from
+  bearing the `n5` keyword. Along with fixing this bug, new tests were added for N5 routines
+  that had previously been excluded from testing, and type annotations were added to the N5 codebase.
+  By :user:`Davis Bennett <d-v-b>`; :issue:`1092`.
```
joshmoore, line 38:
```suggestion
  By :user:`Davis Bennett <d-v-b>`; :issue:`1092`.

```

PR #1096, opened 2022-07-28
add storage_transformers and get/set_partial_values

jstriebel:
This PR adds two features of the [current v3 spec](https://zarr-specs.readthedocs.io/en/latest/core/v3.0.html) to the zarr-python implementation:
* storage transformers
* `get_partial_values` and `set_partial_values`

The current spec is not finalized and waiting for finishing [review](https://github.com/zarr-developers/zarr-specs/pull/149) and approval of [ZEP001](https://zarr.dev/zeps/draft/ZEP0001.html). Since all v3 features are currently behind the `ZARR_V3_EXPERIMENTAL_API` flag, I think it might still be fine to incorporate those features to gain early feedback.

Both features are prerequisites for sharding as described [here](https://zarr-specs.readthedocs.io/en/latest/extensions/storage-transformers/sharding/v1.0.html#sharding-storage-transformer-v1), which I'd like to add as a follow-up PR, together with a new ZEP for it.

TODO:
* [New/modified features documented in docs/tutorial.rst] (v3 features seem not to be documented yet?)

### DIFF
--- a/zarr/_storage/store.py
+++ b/zarr/_storage/store.py
@@ -1,6 +1,8 @@
 import abc
 import os
+from collections import defaultdict
 from collections.abc import MutableMapping
+from copy import copy
 from string import ascii_letters, digits
 from typing import Any, List, Mapping, Optional, Union
 
@@ -254,6 +256,44 @@ class StoreV3(BaseStore):
     def __getitem__(self, key):
         """Get a value."""
 
+    def get_partial_values(self, key_ranges):
+        """Get multiple partial values.
+        key_ranges can be an iterable of key, range pairs,
+        where a range specifies two integers range_start and range_length
+        as a tuple, (range_start, range_length).
+        Length may be None to indicate to read until the end.
+        A key may occur multiple times with different ranges."""
+        results = [None] * len(key_ranges)
+        indexed_ranges_by_key = defaultdict(list)
+        for i, (key, range_) in enumerate(key_ranges):
+            indexed_ranges_by_key[key].append((i, range_))
+        for key, indexed_ranges in indexed_ranges_by_key.items():
+            value = self[key]
+            for i, (range_from, range_length) in indexed_ranges:
+                if range_length is None:
+                    results[i] = value[range_from:]
+                else:
+                    results[i] = value[range_from:range_from + range_length]
+        return results
+
+    def set_partial_values(self, key_start_values):
+        """Set multiple partial values.
+        key_start_values can be an iterable of key, start and value triplets
+        as tuples, (key, start, value), where start defines the offset in bytes.
+        A key may occur multiple times with different starts and non-overlapping values.
+        Also, start may only be beyond the current value if other values fill the gap."""
+        unique_keys = set(next(zip(*key_start_values)))
+        values = {key: bytearray(self.get(key)) for key in unique_keys}
+        for key, start, value in key_start_values:
+            if values[key] is None:
+                assert start == 0
+                values[key] = value
+            else:
+                assert start <= len(values[key])
+                values[key][start:start + len(value)] = value
+        for key, value in values.items():
+            self[key] = value
+
     def clear(self):
         """Remove all items from store."""
         self.erase_prefix("/")
@@ -301,6 +341,115 @@ class StoreV3(BaseStore):
             "if your store exposes the MutableMapping interface wrap it in "
             f"Zarr.storage.KVStoreV3. Got {store}"
         )
+
+
+class StorageTransformer(MutableMapping, abc.ABC):
+    def __init__(self, _type) -> None:
+        assert _type in self.valid_types
+        self.type = _type
+        self._inner_store = None
+
+    def _copy_for_array(self, inner_store):
+        transformer_copy = copy(self)
+        transformer_copy._inner_store = inner_store
+        return transformer_copy
+
+    @abc.abstractproperty
+    def extension_uri(self):
+        pass
+
+    @abc.abstractproperty
+    def valid_types(self):
+        pass
+
+    def get_config(self):
+        """Return a dictionary holding configuration parameters for this
+        storage transformer. All values must be compatible with JSON encoding."""
+        # Override in sub-class if need special encoding of config values.
+        # By default, assume all non-private members are configuration
+        # parameters except for type .
+        return {
+            k: v for k, v in self.__dict__.items()
+            if not k.startswith('_') and k != "type"
+        }
+
+    @classmethod
+    def from_config(cls, _type, config):
+        """Instantiate storage transformer from a configuration object."""
+        # override in sub-class if need special decoding of config values
+
+        # by default, assume constructor accepts configuration parameters as
+        # keyword arguments without any special decoding
+        return cls(_type, **config)
+
+    def is_readable(self):
+        return self._inner_store.is_readable()
+
+    def is_writeable(self):
+        return self._inner_store.is_writeable()
+
+    def is_listable(self):
+        return self._inner_store.is_listable()
+
+    def is_erasable(self):
+        return self._inner_store.is_erasable()
+
+    def __enter__(self):
+        return self._inner_store.__enter__()
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        return self._inner_store.__exit__(exc_type, exc_value, traceback)
+
+    def close(self) -> None:
+        return self._inner_store.close()
+
+    def rename(self, src_path: str, dst_path: str) -> None:
+        return self._inner_store.rename(src_path, dst_path)
+
+    def list_prefix(self, prefix):
+        return self._inner_store.list_prefix(prefix)
+
+    def erase(self, key):
+        return self._inner_store.erase(key)
+
+    def erase_prefix(self, prefix):
+        return self._inner_store.erase_prefix(prefix)
+
+    def list_dir(self, prefix):
+        return self._inner_store.list_dir(prefix)
+
+    def list(self):
+        return self._inner_store.list()
+
+    def __contains__(self, key):
+        return self._inner_store.__contains__(key)
+
+    def __setitem__(self, key, value):
+        return self._inner_store.__setitem__(key, value)
+
+    def __getitem__(self, key):
+        return self._inner_store.__getitem__(key)
+
+    def __delitem__(self, key):
+        return self._inner_store.__delitem__(key)
+
+    def __iter__(self):
+        return self._inner_store.__iter__()
+
+    def __len__(self):
+        return self._inner_store.__len__()
+
+    def get_partial_values(self, key_ranges):
+        return self._inner_store.get_partial_values(key_ranges)
+
+    def set_partial_values(self, key_start_values):
+        return self._inner_store.set_partial_values(key_start_values)
+
+    def clear(self):
+        return self._inner_store.clear()
+
+    def __eq__(self, other):
+        return self._inner_store.__eq__(other)
 
 
 # allow MutableMapping for backwards compatibility
--- a/zarr/core.py
+++ b/zarr/core.py
@@ -289,7 +289,9 @@ class Array:
                 if self._chunk_store is not None:
                     transformed_chunk_store = self._chunk_store
                     for storage_transformer in storage_transformers:
-                        transformed_chunk_store = storage_transformer._copy_for_array(transformed_chunk_store)
+                        transformed_chunk_store = (
+                            storage_transformer._copy_for_array(transformed_chunk_store)
+                        )
                     self._chunk_store = transformed_chunk_store
 
     def _refresh_metadata(self):
--- a/zarr/core.py
+++ b/zarr/core.py
@@ -280,6 +280,18 @@ class Array:
                 filters = [get_codec(config) for config in filters]
             self._filters = filters
 
+            if self._version == 3:
+                storage_transformers = meta.get('storage_transformers', [])
+                transformed_store = self._store
+                for storage_transformer in storage_transformers:
+                    transformed_store = storage_transformer._copy_for_array(transformed_store)
+                self._store = transformed_store
+                if self._chunk_store is not None:
+                    transformed_chunk_store = self._chunk_store
+                    for storage_transformer in storage_transformers:
+                        transformed_chunk_store = storage_transformer._copy_for_array(transformed_chunk_store)
+                    self._chunk_store = transformed_chunk_store
+
     def _refresh_metadata(self):
         if not self._cache_metadata:
             self._load_metadata()
--- a/zarr/creation.py
+++ b/zarr/creation.py
@@ -21,7 +21,7 @@ def create(shape, chunks=True, dtype=None, compressor='default',
            overwrite=False, path=None, chunk_store=None, filters=None,
            cache_metadata=True, cache_attrs=True, read_only=False,
            object_codec=None, dimension_separator=None, write_empty_chunks=True,
-           *, zarr_version=None, **kwargs):
+           storage_transformers=None, *, zarr_version=None, **kwargs):
     """Create an array.
 
     Parameters
@@ -161,7 +161,7 @@ def create(shape, chunks=True, dtype=None, compressor='default',
     init_array(store, shape=shape, chunks=chunks, dtype=dtype, compressor=compressor,
                fill_value=fill_value, order=order, overwrite=overwrite, path=path,
                chunk_store=chunk_store, filters=filters, object_codec=object_codec,
-               dimension_separator=dimension_separator)
+               dimension_separator=dimension_separator, storage_transformers=storage_transformers)
 
     # instantiate array
     z = Array(store, path=path, chunk_store=chunk_store, synchronizer=synchronizer,
--- a/zarr/meta.py
+++ b/zarr/meta.py
@@ -9,7 +9,11 @@ from numcodecs.abc import Codec
 from zarr.errors import MetadataError
 from zarr.util import json_dumps, json_loads
 
-from typing import cast, Union, Any, List, Mapping as MappingType, Optional
+from typing import cast, Union, Any, List, Mapping as MappingType, Optional, TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from zarr._storage.store import StorageTransformer
+
 
 ZARR_FORMAT = 2
 ZARR_FORMAT_v3 = 3
@@ -460,6 +464,31 @@ class Metadata3(Metadata2):
         return codec
 
     @classmethod
+    def _encode_storage_transformer_metadata(cls, storage_transformer: "StorageTransformer") -> Optional[Mapping]:
+        return {
+            "extension": storage_transformer.extension_uri,
+            "type": storage_transformer.type,
+            "configuration": storage_transformer.get_config(),
+        }
+
+    @classmethod
+    def _decode_storage_transformer_metadata(cls, meta: Mapping) -> "StorageTransformer":
+        from zarr.tests.test_storage_v3 import DummyStorageTransfomer
+        KNOWN_STORAGE_TRANSFORMERS = [DummyStorageTransfomer]
+
+        conf = meta.get('configuration', {})
+        extension_uri = meta['extension']
+        transformer_type = meta['type']
+
+        for StorageTransformerCls in KNOWN_STORAGE_TRANSFORMERS:
+            if StorageTransformerCls.extension_uri == extension_uri:
+                break
+        else:
+            raise NotImplementedError
+
+        return StorageTransformerCls.from_config(transformer_type, conf)
+
+    @classmethod
     def decode_array_metadata(cls, s: Union[MappingType, str]) -> MappingType[str, Any]:
         meta = cls.parse_metadata(s)
 
@@ -476,6 +505,11 @@ class Metadata3(Metadata2):
             # TODO: remove dimension_separator?
 
             compressor = cls._decode_codec_metadata(meta.get("compressor", None))
+            storage_transformers = meta.get("storage_transformers", None)
+            if storage_transformers:
+                storage_transformers = [
+                    cls._decode_storage_transformer_metadata(i) for i in storage_transformers
+                ]
             extensions = meta.get("extensions", [])
             meta = dict(
                 shape=tuple(meta["shape"]),
@@ -493,6 +527,8 @@ class Metadata3(Metadata2):
             # compressor field should be absent when there is no compression
             if compressor:
                 meta['compressor'] = compressor
+            if storage_transformers:
+                meta['storage_transformers'] = storage_transformers
 
         except Exception as e:
             raise MetadataError("error decoding metadata: %s" % e)
@@ -514,6 +550,11 @@ class Metadata3(Metadata2):
             object_codec = None
 
         compressor = cls._encode_codec_metadata(meta.get("compressor", None))
+        storage_transformers = meta.get("storage_transformers", None)
+        if storage_transformers:
+            storage_transformers = [
+                cls._encode_storage_transformer_metadata(i) for i in storage_transformers
+            ]
         extensions = meta.get("extensions", [])
         meta = dict(
             shape=meta["shape"] + sdshape,
@@ -532,6 +573,8 @@ class Metadata3(Metadata2):
             meta["compressor"] = compressor
         if dimension_separator:
             meta["dimension_separator"] = dimension_separator
+        if storage_transformers:
+            meta["storage_transformers"] = storage_transformers
         return json_dumps(meta)
 
 
--- a/zarr/meta.py
+++ b/zarr/meta.py
@@ -464,7 +464,10 @@ class Metadata3(Metadata2):
         return codec
 
     @classmethod
-    def _encode_storage_transformer_metadata(cls, storage_transformer: "StorageTransformer") -> Optional[Mapping]:
+    def _encode_storage_transformer_metadata(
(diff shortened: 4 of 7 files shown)


jstriebel:
The Linux tests fail due to some s3 tests, and the Windows test timeout in the s3 fixture setup, which also fail on different branches and seem to be unrelated to this PR.

jstriebel:
@joshmoore & @grlee77 Would one of you have time to review my changes here? Asking since you authored the initial v3 additions here. From my side this PR is ready for review.

zarr/tests/test_storage_v3.py
```diff
@@ -235,6 +254,46 @@ def test_rename_nonexisting(self):
             with pytest.raises(NotImplementedError):
                 store.rename('a', 'b')
 
+    def test_get_partial_values(self):
```
grlee77, line 257: This is a nice test case. One thing that still looks missing here is also including a range that has `None` for the second value? (Based on the `StorageTransformer` class it seems that is also supported)

zarr/meta.py
```diff
@@ -459,6 +463,36 @@ def _decode_codec_metadata(cls, meta: Optional[Mapping]) -> Optional[Codec]:
...
+    ) -> Optional[Mapping]:
+        return {
+            "extension": storage_transformer.extension_uri,
+            "type": storage_transformer.type,
+            "configuration": storage_transformer.get_config(),
+        }
+
+    @classmethod
+    def _decode_storage_transformer_metadata(cls, meta: Mapping) -> "StorageTransformer":
+        from zarr.tests.test_storage_v3 import DummyStorageTransfomer
```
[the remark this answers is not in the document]
grlee77: I guess in answer to my own question, we probably don't want to encourage anyone else to use the `DummyStorageTransformer`? Those not wanting any transformer should just be calling `init_array` with the default `storage_transformers=None`.

grlee77 (changes requested):
Thanks, @jstriebel! I have some minor suggestions, but this looks pretty good already. I think we will get more experience and a better idea of a real-world use case with the follow-up sharding PR.

A couple more thoughts that just came up from an end-user perspective:
- does passing `storage_transformers=[]` work equivalently to `storage_transformers=None`?
- are there plans to expose `storage_transformers` via the higher level convenience/creation functions? (e.g. `zarr/creation.py`)

zarr/tests/test_storage_v3.py
jstriebel: Good point, I added this :+1:

zarr/creation.py
```diff
@@ -21,7 +21,7 @@ def create(shape, chunks=True, dtype=None, compressor='default',
            overwrite=False, path=None, chunk_store=None, filters=None,
            cache_metadata=True, cache_attrs=True, read_only=False,
            object_codec=None, dimension_separator=None, write_empty_chunks=True,
-           *, zarr_version=None, **kwargs):
+           storage_transformers=None, *, zarr_version=None, **kwargs):
```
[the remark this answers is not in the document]
jstriebel: Good idea, especially since this can only be used with the keyword-only `zarr_version` together atm :+1:

zarr/meta.py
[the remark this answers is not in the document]
jstriebel: Exactly, this is only supposed to be used in tests. When adding a proper registry for the transformers, the tests might even add it themselves to the registry. It's currently used both in `test_core.py` and `test_storage_v3.py`. I found it most appropriate to put it into the latter and include it from there.

jstriebel:
> ! I have some minor suggestions, but this looks pretty good already.

Thank you so much for the review, @grlee77! I applied your feedback in the recent commits.

> I think we will get more experience and a better idea of a real-world use case with the follow-up sharding PR.

Yes, I think the followup-PR will help to make more sense for those methods.

> * does passing `storage_transformers=[]` work equivalently to `storage_transformers=None`?

Yes, it does, I added a notice for it in the docstring. I also thought about just having an empty list as the default, but lists as default arguments are somewhat deprecated due to their mutability.

> * are there plans to expose `storage_transformers` via the higher level convenience/creation functions? (e.g. `zarr/creation.py`)

I think it is already exposed via `create`. I also thought about exposing specific shorthand parameters for single storage transformers, e.g. `create(…, chunks_per_shard=(32, 32, 32))`, but that's not strictly necessary. Are there any other places where storage transformers should be passed, that I'm missing?

zarr/meta.py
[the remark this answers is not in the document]
joshmoore: FWIW, during the last community call meeting (27.07) it was discussed for codecs having `[]` be the "dummy" (or "raw") codec.

zarr/_storage/store.py
```diff
@@ -303,6 +348,125 @@ def _ensure_store(store):
         )
 
 
+class StorageTransformer(MutableMapping, abc.ABC):
+    """Base class for storage transformers. The methods simply pass on the data as-is
+    and should be overwritten by sub-classes."""
+
+    _store_version = 3
```
joshmoore, line 355: Slight surprise to find this in store.py rather than under v3.

joshmoore (commented):
> @jstriebel I think it might still be fine to incorporate those features to gain early feedback.

Concur. :heart: for hammering on the Python API a bit while we also debate the spec.

> I also thought about just having an empty list as the default, but lists as default arguments are somewhat deprecated due to their mutability.

And a tuple?

jakirkham:
cc @martindurant (in case this is of interest)

martindurant:
@jakirkham , I must be missing something: allowing partial byte ranges to pass through to the storage layer and transformers seem completely orthogonal to me. 

In the case of this code, on [this line](https://github.com/zarr-developers/zarr-python/pull/1096/files#diff-d214c5fd1f070e07da44757169a21a6fb6de9a59eec35be45e0a09739ea7da22R271) we get the full chunk and then slice it anyway. Is the intend for some storage classes (fsspec!) to subclass this and implement a new get_partial_values?

jstriebel:
> allowing partial byte ranges to pass through to the storage layer and transformers seem completely orthogonal to me.

Yes, it is to some extend. I can also have them as separate PR. My motivation was to implement sharding, and I tried to keep the bases for it as a separate PR, that's how they both ended up here.

> In the case of this code, on [this line](https://github.com/zarr-developers/zarr-python/pull/1096/files#diff-d214c5fd1f070e07da44757169a21a6fb6de9a59eec35be45e0a09739ea7da22R271) we get the full chunk and then slice it anyway. Is the intend for some storage classes (fsspec!) to subclass this and implement a new get_partial_values?

Yes, the idea is to overwrite those functions for stores that allow partial access. Here I just implemented the API and the fallback strategy, I'll open an issue for the optimized paths as a follow-up, also mentioned in the TODOs in the description above:
> * Add follow-up issue to optimize get_partial_values and set_partial_values for specific stores

Does that answer your questions, @martindurant?

martindurant:
I didn't mean to denigrate this PR, I know very little about how the storage transformer or sharding are to be implemented. I am wondering how the new get_partial_values would get called, and then it sounds like we can agree on something that works for us all.

zarr/_storage/store.py
jstriebel: I assumed that the base-classes go into `store.py`, similar to `StoreV3`, but also happy to move this into `_storage/v3.py` if that's a better fit.

jstriebel:
> > I also thought about just having an empty list as the default, but lists as default arguments are somewhat deprecated due to their mutability.
> 
> And a tuple?

Good idea, I changed the default to an empty tuple now :+1:

jstriebel:
I opened the follow-up issue #1106, which is about implementing optimized partial read/write methods for specific stores, and using them for the partial decode routines.

jstriebel:
@grlee77 @joshmoore Thank you very much for your reviews! I incorporated your feedback where I marked the comments as resolved, the other's are still open conversations. Could you give this PR another look? CI finally is completely green, I added some more tests for a higher coverage.

zarr/creation.py
```diff
@@ -84,6 +84,15 @@ def create(shape, chunks=True, dtype=None, compressor='default',
 
         .. versionadded:: 2.11
 
+    storage_transformers : sequence of StorageTransformers, optional
+        Setting storage transformers, changes the storage structure and behaviour
+        of data coming from the underlying store. The transformers are applied in the
+        order of the given sequence. Supplying an empty sequence is the same as omitting
+        the argument or setting it to None. May only be set when using zarr_version 3.
+        Supplying an empty seq
```
grlee77, line 92: looks like some duplicate text snuck in here
```suggestion
```

grlee77 (approved):
The new changes look good to me, thanks @jstriebel

jstriebel:
@grlee77 @joshmoore I've incorporated some more changes that were useful for the sharding PR #1111. I added them here to keep the PRs clean. Could you have a look at those as well? Thanks a ton!

zarr/_storage/store.py
```diff
@@ -303,6 +373,150 @@ def _ensure_store(store):
...
+    def __init__(self, _type) -> None:
+        if _type not in self.valid_types:  # pragma: no cover
+            raise ValueError(
+                f"Storage transformer cannot be initialized with type {_type}, "
+                + f"must be one of {list(self.valid_types)}."
+            )
+        self.type = _type
+        self._inner_store = None
+
+    def _copy_for_array(self, array, inner_store):
```
grlee77, line 392: what is the purpose of the unused `array` argument here? Is it a intended that subclasses would use this?

martindurant:
Above I asked "how would the new get_partial_values would get called"? At it stands, this is all unused code in this PR (from a zarr user's perspective, as opposed to tests), right?

I am still trying to wrap my head around why we need a new method here, as opposed, to just `getitems(keys, ranges=None)`.

**In fact**, I have been mulling over the idea of passing a _context_ (e.g., requested chunk key, chunk slice, versus array position) through the filters/compressors and to the storage layer, such that they can take whatever action they like with that information. For example, could have `Codec.decode(buf, context=None)` which applies a different scale factor as a factor of dimension 3 ( https://github.com/fsspec/kerchunk/issues/212 ) and use the same information object in `getitems(keys, contexts)` to decide what byterange to fetch.

zarr/_storage/store.py
jstriebel: Exactly, I've seen a need for this when implementing the Sharding storage transformer in #1111:
https://github.com/zarr-developers/zarr-python/blob/6e2790c6b5596471bbb47ccdcad8d9f035942f58/zarr/_storage/v3_storage_transformers.py#L99-L102

Related to this: I'm not super happy about the double-initialization using both `__init__` and `_copy_for_array`, but it seems necessary to supply the information in two steps, first from the config, then from the array. Please let me know if you have an idea how to solve this more elegantly.

jstriebel:
> Above I asked "how would the new get_partial_values would get called"? At it stands, this is all unused code in this PR (from a zarr user's perspective, as opposed to tests), right?

Exactly, the methods are only used for "real" code in the sharding PR #1111. From the description of this PR:
> Both features are prerequisites for sharding as described [here](https://zarr-specs.readthedocs.io/en/latest/extensions/storage-transformers/sharding/v1.0.html#sharding-storage-transformer-v1), which I'd like to add as a follow-up PR, together with a new ZEP for it.

> I am still trying to wrap my head around why we need a new method here, as opposed, to just `getitems(keys, ranges=None)`.

That proposal is new to me, what exactly would `ranges` or `context` be here? I wonder how the returned value should look like, especially when multiple ranges can be specified for the same key. Atm `getitems` returns a dict with the storage-keys as dict-keys, but it might be fine to add the ranges to the dict-keys as well, returning something like `{("a_key", (0, 10)): b"…", ("a_key", (-16, None)): b"…"}`.
The main reason for the new methods is that I described them this way in the [v3 spec](https://zarr-specs.readthedocs.io/en/latest/core/v3.0.html#abstract-store-interface), but any other API for the same functionality should be fine :+1: 
It's definitely useful to have some coherence when a codec allows partial reads, or could even request sequential partial reads from a store. Is there some work on this going on already?

I don't have a strong opinion here, please also feel free to directly propose code changes, e.g. as a PR against this PR.

jstriebel:
> I would very much like for partial reads to be available for uncompressed (unsharded) data, that is a major concern of mine, but this PR doesn't provide it as it stands, and I don't see a path to providing it based on the proposed architecture.

This is also implemented in #1111, please have a look at this PR as well, I think it might clarify the usage of the proposed methods quite a bit.

> The other concern is about a codec knowing which chunk its working on when en/decoding. Maybe the two concerns are not part of the same puzzle...

I agree, that's quite important to allow generic partial reads (and maybe writes?) via codecs, not just for blosc. However, I think that's a separate problem, and initially allowing partial reads from the *stores* will make this much easier. Changing the API slightly (e.g. merging the partial reads into `getitems`) might still be fine later, after all v3 is only in a preview state atm. Therefore I see this rather as a minor issue, do you agree @martindurant?

martindurant:
(for the codecs, I was actually thinking about parameters that depend on location in the overall array, not partial reads, although that could be part of it too)

jstriebel:
@grlee77 @joshmoore @MSanKeys963 
I merged the newest changes into this branch. From my POV this seems ready to be merged. I'll also update #1111 (the PR about sharding) as discussed (putting sharding behind a zep 2 flag), which we can merge separately afterwards.

martindurant:
Did you have any thoughts of the interaction of this with https://github.com/zarr-developers/zarr-python/pull/1131 ?

jstriebel:
> Did you have any thoughts of the interaction of this with https://github.com/zarr-developers/zarr-python/pull/1131 ?

Thanks for making me aware of that, not yet. #1131 is mostly about providing a context, which might just get passed through storage transformers. It's not clear to me why stores do handle arrays directly instead of byte-buffers, but that's a completely orthogonal discussion I'd rather like to keep out of this PR. As #1131 stands right now this would not change much wrt this PR. @madsbk, do you have any concerns regarding this PR?

martindurant:
I am not really talking about transformers - the main thrust of this PR - but the ranges.

I raise it, because the byterange (or array slice, more likely) is something that should be included in the context, so then you would not need extra methods on the storage class. That's what I was after in my comments above. It also provides a direct way to actually activate the partial read. As opposed to in the code here, where the methods exist, but there is no obvious path to calling them.

jstriebel:
@martindurant Thanks for the clarification.

> I raise it, because the byterange (or array slice, more likely) is something that should be included in the context, so then you would not need extra methods on the storage class.

Sure, that's possible, but I don't see a benefit. The context would then need to force stores into different behavior, and the returned content would depend heavily on the context. I find it more clean to separate the different behavior into a different method. But not a super-strong opinion.

> As opposed to in the code here, where the methods exist, but there is no obvious path to calling them.

The method's can just be called where appropriate, e.g. for partial reads on uncompressed chunks or sharding. Please see #1111 for usage examples.

martindurant:
> The methods can just be called where appropriate

Can I be super annoying and ask for a smallest possible example here? That other PR and thread is very long and not focussed on partial reads. 
```python
x = group.create_dataset(name="x", data=np.array([1,2,3,4]), compression=None, chunks=(4, ))
x[:2]
```
where the second line should read only 16 bytes from the store, not all 32, if the store supports partials. 

> The context would then need to force stores into different behavior, and the returned content would depend heavily on the context. 

I don't follow this. For the above case, 
```
context={"key": "data/0.0", "out_selection": [slice(0, 2)], "chunk_selection": [slice(0, 2), "meta_array": np.NDArray, "dtype": np.int64]}
```
where all these things are known in Array._chunk_getitem (meta_array is new and shows that we can do more with this idea, from the other PR). Return bytes for the chunk would be
```
b'\x01\x00\x00\x00\x00\x00\x00\x00\x02\x00\x00\x00\x00\x00\x00\x00'
```
and 16 more bytes of or empty/zeros; if we can't just truncate instead.

jstriebel:
@martindurant No worries! This is a short example for your case:
```
x = group.create_dataset(name="x", data=np.array([1,2,3,4]), compression=None, chunks=(4, ))
x[:2]

# with PR #1111 in place, the indexing would do:
store.get_partial_values[(store_key, (0, 2 * itemsize))]
# where store_key is probably sth like "x/0", and itemsize=8 for the int64 dtype
# returning [b'\x01\x00\x00\x00\x00\x00\x00\x00\x02\x00\x00\x00\x00\x00\x00\x00']
```

(For uncompressed partial reads the implementation is [here in PR #1111](https://github.com/zarr-developers/zarr-python/pull/1111/files#diff-f389c1f0162cd60b1b5a4187dad20dfd3d22c98da2ebe22e860ea830bc25a0e9R652-R655)

Also the tests have some more examples:
https://github.com/zarr-developers/zarr-python/pull/1096/files#diff-49f50043d526642ad824a8d6a944b896540238b06ec11f182476ff8fdb2b5977R257-R316

I think I finally understand where you are coming from. With the context variant the parts that are not requested would simply be filled I guess (or truncated, as you mentioned). However, I'd prefer to only have a buffer for the actual parts of the chunk, which might be spread through the chunk, so one single buffer would not be enough. Therefore the return type would need to be different for the partial reads, which fits more nicely with a new method IMO. Does that make sense?

martindurant:
OK, I got it: so the code that calls the methods introduced here actually are in the other PR. That call is not really to do with sharding, which is why I got confused. Actually I don't see how that uncompressed-partial-buffer gets used either, there seems to be many levels here.

Are you saying that you want a single call to chunk-getitem to produce many buffers, which will then be filled into the target output array? That would work with contexts too, get([key1, key1], contexts=[context1, context2]). I think in the proposal contexts= was a dict, so it would indeed need to be a list instead. That makes the API the same as here (and we could even have both).

zarr/core.py
```diff
@@ -1800,7 +1813,7 @@ def _set_selection(self, indexer, value, fields=None):
             check_array_shape('value', value, sel_shape)
 
         # iterate over chunks in range
-        if not hasattr(self.store, "setitems") or self._synchronizer is not None \
+        if not hasattr(self.chunk_store, "setitems") or self._synchronizer is not None \
```
joshmoore, line 1816: I have to wonder if this wouldn't ever need to be the transformed_chunk_store. In general, this chain of replacement stores feels slightly like an anti-pattern.

zarr/meta.py
```diff
@@ -459,6 +463,36 @@ def _decode_codec_metadata(cls, meta: Optional[Mapping]) -> Optional[Codec]:
 
         return codec
 
+    @classmethod
+    def _encode_storage_transformer_metadata(
```
joshmoore, line 467: After the splitting of `storage.py` into separate files, this makes me wonder if meta and some of the other core classes shouldn't also have v3 variants.

joshmoore (approved):
Two more general points more geared towards potential future refactorings. I don't see anything that causes alarm knowing that v3 is still in flux and protected by the environment variables. Primary question is if you would see still getting this into a 2.13 release (especially if one were to come out over the holidays) or would it make more sense to roll it into 2.14.

Anyone else have thoughts?

joshmoore:
I should add re-reviewing the interaction with #1131/meta_array I start to wonder if the definition of "context" and "partial" don't need to end up in the spec or at least a clarifying ZEP together.

zarr/_storage/store.py
```diff
@@ -254,6 +256,74 @@ def __setitem__(self, key, value):
...
 
+    @abc.abstractmethod
+    def rmdir(self, path=None):
+        """Remove a data path and all its subkeys and related metadata.
+        Expects a path without the data or meta root prefix."""
+
+    def supports_efficient_get_partial_values(self):
+        return False
+
+    def get_partial_values(self, key_ranges):
```
rabernat, line 267: A type hint would be very useful here.

zarr/_storage/store.py
```diff
@@ -254,6 +256,74 @@ def __setitem__(self, key, value):
...
+        key_ranges can be an iterable of key, range pairs,
+        where a range specifies two integers range_start and range_length
+        as a tuple, (range_start, range_length).
+        range_length may be None to indicate to read until the end.
+        range_start may be negative to start reading range_start bytes
+        from the end of the file.
+        A key may occur multiple times with different ranges.
+        Inserts None for missing keys into the returned list."""
+        results = [None] * len(key_ranges)
+        indexed_ranges_by_key = defaultdict(list)
```
rabernat, line 278: Type hint would be good here too.

zarr/_storage/store.py
```diff
@@ -254,6 +256,74 @@ def __setitem__(self, key, value):
     def __getitem__(self, key):
         """Get a value."""
 
+    @abc.abstractmethod
+    def rmdir(self, path=None):
+        """Remove a data path and all its subkeys and related metadata.
+        Expects a path without the data or meta root prefix."""
+
+    def supports_efficient_get_partial_values(self):
+        return False
```
rabernat, line 265: A property would be better here, no?

zarr/_storage/store.py
```diff
@@ -254,6 +256,74 @@ def __setitem__(self, key, value):
...
+                continue
+            for i, (range_from, range_length) in indexed_ranges:
+                if range_length is None:
+                    results[i] = value[range_from:]
+                else:
+                    results[i] = value[range_from:range_from + range_length]
+        return results
+
+    def supports_efficient_set_partial_values(self):
+        return False
```
rabernat, line 294: Same

zarr/_storage/store.py
```diff
@@ -303,6 +373,150 @@ def _ensure_store(store):
...
+
+    def get_config(self):
+        """Return a dictionary holding configuration parameters for this
+        storage transformer. All values must be compatible with JSON encoding."""
+        # Override in sub-class if need special encoding of config values.
+        # By default, assume all non-private members are configuration
+        # parameters except for type .
+        return {
+            k: v for k, v in self.__dict__.items()
+            if not k.startswith('_') and k != "type"
```
rabernat, line 413: Why are keys starting with `_` characters ignored? Is this part of the spec?

zarr/tests/test_storage_v3.py
```diff
@@ -235,6 +254,67 @@ def test_rename_nonexisting(self):
...
+        store = self.create_store()
+        store.supports_efficient_get_partial_values()
```
rabernat, line 259:
```suggestion
        assert store.supports_efficient_get_partial_values()
```

zarr/core.py
jstriebel: Yes, I'm also not particularly happy with this pattern, but also couldn't come up with a better solution for now. Happy about any ideas :+1:

jstriebel:
@joshmoore Thanks for the review!

> Primary question is if you would see still getting this into a 2.13 release (especially if one were to come out over the holidays) or would it make more sense to roll it into 2.14.

I wouldn't mind having this in the 2.13 release, so we can play around with it a bit already. I can also add the additional flag for sharding in #1111 so we are on the safe side.

> I should add re-reviewing the interaction with https://github.com/zarr-developers/zarr-python/pull/1131/meta_array I start to wonder if the definition of "context" and "partial" don't need to end up in the spec or at least a clarifying ZEP together.

The partial access is part of the v3 spec already: https://zarr-specs.readthedocs.io/en/latest/core/v3.0.html#abstract-store-interface.

rabernat (commented):
I finally got around to reviewing this PR! 🙃 

Overall it makes sense to me. My impression is that more type hints would be helpful in a few places.

zarr/_storage/store.py
jstriebel: I used to have type-hints there in an early version, but removed them since the other methods didn't have some. Just out of curiosity: Is there some guideline when type-hints should be used? Or even a plan to add them everywhere?

I'll add them here where you commented :+1:

zarr/tests/test_storage_v3.py
jstriebel: This should not be asserted here, since not all stores have the efficient access. I just added it to ensure that the call works. I added an `in [True, False]` check instead.

zarr/_storage/store.py
jstriebel: > Why are keys starting with _ characters ignored? Is this part of the spec?

This seemed to be a good default which I copied from the `Codec` implementation:
https://github.com/zarr-developers/numcodecs/blob/main/numcodecs/abc.py#L79-L95

It's not part of the spec, and might also be overwritten by subclasses where this behavior does not fit, as written in the code comment.

rabernat (approved):

martindurant:
Is there any reason I should not press the big green button? We have three approvals already. Last chance to object!

joshmoore:
Now with the somewhat large 2.13.4 out the door, I'm going to go ahead and click the button (but would have also supported @martindurant's pushing therefore :wink: :heart:)

joshmoore:
@jstriebel: can you please push from https://github.com/scalableminds/zarr-python/pull/3? (I could not push to scalableminds forks as an admin)

cc: @normanrz

jstriebel:
> I'm going to go ahead and click the button

:rocket: 

> @jstriebel: can you please push from [scalableminds#3](https://github.com/scalableminds/zarr-python/pull/3)? (I could not push to scalableminds forks as an admin)

Sure, done :+1:

joshmoore:
Sorry, @jstriebel, but _one_ more needs to go in: https://github.com/zarr-developers/zarr-python/pull/1320

joshmoore:
Ok. Merging to spare you any more pain!

MSanKeys963:
🎉

Issue #1106, opened 2022-08-03
v3 stores: Implement efficient get/set_partial_values


jstriebel:
In #1096 `get/set_partial_values` methods were introduced to Zarr v3 stores. The provided method is a viable fallback for stores that cannot read and write partial objects. Other stores however should implement optimized methods, such as fsspec-based stores (using read_block). It might be useful that stores indicate if they have fast partial read/write methods, so that strategies such as partial decompression can be selected automatically.

As a follow-up, the new `get/set_partial_values` methods could be used for the actual partial decompression in the `PartialReadBuffer`, instead of the current store-specific implementation.

Follow-up to #1096

martindurant:
I believe that the most important use case for this is actually uncompressed arrays! That's a much simpler code path and reads no partial-reader (also happens to be the only one important to me for now).

How are you proposing that get_partial_buffer should be called? At the moment in (v2) Array._get_selection we iterate over the selections for each chunk, so we have the information right before handing off to the store.

jstriebel:
> I believe that the most important use case for this is actually uncompressed arrays! That's a much simpler code path and reads no partial-reader (also happens to be the only one important to me for now).

Indeed, that's a great use-case!

> How are you proposing that get_partial_buffer should be called? At the moment in (v2) Array._get_selection we iterate over the selections for each chunk, so we have the information right before handing off to the store.

I'll try to dump my thoughts about them:

I guess there are at least two ways:
* Simply store the partial data as-is and pass it on,
* use sth. like the PartialReadBuffer or extend it, so that the partial data has a similar interface as the whole chunk.

To solve this more holistically, the compressor (or a dummy for uncompressed arrays) should be able to tell if it can decode partial data, and have some interface for "demanding" data. In the uncompressed use-case, the requested array indices can directly be translated to chunk offsets, but in the blosc or other cases with an index, the decoder might need to read data in several passes (e.g. first getting some index, then getting the actual data, based on the index). For such cases, the PartialReadBuffer is a nice abstraction that allows to reload data in several passes, depending on the decoder. If the pattern is always to maybe get some data upfront for a chunk, and then the decoder can translate indices to offsets, this might be also be a viable option.

PS: First, we still need to implement efficient get/set_partial_values for stores where this is possible, to gain anything from it.

jhamman:
This is done now on the `v3` branch.

PR #1111, opened 2022-08-22
Sharding storage transformer for v3

jstriebel:
## Description

This PR 
* adds a sharding storage transformer:
  * based on the [current spec proposal](https://zarr-specs.readthedocs.io/en/latest/extensions/storage-transformers/sharding/v1.0.html) from [ZEP2](https://zarr.dev/zeps/draft/ZEP0002.html)
  * supports partial reads
  * writes are limited so far: partial writes are not efficient yet, shards are always completely rewritten
    (a more efficient implementation might be a good follow-up)
* adapts the v3 fsspec store to allow partial reads
* uses partial reads for uncompressed v3 arrays
  (some refactoring might be needed in `core.py` to simplify the code-paths, but I'd rather defer this to a separate PR to keep the diff readable)
* added test cases
* resolves parts of #1106

## TODO:
* ~~New/modified features documented in docs/tutorial.rst~~ (v3 is not yet documented)
* [ ] Consider adding a short-cut for sharded array creation (e.g. having a chunks_per_shard argument directly)


jstriebel:
@joshmoore & @grlee77 I think this PR is now ready for a round of review, would you have time for it? As described above, it builds upon #1096 and adds a sharding transformer implementation, together with some optimizations with partial reads. Reviewing this is probably best done on the diff PR https://github.com/scalableminds/zarr-python/pull/2.

I left one TODO open, what do you think about it?
> Consider adding a short-cut for sharded array creation (e.g. having a chunks_per_shard argument directly)

zarr/_storage/v3.py
```diff
@@ -182,6 +182,37 @@ def rmdir(self, path=None):
...
+        A key may occur multiple times with different ranges.
+        Inserts None for missing keys into the returned list."""
+        results = []
+        for key, (range_start, range_length) in key_ranges:
+            key = self._normalize_key(key)
+            path = self.dir_path(key)
+            try:
+                if range_start < 0:
+                    if range_length is None:
+                        result = self.fs.tail(path, size=-range_start)
```
martindurant, line 205: Should use `fs.cat` for one item, `fs.cat_ranges` for multiple instead of tail/read_block, which use the stateful file API. It also handles negative indices and None.

zarr/core.py
```diff
@@ -180,6 +181,7 @@ def __init__(
 
         self._store = store
         self._chunk_store = chunk_store
+        self._transformed_chunk_store = None
```
martindurant, line 184: If we have a transform, do we need to maintain a link to the base store?
Could have ._chunk_store point to the transformed one, and not have to have the `if is None` checks below.

zarr/core.py
```diff
@@ -1238,8 +1252,12 @@ def _get_selection(self, indexer, out=None, fields=None):
             check_array_shape('out', out, out_shape)
 
         # iterate over chunks
-        if not hasattr(self.chunk_store, "getitems") or \
-           any(map(lambda x: x == 0, self.shape)):
+        if (
+            not hasattr(self.chunk_store, "getitems") and not (
+                hasattr(self.chunk_store, "get_partial_values") and
```
martindurant, line 1257: Not sure about this chain of conditions. Would anything "support_efficient" without the method existing (or is it in the base class and always there?) ad why the connection to getitems?

zarr/core.py
```diff
@@ -1854,10 +1872,12 @@ def _process_chunk(
                 # contiguous, so we can decompress directly from the chunk
                 # into the destination array
                 if self._compressor:
-                    if isinstance(cdata, PartialReadBuffer):
+                    if isinstance(cdata, (PartialReadBuffer, UncompressedPartialReadBufferV3)):
```
martindurant, line 1875: How does one end up with UncompressedPartial here? Isn't that what happens simply if there is no compression?

zarr/core.py
```diff
@@ -1879,13 +1899,21 @@ def _process_chunk(
...
                         for i, dim in enumerate(self.chunks)
                     ]
-                    cdata.read_part(start, nitems)
-                    chunk_partial = self._decode_chunk(
-                        cdata.buff,
-                        start=start,
-                        nitems=nitems,
-                        expected_shape=expected_shape,
-                    )
+                    if isinstance(cdata, UncompressedPartialReadBufferV3):
```
martindurant, line 1902: isinstance checks all around the place are unfortunate and make it hard to make new codecs that do something related but different. How would we make this a protocol?

zarr/core.py
```diff
@@ -1980,9 +2008,29 @@ def _chunk_getitems(self, lchunk_coords, lchunk_selection, out, lout_selection,
                 for ckey in ckeys
                 if ckey in self.chunk_store
             }
+        elif (
+            self._partial_decompress
+            and not self._compressor
+            and not fields
```
martindurant, line 2014: fields *are* allowed, and the final dtype might be object if one of the intermediate codecs turns bytes into objects (e.g., categorical). Here we are asserting that any compression must change the number of bytes, which might not be true.

zarr/core.py
```diff
@@ -1980,9 +2008,29 @@ def _chunk_getitems(self, lchunk_coords, lchunk_selection, out, lout_selection,
...
+                ckey: UncompressedPartialReadBufferV3(
+                    ckey, self.chunk_store, itemsize=self.itemsize
+                )
+                for ckey in ckeys
+                if ckey in self.chunk_store
+            }
         else:
             partial_read_decode = False
-            cdatas = self.chunk_store.getitems(ckeys, on_error="omit")
+            if not hasattr(self.chunk_store, "getitems"):
```
martindurant, line 2029: So implicitly get_partial works on multiple keys and *might* do them concurrently, but getitems always is concurrent and so preferred when there are ranges. What about just one of the keys having a range (well, typically it's the end ones, right?) and that's enough to put us on a different code path.

zarr/creation.py
```diff
@@ -21,7 +21,7 @@ def create(shape, chunks=True, dtype=None, compressor='default',
            overwrite=False, path=None, chunk_store=None, filters=None,
            cache_metadata=True, cache_attrs=True, read_only=False,
            object_codec=None, dimension_separator=None, write_empty_chunks=True,
-           *, zarr_version=None, **kwargs):
+           *, zarr_version=None, storage_transformers=(), **kwargs):
```
martindurant, line 24: All the other are None when not used, so suggest that here too

zarr/tests/test_storage_v3.py
```diff
@@ -88,6 +91,18 @@ def keys(self):
...
 
+class DummyStorageTransfomer(StorageTransformer):
+    TEST_CONSTANT = "test1234"
+
+    extension_uri = "https://purl.org/zarr/spec/storage_transformers/dummy/1.0"
+    valid_types = ["dummy_type"]
+
+    def __init__(self, _type, test_value) -> None:
+        super().__init__(_type)
+        assert test_value == self.TEST_CONSTANT
```
martindurant, line 102: An assert within a helper class, even if only used for tests, is a little unconventional

martindurant (commented):
I have given this a once over. I am confused about how we get an array to be recognised as partial read capable, how the path to partial reads is activated (it involved a lot of conditionals), and - critically - worry that this is done serially rather than concurrently with fsspec. The final point would be a massive performance penalty in many workflows.

zarr/core.py
martindurant: Also, we should take note that only the data is getting transformed, not the metadata

zarr/creation.py
jstriebel: Using a tuple instead was explicitly proposed in the comment in https://github.com/zarr-developers/zarr-python/pull/1096#pullrequestreview-1055940239, no strong opinion from my side.

zarr/core.py
jstriebel: That's indeed a very unfortunate spaghetti-code :spaghetti: , I simply took over the design of the previous partial reads, but it's just becoming too complex with more options. I'd love to do a proper refactoring of those code-paths, but not in this PR, to keep them reviewable.  I think it would help to have some of the logic in the compressor itself, but I don't have sth. specific planned out.

zarr/core.py
jstriebel: This if checks if we're in the code-path where we can not or should not use either `getitems` or `get_partial_values`, that's why all of those are checked. Then we use the `_chunk_getitem` code-path. Not super-happy about those code-paths, but I think cleaning those up is better left to a separate PR.

zarr/core.py
martindurant: fsspec's is the only one implemented, no? "By design" the method allows for concurrency.

jstriebel:
@martindurant Thanks a lot for the review! Sorry I'm answering so late, I was quite busy the last weeks, and will be on vacation the next two weeks, so I'll just get back to this PR afterwards. Please feel free to adapt anything also directly, e.g. having a PR against this one, since you are much more familiar with the internals of fsspec and have a great overview of the zarr code-base :+1:

jstriebel:
@joshmoore I just added a `ZARR_V3_SHARDING` flag for sharding usage, as discussed previously.

joshmoore:
@jstriebel : ok to leave you to push the update here? (If not, I'll dig in)

jstriebel:
> @jstriebel : ok to leave you to push the update here? (If not, I'll dig in)

Sure, done :+1:

.github/workflows/minimal.yml
```diff
@@ -24,6 +24,7 @@ jobs:
       shell: "bash -l {0}"
       env:
         ZARR_V3_EXPERIMENTAL_API: 1
+        ZARR_V3_SHARDING: 1
```
joshmoore, line 27: Sorry, review got caught up on trying to figure out whether or not a second environment variable was needed, and then whether or not we need some more general mechanism, etc. etc. I assume it's not an issue to introduce a new one, but moving forward, we should probably make this easier on ZEP developers.

zarr/core.py
joshmoore: @jstriebel: can you capture the outstanding refactorings in an issue (if you don't already have work prepped in a PR... 🙂 )

zarr/core.py
joshmoore: @jstriebel, any comment here?

joshmoore (approved):
One minor comment I've been struggling with that's a bit outwith this PR. Getting this merged and then I'll prep a 2.14.0 (including `ensure_bytes`) unless there are any concrete versioning suggestions.

zarr/core.py
jstriebel: > fields _are_ allowed

I think that's right, we might not need to exclude `fields` usage here. However, I'm wondering the same for `PartialReadBuffer`, why is it excluded there? I assumed the same reason applies now.

> the final dtype might be object if one of the intermediate codecs turns bytes into objects (e.g., categorical).

True, but is there any option to recognize that? Again, this is the same issue as with `PartialReadBuffer` above.

> Here we are asserting that any compression must change the number of bytes, which might not be true.

We use it as an exclusion criterum, yes, because most compressions do and again, there's currently no way to tell if it would change the number of bytes. Again, this is exactly the same as above with `PartialReadBuffer`.

I agree that there might be more cases where `PartialReadBuffer` could be used, but this solves it for the ones where I'm sure that it works.

zarr/core.py
jstriebel: Added issue #1329 for it

jstriebel:
> Getting this merged

:rocket:

> I'll prep a 2.14.0 (including `ensure_bytes`) unless there are any concrete versioning suggestions.

Seems fine to me :+1:

.github/workflows/minimal.yml
jstriebel: I felt that this was quite an easy change, but true, this could be simplified in the future (e.g. a single variable for all enabled ZEPs, and a special value to enable all, which is used in CI).

zarr/core.py
joshmoore: @martindurant: how do you see moving forward here? Do you think these issues (/changes in assumptions) can impact existing code?

martindurant: Existing code, no.

But we want to make something that is as generally useful as possible. The points here suggest to me that such decisions are poorly made at the array layer, and something only the codecs (including compressors) can know, if they are fed the information about what the request was.

jstriebel: Perhaps this should be improved in a follow-up PR, I don't think this should block the current one. If anything, this is purely improving the current state since it allows partial reads for uncompressed chunks at all.

jstriebel:
@joshmoore I don't see any further actionables for this PR, I'd be glad if we could merge it soon.

zarr/core.py
joshmoore: Thanks, both. Had this on my radar for discussion at the meeting and missed that it had happened on GitHub 🤦🏽

rabernat (approved):

joshmoore:
Yessirs.

rabernat:
This is huge! 🙌

jstriebel:
:tada:

PR #1131, opened 2022-09-12
Getitems: support `meta_array`

madsbk:
`BaseStore` now implements `getitems()`, which takes a `meta_array` argument. Besides simplifying the code, this enable stores to read directly to GPU memory: https://github.com/rapidsai/kvikio/pull/131

TODO:
* [x]  Add unit tests and/or doctests in docstrings


quasiben:
Is this the last PR needed by kvik for full Zarr support ?

madsbk:
> Is this the last PR needed by kvik for full Zarr support ?

Yes with this PR and https://github.com/rapidsai/kvikio/pull/131, all basic building blocks should be in place to use Zarr backed by CuPy arrays!

zarr/_storage/store.py
```diff
@@ -129,6 +131,32 @@ def _ensure_store(store: Any):
...
+            A collection mapping the input keys to their results.
+
+        Developer Notes
+        ---------------
+        This default implementation use __getitem__() to read each key sequential and
+        ignores the meta_array argument. Overwrite this method to implement concurrent
+        reads of multiple keys and/or to utilize the meta_array argument.
+        """
+
+        return {k: self[k] for k in keys if k in self}
```
jakirkham, line 158: Know fsspec-based storage layers are using a `dict` return value, but am wondering if we should be doing something different (like returning an iterable). Asking since this would make the read blocking vs. a bit more lazy. The latter can be useful when working with operations that take a bit longer (like reading from the cloud or parallelizing several reads)

cc @martindurant (who also may have thoughts)

jakirkham (commented):
Thanks Mads! This is very helpful 🙏

Asked one question above about API design. Curious to hear yours and others thoughts 🙂

jakirkham:
> > Is this the last PR needed by kvik for full Zarr support ?
> 
> 
> 
> Yes with this PR and https://github.com/rapidsai/kvikio/pull/131, all building blocks should be in place to use Zarr backed by CuPy arrays!

Just to clarify, this is a quite useful optimization (and much appreciated!). Though is it a requirement (in that an error was hit somewhere)? Is there a bit more context about how this came up (particularly if an issue was encountered)?

zarr/_storage/store.py
martindurant: The point of it being blocking, is that  many keys may be being fetched concurrently. If you make it an iterator, you lose that, unless you have something that can wait on an async iterator, which in turn has first-completed working.

madsbk:
> > > Is this the last PR needed by kvik for full Zarr support ?
> > 
> > 
> > Yes with this PR and [rapidsai/kvikio#131](https://github.com/rapidsai/kvikio/pull/131), all building blocks should be in place to use Zarr backed by CuPy arrays!
> 
> Just to clarify, this is a quite useful optimization (and much appreciated!). Though is it a requirement (in that an error was hit somewhere)? Is there a bit more context about how this came up (particularly if an issue was encountered)?

It is required in the sense that a backend doesn’t know the memory destination of a read without this PR. In KvikIO’s backend, we must rely on key names to infer if the data goes to host or device memory (see below). 

```python
  if os.path.basename(fn) in [
      zarr.storage.array_meta_key,
      zarr.storage.group_meta_key,
      zarr.storage.attrs_key,
  ]:
```

This is a problem since a `Store` can define its own special keys with arbitrary names.

zarr/_storage/store.py
madsbk: I agree with both of you.
The API specifies a return type of `Mapping[str, Any]` thus it is possible to return a _lazy_ mapping that only reads `self[k]` when accessed. But let's do that in a follow up PR?

jakirkham:
Thanks for clarifying. Do we need `setitems` as well?

madsbk:
> Thanks for clarifying. Do we need `setitems` as well?

I don't think it is as important as the `getitems` since a backend can always probe the input data and determine if it is host or device memory. 
However, `setitems` would make it possible to implement concurrent write, which could benefit backends such as KvikIO significantly.

zarr/_storage/store.py
martindurant: Understood, and will be interested in how that looks.

By the way, I mentioned elsewhere the possibility of passing key-specific metadata to the get function; that would happen in this same signature. I wonder if you have any use for an array where only *some* of it is destined for the GPU (perhaps because that data already exists there and doesn't need loading at all).

zarr/core.py
```diff
@@ -2008,8 +1969,8 @@ def _chunk_getitems(self, lchunk_coords, lchunk_selection, out, lout_selection,
                 if ckey in self.chunk_store
             }
         else:
-            partial_read_decode = False
-            cdatas = self.chunk_store.getitems(ckeys, on_error="omit")
+            cdatas = self.chunk_store.getitems(ckeys, meta_array=self._meta_array)
```
martindurant, line 1972: Since this is taking away code for the (previous) case that only some stores have `getitems`, can you please comment here with a link to the PR where this changed? Is it possible that there are old Store duck-types out there that lack this method, or do we now strictly require them to be subclasses of Store?

madsbk: AFAICT, all Stores should go through [`_ensure_store()`](https://github.com/zarr-developers/zarr-python/blob/main/zarr/_storage/store.py#L93) at some point.

jakirkham: Also worth noting the `partial_read_decode = False` was just moved above the `if`. So it is still there (just always done now)

jakirkham:
@joshmoore do you have thoughts on this one? 🙂

zarr/_storage/store.py
```diff
@@ -129,6 +131,32 @@ def _ensure_store(store: Any):
...
+            of the output is the same as calling __getitem__() for each key in keys.
+
+        Returns
+        -------
+        Mapping
+            A collection mapping the input keys to their results.
+
+        Developer Notes
+        ---------------
+        This default implementation use __getitem__() to read each key sequential and
```
joshmoore, line 153:
```suggestion
        This default implementation uses __getitem__() to read each key sequentially and
```

zarr/_storage/store.py
```diff
@@ -129,6 +131,32 @@ def _ensure_store(store: Any):
...
+    def getitems(self, keys: Sequence[str], meta_array: NDArrayLike) -> Mapping[str, Any]:
+        """Retrieve data from multiple keys.
+
+        Parameters
+        ----------
+        keys : Iterable[str]
+            The keys to retrieve
+        meta_array : array-like
+            An array instance to use for determining the output type. For now, this is
+            only a hint and can be ignore by the implementation, in which case the type
```
joshmoore, line 143: "for now...only a hint" -- what's the plan after this PR? Will there be compatibility issues?

zarr/_storage/store.py
joshmoore: Do you have an example of what they would look like, @martindurant?

zarr/core.py
```diff
@@ -1930,76 +1923,44 @@ def _process_chunk(
         # store selected data in output
         out[out_selection] = tmp
 
-    def _chunk_getitem(self, chunk_coords, chunk_selection, out, out_selection,
```
joshmoore, line 1933: Is this so private that no one could have made use of it in a subclass?

zarr/core.py
```diff
@@ -1257,20 +1257,13 @@ def _get_selection(self, indexer, out=None, fields=None):
         else:
             check_array_shape('out', out, out_shape)
 
-        # iterate over chunks
-        if not hasattr(self.chunk_store, "getitems") or \
```
joshmoore, line 1261: :heart: for removing `hasattr` hacks in general. :+1:

joshmoore (commented):
Some minor comments, @jakirkham. I do get the feeling that after the refactoring that enabled V3, we likely do need some storage-extension-specific documentation notes.

zarr/_storage/store.py
martindurant:
``` 
def getitems(self, keys: Sequence[str], meta_array: NDArrayLike, key_contexts: Mapping[str, Mapping]) -> Mapping[str, Any]:
```
?

Or did I misunderstand the question here?

zarr/_storage/store.py
madsbk: In the future, we might require stores to return an array of type `meta_array`, which would introduce compatibility issues. Should I can remove the `For now`?

zarr/core.py
madsbk: Yes, AFAICT.

zarr/_storage/store.py
madsbk: I don't know of such a use case. If it is a rare case, it might be sufficient to just divide the read into multiple calls to `getitems()` -- one for each return type.

martindurant: No, please don't - we would loose concurrency. Also, that would need getitems() to change instead, so no win. I will try to find the issue where I raise this.

martindurant: Like https://github.com/zarr-developers/zarr-python/pull/1096#issuecomment-1226045840

martindurant: In summary, we have a few different things that may want to pass to the codec/storage chain, each for a specialised case. Examples
- array type we want to end up with
- load many pieces concurrently
- locally-dependent codec parameters (even different compression per chunk)
- partial read of a chunk
- (perhaps) return a lazily or async evaluated thing
- probably others

Partial read is already implemented specifically for blosc, and it took a lot of hackery to fit it in the current code. Here we are adding another special case by peppering kwargs in the code. I am suggesting that it might make more sense to combine these pieces of specialised and optional information into some "context" which can be passed to both codecs and storage. 

Actually, this raises a confusion of mine with this PR: the storage layer never returns an array, it returns a buffer. Is this not the case for kivi or whatever this is directed at?

madsbk: I think a `contexts` is a very good idea. I will be happy to update this PR to implement a more general context approach. 

> Actually, this raises a confusion of mine with this PR: the storage layer never returns an array, it returns a buffer.

Good point, it should probably be a buffer-like type instead.

martindurant: > it should probably be a buffer-like type instead.

So GPU the flow would be: 
- tell the storage layer to produce a GPU buffer
- tell any codec to operate on the GPU and produce GPU buffers (decompression ) or array pieces (filters)
- tell zarr to use these pieces to fill a GPU array

Is it enough to have `memory="gpu"` or similar for all the stages? Will there be cases where more detail about the specific class is required?

jakirkham: Is the contexts something that is store specific, Martin? If so, wonder if the particular stores could track that information and one could just use [`collections.ChainMap`]( https://docs.python.org/3/library/collections.html#collections.ChainMap ) to combine them all into one store that has this info under-the-hood.

zarr/core.py
jakirkham: Yeah that's my understanding as well. Or at least by prefixing with `_` we have warned users this is an implementation detail (subject to change) that they shouldn't rely on.

zarr/_storage/store.py
martindurant: No, the context is something specific to each key, which a store (or codec) may choose to do something with, similar to `block_info` in a dask array.

jakirkham: Ok. Still trying to wrap my head around this idea (so apologies if this suggestion is naive). Would it help if supplied an `out`/`outs`/`outputs`/etc. argument that (optionally) provided the arrays to write into?

martindurant: If the type of the output was the only thing we cared about, that would do; but I am trying to motivate something more general. Interestingly, codecs do allow passing a buffer to write into, so it be reasonable to do the same for the storage layer, except that we won't usually know how big the buffer should be.

jakirkham: As `getitems` is called by index operations, they should know how big the resulting array will be (and may even have that array provided). So it should be possible to be fairly efficient with memory, which seems important alongside the ability to fill up those pieces asynchronously behind `getitems`.

Could you please share a bit more about the use case you have in mind?

martindurant: I listed some thoughts above, but two immediate concerns are:
- partial read of uncompressed or block-compressed chunks (includes the possibility of subchunks and other sharding style access)
- a codec whose behaviour is dependent on location in the array, e.g., having a scale factor/offset that is a function of the Z dimension of a 3d array

and then the concern immediately addressed by this PR: picking the array class we want to create.

jakirkham: With partial reads/writes, I think we should handle this as part of the sharding work either in PR ( https://github.com/zarr-developers/zarr-python/pull/1111 ) or PR ( https://github.com/zarr-developers/zarr-python/pull/1096 ). It is possible we want a different method (like `get_partial_items` and `set_partial_items`), but am not sure about that. Would want Jonathan's input since he has spent more time thinking about sharding APIs.

With location dependent codecs, know this has come up in a few meetings, but Idk if we have created an issue (am not finding one atm). Would suggest we start with a new issue and scope that out. It will most likely involve more changes than just the API here as Numcodecs would need to become aware of this somehow. Also not sure whether it interacts with sharding. So a fair bit to unpack here.

joshmoore: One quick thought on the signature: while introducing the `dimension_separator` element, I often had the feeling that my life would have been simplified if there were a (extensible) `Key` object rather than just a string. I wonder if this is a case where we might want to support `Union[str, Key]` rather than introducing a separate `contexts: Mapping[str, Mapping]`

zarr/_storage/store.py
joshmoore: Not sure I'm the best to say, but :+1: for thinking through those future compatibility issues.

zarr/core.py
joshmoore: @jakirkham: this pre-dates me, so I defer. But if it's not documented somewhere we might want to review if that holds across the board. For me, `_x` is typically valid for subclassing by developers, otherwise it would be a `__x`. (i.e. public, protected, private in Java-parlance)

zarr/_storage/store.py
martindurant: I am ambivalent about a Key object, but the above suggestion is almost exactly what I had said first. Whether the contexts should be a mapping or a sequence is not important to me. For now, the context will be empty or include the desired class, and we can agree on things that can be added into it.

zarr/core.py
jakirkham: It's [a Python thing generally]( https://stackoverflow.com/a/41761857 ) (not specific to Zarr).

zarr/_storage/store.py
jakirkham: As [noted above]( https://github.com/zarr-developers/zarr-python/pull/1131#discussion_r990402879 ), think this discussion gets hairy enough it should be broken out into an issue (likely two) and discussed separately.

madsbk: > As [noted above](https://github.com/zarr-developers/zarr-python/pull/1131#discussion_r990402879), think this discussion gets hairy enough it should be broken out into an issue (likely two) and discussed separately.

If we want to go the `contexts` route, we don't need a `meta_array` argument

madsbk:
Based on @martindurant idea, I have replaced the `meta_array` argument with the more generic `contexts` argument.  
When `_chunk_getitems()` calls `getitems()`, it now provides the `meta_array` (if non-default) to the store using the new `contexts` argument. 

Thoughts?

zarr/core.py
martindurant: Single underscore if by far the most common thing to do, to show intent rather than enforce any privateness (which double underscore doesn't do either).

joshmoore: @jakirkham: see also though https://stackoverflow.com/questions/41761645/when-to-use-one-or-two-underscore-in-python/41761857#comment99492184_41762806

zarr/_storage/store.py
joshmoore: Does this also imply that `getitem()` should have a `contexts` parameter?

madsbk: > Does this also imply that `getitem()` should have a `contexts` parameter?

Good point, yes `getitem()` should also take a  `contexts` parameter

martindurant:
Where should we go from here, @joshmoore @jakirkham >

joshmoore:
Sorry for dragging my feet. Something about this is making me nervous, but I haven't been able to put my finger on it. Perhaps it's the lack of an interface.

madsbk:
Agree, the lack of an interface/protocol is a concern. Since `contexts` doesn’t enforce anything, we are effectively saying that the client and store should negotiate their own protocol. This might become hard to maintain, if we do not clearly define a protocol for each context. 

To mitigate this, I have added `Context`, which defines the context as a `TypedDict`. All components that want to use `contexts` to exchange information, must define said information in `Context` or in a sub-class. Furthermore, all items are optional thus components must implement a default implementation in the case an item isn’t found in the context. 

 

Thoughts, do you think this will suffice?

martindurant:
Suggested things I would add to context, suiting my purposes and perhaps explaining why I wanted this in the first place. All of these are already known before calling getitem(s) or decode chunk, they will never be None.

- key string
- compressor/filters list
- whole final array shape and type
- selected slice
- slice within this chunk (this is actually calculated when processing the chunk, so I'm not sure it's immediately available without moving things around)

I think you can see how this information would allow for partial load from any storage layer that supports it, in combination with a compressor that supports it (uncompressed, blosc, zstd, maybe others); it also allows for a decoding step which is dependent on the position in the array (if it can be passed to codec.encode/decode too).

joshmoore:
Thanks for the list, @martindurant! Digging in a little bit to make sure I understand you:

* key string: (obvious)
* compressor/filters list: so this would allow _different_ compressors or it would allow you to remove them from processing to get back more similar to what is actually stored?
* whole final array shape and type: as in the static value for the current array? in which case we could likely do this for the user, no?
* selected slice and slice within this chunk: These I'm less sure about. So when the implementation is calculating which key has been asked for, if it knows that less than a full chunk is needed, it passes that along?

martindurant:
All of the information in context if supposed to be static, to allow decisions to be take in the load or codec. The meta-array is the only one so far that the user can have a say in (but it is fixed throughout the call to zarr).

> compressor/filters list: so this would allow different compressors or it would allow you to remove them from processing to get back more similar to what is actually stored?

No, this is just the list from the array metadata, read-only. It allows the equivalent of all the `if compressor == "blosc"` stuff currently peppered around the code to allow block-partial reading.

> selected slice and slice within this chunk: These I'm less sure about. So when the implementation is calculating which key has been asked for, if it knows that less than a full chunk is needed, it passes that along?

Yes, in `zarr.Array._chunk_getitems` we have the chunk_select (which part of the logical chunk we need) and out_select (where does this go in the final array), but they are not passed to the loader or codecs as things stand. See the gnarly code around blosc artials int he same function.

jakirkham:
Share Josh's concern about a protocol.

If we go with the context route, think Mads' proposal might be a reasonable solution (need to think more about this). Though as there will inevitably changes there, would suggest we add a version key-value pair as well. This way we can confirm what is supported.

madsbk:
I think it would be good if we could get some progress here. 
Do we want:
- a `Context` argument as defined in https://github.com/zarr-developers/zarr-python/pull/1131#issuecomment-1303174124 or similar.
- a `meta_array` argument that corresponds to the `meta_array` argument used throughout the code. 
- something else?

joshmoore:
Chatted with @jakirkham, @jbms, and @MSanKeys963 about the context here this evening. We were unsure of the (multiple? overlapping?) use cases. Would it be possible to get some details on them here or discuss at the next community meeting?

- What features are specific to the GPU use case?
- What is fsspec specific?
- Sharding specific?
- Does it need to touch zarr core?
- Do the codecs need to be "aware" and if so, why?
- Would/could @jstriebel's transformers in #1096 provide a way to "capture" the multiple gets and serialize them into a single get?
- If so, does the transformer extension need any additional metadata or features?

Basically, who needs to know what where when and why? 🙂

martindurant:
I will try...

> What features are specific to the GPU use case?

The array_meta itself: what kind of instance will the incoming data be put into. This was originally a standalone kwargs, but could be in the context.

> What is fsspec specific?

As far as existing FSStore workflows are concerned, nothing. However, passing a context would allow codec-level and storage-level per-chunk operations, something kerchunk is interested in, since it often is making a logical dataset out of inputs which do not have homogenous nature.

> Sharding specific?

Passing the part of a desired chunk needed to the loader means that only required shards need be loaded. This would isolate the logic much more cleanly in the IO machinery than having the specific shard- and blosc blocks of code in the higher levels as we have now.

> Does it need to touch zarr core?

It is needed in the implementation. It is not a spec issue, the contract between zarr and the storage classes (and indeed codecs) is I think only described in code. 

> Do the codecs need to be "aware" and if so, why?

In order for the codecs to be able to make use of the context (e.g., position-dependent parameters), it need to be passed to them. I imagine this as an optional kwarg. All codecs should ideally have this in their encode/decode signatures, but currently existing ones would just ignore it. We can use inspect to determine if the kwargs can be passed for compatibility.

> Would/could @jstriebel's transformers in https://github.com/zarr-developers/zarr-python/pull/1096 provide a way to "capture" the multiple gets and serialize them into a single get?
> If so, does the transformer extension need any additional metadata or features?

I don't know. In some ways, the context is a simpler way to do partial get/sets, but nothing here approaches a general transform.
Zarr being the flexible and simple thing it is, it would be possible to achieve some things I am after (chunk-specific ops, irregular chunks, shards) in the storage layer alone and not touch the spec/impl for those. The exception is partial chunks, since the codec and storage has no way of knowing when a key is requested, whether all of that key is needed or not.

zarr/_storage/store.py
```diff
@@ -129,6 +130,34 @@ def _ensure_store(store: Any):
...
+        """Retrieve data from multiple keys.
+
+        Parameters
+        ----------
+        keys : Iterable[str]
+            The keys to retrieve
+        contexts: Mapping[str, Context]
+            A mapping of keys to their context. Each context is a mapping of store
+            specific information. E.g. a context could be a dict telling the store
+            the preferred output array type: `{"meta_array": cupy.empty(())}`
```
rabernat, line 145: I don't understand why the context needs to be key specific. If a user wants to load data into the GPU, surely they want to do that for the entire read, not just some chunks. It seems like we could simplify this a lot by just providing one single context to `getitems`, rather than a mapping.

There is also likely a performance cost to doing it this way, no? If you're reading 1000s of chunks at a time, it seems silly to be carrying around all these big mappings to the same context.

rabernat (approved):
I just reviewed this PR and associated discussion.

The main motivating goal of this work is to make it possible to read data data directly to GPU memory. That is a clearly defined objective.

My impression is that the PR has stalled due suggestions from reviewers about potential other functionality that might eventually be supported by the "context" concept. This is coupled together with other semi-vague ideas about how to improve / refactor the ways different classes in zarr-python communicate with each other (also related to #1329).

We need to unblock @madsbk, who is clearly motivated to integrate this into other exciting projects. Therefore, I recommend we limit the scope creep and focus on the original goal.

So I would be happy to see this PR go in as is, and defer questions about how to refactor zarr-python for future work. 

I do have one suggestion about a potential simplification. I really don't understand the use case for key-specific context.

zarr/_storage/store.py
madsbk: Agree, this is to support the generic _context_ approach. Although, it is possible to limit the performance impact by using a custom `Mapping` that returns the same value for all keys.

Also notice, the current code implements the context approach where the `meta_array` argument is embedded within the context argument. We could rollback to my original approach where `getitems()` takes the `meta_array` directly (`getitems(..., meta_array:ndarray`)

rabernat:
@joshmoore & @jakirkham - any objections to merging this as-is and deferring the broader discussions for future work?

joshmoore:
Thanks for the ping, @rabernat. I have to say I'm liking the typed `Context` better. I will read through again ASAP.

madsbk:
@joshmoore, how can I help progress here? Maybe an online meeting?

joshmoore:
Sorry, @madsbk. My last comment was shortly before leaving on several weeks of travel and I lost of track of this. It wasn’t my intention to hold you up, and big :+1: for the ping. A few thoughts (from smallest to biggest) though I get the feeling that you are being ping-ponged back and forth between design issues:

 * If we're less than confident on this API, perhaps we should make this a keyword-only argument. That way the second position could potentially be used for something else and deprecation/renaming would be more straight-forward.
 * Unlike some of the context that was going to get passed, I have the feeling that the `meta_array` is more likely to be used for all chunks equally. Does that sound right? If so, I'm a bit hesitant to force someone to create dictionaries of N objects all with the same `meta_array` instance wrapped in new `Context` object. As I started to think about "wildcard" keys in the `context` argument or an additional `default_context` argument, I realized that perhaps we've not gotten this abstract right yet.
 * Finally, I've still not got a clear picture of what other fields on the `Context` object are going to help with the likes of partial read. Obviously, that's less to do your PR but just to callback to the original discussions about whether or not we knew what the API should look like.

madsbk:
> If we're less than confident on this API, perhaps we should make this a keyword-only argument. That way the second position could potentially be used for something else and deprecation/renaming would be more straight-forward.

Agree, fixed.

> Unlike some of the context that was going to get passed, I have the feeling that the meta_array is more likely to be used for all chunks equally. Does that sound right? If so, I'm a bit hesitant to force someone to create dictionaries of N objects all with the same meta_array instance wrapped in new Context object. As I started to think about "wildcard" keys in the context argument or an additional default_context argument, I realized that perhaps we've not gotten this abstract right yet.

I have added [`ConstantMap`](https://github.com/zarr-developers/zarr-python/pull/1131/commits/c3ee95fc9f7200f608d220224a7b0aae1650977a), which is a read-only map that maps all keys to the same constant value. This fixes the issue of having to create a `Context` for each key without complicating the semantic of `getitems()`.

> Finally, I've still not got a clear picture of what other fields on the Context object are going to help with the likes of partial read. Obviously, that's less to do your PR but just to callback to the original discussions about whether or not we knew what the API should look like.

I don't think any of us have a clear answer to this. Therefore, I suggest that we merge this PR and let people explore its possibilities. As long as `Context` is optional and we force `Stores` to implement a fallback if it doesn't exist, I think we are fine.

madsbk:
Thanks all!

rabernat:
Thank you for your patience Mads.

PR #1135, opened 2022-09-13
Fix numcodecs dependency

normanrz:
`zarr` actually depends on `numcodecs>=0.10`, because it [imports `numcodecs.compat.ensure_ndarray_like`](https://github.com/zarr-developers/zarr-python/blob/main/zarr/util.py#L13) that has been introduced in `0.10.0`.

TODO:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -10,7 +10,7 @@ dependencies = [
     'asciitree',
     'numpy>=1.7',
     'fasteners',
-    'numcodecs>=0.6.4',
+    'numcodecs>=0.10.0',
 ]
 
 setup(


joshmoore:
Thanks, @normanrz! This is if pip installing with `--pre` or directly from `main`?


cc: @madsbk

normanrz:
We pinned `numcodecs==0.9.1`, which actually shouldn't be allowed (because of the aforementioned import), but is permitted by zarr's `setup.py`.

joshmoore:
Merging to that the main branch has sanity. I find it a bit unfortunate to lose backwards compatibility with the previous versions of numcodecs. If 0.9 and before were numcodecs 1.x, then I think @madsbk's recent changes might have pushed us to 2.x.

PR #1149, opened 2022-09-27
fastpath _array_iter for array_keys

jhamman:
This PR modifies the v3 implementation of `Group._array_iter`. Previous behavior unnecessarily called `contains_array` for keys known to be arrays (i.e. keys that ended in `.array.json`). These changes fast track the iterator to return v3 array keys. In practice, this reduces the number of times a key is searched for in a store.

TODO:

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -75,8 +75,8 @@ The Zarr source code is hosted on GitHub at the following location:
 You will need your own fork to work on the code. Go to the link above and hit
 the "Fork" button. Then clone your fork to your local machine::
 
-    $ git clone git@github.com:your-user-name/zarr.git
-    $ cd zarr
+    $ git clone git@github.com:your-user-name/zarr-python.git
+    $ cd zarr-python
     $ git remote add upstream git@github.com:zarr-developers/zarr-python.git
 
 Creating a development environment
--- a/zarr/hierarchy.py
+++ b/zarr/hierarchy.py
@@ -638,17 +638,19 @@ class Group(MutableMapping):
         else:
             dir_name = meta_root + self._path
             array_sfx = '.array' + self._metadata_key_suffix
+            group_sfx = '.group' + self._metadata_key_suffix
+
             for key in sorted(listdir(self._store, dir_name)):
                 if key.endswith(array_sfx):
                     key = key[:-len(array_sfx)]
-                path = self._key_prefix + key
-                assert not path.startswith("meta/")
-                if key.endswith('.group' + self._metadata_key_suffix):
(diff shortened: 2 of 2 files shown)


zarr/hierarchy.py
```diff
@@ -638,17 +638,19 @@ def _array_iter(self, keys_only, method, recurse):
...
+
                 path = self._key_prefix + key
                 assert not path.startswith("meta/")
-                if key.endswith('.group' + self._metadata_key_suffix):
+                if key.endswith(group_sfx):
                     # skip group metadata keys
                     continue
-                if contains_array(self._store, path):
-                    _key = key.rstrip("/")
-                    yield _key if keys_only else (_key, self[key])
```
rabernat, line 651: Joe and I were looking at this code together, and we couldn't understand the circumstances under which it would be necessary to call `contains_array` here. AFAICT, the V3 spec makes this unnecessary; if you've got an array metadata document, then there is an array at that path. It feels like something leftover from the V2 implementation.

Are we missing something?

docs/contributing.rst
```diff
@@ -75,8 +75,8 @@ The Zarr source code is hosted on GitHub at the following location:
 You will need your own fork to work on the code. Go to the link above and hit
 the "Fork" button. Then clone your fork to your local machine::
 
-    $ git clone git@github.com:your-user-name/zarr.git
-    $ cd zarr
+    $ git clone git@github.com:your-user-name/zarr-python.git
```
joshmoore, line 78: :heart:

zarr/hierarchy.py
joshmoore: Reading the change & explanation again, it makes sense to me. I imagine it's safe to get in with the rest of the work that will be going on on the V3 side.

docs/contributing.rst
joshmoore: Also found in https://github.com/zarr-developers/zarr-python/pull/1171

zarr/hierarchy.py
jhamman: @grlee77 - do you have thoughts on this change?

jhamman:
@joshmoore - can you expand on what you mean here: 

> Reading the change & explanation again, it makes sense to me. I imagine it's safe to get in with the rest of the work that will be going on on the V3 side.

Are you suggesting this should be done in a separate PR?

joshmoore:
> Are you suggesting this should be done in a separate PR?

Sorry, @jhamman. I mean that V3 is in-flux enough it's going to be hammered on enough outside of this PR. Merging.

Issue #843, opened 2021-09-30
Poor performance when slicing with steps


cgohlke:
#### Minimal, reproducible code sample, a copy-pastable example if possible

```python
"""Demonstrate low performance of slicing zarr array with step size."""

import sys
import json
import time
from collections.abc import MutableMapping

import numpy
import zarr
import dask
import dask.array

shape = (149568, 66369, 3)  # 27.7G RGB image
chunks = (1, 66369, 3)  # one scanline
dtype = '|u1'
step = 512
repeat = 2
counter = 0  # global counter, not thread-safe

class ZarrStore(MutableMapping):
    def __init__(self, shape, chunks, dtype):
        self.chunk = numpy.zeros(chunks, dtype)
        self.store = {}
        self.store['.zattrs'] = json_dump({})
        self.store['.zarray'] = json_dump(
            {
                'zarr_format': 2,
                'shape': shape,
                'chunks': chunks,
                'dtype': dtype,
                'fill_value': 0,
                'order': 'C',
                'compressor': None,
                'filters': None,
            }
        )

    def flush(self):
        raise PermissionError

    def clear(self):
        raise PermissionError

    def keys(self):
        return self.store.keys()

    def items(self):
        return self.store.items()

    def values(self):
        return self.store.values()

    def __iter__(self):
        return iter(self.store.keys())

    def __len__(self):
        return len(self.store)

    def __delitem__(self, key):
        raise PermissionError

    def __setitem__(self, key, value):
        raise PermissionError

    def __contains__(self, key):
        return key in self.store

    def __getitem__(self, key):
        global counter
        if key in self.store:
            return self.store[key]
        counter += 1
        return self.chunk


def json_dump(obj):
    return json.dumps(
        obj,
        indent=1,
        sort_keys=True,
        ensure_ascii=True,
        separators=(',', ': '),
    ).encode('ascii')


if 1:
    store = ZarrStore(shape, chunks, dtype)
    z = zarr.open(store, mode='r')
else:
    # use zarr MemoryStore
    z = zarr.zeros(
        shape=shape,
        chunks=chunks,
        dtype=dtype,
        compressor=None,
    )

print('Python', sys.version)
print('Platform', sys.platform)
print('numpy', numpy.__version__)
print('zarr', zarr.__version__)
print('dask', dask.__version__)
print()
print(z.info)

print('z[:]  # slice all chunks')
for _ in range(repeat):
    counter = 0
    start = time.perf_counter()
[...]

PR #1154, opened 2022-10-01
Omit chunks with no elements in slice selection with step

jrs65:
This stops chunks being read unnecessarily when a slice selection with a step was used (reported in #843). 

Previously all chunks spanning the start-end range would be used regardless of whether they contained any elements. The changes in this PR are a small modification to `zarr.indexing.SliceDimIndexer` to stop it yielding empty selections up to `zarr.Array._get_selection`. Previously the chunk would be read and have no elements extracted from it, this stops it ever looking at the empty chunk.

I'm unsure about if a unit test should be added for this. It would be feasible to have a test asserting that only the minimal set of chunks are read, but that feels like it's testing an implementation detail. I'm happy to construct a test if that would be useful. Regardless I'll add some test code in a comment that succinctly shows the issue and the fix.

TODO:

### DIFF
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -14,6 +14,15 @@ Release notes
     # .. warning::
     #    Pre-release! Use :command:`pip install --pre zarr` to evaluate this release.
 
+.. _release_2.13.3:
+
+2.13.3
+------
+
+* Improve performance of slice selections with steps by omitting chunks with no relevant
+  data.
+  By :user:`J. Richard Moore <jrs65>` :issue:`843`.
+
 .. _release_2.13.2:
 
 2.13.2
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -21,7 +21,7 @@ Release notes
 
 * Improve performance of slice selections with steps by omitting chunks with no relevant
   data.
-  By :user:`J. Richard Moore <jrs65>` :issue:`843`.
+  By :user:`Richard Shaw <jrs65>` :issue:`843`.
 
 .. _release_2.13.2:
 
--- a/zarr/indexing.py
+++ b/zarr/indexing.py
@@ -216,6 +216,11 @@ class SliceDimIndexer:
             dim_chunk_sel = slice(dim_chunk_sel_start, dim_chunk_sel_stop, self.step)
             dim_chunk_nitems = ceildiv((dim_chunk_sel_stop - dim_chunk_sel_start),
                                        self.step)
+
+            # If there are no elements on the selection within this chunk, then skip
+            if dim_chunk_nitems == 0:
+                continue
+
             dim_out_sel = slice(dim_out_offset, dim_out_offset + dim_chunk_nitems)
 
             yield ChunkDimProjection(dim_chunk_ix, dim_chunk_sel, dim_out_sel)
--- a/zarr/tests/test_indexing.py
+++ b/zarr/tests/test_indexing.py
@@ -13,6 +13,8 @@ from zarr.indexing import (
     PartialChunkIterator,
 )
 
+from zarr.tests.util import CountingDict
+
 
 def test_normalize_integer_selection():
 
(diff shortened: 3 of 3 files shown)


jrs65:
Here's a short demonstration of this:
```python
import numpy as np
import zarr


def track_chunk_calls(z: zarr.Array):
    """Wrap a zarr.Array such that it prints out the details of the requested chunk."""

    f = z._chunk_getitem

    def _track(chunk_coords, chunk_sel, out, out_sel, **kwargs):
        print(f"{chunk_coords=}, {chunk_sel=}, {out_sel=}")
        return f(chunk_coords, chunk_sel, out, out_sel, **kwargs)

    z._chunk_getitem = _track


# Create a chunked array
N = 50
C = 5
z = zarr.ones((N,), chunks=(C,))

# Define a sparse slice selection
sl = slice(4, 41, 12)

# Calculate the chunks the requested elements are in
chunk_inds = np.arange(N)[sl] // C
print(f"## Chunks that must be read: {chunk_inds}")

print("## Chunks that are actually read:")
track_chunk_calls(z)
z[sl]
```

Prior to the fix in this PR, this gives the output:
```
## Chunks that must be read: [0 3 5 8]
## Chunks that are actually read:
chunk_coords=(0,), chunk_sel=(slice(4, 5, 12),), out_sel=(slice(0, 1, None),)
chunk_coords=(1,), chunk_sel=(slice(11, 5, 12),), out_sel=(slice(1, 1, None),)
chunk_coords=(2,), chunk_sel=(slice(6, 5, 12),), out_sel=(slice(1, 1, None),)
chunk_coords=(3,), chunk_sel=(slice(1, 5, 12),), out_sel=(slice(1, 2, None),)
chunk_coords=(4,), chunk_sel=(slice(8, 5, 12),), out_sel=(slice(2, 2, None),)
chunk_coords=(5,), chunk_sel=(slice(3, 5, 12),), out_sel=(slice(2, 3, None),)
chunk_coords=(6,), chunk_sel=(slice(10, 5, 12),), out_sel=(slice(3, 3, None),)
chunk_coords=(7,), chunk_sel=(slice(5, 5, 12),), out_sel=(slice(3, 3, None),)
chunk_coords=(8,), chunk_sel=(slice(0, 1, 12),), out_sel=(slice(3, 4, None),)
```
demonstrating that it is reading all the chunks spanning the range, not just the minimal set of `[0, 3, 5, 8]`.

After applying the fix only the minimal chunks are actually read:
```
## Chunks that must be read: [0 3 5 8]
## Chunks that are actually read:
chunk_coords=(0,), chunk_sel=(slice(4, 5, 12),), out_sel=(slice(0, 1, None),)
chunk_coords=(3,), chunk_sel=(slice(1, 5, 12),), out_sel=(slice(1, 2, None),)
chunk_coords=(5,), chunk_sel=(slice(3, 5, 12),), out_sel=(slice(2, 3, None),)
chunk_coords=(8,), chunk_sel=(slice(0, 1, 12),), out_sel=(slice(3, 4, None),)
```

rabernat:
Thanks so much for this PR @jrs65! It looks like a simple solution with a clear improvement.

> It would be feasible to have a test asserting that only the minimal set of chunks are read, but that feels like it's testing an implementation detail.

I'm not sure it's an implementation detail. Not accessing data unnecessarily is an important feature, particularly when the access might be very expensive (think cloud storage, large chunks, etc.). From my point of view, it would be great to have tests for this sort of thing.

On thing we do in Xarray's test suite is to define a [special kind of store](https://github.com/pydata/xarray/blob/513ee34f16cc8f9250a72952e33bf9b4c95d33d1/xarray/tests/test_dataset.py#L201) which errors if you try to access any of its data. This allows us to guard against undesirable data access. We could take a similar approach here in Zarr.

jrs65:
> Thanks so much for this PR @jrs65! It looks like a simple solution with a clear improvement.

> I'm not sure it's an implementation detail. Not accessing data unnecessarily is an important feature, particularly when the access might be very expensive (think cloud storage, large chunks, etc.). From my point of view, it would be great to have tests for this sort of thing.
> 
> On thing we do in Xarray's test suite is to define a [special kind of store](https://github.com/pydata/xarray/blob/513ee34f16cc8f9250a72952e33bf9b4c95d33d1/xarray/tests/test_dataset.py#L201) which errors if you try to access any of its data. This allows us to guard against undesirable data access. We could take a similar approach here in Zarr.

Great. That's along the lines I was thinking of anyway, so I will put together a set of unit tests along those lines, and update this PR with them. I'll add them to `zarr/tests/test_indexing.py` unless you have another suggestion.

rabernat:
That sounds great! Thanks so much for volunteering to take that on. It's a lot more work than just fixing the bug! 🙏  We welcome your contributions to zarr-python and will do our best to help with your PR. 🙏

jrs65:
@rabernat very happy to do so! Compared to hacking around in the internals of HDF5 and h5py, the `zarr` codebase is a joy to understand and modify.

I've implemented tests for this case. In the end I used the existing `zarr.tests.util.CountingDict` implementation rather than reinventing the wheels. I use that as a backing store, and then after each operation I check that the expected chunks have been accessed once and that no other chunks have been touched.

Let me know what you think!

joshmoore:
@jrs65: I'll second @rabernat's thanks! I've triggered the GHA actions. Let's see how the tests far. (Also looking forward to what the [asv](https://github.com/airspeed-velocity/asv) benchmarks look like before & after. cc: @MSanKeys963)

joshmoore:
Tests are all green! 🎉 @jrs65, do you want to add yourself to docs/release.rst?

Otherwise, unless there are objections, I could see getting this out for a quick patch release.

jakirkham:
Agree this is a nice improvement. Thank you! 🙏

Are there other indexers that would benefit from the same kind of fix?

jrs65:
> Tests are all green! 🎉 @jrs65, do you want to add yourself to docs/release.rst?

Sure! I'll add it down in a provisional 2.13.3 patch release, presuming you'll edit that around when you decide when to get it out.

jrs65:
> Agree this is a nice improvement. Thank you! 🙏
> 
> Are there other indexers that would benefit from the same kind of fix?

I've just read over the code for the other indexers. `BoolArrayDimIndexer`, `IntArrayDimIndexer`, `CoordinateIndexer`, and `MaskIndexer` (a subclass of `CoordinateIndexer`) look like they explicitly calculate which chunks contain the requested elements and then ignores those that don't. So I believe those are all fine.

I'm unsure about `OrthogonalIndexer` it does not seem to explicitly do that, so I'd need to read in more detail.

Regardless, I think it would be very straightforward to extend the tests to ensure that going forward the other selection types continue to only access what they need. If I find a bit of time to spare in the next week I might give that ago and see what happens.

jakirkham:
Adding to 2.13.3 sounds good. Thanks! 🙏

Also thanks for taking the time to look into other indexers. No worries if you don't have time to look further. Figured this was a good opportunity to revisit since we are already looking at performance here. Though happy to spin off into a new issue for a future date if that's preferable 🙂

jrs65:
Errrmm... ooops. Sorry. I typo'd my own name in the Release Notes 😆  I'm just going to push a quick fix to that one.

jakirkham:
No worries. Thanks again Richard! 🙏

Planning on merging after CI passes.

joshmoore:
Re-launched the job after an unrelated failure. I'll keep an eye on it for release.

jakirkham:
Thanks Josh! 🙏 Looks like it passed 🟢

jakirkham (approved):

joshmoore:
Thanks, @jakirkham. I created the GH release. :+1:

PR #1182, opened 2022-10-13
added license to docs and updated spec_v3

### LINKED ISSUE #1180: Add License to [Zarr documentation](https://zarr.readthedocs.io/)
I think it would be good to have the [LICENSE](https://github.com/zarr-developers/zarr-python/blob/main/LICENSE) added as a page on the [documentation](https://zarr.readthedocs.io/). So it can appear on the side panel.

### LINKED ISSUE #1181: Duplicate label spec_v2 in /docs/spec/v3.rst and /docs/spec/v2.rst
Command
```
tox -e docs
```
ends in an error.
 
I think the label in /docs/spec/v3.rst should be 
```
.. _spec_v3:
```
and not 
```
.. _spec_v2:
```

DON-BRAN:
fixes #1180
fixes #1181

TODO:

### DIFF
--- a/docs/license.rst
+++ b/docs/license.rst
@@ -1,0 +1,24 @@
+License
+=======
+
+The MIT License (MIT)
+
+Copyright (c) 2015-2022 Zarr Developers <https://github.com/zarr-developers>
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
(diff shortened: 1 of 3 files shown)


docs/license.rst
```diff
@@ -0,0 +1,24 @@
+License
+=======
+
+The MIT License (MIT)
```
jakirkham, line 4: Is there some way to include [this license file]( https://github.com/zarr-developers/zarr-python/blob/main/LICENSE ) without copying it?

docs/spec/v3.rst
```diff
@@ -1,4 +1,4 @@
-.. _spec_v2:
+.. _spec_v3:
```
jakirkham, line 1: This change looks ok to me

cc @joshmoore (in case you have thoughts)

jakirkham (commented):
Thanks! 🙏

Had a couple comments below

docs/license.rst
jakirkham: Ah sorry, let me clarify. Is there a way with Sphinx to tell it to load the content from `LICENSE` as opposed to creating a second copy of it here? IOW would it be possible to drop this `rst` file somehow?

jakirkham: No worries. Can see how that might be unclear.

Maybe this [SO question is helpful]( https://stackoverflow.com/q/2921724 )? We could also add a symlink if being outside `docs` is an issue

DON-BRAN: from https://www.sphinx-doc.org/en/master/usage/markdown.html
I think `myst-parser`has to be installed as well

DON-BRAN: Good morning @jakirkham. I wasn't able to completely remove the license.rst file but i could link it to the original LICENSE file without copying the content.
I referenced this issue https://github.com/sphinx-doc/sphinx/issues/7000

jakirkham: Actually this is exactly what I was looking for. Thank you! 😄

docs/spec/v3.rst
jakirkham: Sounds like this is good

xref: https://github.com/zarr-developers/zarr-python/issues/1181#issuecomment-1277444650

jakirkham (approved):
LGTM. Once CI passes this should be in. Thanks @DON-BRAN! 🙏

Issue #828, opened 2021-09-08
Implementing `ZipStore`'s `__delitem__` via overwrite


jakirkham:
Currently [`ZipStore`'s `__delitem__` `raises` `NotImplementedError`]( https://github.com/zarr-developers/zarr-python/blob/4e98567761cb557a76d9f05ab7cae6cc13f89266/zarr/storage.py#L1513-L1514 ), which is a bit awkward to workaround at higher levels since `__delitem__` is usually implemented in other stores. The reason this isn't implemented in `ZipStore` is zip files lack the ability to delete members.

An alternative approach that might be worth considering is overwriting the existing members with an empty string (`b""`). We can then treat them as deleted in Zarr.

jakirkham:
Sure, please do Emmanuel 🙂

jakirkham:
cc @MSanKeys963 @joshmoore (for awareness)

KarlieKKY:
> I'm an outreachy applicant, can I contribute to this project?

Hello, i'm also an outcheachy applicant, would you mind if we work on this issue together?

jakirkham:
Thanks for your interest! 😀

Emmanuel opened PR ( https://github.com/zarr-developers/zarr-python/pull/1184 ) already.

Would suggest picking a different issue 🙂

jakirkham:
Also worth noting there is an upstream issue about adding this functionality in CPython ( https://github.com/python/cpython/issues/51067 )

mohammadZuherJaserAsad:
Hi! I'm a student in the CodePath AI301 program and I'd like to work on this issue. I'll start by reproducing the current `NotImplementedError` behavior, reviewing the two previous stale PRs (#1184 and #2838) to understand what was attempted, and then implement the soft-delete approach for the v3 `ZipStore` — overwriting entries with `b""` and updating `__contains__`, `__getitem__`, and `keylist()` to treat them as deleted. Happy to ask questions if I run into anything unclear!

mkitti:
https://github.com/python/cpython/pull/134627 provides a `zipfile.remove()` which removes the zip entry in the central directory. The zipremove package backports the feature:
https://github.com/danny0838/zipremove

PR #1184, opened 2022-10-13
zipstore delitems via override

I-am-Emmanuel:
Fixes https://github.com/zarr-developers/zarr-python/issues/828

TODO:


I-am-Emmanuel:
I want to know if this is actually expected of me to do. Please if it is not give me a details explanation of what to do.

zarr/storage.py
```diff
@@ -1779,7 +1779,13 @@ def __setitem__(self, key, value):
             self.zf.writestr(keyinfo, value)
 
     def __delitem__(self, key):
-        raise NotImplementedError
+        # raise NotImplementedError
+        try:
+            value = self[key]
+        except KeyError:
+            raise KeyError("Can not delete a nonexistent key")
```
jakirkham, line 1786: Would suggest just checking if we have `key` (instead of reading it in). That way if it is a large value, we don't pay the cost of reading. `__contains__` has [some useful code]( https://github.com/zarr-developers/zarr-python/blob/2cfee9c0f0b2e0782f942ff0dd8681ecfb756ca0/zarr/storage.py#L1808-L1809 ) that we may borrow here.

zarr/storage.py
```diff
@@ -1779,7 +1779,13 @@ def __setitem__(self, key, value):
...
+        
+        self[key] = b""
```
jakirkham, line 1788: Think we need to [hold a mutex]( https://github.com/zarr-developers/zarr-python/blob/2cfee9c0f0b2e0782f942ff0dd8681ecfb756ca0/zarr/storage.py#L1766 ) when writing. Otherwise yes, this seems conceptually correct.

jakirkham (commented):
Thanks Emmanuel! 🙏

Generally this seems like the right approach. Thanks for diving in 😄

Had a couple comments below on improvements we might want to make.

The next thing would be to start updating some tests that currently are not running (like [this one]( https://github.com/zarr-developers/zarr-python/blob/2cfee9c0f0b2e0782f942ff0dd8681ecfb756ca0/zarr/tests/test_storage.py#L1809 )), but should now be able to with this change.

Please let us know if you have any questions 🙂

I-am-Emmanuel:
That's incredible. I'm not sure whether that is what is expected of me. I'll continue with the testing. Thanks

jakirkham:
I think it should be doable and believe you are up to the task. I hope I've given enough info. As always though, please feel free to ask questions. Any question that occurs to you is worth asking 🙂

I-am-Emmanuel:
Oh, just seeing this. I'm on it.

zarr/storage.py
I-am-Emmanuel: This is because I believe the set item method already handles the writing method. Is it really necessary to repeat the mutex?

jakirkham: That's fair. Though this means we are acquiring the mutex twice. Once to check for the `key` in the `if` statement above and once to set the value. Maybe we can copy the code from those methods here and tidy it up so we acquire the mutex only once. What do you think?

zarr/storage.py
```diff
@@ -1779,8 +1779,14 @@ def __setitem__(self, key, value):
...
-        raise NotImplementedError
-
+        # raise NotImplementedError
+        if key not in self:
+            raise KeyError("Can not delete a nonexistent key")
+        
+        self[key] = b""
+
+    def pop(self, key):
+        del self
```
jakirkham, line 1789: This is [implemented in the `MutableMapping` base class]( https://github.com/python/cpython/blob/a834e2d8e1230c17193c19b425e83e0bf736179e/Lib/_collections_abc.py#L933-L945 ). So we can skip implementing it here

```suggestion
```

zarr/tests/test_storage.py
```diff
@@ -1802,12 +1802,24 @@ def test_context_manager(self):
...
-            store.pop(self.root + 'foo')
+
+        store.pop(self.root + 'foo')
+
+        assert store[self.root + 'foo'] == b""
+        # with pytest.raises(NotImplementedError):
+        #     store.pop(self.root + 'foo')
 
     def test_popitem(self):
         # override because not implemented
```
jakirkham, line 1825: We can update this test to run as well (and drop `NotImplementedError` from the test)

zarr/tests/test_storage.py
```diff
@@ -1802,12 +1802,24 @@ def test_context_manager(self):
...
             assert 2 == len(store)
 
+    def test_delitem(self):
+        store = self.create_store()
+        store[self.root + 'foo'] = b'bar'
+
+        del store[self.root + 'foo']
+
+        assert store[self.root + 'foo'] == b''
+
```
jakirkham, line 1812: This is great. Would suggest moving this method to [the `StoreTests` class above]( https://github.com/I-am-Emmanuel/zarr-python/blob/0144f953e99ec73f4d03a7e6825ea48578e9e600/zarr/tests/test_storage.py#L110 )

zarr/tests/test_storage.py
jakirkham, line 1825: Just realized we can actually drop these methods since the `StoreTests` base class already includes them. So would go ahead and delete them. Though we can include any relevant changes in the [`test_pop`]( https://github.com/I-am-Emmanuel/zarr-python/blob/0144f953e99ec73f4d03a7e6825ea48578e9e600/zarr/tests/test_storage.py#L173 ) and [`test_popitem`]( https://github.com/I-am-Emmanuel/zarr-python/blob/0144f953e99ec73f4d03a7e6825ea48578e9e600/zarr/tests/test_storage.py#L195 ) methods above.

I-am-Emmanuel: Okay. I will do that. Thanks

zarr/tests/test_storage.py
I-am-Emmanuel: Alright. I will work on it.

zarr/storage.py
```diff
@@ -1779,8 +1779,24 @@ def __setitem__(self, key, value):
             self.zf.writestr(keyinfo, value)
 
     def __delitem__(self, key):
-        raise NotImplementedError
+        # raise NotImplementedError
+        value = b""
+        try:
+            with self.mutex:
+                self.zf.getinfo(key)
```
jakirkham, line 1786: Maybe we can place the `try...except...` just around this line?

zarr/tests/test_storage.py
```diff
@@ -1800,14 +1808,16 @@ def test_context_manager(self):
         with self.create_store() as store:
             store[self.root + 'foo'] = b'bar'
             store[self.root + 'baz'] = b'qux'
-            assert 2 == len(store)
+            assert 2 == len(store)
```
jakirkham, line 1811:
```suggestion
            assert 2 == len(store)
```

zarr/tests/test_storage.py
```diff
@@ -1792,22 +1800,24 @@ def test_flush(self):
         store.flush()
         assert store[self.root + 'foo'] == b'bar'
         store.close()
-
+
```
jakirkham, line 1803:
```suggestion

```

zarr/storage.py
```diff
@@ -1779,7 +1779,25 @@ def __setitem__(self, key, value):
             self.zf.writestr(keyinfo, value)
 
     def __delitem__(self, key):
-        raise NotImplementedError
+
```
jakirkham, line 1782:
```suggestion
```

zarr/storage.py
```diff
@@ -1779,7 +1779,25 @@ def __setitem__(self, key, value):
             self.zf.writestr(keyinfo, value)
 
     def __delitem__(self, key):
-        raise NotImplementedError
+        
+        value = b""
+
```
jakirkham, line 1784:
```suggestion

```

zarr/storage.py
```diff
@@ -1779,7 +1779,25 @@ def __setitem__(self, key, value):
...
+        
+        value = b""
+    
+        with self.mutex:
+            try:
+                self.zf.getinfo(key)
+            except KeyError:
+                raise KeyError("Cannot delete a non-existent key")
+            keyinfo = zipfile.ZipInfo(filename=key, 
+                                        date_time=time.localtime(time.time())[:6])
```
jakirkham, line 1791:
```suggestion
            keyinfo = zipfile.ZipInfo(
                filename=key, 
                date_time=time.localtime(time.time())[:6]
            )
```

zarr/storage.py
```diff
@@ -1779,7 +1779,25 @@ def __setitem__(self, key, value):
...
+            keyinfo.compress_type = self.compression
+            if keyinfo.filename[-1] == os.sep:
+                keyinfo.external_attr = 0o40775 << 16   # drwxrwxr-x
+                keyinfo.external_attr |= 0x10           # MS-DOS directory flag
+            else:
+                keyinfo.external_attr = 0o644 << 16     # ?rw-r--r--
+
+            self.zf.writestr(keyinfo, value)
+
```
jakirkham, line 1800:
```suggestion
```

jakirkham (commented):
Proposing a few changes below to fix [lints]( https://results.pre-commit.ci/run/github/48049137/1666295952.tXoQYX6eQyasuB4IQ5EQyQ )

zarr/storage.py
```diff
@@ -1779,7 +1779,24 @@ def __setitem__(self, key, value):
...
     def __delitem__(self, key):
-        raise NotImplementedError
+        value = b""
+        with self.mutex:
+            try:
+                self.zf.getinfo(key)
+            except KeyError:
+                raise KeyError("Cannot delete a non-existent key")
+            keyinfo = zipfile.ZipInfo(
+                filename=key,
```
jakirkham, line 1789:
```suggestion
                filename=key,
```

zarr/storage.py
```diff
@@ -1779,7 +1779,24 @@ def __setitem__(self, key, value):
             self.zf.writestr(keyinfo, value)
 
     def __delitem__(self, key):
-        raise NotImplementedError
+        value = b""
```
jakirkham, line 1782:
```suggestion
```

zarr/storage.py
```diff
@@ -1779,7 +1779,24 @@ def __setitem__(self, key, value):
...
+                date_time=time.localtime(time.time())[:6]
+            )
+            keyinfo.compress_type = self.compression
+            if keyinfo.filename[-1] == os.sep:
+                keyinfo.external_attr = 0o40775 << 16   # drwxrwxr-x
+                keyinfo.external_attr |= 0x10           # MS-DOS directory flag
+            else:
+                keyinfo.external_attr = 0o644 << 16     # ?rw-r--r--
+
+            self.zf.writestr(keyinfo, value)
```
jakirkham, line 1799:
```suggestion
            self.zf.writestr(keyinfo, b"")
```

jakirkham (commented):
Since `value` is only used once, suggesting we inline it.

zarr/storage.py
jakirkham: This change has been made

zarr/storage.py
jakirkham: Also done

zarr/tests/test_storage.py
jakirkham: This is done

zarr/tests/test_storage.py
```diff
@@ -1806,8 +1813,10 @@ def test_pop(self):
...
         store[self.root + 'foo'] = b'bar'
-        with pytest.raises(NotImplementedError):
-            store.pop(self.root + 'foo')
+
+        store.pop(self.root + 'foo')
+
+        assert store[self.root + 'foo'] == b""
 
     def test_popitem(self):
         # override because not implemented
```
jakirkham, line 1822: Would also update this test

jakirkham:
Thanks Emmanuel! 🙏

Looks like many things are done. So marked them resolved above.

The linter raised a few stylistic issues. Went ahead and just added these to the PR for simplicity.

There is one more test remaining that I've commented on above.

There is a test failure we need to look at. Will give it some thought.

zarr/tests/test_storage.py
I-am-Emmanuel: Okay.

zarr/tests/test_storage.py
jakirkham: Oh this change has already been made. So this can be ignored

zarr/tests/test_storage.py
```diff
@@ -192,6 +192,14 @@ def test_pop(self):
 
         store.close()
 
+    def test_delitem(self):
+        store = self.create_store()
+        store[self.root + 'foo'] = b'bar'
+
+        del store[self.root + 'foo']
+
+        assert store[self.root + 'foo'] == b''
```
jakirkham, line 201:
```suggestion
        assert store.get(self.root + 'foo', b'') == b''
```

zarr/tests/test_storage.py
```diff
@@ -9,6 +9,7 @@
 import tempfile
 from contextlib import contextmanager
 from pickle import PicklingError
+from xxlimited import foo
```
joshmoore, line 12: I assume this is imported by accident?

zarr/tests/test_storage.py
```diff
@@ -192,6 +192,19 @@ def test_pop(self):
 
         store.close()
 
+    def test_delitem(self):
+        store = self.create_store()
+        foo = self.root + 'foo'
+        store[foo] = b'bar'
+
+        del store[foo]
+
```
jakirkham, line 201: Dropping extra spaces

```suggestion

```

jakirkham:
Maybe we can rewrite `__getitem__` for `ZipStore` like so?

```python
    def __getitem__(self, key):
        with self.mutex:
            with self.zf.open(key) as f:  # will raise KeyError
                data = f.read()
                if data:
                    return data
                else:
                    # Treat empty `data` as deleted (so no `key` present)
                    raise KeyError(key)
```

Perhaps this fixes the remaining tests

jakirkham:
Think we need to tweak a few of `ZipStore`'s methods to treat keys with values equal to `b""` as missing. Once that is done this should fix the remaining test issues (and remove the need for special casing `ZipStore` in the tests)

I-am-Emmanuel:
Waiting for this work to be reviewed.

jakirkham:
Thanks @I-am-Emmanuel! 🙏

That's a good start. Think we still need similar changes to `__contains__` and `keylist` (skipping empty entries)

MSanKeys963:
@I-am-Emmanuel, would you like to finish this?

jakirkham (commented):
Sorry had some review comments that didn't post. So completing to get those out

jhamman:
I'm going to close this as stale. Folks should feel free to reopen if there is interest in continuing this work. Note there is a new implementation of the Zip store in v3 that could use the same patch for whoever is interested.

PR #1219, opened 2022-10-24
Remove `tox` support

### LINKED ISSUE #1211: Re-use tox where possible from GHA workflows
If the tox.ini file is going to be maintained, then existing workflows should make use of it.

e.g. see https://github.com/zarr-developers/zarr-python/pull/1210#issuecomment-1287771539 in which the `tox -e docs` target was not being tested anywhere.

Saransh-cpp:
Resolves #1211

As title!

TODO:


jakirkham:
Thanks Saransh! 🙏

Should we also remove these instances too?

https://github.com/zarr-developers/zarr-python/blob/daad2fa86a32c3ed10871f6ba191ed3356b29e91/release.txt#L1
https://github.com/zarr-developers/zarr-python/blob/f361631384fc8e4ce7a76528eac7e89bd8f58cb6/requirements_dev_optional.txt#L16
https://github.com/zarr-developers/zarr-python/blob/daad2fa86a32c3ed10871f6ba191ed3356b29e91/.gitignore#L38

(not sure if the last one might be worth keeping)

jakirkham (approved):

GbotemiB:
Hi @Saransh-cpp and @jakirkham. since support tox has been removed. Can I go ahead to update the docs to build docs with `'make html'`

jakirkham:
Hi @GbotemiB. Thanks for the offer. Think that is already in Saransh's PR, but if you see somewhere else that needs to be updated please let us know and we can fix that here.

joshmoore:
> [jakirkham](/jakirkham) requested a review from [joshmoore](/joshmoore) [ 13 hours ago](#event-7657620492)

I'm a bit torn, to be honest. The tox.ini is definitely not in an ideal state, especially now with everything being so nice and clean thanks to @Saransh-cpp! However, I do like the idea of having one place that encapsulates the various commands that are needed. I could see preparing a _second_ PR which does the opposite, namely to go all-in on tox and then we could compare the two.

Saransh-cpp:
May I also suggest looking into [`nox`](https://github.com/wntrblm/nox)? We use `nox` in most of the [`Scikit-HEP`](https://github.com/scikit-hep) packages (for example, vector's [`noxfile`](https://github.com/scikit-hep/vector/blob/main/noxfile.py)), and I personally like how one can specify instructions in pure python, instead of having a `.ini` file.

I'll be happy to work on either of them (`tox.ini` or `noxfile.py`)!

jakirkham:
> I'm a bit torn, to be honest. The tox.ini is definitely not in an ideal state, especially now with everything being so nice and clean thanks to @Saransh-cpp! However, I do like the idea of having one place that encapsulates the various commands that are needed. I could see preparing a _second_ PR which does the opposite, namely to go all-in on tox and then we could compare the two.

Given how valuable the `conda-incubator/setup-miniconda` GHA workflow has been in cutting down maintenance, am not eager to replace that with tox (where we have to maintain even more stuff)

The fact that new contributors are stumbling over tox here just makes it that much clear in my mind that it is time to drop it

joshmoore:
Definitely agreed on not reproducing the environmental logic within tox. I rarely if ever use tox for running multiple environments. That type of testing I defer to GHA. For me, it's more a matter of having the individual commands that are hard-coded in the workflows available to be run locally. At the moment, we have different sets of commands spread between:

 * tox.ini
 * the docs/ files
 * the .github workflows

If there were a way to reduce that down to 1️⃣, I'd be pretty happy.

Saransh-cpp:
@joshmoore @jakirkham, could you please have a look at the last commit? I have added `nox` support to address @joshmoore's comments!

I feel like maintaining a `noxfile` would be much easier than maintaining a `tox.ini` file. Plus it is all written in `Python`! I have also added documentation for the same!

jakirkham:
One could run GitHub Actions locally

https://github.com/nektos/act

joshmoore:
Thanks for drafting the nox support, @Saransh-cpp. I agree that the Python syntax is a nice improvement, but I'm not sure it warrants the addition of a new tool. Open to other opinions, but I'd err on the side of reverting those changes. I can understand if you'd like to then just get this merged as is, otherwise happy to discuss other options for the unification.

Saransh-cpp:
Completely understandable! Maybe we can then link or document the usage of https://github.com/nektos/act for Zarr?

jakirkham:
Documenting how to use `act` to run Zarr's tests locally sounds reasonable

joshmoore:
Thanks all. @Saransh-cpp, if you can revert, then let's go ahead and get this merged. Someone can work on documenting akt, and I might look into unifying the docs & the GHA commands.

Saransh-cpp:
Reverted!

joshmoore (approved):
Awesome. Thank you!

tox.ini
```diff
@@ -1,51 +0,0 @@
...
-[testenv:docs]
-basepython = python3.8
-changedir = docs
-deps =
-    -rrequirements_rtfd.txt
-commands =
-    sphinx-build -W -b html -d {envtmpdir}/doctrees .  {envtmpdir}/html
-
-[flake8]
-max-line-length = 100
```
jakirkham, line 51: Just noticed that we lost the `flake8` configuration here. Adding back in PR ( https://github.com/zarr-developers/zarr-python/pull/1249 ).

PR #1225, opened 2022-10-27
Apply some pyupgrade suggestions

DimitriPapadopoulos:
* Yield from an iterable instead of iterating to yield items.
* `io.open()` is an alias for for the builtin `open()` function:
  https://docs.python.org/3/library/io.html#io.open
* `IOError` is kept for compatibility with previous versions; starting from Python 3.3, it is an alias of `OSError`:
  https://docs.python.org/3/library/exceptions.html#IOError

TODO:

### DIFF
--- a/zarr/errors.py
+++ b/zarr/errors.py
@@ -1,5 +1,3 @@
-
-
 class MetadataError(Exception):
     pass
 
--- a/zarr/hierarchy.py
+++ b/zarr/hierarchy.py
@@ -633,8 +633,7 @@ class Group(MutableMapping):
                     yield _key if keys_only else (_key, self[key])
                 elif recurse and contains_group(self._store, path):
                     group = self[key]
-                    for i in getattr(group, method)(recurse=recurse):
-                        yield i
+                    yield from getattr(group, method)(recurse=recurse)
         else:
             dir_name = meta_root + self._path
             array_sfx = '.array' + self._metadata_key_suffix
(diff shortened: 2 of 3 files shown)


Saransh-cpp (commented):
A general good practice -

Such pre-commit changes (`pyupgrade` can be added as a pre-commit) should be carried out in a single commit (that is - changing "some pyupgrade suggestions" to "all pyupgrade suggestions"). 

This commit's hash should then be added in the `.git-blame-ignore-revs` file (does not exist in Zarr, but should be added in the root folder) to preserve `git blame`.

For example, see the `.git-blame-ignore-revs` - https://github.com/JuliaLang/julia/blob/master/.git-blame-ignore-revs - file of Julia!

DimitriPapadopoulos:
One of pyupgrade suggestions doesn't seem to be working `io.open()` → `open()` - I am investigate that - and others may be debatable. My past experience with other projects is that not all the suggestions of a tool are accepted, and often adding a tool to CI or pre-commit isn't accepted either.

DimitriPapadopoulos:
This change doesn't pass tests and that's rather curious because `io.open()` and open() are aliases:
* `io.open()` is an alias for the builtin `open()` function:
  https://docs.python.org/3/library/io.html#io.open

Removed from now, must probably be investigated in a PR of its own.

joshmoore:
> Also, why hide these changes from `git blame`?

I'd also agree. When/if we run `black` on the entire code base, I could definitely see having _that_ in `.git-blame-ignore-revs`, but as long as the changes are fairly manageable, happy for people to get credit (in blame).

@DimitriPapadopoulos: do you expect any more commits here?

PR #1226, opened 2022-10-27
Fix linting errors

caviere:
Fix possible unbound variable error
Avoid extra zarr version check
Add Optional typing to indicate value might be None

### DIFF
--- a/zarr/creation.py
+++ b/zarr/creation.py
@@ -1,3 +1,4 @@
+from typing import Optional
 from warnings import warn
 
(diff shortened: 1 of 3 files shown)


joshmoore (approved):
Thanks, @caviere. Changes look fine and all is green.

Out of curiosity, what linter are these from?

caviere:
I am using pyright currently

joshmoore:
Many thanks!

PR #1264, opened 2022-11-15
Get coverage up to 100%

jakirkham:
TODO:


zarr/hierarchy.py
```diff
@@ -1430,7 +1430,7 @@ def open_group(store=None, mode='a', cache_attrs=True, synchronizer=None, path=N
                                            mode=mode,
                                            zarr_version=zarr_version)
         if getattr(chunk_store, '_store_version', DEFAULT_ZARR_VERSION) != zarr_version:
-            raise ValueError(
+            raise ValueError(  # pragma: no cover
```
joshmoore, line 1433: I don't imagine it would be too difficult to write a test for this one, but no objection to skipping.

joshmoore (approved):
:100: for `100%`

zarr/hierarchy.py
jakirkham: We can turn this into an issue if we like

PR #1268, opened 2022-11-17
Fix double counting V3 groups bug

### LINKED ISSUE #1228: Double counting groups in V3
### Zarr version

2.13.3

### Numcodecs version

n/a

### Python Version

3.10

### Operating System

Mac

### Installation

poetry

### Description

I have discovered that certain zarr methods double count groups in V3. The methods `Group.group_keys()` and `Group.groups()` return the group twice when the group contains children.

Instead, each group should only be returned once.

### Steps to reproduce

The following test should pass

```python
import zarr

def test_double_counting_group_v3():
    store = zarr.MemoryStoreV3()
    root_group = zarr.group(store=store, zarr_version=3)
    sub_group = root_group.create_group("foo")
    array = sub_group.create("bar", shape=10, dtype="i4")
    # this works and is in spec (although the spec says it should be "foo/")
    assert store.listdir("meta/root") == ['foo', 'foo.group.json']
    # these should work but don't: instead we get foo twice
    assert list(root_group.group_keys()) == ['foo']
    assert list(root_group.groups()) == [("foo", sub_group)]
```

What is happening is that the the `listdir` call here is returning both `foo` and `foo.group.json`. Because `contains_group('foo')` is valid, the group is returned twice.

https://github.com/zarr-developers/zarr-python/blob/ce129a560d48854aee533bb2699a3f28b396bc22/zarr/hierarchy.py#L513-L521

If others agree this is a bug, I will submit a PR to fix it.

### Additional output

_No response_

rabernat:
Fixes #1228

TODO:
* ~Add docstrings and API docs for any new/modified user-facing classes and functions~
* ~New/modified features documented in docs/tutorial.rst~

### DIFF
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -14,6 +14,12 @@ Unreleased
     # .. warning::
     #    Pre-release! Use :command:`pip install --pre zarr` to evaluate this release.
 
+* Fix bug that caused double counting of groups in ``groups()`` and ``group_keys()``
+  methods with V3 stores.
+  By :user:`Ryan Abernathey <rabernat>` :issue:`1228`.
+
+.. _release_2.13.2:
+
 Maintenance
 ~~~~~~~~~~~
 
--- a/zarr/hierarchy.py
+++ b/zarr/hierarchy.py
@@ -511,9 +511,15 @@ class Group(MutableMapping):
         else:
             dir_name = meta_root + self._path
             group_sfx = '.group' + self._metadata_key_suffix
-            for key in sorted(listdir(self._store, dir_name)):
+            # The fact that we call sorted means this can't be a streaming generator.
+            # The keys are already in memory.
+            all_keys = sorted(listdir(self._store, dir_name))
+            for key in all_keys:
                 if key.endswith(group_sfx):
                     key = key[:-len(group_sfx)]
+                    if key in all_keys:
+                        # otherwise we will double count this group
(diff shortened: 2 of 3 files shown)


rabernat:
Is anyone up for reviewing this simple PR?

docs/release.rst
```diff
@@ -14,6 +14,12 @@ Unreleased
     # .. warning::
     #    Pre-release! Use :command:`pip install --pre zarr` to evaluate this release.
 
+* Fix bug that caused double counting of groups in ``groups()`` and ``group_keys()``
+  methods with V3 stores.
+  By :user:`Ryan Abernathey <rabernat>` :issue:`1228`.
+
+.. _release_2.13.2:
```
joshmoore, line 21: This might end up being redundant:

https://github.com/zarr-developers/zarr-python/blob/main/docs/release.rst#2132

joshmoore (approved):
Barring the minor release.rst issue (which will need work anyway for the release), this makes sense to me.

docs/release.rst
rabernat: I can't tell if you are requesting changes to my PR or not. Could you be more explicit?

joshmoore: Sorry. I updated the branch and there doesn't look to be a conflict. (My expectation was there were going to be two)

Issue #1279, opened 2022-12-01
Reading/writing F-order data w/o compression is broken


normanrz:
### Zarr version

v2.13.3

### Numcodecs version

v0.10.2

### Python Version

3.10.8

### Operating System

tested on Linux, Mac

### Installation

poetry

### Description

I was trying to upgrade zarr-python in our [webKnossos libs](https://github.com/scalableminds/webknossos-libs) and noticed failing tests. The tests essentially write some data, reopen the array, read the data back, and compare before/after. 
We use F-order to write our data. I tried to replace `F` with `C` order and the tests succeeded. Interestingly, the issue does not occur when writing with compression (tested with Blosc). I didn't check whether the writing or reading part is the cause for this problem.

### Steps to reproduce

```python
from pathlib import Path
from shutil import rmtree

import numpy as np
import zarr


def get_multichanneled_data(dtype: type) -> np.ndarray:
    data: np.ndarray = np.zeros((3, 250, 200, 10), dtype=dtype)
    for h in range(10):
        for i in range(250):
            for j in range(200):
                data[0, i, j, h] = i * 256
                data[1, i, j, h] = j * 256
                data[2, i, j, h] = 100 * 256
    return data


def main():
    ds_path = Path("testoutput") / "test_zarr"
    rmtree(ds_path, ignore_errors=True)

    data = get_multichanneled_data(np.uint16)
    ds = zarr.create(
        shape=data.shape,
        chunks=(3, 32, 32, 32),
        dtype=data.dtype,
        compressor=(None),
        store=zarr.storage.FSStore(url=str(ds_path), mode="a"),
        order="F",
    )

    ds[:, :, :, :] = data

    ds_reopened = zarr.open_array(
        store=zarr.storage.FSStore(url=str(ds_path), mode="r")
    )

    written_data = ds_reopened[:, :, :, :]

    assert data.shape == written_data.shape
    assert np.array_equal(data, written_data)


if __name__ == "__main__":
    main()
```

normanrz:
I just checked: the write path seems to be the culprit.

joshmoore:
Can confirm that using `FSStore` fails while `DirectoryStore` passes:

```
    ds = zarr.create(
        shape=data.shape,
        chunks=(3, 32, 32, 32),
        dtype=data.dtype,
        compressor=(None),
        store=zarr.storage.DirectoryStore(str(ds_path)),
        order="F",
    )

    ds[:, :, :, :] = data

[...]

PR #1285, opened 2022-12-06
FSStore: use `ensure_bytes()`

madsbk:
Fixes https://github.com/zarr-developers/zarr-python/issues/1279

TODO:

### DIFF
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -55,8 +55,8 @@ from zarr.meta import encode_array_metadata, encode_group_metadata
 from zarr.util import (buffer_size, json_loads, nolock, normalize_chunks,
                        normalize_dimension_separator,
                        normalize_dtype, normalize_fill_value, normalize_order,
-                       normalize_shape, normalize_storage_path, retry_call
-                       )
+                       normalize_shape, normalize_storage_path, retry_call,
+                       ensure_contiguous_ndarray_or_bytes)
 
 from zarr._storage.absstore import ABSStore  # noqa: F401
 from zarr._storage.store import (_get_hierarchy_metadata,  # noqa: F401
@@ -1380,14 +1380,17 @@ class FSStore(Store):
             raise ReadOnlyError()
 
         # Normalize keys and make sure the values are bytes
-        values = {self._normalize_key(key): ensure_bytes(val) for key, val in values.items()}
+        values = {
+            self._normalize_key(key): ensure_contiguous_ndarray_or_bytes(val)
+            for key, val in values.items()
+        }
         self.map.setitems(values)
 
     def __setitem__(self, key, value):
         if self.mode == 'r':
             raise ReadOnlyError()
         key = self._normalize_key(key)
-        value = ensure_bytes(value)
+        value = ensure_contiguous_ndarray_or_bytes(value)
         path = self.dir_path(key)
         try:
             if self.fs.isdir(path):
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -1378,13 +1378,16 @@ class FSStore(Store):
     def setitems(self, values):
         if self.mode == 'r':
             raise ReadOnlyError()
-        values = {self._normalize_key(key): val for key, val in values.items()}
+
+        # Normalize keys and make sure the values are bytes
+        values = {self._normalize_key(key): ensure_bytes(val) for key, val in values.items()}
         self.map.setitems(values)
 
     def __setitem__(self, key, value):
         if self.mode == 'r':
             raise ReadOnlyError()
         key = self._normalize_key(key)
+        value = ensure_bytes(value)
         path = self.dir_path(key)
         try:
             if self.fs.isdir(path):
--- a/zarr/util.py
+++ b/zarr/util.py
@@ -5,16 +5,21 @@ import numbers
 from textwrap import TextWrapper
 import mmap
 import time
+from typing import Any, Callable, Dict, Optional, Tuple, Union
 
 import numpy as np
 from asciitree import BoxStyle, LeftAligned
 from asciitree.traversal import Traversal
 from collections.abc import Iterable
-from numcodecs.compat import ensure_text, ensure_ndarray_like
+from numcodecs.compat import (
+    ensure_text,
+    ensure_ndarray_like,
+    ensure_bytes,
+    ensure_contiguous_ndarray_like
+)
+from numcodecs.ndarray_like import NDArrayLike
 from numcodecs.registry import codec_registry
 from numcodecs.blosc import cbuffer_sizes, cbuffer_metainfo
-
-from typing import Any, Callable, Dict, Optional, Tuple, Union
 
 
 def flatten(arg: Iterable) -> Iterable:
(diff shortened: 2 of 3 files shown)


madsbk:
@normanrz can you check if this fixes the issue?

normanrz:
Yes! That looks good to me. Thanks for the quick fix!

rabernat:
> We changed this with #934, now we also allow [`NDArrayLike`](https://github.com/zarr-developers/numcodecs/blob/922ab23d85a9f854352782212a5e057b7e31e754/numcodecs/ndarray_like.py#L45) instances.

I see that this has been implemented in the zarr-python implementation. And I thank everyone who worked on this amazing feature! 🚀  But that doesn't mean it is consistent with what the [spec says](https://zarr-specs.readthedocs.io/en/latest/core/v3.0.html#abstract-store-interface):

> The store interface defines a set of operations involving keys and values. In the context of this interface, a key is any string... A value is a sequence of bytes.

It is certainly straightforward to convert an uncompressed ndarray into a sequence of bytes. So maybe I am overthinking this... Or maybe the spec needs to be updated to reflect these new implementation changes? 🤔  My only goal here is to make sure that the spec and the implementation are consistent with each other. Otherwise we risk interoperability with other zarr implementations.

If we are fine with saying "a value is a sequence of bytes or a raw array" then I am fine with leaving it up to the store implementation whether to convert arrays to bytes.

madsbk:
Good point and I agree, the spec and implementation should match up!

I must admit I read "A value is a sequence of bytes" in a more general sense, not  a reference to the `bytes` class specifically. Otherwise,  types like `bytearray` or `memoryview` are also excluded.

jakirkham:
Hey Mads, thanks for following up on this issue.

That said, don't think we should be making this change. In general it is already the expectation in zarr-python & Numcodecs that [Python Buffer Protocol]( https://docs.python.org/3/c-api/buffer.html ) supporting objects are used (`memoryview`s, `ndarray`s, `bytearray`s, `bytes`, etc.). Part of the reason for this is it allows zero-copy operations on the data (like casting, reshaping, etc.). It also can support things like memory-mapping natively (as `mmap` supports the Python Buffer Protocol).

How to word this in the spec is nuanced. The spec itself is not meant to be Python specific, but to support other languages as well. They may not have `bytearray` or `memoryview` for example or if they do it may be named differently and/or behave differently. So think we should consider language agnostic wording. It sounds like this is how you understood it as well.

Perhaps others who found this wording in the spec confusing can chime in on how to clarify it. Maybe we can say "a value contains binary data"?

madsbk:
@jakirkham, I am not sure I understand. Are you saying that we shouldn't use `ensure_bytes()` in `FSStore` (this PR) or that the spec should stay language agnostic? 

I have no strong opinion about the specifications, but I think it is important that we allow  `NDArrayLike` objects in `values`. If we move `ensure_bytes()` to the `zarr.core` level, we are effectively saying that a `value` is always of type `bytes`.

jakirkham:
Could we call `ensure_ndarray_like` instead?

normanrz:
> Could we call `ensure_ndarray_like` instead?

I'm not sure that would work. I think the trick is that at some point, `tobytes` needs to be called with `value.tobytes(order='A')` to preserve the order. https://numpy.org/doc/stable/reference/generated/numpy.ndarray.tobytes.html

jakirkham:
There's `ensure_contiguous_ndarray_like`, which does that

jstriebel:
Since most other stores use `ensure_bytes` atm, I think this fix is valid for the moment, and important to fix an existing bug. I agree, that there should be a discussion whether every store should do this or not, but this should not block this PR IMO.

rabernat (approved):
I am fine with this fix for now, since it seems consistent with what most other stores do.

We can revisit the question of whether this `ensure_bytes` logic should live elsewhere in the future.

jakirkham:
Sorry to be a stickler. The main reason I bring up `ensure_bytes` here is this typically incurs the cost of a copy (with additional memory being used). Given this is happening with multiple memory segments that are being written together, this could result in more significant memory usage. Especially as we start to rely more on batching multiple reads & writes. While it may not always be possible to avoid these copies, they could be deferred until absolutely necessary (with the resulting `bytes` objects quickly discarded) to keep memory usage as low as possible (since not all memory segments would needing copying at once). Hope that clarifies the concern a bit more.

zarr/storage.py
```diff
@@ -1378,13 +1378,16 @@ def __getitem__(self, key):
     def setitems(self, values):
         if self.mode == 'r':
             raise ReadOnlyError()
-        values = {self._normalize_key(key): val for key, val in values.items()}
+
+        # Normalize keys and make sure the values are bytes
+        values = {self._normalize_key(key): ensure_bytes(val) for key, val in values.items()}
         self.map.setitems(values)
```
jakirkham, line 1384: Maybe something like this would work

```suggestion
        norm_values = {}
        for key, val in values.items():
            norm_key = self._normalize_key(key)
            try:
                val = ensure_contiguous_ndarray_like(val)
            except TypeError:
                val = ensure_bytes(val)
            norm_values[norm_key] = val
        self.map.setitems(norm_values)
```

normanrz:
Any chance we could get this merged and released?

rabernat:
I'd like to hear from @jakirkham - John, would you be you okay with addressing  `ensure_contiguous_ndarray_like` vs. `ensure_bytes` in a follow-on PR (where we can also address it for all the other stores that call `ensure_bytes`)? Or do you think this needs to be resolved before we merge?

normanrz:
I think @jakirkham's [latest comment](https://github.com/zarr-developers/zarr-python/pull/1285#discussion_r1048880317) is a straightforward compromise that could be incorporated as part of this PR.

madsbk:
> I think @jakirkham's [latest suggestion](https://github.com/zarr-developers/zarr-python/pull/1285#discussion_r1048880317) is a straightforward compromise that could be incorporated as part of this PR.

Done

joshmoore:
Re-opening to try to fix the codecov build.

joshmoore:
Now green. @jakirkham: does this match your expectations?

If so, happy to get it in and prep a release during the holidays. (Help with the release notes will be appreciated!)

rabernat:
If we're doing a release, let's get #1304 in too.

joshmoore:
> If we're doing a release, let's get #1304 in too.

see https://github.com/zarr-developers/zarr-python/pull/1309#issuecomment-1384179684

Leaving this PR for @jakirkham to review.

zarr/util.py
```diff
@@ -696,3 +701,28 @@ def all_equal(value: Any, array: Any):
             # using == raises warnings from numpy deprecated pattern, but
             # using np.equal() raises type errors for structured dtypes...
             return np.all(value == array)
+
+
+def ensure_contiguous_ndarray_or_bytes(buf) -> Union[NDArrayLike, bytes]:
```
jakirkham, line 706: We may want to move this to [`numcodecs.compat`]( https://github.com/zarr-developers/numcodecs/blob/67ede4c6b4f1707f1da18351945e10904a5572de/numcodecs/compat.py ), but that can be follow-on work

zarr/storage.py
```diff
@@ -1395,13 +1395,19 @@ def __getitem__(self, key):
     def setitems(self, values):
         if self.mode == 'r':
             raise ReadOnlyError()
-        values = {self._normalize_key(key): val for key, val in values.items()}
+
+        # Normalize keys and make sure the values are bytes
+        values = {
+            self._normalize_key(key): ensure_contiguous_ndarray_or_bytes(val)
+            for key, val in values.items()
+        }
```
jakirkham, line 1403: Something we may consider is delaying the instantiation of the `dict` so these copies occur as these values are requested. This can be a follow-on though

jakirkham (approved):
Thanks Mads! 🙏

LGTM. Made a few comments that we can follow up on separately

joshmoore:
@jakirkham: ok to get this in the 2.13.x series?

@madsbk: I'll get this out ASAP but might wait on any emergency needs for re: #1324

Suggestions for the docs/release.rst statement would be welcome. 🙏🏽

jakirkham:
> Suggestions for the docs/release.rst statement would be welcome. 🙏🏽

Oops sorry should have asked for that here (or just added it). Submitted PR ( https://github.com/zarr-developers/zarr-python/pull/1325 )

PR #1299, opened 2022-12-15
Attrs kwarg

### LINKED ISSUE #538: attrs as kwarg
Is there any reason why it's not possible to set the attrs of an array / group by passing a kwarg to the  relevant creation function? e.g.: 
```python
z = zarr.create((10,10),chunks=(1,1), attrs={'foo':10})
```
(The current behavior is to ignore the `attrs` kwarg)
Looking back in time, I see this discussion: #121. It seems attrs as a kwarg was not pursued because of concerns about matching the behavior of hdf5? How relevant is that concern today>

d-v-b:
This PR adds `attrs` as a keyword argument to `init_array` and `init_group`. In both cases, the value assigned to `attrs` is inserted into the attributes of the group / array upon initialization. 

Resolves #538

### DIFF
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -63,7 +63,8 @@ from zarr._storage.absstore import ABSStore  # noqa: F401
 from zarr._storage.store import (_get_hierarchy_metadata,  # noqa: F401
                                  _get_metadata_suffix,
                                  _listdir_from_keys,
-                                 _prefix_to_attrs_key,
+                                 _prefix_to_array_attrs_key,
+                                 _prefix_to_group_attrs_key,
                                  _rename_from_keys,
                                  _rename_metadata_v3,
                                  _rmdir_from_keys,
@@ -435,17 +436,27 @@ def init_array(
                          object_codec=object_codec,
                          dimension_separator=dimension_separator)
 
+    _init_array_attrs(store, path, attrs)
+
+
+def _init_array_attrs(store: StoreLike, path: Optional[str], attrs: Dict[str, Any]):
     if len(attrs):
         if path:
             key_prefix = path + '/'
         else:
             key_prefix = ''
-        _init_attrs(store, key_prefix, attrs)
+        akey = _prefix_to_array_attrs_key(store, key_prefix)
+        Attributes(store, key=akey, cache=False).update(attrs)
 
 
-def _init_attrs(store: StoreLike, key_prefix: str, attrs: Dict[str, Any]):
-    akey = _prefix_to_attrs_key(store, key_prefix)
-    Attributes(store, key=akey, cache=False).update(attrs)
+def _init_group_attrs(store: StoreLike, path: Optional[str], attrs: Dict[str, Any]):
+    if len(attrs):
+        if path:
+            key_prefix = path + '/'
+        else:
+            key_prefix = ''
+        akey = _prefix_to_group_attrs_key(store, key_prefix)
+        Attributes(store, key=akey, cache=False).update(attrs)
 
 
 def _init_array_metadata(
@@ -653,12 +664,7 @@ def init_group(
                          chunk_store=chunk_store)
 
     # initialize attrs
-    if len(attrs):
-        if path:
-            key_prefix = path + '/'
-        else:
-            key_prefix = ''
-        _init_attrs(store, key_prefix, attrs)
+    _init_group_attrs(store, path, attrs)
 
     if store_version == 3:
         # TODO: Should initializing a v3 group also create a corresponding
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -446,7 +446,7 @@ def _init_array_attrs(store: StoreLike, path: Optional[str], attrs: Dict[str, An
         else:
             key_prefix = ''
         akey = _prefix_to_array_attrs_key(store, key_prefix)
-        Attributes(store, key=akey, cache=False).update(attrs)
+        Attributes(store, key=akey, cache=False).put(attrs)
 
 
 def _init_group_attrs(store: StoreLike, path: Optional[str], attrs: Dict[str, Any]):
@@ -456,7 +456,7 @@ def _init_group_attrs(store: StoreLike, path: Optional[str], attrs: Dict[str, An
         else:
             key_prefix = ''
         akey = _prefix_to_group_attrs_key(store, key_prefix)
-        Attributes(store, key=akey, cache=False).update(attrs)
+        Attributes(store, key=akey, cache=False).put(attrs)
 
 
 def _init_array_metadata(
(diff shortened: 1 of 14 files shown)


zarr/storage.py
```diff
@@ -430,6 +436,28 @@ def init_array(
                          object_codec=object_codec,
                          dimension_separator=dimension_separator)
 
+    _init_array_attrs(store, path, attrs)
```
rabernat, line 439: What actually happens in terms of I/O operations when this is called. Specifically, in the case of V3 for example, do you know if the array metadata are re-read and re-written? That is my impression based on reading the code, but I find the `Attributes` class a bit hard to follow.

zarr/storage.py
```diff
@@ -633,6 +663,9 @@ def init_group(
     _init_group_metadata(store=store, overwrite=overwrite, path=path,
                          chunk_store=chunk_store)
 
+    # initialize attrs
+    _init_group_attrs(store, path, attrs)
```
rabernat, line 667: Same question

rabernat (commented):
Thanks so much for this! I think this optimization should really help speed up the creation of hierarchies with metadata. (I did some diagnostics of this on s3 [here](https://github.com/pangeo-data/pangeo-eosc/issues/39#issuecomment-1279628806) from the perspective of xarray.)

I have on question that is quite relevant to latency on object storage. Basically I am trying to understand whether this implementation truly reduces the number of I/O operations, or whether it simply provides a more convenient way to specify metadata at array creation time.

zarr/storage.py
d-v-b: I believe your impression is correct: calling `_init_array_attrs` calls [`Attributes._update_nosync`](https://github.com/d-v-b/zarr-python/blob/attrs_kwarg/zarr/attrs.py#L181), which eventually calls [`Attributes._get_nosync`](https://github.com/d-v-b/zarr-python/blob/attrs_kwarg/zarr/attrs.py#L40) and `Attributes._put_nosync`[https://github.com/d-v-b/zarr-python/blob/attrs_kwarg/zarr/attrs.py#L131], which read and write the entire attrs blob from storage respectively.

d-v-b:
`Zarr.creation` has lots of functions that independently hit the array / group creation routines, so with `attrs` as a keyword argument with a default, it would be very easy for a future creation routine to invoke `init_group` or `init_array` without supplying attrs, and thus providing an incomplete API (this has bitten me before with the many different creation routines). How do people think about making `attrs` a positional argument, or keyword only but with no default? This would ensure that any function calling `init_array` must deal with the `attrs` argument.

d-v-b:
> I have on question that is quite relevant to latency on object storage. Basically I am trying to understand whether this implementation truly reduces the number of I/O operations, or whether it simply provides a more convenient way to specify metadata at array creation time.

This change is motivated by the convenience. I haven't explicitly checked whether this actually saves I/O budget (but I doubt it).

zarr/storage.py
d-v-b: It looks like I can use `Attributes.put` instead of `Attributes.update` to avoid a read. I will switch to this approach.

d-v-b: @rabernat for v3 it looks like there's unnecessary IO [here](https://github.com/d-v-b/zarr-python/blob/attrs_kwarg/zarr/attrs.py#L158), where we check if the attrs key exists in the store. Since the routines I'm adding are called at initialization time, there's no need for this check. I'm still thinking about ways to avoid this for the specific case of initialization.

rabernat:
> How do people think about making `attrs` a positional argument, or keyword only but with no default? This would ensure that any function calling `init_array` must deal with the `attrs` argument.

I think this would be a breaking API change, so probably best to avoid that.

> it would be very easy for a future creation routine to invoke `init_group` or `init_array` without supplying attrs, and thus providing an incomplete API (this has bitten me before with the many different creation routines)

Can you elaborate more on the downside here? I'm not sure I fully understand the problem.

zarr/storage.py
rabernat: For this PR, let's just focus on the API change. We can collaborate on a subsequent PR that doesn't touch the API but focuses on i/o optimization.

d-v-b:
> > it would be very easy for a future creation routine to invoke `init_group` or `init_array` without supplying attrs, and thus providing an incomplete API (this has bitten me before with the many different creation routines)
> 
> Can you elaborate more on the downside here? I'm not sure I fully understand the problem.

There are a lot of places in the code that end up invoking `init_group`: e.g., [`Group.create_group`](https://github.com/zarr-developers/zarr-python/blob/1af77b63ad8d51a5a8dc2cd923bf73cb8abe5a64/zarr/hierarchy.py#L897), [`Group.create_groups`](https://github.com/zarr-developers/zarr-python/blob/1af77b63ad8d51a5a8dc2cd923bf73cb8abe5a64/zarr/hierarchy.py#L934) [`Group.require_group`](https://github.com/zarr-developers/zarr-python/blob/1af77b63ad8d51a5a8dc2cd923bf73cb8abe5a64/zarr/hierarchy.py#L938), [`Group.require_groups`](https://github.com/zarr-developers/zarr-python/blob/1af77b63ad8d51a5a8dc2cd923bf73cb8abe5a64/zarr/hierarchy.py#L978), [`group`](https://github.com/zarr-developers/zarr-python/blob/1af77b63ad8d51a5a8dc2cd923bf73cb8abe5a64/zarr/hierarchy.py#L1293),  [`open_group`](https://github.com/zarr-developers/zarr-python/blob/1af77b63ad8d51a5a8dc2cd923bf73cb8abe5a64/zarr/hierarchy.py#L1363), and in principle each added optional keyword argument to `init_group` is a new opportunity for the group creation routines to diverge slightly, in case the developer adding the new optional keyword argument does not remember to expose that kwarg to all the different group creation routines (i'm pretty sure I added `attrs` to all of them for this PR). I think for array creation all the high-level creation routines take `**kwargs`, which can defend against this kind of thing at the cost of an opaque function signature. 

Alternatively, we make a breaking change to `init_group` and add `attrs` as a positional argument, and then any higher-level routine that invokes it will break, but this will ensure consistency. 

For this PR I'm fine with `attrs` as a kwarg with a default, but for posterity I want to emphasize the advantages of making it positional.

rabernat:
> Alternatively, we make a breaking change to `init_group` and add `attrs` as a positional argument, and then any higher-level routine that invokes it will break, but this will ensure consistency.

I favor this options. It allows us solve the problem without changing the public API in a way that could break user code. I don't consider `init_group` part of the public API. At https://zarr.readthedocs.io/en/stable/api/storage.html#zarr.storage.init_group it says

> Note that this is a low-level function and there should be no need to call this directly from user code.

d-v-b:
Tests would require a lot of redundant changes to `test_core.py`, so I opened #1305 to set the conditiosn for a much more graceful introduction of the attrs kwarg to the test suite.

jhamman:
@d-v-b - should we close this in favor of `v3`?

d-v-b:
closing in favor of `v3`

Issue #1296, opened 2022-12-15
Read performance regression via Store backed by FSMap (fsspec, GCS)


ravwojdyla:
### Zarr version

2.11.0 and up (including current main)

### Numcodecs version

0.10.2

### Python Version

3.10

### Operating System

Mac and Linux

### Installation

conda, pip and from source

### Description

It appears that https://github.com/zarr-developers/zarr-python/pull/789, commit: [5c71212](https://github.com/zarr-developers/zarr-python/commit/5c71212ccf5abe9f8a5ab7996c5ce3abbbbd61e2) so from zarr 0.11.0, there's a performance regression that affects reading zarr data via Store backed by fsspec/FSMap.

In our test example (in practice we use xarray), we have a zarr array made of 2K files (total 1GB compressed), reading it via:

```py
np.asarray(zarr.open(fsspec.get_mapper(...), mode="r"))
```

* on zarr 0.10.3 took about 12 seconds
* on zarr 0.13.3 took about 90 seconds (so roughly 7x longer)
  * the same problem exists starting from version 0.11.0 

<hr>

Looking at the stacktraces from the different versions, looks like 0.10.3 was asynchronous fetching multiple items, while 0.13.3 is synchronized per storage item?

<details>
<summary>zarr 0.13.3</summary>

```
Thread 3529282 (idle): "MainThread"
    do_futex_wait.constprop.0 (libpthread-2.31.so)
    __new_sem_wait_slow.constprop.0 (libpthread-2.31.so)
    PyThread_acquire_lock_timed.localalias (python3.10)
    lock_PyThread_acquire_lock (python3.10)
    wait (threading.py:324)
    wait (threading.py:607)
    sync (fsspec/asyn.py:86)
    wrapper (fsspec/asyn.py:113)
    __getitem__ (fsspec/mapping.py:143)
    __getitem__ (zarr/storage.py:724)
    _chunk_getitem (zarr/core.py:1966)
    _get_selection (zarr/core.py:1267)
    _get_basic_selection_nd (zarr/core.py:976)
    get_basic_selection (zarr/core.py:933)
    __getitem__ (zarr/core.py:807)
    __array__ (zarr/core.py:589)
    PyArray_FromArrayAttr_int (numpy/core/_multiarray_umath.cpython-310-x86_64-linux-gnu.so)
    _array_from_array_like (numpy/core/_multiarray_umath.cpython-310-x86_64-linux-gnu.so)
[...]

PR #1304, opened 2022-12-20
Handle fsspec.FSMap using FSStore store

ravwojdyla:
Re: https://github.com/zarr-developers/zarr-python/issues/1296

TODO:
* [x] ~Add docstrings and API docs for any new/modified user-facing classes and functions~
* [x] ~New/modified features documented in docs/tutorial.rst~
* [x] ~Changes documented in docs/release.rst~

### DIFF
--- a/zarr/_storage/v3.py
+++ b/zarr/_storage/v3.py
@@ -567,6 +567,16 @@ def _normalize_store_arg_v3(store: Any, storage_options=None, mode="r") -> BaseS
         return store
     if isinstance(store, os.PathLike):
         store = os.fspath(store)
+    if FSStore._fsspec_installed():
+        import fsspec
+        if isinstance(store, fsspec.FSMap):
+            return FSStoreV3(store.root,
+                             fs=store.fs,
+                             mode=mode,
+                             check=store.check,
+                             create=store.create,
+                             missing_exceptions=store.missing_exceptions,
+                             **(storage_options or {}))
     if isinstance(store, str):
         if "://" in store or "::" in store:
             store = FSStoreV3(store, mode=mode, **(storage_options or {}))
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -139,6 +139,16 @@ def _normalize_store_arg_v2(store: Any, storage_options=None, mode="r") -> BaseS
         return store
     if isinstance(store, os.PathLike):
         store = os.fspath(store)
+    if FSStore._fsspec_installed():
+        import fsspec
+        if isinstance(store, fsspec.FSMap):
+            return FSStore(store.root,
+                           fs=store.fs,
+                           mode=mode,
+                           check=store.check,
+                           create=store.create,
+                           missing_exceptions=store.missing_exceptions,
+                           **(storage_options or {}))
     if isinstance(store, str):
         if "://" in store or "::" in store:
             return FSStore(store, mode=mode, **(storage_options or {}))
@@ -1308,6 +1318,8 @@ class FSStore(Store):
                  create=False,
                  missing_exceptions=None,
                  **storage_options):
+        if not self._fsspec_installed():  # pragma: no cover
+            raise ImportError("`fsspec` is required to use zarr's FSStore")
         import fsspec
 
         mapper_options = {"check": check, "create": create}
@@ -1478,6 +1490,13 @@ class FSStore(Store):
         if self.mode == 'r':
             raise ReadOnlyError()
         self.map.clear()
+
+    @classmethod
+    def _fsspec_installed(cls):
+        """Returns true if fsspec is installed"""
+        import importlib.util
+
+        return importlib.util.find_spec("fsspec") is not None
 
 
 class TempStore(DirectoryStore):
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -17,6 +17,8 @@ Unreleased
 * Fix bug that caused double counting of groups in ``groups()`` and ``group_keys()``
   methods with V3 stores.
   By :user:`Ryan Abernathey <rabernat>` :issue:`1228`.
+* Handle fsspec.FSMap using FSStore store
+  By :user:`Rafal Wojdyla <ravwojdyla>` :issue:`1304`.
 
 .. _release_2.13.2:
 
--- a/zarr/tests/test_storage.py
+++ b/zarr/tests/test_storage.py
@@ -2556,8 +2556,13 @@ def test_normalize_store_arg(tmpdir):
         assert isinstance(store, Class)
 
     if have_fsspec:
+        import fsspec
+
         path = tempfile.mkdtemp()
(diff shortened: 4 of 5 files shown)


zarr/_storage/v3.py
```diff
@@ -567,6 +567,16 @@ def _normalize_store_arg_v3(store: Any, storage_options=None, mode="r") -> BaseS
...
+    if FSStore._fsspec_installed():
+        import fsspec
+        if isinstance(store, fsspec.FSMap):
+            return FSStore(store.root,
+                           fs=store.fs,
+                           mode=mode,
+                           check=store.check,
+                           create=store.create,
+                           missing_exceptions=store.missing_exceptions,
+                           **(storage_options or {}))
```
martindurant, line 579: At this point, we switch from the already configured Map and internal FS instance we were given to making new instances with these storage_options, which will in general be different. This is missed in the test, which uses the local filesystem and needs no kwargs. 

(same for the other copy of this call below)

martindurant (commented):
In principle this can be OK, but please see my major comment.

I wonder why this is necessary? Currently the FSMap will be wrapped by KVStore and everything works fine.

ravwojdyla:
> I wonder why this is necessary? Currently the FSMap will be wrapped by KVStore and everything works fine.

@martindurant see https://github.com/zarr-developers/zarr-python/issues/1296. afaiu an alternative to this could be to add set/get items to `KVStore`.

zarr/_storage/v3.py
ravwojdyla: > At this point, we switch from the already configured Map and internal FS instance we were given to making new instances with these storage_options, which will in general be different. This is missed in the test, which uses the local filesystem and needs no kwargs.

@martindurant what's your recommendation?

martindurant:
Ah I see, so previous store.getitems existed, but not when wrapped. So doing what this PR does is worthwhile, but my comment still needs to be solved. Can we have FSStore accept an FSMap instance as an alternative to creating one?

ravwojdyla:
> Can we have FSStore accept an FSMap instance as an alternative to creating one?

@martindurant I don't have a strong opinion about this, but want to point out that right now `FSStore` has a single required argument `url` of type str. There's optional `fs` argument. If we add optional `fs_map` argument, that `fs_map` will already include an "url" (root), so it can be somewhat confusing API (why would the user have an option to pass both `url` and `fs_map`, and what happens when there's discrepancies between these two). wdyt?

martindurant:
OK, so we should pass (mapper.root, fs=mapper.fs), right?

martindurant:
I mean the point is, *not* to pass the storage_options. Maybe it makes no difference, so sorry if I am just confusing things!

martindurant:
OK, I guess you would hit https://github.com/zarr-developers/zarr-python/pull/1304/files#diff-565e487a2f60258b6baa2e4db8ef175cc16b8a949651834bd43d0a9f21e07358R1342 if you tried, never mind...

ravwojdyla:
@martindurant sorry, that link [above](https://github.com/zarr-developers/zarr-python/pull/1304#issuecomment-1359626523) doesn't resolve for me, I probably mean this:

https://github.com/zarr-developers/zarr-python/blob/1af77b63ad8d51a5a8dc2cd923bf73cb8abe5a64/zarr/storage.py#L1329-L1331

So it's not possible to pass both `fs` and `storage_options` ^

martindurant:
Yes that's the one +1

ravwojdyla:
Actually just noticed that need to update this:

image

Should be `FSStoreV3` ^

rabernat:
The use case here is that many users (e.g. @ravwojdyla, others in the Pangeosphere) are used to constructing stores via `fsspec.get_mapper() -> FSMap` and have existing production code that does this; but the FSMap feature set has diverged from FSStore and is therefore less performant.

I support this PR in general because it seems like a lightweight way of upgrading an FSSMap to an FSStore. 

An alternative would be to bring FSMap into feature parity with FSStore.

zarr/_storage/v3.py
```diff
@@ -567,6 +567,16 @@ def _normalize_store_arg_v3(store: Any, storage_options=None, mode="r") -> BaseS
...
+    if FSStore._fsspec_installed():
+        import fsspec
+        if isinstance(store, fsspec.FSMap):
+            return FSStoreV3(store.root,
+                             fs=store.fs,
+                             mode=mode,
+                             check=store.check,
+                             create=store.create,
+                             missing_exceptions=store.missing_exceptions,
+                             **(storage_options or {}))
```
rabernat, line 579: I think it's actually fine to pass `storage_options` through like this. If it is not empty, it will trigger an error via

https://github.com/zarr-developers/zarr-python/blob/1af77b63ad8d51a5a8dc2cd923bf73cb8abe5a64/zarr/storage.py#L1329-L1331

This is the desired behavior. Storage options will already be configured on the `FSMap`, so specifying them twice would not make sense.

zarr/storage.py
```diff
@@ -1308,6 +1318,8 @@ def __init__(self, url, normalize_keys=False, key_separator=None,
                  create=False,
                  missing_exceptions=None,
                  **storage_options):
+        if not self._fsspec_installed():
```
rabernat, line 1321:
```suggestion
        if not self._fsspec_installed():  # pragma: no cover
```

This should keep coverage at 100%

zarr/storage.py
```diff
@@ -1479,6 +1491,13 @@ def clear(self):
             raise ReadOnlyError()
         self.map.clear()
 
+    @classmethod
+    def _fsspec_installed(cls):
+        """Returns true if fsspec is installed"""
+        import importlib.util
+
+        return importlib.util.find_spec("fsspec") is not None
```
rabernat, line 1499: This seems like a suitable way to discover whether fsspec is available.

rabernat (approved):
Thanks a lot for this @ravwojdyla! I see no major issues, just one small suggestion regarding coverage.

joshmoore:
@martindurant, are all your concerns addressed?

From my side, I do start to wonder just looking at the changes needed if fsspec isn't starting to become more than an optional dependency.

martindurant:
To be sure, fsspec is a very small dep and one that has none of its own. I cannot make that decision, however.

Yes, I am happy otherwise.

rabernat:
:+1: to making fsspec a required dependency. However, let's address that in a separate PR. This is @ravwojdyla's first contribution to Zarr, so let's keep it simple.

@ravwojdyla - I invite you to give yourself appropriate credit for this bug fix in `docs/release.rst`. Once that is done we can merge.

ravwojdyla:
@rabernat added the release note in 3a9fab9

joshmoore:
🎉 Thanks, @ravwojdyla 🙏🏽

Issue #1306, opened 2022-12-22
Dask `to_zarr` now complains about read-only after #1304 merged


djhoese:
### Zarr version

`main` branch

### Numcodecs version

0.11.0

### Python Version

3.9

### Operating System

Linux - ubuntu-latest

### Installation

Pip into conda environment with other unstable dependencies

### Description

I'm not sure if there are other changes planned to make this work, but `to_zarr` in dask which calls `create` in zarr now seems to open things as read-only even though it is creating a new file.

### Steps to reproduce

```
In [1]: import dask.array as da

In [2]: a = da.random.random((10, 10), chunks=2)

In [3]: a.to_zarr("/tmp/test.zarr", compute=False)
---------------------------------------------------------------------------
ReadOnlyError                             Traceback (most recent call last)
Input In [3], in <cell line: 1>()
----> 1 a.to_zarr("/tmp/test.zarr", compute=False)

File ~/miniconda3/envs/satpy_py39_unstable/lib/python3.9/site-packages/dask/array/core.py:2926, in Array.to_zarr(self, *args, **kwargs)
   2915 def to_zarr(self, *args, **kwargs):
   2916     """Save array to the zarr storage format
   2917 
   2918     See https://zarr.readthedocs.io for details about the format.
   (...)
   2924     dask.array.to_zarr : equivalent function
   2925     """
-> 2926     return to_zarr(self, *args, **kwargs)

File ~/miniconda3/envs/satpy_py39_unstable/lib/python3.9/site-packages/dask/array/core.py:3685, in to_zarr(arr, url, component, storage_options, overwrite, region, compute, return_stored, **kwargs)
   3681     mapper = url
   3683 chunks = [c[0] for c in arr.chunks]
-> 3685 z = zarr.create(
   3686     shape=arr.shape,
   3687     chunks=chunks,
   3688     dtype=arr.dtype,
   3689     store=mapper,
   3690     path=component,
   3691     overwrite=overwrite,
   3692     **kwargs,
   3693 )
   3694 return arr.store(z, lock=False, compute=compute, return_stored=return_stored)

[...]

PR #1309, opened 2022-12-23
Ensure `zarr.create` uses writeable mode

jrbourbeau:
Closes https://github.com/zarr-developers/zarr-python/issues/1306

cc @ravwojdyla @djhoese

TODO:

### DIFF
--- a/zarr/tests/test_creation.py
+++ b/zarr/tests/test_creation.py
@@ -19,7 +19,7 @@ from zarr.storage import DirectoryStore, KVStore
 from zarr._storage.store import v3_api_available
 from zarr._storage.v3 import DirectoryStoreV3, KVStoreV3
 from zarr.sync import ThreadSynchronizer
-from zarr.tests.util import mktemp
+from zarr.tests.util import mktemp, have_fsspec
 
 _VERSIONS = ((None, 2, 3) if v3_api_available else (None, 2))
 _VERSIONS2 = ((2, 3) if v3_api_available else (2, ))
@@ -427,6 +427,18 @@ def test_create_in_dict(zarr_version, at_root):
 
     a = full(100, 5, store=dict(), **kwargs)
     assert isinstance(a.store, expected_store_type)
+
+
+@pytest.mark.skipif(have_fsspec is False, reason="needs fsspec")
+@pytest.mark.parametrize('zarr_version', _VERSIONS)
+@pytest.mark.parametrize('at_root', [False, True])
+def test_create_writeable_mode(zarr_version, at_root, tmp_path):
+    # Regression test for https://github.com/zarr-developers/zarr-python/issues/1306
+    import fsspec
+    kwargs = _init_creation_kwargs(zarr_version, at_root)
+    store = fsspec.get_mapper(str(tmp_path))
+    z = create(100, store=store, **kwargs)
+    assert z.store.map == store
 
 
 @pytest.mark.parametrize('zarr_version', _VERSIONS)
--- a/zarr/creation.py
+++ b/zarr/creation.py
@@ -145,7 +145,7 @@ def create(shape, chunks=True, dtype=None, compressor='default',
         zarr_version = getattr(chunk_store, '_store_version', DEFAULT_ZARR_VERSION)
 
     # handle polymorphic store arg
-    store = normalize_store_arg(store, zarr_version=zarr_version)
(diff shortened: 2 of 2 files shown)


jrbourbeau:
Can confirm the changes here also fix the related test failures we were seeing in Dask's CI

ravwojdyla (approved):

zarr/tests/test_creation.py
```diff
@@ -429,6 +429,18 @@ def test_create_in_dict(zarr_version, at_root):
     assert isinstance(a.store, expected_store_type)
 
 
+@pytest.mark.skipif(have_fsspec is False, reason="needs fsspec")
```
rabernat, line 432: Is fsspec needed to reproduce this bug? In my reading of the code, it it should only affect stores that have a `mode` argument in their constructor. That is:

- `FSStore`
- `ZipStore`
- `DBMStore`

The fact that not all stores accept the `mode` keyword it, to me, a Zarr code smell.

rabernat (approved):
Thanks a lot @jrbourbeau for a quick fix of this regression.

I'm eager to help fix the problem. At the same time, I want to understand this bug a bit better in order to avoid creating additional technical debt within zarr-python.

When I first saw @djhoese's issue, I was puzzled by the recommend fix. The example was using local file storage. How is it possible that creating arrays in local files is completely broken in zarr? Surely our test suite would have caught such a regression.

I dug deeper and realized that dask.array.to_zarr is always converting string arguments to FSMap objects (see https://github.com/dask/dask/blob/40bc376acb1f631aa80db17ce4fb779e1a888aee/dask/array/core.py#L3677-L3681) using code like

```python
    if isinstance(url, str):
        mapper = get_mapper(url, **storage_options)
```

FSMap has been effectively deprecated by the introduction of zarr.storage.FSStore (beginning with #546). FSStore has additional performance optimization that are not available with FSMap, and therefore we always prefer to use an FSStore (although an FSMap is still a valid store, as it is a valid MutableMapping).

To address this inconsistency, #1304 recently implemented a change to automatically promote an FSMap to an FSStore (thanks @ravwojdyla! 🙌 ). That change is what revealed this latent bug. Before, the FSMap was initialized with whatever mode the user chose (or the default from fsspec, presumably  `mode='w'`). Now we override this with the default keyword argument from `normalize_store_arg` which is indeed `mode='r'`:

https://github.com/zarr-developers/zarr-python/blob/4e633ad9aa434304296900790c4c65e0fa0dfa12/zarr/storage.py#L144-L147

`mode` is not an attribute of an FSMap, but it is an attribute of FSStore. It is used internally within Zarr to implement read-only behavior, e.g.

https://github.com/zarr-developers/zarr-python/blob/4e633ad9aa434304296900790c4c65e0fa0dfa12/zarr/storage.py#L1409-L1411

My concern is that the `mode` argument has a somewhat ambiguous status within Zarr. It is not supported by all stores, but it could be. That inconsistency is what opened the possibility for this bug in the first place.

Having written all that, I think I have convinced myself that I am fine with this fix. 🙃 

The two follow-up actions I would suggest are:
- Do an audit on the status of `mode` in Zarr. All stores could potentially support mode in the same way that FSStore does. However, we also support a different, redundant method of specifying read-only: via the `Array` and `Group` constructors. `DirectoryStore`, for example, doesn't have `mode` at all, but it is still possible to use read-only stores on disk. Check out [`open_array`](https://github.com/zarr-developers/zarr-python/blob/main/zarr/creation.py#L397) for more details.
- Consider refactoring `dask.array.to_zarr`. Now that Zarr implements url parsing internally via fsspec, it should be possible to leave the store creation completely up to Zarr, rather than Dask.

rabernat:
@jrbourbeau - if you can push a release notes update, I will merge this.

martindurant:
Thanks for the thorough thoughts @rabernat . Indeed, storage code logic has been slowly drifting into zarr, which seems like the right direction to go in.

zarr/tests/test_creation.py
joshmoore: @rabernat: would you hold up this PR for using `fsspec` for the regression test?

Also: would you propose an issue to say, "all stores should have a `mode` argument"?

joshmoore:
Just realizing that 2.13.4 is likely broken since #1304 is included but this is not!

jrbourbeau:
Thanks @rabernat @joshmoore!

PR #1312, opened 2022-12-27
Allow reading utf-8 encoded json files 

### LINKED ISSUE #1308: Cannot read attributes that contain non-ASCII characters
### Zarr version

2.13.4.dev68

### Numcodecs version

0.11.0

### Python Version

3.10.6

### Operating System

Linux

### Installation

With pip, using the instructions here https://zarr.readthedocs.io/en/stable/contributing.html

### Description

I expect zarr to be able to read attribute files that contain non-ASCII characters, because JSON files use utf-8 encoding. However, currently, there is a check to error if the JSON file contains any non-ASCII characters.

### Steps to reproduce

```python
import zarr
import tempfile
tempdir = tempfile.mkdtemp()
f = open(tempdir + '/.zgroup','w', encoding='utf-8')
f.write('{"zarr_format":2}')
f.close()
f = open(tempdir + '/.zattrs','w', encoding='utf-8')
f.write('{"foo": "た"}')
f.close()
z = zarr.open(tempdir, mode='r')
z.attrs['foo']
```

```python
  File "/root/github/zarr-python/zarr/attrs.py", line 74, in __getitem__
    return self.asdict()[item]
  File "/root/github/zarr-python/zarr/attrs.py", line 55, in asdict
    d = self._get_nosync()
  File "/root/github/zarr-python/zarr/attrs.py", line 48, in _get_nosync
    d = self.store._metadata_class.parse_metadata(data)
  File "/root/github/zarr-python/zarr/meta.py", line 104, in parse_metadata
    meta = json_loads(s)
  File "/root/github/zarr-python/zarr/util.py", line 56, in json_loads
    return json.loads(ensure_text(s, 'ascii'))
  File "/root/pyenv/zarr-dev/lib/python3.10/site-packages/numcodecs/compat.py", line 181, in ensure_text
    s = codecs.decode(s, encoding)
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe3 in position 9: ordinal not in range(128)
```

### Additional output

_No response_

nhz2:
Fixes #1308 

Currently, zarr-python errors when reading a json file with non-ascii characters encoded with utf-8, however, Zarr.jl writes json files that include non-ascii characters using utf-8 encoding.
This PR will enable zarr attributes written in Zarr.jl to be read by zarr-python.

TODO:


nhz2:
@MSanKeys963 @jakirkham  This PR is ready for review. Also, because I am a first-time contributor, I am not able to trigger Github Actions myself.

MSanKeys963:
Thanks for sending the PR @nhz2.
@jakirkham, @joshmoore and @rabernat, please have a look.

joshmoore:
Thanks for this, @nhz2! I've triggered the actions and will go through the code ASAP.

joshmoore:
Green :tada: in case anyone else wants to now take a closer look. (`release.rst` will need its conflict handled)

nhz2:
This PR should have no effect on the JSON files zarr-python writes, they will still be ASCII only with non-ASCII characters escaped. It only affects what kind of JSON files zarr-python can read.

joshmoore:
Thanks, @nhz2. Guess I was more concerned whether any calling code (and/or subclasses) would need modification.

joshmoore:
Thanks for the analysis, @nhz2! Sounds like it should not impact external consumers.

Any objections to rolling this into 2.14, anyone?

zarr/meta.py
```diff
@@ -92,7 +92,7 @@ class Metadata2:
     ZARR_FORMAT = ZARR_FORMAT
 
     @classmethod
-    def parse_metadata(cls, s: Union[MappingType, str]) -> MappingType[str, Any]:
+    def parse_metadata(cls, s: Union[MappingType, bytes]) -> MappingType[str, Any]:
```
jakirkham, line 95: Should this still include `str` as an option? Same question with the functions below

```suggestion
    def parse_metadata(cls, s: Union[MappingType, bytes, str]) -> MappingType[str, Any]:
```

jakirkham:
Thanks for the PR Nathan! 🙏

This is a good improvement to include.

Noting some changes in the typing to still allow `str` in functions (since it should still technically be valid). Though agree `bytes` should be included in a few places as well for clarity

Ideally we would capture any `bytes`-like type here (since they are all valid), but that is somewhat complicated to do in practice. Maybe there will be guidance on this in the future ( https://github.com/python/typing/issues/593 )

jakirkham (approved):

joshmoore:
I think this has failed on the zarr-feedstock: https://github.com/conda-forge/zarr-feedstock/pull/72

zarr/tests/test_attrs.py
```diff
@@ -42,11 +43,17 @@ def test_storage(self, zarr_version):
...
+        d = json.loads(str(store[attrs_key], 'utf-8'))
         if zarr_version == 3:
             d = d['attributes']
         assert dict(foo='bar', baz=42) == d
 
+    def test_utf8_encoding(self, zarr_version):
+
+        # fixture data
+        fixture = group(store=DirectoryStore('fixture'))
+        assert fixture['utf8attrs'].attrs.asdict() == dict(foo='た')
```
jakirkham, line 55: This test is failing there

fixture/utf8attrs/.zattrs
```diff
@@ -0,0 +1 @@
+{"foo": "た"}
```
jakirkham, line 1: This file is missing from the `sdist` on PyPI

fixture/utf8attrs/.zgroup
```diff
@@ -0,0 +1,3 @@
+{
+    "zarr_format": 2
+}
```
jakirkham, line 3: Same with this one

jakirkham (commented):
The issue is neither of these files were included in the Python `sdist`

fixture/utf8attrs/.zattrs
joshmoore: Thanks, @jakirkham. Argh, I thought I had caught these issues but likely only if there's code to generate it them as well. Are you already working on a fix or shall I?

jakirkham: Feel free to go ahead. I might not get to this for a while

jakirkham: Forget if we already discussed this before, but maybe we can make some fixes to CI to ensure we catch these issues in PRs. Wrote up some thoughts on this in issue ( https://github.com/zarr-developers/zarr-python/issues/1347 )

PR #1358, opened 2023-02-27
merge consecutive reads in shards

jstriebel:
This PR adds a feature for the v3 sharding implementation. When multiple chunks are read at once with a store that supports partial reads, each chunk was requested separately before, even if it was in the same shard. This PR adds the feature to merge consecutive (on a byte-level) requests in the same shard.

TODO:

### DIFF
--- a/zarr/_storage/v3_storage_transformers.py
+++ b/zarr/_storage/v3_storage_transformers.py
@@ -340,12 +340,55 @@ class ShardingStorageTransformer(StorageTransformer):  # lgtm[py/missing-equals]
                     )
                 else:  # pragma: no cover
                     transformed_key_ranges.append((key, range_))
-            values = self.inner_store.get_partial_values(transformed_key_ranges)
+            merged_key_ranges, reverse_lookup = self._merge_ranges(transformed_key_ranges)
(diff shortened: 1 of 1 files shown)


rabernat:
Thanks for working on this important optimization! @martindurant implemented something similar at the fsspec level in https://github.com/fsspec/filesystem_spec/pull/1063. I'm curious how this implementation here would interact with that one. Where in this stack should this optimization live?

normanrz:
Good point. We weren't aware of that optimization. I almost think it makes more sense in the fsspec. @jstriebel wdyt?

rabernat:
> I almost think it makes more sense in the fsspec.

That's a reasonable point of view. However, currently not all zarr I/O goes through fsspec. This is a rather major architectural issue that we need to confront for zarr-python: how dependent do we want to be on fsspec? If we require fsspec as a core dependency, we could remove quite a bit of code from zarr-python itself. Currently, we're somewhere in the middle--we rely on fsspec for some things, but we still have many of our own implementations for different stores.

We need to discuss this as a team.

martindurant:
Totally agree with @rabernat , of course you might know which way I lean.

For merging the requests via fsspec, this should be available via the cat_ranges() method - it is intended but not implemented yet.

PR #1378, opened 2023-03-27
Update release.rst for the upcoming release - 2.15.0

MSanKeys963:
Updated the `release.rst` for the upcoming 2.15.0 release. Please have a look.

CC: @joshmoore 

TODO:

### DIFF
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -15,8 +15,28 @@ Unreleased
     # .. warning::
     #    Pre-release! Use :command:`pip install --pre zarr` to evaluate this release.
 
+.. _release_2.15.0:
+
+2.15.0
+------
+
+Enhancements
+~~~~~~~~~~~~
+
 * Implement more extensive fallback of getitem/setitem for orthogonal indexing.
   By :user:`Andreas Albert <AndreasAlbertQC>` :issue:`1029`.
+
+Documentation
+~~~~~~~~~~~~~
+
+* Add API reference for V3 Implementation in the docs.
+  By :user:`Sanket Verma <MSanKeys963>` :issue:`1345`.
+
+Bug fixes
+~~~~~~~~~
+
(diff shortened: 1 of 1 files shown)


MSanKeys963:
Does any of the @zarr-developers/python-core-devs want to have a look at this and assist with the `2.15.0` release?
I'd appreciate it.

martindurant (approved):

martindurant:
Is this waiting on anything?

martindurant:
@jakirkham @joshmoore - now this is in, we should actually release. I have never released zarr before, but I can have a go if no one else is available.

MSanKeys963:
Hi, @martindurant.

I had a quick chat with @jakirkham yesterday, and we're thinking of going with an alpha release (2.15.0a1) rather than `2.15.0` since #1131 just got in.

What do you think?

martindurant:
Honestly, I don't think anyone will try an alpha that wouldn't be happy to install directly from github.

martindurant:
Since 2.15.0 isn't tagged yet, why not just include the extra PR? It's opt-in, won't change behaviour for users.

MSanKeys963:
@joshmoore, @jakirkham and @rabernat: Any thoughts here?

joshmoore:
If it's in before the tag, then it's definitely "included". It will need the release notes though.

jakirkham:
For context, Sanket and I discussed releasing. Had suggested we go with an alpha to kick the tires on the changes in PR ( https://github.com/zarr-developers/zarr-python/pull/1131 ). That way we are not beholden to a particular API (should it need reworking).

Suppose other ways to accomplish this would be to mark that API as experimental in the docs and possibly elsewhere (at least until we've gotten more confidence with it)

Maybe there are other approaches

jakirkham:
Raising an issue to discuss releasing ( https://github.com/zarr-developers/zarr-python/issues/1401 )

PR #1421, opened 2023-05-28
io.open → open

DimitriPapadopoulos:
In Python 3, `io.open()` is an alias for the builtin `open()` function:
https://docs.python.org/3/library/io.html#io.open

TODO:

### DIFF
--- a/zarr/convenience.py
+++ b/zarr/convenience.py
@@ -1,5 +1,4 @@
 """Convenience functions for storing and loading data."""
-import io
 import itertools
 import os
 import re
@@ -28,6 +27,8 @@ from zarr.util import TreeViewer, buffer_size, normalize_storage_path
 from typing import Union
 
 StoreLike = Union[BaseStore, MutableMapping, str, None]
+
+_builtin_open = open  # builtin open is later shadowed by a local open function
 
 
 def _check_and_update_path(store: BaseStore, path):
@@ -491,7 +492,7 @@ class _LogWriter:
         elif callable(log):
             self.log_func = log
         elif isinstance(log, str):
-            self.log_file = io.open(log, mode="w")
+            self.log_file = _builtin_open(log, mode="w")
             self.needs_closing = True
         elif hasattr(log, "write"):
             self.log_file = log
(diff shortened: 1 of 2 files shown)


DimitriPapadopoulos:
That's disturbing. Despite `io.open()` being (supposedly) an alias for `open()`, changing to `open()` fails. The error message is unclear, that's perhaps a real issue.

DimitriPapadopoulos:
No unexpected failures locally 🤯

Any clue?

DimitriPapadopoulos:
This should be fixed by changing the name of the local function `open`, for example `zarr_open`.

Then change:
https://github.com/zarr-developers/zarr-python/blob/4132f360616a4c8bfa3dd4e979a4793c5d84cdfc/zarr/__init__.py#L3-L5
to:
```python
from zarr.convenience import (consolidate_metadata, copy, copy_all, copy_store,
                              load, open_consolidated, save, save_array,
                              save_group, tree)
from zarr.convenience import zarr_open as open
```

d-v-b:
@DimitriPapadopoulos 
> This should probably be fixed by changing the name of the local function open, perhaps to zarr_open.

I have no objection to this strategy. Do you want to pursue that in this PR, or elsewhere?

zarr/convenience.py
```diff
@@ -491,7 +492,7 @@ def __init__(self, log):
         elif callable(log):
             self.log_func = log
         elif isinstance(log, str):
-            self.log_file = io.open(log, mode="w")
+            self.log_file = _builtin_open(log, mode="w")
```
rabernat, line 495: Are we certain this is the only use of `io.open`?

DimitriPapadopoulos: I think so. A grep of `io.open` yields:
```
$ fgrep -R io.open --exclude-dir=.git
zarr/convenience.py:            self.log_file = io.open(log, mode="w")
build/lib/zarr/convenience.py:            self.log_file = io.open(log, mode='w')
$ 
```
Do you have a specific concern, anything I may have overlooked?

rabernat (approved):
Thanks a lot for this PR!

zarr/convenience.py
rabernat: No not really. Just wanted to make sure we only have to do this once.

PR #1423, opened 2023-05-31
Change occurrences of % and format() to f-strings

DimitriPapadopoulos:
The rationale is that f-strings are faster.

TODO:


d-v-b:
I will fix these conflicts, and then I think this should be good to go in.

d-v-b:
@rabernat any objection to getting this in to the release?

zarr/convenience.py
```diff
@@ -497,7 +497,7 @@ def __init__(self, log):
             self.log_file = log
         else:
             raise TypeError(
-                "log must be a callable function, file path or " "file-like object, found %r" % log
```
DimitriPapadopoulos, line 500: This string was cut in half, looks like a left-over from #1459.

zarr/convenience.py
```diff
@@ -756,7 +754,7 @@ def copy(
     log=None,
     if_exists="raise",
     dry_run=False,
-    **create_kws
+    **create_kws,
```
DimitriPapadopoulos, line 757: That's not me, `black` did it.

zarr/core.py
```diff
@@ -2518,7 +2518,7 @@ def bytestr(n):
         # filters
         if self.filters:
             for i, f in enumerate(self.filters):
-                items += [("Filter [%s]" % i, repr(f))]
```
DimitriPapadopoulos, line 2521: Only the first argument `i` was printed, but I think the main intent mostly was to print `repr(f)`. Please advise.

zarr/convenience.py
```diff
@@ -949,7 +943,7 @@ def _copy(log, source, dest, name, root, shallow, without_attrs, if_exists, dry_
         if do_copy:
 
             # log a message about what we're going to do
-            log("copy {} {} {}".format(source.name, source.shape, source.dtype))
```
DimitriPapadopoulos, line 952: Perhaps `log("{}".format(arg))` should actually be changed to `log("%s", arg)`, so that string formatting interpolation is not executed if logging is not enabled.
* Some developers argue that deferring that string interpolation in the logging function, after checking whether logging is enabled, is a perceptible speed win if logging is disabled. See Pylint's [logging-fstring-interpolation / W1203](https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/logging-fstring-interpolation.html).
* Other developers argue that the speed win is hardly perceptible because the cost of a function call is much higher, while the speed win of f-strings if logging is enabled is more perceptible. See https://github.com/python/cpython/issues/90358.

See also:
* https://docs.python.org/3/howto/logging.html#optimization

jhamman (approved):

joshmoore:
Conflicts fixed. @DimitriPapadopoulos: please note it looks like https://github.com/zarr-developers/zarr-python/pull/1423/commits/46989fd72ed251a6d41628887a3d86348b9cbf08#diff-35a6b5d683684d4a73baf73ec18e5a438a738b98e3138b494834bca5add5ba4bR1399 was missing an `f`?

DimitriPapadopoulos:
> Conflicts fixed. @DimitriPapadopoulos: please note it looks like [46989fd#diff-35a6b5d683684d4a73baf73ec18e5a438a738b98e3138b494834bca5add5ba4bR1399](https://github.com/zarr-developers/zarr-python/commit/46989fd72ed251a6d41628887a3d86348b9cbf08#diff-35a6b5d683684d4a73baf73ec18e5a438a738b98e3138b494834bca5add5ba4bR1399) was missing an `f`?

Indeed, sorry about that. Because `pyupgrade` cannot be restricted to a specific rule and misses some cases, it wasn't easy to automate such changes. I guess `ruff` is a better alternative because you can `--select` specific rules and `--fix` issues.

Issue #991, opened 2022-03-23
Proposal: Add Array.blocks using new BlockIndexer (Prototype Code Included)


tasansal:
#### Problem description

I have a use case downstream where we want to access "blocks" of chunks. I have implemented a prototype that functions like [`dask.array.Array.blocks`](https://docs.dask.org/en/latest/generated/dask.array.Array.blocks.html#dask-array-array-blocks).
It uses the existing `zarr.indexing` machinery and matches the API of existing indexers.

This allows us to pull a "block" of chunks from data using slicing logic.
For instance, if we have an array with shape `(10, 20, 30)` with chunk sizes `(5, 4, 10)`

- `array.blocks[0]` maps to `array[:5]`
- `array.blocks[..., 0]` maps to `array[:, :, :10]`
- `array.blocks[1, 1, 1]` maps to `array[5:10, 4:8, 10:20]`
- `array.blocks[:, 1:4, :]` maps to `array[:, 4:16, :]`

Why not just use `dask` array and method?
- We don't want to have `dask` as a hard requirement in our library.
- In many simple queries, vanilla zarr is much faster than converting to `dask.array` + scheduling overhead.
- We don't want to install all dependencies of `dask` for just this functionality to keep our library lightweight.

If there are no objections I can start adding this in a new PR as soon as possible.
If we don't want this in `zarr`, I can keep this in our library as an extension to zarr.

Once the code below are implemented, all of the following evaluates to `True` as expected.
```python
np.array_equal(array.blocks[0], array[:5])
np.array_equal(array.blocks[..., 0], array[:, :, :10])
np.array_equal(array.blocks[1, 1, 1], array[5:10, 4:8, 10:20])
np.array_equal(array.blocks[:, 1:4, :], array[:, 4:16, :])
```

Tests and docs must be written, of course, I don't have that yet.

#### Implementation Details
Methods and attributes that would go into `zarr.core.Array`
```python
class Array:
    def __init__(...):
        ...
        self._blocks = Blocks(self)

    @property
    def blocks(self):
        return self._blocks

    def get_block_selection(self, selection, out=None, fields=None):
        if not self._cache_metadata:
            self._load_metadata()

        # check args
        check_fields(fields, self._dtype)

        # setup indexer
        indexer = BlockIndexer(selection, self)

        return self._get_selection(indexer=indexer, out=out, fields=fields)

    def set_block_selection(self, selection, value, fields=None):
        # guard conditions
        if self._read_only:
[...]

Issue #545, opened 2020-03-09
Utility to turn block-ID into equivalent slice 


jhamman:
Suppose I have a array block id (e.g. "0.0.0") and I want its equivalent slice on the full array. Does Zarr currently have a function to do this? I'm looking for something like this:

```python
In [1]: block_id = '0.0.0'

In [2]: block_size = (10, 20, 30)

In [3]: def slice_axis(key, chunk_size):
   ...:     return slice(key * chunk_size, (key + 1) * chunk_size)
   ...:
   ...:
   ...: def get_indexers(key, chunks):
   ...:     ikeys = key.split(".")
   ...:     return tuple(slice_axis(int(i), c) for i, c in zip(ikeys, chunks))
   ...:

In [4]: get_indexers(block_id, block_size)
Out[4]: (slice(0, 10, None), slice(0, 20, None), slice(0, 30, None))

In [5]: import numpy as np

In [6]: data = np.random.random((100, 100, 100))

In [7]: slices = get_indexers(block_id, block_size)

In [8]: data[slices]
```

I recognize this is actually the inverse operation that Zarr takes when slicing a zarr array (usually we go from slice -> block_id).

jakirkham:
Similar to issue ( https://github.com/zarr-developers/zarr-python/issues/543 ).

Issue #543, opened 2020-03-03
API for direct block access


clbarnes:
There are internal methods for retrieving individual blocks, but there are some circumstances where addressing data one block at a time is helpful for end users, and would avoid the user having to do their own pipeline of chunk size -> block indices -> slicing only for zarr to then go slicing -> block indices etc. again.

I envision something like 

```python
@dataclass
class ChunkWrapper:
    chunk_idx: Tuple[int, ...]
    chunk_slice: Tuple[slice, ...]  # or an offset-shape pair, or a start-stop pair
    data: np.ndarray

class Array:
    ...
    def get_chunk(self, chunk_idx: Tuple[int, ...]) -> ChunkWrapper:
        ...

    def set_chunk(self, chunk_idx: Tuple[int, ...], data: np.ndarray) -> None:
        # check data is the right shape, handling edge blocks
        ...

    def iter_chunk_idxs(self) -> Iterator[ChunkWrapper]:
        ...
```

Then e.g. a blockwise operation could be trivially implemented with

```python
for idx in my_array.iter_chunk_idxs():
    chunk = my_array.get_chunk(idx)
    my_array.set_chunk(idx, chunk.data  * 2)
```

Obviously in this particular case, you could use dask, but the principle is useful elsewhere. My use case is that I have an array of labels which I want to relate to point annotations: I want to get a chunk, see which point annotations exist inside it, and find the relationships, preferably without chunk-mangling boilerplate :grin:

alimanfoo:
Hi @clbarnes, could `get_chunk()` return the numpy array with chunk data directly, no need for a wrapper class? E.g.:

```
for idx in my_array.iter_chunk_idxs():
    chunk = my_array.get_chunk(idx)
    my_array.set_chunk(idx, chunk * 2)
```

clbarnes:
It could, although working with blocks like this before, I've found myself shipping round the same tuple of index and data to different functions so thought it might be handy to have them in the same place. Not critical, though. You could get the convenience of both in one go by subclassing np.ndarray to add a chunk_idx member, although that introduces more maintenance overhead.

PR #1428, opened 2023-06-08
Support Block (Chunk) Indexing

### LINKED ISSUE #518: Enhancement: Chunk level access api / indexing by chunk rather than voxel
I may be missing something, but zarr currently seems to offers a high level interface to the chunked data that appears to be largely transparent to the fact that the underlying data is in fact chunked (much like HDF5). I've got a few use cases where I'd like to be able to directly operate on the underlying chunks rather than the whole large dataset.

My target application is multi-dimensional image processing in large 3D biomedical datasets which there are many situations where it would make sense to perform operations on individual chunks in parallel. 

Whilst it would be possible to a) read the chunk size of an array and b) work out how to slice the dataset in multiples of that chunk size, a direct chunk level access might be easier and more efficient. In essence I'm suggesting something which exposes a simplified version of `Array._chunk_getitem`  and `Array._chunk_setitem` which would only ever get  or set a complete chunk.

If you extended this concept somewhat by having the functions return a lightweight`Chunk` object which was essentially a reference to the location of the array, a chunk ID, and a `.data` property which gave you the data for that chunk and additionally exposed an iterator in the `Array` class , you could conceivably write code like:

```python
def do_something(chunk):
    res = some_processing_function(chunk.data)
    with zarr.open('output_file_uri') as z1:
        z1.save_chunk(res, chunk.chunk_id)

with multiprocessing.Pool() as pool:
    pool.map(do_something, array.chunk_iterator)
```

In the longer term (and I'm not sure how to go about this - I might be better aiming for API compatibility with zarr rather than inclusion in zarr)  I'd want to enable data-local processing of individual chunks of an array which was chunked across a distributed file system. I've currently got a python solution for this for 2.5 dimensional data (xyt) but it's pretty specific to one use case and I would like to avoid duplicating other efforts as we make it more general.

tasansal:
Closes #518, #543, #545, #991.

This feature has been requested many times.
This adds the basic functionality of Dask's [`dask.array.Array.blocks`](https://docs.dask.org/en/stable/generated/dask.array.Array.blocks.html) indexing.

I explained the implementation in detail in #991. Please see details there.

This PR implements #991 and adds tests, tutorial, and docs.

This does allow integer and slice (with step of 1) block indexing, or a mix of the two, for Zarr arrays.

**QUESTION**
* I wasn't exactly sure where to put it in the change log. Currently its in 2.15.0, however, should I move that to Unreleased section??

TODO:
* [X] New/modified features documented in docs/tutorial.rst
* [X] GitHub Actions have all passed
* [X] Test coverage is 100% (Codecov passes)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ __pycache__/
 # Distribution / packaging
 .Python
 env/
+.venv/
 build/
 develop-eggs/
 dist/
--- a/docs/api/core.rst
+++ b/docs/api/core.rst
@@ -10,6 +10,8 @@ The Array class (``zarr.core``)
     .. automethod:: set_basic_selection
     .. automethod:: get_mask_selection
     .. automethod:: set_mask_selection
+    .. automethod:: get_block_selection
+    .. automethod:: set_block_selection
     .. automethod:: get_coordinate_selection
     .. automethod:: set_coordinate_selection
     .. automethod:: get_orthogonal_selection
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -31,6 +31,8 @@ Enhancements
 * ``open_array()`` now takes the ``meta_array`` argument.
   By :user:`Mads R. B. Kristensen <madsbk>` :issue:`1396`.
 
+* **Block Indexing**: Implemented blockwise (chunk blocks) indexing to ``zarr.Array``.
+  By :user:`Altay Sansal <tasansal>` :issue:`1428`
 Maintenance
 ~~~~~~~~~~~
 
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -11,6 +11,9 @@ Release notes
     Unreleased
     ----------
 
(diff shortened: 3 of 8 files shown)


joshmoore:
> * I wasn't exactly sure where to put it in the change log. Currently its in 2.15.0, however, should I move that to Unreleased section??

Yes please, and update your comments of `2.15` to be `2.16`? Huge thanks!

tasansal:
> > * I wasn't exactly sure where to put it in the change log. Currently its in 2.15.0, however, should I move that to Unreleased section??
> 
> Yes please, and update your comments of `2.15` to be `2.16`? Huge thanks!

Done! Seems like all the CI passed too.

tasansal:
@joshmoore ready to merge?

joshmoore:
In general, this looks really well done. I'd still ping a few daskanistas (daskonauts?) for some feedback.

cc: @jrbourbeau @jni but of course others welcome!

jni (approved):

rabernat (approved):
This is a great new feature.

PR #1447, opened 2023-06-27
[V3] Allow for incomplete codec metadata using numcodecs.get_codec

### LINKED ISSUE #1443: Will V3 compressor codecs require all parameters in array metadata?
### Zarr version

v2.14.3

### Numcodecs version

v0.11.0

### Python Version

3.10

### Operating System

Mac

### Installation

pip

### Description

The current V3 implementation seems to require a more specific set of parameters for codecs/compressors than V2 did. The code sample below demonstrates how all configuration parameters are extracted separately:

https://github.com/zarr-developers/zarr-python/blob/b3cafa96b88c362e7843df8e2a07499588b7fa49/zarr/meta.py#L443-L460

Whereas in V2, the compressor parameters were simply passed on to [`numcodecs.get_config`](https://github.com/zarr-developers/numcodecs/blob/4f194b661faf1f945cb21b1ac1f19c4314ff8f05/numcodecs/registry.py#L51C1-L52C1)

https://github.com/zarr-developers/zarr-python/blob/b3cafa96b88c362e7843df8e2a07499588b7fa49/zarr/meta.py#L169-L171

## Question:
Was this an intentional change? If so, I foresee a number of downstream projects that will need to adjust their expectations (e.g. Kerchunk, Tifffile). 

### Steps to reproduce

I ran into this working on a V2->V3 conversion tool for Kerchunk. A minimal example is not available at this time.

xref: https://github.com/cgohlke/tifffile/issues/211

jhamman:
This PR refactors how codecs are instantiated in the current V3 implementation. This allows for default parameters to be omitted.

TODO:

closes #1443

### DIFF
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -13,10 +13,13 @@ Release notes
     # to document your changes. On releases it will be
     # re-indented so that it does not show up in the notes.
 
-    .. _unreleased:
+.. _unreleased:
 
-    Unreleased
-    ----------
+Unreleased
+----------
+
+* Allow for partial codec specification in V3 array metadata.
+  By :user:`Joe Hamman <jhamman>` :issue:`1443`.
 
 .. _release_2.15.0:
 
--- a/requirements_dev_optional.txt
+++ b/requirements_dev_optional.txt
@@ -14,7 +14,7 @@ types-setuptools
 pymongo==4.4.0
 # optional test requirements
 coverage
-pytest-cov==4.0.0
+pytest-cov==4.1.0
 pytest-doctestplus==0.13.0
 pytest-timeout==2.1.0
 h5py==3.9.0
--- a/zarr/meta.py
+++ b/zarr/meta.py
@@ -441,25 +441,21 @@ class Metadata3(Metadata2):
         uri = 'https://purl.org/zarr/spec/codec/'
         conf = meta['configuration']
         if meta['codec'].startswith(uri + 'gzip/'):
-            codec = numcodecs.GZip(level=conf['level'])
+            conf["id"] = "gzip"
         elif meta['codec'].startswith(uri + 'zlib/'):
-            codec = numcodecs.Zlib(level=conf['level'])
+            conf["id"] = "zlib"
         elif meta['codec'].startswith(uri + 'blosc/'):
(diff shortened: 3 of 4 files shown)


rabernat (changes requested):
I am in favor of making the handling of codec configuration the same in V2 vs V3.

I do think this needs a test to verify that the situation described in https://github.com/cgohlke/tifffile/issues/211 is handled correctly after this change.

jhamman:
Test failures are unrelated and are fixed here https://github.com/zarr-developers/zarr-python/pull/1450

jhamman:
@zarr-developers/python-core-devs - looking for some final feedback here. If I don't hear any, I'll plan to merge ~7/12.

martindurant:
LGTM. I suppose if the conf has invalid values, you get an exception when the codec is instantiated -  do we have any expectation/test of how that should go?

For the related issue in TIFFfile, I will comment that many codec options, specifically for compressors, are only actually important at compress time, not when reading. It would be useful to know, though, if the defaults in python implementations are compatible with whatever TIFF assumes - I don't know how to do that rigorously.

cgohlke:
> It would be useful to know, though, if the defaults in python implementations are compatible with whatever TIFF assumes - I don't know how to do that rigorously.

The only compression method specified in the TIFF standard that has a codec in the numcodecs package is "deflate". The spec says that "each image segment consists of a single complete zlib data stream.". This is compatible what numcodecs.Zlib handles.

All the other dozens of compression methods found in TIFF are either nowhere specified specifically for TIFF (e.g. LZMA, Zstd) , not supported by numcodecs (PackBits, LZW, CCITT, JPEG), or both (e.g., JPEG2000, WebP, PNG, JetRaw).

The imagecodecs package provides many of those codecs made for TIFF. Few of them might require setting extra arguments, such as JPEG (`jpegtables`), WebP (`hasalpha`), or JetRaw (`shape`). This is handled through tifffile. If not, open an issue there.

rabernat:
@jhamman - we just need release notes for this. That's important, since this change could hypothetically impact existing code.

martindurant:
@cgohlke , thank you for the thorough reply. It seems we have nothing to worry about here.

PR #1454, opened 2023-07-10
Add __contains__ method to KVStore

cgohlke:
Over at https://github.com/bayer-science-for-a-better-life/tiffslide/issues/72, we noticed that reading from a `tifffile.ZarrTiffStore` calls the store's `__getitem__` member function twice for each chunk. This is due to `k in self` below being routed through `KVStore.__getitem__`. 

https://github.com/zarr-developers/zarr-python/blob/8c98f4518ea20251c8ef3258276860a3cbef9eeb/zarr/_storage/store.py#L160

This patch adds a `KVStore.__contains__` member function, which does not require reading and decoding chunks for the membership test.

Using this patch, performance almost doubled using the benchmarks in the linked issue.

TODO:
* [ ] ~Add unit tests and/or doctests in docstrings~

### DIFF
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -13,10 +13,15 @@ Release notes
     # to document your changes. On releases it will be
     # re-indented so that it does not show up in the notes.
 
-    .. _unreleased:
+.. _unreleased:
 
-    Unreleased
-    ----------
+Unreleased
+----------
+
+Bug fixes
+~~~~~~~~~
+
+* Add ``__contains__`` method to ``KVStore``. By :user:`Christoph Gohlke <cgohlke>` :issue:`1454`.
 
 .. _release_2.15.0:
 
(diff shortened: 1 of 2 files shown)


rabernat (approved):
Thanks a lot for this @cgohlke!

Having also worked with the various store classes exposed by zarr-python, I have to say that I find their behavior pretty confusing. There are a lot of fallback methods like this one which are very expensive, and when developing a custom store, it's hard to know which methods you should have to implement in order to get good performance. I think a broader refactor of this part of the code is needed.

In the meantime, this seems like a much needed improvement.

rabernat:
@cgohlke - if you can just update the release notes, we should be gtg here.

cgohlke:
> if you can just update the release notes, we should be gtg here.

Will do later today.

PR #1462, opened 2023-07-14
Refactor core tests, round 2

d-v-b:
It's the same thing as https://github.com/zarr-developers/zarr-python/pull/1305, but without ruff + black linting applied to the whole codebase, and thus a much cleaner diff.  

Some type hints were improved outside the core tests to ensure that mypy was happy.

TODO:
* [] Add unit tests and/or doctests in docstrings
* [] New/modified features documented in docs/tutorial.rst

### DIFF
--- a/zarr/tests/test_core.py
+++ b/zarr/tests/test_core.py
@@ -2060,6 +2060,7 @@ class TestArrayWithDBMStoreBerkeleyDB(TestArray):
 class TestArrayWithLMDBStore(TestArray):
 
     def create_store(self):
+        pytest.importorskip("lmdb")
         path = mktemp(suffix=".lmdb")
         atexit.register(atexit_rmtree, path)
         store = LMDBStore(path, buffers=True)
--- a/zarr/tests/test_core.py
+++ b/zarr/tests/test_core.py
@@ -2415,7 +2415,7 @@ class TestArrayWithFSStore(TestArray):
                         auto_mkdir=True,
                         check=True,
                         create=True,
-                        missing_exceptions=True)
+                        missing_exceptions=None)
         return store
 
     def expected(self):
--- a/zarr/tests/test_core.py
+++ b/zarr/tests/test_core.py
@@ -85,7 +85,10 @@ class TestArray():
     def create_chunk_store(self) -> Optional[BaseStore]:
         return None
 
-    def create_storage_transformers(self, shape) -> Tuple[Any, ...]:
+    def create_storage_transformers(self, shape: Union[int, Tuple[int, ...]]) -> Tuple[Any, ...]:
+        return ()
+
+    def create_filters(self, dtype: Optional[str]) -> Tuple[Any, ...]:
         return ()
 
     def create_array(self, shape: Union[int, Tuple[int, ...]], **kwargs):
@@ -94,9 +97,10 @@ class TestArray():
         # keyword arguments for array initialization
         init_array_kwargs = {
             "path": kwargs.pop("path", self.path),
-            "compressor": self.compressor,
+            "compressor": kwargs.pop("compressor", self.compressor),
             "chunk_store": chunk_store,
             "storage_transformers": self.create_storage_transformers(shape),
+            "filters": kwargs.pop("filters", self.create_filters(kwargs.get("dtype", None)))
         }
 
         # keyword arguments for array instantiation
@@ -109,6 +113,7 @@ class TestArray():
             "partial_decompress": kwargs.pop("partial_decompress", self.partial_decompress),
             "write_empty_chunks": kwargs.pop("write_empty_chunks", self.write_empty_chunks),
         }
+
         init_array(store, shape, **{**init_array_kwargs, **kwargs})
 
         return Array(store, **access_array_kwargs)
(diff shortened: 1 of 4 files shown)


d-v-b:
@rabernat this is ready for eyes

zarr/tests/test_core.py
```diff
@@ -62,16 +64,59 @@
...
 
 
-class TestArray(unittest.TestCase):
-
+class TestArray():
     version = 2
     root = ''
-    KVStoreClass = KVStore
+    path = ''
+    compressor = Zlib(level=1)
```
rabernat, line 71: Curious about why we are setting a default compressor here, and / or why it is different from the built-in default.

zarr/tests/test_core.py
```diff
@@ -62,16 +64,59 @@
...
+    def create_storage_transformers(self, shape) -> Tuple[Any, ...]:
+        return ()
+
+    def create_array(self, shape: Union[int, Tuple[int, ...]], **kwargs):
+        store = self.create_store()
+        chunk_store = self.create_chunk_store()
+        # keyword arguments for array initialization
+        init_array_kwargs = {
+            "path": kwargs.pop("path", self.path),
+            "compressor": self.compressor,
```
rabernat, line 97: Why not 
```suggestion
            "compressor": kwargs.pop("compressor", self.compressor),
```
?

Just trying to understand how this class is intended to be used by developers...

zarr/tests/test_core.py
```diff
@@ -80,15 +125,11 @@ def test_array_init(self):
...
-        if self.version == 2:
-            assert "8fecb7a17ea1493d9c1430d04437b4f5b0b34985" == a.hexdigest()
-        else:
-            assert "968dccbbfc0139f703ead2fd1d503ad6e44db307" == a.hexdigest()
+        assert a.store == normalize_store_arg(store)
         store.close()
 
         # initialize at path
-        store = self.KVStoreClass(dict())
+        store = self.create_store()
```
rabernat, line 132: Kind of feels like we should be in a separate test here. Is anything reused below from above?

zarr/tests/test_core.py
```diff
@@ -2051,75 +2025,46 @@ def expected(self):
...
-        cache_metadata = kwargs.pop('cache_metadata', True)
-        cache_attrs = kwargs.pop('cache_attrs', True)
-        write_empty_chunks = kwargs.pop('write_empty_chunks', True)
-        kwargs.setdefault('compressor', Zlib(1))
-        init_array(store, **kwargs)
-        return Array(store, read_only=read_only, cache_attrs=cache_attrs,
-                     cache_metadata=cache_metadata, write_empty_chunks=write_empty_chunks)
+    def create_store(self):
+        path = mktemp(suffix=".anydbm")
+        atexit.register(atexit_rmglob, path + "*")
```
rabernat, line 2039: It would really be best to replace all of this stuff with [pytest tempdir fixtures](https://docs.pytest.org/en/6.2.x/tmpdir.html).

zarr/tests/test_core.py
```diff
@@ -2215,16 +2127,7 @@ def expected(self):
...
-        store = KVStore(dict())
-        compressor = Blosc(cname='zstd', clevel=1, shuffle=1)
-        kwargs.setdefault('compressor', compressor)
-        cache_metadata = kwargs.pop('cache_metadata', True)
-        cache_attrs = kwargs.pop('cache_attrs', True)
-        write_empty_chunks = kwargs.pop('write_empty_chunks', True)
-        init_array(store, **kwargs)
-        return Array(store, read_only=read_only, cache_metadata=cache_metadata,
-                     cache_attrs=cache_attrs, write_empty_chunks=write_empty_chunks)
+    compressor = Blosc(cname="zstd", clevel=1, shuffle=1)
```
rabernat, line 2130: Feels like we should be using [parametrized fixtures](https://docs.pytest.org/en/7.3.x/how-to/parametrize.html) here rather than making a class for each compressor.

zarr/tests/test_core.py
```diff
@@ -2268,23 +2162,35 @@ def expected(self):
 
 class TestArrayWithFilters(TestArray):
 
-    @staticmethod
-    def create_array(read_only=False, **kwargs):
-        store = KVStore(dict())
-        dtype = kwargs.get('dtype')
+    compressor = Zlib(1)
+
+    def create_array(self, shape: Union[int, Tuple[int, ...]], **kwargs):
```
rabernat, line 2167: Rather than re-implementing this whole method in the subclass, could you add `filters` to the base `TestArray` class attributes?

rabernat (approved):
This is fantastic work @d-v-b! 🙌 

Overall, what I love is that you have eliminated a ton of boilerplate code. I have a few comments mostly in the vein of "could we go even farther"? 

Another open question I have is whether to bring in pytest conveniences like parametrization and temporary directories. 

Happy to leave both thoseitems for future work if you prefer. Or we could keep pushing forward here.

zarr/tests/test_core.py
d-v-b: yeah, I had the same feeling. I wanted to keep the initial PR close to the original tests, but I'm experimenting with a fixture-style locally.

zarr/tests/test_core.py
d-v-b: I did that to make the test pass :) My observation is that historically, different test classes used different compressors (I have no idea why), and because of the hexdigest tests, these compressors get immured with the rest of the tests in that test class. I think this alone argues for rethinking the hexdigest tests.

zarr/tests/test_core.py
d-v-b: no, nothing is reused, so it's a great opportunity for just making a separate test

zarr/tests/test_core.py
d-v-b: done in [9d5b41f](https://github.com/zarr-developers/zarr-python/pull/1462/commits/9d5b41ffc6cfe0739becdaa373634779837d2938)

d-v-b:
I'm making progress even more refactoring of the tests, but it's getting a bit more involved -- I'm entering territory where I'm merging  / splitting / parametrizing individual tests, and I'm less sure that coverage will be preserved. This messier refactoring could take a while, so I think I recommend we settle for the current state of this PR for now. I can follow up with deeper refactors later on. How does this sound @rabernat ?

Issue #1121, opened 2022-09-04
`FSStore` not handling `.zmetadata` correctly when `dimension_separator` is set to `/` at store level.


tasansal:
### Zarr version

v2.12.0

### Numcodecs version

0.10.2

### Python Version

3.10

### Operating System

Linux

### Installation

using pip into virtual environment

### Description

When creating an `FSStore` with dimension separator `/` the `.zmetadata` file also gets renamed to `/zmetadata`. 
I have tested this on Google Cloud Storage.

This causes inconsistent behavior in local vs. cloud stores. The local files get structured like this:

```
/root
|_ my_array
|_ .zgroup
|_ zmetadata
```
Not the missing `.` from `.zmetadata`.

And on the cloud storage, this is what happens:
```
/root
|_ my_array
|_ .zgroup
|_ /
  |_ zmetadata
```
Note that extra prefix `/` before `.zmetadata` and the missing `.`.

It appears that the local FSStore (or path handler) removes one of the slashes from `path_to_root//zmetadata`, whereas, on the cloud stores, it doesn't. 

When we use `zarr.open_consolidated`, in some edge cases this fails. Such as: creating a Zarr on-prem and copying it to Google Cloud doesn't work because file structures are different.

The ideal behavior would be 
```
/root
|_ my_array
|_ .zgroup
|_ .zmetadata
```

as usual, `my_array` has a dimension separator set to `/`, and Zarr should parse that properly.

One possible workaround is NOT to use the `dimension_separator` at the store level but use it when creating arrays.
However, this is prone to error since we would have to specify it every time we create an array, or else there could be inconsistent `.` or `/` arrays within the store.

This works as expected, on both cloud and local:
```python
import zarr


store = zarr.storage.FSStore("test.zarr", mode="w")
root = zarr.open_group(store)
ds = root.create_dataset("my_dataset", shape=(5, 5), chunks=(1, 1), dtype='float32', overwrite=True, dimension_separator='/')
zarr.consolidate_metadata(store)
```

### Steps to reproduce

```python
import zarr


store = zarr.storage.FSStore("test.zarr", mode="w", dimension_separator="/")
root = zarr.open_group(store)
root.create_dataset("my_dataset", shape=(5, 5), chunks=(1, 1), dtype='float32', overwrite=True)
zarr.consolidate_metadata(store)
```

LucaMarconato:
[...]

PR #1467, opened 2023-07-17
Hard-code .zmetadata in FSStore._normalize_key

joshmoore:
The normalization code in FSStore uses a list of well-known files to prevent incorrectly re-writing keys when dimension separator is "/" rather than ".". This was initialized only with the names specified in the spec, which left out ".zmetadata" from consolidated metadata.

see: #1121

TODO:


joshmoore:
It would be good to have a failing test in the suite before this gets merged.

zarr/storage.py
```diff
@@ -1378,7 +1378,12 @@ def _normalize_key(self, key):
         if key:
             *bits, end = key.split('/')
 
-            if end not in (self._array_meta_key, self._group_meta_key, self._attrs_key):
+            if end not in (
+                self._array_meta_key,
+                self._group_meta_key,
+                self._attrs_key,
+                ".zmetadata",  # see: #1121
```
d-v-b, line 1385: could we do this via a class attribute like `_consolidated_meta_key`?

joshmoore:
Any thoughts about getting this into the 2.17.x line?

jhamman:

PR #1481, opened 2023-07-31
Improve LRU Store Cache Performance by Partial Key Cache Update

Yannicked:
When data is added/updated in an LRU store, the `keys_cache` and `contains_cache` are invalidated. In my case this causes a large slowdown due to occasional writes and many reads because my underlying store has a slow `keys()` function.

To address this problem, I propose introducing a `_update_key` function that will be called from the `__setitem__` function. This new function will specifically add the key to the key/contains cache, without wiping the entire cache. By doing so, we can avoid unnecessary cache invalidations and reduce the slowdown associated with occasional writes.

Furthermore, I believe a similar enhancement can be made to the `__delitem__` function. This would allow us to remove a key from the key/contains cache, without invalidating it.

TODO:

### DIFF
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -2448,7 +2448,7 @@ class LRUStoreCache(Store):
 
     def _keys(self):
         if self._keys_cache is None:
-            self._keys_cache = list(self._store.keys())
+            self._keys_cache = set(self._store.keys())
         return self._keys_cache
 
     def listdir(self, path: Path = None):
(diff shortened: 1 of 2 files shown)


joshmoore:
Workflows launched if anyone wants to jump in for review.

zarr/storage.py
```diff
@@ -2504,6 +2504,18 @@ def invalidate_keys(self):
         with self._mutex:
             self._invalidate_keys()
 
+    def _update_key(self, key):
+        if self._keys_cache is None:
+            self._keys_cache = set(self._store.keys())
+        else:
+            self._keys_cache.add(key)
```
madsbk, line 2511:
```suggestion
        if self._keys_cache is not None:
            self._keys_cache.add(key)
```

madsbk (approved):
Looks good to me, I only have a minor simplification.

jhamman:

PR #1495, opened 2023-08-10
[Draft] refactor synchronization API

d-v-b:
This PR alters the internal synchronization API, with the goal of making things simpler and more consistent.

## The current sync API

`Attrs`, `Group`, and `Array` all have methods that require locking / synchronization, e.g. operations that change metadata. The current API for these methods is approximately as follows: if `foo` is some operation that should be synchronized / locked, then for every `foo`, there's a `_foo_nosync()` method that actually does the business of performing `foo`, and a `foo()` method that passes the `foo_nosync()` method, and all its arguments, to `_write_op(func, *args, **kwargs)`, which acquires a lock before calling `_foo_nosync()`. 

`Attrs`, `Group`, and `Array` all use this pattern, but without any formal protocol to enforce consistency.

## What could be improved

 - The current synchronization API has a lot of indirection -- in `Array`, a given method like [`append`](https://github.com/zarr-developers/zarr-python/blob/6ed4d785db71d7087b276f0f6dc5b3f7cfedd7f1/zarr/core.py#L2720) passes [another method](https://github.com/zarr-developers/zarr-python/blob/6ed4d785db71d7087b276f0f6dc5b3f7cfedd7f1/zarr/core.py#L2757) to [another method](https://github.com/zarr-developers/zarr-python/blob/6ed4d785db71d7087b276f0f6dc5b3f7cfedd7f1/zarr/core.py#L2638) which passes it to yet [another method](https://github.com/zarr-developers/zarr-python/blob/6ed4d785db71d7087b276f0f6dc5b3f7cfedd7f1/zarr/core.py#L2621), wherein the `append` routine is finally called, and the return value generated. All this indirection makes it hard to reason about the flow of data through the program, either informally or formally (via types).
 - `Attrs`, `Group`, and `Array` all implement this API, but with subtle differences. Ideally, the synchronization behavior would be defined abstractly somewhere and these classes would implement that behavior as needed.

## What's in this PR

I defined a protocol [`Synchronized`](https://github.com/d-v-b/zarr-python/blob/6f36205fc4612a6c70a536a139ee7ebba9118d3f/zarr/sync.py#L11) which has a single method, `_write_context`, which takes a string key and returns a context manager (e.g., a lock on that key). `Attrs`, `Array`, and `Group` all inherit from `Synchronized`, and so they must define `_write_context` accordingly. Methods of these classes that require synchronization now use `_write_context(key)` to lock the key that must be synchronized on, e.g., 

```python
...
# inside some method of Array
with self._write_context(array_metadata_key): 
    # do stuff
...
```
The redundant `_write_op` and `_synchronized_op` methods have been removed.

I also made a [synchronizer class](https://github.com/d-v-b/zarr-python/blob/6f36205fc4612a6c70a536a139ee7ebba9118d3f/zarr/sync.py#L34) that does nothing but return instances of [`NoLock`](https://github.com/d-v-b/zarr-python/blob/6f36205fc4612a6c70a536a139ee7ebba9118d3f/zarr/sync.py#L24). Thus all instances of `Array`, `Attrs`, and `Group` can have a synchronizer, even if no actual synchronization is expected. As a result, the `._synchronizer` property of these objects is no longer nullable, and this removes a few `if self._synchronizer is None...` checks.

## Other changes

- added `_attrs_key` and `_meta_key` properties to `Array`, which are set in `__init__`. This replaces many redundant invocations of `_prefix_to_attrs_key` and `_prefix_to_meta_key`.
- added a few type annotations here and there.

## Going farther

- There is still a layer of indirection over the methods that have to be synchronized. I'm open for suggestions on how to reduce this.

- `synchronizer` is a public attribute of `Attrs`, but a private attribute of `Array` and `Group` (`._synchronizer`). We should make the APIs consistent.

- If we ever want to support performant distributed locking, then we can't have workers requesting a separate lock for every single chunk they want to reserve. But it might work if workers can request locks for _sets_ of chunks. This would require changes to the basic synchronization API, but these changes would be backwards compatible. I think this is a good direction to explore.

TODO:


jhamman:
Closing as stale. Let's continue this work on the v3 branch.

PR #1524, opened 2023-09-14
Faster delitems

### LINKED ISSUE #1336: performance optimization in `delitems`
in [`FSStore.delitems`](https://github.com/zarr-developers/zarr-python/blob/6f11ae78b142e242d2ac8bc67019e7528539fe73/zarr/storage.py#L1432), only keys that exist in the store are deleted. This requires [checking each key in a loop](https://github.com/zarr-developers/zarr-python/blob/6f11ae78b142e242d2ac8bc67019e7528539fe73/zarr/storage.py#L1436), which incurs a lot of overhead for filesystems like s3. I suspect that sequentially calling `FSStore.__contains__` here erodes a lot of the benefit of using a bulk delete.

From the zarr side, the solution would be to simply request deletion of keys regardless of whether they exist in the store or not, however fsspec raises an exception if you try to `rm` a file that doesn't exist. So perhaps we need to add an exception-handling kwarg to `rm` like the `on-omit` kwarg for [`FSMap.getitems`](https://github.com/fsspec/filesystem_spec/blob/012816bb142de1e507423b81dc3b2c925d4424c3/fsspec/mapping.py#L70). This would be more "s3ish" anyway, since the boto3 [`delete_objects`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Bucket.delete_objects) method silently treats missing keys as deleted.

thoughts on this @martindurant ? I'm happy to open a PR on the FSSpec side.

d-v-b:
fixes #1336

We first try to delete keys without checking if they exist in storage. If the storage backend treats deleting missing keys as error, and it signals such an error by raising `FileNotFoundError`, then keys will be filtered based on whether they exist in storage. This is much slower.

TODO:


martindurant:
This is fine, but rm() in fsspec should really allow on_error on all calls consistently. It's worth noting that the backend that doesn't ignore FileNotFound (i.e., local) is also the one that deletes files sequentially as opposed to async.

martindurant:
> is also the one that deletes files sequentially

I mean, we could try/except on individual files in this case and still not bother explicitly checking existence first.

jhamman:

PR #1536, opened 2023-10-01
Add type hints to zarr.create

dstansby:
This PR adds type hints to `zarr.creation.create()`.

I couldn't find any discussion of adding type hints to `zarr` in the issues or discussion forums, so not sure if this is welcome - I would definitely personally find it helpful to have type hints throughout the codebase, and would be happy to chip away at adding them if they are. But definitely understand if it would be useful to have a wider discussion separate to this PR about adding type hinting to `zarr` before going any further.

TODO:


zarr/creation.py
```diff
@@ -19,32 +22,35 @@
...
-    read_only=False,
-    object_codec=None,
-    dimension_separator=None,
-    write_empty_chunks=True,
+    order: MEMORY_ORDER = "C",
+    store: Optional[Union[str, MutableMapping]] = None,
+    synchronizer: Optional[Synchronizer] = None,
+    overwrite: bool = False,
+    path: Optional[str] = None,
+    chunk_store: Optional[MutableMapping] = None,
```
d-v-b, line 42: maybe there's an opportunity to specialize the `MutableMapping` annotations with the type parameters for the keys and values? e.g., `MutableMapping[str, bytes]`

zarr/creation.py
```diff
@@ -19,32 +22,35 @@
...
+    read_only: bool = False,
+    object_codec: Optional[Codec] = None,
+    dimension_separator: Optional[DIMENSION_SEPARATOR] = None,
+    write_empty_chunks: bool = True,
     *,
-    zarr_version=None,
-    meta_array=None,
-    storage_transformers=(),
+    zarr_version: Optional[ZARR_VERSION] = None,
+    meta_array: Optional[npt.ArrayLike] = None,
```
d-v-b, line 52: I'm not sure that `meta_array` is an instance of `npt.ArrayLike` -- if I understand numpy's type hints correctly (and I may not!), `npt.ArrayLike` is basically `Union[Numpy array, thing that can be coerced into a numpy array]`, so I think `List[int]` counts as `ArrayLike`, but I'm not sure that `List[int]` would work for `meta_array`. cc @madsbk , who knows more about this kwarg

zarr/sync.py
```diff
@@ -5,7 +5,13 @@
 import fasteners
 
 
-class ThreadSynchronizer:
+class Synchronizer:
```
d-v-b, line 8: over in https://github.com/zarr-developers/zarr-python/pull/1495/files#diff-9b925c58923bb1c258699fe5c0f20a9a8cff470568d6badb9498156dd19a44e9 I use protocols here, in case that's of interest. I'm not sure when / if my PR will get merged (I haven't been pushing it at all), and I'm happy to rebase on top of your PR if yours gets merged first.

zarr/creation.py
```diff
@@ -19,32 +22,35 @@
...
 from zarr.util import normalize_dimension_separator
 
 
 def create(
-    shape,
-    chunks=True,
-    dtype=None,
+    shape: Union[int, Tuple[int, ...]],
+    chunks: bool = True,
+    dtype: Optional[Union[str, npt.DTypeLike]] = None,
```
d-v-b, line 34: I think `str` is a subtype of `dtypeLike`: https://numpy.org/doc/stable/reference/typing.html#numpy.typing.DTypeLike

d-v-b: (I should say, "the subset of `str` that can be coerced into numpy dtypes" is a subtype of `DtypeLike`)

zarr/creation.py
```diff
@@ -19,32 +22,35 @@
...
-    cache_attrs=True,
-    read_only=False,
-    object_codec=None,
-    dimension_separator=None,
-    write_empty_chunks=True,
+    order: MEMORY_ORDER = "C",
+    store: Optional[Union[str, MutableMapping]] = None,
+    synchronizer: Optional[Synchronizer] = None,
+    overwrite: bool = False,
+    path: Optional[str] = None,
```
d-v-b, line 41: maybe check if we accept instances of `Pathlib.Path` here

d-v-b:
I like this! I left some comments. Having started one of these "add types to zarr" endeavors before, I understand that it can feel unbounded, so I respect your restraint in only adding types where you did :)

zarr/creation.py
madsbk: Agree, `meta_array` relies on [NEP18](https://numpy.org/neps/nep-0018-array-function-protocol.html). 
Maybe a more precise hint would be something like:
```python

from numcodecs.compat import NDArrayLike
class MetaArray(NDArrayLike):
    def __array_function__(self, func, types, args, kwargs) -> Any:
        ...   
```

zarr/creation.py
dstansby: I might leave that for another PR to keep this one simple?

zarr/creation.py
dstansby: 👍  - I went for adding a protocol that specifies `__array_funciton__` as a method.

zarr/sync.py
dstansby: Ah, thanks for reminding me of Protocols, I added a simple one for `Synchronizer` for now, but I imagine that will be made redundant by your PR anyway.

zarr/creation.py
dstansby: 👍

zarr/creation.py
dstansby: I checked and fixed this to be the same type that is used throughout `zarr.storage` (that doesnt' currently explicitly allow `pathlib.Path`).

zarr/creation.py
d-v-b: that's fine!

d-v-b (approved):
👍

d-v-b:
any objections to merging this?

d-v-b:
I'm kicking off the gha workflows, assuming those pass I'm going to merge this unless there are any objections.

zarr/util.py
```diff
@@ -286,9 +287,9 @@ def normalize_order(order: str) -> str:
     return order
 
 
-def normalize_dimension_separator(sep: Optional[str]) -> Optional[str]:
+def normalize_dimension_separator(sep: Optional[str]) -> Optional[DIMENSION_SEPARATOR]:
     if sep in (".", "/", None):
-        return sep
+        return cast(DIMENSION_SEPARATOR, sep)
```
d-v-b, line 292: If `sep` is `None`, then I think this cast is incorrect?

dstansby: 👍 I fixed the casting

joshmoore:
I know that @d-v-b and I tend to be fans, but this feels like something that others may want to take a look. Anyone want us to hold merging, @zarr-developers/python-core-devs?

d-v-b:
Since this doesnt affect runtime at all, I think its safe to merge this.

joshmoore:
It is adding a new public module though. I would give it until next week following https://github.com/zarr-developers/governance/blob/main/GOVERNANCE.md#decision-making-process

zarr/sync.py
```diff
@@ -1,11 +1,19 @@
...
 
 import fasteners
 
 
-class ThreadSynchronizer:
+class Synchronizer(Protocol):
+    """Base class for synchronizers."""
+
+    def __getitem__(self, item):
+        ...
```
joshmoore, line 13: Thoughts welcome on how to make codecov happy with these lines.

d-v-b: I couldn't find anything obvious in https://docs.codecov.com/docs/ignoring-paths, seems like it only has per-file granularity. Since we know the decrease in coverage isn't meaningful, I'd be fine ignoring it here. In the future maybe we put protocols in `types.py` (either a module-local one, or a zarr-wide one)  and tell codecov to ignore it.

dstansby: I've updated the code coverage config to ignore lines that are "...".

d-v-b: good idea!

zarr/creation.py
```diff
@@ -19,32 +22,35 @@
...
+from zarr.types import ZARR_VERSION, DIMENSION_SEPARATOR, MEMORY_ORDER, MetaArray, PathLike
 from zarr.util import normalize_dimension_separator
 
 
 def create(
-    shape,
-    chunks=True,
-    dtype=None,
+    shape: Union[int, Tuple[int, ...]],
+    chunks: bool = True,
```
d-v-b, line 33: chunks can also be `Tuple[int, ...]`, so i think the full type should be `Union[Tuple[int, ...], bool]`

jhamman:
@d-v-b - if you are happy here, would you mind giving this a merge?

Thanks @dstansby for the contribution here. FYI - we're planning much deeper type coverage for v3, if you are up to help in that effort, your contributions would be more than welcome (#1593, etc.).

d-v-b:
thanks @dstansby! and please feel free to dive in to the v3 effort!

PR #1556, opened 2023-10-31
Single startswith() call instead of multiple ones

DimitriPapadopoulos:
It's faster and probably more readable.

[Description of PR]

TODO:

### DIFF
--- a/zarr/_storage/store.py
+++ b/zarr/_storage/store.py
@@ -221,9 +221,8 @@ class StoreV3(BaseStore):
             )
 
         if (
-            not key.startswith("data/")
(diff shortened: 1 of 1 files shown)


jakirkham:
Wonder if this is part of performance degradation with v3 that was called out before

cc @rabernat (for awareness)

DimitriPapadopoulos:
Replacing two function calls with a single function call should be faster, **unless** `key.startswith("data/")` is usually false. In that case, the second function call `key.startswith("meta/")` is actually seldom triggered. But then it would be nice to add a comment here.

Issue #1552, opened 2023-10-27
Drop `sphinx-rtd-theme` dependency


jakirkham:
IIUC the `sphinx-rtd-theme` dependency is no longer used. Should we drop this?

https://github.com/zarr-developers/zarr-python/blob/1ed37f5fa7039cbc8849bf9b0e9aae9c890e7ced/pyproject.toml#L51

d-v-b:
sounds good to me!

PR #1563, opened 2023-11-02
Remove  dependency

MSanKeys963:
Fixes: #1552.

TODO:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 
 [project]
-name = "Zarr"
+name = "zarr"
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 
 [project]
-name = "zarr"
+name = "Zarr"
```
jakirkham, line 7: Let's keep the name as-is. Would like to avoid confusing package tooling

jakirkham (commented):
Thanks Sanket! 🙏

Had one suggestion below

pyproject.toml
joshmoore, line 7:
```suggestion
name = "zarr"
```

PR #1581, opened 2023-11-28
Cache result of FSStore._fsspec_installed()

ph03:
`FSStore._fsspec_installed()` is run in a lot of hierarchy operations, and has a significant runtime overhead for each invocation. Given the result will not change between invocations, cache the very first invocation for subsequent runs.

Example profiling of effects without (`main`) behaviour (6sec total runtime)

image

and with caching the invocation (2sec total runtime)

image

Most of the runtime is spent in the repeated lookup of `FSStore._fsspec_installed()`
image

TODO:

### DIFF
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -43,6 +43,9 @@ Docs
 Maintenance
 ~~~~~~~~~~~
 
+* Cache result of ``FSStore._fsspec_installed()``.
+  By :user:`Janick Martinez Esturo <ph03>` :issue:`1581`.
+
 * Extend copyright notice to 2023.
   By :user:`Jack Kelly <JackKelly>` :issue:`1528`.
 
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -28,6 +28,7 @@ import warnings
 import zipfile
 from collections import OrderedDict
 from collections.abc import MutableMapping
+from functools import lru_cache
 from os import scandir
 from pickle import PicklingError
 from threading import Lock, RLock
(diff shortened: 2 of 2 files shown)


d-v-b (approved):

d-v-b:
i was just noticing sluggish fsstore performance today. this change looks good to me!

jhamman:
This looks good to me. Would be nice to get @martindurant to take a look here.

martindurant:
Agree. I suppose checking `"fsspec" in sys.modules` would be much faster, after the first use, but this is fine.

Not sure why we didn't do something more typical like
```
        if not self._fsspec_installed():  # pragma: no cover
            raise ImportError("`fsspec` is required to use zarr's FSStore")
```
->
```
        try:
            import fsspec
        except ImportError:
```

ph03:
Should I still adapt this?

martindurant:
I am happy to go with @d-v-b and @jhamman on this one - this fix is fine, and we can always simplify another time.

ph03:
> I am happy to go with @d-v-b and @jhamman on this one - this fix is fine, and we can always simplify another time.

Awesome - perhaps you can let me know if there are additional points missing for merging?

jhamman:
I think the main missing thing here is a changelog entry. Then we should be good to merge.

ph03:
> I think the main missing thing here is a changelog entry. Then we should be good to merge.

Cool thanks, I wasn't aware, thanks for pointing out - added a changelog entry now :)

ph03:
@martindurant are you ok with the PR / can you let me know which remaining parts are missing (would like to start pointing dependent code to a main repo commit / not a personal one ^^)?

PR #1588, opened 2023-12-05
Extensible codecs for V3

normanrz:
Makes codecs extensible in the v3 branch by adding a codec registry.


zarr/tests/test_codecs_v3.py
```diff
@@ -0,0 +1,989 @@
...
+
+
+@pytest.fixture
+def store() -> Iterator[Store]:
+    yield MemoryStore()
+
+
+@pytest.fixture
+def sample_data() -> Iterator[np.ndarray]:
+    yield np.arange(0, 128 * 128 * 128, dtype="uint16").reshape((128, 128, 128), order="F")
```
jhamman, line 46:
```suggestion
def sample_data() -> np.ndarray:
    return np.arange(0, 128 * 128 * 128, dtype="uint16").reshape((128, 128, 128), order="F")
```
nit: I don't think there there is any value in returning this as an iterator.

zarr/v3/codecs/registry.py
```diff
@@ -0,0 +1,56 @@
...
+    item = __codec_registry.get(key)
+    if item is None:
+        if key in __lazy_load_codecs:
+            # logger.debug("Auto loading codec '%s' from entrypoint", codec_id)
+            cls = __lazy_load_codecs[key].load()
+            register_codec(key, cls)
+            item = __codec_registry.get(key)
+    if item:
+        return item
+    raise KeyError
```
jhamman, line 45:
```suggestion
    raise KeyError(key)
```

zarr/v3/codecs/sharding.py
```diff
@@ -260,7 +318,7 @@ async def decode_partial(
         store_path: StorePath,
         selection: SliceSelection,
     ) -> Optional[np.ndarray]:
-        # print("decode_partial")
+        print("decode_partial")
```
jhamman, line 321:
```suggestion
```

zarr/v3/common.py
```diff
@@ -59,24 +53,8 @@ def _structure_chunk_key_encoding_metadata(d: Dict[str, Any], _t) -> ChunkKeyEnc
     )
 
     def _structure_codec_metadata(d: Dict[str, Any], _t=None) -> CodecMetadata:
```
jhamman, line 55: not for now but it would be nice if the codec metadata was a `TypedDict` so we could expect for keys like "name".

jhamman (approved):
I mostly reviewed the entrypoints stuff since the rest seemed to be reorganization / porting from zarrita. Looks really nice!

zarr/v3/common.py
normanrz: `name` is really the only shared attribute among all codecs. Even `configuration` is optional. I don't think there'd be much value in a TypedDict. This piece of code is in the middle of the deserialization; lots of untyped magic happening here anyways :-)

zarr/tests/test_codecs_v3.py
```diff
@@ -0,0 +1,989 @@
...
+@frozen
+class _AsyncArraySelectionProxy:
+    array: AsyncArray
+    selection: Selection
+
+    async def get(self) -> np.ndarray:
+        return await self.array.getitem(self.selection)
+
+    async def set(self, value: np.ndarray):
+        return await self.array.setitem(self.selection, value)
```
normanrz, line 36: @jhamman @d-v-b What do you think about this making it into the AsyncArray? It allows for writing something like this
```
await _AsyncArrayProxy(array)[:10, 10:].get()
```
instead of 
```
await array.getitem((slice(None, 10), slice(10, None)))
```

PR #1590, opened 2023-12-05
basic support for v2 and v3 groups

jhamman:
This begins the work of harmonizing the V2 and V3 group APIs from Zarrita with the Zarr-Python Group API. 

I'll leave some comments inline.

TODO:


zarr/v3/group.py
```diff
@@ -1,97 +1,106 @@
...
-from zarr.v3.sync import sync
+from zarr.v3.sync import SyncMixin
 
 
 @frozen
 class GroupMetadata:
     attributes: Dict[str, Any] = field(factory=dict)
-    zarr_format: Literal[3] = 3
-    node_type: Literal["group"] = "group"
+    zarr_format: Literal[2, 3] = 3  # field(default=3, validator=validators.in_([2, 3]))
```
jhamman, line 20: @normanrz - I tried this and failed! Happy to return to it later.

zarr/v3/abc/group.py
```diff
@@ -17,70 +19,188 @@ def info(self) -> Any:  # TODO: type this later
...
+        ...
+
+    @abstractmethod
+    def __getitem__(self, child: str) -> Union[SyncArray, "SyncGroup"]:
+        """get child"""
+        ...
+
+    @abstractmethod
+    def __setitem__(self, key: str, value: Union[SyncArray, "SyncGroup"]) -> None:
+        """get child"""
```
jhamman, line 137:
```suggestion
```
Proposing that we do away with `Group.__setitem__` all together. It really doesn't make sense to include in the api.

zarr/v3/abc/group.py
```diff
@@ -17,70 +19,188 @@ def info(self) -> Any:  # TODO: type this later
...
+        ...
+
+    @abstractmethod
+    def group_keys(self) -> AsyncIterator[str]:
+        """iterate over child group keys"""
+        ...
```
jhamman, line 143: all of these iterators need listable stores before they can be built.

zarr/v3/config.py
```diff
@@ -0,0 +1,19 @@
...
+
+from asyncio import AbstractEventLoop
+from typing import Literal, Optional
+from attr import frozen
+
+
+@frozen
+class SyncConfiguration:
+    concurrency: Optional[int] = None
+    asyncio_loop: Optional[AbstractEventLoop] = None
```
jhamman, line 11: this was a half hearted attempt at rethinking the config topic. I expect this to go away soon.

zarr/v3/group.py
```diff
@@ -107,73 +116,276 @@ def from_json(
...
+        raise NotImplementedError
+
+    def __iter__(self):
+        raise NotImplementedError
+
+    def __len__(self):
+        raise NotImplementedError
+
+    def __setitem__(self, key, value):
+        raise NotImplementedError
```
jhamman, line 318:
```suggestion
```

zarr/v3/group.py
```diff
@@ -107,73 +116,276 @@ def from_json(
...
+    def update_attributes(self, new_attributes: Dict[str, Any]):
+        self._sync(self._async_group.update_attributes, new_attributes)
+        return self
+
+    @property
+    def nchildren(self) -> int:
+        return self._sync(self._async_group.nchildren)
+
+    @property
+    def children(self) -> List[Array, "Group"]:
```
jhamman, line 337: some exploratory work is needed to figure out how to make this (and similar methods) a proper iterator/generator instead of a list. The tricky bit is that the async method is returning an AsyncIterator and we need to synchronize that into a regular Iterator.

jhamman (commented):
A few thoughts below 👇

zarr/v3/group.py
```diff
@@ -1,97 +1,106 @@
...
-        runtime_configuration: RuntimeConfiguration = RuntimeConfiguration(),
-    ) -> Group:
-        return sync(
-            cls.open_async(store, runtime_configuration),
-            runtime_configuration.asyncio_loop,
-        )
+        # TODO: consider trying to autodiscover the zarr-format here
+        if zarr_format == 3:
+            # V3 groups are comprised of a zarr.json object
+            # (it is optional in the case of implicit groups)
```
normanrz, line 82: Are implicit groups really a thing? I don't think they show up in the v3 spec.

zarr/v3/group.py
```diff
@@ -107,73 +116,276 @@ def from_json(
...
-        if zarr_json["node_type"] == "array":
-            return Array.from_json(
-                store_path, zarr_json, runtime_configuration=runtime_configuration
+
+        store_path = self.store_path / key
+
+        if self.zarr_format == 3:
+            zarr_json_bytes = await (store_path / ZARR_JSON).get_async()
+            if zarr_json_bytes is None:
+                # implicit group?
```
normanrz, line 129: This is a bit dangerous because we have no good way of telling whether this path even exists.

zarr/v3/group.py
```diff
@@ -107,73 +116,276 @@ def from_json(
...
+            else:
+                zarr_json = json.loads(zarr_json_bytes)
+            if zarr_json["node_type"] == "group":
+                return type(self).from_json(store_path, zarr_json, self.runtime_configuration)
+            if zarr_json["node_type"] == "array":
+                return Array.from_json(
+                    store_path, zarr_json, runtime_configuration=self.runtime_configuration
+                )
+        elif self.zarr_format == 2:
+            # Q: how do we like optimistically fetching .zgroup, .zarray, and .zattrs?
```
normanrz, line 144: I think the overfetching is fine because it is happening in parallel.

zarr/v3/group.py
```diff
@@ -1,97 +1,106 @@
...
+    async def create(
         cls,
         store: StoreLike,
         *,
         attributes: Optional[Dict[str, Any]] = None,
         exists_ok: bool = False,
+        zarr_format: Literal[2, 3] = 3,  # field(default=3, validator=validators.in_([2, 3])),
         runtime_configuration: RuntimeConfiguration = RuntimeConfiguration(),
-    ) -> Group:
+    ) -> "AsyncGroup":
```
normanrz, line 54: No need for the quotes because we have `from __future__ import annotations`

zarr/v3/group.py
```diff
@@ -107,73 +116,276 @@ def from_json(
...
 
-    @classmethod
-    async def open_or_array(
-        cls,
-        store: StoreLike,
-        runtime_configuration: RuntimeConfiguration = RuntimeConfiguration(),
+    async def getitem(
+        self,
+        key: str,
     ) -> Union[Array, Group]:
```
normanrz, line 122:
```suggestion
    ) -> Union[AsyncArray, AsyncGroup]:
```

normanrz: Probably needs fixing in other places, too

zarr/v3/group.py
```diff
@@ -107,73 +116,276 @@ def from_json(
...
+                    "zarr_format": self.zarr_format,
+                    "node_type": "group",
+                    "attributes": {},
+                }
+            else:
+                zarr_json = json.loads(zarr_json_bytes)
+            if zarr_json["node_type"] == "group":
+                return type(self).from_json(store_path, zarr_json, self.runtime_configuration)
+            if zarr_json["node_type"] == "array":
+                return Array.from_json(
```
normanrz, line 140:
```suggestion
                return AsyncArray.from_json(
```

zarr/v3/abc/group.py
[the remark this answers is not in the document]
jhamman: @d-v-b and I were just talking about that last night. I think the answer is NO. For ABCs, I we probably only need the Store and Codec base classes.

zarr/v3/group.py
jhamman: They do show up in the spec (in quite a few places actually)! https://zarr-specs.readthedocs.io/en/latest/v3/core/v3.0.html#explicit-vs-implicit-groups

zarr/v3/group.py
jhamman: I agree and we should return to this. For now I'm going to put a log statement there and we can address this later.

zarr/v3/group.py
rabernat: I was thinking about this issue the other day. One possible behavior when you "open" anything in Zarr would be to call the `store.tree(depth=1)` method, fetching all existing metadata docs at the root level and one level below.

zarr/v3/group.py
```diff
@@ -1,97 +1,108 @@
...
 
-    def to_bytes(self) -> bytes:
-        return json.dumps(asdict(self)).encode()
+    def to_bytes(self) -> Dict[str, bytes]:
+        if self.zarr_format == 3:
+            return {ZARR_JSON: json.dumps(asdict(self)).encode()}
+        elif self.zarr_format == 2:
+            return {
+                ZGROUP_JSON: self.zarr_format,
+                ZATTRS_JSON: json.dumps(self.attributes).encode(),
```
d-v-b, line 32: how would custom JSON serialization of user attributes plug in here? e.g., i don't think numpy dtype objects serialize to JSON easily. Maybe `to_bytes` can take a `attrs_serializer` kwarg or something.

zarr/v3/group.py
```diff
@@ -107,73 +118,294 @@ def from_json(
...
+    def tree(self, expand=False, level=None) -> Any:
+        return self._sync(self._async_group.tree(expand=expand, level=level))
+
+    def create_group(self, name: str, **kwargs) -> Group:
+        return Group(self._sync(self._async_group.create_group(name, **kwargs)))
+
+    def create_array(self, name: str, **kwargs) -> Array:
+        return Array(self._sync(self._async_group.create_array(name, **kwargs)))
+
+    def empty(self, **kwargs) -> Array:
```
d-v-b, line 386: thoughts on ultimately putting these routines in a dedicated namespace?

zarr/v3/store.py
```diff
@@ -65,6 +65,14 @@ def __str__(self) -> str:
     def __repr__(self) -> str:
         return f"StorePath({self.store.__class__.__name__}, {repr(str(self))})"
 
+    def __eq__(self, other: Any) -> bool:
+        try:
+            if self.store == other.store and self.path == other.path:
+                return True
+        except Exception:
```
d-v-b, line 72: we can probably be a bit more specific about the exception type here

PR #1146, opened 2022-09-27
2.13.2: another fix for the conda-forge tests

joshmoore:
See description in https://github.com/zarr-developers/zarr-python/pull/1143 Testing locally, however, does not suffice. In this case, the relative path in the test worked for me, but doesn't work on conda-forge. The only reliable way to be sure is to run `./build-locally.py` in the zarr-feedstock **against pre-tagged code**. ⚠️ 

TODO:
* [x] Add unit tests and/or doctests in docstrings
* [x] Add docstrings and API docs for any new/modified user-facing classes and functions
* [x] New/modified features documented in docs/tutorial.rst
* [x] Changes documented in docs/release.rst
* [x] GitHub Actions have all passed
* [x] Test coverage is 100% (Codecov passes)

PR #1592, opened 2023-12-07
use src layout and use `hatch` for packaging

d-v-b:
This PR changes the layout of the repo by putting the python source code in the `src` directory at the root of the repo. The logic for this layout can be found in lots of places online, but the [scientific python packaging guide](https://learn.scientific-python.org/development/guides/packaging-simple/#package-structure) expresses it fairly concisely. Basically putting the source code in `src` helps prevent running tests with the local code instead of the installed code. 

This change broke some tests, specifically all the tests that were trying to treat `zarr.tests` as a module, which was simple to fix. Oddly, there's code in `main` that _imports from the test suite_: see [here](https://github.com/zarr-developers/zarr-python/blob/25dbeeda7d3a300569b358c157c5bd1c02ddaec3/zarr/meta.py#L462C3-L462C3) which seems pretty undesirable, if not a bug. I fixed this by reversing the direction of the imports: the relevant class is now defined in `meta.py` and the tests that use this class import it from `zarr`.

Something else I noticed is that the test suite creates a persistent directory called `fixture` that contains a bunch of data that outlives the tests. These tests can be found [here](https://github.com/zarr-developers/zarr-python/blob/25dbeeda7d3a300569b358c157c5bd1c02ddaec3/zarr/tests/test_dim_separator.py#L32) in `main`. The tests put this directory in the same directory as the root of the module, which means after running tests we end up with a `src/fixture` directory that has a bunch of stuff in it. This seems like something we should not be doing. For example, perhaps we could create temporary directories scoped to the test session. Maybe someone who understands these tests better can shed some light on how we ended up here.

In addition to changing the layout of the source code, I modified `pyproject.toml` to support use of the [`hatch`](https://hatch.pypa.io/latest/) project manager. I also considered [PDM](https://pdm-project.org/latest/) and [Poetry](https://python-poetry.org/) (which I have the most experience with). I chose `hatch` because it was used in the scientific python packaging guide, and I've never heard anyone complain about it, so it seems safe. I think a modern project manager could make development a lot easier. We can replace the `requirements_dev_*.txt` files with entries in `pyproject.toml`, define environment variables for workflows in `project.toml`, and manage separate environments more easily than with our current setup. 

This PR is WIP because
- It's a work in progress. I'm not done folding all our build / test / ci configuration into `pyproject.toml`
- I anticipate that lots of people will want to weigh in on how packaging for the repo should be configured. 

I'm very interested to hear what people think!


d-v-b:
Git history has been restored via me using `git mv` instead of vanilla `mv`. 

I opened #1601 to separate the test importing issues from the actual layout change.

Things I would appreciate help with:
- getting CI to use hatch environments / commands
- Setting up hatch environments for our different test contexts
- getting dynamic versioning working
- expressing the doc generation process via hatch

pyproject.toml
```diff
@@ -1,42 +1,38 @@
 [build-system]
-requires = ["setuptools>=64.0.0", "setuptools-scm>1.5.4"]
-build-backend = "setuptools.build_meta"
+requires = ["hatchling"]
```
Saransh-cpp, line 2: There is an equivalent `setuptools-scm` for `hatch` - `hatch-vcs`. I feel that going from automated versioning to manual versioning would be a downgrade and an extra overhead for the maintainers.

pyproject.toml
```diff
@@ -55,85 +51,135 @@ docs = [
...
 addopts = [
-    "--durations=10",
-]
-filterwarnings = [
-    "error:::zarr.*",
-    "ignore:PY_SSIZE_T_CLEAN will be required.*:DeprecationWarning",
-    "ignore:The loop argument is deprecated since Python 3.8.*:DeprecationWarning",
-]
+    "--durations=10"
+]
```
Saransh-cpp, line 185: Good practice to have empty lines at the end of every file
```suggestion
]

```

pyproject.toml
```diff
@@ -55,85 +51,135 @@ docs = [
...
-)/
-'''
+
+[tool.ruff.isort]
+known-first-party = ["zarr"]
+
+[tool.ruff.flake8-tidy-imports]
+ban-relative-imports = "all"
+
+[tool.ruff.per-file-ignores]
```
Saransh-cpp, line 149: ruff can now format too!
```suggestion
[tool.ruff.lint.per-file-ignores]
```

pyproject.toml
```diff
@@ -55,85 +51,135 @@ docs = [
...
-  | build
-  | dist
-  | docs
-)/
-'''
+
+[tool.ruff.isort]
+known-first-party = ["zarr"]
+
+[tool.ruff.flake8-tidy-imports]
```
Saransh-cpp, line 146: Ditto!
```suggestion
[tool.ruff.lint.flake8-tidy-imports]
```

pyproject.toml
```diff
@@ -55,85 +51,135 @@ docs = [
...
-  | \.venv
-  | _build
-  | buck-out
-  | build
-  | dist
-  | docs
-)/
-'''
+
+[tool.ruff.isort]
```
Saransh-cpp, line 143:
```suggestion
[tool.ruff.lint.isort]
```

pyproject.toml
```diff
@@ -55,85 +51,135 @@ docs = [
...
+]
+all = [
+  "style",
+  "typing",
 ]
 
 [tool.black]
+target-version = ["py38"]
+line-length = 100
+skip-string-normalization = true
```
Saransh-cpp, line 137: black can now be replaced with ruff format but maybe not in this PR

src/zarr/__about__.py
```diff
@@ -0,0 +1,4 @@
+# SPDX-FileCopyrightText: 2023-present Zarr Developers
+#
+# SPDX-License-Identifier: MIT
+__version__ = "0.0.1"
```
Saransh-cpp, line 4: See the message about `hatch-vcs` above.

pyproject.toml
```diff
@@ -1,42 +1,38 @@
...
 
 
 [project]
 name = "zarr"
+dynamic = ["version"]
 description = "An implementation of chunked, compressed, N-dimensional arrays for Python"
 readme = { file = "README.md", content-type = "text/markdown" }
 maintainers = [
-    { name = "Alistair Miles", email = "alimanfoo@googlemail.com" }
+    { name = "Alistair Miles", email = "alimanfoo@googlemail.com"}
```
Saransh-cpp, line 12:
```suggestion
    { name = "Alistair Miles", email = "alimanfoo@googlemail.com" }
```

Saransh-cpp (changes requested):
Saw the help required label in my GitHub feed 😄

pyproject.toml
d-v-b: that makes sense! We definitely want to avoid manual versioning. I will look into adding `hatch-vcs`

clbarnes: Looks like this is referred to in the scientific packaging guide too https://learn.scientific-python.org/development/guides/packaging-simple/#versioning

clbarnes:
What's the rationale behind the `__about__.py` files?

clbarnes:
Hatch has just released a new version, 1.8, with some nice features https://hatch.pypa.io/latest/blog/2023/12/11/hatch-v180/

jhamman:
@d-v-b - would love to see this move forward. Is the list of help-needed here still relevant?

d-v-b:
Yes! Help is definitely still wanted here. I'm not sure when I will have time to advance this.

Saransh-cpp:
@d-v-b would it be possible for you to give me access to your fork? I might get some time this weekend to work on the hatch-vcs part. Otherwise, if you have time to work on it yourself, I could answer your questions or help you with the issues you are facing with hatch-vcs.

> getting CI to use hatch environments / commands
Setting up hatch environments for our different test contexts
expressing the doc generation process via hatch

These are add-ons. Everything working in the CI with `setuptools` should work with `hatch`. I would suggest just migrating to `hatch` and `hatch-vcs` in this PR and then switching to `hatch` commands in the CI in the follow-up PRs.

d-v-b:
@Saransh-cpp thank you! I added you as a collaborator to my fork of zarr-python.

d-v-b:
> These are add-ons. Everything working in the CI with setuptools should work with hatch. I would suggest just migrating to hatch and hatch-vcs in this PR and then switching to hatch commands in the CI in the follow-up PRs.

+1 to this approach

Saransh-cpp:
Thanks! Will work on this on weekend.

src/zarr/_storage/v3.py
```diff
@@ -163,7 +163,7 @@ def setitems(self, values):
         values = {self._normalize_key(key): val for key, val in values.items()}
 
         # initialize the /data/root/... folder corresponding to the array!
-        # Note: zarr.tests.test_core_v3.TestArrayWithFSStoreV3PartialRead fails
+        # Note: .tests.test_core_v3.TestArrayWithFSStoreV3PartialRead fails
```
Saransh-cpp, line 166:
```suggestion
        # Note: tests.test_core_v3.TestArrayWithFSStoreV3PartialRead fails
```

docs/contributing.rst
```diff
@@ -335,7 +335,7 @@ of the storage specification that is currently implemented is stored under the
 
 Note that the Zarr test suite includes a data fixture and tests to try and ensure that
 data format compatibility is not accidentally broken. See the
-:func:`test_format_compatibility` function in the :mod:`zarr.tests.test_storage` module
+:func:`test_format_compatibility` function in the :mod:`.tests.test_storage` module
```
Saransh-cpp, line 338:
```suggestion
:func:`test_format_compatibility` function in the :mod:`tests.test_storage` module
```

src/zarr/meta.py
```diff
@@ -459,7 +459,7 @@ def _encode_storage_transformer_metadata(
 
     @classmethod
     def _decode_storage_transformer_metadata(cls, meta: Mapping) -> "StorageTransformer":
-        from zarr.tests.test_storage_v3 import DummyStorageTransfomer
+        from tests.test_storage_v3 import DummyStorageTransfomer
```
Saransh-cpp, line 462: Why is a class from tests imported in user-facing functionality?

d-v-b: we don't know! see https://github.com/zarr-developers/zarr-python/pull/1601

Saransh-cpp:
This should be ready for a review. The tests pass for me locally and the package builds correctly.

The failing mypy test looks unrelated to this PR.

I have also force-pushed because it was much easier to migrate from scratch than resolving the conflicts in this PR.

src/zarr/meta.py
Saransh-cpp: Oh, I see, haha! No worries, I'll rebase this once your PR is merged!

Saransh-cpp:
Should be ready now

.git_archival.txt
```diff
@@ -0,0 +1,4 @@
+node: $Format:%H$
+node-date: $Format:%cI$
+describe-name: $Format:%(describe:tags=true,match=*[0-9]*)$
+ref-names: $Format:%D$
```
jhamman, line 4: can you explain why we need this file? Looking at hatch-vcs, I don't see any requirement of this configuration.

tests/test_attrs.py
```diff
@@ -8,7 +8,7 @@
 from zarr.attrs import Attributes
 from zarr.storage import KVStore, DirectoryStore
 from zarr._storage.v3 import KVStoreV3
-from .util import CountingDict, CountingDictV3
+from tests.util import CountingDict, CountingDictV3
```
jhamman, line 11: this is a bit of surprising change. Is it required? Is this what hatch suggests packages do? Generally, I prefer a relative imports within the tests directory.

jhamman (approved):
Two small questions around the edges here. I think we should be good to go here soon though.

.git_archival.txt
Saransh-cpp: This file allows users to install `Zarr` from Git archives without encountering version issues. This is not a `hatch-vcs` thing, instead this file should've been introduced when `setuptools-scm` was introduced in Zarr - https://setuptools-scm.readthedocs.io/en/latest/usage/#builtin-mechanisms-for-obtaining-version-numbers. `hatch-vcs` uses `setuptools_scm` underneath.

Scientific Python also [recommends adding git archival support](https://learn.scientific-python.org/development/guides/packaging-simple/) -
```
You should also add these two files:

.git_archival.txt:

node: $Format:%H$
node-date: $Format:%cI$
describe-name: $Format:%(describe:tags=true,match=*[0-9]*)$
ref-names: $Format:%D$

And .gitattributes (or add this line if you are already using this file):

.git_archival.txt  export-subst

This will allow git archives (including the ones generated from GitHub) to also support versioning.
```

tests/test_attrs.py
Saransh-cpp: Oh yes, this was just a personal preference. Happy to revert it!

jhamman: Let's revert it. Happy to discuss this change elsewhere but let's keep this pr scope as small as possible.

.git_archival.txt
jhamman: Makes sense. Thanks for the explanation.

joshmoore:
There have been a number of breakages in conda land related to the `fixtures/` directory. Considering the scale of the rework, this might be a time to address the root issue (if it doesn't just do it by chance!)

see:
 - https://github.com/conda-forge/zarr-feedstock/pull/50
 - https://github.com/zarr-developers/zarr-python/pull/1146
 - etc. etc.

jhamman:
@joshmoore - let's sort out CI separately. Hatch is going to help us manage our various envs so let's get right on that after merging this.

Saransh-cpp:
This will require some minor changes in the feedstock. I'll keep a watch, or if I forget, please ping me on the next version update PR in Zarr's feedstock.

PR #1601, opened 2023-12-08
Don't import from tests

d-v-b:
In `main`, `DummyStorageTransformer` is defined in the test suite and imported by `zarr/storage/v3_storage_transformers`. Because the distributed version of `zarr` should not be importing code from the tests, I moved the definition of `DummyStorageTransformer` to `zarr/storage/v3_storage_transformers`, and now the tests that use this class can import it like any other zarr code.

In the test suite, I also changed imports of the form `from zarr.test.module.. import bla` to `from .module import bla`. 

Both of these changes are necessary for moving to a `src/zarr` layout.

xref: #1592

### DIFF
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -18,6 +18,8 @@ Release notes
 Unreleased (v3)
 ---------------
 
+Maintenance
+~~~~~~~~~~~
 
 * Remedy a situation where ``zarr-python`` was importing ``DummyStorageTransformer`` from the test suite. 
   The dependency relationship is now reversed: the test suite imports this class from ``zarr-python``.
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -13,10 +13,20 @@ Release notes
     # to document your changes. On releases it will be
     # re-indented so that it does not show up in the notes.
 
+.. _unreleased(v3):
+
+Unreleased (v3)
+---------------
+
+
+* Remedy a situation where ``zarr-python`` was importing ``DummyStorageTransformer`` from the test suite. 
+  The dependency relationship is now reversed: the test suite imports this class from ``zarr-python``.
+  By :user:`Davis Bennett <d-v-b>` :issue:`1601`.
+
 .. _unreleased:
 
-Unreleased
-----------
+Unreleased (v2)
+---------------
 
 Docs
 ~~~~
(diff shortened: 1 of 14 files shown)


d-v-b:
Does anyone have a problem with merging this? It will efforts to modernize the packaging of `zarr-python` over in #1592

d-v-b:
if nobody objects, I'm going to merge this in under 12 hours.

d-v-b:
@jhamman has offered to review, so I will hold off on merging unilaterally

docs/release.rst
```diff
@@ -61,6 +61,9 @@ Maintenance
 * Remove ``sphinx-rtd-theme`` dependency from ``pyproject.toml``.
   By :user:`Sanket Verma <MSanKeys963>` :issue:`1563`.
 
+* Remedy a situation where ``zarr-python`` was importing ``DummyStorageTransformer`` from the test suite. 
+  The dependency relationship is now reversed: the test suite imports this class from ``zarr-python``.
+  By :user:`Davis Bennett <d-v-b>` :issue:`1601`.
```
jhamman, line 66: Is this in the right place? We have not been doing release notes for the v3 branch and this seems like its mixed in with 2.17.

jhamman (approved):
Thanks @d-v-b! Just one question below.

docs/release.rst
d-v-b: oops, you are right! where should we put release notes for v3 stuff? :thinking:

jhamman: I've been thinking we'll do a review of the v3 branch and aggregate the meaningful things into the release notes based on the commit history and code review. But we can start adding notes now (just not where you put them 😉 ).

d-v-b: thoughts on something like this? 

```
[normal release notes above]

Unreleased (v3)
---------------


* Remedy a situation where ``zarr-python`` was importing ``DummyStorageTransformer`` from the test suite. 
  The dependency relationship is now reversed: the test suite imports this class from ``zarr-python``.
  By :user:`Davis Bennett <d-v-b>` :issue:`1601`.

.. _unreleased:

[v2 release notes appended below]

Unreleased (v2)
---------------

Docs
~~~~

* Minor correction and changes in documentation.
  By :user:`Sanket Verma <MSanKeys963>` :issue:`1509`.

...
```

jhamman: lgtm

d-v-b: added in https://github.com/zarr-developers/zarr-python/pull/1601/commits/8e790a73a622464f12cebd3fa097bb4599c3379d

d-v-b:
This is going in with temporary release notes that distinguish between v2 and v3 contributions. When we merge v3 into main, we will revert this alteration to the release notes.

PR #1237, opened 2022-10-30
Make sure fs exceptions are raised if not Missing

martindurant:
(for v2 for now)

When FSStore gets exceptions, these correctly become the fill value when they are translated to KeyError. However, some exceptions do not equate to missing keys but to other permanent problems such as PermissionError. This change makes sure they are correctly raised to the caller even within getitems (__getitem__ already was correct).

TODO:
* [ ] Add unit tests and/or doctests in docstrings
* [ ] Add docstrings and API docs for any new/modified user-facing classes and functions
* [ ] New/modified features documented in docs/tutorial.rst
* [ ] Changes documented in docs/release.rst
* [ ] GitHub Actions have all passed
* [ ] Test coverage is 100% (Codecov passes)

PR #1604, opened 2023-12-10
Make sure fs exceptions are raised if not MissingFs exceptions (clone)

itcarroll:
**note**: This PR is a copy of PR #1237 which had become stale. Description from @martindurant is:

> (for v2 for now)
> 
> When FSStore gets exceptions, these correctly become the fill value when they are translated to KeyError. However, some exceptions do not equate to missing keys but to other permanent problems such as PermissionError. This change makes sure they are correctly raised to the caller even within getitems (getitem already was correct).

TODO:
* (NA) Add docstrings and API docs for any new/modified user-facing classes and functions
* (NA) New/modified features documented in docs/tutorial.rst

### DIFF
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -1415,13 +1415,14 @@ class FSStore(Store):
     ) -> Mapping[str, Any]:
         keys_transformed = [self._normalize_key(key) for key in keys]
         results = self.map.getitems(keys_transformed, on_error="return")
-        # The function calling this method may not recognize the transformed keys
-        # So we send the values returned by self.map.getitems back into the original key space.
+        # Only recognized errors will prompt KeyErrors in the function calling this method
         for k, v in results.copy().items():
             if isinstance(v, self.exceptions):
                 results.pop(k)
             elif isinstance(v, Exception):
                 raise v
+        # The function calling this method may not recognize the transformed keys, so we
+        # send the values returned by self.map.getitems back into the original key space.
         return {keys[keys_transformed.index(rk)]: rv for rk, rv in results.items()}
 
     def __getitem__(self, key):
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -1414,9 +1414,14 @@ class FSStore(Store):
         self, keys: Sequence[str], *, contexts: Mapping[str, Context]
     ) -> Mapping[str, Any]:
         keys_transformed = [self._normalize_key(key) for key in keys]
-        results = self.map.getitems(keys_transformed, on_error="omit")
+        results = self.map.getitems(keys_transformed, on_error="return")
         # The function calling this method may not recognize the transformed keys
         # So we send the values returned by self.map.getitems back into the original key space.
+        for k, v in results.copy().items():
+            if isinstance(v, self.exceptions):
+                results.pop(k)
+            elif isinstance(v, Exception):
+                raise v
         return {keys[keys_transformed.index(rk)]: rv for rk, rv in results.items()}
 
     def __getitem__(self, key):
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -1413,17 +1413,23 @@ class FSStore(Store):
     def getitems(
         self, keys: Sequence[str], *, contexts: Mapping[str, Context]
     ) -> Mapping[str, Any]:
-        keys_transformed = [self._normalize_key(key) for key in keys]
-        results = self.map.getitems(keys_transformed, on_error="return")
-        # Only recognized errors will prompt KeyErrors in the function calling this method
-        for k, v in results.copy().items():
+        keys_transformed = {self._normalize_key(key): key for key in keys}
+        results_transformed = self.map.getitems(list(keys_transformed), on_error="return")
+        results = {}
+        for k, v in results_transformed.items():
             if isinstance(v, self.exceptions):
-                results.pop(k)
+                # Cause recognized exceptions to prompt a KeyError in the
+                # function calling this method
+                continue
             elif isinstance(v, Exception):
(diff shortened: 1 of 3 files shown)


zarr/tests/test_storage.py
```diff
@@ -1337,6 +1337,24 @@ def test_s3_complex(self):
...
+        del m.store["/test/out.zarr/data/1"]
+        assert g.store.getitems(["data/1"], contexts={}) == {}  # not found
+        with pytest.raises(Exception):
+            # None is bad data, as opposed to missing
+            g.store.getitems(["data/0", "data/1"], contexts={})
+        with pytest.raises(Exception):
+            # None is bad data, as opposed to missing
+            arr[:]
+        # clear the global memory filesystem's store for use in other tests
+        m.store.clear()
```
jhamman, line 1356: This seems somewhat problematic. Is there a better way to isolate this filesystem, perhaps by moving the filesystem (`m`) to a fixture? The concern here is that if this test fails before this step, the clean up will not happen.

zarr/storage.py
```diff
@@ -1414,9 +1414,15 @@ def getitems(
         self, keys: Sequence[str], *, contexts: Mapping[str, Context]
     ) -> Mapping[str, Any]:
         keys_transformed = [self._normalize_key(key) for key in keys]
-        results = self.map.getitems(keys_transformed, on_error="omit")
-        # The function calling this method may not recognize the transformed keys
-        # So we send the values returned by self.map.getitems back into the original key space.
+        results = self.map.getitems(keys_transformed, on_error="return")
+        # Only recognized errors will prompt KeyErrors in the function calling this method
+        for k, v in results.copy().items():
```
jhamman, line 1419: It would be nice to rework this to avoid the `copy` here.

zarr/tests/test_storage.py
itcarroll: Resolved with fixture, thanks for the suggestion.

zarr/storage.py
itcarroll: With needing to return a new dictionary (with untransformed keys), there was essentially already a shallow copy. Re-organized to take advantage of that and got good speedup.

itcarroll:
It would be nice to get the workflows authorized so I could see what more to fix and document in release notes.

itcarroll:
Failing test is the same seen in #1611, something with Azure connection.

joshmoore:
> It would be nice to get the workflows authorized so I could see what more to fix and document in release notes.

Apologies for the slowness. It's the end of year crunch. Actions launched.

> Failing test is the same seen in #1611, something with Azure connection.

IIUC, this has been happening more frequently recently. A relaunch usually helps. I'll keep an eye on it.

MSanKeys963:
All green! 🚀

zarr/storage.py
itcarroll: In my view this is resolved, but I'm leaving the comment open for further consideration or approval.

martindurant: Right, this is not copying the data blobs, so I don't think it's a problem.

zarr/tests/test_storage.py
```diff
@@ -1337,6 +1343,21 @@ def test_s3_complex(self):
         )
         assert (a[:] == -np.ones((8, 8, 8))).all()
 
+    def test_exceptions(self, memory_store):
+        path = memory_store.path
+        m = memory_store.fs
+        g = zarr.open_group(memory_store, mode="w")
+        arr = g.create_dataset("data", data=[1, 2, 3, 4], dtype="i4", compression=None, chunks=[2])
+        m.store[path + "/data/0"] = None
```
martindurant, line 1351: Can there be a mock (maybe on fsspec.FSMap) to give a specific expected exception? Without that, the `raise` case in the changed code is not being tested, only the OK case, because a value is getting passed back.

Do we also need to test the single-item case?

joshmoore: @itcarroll : any thoughts here?

itcarroll: I can take a stab at the suggesiton next week. Thanks for the reminder.

itcarroll: I think I did this without a mock by taking advantage of being able to "corrupt" the fsspec memory store: setting file objects to `None` rather than some bytes causes an `AttributeError` on reads which is not one of the FSStore.exceptions. The test now traverses both the `continue` case (silences its own `KeyError`) and the `raise` case (from the `AttributeError`) for `getitems`. It also covers the raise case for `__getitem__`, which gets called for scalar datasets.

martindurant: +1

itcarroll: @joshmoore Could you authorize workflows, plz? Thank you!

itcarroll:
@joshmoore Trying to help with another merge to main. Maybe not helpful though ...

itcarroll:
Scan of the 178 failed tests indicates `azure.core.exceptions.HttpResponseError` from tests `WithABSStore` are the culprit.

MSanKeys963:
@itcarroll, looks like re-running the jobs worked!

MSanKeys963:
@itcarroll, it looks like re-running the jobs worked!

itcarroll:
@joshmoore I know that no issue was created to precipitate @martindurant's original PR, but this fixes an actual, real, silent, bug, so it would be great to finish it off. Thanks!

jhamman (approved):

jhamman:
This has been quite the quest -- thanks @itcarroll for hanging in there. @martindurant - are you happy with this solution? Would love to get this into a release (tomorrow 🤞).

PR #1612, opened 2023-12-18
Enable checking of untyped definitions

dstansby:
This enables some more mypy checks. The first two didn't need any fixes. `check_untyped_defs` requires quite a lot of fixes. I've made some of them, and added a list of modules that need a lot of fixes to ignore for now. The idea here is that at least in any other newly added modules untyped definitions will be checked.

TODO:


zarr/storage.py
```diff
@@ -833,7 +834,7 @@ def __getstate__(self):
 
     def __setstate__(self, state):
         root, cls = state
-        self.__init__(root=root, cls=cls)
+        self.__init__(root=root, cls=cls)  # type: ignore[misc]
```
dstansby, line 837: These igore statements are all for the following warning. I'm not sure if there's a good way to fix instead of ignoring the warning?

> Accessing "__init__" on an instance is unsound, since instance.__init__ could be from an incompatible subclass

d-v-b: pretty sure the right thing here is to define`__reduce__`, which I started over in #1089. It's stalled, but very simple to pick up I think

dstansby: 👍  - do you want #1089 to be finished before merging this PR, or would it be okay to merge this with the `# type: ignore` comments and then get rid of them when #1089 is done?

d-v-b: I think blinding mypy to our sins with `#type: ignore` is fine for now

dstansby:
I'll close this since it's gone in to the v3 branch

Issue #1623, opened 2024-01-03
[v3] Refactor metadata classes


normanrz:
In the current V3 draft version, all metadata classes are closely modeled after the json representation. For example, the regular chunk grid is represented as this json object:
```json
{
  "name": "regular",
  "configuration": { "chunk_shape": [2, 5] }
}
```
In Python, these objects are implemented as 2 classes
```python
@frozen
class RegularChunkGridConfigurationMetadata:
    chunk_shape: ChunkCoords


@frozen
class RegularChunkGridMetadata:
    configuration: RegularChunkGridConfigurationMetadata
    name: Literal["regular"] = "regular"

ChunkGridMetadata = Union[RegularChunkGridMetadata]
```

This is an artifact of using cattrs for (de)serialization. However, we were thinking of dropping the attrs dependency. That would open the way for a more ergonomic class design that serializes to the same json structure.
```python
class ChunkGrid(ABC):
    pass

class RegularChunkGrid(ChunkGrid):
    chunk_shape: ChunkCoords
```

jhamman:
@normanrz - I like this suggestion. Perhaps you could expand on what the serialization/deserialization flow might look like. In the proposal to basically unpack the JSON object by key in a sequential model like this: 

```python
zarr_json = json.loads(...)

if zarr_json["node_type"] == "array":
    ...

    if zarr_json["chunk_grid"]["name"] == "regular":
        chunk_grid = RegularChunkGrid(chunk_shape=zarr_json["chunk_grid"]["configuration"]["chunk_shape"])

array_meta = ArrayMetadata(..., chunk_grid=chunk_grid, ...)
```

normanrz:
I think it could work via delegation from the parent classes:

```python
class ChunkGrid:
    @classmethod
    def from_dict(cls, val: Dict[str, Any]) -> ChunkGrid:
        assert isinstance(val, dict)
        name = val["name"]
        if name == "regular":
            return RegularChunkGrid.from_dict(val)
        else:
            raise NotImplementedError

class RegularChunkGrid(ChunkGrid):
    @classmethod
    def from_dict(cls, val: Dict[str, Any]) -> RegularChunkGrid:
        assert isinstance(val, dict)
        assert val["name"] == "regular"
        assert _validate_int_list(val["configuration"]["chunk_shape"])
        return cls(chunk_shape=tuple(val["configuration"]["chunk_shape"]))

...

zarr_json = json.loads(...)

[...]

PR #1632, opened 2024-01-12
Codecs without array metadata

normanrz:
This PR removes the `array_metadata` attribute from the codecs. That makes it possible to construct Codec classes that can be reused for multiple arrays. It is kind of a precursor for #1623.

Instead there is now a `chunk_metadata` argument passed to the `encode` and `decode` methods. This chunk metadata can be the same object for all chunks, which is valid for regular chunk grids. When we have variable chunking in the future, each chunk can have its own metadata passed through the codec pipeline.

There are 2 new methods on the Codec interface:
- `validate` validates a codec configuration against array metadata. Consequently, it can only be called after the array metadata has been constructed (which also contains a list of codecs). For example, `validate` can be used to validate that the number of dimensions in the `order` attribute in the `transpose` configuration matches the number of dimensions in the array. It cannot modify the codec configuration but raise exceptions.
- `evolve` can be used to create a new codec with "fixed" configuration. For example, the `blosc` codec has a `typesize` attribute, which can be trivially be determined when knowing the dtype of the array. It would be poor UX to require this to be explicitly set by the user when constructing the codec. The `evolve` method can perform this modification. It only has 2 arguments for now `ndim` and `data_type`. It is called before constructing the array metadata so we can't pass that in. I don't think this is the best interface we can come up with. Ideas welcome.

TODO:


zarr/v3/metadata.py
```diff
@@ -154,20 +163,19 @@ class ShardingCodecIndexLocation(Enum):
 
 
 @frozen
-class CoreArrayMetadata:
-    shape: ChunkCoords
+class ChunkMetadata:
```
d-v-b, line 166: I'm thinking about the name for this class:

- the 'chunk" part.  None of the attributes of this class are specific to chunks per se -- if we aliased `chunk-shape` with `shape`, `data_type` to `dtype`, then we have a generic specification of a NDArray-like that wouldn't look out of place in numpy. So what if we drop the "chunk" references in the name and attributes?
The old name was `CoreArrayMetadata`. "core array metadata" feels like a better _description_ than "chunk metadata", but it's a bit long, and I don't think we need to semantically distinguish between "core" arrays and "non-core arrays (i.e., chunks)" in code like this. 

- the "metadata" part: I'm not sure that I would call the shape and dtype of an array "metadata" in the conventional sense -- "property" seems like a better fit? But i'm not sure. And we are already in a module called `metadata`, so repeating that word in class names might be redundant; what do you think of moving from the `<Thing>Metadata` pattern to `<Thing>`, as long as the "metadata-ness" is conveyed by the module name?

A couple of alternative names for this class, off the top of my head:
- `ArraySpec`
- `ArrayInfo`
- `ArrayProps`
- `NDArray` (I kind of prefer this one, as long as we namespace it in the right module to make things unambiguous)

zarr/v3/metadata.py
```diff
@@ -154,20 +163,19 @@ class ShardingCodecIndexLocation(Enum):
...
     chunk_shape: ChunkCoords
     data_type: DataType
     fill_value: Any
-    runtime_configuration: RuntimeConfiguration
+    runtime_configuration: RuntimeConfiguration = runtime_configuration("C")
```
d-v-b, line 170: Following up on the comment above about the semantic target for this class, I'm not sure `runtime_configuration` is a good fit on a class that basically is an abstract (no data) description of an array. I'm pretty sure that`runtime_configuration` is included here because functions that consume this class need  a `runtime_configuration` argument, but I would argue that this is a problem for the callers of those functions, i.e. if code is calling a function that requires a `runtime_configuration`, then lets just make it part of the signature and callers can deal with it formally.

d-v-b: that being said, we can deal with this kind of refactoring separately from this PR

zarr/v3/metadata.py
```diff
@@ -191,9 +199,13 @@ def dtype(self) -> np.dtype:
     def ndim(self) -> int:
         return len(self.shape)
 
-    def get_core_metadata(self, runtime_configuration: RuntimeConfiguration) -> CoreArrayMetadata:
-        return CoreArrayMetadata(
-            shape=self.shape,
+    def get_chunk_metadata(
+        self, _chunk_coords: ChunkCoords, runtime_configuration: RuntimeConfiguration
```
d-v-b, line 203: Neither `_chunk_coords` nor `runtime_configuration` get used here. I'm guessing that this function takes `_chunk_coords` as an argument in anticipation of irregular chunking, but what's the use for `runtime_configuration`? Maybe a docstring could help here

zarr/v3/codecs/transpose.py
```diff
@@ -32,82 +31,70 @@ class TransposeCodecMetadata:
...
+            runtime_configuration=chunk_metadata.runtime_configuration,
         )
 
-    async def decode(
-        self,
-        chunk_array: np.ndarray,
-    ) -> np.ndarray:
-        inverse_order = [0 for _ in range(self.array_metadata.ndim)]
+    async def decode(self, chunk_array: np.ndarray, chunk_metadata: ChunkMetadata) -> np.ndarray:
+        inverse_order = [0 for _ in range(chunk_metadata.ndim)]
```
d-v-b, line 85: A shorter alternative here is `[0] * chunk_metadata.ndim`

zarr/v3/codecs/transpose.py
```diff
@@ -32,82 +31,70 @@ class TransposeCodecMetadata:
...
         for x, i in enumerate(self.order):
             inverse_order[x] = i
         chunk_array = chunk_array.transpose(inverse_order)
         return chunk_array
 
     async def encode(
-        self,
-        chunk_array: np.ndarray,
+        self, chunk_array: np.ndarray, chunk_metadata: ChunkMetadata
```
d-v-b, line 92: is it important that this method operate on _chunks_, as opposed to _arrays_ (of which chunks are a subset)? If we want these functions to be agnostic to the semantics of chunking, and scoped more broadly to transforming arrays / bytes, then maybe renaming the `chunk_array` and `chunk_metadata` parameters to `array` and `array_metadata` would make sense?

zarr/v3/codecs/transpose.py
```diff
@@ -32,82 +31,70 @@ class TransposeCodecMetadata:
...
     ) -> Optional[np.ndarray]:
         chunk_array = chunk_array.transpose(self.order)
         return chunk_array
 
-    def compute_encoded_size(self, input_byte_length: int) -> int:
+    def compute_encoded_size(self, input_byte_length: int, _chunk_metadata: ChunkMetadata) -> int:
```
d-v-b, line 97: Thoughts on `compute_encoded_size` vs `encoded_size`, or `get_encoded_size`?

zarr/v3/array_v2.py
```diff
@@ -476,7 +484,11 @@ async def convert_to_v3_async(self) -> Array:
 
         if self.metadata.order == "F":
             codecs.append(
-                TransposeCodecMetadata(configuration=TransposeCodecConfigurationMetadata(order="F"))
+                TransposeCodecMetadata(
+                    configuration=TransposeCodecConfigurationMetadata(
+                        order=tuple(self.metadata.ndim - x - 1 for x in range(self.metadata.ndim))
```
d-v-b, line 489: possible alternative: `order=tuple(reversed(range(self.metadata.ndim)))`

zarr/v3/metadata.py
normanrz: > I'm guessing that this function takes _chunk_coords as an argument in anticipation of irregular chunking,

Exactly! We can leave it out for now.

zarr/v3/codecs/transpose.py
normanrz: `transpose` is an `ArrayArrayCodec` that is why the arguments are called `chunk_array` in contrast to `chunk_bytes`.

zarr/v3/codecs/transpose.py
normanrz: All fine with me. I just copied it straight from the implementation recommendations in the spec.

zarr/v3/metadata.py
normanrz: Fair enough, we can add `runtime_configuration` as separate argument to the codec signature.

zarr/v3/metadata.py
normanrz: I would find `NDArray` confusing, because it is not an array but only contains information about an array. `ArraySpec` or `ArrayInfo` are fine with me.

d-v-b:
> evolve can be used to create a new codec with "fixed" configuration. For example, the blosc codec has a typesize attribute, which can be trivially be determined when knowing the dtype of the array. It would be poor UX to require this to be explicitly set by the user when constructing the codec. The evolve method can perform this modification. It only has 2 arguments for now ndim and data_type. It is called before constructing the array metadata so we can't pass that in. I don't think this is the best interface we can come up with. Ideas welcome.

Am I right in thinking that `evolve` is basically a copy, with constraints? Personally I prefer the word "copy" over "evolve", because that makes the semantics a bit more clear -- based on the name alone, it's not intuitive to me whether "evolve" modifies an object in place, or returns a new object. Of course intuitions can change with time :wink:, so if lots of people are fine with "evolve" then I can adapt.

Can we be sure that `ndim` and `data_type` will be the _only_ free parameters for all the codecs? If not, it might make sense to have a more open-ended API like the one pydantic uses for [`model_copy`](https://docs.pydantic.dev/latest/api/base_model/#pydantic.main.BaseModel.model_copy), wherein you pass an optional `update` parameter, which is a dict of `{field_name: new_value}`. Even if we keep `evolve(self, ndim, data_type)`, it probably makes sense for that function to call a more general purpose copy method, unless we are super confident that `ndim` and `data_type` are the only free parameters.

normanrz:
> > evolve can be used to create a new codec with "fixed" configuration. For example, the blosc codec has a typesize attribute, which can be trivially be determined when knowing the dtype of the array. It would be poor UX to require this to be explicitly set by the user when constructing the codec. The evolve method can perform this modification. It only has 2 arguments for now ndim and data_type. It is called before constructing the array metadata so we can't pass that in. I don't think this is the best interface we can come up with. Ideas welcome.
> 
> Am I right in thinking that `evolve` is basically a copy, with constraints? Personally I prefer the word "copy" over "evolve", because that makes the semantics a bit more clear -- based on the name alone, it's not intuitive to me whether "evolve" modifies an object in place, or returns a new object. Of course intuitions can change with time 😉, so if lots of people are fine with "evolve" then I can adapt.

attrs calls it [`evolve`](https://www.attrs.org/en/stable/api.html#attrs.evolve) and dataclasses calls it [`replace`](https://docs.python.org/3/library/dataclasses.html#dataclasses.replace). I think it is a bit more than a `copy` because it changes the attributes.



> Can we be sure that `ndim` and `data_type` will be the _only_ free parameters for all the codecs? If not, it might make sense to have a more open-ended API like the one pydantic uses for [`model_copy`](https://docs.pydantic.dev/latest/api/base_model/#pydantic.main.BaseModel.model_copy), wherein you pass an optional `update` parameter, which is a dict of `{field_name: new_value}`. Even if we keep `evolve(self, ndim, data_type)`, it probably makes sense for that function to call a more general purpose copy method, unless we are super confident that `ndim` and `data_type` are the only free parameters.

Very likely, there will be additional required arguments in the future. I already declared `ndim` and `data_type` as keyword-only arguments. That essentially provides the dict-like semantics that you proposed with typing. Codecs can add a `**_kwargs` argument to their signature to eat up unknown arguments.

d-v-b:
> **attrs calls it [evolve](https://www.attrs.org/en/stable/api.html#attrs.evolve) and dataclasses calls it [replace](https://docs.python.org/3/library/dataclasses.html#dataclasses.replace). I think it is a bit more than a copy because it changes the attributes.**

ok! I'm fine with `evolve`. 

> Very likely, there will be additional required arguments in the future. I already declared ndim and data_type as keyword-only arguments. That essentially provides the dict-like semantics that you proposed with typing. Codecs can add a **_kwargs argument to their signature to eat up unknown arguments.

Another approach that could keep a constant function signature would be to use a single keyword argument, `update`, and implementations of `evolve` would define their own `TypedDict` for that kwarg. But I don't think this is a big deal either way.

One broad question: what do you think about using `dtype` internally instead of `data_type`? I know that the v3 spec uses the name `data_type`, but I think the user base of `zarr-python` is largely people working in the broader pydata ecosystem where `dtype` is standard, and so `data_type` as a kwarg / attriibute when `dtype` was available might be a source of friction.

normanrz:
> > Very likely, there will be additional required arguments in the future. I already declared ndim and data_type as keyword-only arguments. That essentially provides the dict-like semantics that you proposed with typing. Codecs can add a **_kwargs argument to their signature to eat up unknown arguments.
> 
> Another approach that could keep a constant function signature would be to use a single keyword argument, `update`, and implementations of `evolve` would define their own `TypedDict` for that kwarg. But I don't think this is a big deal either way.

Wouldn't that weaken the type "safety"?



> One broad question: what do you think about using `dtype` internally instead of `data_type`? I know that the v3 spec uses the name `data_type`, but I think the user base of `zarr-python` is largely people working in the broader pydata ecosystem where `dtype` is standard, and so `data_type` as a kwarg / attriibute when `dtype` was available might be a source of friction.

In zarrita, I only used `dtype` if it actually is an `np.dtype`. While `data_type` and `dtype` are similar, they are not exactly the same. Notably, the `dtype` also contains information about endianness. In the future, there will likely be data types that are not 1-to-1 translatable to dtypes.

d-v-b:
> In zarrita, I only used dtype if it actually is an np.dtype. While data_type and dtype are similar, they are not exactly the same. Notably, the dtype also contains information about endianness. In the future, there will likely be data types that are not 1-to-1 translatable to dtypes.

Does the the `byteorder` attribute of a `np.dtype` object not convey the endianness? And more broadly, I'm a little nervous about having two representations (`dtype` and `data_type`) of the same thing (same in the conceptual sense), as I can see this becoming a point of confusion. If `data_type` does end up using extra attributes beyond those present on `np.dtype`, I think the best way to do this would be by implementing an `np.dtype`-like class that either inherits from `np.dtype` or acts like it inherits from `np.dtype`, and I think we should call this attribute `dtype` inside the API, even if it's called `data_type` in the metadata document. In my opinion, there should probably be just 1 place where we convert an attribute called `data_type` to an attribute called `dtype`, and that's where in a function that consumes or generates a metadata document.

normanrz:
> Does the the `byteorder` attribute of a `np.dtype` object not convey the endianness?

It is the other way around: `dtype` has endianness via `byteorder`, `data_type` does not know about endianness.

d-v-b:
> > > Very likely, there will be additional required arguments in the future. I already declared ndim and data_type as keyword-only arguments. That essentially provides the dict-like semantics that you proposed with typing. Codecs can add a **_kwargs argument to their signature to eat up unknown arguments.
> > 
> > 
> > Another approach that could keep a constant function signature would be to use a single keyword argument, `update`, and implementations of `evolve` would define their own `TypedDict` for that kwarg. But I don't think this is a big deal either way.
> 
> Wouldn't that weaken the type "safety"?

In what way? I had a think about this, and a fundamental issue is that if the `evolve` function _itself_ takes any keyword arguments (e.g., `deep`, to specify whether a deep or shallow copy should be used), then we need those keyword arguments to be separate from the names of the attributes we are updating. Putting all the updated attributes in their own dict avoids potential name collisions. 

> > Does the the `byteorder` attribute of a `np.dtype` object not convey the endianness?
> 
> It is the other way around: `dtype` has endianness via `byteorder`, `data_type` does not know about endianness.

Aha, sorry for misreading you. In that case, If `np.dtype` is semantically "broader" than `data_type`, then it seems OK to use `np.dtype` internally, unless there's some something we need that it can't provide?

zarr/v3/metadata.py
d-v-b: let's just go with `ArraySpec` and if people hate it we can find a better name at that time?

zarr/v3/codecs/__init__.py
```diff
@@ -31,35 +30,29 @@
     from zarr.v3.codecs.crc32c_ import Crc32cCodecMetadata
 
 
+def _find_array_bytes_codec(
+    codecs: Iterable[Tuple[Codec, ArraySpec]]
+) -> Tuple[ArrayBytesCodec, ArraySpec]:
+    for codec, chunk_metadata in codecs:
+        if isinstance(codec, ArrayBytesCodec):
+            return (codec, chunk_metadata)
```
jhamman, line 38: @normanrz - can you comment on whether we have a check elsewhere to make sure we never have two `ArrayBytesCodec`s in the array metadata object?

zarr/v3/codecs/__init__.py
```diff
@@ -86,72 +79,87 @@ def _validate_codecs(codecs: List[Codec], array_metadata: CoreArrayMetadata) ->
...
+            yield (codec, chunk_metadata)
+            chunk_metadata = codec.resolve_metadata(chunk_metadata)
+
+    async def decode(
+        self,
+        chunk_bytes: BytesLike,
+        chunk_metadata: ArraySpec,
+        runtime_configuration: RuntimeConfiguration,
+    ) -> np.ndarray:
+        codecs = list(self._codecs_with_resolved_metadata(chunk_metadata))[::-1]
```
jhamman, line 103: do you think it would be easier to reason about this if we split the codec pipeline into three groups:

1. `BytesBytesCodec`s
2. `BytesArrayCodec`s
3. `ArrayArrayCodec`s

That would cut down on the number of instance checks later on.

zarr/v3/codecs/__init__.py
normanrz: Yes, that is in here https://github.com/normanrz/zarr-python/blob/codecs-without-array-metadata/zarr/v3/codecs/__init__.py#L49-L51

zarr/v3/codecs/__init__.py
d-v-b: alternatively, could we do a tagged union by adding a `codec_type` attribute with type `Literal["BytesBytes", "BytesArray", "ArrayArray"]`? then we just need 1 class definition and attribute checks instead of `isinstance`

d-v-b: ah, my proposal is actually not an alternative to @jhamman's idea. i think we could do both of these things.

d-v-b: Worth noting that the proposal to partition the codecs by type actually brings us back to the design of hdf5 / v2, where there is a separate filter pipeline (filters basically being `ArrayArray` transformations).

normanrz: Not sure it makes anything more readable: https://github.com/zarr-developers/zarr-python/pull/1632/commits/b825b4fab6043e3b36b13b9638aa58ade7c20057

zarr/v3/metadata.py
```diff
@@ -154,20 +163,18 @@ class ShardingCodecIndexLocation(Enum):
 
 
 @frozen
-class CoreArrayMetadata:
-    shape: ChunkCoords
+class ArraySpec:
     chunk_shape: ChunkCoords
```
d-v-b, line 167: Should this be `shape` instead of `chunk_shape`?

zarr/v3/codecs/__init__.py
```diff
@@ -110,48 +97,150 @@ def _validate_codecs(codecs: List[Codec], array_metadata: CoreArrayMetadata) ->
...
+            )
             if chunk_bytes_maybe is None:
                 return None
             chunk_bytes = chunk_bytes_maybe
 
         return chunk_bytes
 
-    def compute_encoded_size(self, byte_length: int) -> int:
-        return reduce(lambda acc, codec: codec.compute_encoded_size(acc), self.codecs, byte_length)
+    async def encode_partial(
```
d-v-b, line 225: what would the consequences be for removing `encode`, renaming `encode_partial` to `encode`, and using using exceptions for codecs that can't do partial encoding (i.e., for these codecs, if `selection` is anything other than a full selection, raise an exception, which the caller can handle). logically, `encode_partial` and `decode_partial` are just more general versions of `encode` and `decode`, so it feels safe to rely on this fact in the API design.

normanrz: I think `encode` and `encode_partial` are different enough to keep them separate. `encode_partial` has a number of limitations, the main one being that it is currently only works with a single codec. Also, `encode_partial` takes a reference to the store to load the data itself; `encode` takes a buffer.

zarr/v3/codecs/__init__.py
d-v-b: An attribute check would be better if we wanted to enable developers to use codecs without requiring that they inherit from our base classes (we could also achieve this with structural subtyping). And `isinstance(x, `ArrayArrayCodec`) is teeny bit more work to read / write than `x.codec_type == "ArrayArray"`. That being said, I don't think either of these complaints should block your efforts here, so feel free to ignore

PR #1654, opened 2024-02-07
Add env variables to sprint setup instructions

maxrjones:
A minor addition to the readme Ryan created for today's Zarr V3 sprint.


jhamman:
FYI - These variables are not used in the code path we're developing in the v3 branch so I'm not sure this addition was truly needed. We can discuss at the sprint kick off.

Issue #1648, opened 2024-02-06
[v3] Test environments and CI


jhamman:
We just merged #1592 which should make managing our various test environments much easier going forward. This ticket lays out a number of small tasks that can be taken to modernize our test environments and GitHub Actions Workflows. I'm sure I'm missing a few things here so feel free to comment / suggest alternatives.

### Hatch related improvements

- [x] Define dev environments in pyproject.toml using [Hatch environments](https://hatch.pypa.io/1.9/environment/)
  - Move `requirements_dev_*.txt` defining separate environments for "minimal", "dev". These should include parameterizations over python versions (3.10 and 3.11). 
  - Use these environments in our GitHub actions
- [x] Split out "minimal" environment into two separate environments
  - Minimum supported versions of all dependencies (see #1616)
  - Minimal dependencies (latest versions)
- [x] Create "docs" environment so developers can simply run: `hatch run docs:serve` 

### GitHub Actions related improvements

- [x] Explore removing the need to install a conda environment, instead relying on pure hatch environments
- [ ] Explore combining the "windows", "minimal", and "python-package" workflows into a single workflow
- [x] Run Mypy checks in each matrix run in addition to our pre-commit script

rabernat:
I want to help with this and am motivated to get it done by tomorrow in time for the Zarr sprint.

jhamman:
Quick update here. @rabernat got a good start on an initial Hatch env / CI setup in #1650. Still more to do here thought so leaving this open with the flag: `help wanted`.

maxrjones:
I'm planning to work on this at the Zarr sprint today, unless Ryan or others already plan to

dstansby:
Yep, I think this can be closed

jhamman:
Before we do, it would be nice to get a windows test environment up and running.

jhamman:
After https://github.com/zarr-developers/zarr-python/pull/2423 and #2418  go in, the last thing here is to get some diversity in platform into the test suite. Would be good to get at least 1 windows and MacOS entry in the main test matrix.

Issue #1593, opened 2023-12-07
Resolve Mypy errors in `v3` branch


jhamman:
The `v3` branch needs some Mypy attention. As expected, our port of the zarrita code (see https://github.com/zarr-developers/zarr-python/pull/1584) has left a number of mypy errors that would be great to resolve.

This ticket can be closed when:

- [x] we have CI running against the `v3` codebase
- [x] we have zero mypy errors when checked in a github action

Marking this as _help-wanted_!

jhamman:
This is feeling like a fairly high priority to start ticking off. For now, I've disabled the mypy checks in pre-commit (#1649) but this should be a short lived state.

DahnJ:
I wanted to get into zarr a little and this issue had "help wanted" tag, so I picked it up here: https://github.com/zarr-developers/zarr-python/pull/1692

I made a few choices in fixing some of the issues, hope it helps more than adds noise 🙂

jhamman:
closed by https://github.com/zarr-developers/zarr-python/pull/1692 and lots more work by @dstansby. See https://github.com/zarr-developers/zarr-python/issues/1783 for ongoing work on this topic.

PR #1656, opened 2024-02-07
Add test matrix for V3

maxrjones:
Part of #1648 

The `extras` are packages that should be optional (with import skipping), which could be addressed separately.

The current setup includes 8 combinations (below), would appreciate feedback on this.

```
$ hatch env show --ascii
                Standalone                
+---------+---------+----------+---------+
| Name    | Type    | Features | Scripts |
+=========+=========+==========+=========+
| default | virtual |          |         |
+---------+---------+----------+---------+
| docs    | virtual | docs     | build   |
|         |         |          | rtd     |
|         |         |          | serve   |
+---------+---------+----------+---------+
                                        Matrices                                         
+------+---------+---------------------------+----------+----------------+--------------+
| Name | Type    | Envs                      | Features | Dependencies   | Scripts      |
+======+=========+===========================+==========+================+==============+
| test | virtual | test.py3.10-1.24-minimal  | extra    | coverage       | run          |
|      |         | test.py3.10-1.26-minimal  |          | mypy           | run-coverage |
|      |         | test.py3.11-1.24-minimal  |          | pytest         | run-mypy     |
|      |         | test.py3.11-1.26-minimal  |          | pytest-asyncio | run-verbose  |
|      |         | test.py3.10-1.24-optional |          | pytest-cov     |              |
|      |         | test.py3.10-1.26-optional |          |                |              |
|      |         | test.py3.11-1.24-optional |          |                |              |
|      |         | test.py3.11-1.26-optional |          |                |              |
+------+---------+---------------------------+----------+----------------+--------------+
```

### DIFF
--- a/.github/workflows/test-v3.yml
+++ b/.github/workflows/test-v3.yml
@@ -30,3 +30,6 @@ jobs:
     - name: Run Tests
       run: |
         hatch run test:run
+    - name: Run mypy
+      run: |
+        hatch run test:run-mypy
--- a/.github/workflows/test-v3.yml
+++ b/.github/workflows/test-v3.yml
@@ -31,5 +31,6 @@ jobs:
       run: |
         hatch run test:run
     - name: Run mypy
+      continue-on-error: true
       run: |
         hatch run test:run-mypy
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -99,26 +99,13 @@ features = ["extra"]
 
 [[tool.hatch.envs.test.matrix]]
 python = ["3.10", "3.11"]
-numpy = ["1.24"]
-version = ["minimal-legacy"]
+numpy = ["1.24", "1.26"]
+version = ["minimal"]
 
 [[tool.hatch.envs.test.matrix]]
 python = ["3.10", "3.11"]
-numpy = ["1.26"]
+numpy = ["1.24", "1.26"]
 features = ["optional"]
-version = ["minimal-latest"]
-
-[[tool.hatch.envs.test.matrix]]
-python = ["3.10", "3.11"]
-numpy = ["1.24"]
-features = ["optional"]
-version = ["legacy"]
-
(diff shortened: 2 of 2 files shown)


maxrjones:
@jhamman I think this is ready for review if you have time. Mypy predictably fails (https://github.com/zarr-developers/zarr-python/issues/1593), so let me know if you think we should comment that step out in the CI for now.

.github/workflows/test-v3.yml
```diff
@@ -29,4 +29,7 @@ jobs:
         hatch env create
     - name: Run Tests
       run: |
-        hatch run test:run
\ No newline at end of file
+        hatch run test:run
```
jhamman, line 32: It would be nice if we could use github action's matrix feature here as well. Obviously, we're skipping lots of tests at the moment so running the whole matrix in one action is fine. But as the test suite comes back to life, I think we'll spread out. It seems like Hatch can support this with something like: 

```
hatch run +py=${{ matrix.python-version }} test:run
```

_Note: I wasn't able to figure out how to run individual matrix members with hatch variable selection but it seems like it should be possible: https://hatch.pypa.io/1.9/cli/reference/#hatch-env-run_

jhamman (commented):
Looks great @maxrjones. A few suggestions but this is close to going in.

jhamman:
I'm going to merge this now since I think its a solid improvement. 

@maxrjones - if you come back to #1648, we can look into splitting the matrix runs out into a GHA matrix.

maxrjones:
> I'm going to merge this now since I think its a solid improvement.
> 
> @maxrjones - if you come back to #1648, we can look into splitting the matrix runs out into a GHA matrix.

Thanks @jhamman! I'll be offline Feb 14 - March 5 and doubt that I will have time to work on this more before then. But, I'll be glad to contribute more after returning if there are still open V3 refactor issues.

PR #1661, opened 2024-02-08
Prototype of `object-store`-based Store implementation

kylebarron:
Prototype of [object-store](https://github.com/roeap/object-store-python) based store. 

[object-store](https://docs.rs/object_store/latest/object_store/) is a rust crate for interoperating with remote object stores like S3, GCS, Azure, etc. See the [highlights](https://docs.rs/object_store/latest/object_store/#highlights) section of its docs. It [doesn't even try to implement](https://docs.rs/object_store/latest/object_store/#why-not-a-filesystem-interface) a filesystem interface, instead focusing on the core atomic operations supported across object stores. This makes it a good candidate for use with a Zarr v3 Store.

[object-store-python](https://github.com/roeap/object-store-python) is a Python binding to object-store. With https://github.com/roeap/object-store-python/pull/6, I added async methods to the library. So the underlying Rust binary will return a Python coroutine that can be awaited.

That and related PRs haven't been merged yet, but you can try this out locally by installing
```
pip install git+https://github.com/kylebarron/object-store-python.git@dev#subdirectory=object-store
```
note that you need the Rust compiler on your computer. Install that by following [these docs](https://www.rust-lang.org/tools/install).

TODO:

* [ ] Implement multiple-range support in the underlying object-store-python library

### DIFF
--- a/src/zarr/v3/store/object_store.py
+++ b/src/zarr/v3/store/object_store.py
@@ -1,0 +1,78 @@
+from __future__ import annotations
+
+import asyncio
+from typing import List, Optional, Tuple
+
+from object_store import ObjectStore as _ObjectStore
+from object_store import Path as ObjectPath
+
+from zarr.v3.abc.store import Store
+
+
+class ObjectStore(Store):
+    supports_writes: bool = True
+    supports_partial_writes: bool = False
+    supports_listing: bool = True
+
+    store: _ObjectStore
+
+    def init(self, store: _ObjectStore):
+        self.store = store
+
+    def __str__(self) -> str:
+        return f"object://{self.store}"
+
+    def __repr__(self) -> str:
+        return f"ObjectStore({repr(str(self))})"
+
+    async def get(
+        self, key: str, byte_range: Optional[Tuple[int, Optional[int]]] = None
+    ) -> Optional[bytes]:
+        if byte_range is None:
+            return await self.store.get_async(ObjectPath(key))
+
+        start, end = byte_range
+        if end is None:
+            # Have to wrap a separate object-store function to support this
+            raise NotImplementedError
+
+        return await self.store.get_range_async(ObjectPath(key), start, end - start)
+
+    async def get_partial_values(
+        self, key_ranges: List[Tuple[str, Tuple[int, int]]]
+    ) -> List[bytes]:
+        # TODO: use rust-based concurrency inside object-store
+        futs = [self.get(key, byte_range=byte_range) for (key, byte_range) in key_ranges]
+
+        # Seems like a weird type match where `get()` returns `Optional[bytes]` but
+        # `get_partial_values` is non-optional?
+        return await asyncio.gather(*futs)  # type: ignore
+
+    async def exists(self, key: str) -> bool:
+        try:
+            _ = await self.store.head_async(ObjectPath(key))
+            return True
+        except FileNotFoundError:
+            return False
+
+    async def set(self, key: str, value: bytes) -> None:
+        await self.store.put_async(ObjectPath(key), value)
+
+    async def delete(self, key: str) -> None:
+        await self.store.delete_async(ObjectPath(key))
+
+    async def set_partial_values(self, key_start_values: List[Tuple[str, int, bytes]]) -> None:
+        raise NotImplementedError
+
+    async def list(self) -> List[str]:
+        objects = await self.store.list_async(None)
+        return [str(obj.location) for obj in objects]
+
+    async def list_prefix(self, prefix: str) -> List[str]:
+        objects = await self.store.list_async(ObjectPath(prefix))
+        return [str(obj.location) for obj in objects]
+
+    async def list_dir(self, prefix: str) -> List[str]:
+        list_result = await self.store.list_with_delimiter_async(ObjectPath(prefix))
+        common_prefixes = set(list_result.common_prefixes)
+        return [str(obj.location) for obj in list_result.objects if obj not in common_prefixes]


jhamman:
Amazing @kylebarron! I'll spend some time playing with this today.

kylebarron:
With https://github.com/roeap/object-store-python/pull/9 it should be possible to fetch multiple ranges within a file concurrently with range coalescing (using `get_ranges_async`). Note that this object-store API accepts multiple ranges _within one object_, which is still not 100% aligned with the Zarr `get_partial_values` because that allows fetches _across multiple objects_. 

That PR also adds a `get_opts` function which now supports "offset" and "suffix" ranges, of the sort `Range:N-` and `Range:-N`, which would allow removing the `raise NotImplementedError` on line 37.

martindurant:
https://github.com/martindurant/rfsspec/issues/3

src/zarr/v3/store/object_store.py
```diff
@@ -0,0 +1,78 @@
...
+        if end is None:
+            # Have to wrap a separate object-store function to support this
+            raise NotImplementedError
+
+        return await self.store.get_range_async(ObjectPath(key), start, end - start)
+
+    async def get_partial_values(
+        self, key_ranges: List[Tuple[str, Tuple[int, int]]]
+    ) -> List[bytes]:
+        # TODO: use rust-based concurrency inside object-store
```
martindurant, line 44: How I did it in rfsspec: https://github.com/martindurant/rfsspec/blob/main/src/lib.rs#L141

kylebarron: object-store has a built-in function for this: [`get_ranges`](https://docs.rs/object_store/latest/object_store/trait.ObjectStore.html#method.get_ranges). With the caveat that it only manages multiple ranges in a single file.

martindurant: Yes I know, but mine already did the whole thing, so I am showing how I did that.

martindurant:
> What are everbody's thoughts on having this in zarr-python vs. spinning it out as a separate package?

I suggest we see whether it makes any improvements first, so it's author's choice for now.

kylebarron:
While @rabernat has seen some impressive perf improvements in some settings when making many requests with Rust's tokio runtime, which would possibly also trickle down to a Python binding,  the biggest advantage I see is improved ease of use in installation.

A common hurdle I've seen is handling dependency management, especially around boto3, aioboto3, etc dependencies. Versions need to be compatible at runtime with any other libraries the user also has in their environment. And Python doesn't allow multiple versions of the same dependency at the same time in one environment. With a Python library wrapping a statically-linked Rust binary, you can remove all Python dependencies and remove this class of hardship.

The underlying Rust object-store crate is stable and under open governance via the Apache Arrow project. We'll just have to wait on [some discussion in object-store-python](https://github.com/roeap/object-store-python/pull/4#issuecomment-1937751934) for exactly where that should live. 

I don't have an opinion myself on where this should live, but it should be on the order of 100 lines of code wherever it is (unless the v3 store api changes dramatically)

jhamman:
> I suggest we see whether it makes any improvements first, so it's author's choice for now.

👍 

> What are everbody's thoughts on having this in zarr-python vs. spinning it out as a separate package?

I want to keep an open mind about what the core stores provided by Zarr-Python are. My current thinking is that we should just do a `MemoryStore` and a `LocalFilesystemStore`. Everything else can be opt-in by installing a 3rd party package. That said, I like having a few additional stores in the mix as we develop the store interface since it helps us think about the design more broadly.

normanrz:
> I want to keep an open mind about what the core stores provided by Zarr-Python are. My current thinking is that we should just do a `MemoryStore` and a `LocalFilesystemStore`. Everything else can be opt-in by installing a 3rd party package.

I agree with that. I think it is beneficial to keep the number of dependencies of core zarr-python small. But, I am open for discussion.

> That said, I like having a few additional stores in the mix as we develop the store interface since it helps us think about the design more broadly.

Sure! That is certainly useful.

itsgifnotjiff:
This is awesome work, thank you all!!!

src/zarr/v3/store/object_store.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+class ObjectStore(Store):
+    supports_writes: bool = True
+    supports_partial_writes: bool = False
+    supports_listing: bool = True
+
+    store: _ObjectStore
+
+    def init(self, store: _ObjectStore):
```
dcherian, line 19:
```suggestion
    def __init__(self, store: _ObjectStore):
```

kylebarron: ...oops! 🙈

kylebarron:
The `object-store-python` package is not very well maintained https://github.com/roeap/object-store-python/issues/24, so I took a few days to implement my own wrapper around the Rust `object_store` crate: https://github.com/developmentseed/object-store-rs

I'd like to update this PR soonish to use that library instead.

martindurant:
If the zarr group prefers [object-store-rs](https://github.com/developmentseed/object-store-rs), we can move it into the zarr-developers org, if you like. I would like to be involved in developing it, particularly if it can grow more explicit fsspec compatible functionality.

kylebarron:
I have a few questions because the `Store` API has changed a bit since the spring.

- There's a new `BufferPrototype` object. Is the `BufferPrototype` chosen by the store implementation or the caller? It would be very nice if this prototype could be chosen by the store implementation, because then we could return a `RustBuffer` object that implements the Python buffer protocol, but doesn't need to copy the buffer into Python memory.
- Similarly for puts, is `Buffer` guaranteed to implement the buffer protocol? Contrary to fetching, we can't do zero-copy puts right now with object-store

TomAugspurger:
> Is the BufferPrototype chosen by the store implementation or the caller? It would be very nice if this prototype could be chosen by the store implementation, because then we could return a RustBuffer object that implements the Python buffer protocol, but doesn't need to copy the buffer into Python memory.

This came up in the discussion at https://github.com/zarr-developers/zarr-python/pull/2426/files/5e0ffe80d039d9261517d96ce87220ce8d48e4f2#diff-bb6bb03f87fe9491ef78156256160d798369749b4b35c06d4f275425bdb6c4ad. By default, it's passed as `default_buffer_prototype` though I think the user can override at the call site or globally.

Does it look compatible with what you need?

kylebarron:
I think I'm confused why it's a parameter at all. Why shouldn't it return a protocol, and the store can implement whatever interface is most convenient to return data. 

Put another way: when the store chooses the return interface, it can ensure no memory copies, and then the caller of the store can decide whether they need to copy the memory elsewhere.

d-v-b:
I do think making the concrete return type of `store.get(key, prototype)` depend on its `prototype` argument is a bit of an API smell. If we had no other constraints, the obvious play would be to let `store.get` return what the underlying storage medium returns, and ask the caller of `store.get` to convert that return value into the buffer of choice. I am guessing there is some reason today why we don't do this, but I wonder what abstractions we should add / change to remove this reason.

kylebarron:
It makes sense that we'll always need a copy for CPU -> GPU, but I'd like to avoid situations where a store must copy data for CPU -> CPU. Right now that could be unavoidable depending on the buffer class the user passes in. Are we saying that the user needs to know the copy semantics of the underlying store?

martindurant:
> It makes sense that we'll always need a copy for CPU -> GPU

Actually no, I fully expect that the rapids team should be able to make a direct object-store/NIC->GPU store class and also to do filter decoding on the GPU ( https://docs.rapids.ai/api/kvikio/stable/zarr/ ). Whether any of that ends up here, is another matter.

jhamman:
@kylebarron - in terms of testing this, you should take a look at how we're doing this for other stores.

Basically, we've created a reusable test harness in `zarr.testing.store.StoreTests`. You can subclass that and it will run a bunch of store-only tests for you.

https://github.com/zarr-developers/zarr-python/blob/5807cba192773630dfd172715558423d5a32bb01/tests/test_store/test_remote.py#L107-L134

madsbk:
The idea with `BufferPrototype` is to let the user specify how he/she wants the data to the **_whole_** stack. Stores and codecs can then optimize for a specific buffer type or just call `.as_numpy_array()` to get host memory access (zero-copy if the data is already on in host memory). Similarly, they can use `.from_bytes()` to create a new Buffer from a `memoryview` (zero-copy). See [`MemoryStore.get()`](https://github.com/zarr-developers/zarr-python/blob/8a33df7fb0568c92a40c57874e07b68e371a4a59/src/zarr/storage/memory.py#L92). 

@kylebarron you should be able to create a Buffer from a `RustBuffer` zero-copy like:
```python

async def get(
        self,
        key: str,
        prototype: BufferPrototype,
        byte_range: tuple[int | None, int | None] | None = None,
    ) -> Buffer | None:

    the_rust_buffer: RustBuffer = # load data into a rust buffer
    return prototype.buffer.from_buffer(memoryview(the_rust_buffer))
```

Now, if the user request a GPU buffer, a later codec can decide to move the data to the GPU and maybe use nvCOMP to decompress the data etc.

kylebarron:
> The idea with `BufferPrototype` is to let the user tell how he/she wants the data to the **_whole_** stack. 

I'm not really a fan of this API, but I don't know the GPU side well enough to propose something else.

martindurant:
> This is what I think it means:

That is certainly what it means when used in fsspec; None in the first place is the same as 0 and None in the second place is the same as "end"/"size". Note that they can can be negative, so a suffix range would be `(-N, None)`.

I can't guarantee if the same convention is used here, but zarr blocks are either whole (None, None) or the exact range (start, stop) is known.

kylebarron:
> I can't guarantee if the same convention is used here, but zarr blocks are either whole (None, None) or the exact range (start, stop) is known.

If zarr blocks are either whole or known, then shouldn't the type hint for the store be

```py
ByteRangeRequest: TypeAlias = tuple[int, int] | None
```

?

d-v-b:
> @d-v-b it looks like you added the type hint in #2065, can you shed some light on this?

I introduced that type so that we could have exactly this conversation -- prior to its definition, we had various functions across the codebase that were taking a byte range parameter, but the type of that parameter wasn't defined in a central place. I'm not attached to this particular type! We can totally change it to something nicer, provided the semantics of that type covers all required use cases.

kylebarron:
> The suffix request is required for sharding.

Is there an example of a suffix request somewhere in this repo, so we can see how the range is passed as an argument to the store?

> Most HTTP servers including Object Storage services support suffix requests.

Except for Azure 😢. The default implementation of the `object-store` crate is to use suffix requests for other stores, but two requests for azure.

kylebarron:
> we had various functions across the codebase that were taking a byte range parameter, but the type of that parameter wasn't defined in a central place.

I think it's worth considering changing it to a dataclass, because the semantics of the tuple are not always clear. And elsewhere in the codebase, a "chunk slice" [refers to](https://github.com/zarr-developers/zarr-python/blob/5f3a5128a4e0cc7b160c5461f6e2575aaa3a3e77/src/zarr/codecs/sharding.py#L126-L128) start and _length_, not start and _end_.

d-v-b:
> > we had various functions across the codebase that were taking a byte range parameter, but the type of that parameter wasn't defined in a central place.
> 
> I think it's worth considering changing it to a dataclass, because the semantics of the tuple are not always clear. And elsewhere in the codebase, a "chunk slice" [refers to](https://github.com/zarr-developers/zarr-python/blob/5f3a5128a4e0cc7b160c5461f6e2575aaa3a3e77/src/zarr/codecs/sharding.py#L126-L128) start and _length_, not start and _end_.

Agreed, we definitely need to bump up the literacy of this type. I opened #2437 for this discussion.

normanrz:
> > The suffix request is required for sharding.
> 
> Is there an example of a suffix request somewhere in this repo, so we can see how the range is passed as an argument to the store?

https://github.com/zarr-developers/zarr-python/blob/main/src/zarr/codecs/sharding.py#L700-L702

kylebarron:
There's probably a bug or two in `get_partial_values` that I'd like to fix on Monday, but otherwise this PR should be ready to test with `obstore v0.3.0-beta.2` if anyone is so inclined.

It's particularly nice that we're able to match the `AsyncGenerator[str]` semantics of the list operations.

kylebarron:
I rewrote the implementation of `get_partial_values`, which should now support offset and suffix requests as well.

The type hinting in this PR passes. I'd love to get someone excited about trying this out.

martindurant:
>  I'd love to get someone excited about trying this out.

Is there an obstore release with all of your latest changes?

kylebarron:
> Is there an obstore release with all of your latest changes?

You can use https://pypi.org/project/obstore/0.3.0b2/ for now

TomAugspurger:
https://github.com/zarr-developers/zarr-python/actions/runs/11931641418/job/33291263276?pr=1661, which failed with


```
Creating environment: test.py3.13-1.25-minimal
Installing project in development mode
ERROR: Ignored the following versions that require a different python version: 1.21.2 Requires-Python >=3.7,<3.11; 1.21.3 Requires-Python >=3.7,<3.11; 1.21.4 Requires-Python >=3.7,<3.11; 1.21.5 Requires-Python >=3.7,<3.11; 1.21.6 Requires-Python >=3.7,<3.11; 1.26.0 Requires-Python <3.13,>=3.9; 1.26.1 Requires-Python <3.13,>=3.9
ERROR: Could not find a version that satisfies the requirement obstore==0.3.0b5; extra == "test" (from zarr[extra,test]) (from versions: 0.1.0b1)
```

seems to be because there isn't a wheel for 3.13 at https://pypi.org/project/obstore/0.3.0b5/#files.

I think that https://github.com/zarr-developers/zarr-python/actions/runs/11931641418/job/33291259870?pr=1661 has just our required dependencies. Your test file will need something like a `pytest.importorskip("obstore")` before you import any module importing `obstore`

jhamman:
@kylebarron - thanks for the continued work on this. We discussed this PR in our dev call today. Here's a brief summary of the outcome:

1. We want to include this in Zarr-Python so long as `obstore` is an optional dependency
2. We want to document the store as experimental at first (in the docstring)
3. Over time, we will figure out a way to allow the user to select which remote store is used when a string like `s3://bucket/my.zarr` is used but for now, we'll continue to use the fsspec-based `RemoteStore` for this. For now, users will have to construct the `ObjectStore` themselves.

How does this sound to you?

src/zarr/storage/object_store.py
```diff
@@ -0,0 +1,306 @@
...
+) -> AsyncGenerator[str, None]:
+    # We assume that the underlying object-store implementation correctly handles the
+    # prefix, so we don't double-check that the returned results actually start with the
+    # given prefix.
+    prefix_len = len(prefix)
+    async for batch in list_stream:
+        for item in batch:
+            # Yield this item if "/" does not exist after the prefix.
+            if "/" not in item["path"][prefix_len:]:
+                yield item["path"]
```
kylebarron, line 153: It was brought to my attention in https://github.com/developmentseed/obstore/issues/101 that in the current `object_store` implementation a final `/` is stripped from directories. So this `_transform_list_dir` will include the names of any sub directories.

Does zarr expect zero-byte files in normal usage? Or we could assume that a zero-byte file may be a directory.

kylebarron:
> 3. Over time, we will figure out a way to allow the user to select which remote store is used when a string like `s3://bucket/my.zarr` is used but for now, we'll continue to use the fsspec-based `RemoteStore` for this. For now, users will have to construct the `ObjectStore` themselves.
> 
> How does this sound to you?

I agree it should be an optional dependency. Happy to document it as experimental. Also happy for users to construct `ObjectStore` manually for now; we can document that usage.

I think we should get some benchmarks before deciding how to further integrate obstore.

src/zarr/storage/object_store.py
martindurant: fsspec has struggled with this for the longest time. My recommendation: you won't come up with a convention that matches the expectations of all users and backends, so choose something you're happy with and maintain it consistently. As long as the documentation is clear about how the convention works.

kylebarron: Thanks, that's useful background. What does that mean for Zarr? What does Zarr expect? Should we document behavior at the `Store` level for what all stores should implement? So at least the fsspec-based and obstore-based stores could have the same behavior

martindurant: I thought that listing and caring about directories versus files is something that only matters for V2, no?

kylebarron: I'm not intricately familiar with Zarr v3 but there are still `Store` methods for `list`, `list_prefix` and `list_dir`. Will it give incorrect behavior if one of the listed files is a directory prefix with no content?

martindurant: I am unsure, but I would have thought not.

kylebarron: Well, good for this backend to be considered experimental for now 🙂

maxrjones:
FYI Kyle and I chatted a bit about this PR today, resolving one test failure but hitting others. I'm hoping to work on improving the tests while at AGU a bit, possibly at the Pangeo hackday.

maxrjones:
I'm making some progress on the failing tests in https://github.com/kylebarron/zarr-python/pull/3 (currently up to 37/67 passing from 3/67).

I need some help understanding the expected behavior of this test: https://github.com/zarr-developers/zarr-python/blob/122760fc3440f2f35f5904e9dc7973d5188bdaca/src/zarr/testing/store.py#L297-L315 The test writes the following local Zarr store.

```
.
└── foo
    ├── c
    │   └── 1
    └── zarr.json
```
The expected keys for store.list_dir('foo') are `{'zarr.json', 'c'}`. The returned keys for ObjectStore is `('zarr.json',)`.  The docs state `Retrieve all keys and prefixes with a given prefix and which do not contain the character “/” after the given prefix.`. To me the obstore behavior seems correct and the fsspec behavior seems wrong based on this description because `c` is a folder (local store) and contains `"/"` after the given prefix. @jhamman @d-v-b could you please provide guidance here?

src/zarr/storage/object_store.py
```diff
@@ -0,0 +1,319 @@
...
+    def supports_writes(self) -> bool:
+        return True
+
+    async def set(self, key: str, value: Buffer) -> None:
+        buf = value.to_bytes()
+        await obs.put_async(self.store, key, buf)
+
+    async def set_if_not_exists(self, key: str, value: Buffer) -> None:
+        buf = value.to_bytes()
+        await obs.put_async(self.store, key, buf, mode="create")
```
jhamman, line 112:
```suggestion
    async def set_if_not_exists(self, key: str, value: Buffer) -> None:
        buf = value.to_bytes()
        try:
            await obs.put_async(self.store, key, buf, mode="create")
        except AlreadyExistsError:
            pass
```

maxrjones: thanks Joe, pre-commit recommended a slightly different version (https://github.com/kylebarron/zarr-python/pull/3/commits/86951b8eb6f87459d77318d37360b6424d2df49c) but this fixed one of the failing tests

maxrjones:
Leaving an update that we're down to 6 failing tests in https://github.com/kylebarron/zarr-python/pull/3. I probably won't have time to work on the tests again for several days at least.
#### Test 1
The store will need to be make serializable for Dask to work
`FAILED tests/test_store/test_object.py::TestObjectStore::test_serializable_store - TypeError: cannot pickle 'builtins.LocalStore' object`
#### Test 2-4
Obstore has the end byte on a range request as exclusive whereas Zarr-Python tests have this as inclusive. I want to confirm the intended behavior before proposing any more changes.
```
FAILED tests/test_store/test_object.py::TestObjectStore::test_get[byte_range3-c/0] - AssertionError
FAILED tests/test_store/test_object.py::TestObjectStore::test_get[byte_range3-foo/c/0.0] - AssertionError
FAILED tests/test_store/test_object.py::TestObjectStore::test_get[byte_range3-foo/0/0] - AssertionError
```
#### Test 5
Obstore doesn't like `None` as the start of the range request. I proposed updating `None` be the same as 0 for `get` but refusing `None` for  `get_partial_values` seems reasonable to me 🤷‍♂️ 
```
FAILED tests/test_store/test_object.py::TestObjectStore::test_get_partial_values[key_ranges3] - ValueError: Cannot pass `None` for the start of the range request.
```
#### Test 6
See https://github.com/zarr-developers/zarr-python/pull/1661#issuecomment-2543361881 for an explanation
```
FAILED tests/test_store/test_object.py::TestObjectStore::test_list_dir - AssertionError
```

maxrjones:
I decided to try out an S3Store in https://gist.github.com/maxrjones/0318a87d6179faf4eacbe01beb6d3eb6 (all previous tests were run with local or memory stores). I ran into an issue where `S3Store.from_url` only seems to consider the bucket. In the example included in the gist, a store is created with `from_url(nasa-veda-scratch/zarr-obstore-test/...)`. When listing objects within that store, objects with keys starting with `nasa-veda-scratch/icechunk-test/...` are also included in the results. @kylebarron do you know a way to specify a prefix to use for all objects when creating an ObjectStore?

d-v-b:
> I don't think there should be two different ways to specify "the beginning of the buffer".

this also holds for `zarr-python` -- I think we currently use [`None` and `0` to express the same thing](https://github.com/zarr-developers/zarr-python/blob/775979fce31175149a35aee04f0f7b8cb29ab9a4/src/zarr/abc/store.py#L22), which should be fixed.

maxrjones:
> > Test 2-4
Obstore has the end byte on a range request as exclusive whereas Zarr-Python tests have this as inclusive. I want to confirm the intended behavior before proposing any more changes.

> Obstore defines range requests as start and end where the end is exclusive. I do [explicitly document](https://developmentseed.org/obstore/latest/api/get/#obstore.get_range) that the range is exclusive in the docs. While the HTTP spec defines range request strings as inclusive byte ranges I believe. Perhaps fsspec and zarr-python implement the latter.

Sorry for all the churn here, but I realized the actual difference seems to be that ByteRangeRequest is actually [start, length] rather than [start, end]. I guess this makes the inclusive/exclusive end debate moot, but was still super surprising to me. See https://github.com/zarr-developers/zarr-python/issues/2567 for any further discussions on byte range representations.

maxrjones:
An update for anyone following this PR following a sync with @kylebarron today - the implementations [obstore.store.LocalStore](https://developmentseed.org/obstore/latest/api/store/local/) and [obstore.store.MemoryStore](https://developmentseed.org/obstore/latest/api/store/memory/) backed Zarr stores seem good to go, but AzureStore, GCSStore, and S3Stores will require more work to make the Zarr stores all relative to a specified root prefix (the object_store rust library only supports relativity to the bucket). Kyle and I plan to pair on finishing up this PR in the new year. I hope that work will unblock performance testing of Zarr V3 + obstore relative to Zarr V2 and Zarr V3 + icechunk, and would like to collaborate with Earthmover + others on those performance tests.

JoshCu:
Apologies if this too far off topic but this was the most active and relevant thread I could find for this. 

When using s3fs with zarr xarray and dask, I found that if the data subset being fetched was all within a single chunk then the download speed was single threaded and as a result quite slow. 
For larger subsets spanning multiple chunks this isn't an issue as it quickly becomes network bandwidth limited. With the exception of the end of a download where as threads finish downloading, you're stuck waiting for the last few threads to finish up with network bandwidth to spare. 

The [s3fs cat_file function](https://github.com/fsspec/s3fs/blob/51e3c80ef380a82081a171de652e2b699753be2b/s3fs/core.py#L1118) does take range input parameters but as far as I could tell they weren't being used. 
For my use case, I [overwrote the s3fs function](https://github.com/CIROH-UA/NGIAB_data_preprocess/blob/02f5b6065501b685b26c0111085b339b06a63a38/modules/data_processing/s3fs_utils.py#L13) to always check the size of the file being downloaded and chunk the download with range requests if it was over a certain size. Even with this additional logic slowing down all other requests for metadata etc, it cut down the ~5 minute download to 20s.

Happy to expand on this with more info, tests, and examples if it's useful?

jhamman:
@JoshCu - I think your comment is sufficiently different from the content of this PR to warrant a dedicated issue. Mind creating one and we can discuss there?

JoshCu:
> @JoshCu - I think your comment is sufficiently different from the content of this PR to warrant a dedicated issue. Mind creating one and we can discuss there?

Thanks :) 
https://github.com/zarr-developers/zarr-python/issues/2662

src/zarr/storage/object_store.py
```diff
@@ -0,0 +1,330 @@
...
+    store: _ObjectStore
+    """The underlying obstore instance."""
+
+    def __eq__(self, value: object) -> bool:
+        if not isinstance(value, ObjectStore):
+            return False
+
+        return bool(self.store.__eq__(value.store))
+
+    def __init__(self, store: _ObjectStore, *, read_only: bool = False) -> None:
```
jhamman, line 44: if `store` is not picklable, you'll need to implement `__getstate__` and `__setstate__` below. The basic strategy you'll want to use is:

- in `__getstate__`, extract the needed config to recreate an the `_ObjectStore`
- in `__setstate__`, use the config from above to recreate the `_ObjectStore` and set `self.store` directly

maxrjones: Thanks @jhamman. @kylebarron and I discussed xfailing the pickling tests in this PR and implementing `__getstate__` and `__setstate__`  later on, since `ObjectStore` could be used for non-distributed reading/writing in the meantime. Would you be open to punting this a bit down the road? IIUC Kyle was supportive of adding pickling eventually and I'd be glad to help but it doesn't seem crucial to us right now given that this store will be marked as experimental.

jhamman: Sure, let's add the `__getstate__` methods with a `NotImplementedError` so we can provide a clear error message when things go wrong.

kylebarron:
I released obstore 0.3 today, so I think we're primed to get this PR ready to go.

The failing test is this? https://github.com/zarr-developers/zarr-python/pull/1661#issuecomment-2543361881

maxrjones:
> I released obstore 0.3 today, so I think we're primed to get this PR ready to go.
> 
> The failing test is this? [#1661 (comment)](https://github.com/zarr-developers/zarr-python/pull/1661#issuecomment-2543361881)

Amazing! There are two remaining issues that I know of:
- the difference in `list_dir` from https://github.com/zarr-developers/zarr-python/pull/1661#issuecomment-2543361881
-  `get` with a byte range that doesn't exist errors with object_store and returns None IIRC with FsspecStore (xref https://github.com/zarr-developers/zarr-python/pull/2560)

IMO the two major decision affecting how much work is left are:
1. do we want this PR to focus only on LocalStore, which is pretty well tested, or also PrefixStore-wrapped S3Stores, etc which is currently untested?
2. should this PR be merged into Zarr-Python or exist outside via some extension/registry mechanism?

.pre-commit-config.yaml
```diff
@@ -34,6 +34,7 @@ repos:
           - numpy==2.1  # until https://github.com/numpy/numpy/issues/28034 is resolved
           - typing_extensions
           - universal-pathlib
+          - obstore==0.3.0-beta.8
```
maxrjones, line 37:
```suggestion
          - obstore==0.3.0
```

martindurant:
I don't see why we would need to move the implementation elsewhere, so long as the documentation is clear about the current state. So long as it is "experimental", we are not supposing that other members of zarr can necessarily handle maintenance of this piece (not that I expect @kylebarron to abandon it for any reason).

normanrz:
> > should this PR be merged into Zarr-Python or exist outside via some extension/registry mechanism?
> 
> I want to hear from others here at @zarr-developers/python-core-devs on this. Personally, I'd like to have it here as an "Experimental" store but it wouldn't be too hard to put in a standalone package either.

How mature is `obstore` at this point?
I am a bit torn. On the one hand, I would like zarr-python to remain a small library with lots of external extensions. On the other hand, I also want it to be batteries-included for users to get going quickly. However, this store seems to duplicate a lot of functionality that the FsspecStore already provides.

kylebarron:
> How mature is `obstore` at this point?

Eh, that's hard to answer. You could define maturity as the length of time between breaking releases?

Yesterday's 0.3 release had a bunch of new features like improved DX and async streaming uploads, but it only had two very minor breaking changes: using `(start, end)` for range requests consistently throughout the API, and returning Rust `Bytes` objects that implement the buffer protocol instead of returning a Python `bytes`. [^0]

> However, this store seems to duplicate a lot of functionality that the FsspecStore already provides.

Yes, you could argue this isn't providing anything totally "new" that fsspec doesn't do.

But I hope and expect it will provide performance improvements for large concurrent downloads, and thus be useful for Zarr. I'd certainly like to see similar numbers as what Earthmover found in their icechunk announcement.

We have multiple efforts going on to perform obstore benchmarking. @maxrjones is doing awesome work in https://github.com/maxrjones/zarr-obstore-performance/ specific to Zarr and https://github.com/geospatial-jeff/pyasyncio-benchmark is progressing on totally-general high-throughput benchmarking.

[^0]: Though incidentally, I'm thinking of making another tiny breaking change before 0.3 is too widespread, to require fully explicit semantics of the range request (similar to https://github.com/zarr-developers/zarr-python/pull/2585#issuecomment-2578172415)

src/zarr/storage/object_store.py
```diff
@@ -0,0 +1,361 @@
...
+        if not isinstance(
+            store,
+            (
+                obs.store.AzureStore,
+                obs.store.GCSStore,
+                obs.store.HTTPStore,
+                obs.store.S3Store,
+                obs.store.LocalStore,
+                obs.store.MemoryStore,
```
maxrjones, line 65:
```suggestion
                obs.store.MemoryStore,
                obs.store.PrefixStore,
```

maxrjones:
> But I hope and expect it will provide performance improvements for large concurrent downloads, and thus be useful for Zarr. I'd certainly like to see similar numbers as what Earthmover found in their icechunk announcement.

For a super simple first estimate, ObjectStore loads a 3.7 GB chunk of shape (24, 37, 721, 1440) with chunks (1, 1, 721, 1440) in ~65% of the time required by FsspecStore (4.5s vs 7.0s with n=50; from https://github.com/maxrjones/zarr-obstore-performance). I'd like to add Icechunk to this comparison and possibly include other benchmarks, but IMO this is super promising for the value of an object-store (via obstore) backed Zarr store.

.pre-commit-config.yaml
kylebarron, line 37: TODO: update to latest obstore release when we merge

src/zarr/storage/_object.py
```diff
@@ -0,0 +1,443 @@
...
+    Warnings
+    --------
+    ObjectStore is experimental and subject to API changes without notice. Please
+    raise an issue with any comments/concerns about the store.
+    """
+
+    store: _ObjectStore
+    """The underlying obstore instance."""
+
+    def __eq__(self, value: object) -> bool:
```
maxrjones, line 68:
```suggestion
    def __eq__(self, value: ObjectStore) -> bool:  # type: ignore[override]
```

src/zarr/storage/_object.py
```diff
@@ -0,0 +1,443 @@
...
+        super().__init__(read_only=read_only)
+        self.store = store
+
+    def __str__(self) -> str:
+        return f"object://{self.store}"
+
+    def __repr__(self) -> str:
+        return f"ObjectStore({self})"
+
+    def __getstate__(self) -> dict:
```
maxrjones, line 136:
```suggestion
    def __getstate__(self) -> dict[Any, Any]:
```

src/zarr/storage/_object.py
```diff
@@ -0,0 +1,443 @@
...
+        self.__dict__.update({"store": pickle.dumps(self.store)})
+        return self.__dict__.copy()
+
+    def __setstate__(self, state: dict) -> None:
```
maxrjones, line 140:
```suggestion
    def __setstate__(self, state: dict[Any, Any]) -> None:
```

maxrjones (commented):
These suggestions fix the typing failures.

kylebarron:
It looks like this is now passing all tests, with just Read the docs and codecov targets not hit. 

What are the final steps for this PR? Where should we write documentation?

.pre-commit-config.yaml
kylebarron: Note: obstore is currently on version 0.4.0-beta.4. I'm ready to publish 0.4.0 but wanted to ensure it works with zarr-python before publishing it.

jhamman:
> It looks like this is now passing all tests, with just Read the docs and codecov targets not hit.

🎉 

> What are the final steps for this PR? Where should we write documentation?

- [ ] Make CI happy (@maxrjones is on this)
- [ ] Write a short bit of documentation here: https://github.com/zarr-developers/zarr-python/blob/main/docs/user-guide/storage.rst. Note that this store is "currently experimental".

.pre-commit-config.yaml
kylebarron: Done

kylebarron:
obstore 0.4.0 [was released](https://developmentseed.org/obstore/latest/blog/2025/02/10/releasing-obstore-04/) and this PR was updated to use that latest version.

### Optional dependency

From [comment above](https://github.com/zarr-developers/zarr-python/pull/1661#issuecomment-2494126052)

> 1. We want to include this in Zarr-Python so long as `obstore` is an optional dependency

CI [had been failing](https://github.com/zarr-developers/zarr-python/actions/runs/13244698002/job/36968239552?pr=1661) because `obstore` was marked as an optional dependency but then unconditionally exported from `zarr.storage`.

I removed the re-export from `zarr.storage` but renamed `_object.py` to `obstore.py`. So anyone who wishes to use the `obstore`-backed class needs to run 
```py
import zarr.storage.obstore
```
themselves. When that import is run, `obstore` will be imported, and so `obstore` must exist in the environment at that time.

**Alternatively**

We could keep the re-export but ensure that `obstore.py` doesn't import obstore in any public scope. Thoughts?

### Future follow up PRs:

- https://github.com/zarr-developers/zarr-python/issues/2814. [This PR](https://github.com/zarr-developers/zarr-python/pull/1661/commits/5cc35081a0e7ed2536646ef71ad3d1552aba26eb) added `# type: ignore[arg-type]` to satisfy passing a buffer protocol object into `prototype.buffer.from_bytes`

jhamman:
Regarding the import of obstore, I suggest moving all imports from obstore into the class itself. Users should get an `ImportError` if they attempt to instantiate the Store without the required dependency.

martindurant:
You can have 
```
import obstore
self.obs = obstore
```
in `__init__` to avoid having to have import all over the place.

kylebarron:
We can move the imports into the class; that's fine.

> You can have
> 
> ```
> import obstore
> self.obs = obstore
> ```
> 
> in `__init__` to avoid having to have import all over the place.

It's not clear to me how the typing would work there? How would you define the type hint for `obs` so that `self.obs.get` would correctly access the same type hint as `obstore.get`? I'd rather import obstore a bunch than lose that type safety.

martindurant:
I'm not usually too worried by typing. I'm sure it can be fixed, but fighting with mypy is never fun.

kylebarron:
> We can move the imports into the class; that's fine.

Hmm, moving the imports inside the class caused a pickling error. Maybe this won't work:

```
FAILED tests/test_store/test_object.py::TestObjectStore::test_serializable_store - TypeError: cannot pickle 'module' object
```

https://github.com/zarr-developers/zarr-python/actions/runs/13295245421/job/37125558963?pr=1661#step:6:299

src/zarr/storage/_obstore.py
```diff
@@ -0,0 +1,475 @@
...
+                    return False
+            case MemoryStore():
+                if self.store is not value.store:
+                    return False  # Two memory stores can't be equal because we can't pickle memory stores
+        return True
+
+    def __init__(self, store: _UpstreamObjectStore, *, read_only: bool = False) -> None:
+        import obstore as obs
+
+        self.obs = obs
```
maxrjones, line 124:
```suggestion
```
@kylebarron this is what is causing the pickling issue. What's the motivation for storing the module on the instance?

maxrjones: hmm, nevermind - I see Martin's comment above. the tests do pass without `self.obs = obs` anyways with the current version, probably due to the other import locations

maxrjones: Last comment here - https://github.com/kylebarron/zarr-python/pull/14 shows how to use @martindurant's suggested approach while still being able to pickle the ObjectStore instance by removing `obs` in `__getstate__`.

danielgafni:
Typically optional runtime dependencies are handled the following way:

1. Avoid top-level imports of these dependencies (it is possible to allow them in some modules by following a certain convention but IMO it's easier to just forbid top-level imports completely). Forward references can be used for type checking to work:

```python
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from obstore import ObjectStore

def my_func(store: ObjectStore): ...

```

2. Actual imports are moved into functions and methods which use them at runtime: 

```python
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from obstore import ObjectStore

def my_func(store: ObjectStore):
    import obstore
    data = obstore.get(store, ...)
    ....

```

3. This can be nicely enforced by [banned-module-level-imports](https://docs.astral.sh/ruff/rules/banned-module-level-imports/#banned-module-level-imports-tid253)](https://docs.astral.sh/ruff/rules/banned-module-level-imports/#banned-module-level-imports-tid253) ruff rule.

kylebarron:
I think the latest version of this PR does what you suggest?

maxrjones:
@jhamman @d-v-b Kyle and I think this PR is ready for review. Would you have time to take a look at it or would you be willing to ask another zarr developer to help us with the review process? Thanks for the help so far

kylebarron:
This PR has been ready to merge for a while (there were some small edits after obstore 0.5 was released that made the code here cleaner, but which were not backwards incompatible). 

I'm inclined to provisionally vendor this code into an `obstore.zarr` module, documenting that the `obstore.zarr` module will be removed whenever this PR lands and is released. 

I want people to be able to start playing with this integration and to have some full-featured demos ready for the CNG conference in a month.

jhamman:
@kylebarron - thanks for the ping. We'll give this a review and get it into the next Zarr release.

src/zarr/storage/_obstore.py
```diff
@@ -0,0 +1,419 @@
...
+
+        if not self.read_only == value.read_only:
+            return False
+
+        return self.store == value.store
+
+    def __init__(self, store: _UpstreamObjectStore, *, read_only: bool = False) -> None:
+        import obstore as obs
+
+        if not isinstance(
```
TomAugspurger, line 70: What's the rationale for this check?

A potential downside is that obstore grows a new store implementation in the future, and we'll need to add it to the list allowed here.

The upside of this check is maybe catching an issue at `__init__` time rather than when you call a method (and it presumably fails. Unless it's duck typed appropriately).

If you want to keep this, then does obstore have a base class we could check against?

src/zarr/storage/_obstore.py
```diff
@@ -0,0 +1,419 @@
...
+        ):
+            raise TypeError(f"expected ObjectStore class, got {store!r}")
+        super().__init__(read_only=read_only)
+        self.store = store
+
+    def __str__(self) -> str:
+        return f"object://{self.store}"
+
+    def __repr__(self) -> str:
+        return f"ObjectStore({self})"
```
TomAugspurger, line 89:
```suggestion
        return f"type(self)({self})"
```

So that `__repr__` is correct for subclasses.

src/zarr/storage/_obstore.py
```diff
@@ -0,0 +1,419 @@
...
+
+    async def set(self, key: str, value: Buffer) -> None:
+        # docstring inherited
+        import obstore as obs
+
+        self._check_writable()
+        if not isinstance(value, Buffer):
+            raise TypeError(
+                f"ObjectStore.set(): `value` must be a Buffer instance. Got an instance of {type(value)} instead."
+            )
```
TomAugspurger, line 162: Any reason to do this check here, but not in `self.set_if_not_exists`? (I'd prefer to remove it)

src/zarr/storage/_obstore.py
```diff
@@ -0,0 +1,419 @@
...
+    """
+    Transform the result of list_with_delimiter into an async generator of paths.
+    """
+    list_result = await list_result_coroutine
+
+    # We assume that the underlying object-store implementation correctly handles the
+    # prefix, so we don't double-check that the returned results actually start with the
+    # given prefix.
+    prefixes = [obj.lstrip(prefix).lstrip("/") for obj in list_result["common_prefixes"]]
+    objects = [obj["path"].lstrip(prefix).lstrip("/") for obj in list_result["objects"]]
```
TomAugspurger, line 248: Would `.removeprefix` be correct here? (https://peps.python.org/pep-0616/)

src/zarr/testing/store.py
```diff
@@ -158,6 +158,10 @@ async def test_get(self, store: S, key: str, data: bytes, byte_range: ByteReques
         """
         Ensure that data can be read from the store using the store.get method.
         """
+        if isinstance(store, ObjectStore) and not data:
```
TomAugspurger, line 161: Would it be possible to put this skip / xfail on the `ObjectStore` subclass (and call `super().test_get(...)` for the cases you don't skip)?

Or we could remove this test case if we think it's not generally valid for all store implementations.

src/zarr/storage/_obstore.py
TomAugspurger: There's a test for this below that would need to be updated.

src/zarr/storage/_obstore.py
```diff
@@ -0,0 +1,419 @@
...
+        futs.append(_make_bounded_requests(store, path, bounded_ranges, prototype))
+
+    for request in other_requests:
+        futs.append(_make_other_request(store, request, prototype))  # noqa: PERF401
+
+    buffers: list[Buffer | None] = [None] * len(key_ranges)
+
+    # TODO: this gather a list of list of Response; not sure if there's a way to
+    # unpack these lists inside of an `asyncio.gather`?
+    for responses in await asyncio.gather(*futs):
```
TomAugspurger, line 415: I thought we have a `gather` implementation inside zarr that respects a concurrency limit. But now I'm not seeing it (just `zarr.core.common.concurrent_map`, which I think isn't quite right).

Anyway, maybe we'd want to do acquire an `asyncio.Semaphore` with a limit of `zarr.config.get("async.concurrency")` in each of these coroutines.

TomAugspurger (approved):
I think this looks great. My only questions are around a couple (IMO) unnecessary isinstance checks.

src/zarr/storage/_obstore.py
```diff
@@ -0,0 +1,419 @@
...
+                obs.store.LocalStore,
+                obs.store.MemoryStore,
+            ),
+        ):
+            raise TypeError(f"expected ObjectStore class, got {store!r}")
+        super().__init__(read_only=read_only)
+        self.store = store
+
+    def __str__(self) -> str:
+        return f"object://{self.store}"
```
dcherian, line 86: nit: this seems clearer to me.

```suggestion
        return f"object_store://{self.store}"
```

src/zarr/storage/_obstore.py
```diff
@@ -0,0 +1,419 @@
...
+
+    def __repr__(self) -> str:
+        return f"ObjectStore({self})"
+
+    def __getstate__(self) -> dict[Any, Any]:
+        state = self.__dict__.copy()
+        state["store"] = pickle.dumps(self.store)
```
dcherian, line 93: don't we have to pickle `read_only` too?

src/zarr/storage/_obstore.py
dcherian: nice catch~ `lstrip(prefix)` should be wrong . Worrying that a Store test hasn't caught that.

src/zarr/storage/_obstore.py
```diff
@@ -0,0 +1,419 @@
...
+        buf = value.to_bytes()
+        with contextlib.suppress(obs.exceptions.AlreadyExistsError):
+            await obs.put_async(self.store, key, buf, mode="create")
+
+    @property
+    def supports_deletes(self) -> bool:
+        # docstring inherited
+        return True
+
+    async def delete(self, key: str) -> None:
```
dcherian, line 180: Is there an obstore method we can wrap for `delete_dir` too? That gets called somewhat frequently, so it could be a nice but small improvement.

src/zarr/storage/_obstore.py
dcherian: I don't think we have this utility but are instead creating semaphores manually: https://github.com/zarr-developers/zarr-python/blob/f95c453f3835da7f8097a1b5302ea1b7a3d6e57f/src/zarr/core/group.py#L1384

dcherian (approved):
I don't have much to add to @TomAugspurger 's excellent review.

I pushed a commit that wires up one of our stateful tests. Things pass, which is great!

src/zarr/storage/_obstore.py
kylebarron: Is that not included in `self.__dict__`?

src/zarr/storage/_obstore.py
kylebarron: No, there's no specific `delete_dir` method.

src/zarr/storage/_obstore.py
kylebarron: Closed in [`0735a43` (#1661)](https://github.com/zarr-developers/zarr-python/pull/1661/commits/0735a437bf38740b2fe47a613652eb2bb17e486c)

src/zarr/storage/_obstore.py
TomAugspurger: I would expect it to be.

src/zarr/storage/_obstore.py
```diff
@@ -0,0 +1,419 @@
...
+        # docstring inherited
+        import obstore as obs
+
+        self._check_writable()
+        if not isinstance(value, Buffer):
+            raise TypeError(
+                f"ObjectStore.set(): `value` must be a Buffer instance. Got an instance of {type(value)} instead."
+            )
+        buf = value.to_bytes()
+        await obs.put_async(self.store, key, buf)
```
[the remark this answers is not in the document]
TomAugspurger: I'm planning to do some more testing with obstore and GPU arrays soonish and so would be happy to look into this more.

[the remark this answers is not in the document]
kylebarron: Created https://github.com/zarr-developers/zarr-python/issues/2925

src/zarr/storage/_obstore.py
kylebarron: I think I fixed this in https://github.com/zarr-developers/zarr-python/pull/1661/commits/c45c7d1bc27a720185dbcf0cdada58e8c84b7f0a if you can check my work.

src/zarr/storage/_obstore.py
kylebarron: Removed in [`50fd2b9` (#1661)](https://github.com/zarr-developers/zarr-python/pull/1661/commits/50fd2b9010a590ee89ba55a816aa44235b967c81)

pyproject.toml
```diff
@@ -64,6 +64,7 @@ keywords = ["Python", "compressed", "ndimensional-arrays", "zarr"]
 # User extras
 remote = [
     "fsspec>=2023.10.0",
+    "obstore==0.5.1",
```
TomAugspurger, line 67: Does this need to be an exact pin?

src/zarr/storage/_obstore.py
TomAugspurger: Looks good.

src/zarr/testing/store.py
kylebarron: Actually here we do still need this skip because the test cases are **not even partially satisfiable**. What https://github.com/apache/arrow-rs/pull/6751 fixed is that if there's a file with 100 bytes and you request range 90-110, you'll get the last 10 bytes. But these tests are akin to having a 100 byte-long file and requesting a range 110-120. That [isn't satisfiable](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Range_requests#partial_request_responses), and the HTTP spec says it should error with [416 Range Not Satisfiable](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/416).

TomAugspurger (approved):
Thanks. One last question about the dependency in the pyproject.toml.

src/zarr/testing/store.py
TomAugspurger: I ran into some issues with these tests in https://github.com/rapidsai/kvikio/pull/646 as well ("reading" pass the end of a file).

Removing these edge cases entirely is probably a bigger discussion, but IMO moving the skip to the subclass is worth keeping the base class agnostic to the store implementation.

TomAugspurger: Works for me.

kylebarron:
Now I'm just trying to get the tests to pass (re https://github.com/zarr-developers/zarr-python/pull/1661#discussion_r2006708116) and we should be good. (I can't get the tests to pass locally anyways; I get `botocore.exceptions.ClientError: An error occurred (IllegalLocationConstraintException)` on all the fsspec tests)

src/zarr/storage/_obstore.py
kylebarron: This needed to be
```py
    def __repr__(self) -> str:
        return f"{type(self).__name__}({self})"
```
or else the beginning of the repr was 
```
'<class \'zarr.storage._obstore.ObjectStore\'>
```

kylebarron:
In [`a3afa44` (#1661)](https://github.com/zarr-developers/zarr-python/pull/1661/commits/a3afa44ad39da8646e48c7b549925b8cb88115ee) I added [intersphinx](https://docs.readthedocs.com/platform/stable/guides/intersphinx.html) support, which allows for automatic interlinking with the obstore docs. But restructured text drives me insane, so I hope those docs are good enough

TomAugspurger:
Planning to merge this tomorrow if there aren't any objections.

TomAugspurger:
Thanks for the great work everyone!

kylebarron:
Thanks all! Just [published obstore 0.6](https://developmentseed.org/obstore/latest/blog/2025/03/24/releasing-obstore-06/), which adds easier, automatic-token-refreshing integration with Planetary Computer. And I was able to get [their zarr example](https://planetarycomputer.microsoft.com/docs/quickstarts/reading-zarr-data/) working with this latest main!




```py
import matplotlib.pyplot as plt
import pystac_client
import xarray as xr
from obstore.auth.planetary_computer import PlanetaryComputerCredentialProvider
from obstore.store import AzureStore

catalog = pystac_client.Client.open(
    "https://planetarycomputer.microsoft.com/api/stac/v1/"
)
collection = catalog.get_collection("daymet-daily-hi")
asset = collection.assets["zarr-abfs"]

credential_provider = PlanetaryComputerCredentialProvider.from_asset(asset)
azure_store = AzureStore(credential_provider=credential_provider)
zarr_store = ObjectStore(azure_store, read_only=True)
ds = xr.open_dataset(zarr_store, consolidated=True, engine="zarr")

fig, ax = plt.subplots(figsize=(12, 12))
ds.sel(time="2009")["tmax"].mean(dim="time").plot.imshow(ax=ax, cmap="inferno")
fig
```

<details>
<summary>uv pyproject.toml</summary>

```toml
[project]
name = "zarr-obstore-pc"
version = "0.1.0"
description = "Add your description here"
readme = "README.md"
requires-python = ">=3.12"
dependencies = [
    "matplotlib>=3.10.1",
    "obstore>=0.6.0",
    "pystac-client>=0.8.6",
    "xarray>=2025.3.0",
    "zarr",
]

[tool.uv.sources]
zarr = { git = "https://github.com/zarr-developers/zarr-python" }

[dependency-groups]
dev = [
    "ipykernel>=6.29.5",
]
```

</details>

ilan-gold:
Hi this PR is very exciting - I am curious, is the performance expected to be better than `fsspec` for similar ops (fetching from remote)? It could be good to highlight a bit why to choose this store specifically.  I would love to understand!

kylebarron:
Yes, I expect it to be significantly faster, but we don't have rigorous benchmarks yet. I'd love to see some Zarr benchmarks, and then maybe we can update the docs to reflect those.

itsgifnotjiff:
I am not sure if this is within the scope of your benchmarking but if you can test the single point query times and performance for Zarr store in the 100 Tb range that would be great. Zarr v2 had problems with both number of inodes required and performance in my experience.

The groups/tree addition along with the explosion of large scale data means Zarr stores are either already performant enough or not performant at all for different use cases (geospatial in mine).

kylebarron:
I don't personally use Zarr much, so ideally I want to enable other people to do benchmarking. But happy to pair or support in any way I can.

itsgifnotjiff:
Makes perfect sense. I hope I get to benchmark it later this year. I will link my potential findings here as well 😊. Thank you so much for your work.

maxrjones:
> I am not sure if this is within the scope of your benchmarking but if you can test the single point query times and performance for Zarr store in the 100 Tb range that would be great. Zarr v2 had problems with both number of inodes required and performance in my experience.
> 
> The groups/tree addition along with the explosion of large scale data means Zarr stores are either already performant enough or not performant at all for different use cases (geospatial in mine).

Hey @itsgifnotjiff Davis Bennett wrote a great blog post for Earthmover that explains the general improvements in Zarr V3 with opening 100TB range datasets (which accounts for much of the time of single point queries) - you can read that [here](https://earthmover.io/blog/xarray-open-zarr-improvements). The obstore store offers further improvements, as shown below. Full details are available in https://github.com/maxrjones/zarr-obstore-performance.

zarr_load_performance_comparison

xarray_query_performance_comparison
xarray_open_performance_comparison

itsgifnotjiff:
Thank you very much for this. I can't wait to see if these kind of performance improvements also apply to pseudo zarrs (zarrs backed by our binary files).

TomNicholas:
@itsgifnotjiff what are "pseudo zarrs"? It is similar to a virtual Zarr?  https://github.com/zarr-developers/VirtualiZarr

itsgifnotjiff:
Yes I am trying to see if I can create Icechunk Arrays and/or Zarr stores for Petabytes of binary format data. 

I work with Environment and Climate Change Canada where we have a wonderful binary format for NWP model outputs and like all the organisations I've talked to we can not abandon it but if we can build on top of it .... (Bit like gribjump from ECMWF or even some slides from Icechunk).

TomNicholas:
> like all the organisations I've talked to we can not abandon it but if we can build on top of it

Yes that's exactly the problem VirtualiZarr was built to solve.

> even some slides from Icechunk

Those slides are referring to VirtualiZarr, which has facility for writing "virtual" zarr chunks into Icechunk (see [virtualizarr docs](https://virtualizarr.readthedocs.io/en/stable/usage.html#writing-to-an-icechunk-store) or [icechunk docs](https://icechunk.io/en/latest/icechunk-python/virtual/)).

> gribjump from ECMWF

Interesting - I hadn't heard of this.

But this issue is closed - @itsgifnotjiff let's continue this discussion on the [VirtualiZarr repo](https://github.com/zarr-developers/VirtualiZarr) - perhaps on [this issue](https://github.com/zarr-developers/VirtualiZarr/issues/218) (or feel free to open a new one).

PR #1621, opened 2023-12-27
Update release.rst

MSanKeys963:
Hi all. 👋🏻 

This PR updates the [release.rst](https://github.com/zarr-developers/zarr-python/blob/main/docs/release.rst) to the latest version. 

Please note that I have not updated it under a new release, i.e. **2.16.2**. If we plan to go with **2.16.2** with current updates, I'm happy to modify the `release.rst` accordingly.

Please review. Thanks!

TODO:
* [x] GitHub Actions have all passed
* [x] Test coverage is 100% (Codecov passes)

PR #1663, opened 2024-02-09
Pyodide support: don't require fasteners on emscripten

hoodmane:
This patch makes fasteners dependencies conditional on not-emscripten and moves
the fasteners import from top level into the function where it is used. It is the one patch
needed for zarr to work in Pyodide/

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,7 +14,7 @@ requires-python = ">=3.9"
 dependencies = [
     'asciitree',
     'numpy>=1.21.1',
-    'fasteners',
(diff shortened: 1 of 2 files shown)


jhamman (approved):

joshmoore:
Merging for 2.17.0 (cc: @MSanKeys963 for updating the notes)

I'll keep checking that things go green in #1621

hoodmane:
Thanks!

MSanKeys963:
> (cc: @MSanKeys963 for updating the notes)

Done.

PR #1670, opened 2024-02-16
Reworked codec pipelines

normanrz:
This PR refactors the codec pipelines in the v3 codebase. There are now a new default implementation: 
- `BatchedCodecPipeline`, which divides the chunk batches into configurable "mini-batches". In a mini-batch all steps are run in lock step (e.g. fetching from store, decoding, encoding, writing to store). Multiple mini-batches are processed concurrently.

This PR moves a lot of code from the Array to the codec pipeline, which is an opportunity to share more code between the `Array` and the `ShardingCodec`. To make that work new `ByteGetter` and `ByteSetter` protocols to generalize the existing `StorePath` are introduced.

It also changes the Codec API by making `decode` and `encode` methods take chunk batches.

TODO:


src/zarr/v3/codecs/batched_pipeline.py
```diff
@@ -0,0 +1,283 @@
...
+
+            chunk_bytes_batch = await concurrent_map(
+                [
+                    (None if is_total_slice(chunk_selection, chunk_spec.shape) else store_path,)
+                    for store_path, chunk_spec, chunk_selection, _ in batch_info
+                ],
+                _read_key,
+                runtime_configuration.concurrency,
+            )
+            chunk_array_batch = await self.decode_batched(
```
akshaysubr, line 223: Does this not require a similar `None` guard like with `_read_key` above?

src/zarr/v3/codecs/batched_pipeline.py
```diff
@@ -0,0 +1,283 @@
...
+                        chunk_spec.shape,
+                        dtype=chunk_spec.dtype,
+                    )
+                    chunk_array.fill(chunk_spec.fill_value)
+                else:
+                    chunk_array = existing_chunk_array.copy()  # make a writable copy
+                chunk_array[chunk_selection] = new_chunk_array_slice
+                return chunk_array
+
+            chunk_array_batch = [
```
akshaysubr, line 250: Any way to parallelize these portions between reading and decoding and subsequent encoding?

src/zarr/v3/metadata.py
```diff
@@ -115,7 +115,7 @@ class ArrayMetadata(Metadata):
     chunk_grid: ChunkGrid
     chunk_key_encoding: ChunkKeyEncoding
     fill_value: Any
-    codecs: CodecPipeline
+    codecs: BatchedCodecPipeline
```
akshaysubr, line 118: Does this mean the use the batched pipeline by default always? That might not be the worst thing, but wondering if it might be useful to implement some mechanism to pass a policy to `parse_codecs` that could allow implementing an `InterleavedCodecPipeline` or something down the line?

akshaysubr:
@normanrz Overall, this looks quite good to me. A couple of questions I had:

1. It makes sense for the default behavior to be just dispatching decoding to multiple threads in `decode_batch` and similarly for encoding. For codecs that want a different parallelization strategy, the idea would be that they just override `decode_batch` right?
2. All of the array types are currently `np.ndarray`. Would it be possible to generalize this to something that is agnostic to the specific library and uses the buffer protocol, `__array_interface__`, `__cuda_array_interface__` or [DLPack](https://dmlc.github.io/dlpack/latest/python_spec.html)? This would allow using numpy arrays as usual, but also cupy arrays, PyTorch tensors, JAX arrays, etc.

src/zarr/v3/metadata.py
normanrz: I definitely want to add a `InterleavedCodecPipeline` as well. We could make that configurable through the runtime configuration.

src/zarr/v3/codecs/batched_pipeline.py
normanrz: We could certainly parallelize that with `concurrent_map`. Not sure how that would be parallelized using the GPU.

normanrz:
I refactored the codec pipeline quite a bit in the last commit.
- There are now a `BatchedCodecPipeline` and a `InterleavedCodecPipeline` building upon an abstract `CodecPipeline`
- There are new protocol types for `ByteGetter` and `ByteSetter`. These are basically a generalization of `StorePath`. This allows the sharding codec to use the same codec pipeline abstraction as the array. That moves us one step closer to treating shards as sub-arrays.
- The `decode`, `decode_partial`, `encode`, `encode_partial` now have the batched interface. 
- This still uses `np.ndarray` everywhere. I leave that to @akshaysubr to change to a generalized array interface. 

Currently, the choice of codec pipeline is hard-coded. I am still looking for a way to specify that. Should that go into `RuntimeConfiguration`? @d-v-b @jhamman

src/zarr/v3/abc/codec.py
```diff
@@ -16,6 +25,44 @@
     from zarr.v3.metadata import ArrayMetadata
     from zarr.v3.config import RuntimeConfiguration
 
+T = TypeVar("T")
+U = TypeVar("U")
+
+
+def noop_for_none(
+    func: Callable[[T, ArraySpec, RuntimeConfiguration], Awaitable[Optional[U]]],
```
d-v-b, line 33: I'm having trouble understanding why this function exists just from its signature -- could you add a docstring that gives some context for it?

d-v-b: I think I'm also struggling to understand the role of `None` in the chunk encoding / decoding process

normanrz: Since this is the batch mode, we're dealing with multiple chunk requests. If the store returns `None` (i.e. missing chunk = fill value-only chunk) for any of these, we don't try to decode these. The `noop_for_none` function is just a higher order function so that the decode (or encode) functions don't have to do the early return on their own.

d-v-b: Do you think we could simplify the logic here if stores didn't return `None`?

normanrz: Not really. In that case, the codec pipeline would have to handle KeyError or similar.

d-v-b: > Not really. In that case, the codec pipeline would have to handle KeyError or similar.

But if handling the key error was in just one place, and after handling keyerror the empty keys were turned into arrays of `fill_value` (or just ignored completely, if there was already a buffer pre-filled with `fill_value`), then I think there would be no need to handle `None` in any of the chunk processing logic, which would be kind of nice

normanrz: I might be wrong but handling a `KeyError` or a `None` is the same complexity in the chunk processing logic. Both still have to happen per-chunk and coordinated with the other chunks in a batch request.

d-v-b: I wonder if missing chunks should never enter the encoding / decoding process at all?

if we handle missing chunks early (e.g., right when we fetch them from the store, for decoding, or when we fetch them from the array, for encoding), then the chunk encoding / decoding routines never need to handle missing chunks, which means that `noop_for_none` wouldn't need to exist, because `chunk` could never be `None`, and likewise for all the other methods that take `Optional[np.ndarray]` as an argument.

normanrz: The codec pipeline does the fetching so that code needs to live there. It also merges chunks into the decoded ndarray, for that it also needs to know which chunks are missing to fill them with fill_value. Using None to propagate that information seems ok.

normanrz:
I think this PR is ready for a final review. I updated the PR description with the major changes. The only thing missing from my pov is the user-configurable batch size and codecpipeline selection. I'll add that after #1855 lands.

src/zarr/abc/codec.py
```diff
@@ -16,6 +25,40 @@
     from zarr.metadata import ArrayMetadata
     from zarr.config import RuntimeConfiguration
 
+T = TypeVar("T")
+U = TypeVar("U")
+
+
+def noop_for_none(
+    func: Callable[[T, ArraySpec, RuntimeConfiguration], Awaitable[Optional[U]]],
```
d-v-b, line 33:
```suggestion
    func: Callable[[T, ArraySpec, RuntimeConfiguration], Awaitable[U | None]],
```

src/zarr/abc/codec.py
```diff
@@ -16,6 +25,40 @@
...
+) -> Callable[[Optional[T], ArraySpec, RuntimeConfiguration], Awaitable[Optional[U]]]:
+    async def wrap(
+        chunk: Optional[T], chunk_spec: ArraySpec, runtime_configuration: RuntimeConfiguration
+    ) -> Optional[U]:
+        if chunk is None:
+            return None
+        return await func(chunk, chunk_spec, runtime_configuration)
+
+    return wrap
```
d-v-b, line 42:
```suggestion
) -> Callable[[T | None, ArraySpec, RuntimeConfiguration], Awaitable[U | None]]:
    async def wrap(
        chunk: T | None, chunk_spec: ArraySpec, runtime_configuration: RuntimeConfiguration
    ) -> U | None:
        if chunk is None:
            return None
        return await func(chunk, chunk_spec, runtime_configuration)

    return wrap
```

src/zarr/abc/codec.py
normanrz: Can we do these changes automated?

src/zarr/abc/codec.py
```diff
@@ -16,6 +25,40 @@
...
+@runtime_checkable
+class ByteSetter(Protocol):
+    async def get(
+        self, byte_range: Optional[Tuple[int, Optional[int]]] = None
+    ) -> Optional[BytesLike]: ...
+
+    async def set(self, value: BytesLike, byte_range: Optional[Tuple[int, int]] = None) -> None: ...
+
+    async def delete(self) -> None: ...
+
```
d-v-b, line 61:
```suggestion
@runtime_checkable
class ByteGetter(Protocol):
    async def get(
        self, byte_range: tuple[int, int | None] | None = None
    ) -> BytesLike | None: ...


@runtime_checkable
class ByteSetter(Protocol):
    async def get(
        self, byte_range: tuple[int, int | None] | None = None
    ) -> BytesLike | None: ...

    async def set(self, value: BytesLike, byte_range: tuple[int, int] | None = None) -> None: ...

    async def delete(self) -> None: ...

```

src/zarr/abc/codec.py
```diff
@@ -44,6 +87,20 @@ async def decode(
     ) -> np.ndarray:
         pass
 
+    async def decode_batch(
+        self,
+        chunk_arrays_and_specs: Iterable[Tuple[Optional[np.ndarray], ArraySpec]],
+        runtime_configuration: RuntimeConfiguration,
+    ) -> Iterable[Optional[np.ndarray]]:
```
d-v-b, line 94:
```suggestion
        chunk_arrays_and_specs: Iterable[tuple[np.ndarray | None, ArraySpec]],
        runtime_configuration: RuntimeConfiguration,
    ) -> Iterable[np.ndarray | None]:
```

src/zarr/abc/codec.py
```diff
@@ -53,17 +110,45 @@ async def encode(
     ) -> Optional[np.ndarray]:
         pass
 
+    async def encode_batch(
+        self,
+        chunk_arrays_and_specs: Iterable[Tuple[Optional[np.ndarray], ArraySpec]],
+        runtime_configuration: RuntimeConfiguration,
+    ) -> Iterable[Optional[np.ndarray]]:
```
d-v-b, line 117:
```suggestion
        chunk_arrays_and_specs: Iterable[tuple[np.ndarray | None, ArraySpec]],
        runtime_configuration: RuntimeConfiguration,
    ) -> Iterable[np.ndarray | None]:
```

src/zarr/abc/codec.py
```diff
@@ -53,17 +110,45 @@ async def encode(
...
         chunk_spec: ArraySpec,
         runtime_configuration: RuntimeConfiguration,
     ) -> np.ndarray:
         pass
 
+    async def decode_batch(
+        self,
+        chunk_bytes_and_specs: Iterable[Tuple[Optional[BytesLike], ArraySpec]],
+        runtime_configuration: RuntimeConfiguration,
+    ) -> Iterable[Optional[np.ndarray]]:
```
d-v-b, line 142:
```suggestion
    async def decode_batch(
        self,
        chunk_bytes_and_specs: Iterable[tuple[BytesLike | None, ArraySpec]],
        runtime_configuration: RuntimeConfiguration,
    ) -> Iterable[np.ndarray | None]:
```

src/zarr/abc/codec.py
```diff
@@ -73,42 +158,98 @@ async def encode(
     ) -> Optional[BytesLike]:
         pass
 
+    async def encode_batch(
+        self,
+        chunk_arrays_and_specs: Iterable[Tuple[Optional[np.ndarray], ArraySpec]],
+        runtime_configuration: RuntimeConfiguration,
+    ) -> Iterable[Optional[BytesLike]]:
```
d-v-b, line 165:
```suggestion
    async def encode_batch(
        self,
        chunk_arrays_and_specs: Iterable[tuple[np.ndarray | None, ArraySpec]],
        runtime_configuration: RuntimeConfiguration,
    ) -> Iterable[BytesLike | None]:
```

src/zarr/abc/codec.py
```diff
@@ -73,42 +158,98 @@ async def encode(
...
         chunk_spec: ArraySpec,
         runtime_configuration: RuntimeConfiguration,
     ) -> Optional[np.ndarray]:
         pass
 
+    async def decode_partial_batch(
+        self,
+        batch_info: Iterable[Tuple[ByteGetter, SliceSelection, ArraySpec]],
+        runtime_configuration: RuntimeConfiguration,
+    ) -> Iterable[Optional[np.ndarray]]:
```
d-v-b, line 191:
```suggestion
    async def decode_partial_batch(
        self,
        batch_info: Iterable[tuple[ByteGetter, SliceSelection, ArraySpec]],
        runtime_configuration: RuntimeConfiguration,
    ) -> Iterable[np.ndarray | None]:
```

src/zarr/abc/codec.py
```diff
@@ -73,42 +158,98 @@ async def encode(
...
         chunk_array: np.ndarray,
         selection: SliceSelection,
         chunk_spec: ArraySpec,
         runtime_configuration: RuntimeConfiguration,
     ) -> None:
         pass
 
+    async def encode_partial_batch(
+        self,
+        batch_info: Iterable[Tuple[ByteSetter, np.ndarray, SliceSelection, ArraySpec]],
```
d-v-b, line 216:
```suggestion
        batch_info: Iterable[tuple[ByteSetter, np.ndarray, SliceSelection, ArraySpec]],
```

src/zarr/abc/codec.py
```diff
@@ -73,42 +158,98 @@ async def encode(
...
         chunk_spec: ArraySpec,
         runtime_configuration: RuntimeConfiguration,
     ) -> BytesLike:
         pass
 
+    async def decode_batch(
+        self,
+        chunk_bytes_and_specs: Iterable[Tuple[Optional[BytesLike], ArraySpec]],
+        runtime_configuration: RuntimeConfiguration,
+    ) -> Iterable[Optional[BytesLike]]:
```
d-v-b, line 243:
```suggestion
        chunk_bytes_and_specs: Iterable[tuple[BytesLike | None, ArraySpec]],
        runtime_configuration: RuntimeConfiguration,
    ) -> Iterable[BytesLike | None]:
```

src/zarr/abc/codec.py
```diff
@@ -117,3 +258,17 @@ async def encode(
         runtime_configuration: RuntimeConfiguration,
     ) -> Optional[BytesLike]:
         pass
+
+    async def encode_batch(
+        self,
+        chunk_bytes_and_specs: Iterable[Tuple[Optional[BytesLike], ArraySpec]],
+        runtime_configuration: RuntimeConfiguration,
+    ) -> Iterable[Optional[BytesLike]]:
```
d-v-b, line 266:
```suggestion
        chunk_bytes_and_specs: Iterable[tuple[BytesLike | None, ArraySpec]],
        runtime_configuration: RuntimeConfiguration,
    ) -> Iterable[BytesLike | None]:
```

src/zarr/codecs/pipeline/batched.py
```diff
@@ -0,0 +1,281 @@
...
+    from zarr.common import ArraySpec, BytesLike, SliceSelection
+
+T = TypeVar("T")
+U = TypeVar("U")
+
+
+def _unzip2(iterable: Iterable[Tuple[T, U]]) -> Tuple[List[T], List[U]]:
+    out0: List[T] = []
+    out1: List[U] = []
+    for item0, item1 in iterable:
```
d-v-b, line 33:
```suggestion
def _unzip2(iterable: Iterable[tuple[T, U]]) -> tuple[list[T], list[U]]:
    out0: list[T] = []
    out1: list[U] = []
    for item0, item1 in iterable:
```

src/zarr/abc/codec.py
normanrz: pylint has https://pylint.readthedocs.io/en/stable/user_guide/messages/refactor/consider-alternative-union-syntax.html

src/zarr/codecs/pipeline/batched.py
```diff
@@ -0,0 +1,281 @@
...
+if TYPE_CHECKING:
+    from typing import List, Optional, Tuple
+    from zarr.metadata import RuntimeConfiguration
+    from zarr.common import ArraySpec, BytesLike, SliceSelection
+
+T = TypeVar("T")
+U = TypeVar("U")
+
+
+def _unzip2(iterable: Iterable[Tuple[T, U]]) -> Tuple[List[T], List[U]]:
```
d-v-b, line 30: what's the difference between this function and zip with tuple unpacking (the last line in my example)?

```python
>>> a = ('a','b','c')
>>> b = (1, 2, 3)
>>> tuple(zip(a,b))
(('a', 1), ('b', 2), ('c', 3))
>>> z = tuple(zip(a,b))
>>> tuple(zip(*z))
(('a', 'b', 'c'), (1, 2, 3))
```

src/zarr/abc/codec.py
d-v-b: i'm not aware of an automated way to do it, but it would be a lot nicer than the manual way. you can also do an IDE-driven style where you remove the `Union` import from typing and find all the errors that result.

src/zarr/codecs/pipeline/hybrid.py
```diff
@@ -0,0 +1,137 @@
...
+    from typing_extensions import Self
+    from zarr.config import RuntimeConfiguration
+    from zarr.common import ArraySpec, BytesLike, SliceSelection
+
+DEFAULT_BATCH_SIZE = 1000
+
+T = TypeVar("T")
+
+
+def batched(iterable: Iterable[T], n: int) -> Iterable[Tuple[T, ...]]:
```
d-v-b, line 28: can we use [`itertools.batched`](https://docs.python.org/3/library/itertools.html#itertools.batched) instead?

normanrz: Unfortunately, no. `itertools.batched` is 3.12+

d-v-b: :(

src/zarr/codecs/pipeline/batched.py
normanrz: Functionally equivalent, but mypy didn't understand the types. I got `Tuple[List[T | U], ...]` as return type.

src/zarr/abc/codec.py
```diff
@@ -44,6 +87,20 @@ async def decode(
     ) -> np.ndarray:
         pass
 
+    async def decode_batch(
+        self,
+        chunk_arrays_and_specs: Iterable[tuple[np.ndarray | None, ArraySpec]],
+        runtime_configuration: RuntimeConfiguration,
+    ) -> Iterable[np.ndarray | None]:
```
d-v-b, line 94: What would be the problem with making `decode` just a special case of `decode_batch`, where the length of `chunk_arrays_and_specs` is 1? 

Also, it seems odd that `decode_batch` with a batch size of 1 could be nullable, while `decode` with the same argument would not be.

normanrz: That would be possible. If we want to make it easier for the community to develop codecs, the batch-style is a bit more complicated. We can provide other means of making that easier, though (helper functions, or `Base*Codec` classes).

d-v-b: to be clear, even if there are performance differences between "do 1" and "do N", I think we should consider hiding that behind a single "do N" api that dispatches to the efficient sub-case as needed.

src/zarr/codecs/gzip.py
```diff
@@ -44,15 +44,15 @@ def from_dict(cls, data: Dict[str, JSON]) -> Self:
     def to_dict(self) -> Dict[str, JSON]:
         return {"name": "gzip", "configuration": {"level": self.level}}
 
-    async def decode(
+    async def decode_single(
```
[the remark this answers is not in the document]
normanrz: That is what the `BytesBytesBatchMixin` does.
Btw, let me know if you know of a better name than `BytesBytesBatchMixin`. Fwiw, they are not real mixins more like base classes.

src/zarr/abc/codec.py
normanrz: Done

src/zarr/codecs/pipeline/batched.py
```diff
@@ -0,0 +1,258 @@
...
+        out1.append(item1)
+    return (out0, out1)
+
+
+def resolve_batched(codec: Codec, chunk_specs: Iterable[ArraySpec]) -> Iterable[ArraySpec]:
+    return [codec.resolve_metadata(chunk_spec) for chunk_spec in chunk_specs]
+
+
+@dataclass(frozen=True)
+class BatchedCodecPipeline(CodecPipeline):
```
d-v-b, line 44: Since `CodecPipeline` defines `read_batch` and `write_batch`, what is the relationship between those methods and the new functionality offered by this class? I think a clarifying docstring for the class might be useful here, because I don't find the inheritance relationship intuitive (and ditto for the `HybridPipeline` -- I have no intuition for that that one is for :) )

src/zarr/codecs/bytes.py
```diff
@@ -25,7 +25,7 @@ class Endian(Enum):
 
 
 @dataclass(frozen=True)
-class BytesCodec(ArrayBytesCodec):
+class BytesCodec(ArrayBytesCodecBatchMixin):
```
d-v-b, line 28: Will `ArrayBytesCodec` be used anywhere other than as the base class fro `ArrayBytesCodecBatchMixin`? If not, we might want to consider simplifying the inheritance structure a bit.

src/zarr/codecs/mixins.py
```diff
@@ -0,0 +1,183 @@
...
+        chunk_spec: ArraySpec,
+    ) -> BytesLike | None:
+        pass
+
+    async def encode(
+        self,
+        chunk_arrays_and_specs: Iterable[tuple[np.ndarray | None, ArraySpec]],
+    ) -> Iterable[BytesLike | None]:
+        return await batching_helper(self.encode_single, chunk_arrays_and_specs)
+
```
d-v-b, line 105: The amount of logic duplicated here really stands out to me. I wonder if there is an abstraction we are missing that could simplify this

src/zarr/codecs/bytes.py
normanrz: I would anticipate that folks want to build a codec that implements their own batching, e.g. in rust or on the GPU. That is why, we should keep both classes.

src/zarr/codecs/pipeline/hybrid.py
```diff
@@ -0,0 +1,115 @@
...
+    if n < 1:
+        raise ValueError("n must be at least one")
+    it = iter(iterable)
+    while batch := tuple(islice(it, n)):
+        yield batch
+
+
+@dataclass(frozen=True)
+class HybridCodecPipeline(CodecPipeline):
```
d-v-b, line 35: What are the operational differences between this class and the `BatchedCodecPipeline`?

src/zarr/codecs/pipeline/hybrid.py
```diff
@@ -0,0 +1,115 @@
...
+    batch_size: int
+    batched_codec_pipeline: BatchedCodecPipeline
```
d-v-b, line 37: Having a sibling class as an attribute is rather surprising. I see that this attribute is used quite extensively in this class, in combination with the `batched` function. This makes me wonder whether we need `HybridCodecPipeline` at all, as opposed to just putting a `batch_size` parameter on `BatchedCodecPipeline`

src/zarr/codecs/pipeline/hybrid.py
normanrz: This is a good question. `BatchedCodecPipeline` executes all steps in lock step for all chunks that are requested. `HybridCodecPipeline` adds a batch size where each batch can run asynchronously interleaved. We could consolidate both into 1 class. That would leave the abstract `CodecPipeline` (for implementing custom codec pipelines) and one implementation.

src/zarr/codecs/mixins.py
normanrz: Can you elaborate? The logic is reused through the `batching_helper`. Large parts of the signature are duplicated but differ in details due to the nature of the 3 separate codec classes. What abstraction would you propose?

madsbk:
I think it would be helpful with some docstrings. At least, I find it hard to follow the intention without any help :)

src/zarr/codecs/pipeline/hybrid.py
normanrz: I got rid of the `HybridCodecPipeline` and added a `batch_size` parameter to the `BatchedCodecPipeline`.

src/zarr/codecs/mixins.py
normanrz: I was able to get rid of a lot of this code by using `typing.Generic`.

normanrz:
I added a few doc strings and implemented the abstract Codec classes with Generics.

src/zarr/codecs/batched_codec_pipeline.py
```diff
@@ -0,0 +1,439 @@
...
+
+
+@dataclass(frozen=True)
+class BatchedCodecPipeline(CodecPipeline):
+    """Default codec pipeline.
+
+    This batched codec pipeline divides the chunk batches into batches of a configurable
+    batch size ("mini-batch"). Fetching, decoding, encoding and storing are performed in
+    lock step for each mini-batch. Multiple mini-batches are processing concurrently.
+    """
```
jhamman, line 62: this can come later but we're going to want some additional documentation on the behavior here. Reading this, I'm not entirely sure I get it.

src/zarr/codecs/batched_codec_pipeline.py
```diff
@@ -0,0 +1,439 @@
...
+                        f"ArrayArrayCodec '{type(codec)}' cannot follow after "
+                        + f"BytesBytesCodec '{type(prev_codec)}'."
+                    )
+            prev_codec = codec
+
+        if any(isinstance(codec, ShardingCodec) for codec in codecs) and len(codecs) > 1:
+            warn(
+                "Combining a `sharding_indexed` codec disables partial reads and "
+                + "writes, which may lead to inefficient performance."
+            )
```
jhamman, line 130:
```suggestion
                + "writes, which may lead to inefficient performance.",
                stacklevel=3
            )
```

src/zarr/codecs/batched_codec_pipeline.py
```diff
@@ -0,0 +1,439 @@
...
+            array_bytes_codec=array_bytes_codec,
+            bytes_bytes_codecs=bytes_bytes_codecs,
+            batch_size=batch_size or config.get("codec_pipeline.batch_size"),
+        )
+
+    @property
+    def supports_partial_decode(self) -> bool:
+        return (len(self.array_array_codecs) + len(self.bytes_bytes_codecs)) == 0 and isinstance(
+            self.array_bytes_codec, ArrayBytesCodecPartialDecodeMixin
+        )
```
jhamman, line 153: let's add a docstring or comment here. The comment should answer why having an array_array_codec or bytes_bytes_codec in the pipeline means `supports_partial_decode is False`.

src/zarr/codecs/batched_codec_pipeline.py
```diff
@@ -0,0 +1,439 @@
...
+
+    @property
+    def supports_partial_encode(self) -> bool:
```
jhamman, line 156: same comment as above.

jhamman (approved):
@normanrz -- this is an impressive piece of work. I should admit that the size of it made it hard to review so I just have a few comments. Overall, I think its a big step forward and I want to get my hands on it so I favor moving it into the v3 branch asap.

d-v-b (approved):

src/zarr/codecs/mixins.py
```diff
@@ -0,0 +1,131 @@
...
+)
+from zarr.common import ArraySpec, BytesLike, SliceSelection, concurrent_map
+from zarr.config import config
+
+
+CodecInput = TypeVar("CodecInput", bound=np.ndarray | BytesLike)
+CodecOutput = TypeVar("CodecOutput", bound=np.ndarray | BytesLike)
+
+
+async def batching_helper(
```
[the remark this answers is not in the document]
d-v-b: This doesn't need to be addressed here, so feel free to ignore for now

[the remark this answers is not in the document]
normanrz: How would you feel about this?
```python
class _Codec(Generic[CodecInput, CodecOutput], Metadata):
    ...

    async def _decode_single(self, chunk_data: CodecOutput, chunk_spec: ArraySpec) -> CodecInput:
        raise NotImplementedError

    async def decode(
        self, chunk_data_and_specs: Iterable[tuple[CodecOutput | None, ArraySpec]]
    ) -> Iterable[CodecInput | None]:
        return await batching_helper(self.decode_single, chunk_data_and_specs)

    # same for encode
    ...
```
Batch-aware codecs would then only override the `decode` method and ignore the `_decode_single` method. That would be fine because `_decode_single` is a protected method not intended to be part of the public interface. Single-chunk codecs could override `_decode_single` and won't have to care about the batching. We could drop the batch mixins, then.

src/zarr/codecs/_v2.py
```diff
@@ -0,0 +1,107 @@
...
+        chunk_array: NDBuffer,
+        chunk_spec: ArraySpec,
+    ) -> NDBuffer | None:
+        chunk_numpy_array = chunk_array.as_numpy_array().ravel(order=chunk_spec.order)
+
+        for filter_metadata in self.filters:
+            filter = numcodecs.get_codec(filter_metadata)
+            chunk_numpy_array = await to_thread(filter.encode, chunk_numpy_array)
+
+        return NDBuffer.from_numpy_array(chunk_numpy_array)
```
normanrz, line 104: @madsbk I added a very naive Buffer integration in these codecs. Could you please help me get that right?

normanrz: I'll merge this PR for now. We can work on a new PR to optimize this.

src/zarr/abc/codec.py
```diff
@@ -1,110 +1,362 @@
...
+        ----------
+        chunk_spec : ArraySpec
+
+        Returns
+        -------
+        ArraySpec
+        """
         return chunk_spec
 
     def evolve(self, array_spec: ArraySpec) -> Self:
```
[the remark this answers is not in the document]
normanrz: What about naming the function `evolve_from_array_spec`?

src/zarr/codecs/mixins.py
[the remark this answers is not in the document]
d-v-b: sounds good!

src/zarr/abc/codec.py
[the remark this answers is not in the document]
d-v-b: i think that works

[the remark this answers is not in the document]
d-v-b: alternatively, we could make `from_array_spec` a class method that takes keyword arguments to cover the attributes that the input array spec doesn't convey

src/zarr/codecs/_v2.py
madsbk: Sounds good!

PR #1703, opened 2024-03-11
Enable ruff/pyupgrade rules (UP) and fix issues

DimitriPapadopoulos:
Keep applying repo-review suggestions (https://github.com/zarr-developers/zarr-python/discussions/1701).

TODO:


zarr/_storage/store.py
```diff
@@ -183,6 +184,9 @@ def rmdir(self, path: str = "") -> None:
         _rmdir_from_keys(self, path)
 
 
+_shadowed_list = list
```
DimitriPapadopoulos, line 187: That's a bit ugly, but I cannot find a better fix.

zarr/_storage/store.py
```diff
@@ -294,8 +298,8 @@ def supports_efficient_get_partial_values(self):
         return False
 
     def get_partial_values(
-        self, key_ranges: Sequence[Tuple[str, Tuple[int, Optional[int]]]]
-    ) -> List[Union[bytes, memoryview, bytearray]]:
+        self, key_ranges: Sequence[tuple[str, tuple[int, Optional[int]]]]
+    ) -> _shadowed_list[Union[bytes, memoryview, bytearray]]:
```
Saransh-cpp, line 302: `Optional` and `Union` can be updated to use the modern syntax (given that Zarr now supports only Python 3.9+). I'm surprised that this wasn't handled by `U` automatically. There must be a setting to do this automatically, saving the manual work.
```suggestion
        self, key_ranges: Sequence[tuple[str, tuple[int, int | None]]]
    ) -> _shadowed_list[bytes | memoryview | bytearray]:
```

This would not be possible everywhere, but it should be possible in the majority of the type annotations (maybe even all of them).

DimitriPapadopoulos: Indeed, ruff is supposed to catch `requires-python = ">=3.9"`.

However, [pyupgrade](https://github.com/asottile/pyupgrade) does not apply such changes - that would be [refurb](https://github.com/dosisod/refurb). I can run try `ruff check --select FURB`, but I'd rather do that in a different pull request, if you agree.

DimitriPapadopoulos: None of these catch such issues:
* `pyupgrade --py39-plus */*.py` / `pyupgrade --py39-plus */*/*.py`
* `ruff check --target-version py39 --select UP`
* `refurb --python-version 3.9 --enable-all .`
* `ruff check --preview --select FURB`

Strange. I was almost certain there are related ruff rules, but I cannot find them.

DimitriPapadopoulos: Got it, it's a pyupgrade rule, [non-pep604-annotation](https://docs.astral.sh/ruff/rules/non-pep604-annotation/) (UP007). However, it is enabled only when targeting Python 3.10 or later.

Saransh-cpp: Ahh, we could use the `isort` rule to propagate the `from __future__ import annotations` import in all the python files. This would allow the `pyupgrade` rules to act on these type annotations while keeping everything backward compatible.

(We should do this, but another PR sounds good too!)

DimitriPapadopoulos: In any case, automating the update process would probably involve two steps
1. `ruff check --target-version py310 --select UP` (to enforce Python 3.10 type annotations)
2. `ruff check --target-version py39 --select FA` (to add `__future__ import annotations` where needed in Python 3.9)

jhamman:
@DimitriPapadopoulos - thanks for moving your focus over to the `v3` branch. I'm going to close this in favor of the contributions you've already made there. 🙏

PR #1705, opened 2024-03-12
Apply assorted Repo-Review suggestions

DimitriPapadopoulos:
Mostly stricter pytest and MyPy settings.

TODO:

### DIFF
--- a/zarr/storage.py
+++ b/zarr/storage.py
@@ -134,11 +134,11 @@ def contains_group(store: StoreLike, path: Path = None, explicit_only=True) -> b
             return True
         # for v3, need to also handle implicit groups
 
-        sfx = _get_metadata_suffix(store)  # type: ignore
+        sfx = _get_metadata_suffix(store)  # type: ignore[arg-type]
         implicit_prefix = key.replace(".group" + sfx, "")
         if not implicit_prefix.endswith("/"):
             implicit_prefix += "/"
-        if store.list_prefix(implicit_prefix):  # type: ignore
+        if store.list_prefix(implicit_prefix):  # type: ignore[union-attr]
             return True
         return False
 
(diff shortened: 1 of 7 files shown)


zarr/storage.py
```diff
@@ -446,7 +446,11 @@ def init_array(
 
     if store_version == 3 and "zarr.json" not in store:
         # initialize with default zarr.json entry level metadata
-        store["zarr.json"] = store._metadata_class.encode_hierarchy_metadata(None)  # type: ignore
+        # fmt: off
```
DimitriPapadopoulos, line 449: That's the only way I have found to ensure `black` doesn't create lines > 100 characters, while not moving `# type: ignore[union-attr]` to an unrelated line (the last line with the parenthesis).

Suggestions welcome.

d-v-b: the ongoing v3 effort will be stripping out a lot of the old v3 logic, so do whatever appeases the linter here, no matter how ugly it is

zarr/storage.py
```diff
@@ -2497,7 +2505,7 @@ def _cache_value(self, key: Path, value):
         value_size = buffer_size(value)
         # check size of the value against max size, as if the value itself exceeds max
         # size then we are never going to cache it
-        if self._max_size is None or value_size <= self._max_size:
+        if self._max_size is None or value_size <= self._max_size:  # type: ignore[redundant-expr]
```
DimitriPapadopoulos, line 2508: `self._max_size` is initialised here:
https://github.com/zarr-developers/zarr-python/blob/f58065b221452acd70235902ad59d920da6fb02f/zarr/storage.py#L2406-L2408
and here:
https://github.com/zarr-developers/zarr-python/blob/f58065b221452acd70235902ad59d920da6fb02f/zarr/storage.py#L2417-L2420

In the first case it shouldn't be `None`, but then typing cannot prevent giving the `max_size` parameter the `None` value. In the second case, no typing is involved.?Not sure why ruff thinks `self._max_size is None` is redundant.

jhamman:
@DimitriPapadopoulos - thanks for moving your focus over to the `v3` branch. I'm going to close this in favor of the contributions you've already made there. 🙏

PR #1712, opened 2024-03-18
Minor improvements to contributing guide

dstansby:
- Highlight code with bash highlighting
- Fix highlighting in markdown code block
- Remove leading "$" in code blocks to make it easier to copy/paste commands
- Add quotes around `[docs]` so that the command works as intended in `zsh`.
- Remove `open` command in doc building section, as this command is only available on macOS.

TODO:


docs/contributing.rst
```diff
@@ -246,9 +245,9 @@ notes (``docs/release.rst``).
 
 The documentation can be built locally by running::
 
-    $ cd docs
-    $ make clean; make html
-    $ open _build/html/index.html
```
MSanKeys963, line 251: How about instead of removing `open _build/html/index.html`, we add [`xdg-open`](https://linux.die.net/man/1/xdg-open) for Linux?

docs/contributing.rst
```diff
@@ -246,9 +245,9 @@ notes (``docs/release.rst``).
...
+    cd docs
+    make clean
+    make html
```
MSanKeys963, line 250:
```suggestion
    make html
    open _build/html/index.html #macos
    xdg-open _build/html/index.html #linux
```

Maybe something like this?

docs/contributing.rst
dstansby: I'd be minded not to include because then there would have to be multiple commands for each OS, and I'm not sure there is even one for Windows?

docs/contributing.rst
dstansby: Again, I it's best to leave out so there's code that works on all operating systems that you can just copy/paste.

jhamman:
@dstansby - would you be willing to take another pass on this w.r.t. our v3 dev process (hatch, etc)?

dstansby:
I think that would be sufficiently different (and I've never used hatch personally) to be worth rebasing this PR, so I'll close

PR #1726, opened 2024-03-27
implement `group.members`

d-v-b:
Implements `Group.children` for v3.

Tests are passing with the `LocalStore`. I still need to mock up a remote store for testing `RemoteStore`, which will be easy to do with the pre-existing v2 tests fixtures (when this is done I will remove the `in progress` label). 

I would appreciate a critical eye on how I [implemented `children`](https://github.com/zarr-developers/zarr-python/compare/v3...d-v-b:group_children_fix?expand=1#diff-88c0f115f2fd8ace004481e7cac73e7603856007810f1cf8ca6cf210b2acebf7R293-R318), because I don't have a lot of experience with async iteration in python and maybe I did something that could be easily improved on.

A few questions, that don't necessarily need answers in this PR:

 - Do we _have_ to call this property `children`? For a variety of reasons I prefer `members`, and that's what I settled on over in the [zom zep](https://github.com/zarr-developers/zeps/pull/46/files#diff-c32ddfc1565df9524bfa5c7024cebc9c077caf76b46872572a2c497e327820a2R54). 
- Should we guarantee a return order for `children`? I think [natural sort order](https://en.wikipedia.org/wiki/Natural_sort_order) on the keys of the subarrays / subgroups is good. This requires that we get all the sub-array / subgroup keys at once, but I think that's how the storage APIs work anyway. 

TODO:

closes https://github.com/zarr-developers/zarr-python/issues/1753


d-v-b:
Seems like `RemoteStore` isn't finished yet, so it is excluded from the set of fixtures used for tests in this PR (currently, `LocalStore` and `MemoryStore`).

```
   def remote_store():
>       return RemoteStore()
E       TypeError: Can't instantiate abstract class RemoteStore with abstract methods get_partial_values, list, list_dir, list_prefix, set_partial_values
```
Implementing `RemoteStore` completely is out of scope for this PR, so I'm removing the `in progress` label.

src/zarr/v3/group.py
```diff
@@ -271,8 +290,32 @@ def __repr__(self):
...
+                "does not support listing, "
+                "specifically via the `list_dir` method. "
+                "This function requires a store that supports listing."
+            )
+
+            raise ValueError(msg)
+        subkeys = await self.store_path.store.list_dir(self.store_path.path)
+        # would be nice to make these special keys accessible programmatically,
+        # and scoped to specific zarr versions
+        subkeys_filtered = filter(lambda v: v not in ("zarr.json", ".zgroup", ".zattrs"), subkeys)
```
jhamman, line 310:
```suggestion
        # and scoped to specific zarr versions
        subkeys_filtered = filter(lambda v: v not in ("zarr.json", ".zgroup", ".zarray"), subkeys)
```

src/zarr/v3/group.py
```diff
@@ -271,8 +290,32 @@ def __repr__(self):
...
+        if not self.store_path.store.supports_listing:
+            msg = (
+                f"The store associated with this group ({type(self.store_path.store)}) "
+                "does not support listing, "
+                "specifically via the `list_dir` method. "
+                "This function requires a store that supports listing."
+            )
+
+            raise ValueError(msg)
+        subkeys = await self.store_path.store.list_dir(self.store_path.path)
```
jhamman, line 307: Following https://github.com/zarr-developers/zarr-python/discussions/1686#discussioncomment-8620826, this method should also return an async generator. Just calling that out as a change that will come here soon, what you've done is fine for now.

What I would love to see is something like:

```python
async for subkey in self.store_path.store.list_regex(self.store_path.path, pattern=pattern):
    yield await self.getitem(subkey)
```

src/zarr/v3/group.py
```diff
@@ -271,8 +290,32 @@ def __repr__(self):
...
+        # and scoped to specific zarr versions
+        subkeys_filtered = filter(lambda v: v not in ("zarr.json", ".zgroup", ".zattrs"), subkeys)
+        # is there a better way to schedule this?
+        for subkey in subkeys_filtered:
+            try:
+                yield await self.getitem(subkey)
+            except KeyError:
+                # keyerror is raised when `subkey``names an object in the store
+                # in which case `subkey` cannot be the name of a sub-array or sub-group.
+                pass
```
jhamman, line 318: I might be missing something here but if `subkey` was just found in the store, why would we be getting a `KeyError`? This `pass` seems like it could silently ignore something that we care about.

tests/conftest.py
```diff
@@ -1,8 +1,76 @@
...
+from zarr.v3.store.remote import RemoteStore
 
 
 @pytest.fixture(params=[str, pathlib.Path])
 def path_type(request):
     return request.param
+
+
+@pytest.fixture()
+def mock_s3(request):
```
jhamman, line 15: as you said in another comment, let's move the remote store setup / testing to another PR.

src/zarr/v3/group.py
```diff
@@ -377,7 +420,7 @@ def __setitem__(self, key, value):
         """__setitem__ is not supported in v3"""
         raise NotImplementedError
 
-    async def update_attributes_async(self, new_attributes: Dict[str, Any]) -> Group:
+    async def update_attributes_async(self, new_attributes: dict[str, Any]) -> Group:
```
jhamman, line 423: I'm a bit surprised that these `_async` methods are still on the Group class. Did I miss these in #1590 / #1634? 

In any event, not your problem but something I should revisit.

jhamman (commented):
Thanks for taking this on @d-v-b. Left you some open-ended questions but this seems like its off to a good start.

src/zarr/v3/group.py
d-v-b: If `"foo"` is the name of a random non-zarr-related file in the store, `group.getitem("foo")` raises `KeyError`, because `"foo"` is not the name of a sub-array or sub-group. So in this case, a `KeyError` just means there's a random file in there, that we want to ignore. I'm not sure that there are sources of `KeyError` (or any other exception) that we would want to handle here?

tests/conftest.py
d-v-b: removed in b28eaee2f328e8e9ee2aed9f51d14edf9d49630b

src/zarr/v3/group.py
d-v-b: when will `.zarray` be an immediate subkey of a group? By contrast, we do expect to find `.zattrs` under the prefix of any v2 group with attributes.

d-v-b:
> Should we guarantee a return order for children? I think [natural sort order](https://en.wikipedia.org/wiki/Natural_sort_order) on the keys of the subarrays / subgroups is good. This requires that we get all the sub-array / subgroup keys at once, but I think that's how the storage APIs work anyway.

I think this is answered (in the negative) by the need to support asynchronous iteration over keys, and the fact that cloud backends (e.g., [s3](https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectsV2.html)) will reasonably paginate requests for lists of keys.

d-v-b:
the name of this method was discussed in an offline discussion in the biweekly refactoring meeting, and the conclusion was that we will switch the name from `children` to `members`. The v2 codebase uses both terms internally, but only the term `members` is used in the [`Group` API](https://github.com/zarr-developers/zarr-python/blob/6105ef203e3e5c5390aa01db613bc42e8fb98c1a/zarr/hierarchy.py). The terms "member" or "members" appears 13 times there, while "children" or "child" are not used.

src/zarr/v3/sync.py
```diff
@@ -113,7 +113,7 @@ def _sync(self, coroutine: Coroutine[Any, Any, T]) -> T:
     def _sync_iter(self, coroutine: Coroutine[Any, Any, AsyncIterator[T]]) -> List[T]:
         async def iter_to_list() -> List[T]:
             # TODO: replace with generators so we don't materialize the entire iterator at once
-            async_iterator = await coroutine
-            return [item async for item in async_iterator]
+            # async_iterator = await coroutine
+            return [item async for item in coroutine()]
```
d-v-b, line 117: @DahnJ I had to revert some of your changes from #1692 here in order to get the code to run. The type signatures are currently incorrect, but the code works, so we should either adjust the annotations to match the actual types, or adjust the implementation (and the actual types) to match the annotation.

@jhamman, I know you have some ideas about the direction for this

d-v-b:
a design question: in the latest changes to this PR, `Group.members` returns a tuple of `(str, Group | Array)` pairs. This is a departure from returning `dict[str, Group | Array]`, but it facilitates streaming individual members instead of prematurely collecting them. I am wondering if "members returns a dict" vs "members returns a tuple of tuples" is a false dichotomy -- can we just do both, by making `Group.members` a `MutableMapping`? (this relates to the discussion over in #1787)

src/zarr/group.py
```diff
@@ -269,26 +287,68 @@ async def update_attributes(self, new_attributes: Dict[str, Any]):
...
 
-    async def array_keys(self) -> AsyncIterator[str]:
+    async def contains(self, member: str) -> bool:
         raise NotImplementedError
 
+    # todo: decide if this method should be separate from `groups`
+    async def group_keys(self) -> AsyncGenerator[str, None]:
+        async for key, value in self.members():
+            if isinstance(value, AsyncGroup):
+                yield key
```
jhamman, line 333: A **future** optimization here would be to avoid constructing the `AsyncGroup` and instead just generate the key based on filtering the metadata doc. As it is currently written, we construct all the members as `AsyncGroup | AsyncArray`, only to return their name.

jhamman: Additionally, we could allow `Group.members` to filter to `only='groups'` so that we never need construct arrays in these cases.

src/zarr/group.py
```diff
@@ -401,50 +465,54 @@ def attrs(self) -> Attributes:
...
 
     @property
-    def children(self) -> List[Union[Array, Group]]:
-        raise NotImplementedError
-        # Uncomment with AsyncGroup implements this method
-        # _children: List[Union[AsyncArray, AsyncGroup]] = self._sync_iter(
-        #     self._async_group.children()
-        # )
-        # return [Array(obj) if isinstance(obj, AsyncArray) else Group(obj) for obj in _children]
+    def members(self) -> tuple[tuple[str, Array | Group], ...]:
```
jhamman, line 477: Let's continue to think about the return type here. I could see a (frozen) dict being a more ergonomic return type.

jhamman (approved):
Let's get this in! It unlocks a bunch of work on the hierarchy. I left a few comments around future optimizations / directions but this is a good first step.

src/zarr/group.py
d-v-b: I'm also a fan of something immutable and dict-like here. I think we can use  #1787 to continue brainstorming

src/zarr/group.py
d-v-b: yes, i think checking the metadata should be much cheaper than instantiating `AsyncArray` / `AsyncGroup`. I will have to think about what those changes would mean for the `AsyncGroup.members` API, since that's what `group_keys` relies on.

PR #1734, opened 2024-03-29
Array & Group: Use already loaded attributes to populate cache.

dcherian:
[Description of PR]

Spotted this quick (possible) optimization.

`_load_metadata` (two lines above) has already downloaded the attributes. May as well populate the cache ahead of the time if `cache_attrs is True` (which is the default)

TODO:

### DIFF
--- a/zarr/core.py
+++ b/zarr/core.py
@@ -174,6 +174,8 @@ class Array:
         self._attrs = Attributes(
             store, key=akey, read_only=read_only, synchronizer=synchronizer, cache=cache_attrs
         )
+        if cache_attrs:
+            self._attrs._cached_asdict = self._meta["attributes"]
 
         # initialize info reporter
         self._info_reporter = InfoReporter(self)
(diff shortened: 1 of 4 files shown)


zarr/core.py
```diff
@@ -174,6 +174,8 @@ def __init__(
         self._attrs = Attributes(
             store, key=akey, read_only=read_only, synchronizer=synchronizer, cache=cache_attrs
         )
+        if cache_attrs:
+            self._attrs._cached_asdict = self._meta["attributes"]
```
jhamman, line 178: It seems like this optimization really belongs in the `Attributes.__init__` method. What do you think about passing the attributes through like this?

```suggestion
        self._attrs = Attributes(
            store, key=akey, read_only=read_only, synchronizer=synchronizer, cache=cache_attrs,
            cache_init=self._meta["attributes"]
        )
        if cache_attrs:
            self._attrs._cached_asdict = self._meta["attributes"]
```

dcherian: Would be fine, wasn't sure if you wanted to preserve the signature for some reason. Also there is some error in the tests, presumably v2 vs v3?

dcherian: It's there for arrays on V3 I think.

rabernat (approved):

jhamman (approved):

PR #1742, opened 2024-04-04
Remove old v3

d-v-b:
The plan here is to remove the old implementation of Zarr v3 (the old version 3). I think fully removing the old v3 stuff would be a minor accelerator for the new V3 effort because it removes the possibility of confusing the old v3 with the new v3 in the codebase, and it reduces the LOC of the codebase which makes the old code easier to read / test.

At the moment, I have just slathered deprecation warnings atop the old V3 routines, which lets my IDE (vscode) visually indicate which stuff is using those old routines. If we feel OK about this effort, I would go ahead and prune the old v3 classes entirely, and remove the WIP tag from this PR.

closes https://github.com/zarr-developers/zarr-python/issues/1771


jhamman:
@d-v-b - do you think this code will ever be released? Since your target is the v3 branch, I think its reasonable to just remove these sections now.

d-v-b:
This is ready for review. 

To summarize the changes, because it's a lot of LOC:
- Removed `.zarr_version` attributes from `Group` and `Array` classes, and the constructors for those classes, and the functions that wrapped the constructors for those classes. As of this PR, `Array` and `Group` are V2-only.
- Removed all legacy V3 code + tests, and anything that was conditional on the zarr version (there was a lot of this in various places in the codebase)
- Modified relevant tests to no longer parametrize over zarr versions, (and longer reference zarr version at all)
- V2 Stores still have an `_version` attribute, which is the literal 2. We can get rid of that now or later if we want.

jhamman (approved):
I did not review every single line here but this seems safe to move forward. Much of the remaining legacy code is likely to be retired as well but this certainly helps clean things up in the meantime.

normanrz (approved):
Let's get this merged :+1:

PR #1743, opened 2024-04-04
groundwork for V3 group tests

d-v-b:
Adds tests for the Group API in v3. In progress. I will remove the in progress label when it's ready for a final review, but I encourage people with an interest in the testing infrastructure to look at the early changes and weigh in if you see anything you don't like. It will be easier to course-correct before a lot of code is written.

To outline the strategy I'm using:
- no classes (or at least, no inheritance)
- parameterize / fixturize as much as possible
- every method of the Group / AsyncGroup classes gets a test, that should parametrize over the arguments of that method

This is a deliberate departure from the strategy used in the v2 tests, where test classes are defined that inherit from a base test class, potentially overriding methods. I'm pretty sure we can use pytest fixtures to express the same functionality without needing the cognitive overhead of inheritance, but this is my personal judgment, and I'm open to other ideas here. The goal is just to have a test suite that's easy for developers to understand, which (in my experience) is not a criterion the v2 tests always meet.

<del>I don't think I will fix failing tests here. That should probably done in separate PRs, once we agree on the shape of the testing infrastructure.</del> edit: amendment to this: since test failures do slow down the creation of new tests, I'm going to apply some fixes to obvious bugs revealed in the process of adding tests. These fixes should be separable in the commit history, and on the basis of the files they affect.

depends on #1726 

TODO:


d-v-b:
Some of my group tests are running into bugs from the storage side. So I'm broadening the scope of this effort to include tests for the stores.

src/zarr/v3/group.py
```diff
@@ -150,15 +158,25 @@ def from_dict(
...
-    ) -> Union[AsyncArray, AsyncGroup]:
+    ) -> AsyncArray | AsyncGroup:
 
         store_path = self.store_path / key
 
+        # Note:
+        # in zarr-python v2, we first check if `key` references an Array, else if `key` references
+        # a group,using standalone `contains_array` and `contains_group` functions. These functions
+        # are reusable, but for v3 they would perform redundant I/O operations.
+        # Not clear how much of that strategy we want to keep here.
```
jhamman, line 169: I don't think we want to keep it at all! 

xref: https://github.com/zarr-developers/zarr-python/pull/1739

src/zarr/v3/group.py
```diff
@@ -150,15 +158,25 @@ def from_dict(
...
+
+        # if `key` names an object in storage, it cannot be an array or group
+        if await store_path.exists():
+            raise KeyError(key)
```
jhamman, line 173: I don't think we should include this check. This adds a roundtrip to the store to protect against a situation I have a hard time imaging being a real issue.


For one, most storage backends forbid such a situation but I don't think the spec really cares.

```
/store/a/b/c
/store/a/b/c/zarr.json
```

src/zarr/v3/group.py
```diff
@@ -271,8 +289,32 @@ def __repr__(self):
...
+                f"The store associated with this group ({type(self.store_path.store)}) "
+                "does not support listing, "
+                "specifically via the `list_dir` method. "
+                "This function requires a store that supports listing."
+            )
+
+            raise ValueError(msg)
+        subkeys = await self.store_path.store.list_dir(self.store_path.path)
+        # would be nice to make these special keys accessible programmatically,
+        # and scoped to specific zarr versions
```
jhamman, line 308: please rework this comment to explain what the subkeys_filtered variable is.

jhamman: You suggestion to move the tuple below is a good one though.

src/zarr/v3/group.py
```diff
@@ -271,8 +289,32 @@ def __repr__(self):
...
+        subkeys_filtered = filter(lambda v: v not in ("zarr.json", ".zgroup", ".zattrs"), subkeys)
```
jhamman, line 309: I'm still confused here. Don't we want to be looking for `.zarray` to find v2 arrays?

src/zarr/v3/group.py
```diff
@@ -271,8 +289,32 @@ def __repr__(self):
...
+        if not self.store_path.store.supports_listing:
+            msg = (
+                f"The store associated with this group ({type(self.store_path.store)}) "
+                "does not support listing, "
+                "specifically via the `list_dir` method. "
+                "This function requires a store that supports listing."
+            )
+
+            raise ValueError(msg)
+        subkeys = await self.store_path.store.list_dir(self.store_path.path)
```
jhamman, line 306: I think the ideal async way to do this would be for `list_dir` to return an `AsyncGenerator`. Then your for loop below would look something like:

```python
async for key in subkeys:
    # filter
    if key in ("zarr.json", ".zgroup", ".zattrs"):
         ...
         yield await self.getitem(subkey)
         
     ...
```

jhamman (commented):
Didn't get to reviewing the tests yet but this is looking good.

src/zarr/v3/group.py
d-v-b: if `foo/bar/` is the prefix for a group, then we might expect to find either v2 group stuff (`foo/bar/.zattrs`, `foo/bar/.zgroup`) or v3 group stuff (`foo/bar/zarr.json`), but `foo/bar/.zarray` would be unexpected, since we know `foo/bar` is a group, and the same path can't correspond to both a group and an array.

src/zarr/v3/group.py
d-v-b: I opened https://github.com/zarr-developers/zarr-specs/pull/291 to discuss removing implicit groups from the spec. the PR is at https://github.com/zarr-developers/zarr-specs/pull/292

d-v-b:
Because of the intersection of the store design and implicit groups, we are a bit squeezed when it comes to performance / ergonomics for discovery routines like `Group.open()`.

Consider this code:
```python
# create a group
g = Group.open("group_path") 
g.open("X")
```
there are three possibilities to handle:
- `group_path/X/zarr.json` exists. Then we attempt to parse it, and produce a group or array.
- `group_path/X` is the key of an object in the store. Then `group_path/X/zarr.json` cannot exist, and so we should raise a `KeyError`
- The key `group_path/X/zarr.json` does not exist. This has two sub-cases.
-   - `group_path/X/` is a prefix for the key of some Zarr metadata object, like `group_path/X/Y/zarr.json`. Then `group_path/X` should be treated as the key for an implicit group according to the spec.
-  - `group_path/X/` is not a prefix for any key in in the store, (e.g., because `X` is just a random string a user typed in by accident). Numerically, this is the most abundant situation (most stores only contain a small subset of all possible keys). The only way we can test for this is by calling `store.list_dir` and checking whether `X` appears in the list of prefixes returned by `list_dir`. I don't think calling `list_dir` for every `Group.open()` call is a good idea, but that's our only option with the current v3 spec and the current store design.

Please someone correct me if I'm wrong here, but it looks like if we support implicit groups, and we don't want `Group.open(blabalabla)` to nearly always succeed, then we have to accept terrible performance (due to the need to run a directory listing). This entails that stores that don't support directory listing will appear to contain an infinite number of implicit groups. Not great. We should really remove implicit groups from the spec.

src/zarr/group.py
```diff
@@ -42,6 +42,26 @@ def parse_attributes(data: Any) -> dict[str, Any]:
...
 
+@overload
+def _parse_async_node(node: AsyncArray) -> Array: ...
+
+
+@overload
+def _parse_async_node(node: AsyncGroup) -> Group: ...
+
+
+def _parse_async_node(node: AsyncArray | AsyncGroup) -> Array | Group:
```
d-v-b, line 53: this exists as something that can be used with `map` over an iterable of `AsyncArray` / `AsyncGroup` to transform each one to its synchronous counterpart. Previously we used a tuple around a generator expression, but mypy couldn't type check this correctly.

src/zarr/group.py
```diff
@@ -494,19 +513,13 @@ def nmembers(self) -> int:
...
-        ret: list[tuple[str, Array | Group]] = []
-        for key, value in _members:
-            if isinstance(value, AsyncArray):
-                ret.append((key, Array(value)))
-            else:
-                ret.append((key, Group(value)))
-        return tuple(ret)
+        _members = self._sync_iter(self._async_group.members())
+
+        result = tuple(map(lambda kv: (kv[0], _parse_async_node(kv[1])), _members))
```
d-v-b, line 521: see https://github.com/zarr-developers/zarr-python/pull/1743/files#r1596469236

d-v-b:
I think this is ready to go

src/zarr/array.py
```diff
@@ -100,31 +102,33 @@ async def create(
                 fill_value = False
             else:
                 fill_value = 0
+        if zarr_format == 3:
```
normanrz, line 105: Ok. This will need some careful merging with #1857

tests/v2/conftest.py
```diff
@@ -1,5 +1,5 @@
-import pathlib
 import pytest
+import pathlib
```
normanrz, line 2: Should we use isort or something else automated?

normanrz (approved):

tests/v2/conftest.py
d-v-b: yes, I would love this. i think ruff format takes a parameter that does what we want. I will add that to this PR unless there any objections.

d-v-b: instead of adding isort-style linting to this PR, it looks like we are going to do that in a separate effort.

src/zarr/group.py
```diff
@@ -248,16 +263,42 @@ def attrs(self):
...
         return await type(self).create(
             self.store_path / path,
-            **kwargs,
+            attributes=attributes,
+            exists_ok=exists_ok,
+            zarr_format=self.metadata.zarr_format,
         )
 
-    async def create_array(self, path: str, **kwargs) -> AsyncArray:
+    async def create_array(
```
d-v-b, line 276: `group.create_group` and `group.create_array` should preserve the zarr format of the `group`, which means we can't accept `**kwargs` in these methods without checking the contents of **kwargs for the absence of `zarr_format`. Since I'm not a huge fan of `**kwargs` anyway, I just elevated the underlying function signatures to this method, sans `zarr_format`, which is set automatically based on the zarr format of the `group`.

normanrz: We need to redo that in #1857

src/zarr/fixture/.zgroup
```diff
@@ -0,0 +1,3 @@
+{
```
jhamman, line 1: I believe this was added in error

tests/v3/test_storage.py
```diff
@@ -1,10 +1,802 @@
+# import array
```
jhamman, line 1: maybe a bad merge conflict resolution here?

src/zarr/group.py
```diff
@@ -41,6 +39,26 @@ def parse_attributes(data: Any) -> dict[str, Any]:
...
+def _parse_async_node(node: AsyncArray | AsyncGroup) -> Array | Group:
+    """
+    Wrap an AsyncArray in an Array, or an AsyncGroup in a Group.
+    """
+    if isinstance(node, AsyncArray):
+        return Array(node)
+    elif isinstance(node, AsyncGroup):
+        return Group(node)
+    else:
+        assert False
```
jhamman, line 59: we could consider adding a `to_sync()` method to `AsyncGroup` and `AsyncArray` to avoid the instance checks.

jhamman (approved):
Looking great @d-v-b. A few small comments for you.

src/zarr/fixture/.zgroup
d-v-b: ah right, this is the zombie fixture directory that gets created by tests, see https://github.com/zarr-developers/zarr-python/pull/1813

src/zarr/group.py
d-v-b: yeah, that could be convenient. I will keep this in mind if these `isinstance` checks start causing undue friction

d-v-b:
one large, but minor (or petty) change in the latest version of this effort: I renamed `test_storage.py` to `test_store.py`, in order to be consistent with the naming: we have `zarr.store`, no `zarr.storage`.

jhamman (approved):

PR #1746, opened 2024-04-05
Add example hypothesis roundtrip test

dcherian:
After `mamba install hypothesis`

Use
```
python -m pytest --capture=no --hypothesis-verbosity=verbose test_properties.py
```

to see all the things it tries.

This was a quick attempt at a property test.

The other thing you can do is a "Stateful" test (e.g. https://github.com/pydata/xarray/blob/main/properties/test_index_manipulation.py)

which runs an arbitrary sequence of manipulations (e.g. add array, rename, move, delete, modify array, copy), and checks for consistency at each point.

TODO:


tests/test_properties.py
```diff
@@ -0,0 +1,109 @@
...
+        array_path = path
+        array_name = None
+    elif path == "/":
+        assert name is not None
+        array_path = name
+        array_name = "/" + name
+    else:
+        assert name is not None
+        array_path = f"{path}/{name}"
+        array_name = "/" + array_path
```
dcherian, line 84: totally not obvious!

tests/test_properties.py
```diff
@@ -0,0 +1,109 @@
...
+
+    expected_attrs = {} if attrs is None else attrs
+
+    init_array(store, shape=nparray.shape, chunks=chunks, path=array_path, dtype=nparray.dtype.str)
+    a = Array(store, path=array_path, zarr_version=3)
+    if attrs is not None:
+        a.attrs.put(attrs)
+
+    assert isinstance(a, Array)
```
Zac-HD, line 93: I'd probably move all of this into a `zarr_arrays()` function, which returns an arbitrary Zarr array.

The general design pattern is that the strategy should generate arbitrarily weird and nasty instances by default, and then you can provide various arguments to constrain it (e.g., the arguments you've passed to `@given()` here).  For example, why constrain names to Latin Extended-A?  If that's part of the Zarr spec it'd be nice to have a comment, if not we should check for bugs in the rest of utf-8 at least (if not all of Unicode).

dcherian: Ah yes, I agree.

I'm a bit confused with how to do that since Arrays need to be backed by stores, but I can think about it.

Good point about the unicode thing, that was an Xarray restriction that is not needed here.

Zac-HD: `zarr_arrays(..., store: st.SearchStrategy[store] = ...)`?  You can pass a specific store with `st.just()`, and access a generated store via the `arr.store` attribute.

dcherian:
Some input needed:
1. Should this be a separate action or folded in to the existing actions? I lean former. It's fast now (47s) but it could get slower in the future.
2. There's more to make `strategies.py` actually useful as public API but I'd like to punt that to the future.
3. I'll make comments where I'd like a careful review.

src/zarr/strategies.py
```diff
@@ -0,0 +1,139 @@
...
+_attr_keys = st.text(st.characters(), min_size=1)
+_attr_values = st.recursive(
+    st.none() | st.booleans() | st.text(st.characters(), max_size=5),
+    lambda children: st.lists(children) | st.dictionaries(_attr_keys, children),
+    max_leaves=3,
+)
+
+# No '/' in array names?
+# No '.' in paths?
+zarr_key_chars = st.sampled_from("-0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ_abcdefghijklmnopqrstuvwxyz")
```
dcherian, line 22: is this right?

src/zarr/strategies.py
```diff
@@ -0,0 +1,139 @@
...
+
+# The following should be public strategies
+attrs = st.none() | st.dictionaries(_attr_keys, _attr_values)
+paths = st.none() | st.text(zarr_key_chars, min_size=1) | st.just("/")
+array_names = st.text(zarr_key_chars | st.just("."), min_size=1).filter(
+    lambda t: not t.startswith((".", ".."))
```
dcherian, line 28: Array names acannot start with `.` or `..` is that right?

src/zarr/strategies.py
```diff
@@ -0,0 +1,139 @@
...
+)
+np_arrays = npst.arrays(
+    # FIXME: re-enable timedeltas once we figure out the fill_value issue.
+    dtype=npst.scalar_dtypes().filter(lambda x: x.kind != "m"),
+    shape=npst.array_shapes(max_dims=4),
+)
+stores = st.builds(MemoryStore, st.just({}), mode=st.just("w"))
+compressors = st.sampled_from([None, "default"])
```
dcherian, line 36: these could be expanded in the future

src/zarr/strategies.py
```diff
@@ -0,0 +1,139 @@
...
+) -> Array:
+    store = draw(stores)
+    nparray, chunks = draw(np_array_and_chunks(arrays=arrays))
+    path = draw(paths)
+    name = draw(array_names)
+    attributes = draw(attrs)
+    # compressor = draw(compressors)
+
+    # TODO: clean this up
+    if path is None and name is None:
```
dcherian, line 75: Can this mess be cleaned up?

src/zarr/strategies.py
```diff
@@ -0,0 +1,139 @@
...
+    )
+
+    assert isinstance(a, Array)
+    assert nparray.shape == a.shape
+    # assert chunks == a.chunks  # TODO: adapt for v2, v3
+    assert array_path == a.path
+    assert array_name == a.name
+    # assert a.basename is None  # TODO
+    # assert a.store == normalize_store_arg(store)
+    assert dict(a.attrs) == expected_attrs
```
dcherian, line 114: Would be good to decide what invariants to check here. Potentially could move out to its own test.

src/zarr/strategies.py
```diff
@@ -0,0 +1,139 @@
...
+
+# The following should be public strategies
+attrs = st.none() | st.dictionaries(_attr_keys, _attr_values)
+paths = st.none() | st.text(zarr_key_chars, min_size=1) | st.just("/")
+array_names = st.text(zarr_key_chars | st.just("."), min_size=1).filter(
+    lambda t: not t.startswith((".", ".."))
+)
+np_arrays = npst.arrays(
+    # FIXME: re-enable timedeltas once we figure out the fill_value issue.
+    dtype=npst.scalar_dtypes().filter(lambda x: x.kind != "m"),
```
dcherian, line 32: There's an error with creating timedelta arrays. Are those supported?

src/zarr/strategies.py
```diff
@@ -0,0 +1,139 @@
...
+    root = Group.create(store)
+    a = root.create_array(
+        array_path,
+        shape=nparray.shape,
+        chunks=chunks,
+        dtype=nparray.dtype.str,
+        attributes=attributes,
+        # compressor=compressor,  # TODO: FIXME
+        # TODO: FIXME seems to break with booleans and timedelta
+        # fill_value=nparray.dtype.type(0),
```
dcherian, line 104: This seems to be a bug.

dcherian:
Some input here would be very helpful

jhamman (commented):
Long overdue. Thanks @dcherian for working on this. Once this is in, I'd love some guidance on writing a similar thing for the store tests.

src/zarr/strategies.py
dcherian: Fixed by https://github.com/zarr-developers/zarr-python/pull/2020!

src/zarr/strategies.py
```diff
@@ -0,0 +1,147 @@
...
+    max_leaves=3,
+)
+
+# From https://zarr-specs.readthedocs.io/en/latest/v3/core/v3.0.html#node-names
+# 1. must not be the empty string ("")
+# 2. must not be the empty string ("")
+# 3. must not include the character "/"
+# 4. must not be a string composed only of period characters, e.g. "." or ".."
+# 5. must not start with the reserved prefix "__"
+zarr_key_chars = st.sampled_from(
```
dcherian, line 26: TODO: Add `_`.

dcherian:
This looks like a real failure:

```
 E           AssertionError: 
E           Arrays are not equal
E           
E           Mismatched elements: 1 / 1 (100%)
E           Max absolute difference: 1
E           Max relative difference: inf
E            x: array([1], dtype=int8)
E            y: array([0], dtype=int8)
```

Apparently it is not roundtripping `np.array([1], dtype=np.int8)` @d-v-b are you able to take a look?

d-v-b:
> This looks like a real failure:
> 
> ```
>  E           AssertionError: 
> E           Arrays are not equal
> E           
> E           Mismatched elements: 1 / 1 (100%)
> E           Max absolute difference: 1
> E           Max relative difference: inf
> E            x: array([1], dtype=int8)
> E            y: array([0], dtype=int8)
> ```
> 
> Apparently it is not roundtripping `np.array([1], dtype=np.int8)` @d-v-b are you able to take a look?

definitely! Given that I know nothing about hypothesis, how can I replicate this failure locally?

dcherian:
This may be user error. It's sensitive to the `store.close()`

```python
import zarr
from zarr.array import Array
from zarr.group import Group
from zarr.store import MemoryStore
import numpy as np

store = MemoryStore(mode="w")
root = Group.create(store)
nparray = np.array([1], dtype=np.int8)
    
a = root.create_array(
    "/0/0",
    shape=nparray.shape,
    chunks=(1,),
    dtype=nparray.dtype.str,
    attributes={},
    # compressor=compressor,  # TODO: FIXME
    fill_value=nparray.dtype.type(0),
)
a[:] = nparray
print(a[:]) # [1]
store.close()
print(a[:]) # [0]
```

d-v-b:
I can replicate this locally. `store.close` is [quite minimal](https://github.com/zarr-developers/zarr-python/blob/0e186719ed1750f3ba43ac18cb95c15305ab97e5/src/zarr/abc/store.py#L219) so I think our problem lies elsewhere. I will investigate.

src/zarr/strategies.py
```diff
@@ -0,0 +1,147 @@
...
+    # assert array_path == a.name, (path, name, array_path, a.name, a.path)
+    # assert a.basename is None  # TODO
+    # assert a.store == normalize_store_arg(store)
+    assert dict(a.attrs) == expected_attrs
+
+    a[:] = nparray
+
+    store.close()
+
+    return a
```
dcherian, line 128: @d-v-b how do I fix this now? Reopen a new array with the store in read-only mode and return that?

d-v-b: do you need to call `store.close` in this test? I don't think we have any stores that actually do anything in their `close` method. I just checked and none of the stores have even implemented `close`, so it really does nothing at this point.

d-v-b: to expand a bit, here is the abstract base class `close` method; none of the concrete classes have overridden it: https://github.com/zarr-developers/zarr-python/blob/064b2e09b188b8e09536fb327d2c0f5fe8492f2b/src/zarr/abc/store.py#L219-L222

src/zarr/strategies.py
```diff
@@ -0,0 +1,147 @@
...
+    assert chunks == a.chunks
+    assert array_path == a.path, (path, name, array_path, a.name, a.path)
+    # assert array_path == a.name, (path, name, array_path, a.name, a.path)
+    # assert a.basename is None  # TODO
+    # assert a.store == normalize_store_arg(store)
+    assert dict(a.attrs) == expected_attrs
+
+    a[:] = nparray
+
+    store.close()
```
dcherian, line 126:
```suggestion
```

src/zarr/strategies.py
```diff
@@ -0,0 +1,146 @@
...
+    max_leaves=3,
+)
+
+# From https://zarr-specs.readthedocs.io/en/latest/v3/core/v3.0.html#node-names
+# 1. must not be the empty string ("")
+# 2. must not include the character "/"
+# 3. must not be a string composed only of period characters, e.g. "." or ".."
+# 4. must not start with the reserved prefix "__"
+zarr_key_chars = st.sampled_from(
+    "_.-0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ_abcdefghijklmnopqrstuvwxyz"
```
dcherian, line 26:
```suggestion
    ".-0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ_abcdefghijklmnopqrstuvwxyz"
```

dcherian:
pre-commit.ci autofix

Issue #1757, opened 2024-04-05
[v3] remote store support (s3, gcs, azure, http)


jhamman:
The v3 branch currently has an initial (but broken) implementation of a remote store. This issue tracks getting that operational so we car start using v3 against s3, gcs, azure, http, etc.

TODOs:

- [ ] rename to `FsspecStore`
- [ ] limit store to only fsspec implementation that are async-friendly
- [ ] decide on approach to testing

@martindurant - any chance you can lend a hand on this one? I optimistically put you in as the assignee but let us know if this is not something you can engage on.

Blocked by #1755

martindurant:
Hi, I'm watching. I can help, but I don't have a huge number of hours, of course. 

I notice the "blocked by" - and there has been no progress on that issue yet,

martindurant:
Is the use of universal_pathlib desired/required?

d-v-b:
@martindurant `universal_pathlib` was carried over from zarrita; maybe @normanrz can weigh in on the pros and cons of keeping it in.

normanrz:
I think `upath` is quite useful to compose paths. It would be great if the `FsspecStore` would support opening from a `UPath` object. Maybe it should be an optional dependency, though.

martindurant:
OK, I can have it optional. Honestly I doubt we'll see it much, since typical invocation would be like `zarr.open("s3://bucket/data.zarr")` (which already works).

normanrz:
UPath becomes interesting if you store some `storage_options` like this and use it in different places:
```python
path = UPath("s3://bucket", key='some', secret='credentials')
z1 = zarr.open(path / "data.zarr")
z2 = zarr.open(path / "other.zarr")
```

PR #1785, opened 2024-04-10
Basic working FsspecStore

martindurant:
This works. 

- We don't have any list methods, do we need them?
- We don't have a bulk delete
- No exception handling is done here, which has been a thorny issue. Shall we do the same as in v2 with expected exceptions -> KeyError? I don't see in `Store`'s API what the expectation is.

Fixes https://github.com/zarr-developers/zarr-python/issues/1757

TODO:


martindurant:
@jhamman @d-v-b , for discussion

martindurant:
I see we have no storage tests, so don't know how to push this any further.

d-v-b:
I will get some tests for you shortly

jhamman:
@martindurant - the test suite is in a better place now. Are you up for picking this up?

jhamman:
@martindurant - which fsspec implementations have an async-api available?

For now, a mocked s3 backend seems like the way to go.

martindurant:
> which fsspec implementations have an async-api

I think this is a complete list

- s3
- gcs
- azure (both blob and datalake2)
- http
- sshfs (not the builtin ssh/sftp)
- anaconda (released version is still sync)

The following can pass through async, if they wrap an async FS:
- generic
- reference
- dir/prefix

jhamman:
Got it. Makes sense. For now, let's target a mocked test against s3. We can add a few others as we approach a full release.

martindurant:
> a mocked test against s3

Well, I would use the moto server as s3fs does, which is a real S3 implementation with most of the functionality of the real one (minus permissions, object lifetimes and other unimportant things).

src/zarr/store/remote.py
```diff
@@ -11,89 +14,135 @@
...
+        path = _dereference_path(self.path, key)
         # write data
         if byte_range:
-            with fs._open(path, "r+b") as f:
-                f.seek(byte_range[0])
-                f.write(value)
-        else:
-            await fs._pipe_file(path, value)
+            raise NotImplementedError
+        await self._fs._pipe_file(path, value)
```
jhamman, line 103:
```suggestion
        await self._fs._pipe_file(path, value.to_bytes())
```

src/zarr/store/remote.py
```diff
@@ -11,89 +14,135 @@
...
-        fs, root = self._make_fs()
-        path = _dereference_path(root, key)
-
+    async def set(
+        self,
+        key: str,
+        value: Buffer,
+        byte_range: tuple[int, int] | None = None,
+    ) -> None:
+        path = _dereference_path(self.path, key)
```
jhamman, line 99:
```suggestion
        self._check_writable()
        path = _dereference_path(self.path, key)
```

src/zarr/store/remote.py
```diff
@@ -11,89 +14,135 @@
...
+            raise NotImplementedError
+        await self._fs._pipe_file(path, value)
 
     async def delete(self, key: str) -> None:
-        self._check_writable()
-        fs, root = self._make_fs()
-        path = _dereference_path(root, key)
-        if await fs._exists(path):
-            await fs._rm(path)
+        path = _dereference_path(self.path, key)
```
jhamman, line 106:
```suggestion
        self._check_writable()
        path = _dereference_path(self.path, key)
```

src/zarr/store/remote.py
```diff
@@ -11,89 +14,135 @@
...
     ) -> Buffer | None:
-        assert isinstance(key, str)
-        fs, root = self._make_fs()
-        path = _dereference_path(root, key)
+        path = _dereference_path(self.path, key)
 
         try:
-            value = await (
-                fs._cat_file(path, start=byte_range[0], end=byte_range[1])
+            return await (
```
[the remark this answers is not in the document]
jhamman: A similar change will be needed in `get_partial_values`.

jhamman (commented):
@martindurant -- this is coming together. I took it for a spin today and ended up needing to make a few changes (suggestions below). But it works!

martindurant:
I merged from v3 and the suggestions above. I cannot get mypy to play.

By the way, I would suggest that memoryview() and bytes() really ought to work as expected for a Buffer.

src/zarr/store/remote.py
```diff
@@ -11,90 +14,141 @@
...
+            self.path = url.path
+            self._fs = url._fs
         else:
-            assert (
-                len(storage_options) == 0
-            ), "If constructed with a UPath object, no additional storage_options are allowed."
-            self.root = url.rstrip("/")
-
+            raise ValueError("URL not understood, %s", url)
+        self.exceptions = allowed_exceptions
```
jhamman, line 67: @dstansby  - can you take a look here?
```
src/zarr/store/remote.py:67: error: Incompatible types in assignment (expression has type "tuple[type[Exception], ...]", variable has type "type[tuple[Any, ...]]")  [assignment]
```

jhamman:
@martindurant -- thanks for pushing this forward...

> I merged from v3 and the suggestions above. I cannot get mypy to play.

I'm hoping @dstansby can take a look... we can find a way!

> By the way, I would suggest that memoryview() and bytes() really ought to work as expected for a Buffer.

cc @madsbk, @akshaysubr, @normanrz

martindurant:
I'll add moto-based s3 tests by Thursday

src/zarr/store/remote.py
```diff
@@ -1,7 +1,10 @@
 from __future__ import annotations
 
+from collections.abc import AsyncGenerator
 from typing import TYPE_CHECKING, Any
 
+import fsspec
```
jhamman, line 6: can we do this import in `RemoteStore.__init__`? Would help us keep this an optional dep here.

martindurant: Since fsspec is very small and itself has no further dependencies, making it required will reduce friction for users

martindurant:
I put in a simple test to show it works, and will try to find time to hook in the existing StoreTests

src/zarr/store/remote.py
jhamman: I suppose this is fine -- at least for now.

src/zarr/store/remote.py
normanrz: There was a typo in the definition of `exceptions` https://github.com/zarr-developers/zarr-python/pull/1785/commits/5315fda35c77bfa2cf512c7ba1cd1e8ec35256fe#diff-22c6cf4de485dcc3037bd442d219d8d1c0c56ac62ae3a40b337a0ca088c2fc25R29

src/zarr/store/remote.py
```diff
@@ -11,90 +14,143 @@
...
-
-    def _make_fs(self) -> tuple[AsyncFileSystem, str]:
-        import fsspec
-
-        storage_options = self.root._kwargs.copy()
-        storage_options.pop("_url", None)
-        fs, root = fsspec.core.url_to_fs(str(self.root), asynchronous=True, **self.root._kwargs)
-        assert fs.__class__.async_impl, "FileSystem needs to support async operations."
-        return fs, root
+        return f"<FsspecStore({self.path})>"
```
normanrz, line 77:
```suggestion
        return f"<RemoteStore({self.path})>"
```

dstansby:
Happy to take a look at typing, maybe once the tests are passing? Feel free to ping me again.

normanrz:
Are the tests failing due to dependency upgrades? Do we really need `fsspec>2024` or could we relax that constraint?

martindurant:
> Do we really need fsspec>2024

I recommend it.

I had thought these particular errors were already addressed

src/zarr/store/remote.py
```diff
@@ -11,93 +14,148 @@
...
+        **storage_options: Any,
     ):
-        import fsspec
-        from upath import UPath
+        """
+        Parameters
+        ----------
+        url: root of the datastore. In fsspec notation, this is usually like "protocol://path/to".
+            Can also be a upath.UPath instance/
+        allowed_exceptions: when fetching data, these cases will be deemed to correspond to missinf
```
d-v-b, line 47:
```suggestion
        allowed_exceptions: when fetching data, these cases will be deemed to correspond to missing
```
typo

src/zarr/store/remote.py
```diff
@@ -11,93 +14,148 @@
...
-
-    def _make_fs(self) -> tuple[AsyncFileSystem, str]:
-        import fsspec
-
-        storage_options = self.root._kwargs.copy()
-        storage_options.pop("_url", None)
-        fs, root = fsspec.core.url_to_fs(str(self.root), asynchronous=True, **self.root._kwargs)
-        assert fs.__class__.async_impl, "FileSystem needs to support async operations."
-        return fs, root
+        return f"<RemoteStore({self.path})>"
```
d-v-b, line 77: should we include the protocol here as well?

src/zarr/store/remote.py
```diff
@@ -11,93 +14,148 @@
...
+        self.exceptions = allowed_exceptions
         # test instantiate file system
-        fs, _ = fsspec.core.url_to_fs(str(self.root), asynchronous=True, **self.root._kwargs)
-        assert fs.__class__.async_impl, "FileSystem needs to support async operations."
+        if not self._fs.async_impl:
+            raise TypeError("FileSystem needs to support async operations")
 
     def __str__(self) -> str:
-        return str(self.root)
+        return f"Remote fsspec store: {self.path}"
```
d-v-b, line 74: I think we should put the protocol in here

martindurant:
OK, so the problem with the tests, is that they call get/set in blocking sync code, but the store implementation works in async mode. The test instance is created in sync mode, and gets put on a dedicated fsspec IO thread; but then the async store on the main thread calls the same instance in async mode. Note that moto3 is also running on a thread, to complicate things. I am looking at it.

d-v-b:
> OK, so the problem with the tests, is that they call get/set in blocking sync code, but the store implementation works in async mode. The test instance is created in sync mode, and gets put on a dedicated fsspec IO thread; but then the async store on the main thread calls the same instance in async mode. Note that moto3 is also running on a thread, to complicate things. I am looking at it.

That's great insight, thank you. Please let me know if there's anything about the `StoreTests` design that we should change to make this process simpler.

jhamman (approved):
I see ✅! 

Thanks @martindurant for working out the kinks here! (and @d-v-b and @dstansby for chipping in)

martindurant:
Highlighting [this line](https://github.com/zarr-developers/zarr-python/pull/1785/files#diff-22c6cf4de485dcc3037bd442d219d8d1c0c56ac62ae3a40b337a0ca088c2fc25R109), which works around tests passing offset-length for data which is b"". Python allows you to slice bytes like that (b""[1:2] == b""), but remote stores do not allow this.

src/zarr/store/remote.py
```diff
@@ -1,103 +1,190 @@
...
 class RemoteStore(Store):
+    # based on FSSpec
     supports_writes: bool = True
     supports_partial_writes: bool = False
     supports_listing: bool = True
 
-    root: UPath
+    _fs: AsyncFileSystem
+    path: str
+    exceptions: tuple[type[Exception], ...]
```
d-v-b, line 29: I think we should call this attribute `allowed_exceptions` to match the kwarg and to more clearly convey what it does.

src/zarr/store/remote.py
```diff
@@ -1,103 +1,190 @@
...
-        path = _dereference_path(root, key)
+        path = _dereference_path(self.path, key)
 
         try:
-            value: Buffer | None = await (
-                fs._cat_file(path, start=byte_range[0], end=byte_range[1])
-                if byte_range
-                else fs._cat_file(path)
+            if byte_range:
+                # fsspec uses start/end, not start/length
```
d-v-b, line 89: I count this as additional evidence that we should switch to start/end semantics for the rest of the stores

src/zarr/store/remote.py
```diff
@@ -1,103 +1,190 @@
...
             )
-        except (FileNotFoundError, IsADirectoryError, NotADirectoryError):
-            return None
+            return value
 
-        return value
+        except self.exceptions:
+            return None
+        except OSError as e:
+            if "not satisfiable" in str(e):
```
d-v-b, line 109: flagging this as s3 abstraction leakage that we might want to address later on by making an s3-specific storage class

src/zarr/store/remote.py
```diff
@@ -1,103 +1,190 @@
...
+                        ((k[1][0] or 0) + k[1][1]) if k[1][1] is not None else None,
+                    )
+                    for k in key_ranges
+                ),
+                strict=False,
+            )
+        else:
+            return []
+        # TODO: expectations for exceptions or missing keys?
+        res = await self._fs._cat_ranges(list(paths), starts, stops, on_error="return")
```
d-v-b, line 162: i think returning the exceptions is the right thing here

tests/v2/test_storage.py
```diff
@@ -399,7 +399,9 @@ def test_hierarchy(self):
             assert [] == store.listdir(self.root + "c/x/y")
             assert [] == store.listdir(self.root + "c/d/y")
             assert [] == store.listdir(self.root + "c/d/y/z")
-            assert [] == store.listdir(self.root + "c/e/f")
+            # the following is listdir(filepath), for which fsspec gives [filepath]
```
d-v-b, line 402: what's the advantage of going with POSIX semantics here?

tests/v3/test_store/test_remote.py
```diff
@@ -0,0 +1,118 @@
...
+def get_boto3_client():
+    from botocore.session import Session
+
+    # NB: we use the sync botocore client for setup
+    session = Session()
+    return session.create_client("s3", endpoint_url=endpoint_uri)
+
+
+@pytest.fixture(autouse=True, scope="function")
+def s3(s3_base):
```
d-v-b, line 47: @martindurant could you explain what's happening in this test fixture? e.g., why do we need to manipulate the cache, why do we need to create an instance of `S3FileSystem`?

martindurant: pytest-asyncio creates a new event loop for each async test. When an async-mode s3fs instance is made from async, it will be assigned to the loop from which it is made. That means that if you use s3fs again from a subsequent test, you will have the same identical instance, but be running on a different loop - which fails.

For the rest: it's very convenient to clean up the state of the store between tests, make sure we start off blank each time.

src/zarr/store/remote.py
martindurant: Kerchunk is the exception, storing start/length, mostly because length is generally smaller for chunks in big files.

d-v-b (approved):
looks good, thank you @martindurant et al

PR #1789, opened 2024-04-11
fix: add mypy to test dependencies

### LINKED ISSUE #1788: [V3] mypy not running properly in CI
I'm seeing the following output in CI in the v3 branch, under the `run mypy` stage:
```
Run hatch run test:run-mypy
─────────────────────────── test.py3.10-1.24-minimal ───────────────────────────
/bin/sh: 1: mypy: not found
Error: Process completed with exit code 127.
```

Nonetheless, GHA reports this stage as succeeding :upside_down_face: Maybe because 127 is not one of the exit codes it considers a failure. However I think we can consider this test a failure. How should we fix this?

cc @dstansby

d-v-b:
mypy was not in our dev dependencies. this PR adds it. expect CI to fail as mypy actually does something, instead of silently erroring out.

closes #1788 

TODO:


dstansby:
Since `mypy` runs as part of `pre-commit` anyway, how about just getting rid of the GitHub actions mypy stage?

d-v-b:
> Since mypy runs as part of pre-commit anyway, how about just getting rid of the GitHub actions mypy stage?

We don't run pre-commit as part of CI (and I don't think we necessarily want to) so I do think it makes sense to have mypy as a stage, but I'm open to other opinions here. I am noticing that even with mypy running and reporting errors, CI is still green :thinking:

dstansby:
I suspect https://github.com/zarr-developers/zarr-python/blob/3a9d968dd856efdb3d56cd73ad527baac6ffeded/.github/workflows/test-v3.yml#L41 might be the culprit 😄

Saransh-cpp:
pre-commit is running in the CI (as seen in this PR, but maybe only on PRs? pre-commit.ci is faster than GH Actions so it should be preferred if there are no blockers) - 

image

d-v-b:
> pre-commit is running in the CI (as seen in this PR, but maybe only on PRs? pre-commit.ci is faster than GH Actions so it should be preferred if there are no blockers) -
> image

I stand corrected! in this case, i'm confused that mypy is failing (silently) in GHA, but passing in pre-commit

dstansby:
It's because `types-redis` is installed as part of the pre-commit checks, but not installed on GitHub actions. I think it makes a bit more sense to run `mypy` on the pre-commit check, to avoid duplicating config between the pre-commit config and GitHub actions setup. Alternatively, in GitHub actions `pre-commit run mypy` could be run to make sure the environment is the same as the one developers will see when they run `pre-commit`, but this is kind of what pre-commit.ci is set up to do anyway.

d-v-b:
I'm open to either running mypy in pre-commit or as a stage in CI, but it doesn't seem like we need to run it in both places. It looks like @dstansby and @Saransh-cpp recommend the pre-commit approach; does anyone disagree with this, or otherwise object to removing the mypy stage from CI?

jhamman:
I've found running mypy in pre-commit to be a bit hard to manage -- mostly because we only execute mypy for one Python version. Perhaps there are ways to make sure we have mypy coverage across our matrix of supported Python versions?

d-v-b:
Does the version of python mypy is running on matter? I thought mypy was configured to type-check against a specific python version in [`pyproject.toml`](https://github.com/zarr-developers/zarr-python/blob/ce6fcbbee24f4429a740963cec8b239a0b76cd34/pyproject.toml#L151)

dstansby:
Since `mypy` is a static type checker I don't think it matters which version of Python you run it with, and the results shouldn't depend on that version.

jhamman (approved):

PR #1790, opened 2024-04-11
[v3] Fix some untyped defs

dstansby:
Some of the easier fixes for some of the untyped definitions.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -161,7 +161,6 @@ check_untyped_defs = true
 
 [[tool.mypy.overrides]]
 module = [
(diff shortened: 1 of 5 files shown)


jhamman:
@dstansby -- first, thanks for the continued effort here. With #1809 in, you have a few conflicts to resolve. I also think you can stop making changes to everything in the `v2` directory going forward. We plan to remove that code path once the v3 effort wraps up.

dstansby:
Good to know! I think all the changes here are now redundant, so I'll close.

PR #1792, opened 2024-04-13
implement __eq__ for LocalStore

Charoula-Kyriakides:
implement __eq__ for LocalStore

### DIFF
--- a/src/zarr/v3/store/local.py
+++ b/src/zarr/v3/store/local.py
@@ -67,6 +67,9 @@ class LocalStore(Store):
     def __repr__(self) -> str:
         return f"LocalStore({repr(str(self))})"
 
+    def __eq__(self, other: 'LocalStore') -> bool:
+        return isinstance(other, type(self)) and self.root == other.root
+
     async def get(
         self, key: str, byte_range: Optional[Tuple[int, Optional[int]]] = None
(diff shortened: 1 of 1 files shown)


src/zarr/v3/store/local.py
```diff
@@ -67,6 +67,9 @@ def __str__(self) -> str:
     def __repr__(self) -> str:
         return f"LocalStore({repr(str(self))})"
 
+    def __eq__(self, other: 'LocalStore') -> bool:
```
d-v-b, line 70: After adding `from typing import Self` to the top of the file, I think you can do this:

```suggestion
    def __eq__(self, other: Self) -> bool:
```

d-v-b:
@Charoula-Kyriakides thanks for the contribution! It might be good to have some more information in the title and description of the PR. I can fill those in if you like.

Charoula-Kyriakides:
That'd be great, thanks. I'm still learning. Should it be something similar
to the commit message?

d-v-b:
> That'd be great, thanks. I'm still learning. Should it be something similar to the commit message?

Maybe something simple like "implement `__eq__` for `LocalStore`"

d-v-b (approved):

PR #1802, opened 2024-04-19
Deprecate the experimental v3 implementation

jhamman:
This PR adds a FutureWarning to `assert_zarr_v3_api_available` to alert users that this version of Zarr is being replaced with a spec complaint v3 version. 

TODO:


zarr/_storage/store.py
```diff
@@ -22,9 +23,21 @@
 DEFAULT_ZARR_VERSION = 2
 
 v3_api_available = os.environ.get("ZARR_V3_EXPERIMENTAL_API", "0").lower() not in ["0", "false"]
+_has_warned_about_v3 = False  # to avoid printing the warning multiple times
```
joshmoore, line 26: I thought this was the default but 🤷🏽 

https://docs.python.org/3/library/warnings.html#the-warnings-filter

zarr/_storage/store.py
```diff
@@ -22,9 +23,21 @@
...
 
 
 def assert_zarr_v3_api_available():
+    # we issue a warning about the experimental v3 implementation when it is first used
+    global _has_warned_about_v3
+    if v3_api_available and not _has_warned_about_v3:
+        warnings.warn(
+            "The experimental Zarr V3 implementation in this version of Zarr-Python is not "
```
joshmoore, line 34: :hearT:

joshmoore (approved):
Great idea. I could also see warning if the environment variable is even set, but I defer on that.

zarr/_storage/store.py
jhamman: I certainly tried this first. It's possible this would have worked if I had put warning elsewhere, but as implemented, the warning is emitted multiple times without this switch. I think this is due to the fact that `assert_zarr_v3_api_available` is used in various places around the library.

PR #1804, opened 2024-04-21
[v3] Sync with futures

d-v-b:
Adjusts the `sync` function. It now creates a `Future` from a coroutine, wait for that future to complete in another thread, and access the result value with the `.result()` method. 

The previous implementation relied on mutating a list which was initialized to `[None]`, which is ambiguous if the coroutine being awaited happens to return `None`.

In order to pass pre-commit, I also had to fix type annotations in `v3/array.py`.

TODO:


src/zarr/v3/sync.py
```diff
@@ -65,22 +69,19 @@ def sync(coro: Coroutine, loop: Optional[asyncio.AbstractEventLoop] = None):
...
+    return_result = list(done)[0].result()
+
     if isinstance(return_result, BaseException):
         raise return_result
     else:
         return return_result
 
 
-def _get_loop():
+def _get_loop() -> asyncio.AbstractEventLoop | None:
```
jhamman, line 84:
```suggestion
def _get_loop() -> asyncio.AbstractEventLoop:
```

I suspect you had this here because of some mypy thing but it should be possible to get this to always return a loop. Perhaps an `assert loop[0] is not None` right before the return is all that is needed?

src/zarr/v3/sync.py
```diff
@@ -65,22 +69,19 @@ def sync(coro: Coroutine, loop: Optional[asyncio.AbstractEventLoop] = None):
...
-        # this loops allows thread to get interrupted
-        if event.wait(1):
-            break
-
-    return_result = result_box[0]
+
+    future = asyncio.run_coroutine_threadsafe(_runner(coro), loop)
+
+    done, _ = wait([future])
+    return_result = list(done)[0].result()
```
jhamman, line 76: Can you comment on what is happening here?

Some specific questions / ideas:

1. `wait` returns `done` and `not_done`. Do we need to assert anything about `not_done`? I guess we're expecting expecting that `not_done` is always an empty set because we have set `return_when=ALL_COMPLETED`. 
2. Should we consider specifying the `timeout` parameter top `wait`? Would a future that hangs never return here?
3. Can we check that `len(done)` is exactly 1 before indexing out the result?

src/zarr/v3/sync.py
```diff
@@ -1,27 +1,29 @@
...
 from zarr.v3.config import SyncConfiguration
 
+P = ParamSpec("P")
+T = TypeVar("T")
 
 # From https://github.com/fsspec/filesystem_spec/blob/master/fsspec/asyn.py
 
-iothread: List[Optional[threading.Thread]] = [None]  # dedicated IO thread
-loop: List[Optional[asyncio.AbstractEventLoop]] = [
+iothread: list[threading.Thread | None] = [None]  # dedicated IO thread
```
martindurant, line 22: So we are targeting py >=3.9?

src/zarr/v3/sync.py
```diff
@@ -36,16 +38,18 @@ def _get_lock() -> threading.Lock:
...
+        return await coro
     except Exception as ex:
-        result_box[0] = ex
-    finally:
-        event.set()
+        return ex
 
 
-def sync(coro: Coroutine, loop: Optional[asyncio.AbstractEventLoop] = None):
+def sync(coro: Coroutine[Any, Any, T], loop: asyncio.AbstractEventLoop | None = None) -> T:
```
martindurant, line 52: Is there any reason not to directly re-use the same functions in fsspec, since they exist and have a lot of eyes on them? If they need improvement, that would be gladly received. Also, you should make triple sure that if this code appears in multiple places, there are not accidentally multiple loops on multiple async threads.

src/zarr/v3/sync.py
```diff
@@ -65,22 +69,19 @@ def sync(coro: Coroutine, loop: Optional[asyncio.AbstractEventLoop] = None):
             raise NotImplementedError("Calling sync() from within a running loop")
     except RuntimeError:
         pass
-    result_box: List[Optional[Any]] = [None]
-    event = threading.Event()
-    asyncio.run_coroutine_threadsafe(_runner(event, coro, result_box), loop)
-    while True:
-        # this loops allows thread to get interrupted
```
martindurant, line 72: This comment here explains the way to loop was written. Does the changed code allow for an exception in the main thread (timeout, interrupt and other signals)? Does the GC run while waiting?

src/zarr/v3/sync.py
martindurant: Might depend on whether it is called in the main thread, the IO thread of elsewhere,

src/zarr/v3/sync.py
jhamman: 3.10+ actually!

src/zarr/v3/sync.py
d-v-b: I wasn't aware that these functions existed in fsspec! I'm reading https://github.com/fsspec/filesystem_spec/blob/master/fsspec/asyn.py and the relevant tests right now.

One reason I can think of to not simply import these functions from fsspec is that it ties `zarr-python` to fsspec in a rather unexpected way. That's a rather abstract problem, and surely something we could defer solving if necessary.

Is FSSpec the _only_ place where this kind of functionality exists, or are there more places? Surely we aren't the only python users setting synchronous barriers for async code.

src/zarr/v3/sync.py
d-v-b: > Does the changed code allow for an exception in the main thread (timeout, interrupt and other signals)? Does the GC run while waiting?

We don't test for these things at present, so I have no idea! Tests are needed, regardless of the efforts in this PR, and I see that fsspec basically has this covered.

src/zarr/v3/sync.py
d-v-b: As I understand it, `done, _ = wait([future])` is guaranteed to return `tuple[set with one element, the empty set]`, because we are waiting on a list of futures with 1 element, and `wait` returns when the futures are all done. 

Regarding a timeout, that's a broader question about what error boundaries we want to define for IO operations. At the moment, we basically assume that IO either succeeds or fails after a reasonable amount of time. If we want to include "IO takes forever" as a failure mode, then we would need to start building infrastructure around that higher up in the stack.

src/zarr/v3/group.py
```diff
@@ -415,24 +415,36 @@ def nchildren(self) -> int:
...
     @property
     def children(self) -> List[Union[Array, Group]]:
-        _children = self._sync_iter(self._async_group.children())
-        return [Array(obj) if isinstance(obj, AsyncArray) else Group(obj) for obj in _children]
+        raise NotImplementedError
+        # Uncomment with AsyncGroup implements this method
+        # _children: List[Union[AsyncArray, AsyncGroup]] = self._sync_iter(
+        #     self._async_group.children()
+        # )
+        # return [Array(obj) if isinstance(obj, AsyncArray) else Group(obj) for obj in _children]
```
jhamman, line 423: all of these changes were to make mypy happy.

src/zarr/v3/sync.py
```diff
@@ -57,30 +65,31 @@ def sync(coro: Coroutine, loop: Optional[asyncio.AbstractEventLoop] = None):
...
-    if loop is None or loop.is_closed():
+    if not isinstance(loop, asyncio.AbstractEventLoop):
+        raise TypeError(f"loop cannot be of type {type(loop)}")
+    if loop.is_closed():
         raise RuntimeError("Loop is not running")
     try:
         loop0 = asyncio.events.get_running_loop()
         if loop0 is loop:
-            raise NotImplementedError("Calling sync() from within a running loop")
+            raise SyncError("Calling sync() from within a running loop")
```
jhamman, line 75: I believe this was a bug before as `NotImplementedError` inherits from `RuntimeError` so this exception never raised.

tests/v3/test_sync.py
```diff
@@ -0,0 +1,118 @@
...
+
+    async def foo() -> str:
+        return "foo"
+
+    with patch.object(loop, "is_closed", return_value=True):
+        with pytest.raises(RuntimeError):
+            sync(foo(), loop=loop)
+
+
+@pytest.mark.filterwarnings("ignore:coroutine.*was never awaited")
```
jhamman, line 69: despite this, we're still seeing the below warning. Why?

```
sys:1: RuntimeWarning: coroutine 'test_sync_raises_if_calling_sync_from_within_a_running_loop.<locals>.foo' was never awaited
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
```

d-v-b:
thanks to @jhamman this PR also adds some tests. 

I also added a `timeout` kwarg to `sync`, and made `timeout` an attribute of `SyncConfiguration`, so that the `SyncMixin` class can use it, and I added a test that this works as expected.

@martindurant [wrotehttps://github.com/zarr-developers/zarr-python/pull/1804#discussion_r1573863272)
> This comment here explains the way to loop was written. Does the changed code allow for an exception in the main thread (timeout, interrupt and other signals)? Does the GC run while waiting?

How would you propose we test for these properties? Since `wait` blocks the main thread, it's not clear to me what signals we should be looking for. And can you say more about the GC? How would we test this?

tests/v3/test_sync.py
martindurant: This warning comes after the test has finished, during garbage collection. Maybe doing `del` on the objects (and closure) would fix. Or run the coroutine after the check.

src/zarr/v3/sync.py
jhamman: I suggest we open a ticket for finding a test that exercises this concern. I don't think we should hold back this PR though.

martindurant:
Meta question: is it actually possible to open multiple groups/arrays concurrently from sync code?

d-v-b:
> Meta question: is it actually possible to open multiple groups/arrays concurrently from sync code?

There should not be a technical barrier to this, since this question translates to "is it possible to concurrently open some JSON documents and apply very lightweight parsing to them", but (to my knowledge) we don't have an API that supports this yet, emphasis on "yet". Bulk hierarchy access is a target for v3, and we are trying to create a foundation for doing this concurrently.

d-v-b:
> > we don't have an API that supports this yet, emphasis on "yet".
> 
> I encourage early planning for this, since data IO access will predominantly come from sync code (as part of data processing). Current zarr 2 already has concurrency on data read (with fsspec) within an array, so this would be a clear benefit of the new approach. We've been talking about it for [two years](https://github.com/martindurant/async-zarr/) or more :)

IMO this API requires a data structure that represents an un-stored Zarr hierarchy, which `zarr-python` currently lacks.

The plan is to implement a declarative hierarchy API, which will allow something like this:

```python
g = GroupModel(
  attributes={'foo': 10}, 
  members={
    'group': GroupModel(), 
    'array': ArrayModel(shape=(10,))
    }
  ).to_storage('foo.zarr/path')
```
where `to_storage` is a blocking call that waits for the hierarchy (2 groups and 1 array) to be created.

martindurant:
I was thinking more of data access, such as when xarray wants to eagerly load multiple smallish coordinate arrays, eagerly at dataset open time.

martindurant:
Exactly, this is the kind of design I am after. Maybe it should be a separate issue/discussion.

d-v-b:
> Exactly, this is the kind of design I am after. Maybe it should be a separate issue/discussion.

I opened https://github.com/zarr-developers/zarr-python/issues/1805 so we can chat about it

jhamman (approved):

src/zarr/v3/sync.py
normanrz: I had loosely copied the code from fsspec for zarrita.

Issue #1812, opened 2024-04-24
[v3] Files created during tests


d-v-b:
Running the [dimension separator tests](https://github.com/zarr-developers/zarr-python/blob/v3/tests/v2/test_dim_separator.py) in `v3` results in a directory called `src/fixture` being created, which is not the correct thing for a test to be doing. Files generated for testing should be written to temporary directories. We should change `test_dim_separator` so that it does not write these files.

The files being written come from [this fixture](https://github.com/zarr-developers/zarr-python/blob/0f755cc02e5b501cc99c25a31ec69ee0dbacbbff/tests/v2/test_dim_separator.py#L32). I don't really understand what this fixture is doing, since it takes `tmpdir`, but contains a branch that ignores `tmpdir` and tries to write array data to a hard-coded path in the project directory, if that path is empty. It looks like it's trying to reference [these files](https://github.com/zarr-developers/zarr-python/tree/v3/fixture)? If those files are static test fixtures, then should we move them to the tests directory, and remove the capacity for the pytest fixture in `test_dim_separator.py` to overwrite those fixtures?

@joshmoore do you have any ideas here?

d-v-b:
I found a few more examples of this pattern: https://github.com/search?q=repo%3Azarr-developers%2Fzarr-python+.resolve%28%29&type=code

I feel like if these tests _can_ work with data generated at testing time, then that's preferable to a hard-coded fixture, in which could we replace the logic for locating the project root with a `tmpdir` fixture?

joshmoore:
These aren't temporary files. These are the fixture that are expected to exist. They are getting created, more or less to keep conda-forge (and first time fixture developers) happy. There are extended conversations around this on the feedstock. See https://github.com/zarr-developers/zarr-python/issues/1347

d-v-b:
I'm trying to wrap my head around that issue you linked @joshmoore, maybe I don't understand enough about the conda feedstock process, but why can't we just put these fixture files in the directory with the tests?

joshmoore:
Because then they need to be present after `pip install`.

joshmoore:
conda-forge expects all tests to pass when installed "out of tree".

d-v-b:
but if the fixtures are moved inside the test suite, then aren't they available if and only if the tests are available?

d-v-b:
we have been running all our tests in v3 out of tree for a while now...

d-v-b:
[...]

Issue #138, opened 2017-03-02
Data fixture


alimanfoo:
Create a fixture of Zarr arrays saved into the repo using the current master. Add tests to check these can be read by all future versions of Zarr (supporting the same spec version).

alimanfoo:
Resolved in #146

PR #1813, opened 2024-04-24
Move fixtures to `tests`

d-v-b:
Move static fixtures from `fixture` to `tests/fixture`, and alter the corresponding tests to look for fixtures in the new location. These tests contained code for regenerating these fixtures if they were not found; I removed this code, because the fixtures are guaranteed to exist. 

I think generating fixtures at testing time is better than using static files, but the changes here were easier to make. Down the road we can deprecate the static fixtures in a separate PR.

Also, this diff is extremely unwieldy because of the large number of objects in the fixtures directory. Do we really need thousands of files here? Here are the test files modified by this PR:

- https://github.com/d-v-b/zarr-python/blob/00b9b8aa5f2a5fecc278fe8fecb104105a44c061/tests/v2/test_attrs.py#L38
- https://github.com/d-v-b/zarr-python/blob/00b9b8aa5f2a5fecc278fe8fecb104105a44c061/tests/v2/test_dim_separator.py#L42
- https://github.com/d-v-b/zarr-python/blob/array_fixture_fix/tests/v2/conftest.py

Addresses #1812 

TODO:
* [] Add docstrings and API docs for any new/modified user-facing classes and functions


d-v-b:
note that tests are passing, but pre-commit and appveyor are not, presumably because there are too many files, leading to timeouts.

d-v-b:
an alternative is to revisit #138 and make these arrays smaller before merging

d-v-b:
I think this is basically finished. A summary of the changes:
- Since the static fixtures are only used by the v2 tests, I moved them all to `tests/v2/fixture`
- There's a large number of static array fixtures that exist only to check that the array can be ready by a future version of zarr. This test is quetionably designed, because if the fixture files are not found, they are programmatically generated at test time, which undermines the premise of the test. Nonetheless, _I kept these tests but reduced the number of chunks in each fixture array to 2_, which keeps the number of checked in files low. 
- There are three tests that refer to static fixtures, but do not really need to -- one test checks whether a group can be created with the name "meta", and another test checks whether unicode strings can be written to attributes, and the last are checking dimension separator stuff. The first two now use the tmpdir fixture instead of static data, the latter uses static fixtures in the new `tests/v2/fixture` directory, but even that test could be done with a vanilla `tmpdir` fixture. 
 
The diff is unreadable. I bundled the changes to the test code into a [single commit](https://github.com/zarr-developers/zarr-python/pull/1813/commits/3b2d1ec0ec1cacf3b742c126ac3d86fc29455c64) to make this more reviewable. That's what reviewers should look at.

d-v-b:
In a separate effort, I just checked auto-generated fixture files by mistake again :face_exhaling: can we get some eyes on this PR so we can fix this behavior.

normanrz (approved):

PR #1815, opened 2024-04-24
Modify getsize to return total size, not just the top level

### LINKED ISSUE #253: Zarr DirectoryStore: getsize is wrong
https://github.com/zarr-developers/zarr/blob/3c8e9291e96e090e20caf6950da69a3cc180652f/zarr/storage.py#L864-L871

This code does not account for subfolders created by this example:

```python
store = zarr.TempStore(dir='/tmpfs')
for name, values in data.items():
    zarr.array(values, store=store, path=name, compressor=compressor, filters=filter)
total_size = store.getsize()
```

`total_size` is `26`, (24+2 for `.zattrs` and `.zgroup`), which is wrong. Subfolders are not accounted for.

My TempStore has the following structure:
```
.zattrs
.group
name1/
  1
  2
  3
name2/
  1
  2
  3
```

The file sizes of name1 and name2 are unaccounted for.

Tested with Zarr 2.2.0.

benjeffery:
Fixes #253 

TODO:


jeromekelleher:
I guess it's worth asking what the fate of this will be for the v3 branch - if the code is just in as a short-lived bugfix that dies when we switch over the v3 then I'm not sure it's worth pushing through for our purposes @benjeffery

benjeffery:
> I guess it's worth asking what the fate of this will be for the v3 branch - if the code is just in as a short-lived bugfix that dies when we switch over the v3 then I'm not sure it's worth pushing through for our purposes @benjeffery

Yes, I'm not sure on the time scales here, so would be good to know.

I'm missing some tests here that I didn't spot locally as they were skipped for optional deps, happy to patch those up if this looks like it's needed.

dstansby:
👋 thanks for the PR, and sorry it's been left to wilt a bit! I've just run into this too at https://github.com/zarr-developers/zarr-python/issues/2174, so I would be keen to see this PR finished. If you can patch up the remaining test failures, and rebase on to the current `main` branch, I'd be happy to review and hopefully merge this!

jhamman:
I've changed the base branch from `main` to `support/2.x` in case there is interest in continuing this work.

dstansby:
Hmmm I had a look at fixing the tests, and it looks non-trivial...

benjeffery:
> Hmmm I had a look at fixing the tests, and it looks non-trivial...

Yes, I also had a go and the current tests really aren't set up for the directory size to be different across store types.

benjeffery:
I don't plan to work on this as we've moved over to v3.

Issue #1751, opened 2024-04-05
[v3] Generalized array support


jhamman:
Recent discussions at the Refactor meeting have indicated indicated that adding generalized array support to the v3 roadmap would be useful.

@akshaysubr - can I ask you to take this ticket and push it forward?

akshaysubr:
Would propose renaming this issue to `Generalized NDArray support` to disambiguate between zarr arrays and the underlying ndarray datastructures used.

## Proposal for adding generalized ndarray support to zarr

- Create a native zarr NDArray class for typing and to interface with existing protocols. This includes:
  - Buffer protocol (bytes arrays, etc.)
  - `__array_interface__` (Numpy)
  - `__cuda_array_interface__` (PyTorch, CuPy, etc.)
  - DLPack (TensorFlow, JAX, etc.)
  - Raw pointers

- What this generalized ndarray class might look like:
    ```cpp
    namespace zarr
    {

    namespace py = pybind11;
    using namespace py::literals;

    class Array
    {
    public:
    Array(zarrArrayInfo_t* array_info, int device_id);
    Array(py::object o, intptr_t cuda_stream = 0);

    py::dict array_interface() const;
    py::dict cuda_interface() const;

    py::tuple shape() const;
    py::tuple strides() const; // Strides of axes in bytes
    py::object dtype() const;

    zarrArrayBufferKind_t getBufferKind() const; // Device or Host buffer
    py::capsule dlpack(py::object stream) const; // Export to DLPack

    py::object cpu(); // Move array to CPU
    py::object cuda(bool synchronize, int device_id) const; // Move array to GPU

    const zarrArrayInfo_t& getArrayInfo() const
    {
    return array_info_;
    };
    static void exportToPython(py::module& m);
    };

    } // namespace zarr
    ```

- Interoparability with Numpy
    ```python
    ascending = np.arange(0, 4096, dtype=np.int32)
    zarray_h = zarr.ndarray.as_array(ascending)

    print(ascending.__array_interface__)
    print(zarray_h.__array_interface__)
    print(zarray_h.__cuda_array_interface__)
    print(zarray_h.buffer_size)
    print(zarray_h.buffer_kind)
    print(zarray_h.ndim)
    print(zarray_h.dtype)
    ```

- Interoparability with Cupy
    ```python
    data_gpu = cp.array(ascending)
    zarray_d = zarr.ndarray.as_array(data_gpu)
    print(data_gpu.__cuda_array_interface__)
[...]

PR #1826, opened 2024-05-01
[v3] First step to generalizes ndarray and bytes 

madsbk:
This PR introduces two new classes `NDBuffer` and `Buffer` to represent the data argument between components.  

Currently, we use `numpy.ndarray` and `bytes` to pass around data between components. As discussed in #1751, it would be good with a generalization of the data containers to enable other memory types. 

As a first step, `NDBuffer` and `Buffer` are backed by numpy arrays. In follow-up PRs, we can discuss/implement alternative backends such as cupy arrays or pytorch tensors.  The important point here is to agree on an abstraction that can facilitate a broad range of array/data types. 

cc. @akshaysubr, @jakirkham


src/zarr/buffer.py
```diff
@@ -0,0 +1,157 @@
...
+        )
+
+
+def as_nd_buffer(data: Any) -> NDBuffer:
+    if isinstance(data, NDBuffer):
+        return data
+    return NDBuffer(np.asanyarray(data))
+
+
+def as_ndarray(data: Optional[NDBuffer]) -> Optional[np.ndarray]:
```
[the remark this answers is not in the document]
madsbk: I think it might be good to keep data conversion explicit in the zarr-python codebase. That is, I think all components (stores, codecs, etc.) should be forced to use a method that makes the data type explicit. This will make the code clearer by showing intent and avoid unnecessary implicit data conversion. E.g., if a component access `__cuda_array_interface__` to see if the buffer is on device, the buffer might have to move data from host to device in order to provide said `__cuda_array_interface__`. 

But I agree, the user facing API should support the different protocols implicitly but let’s address this in a follow-up PR?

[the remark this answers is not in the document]
akshaysubr: So, we would want to add a `.to(device)` method to the `NDBuffer` to do H2D and D2H transfers down the line and then how would this function change? Something like this?
```python
def as_ndarray(data: Optional[NDBuffer], device: Optional[Device]) -> Optional[Union[np.ndarray, cp.ndarray]]:
```

src/zarr/buffer.py
```diff
@@ -0,0 +1,150 @@
...
+        self.as_numpy_array().fill(value)
+
+    def copy(self) -> Self:
+        return self.__class__(self.as_numpy_array().copy())
+
+    def transpose(self, *axes: np.SupportsIndex) -> Self:
+        return self.__class__(self.as_numpy_array().transpose(*axes))
+
+
+class Buffer(NDBuffer):
```
normanrz, line 89: Why is the `Buffer` subclassing `NDBuffer`? I suppose the API of the `Buffer` would look more like `bytes` than `ndarray`.

src/zarr/buffer.py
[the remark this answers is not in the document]
madsbk: Yes, something like `.to(device: str)` where `device` is either a predefined location like `host` or `cuda`, or some new user-defined location that each user-provided component in the pipeline understand. The idea is to make it as flexible as possible -- an user should be able to plug-in a whole custom made pipeline.

---

I was thinking more like:
```python
def as_ndarray(
    data: Optional[NDBuffer], 
    factory_function: Callable[[NDBuffer], ArrayLike] = np.asarray
) -> Optional[ArrayLike]:
```
where `factory_function` should return an array given a `NDBuffer`. 

Now, this is where it would be helpful if `NDBuffer` implements the different buffer protocols like DLPack, `__array_interface__`, `__cuda_array_interface__`, etc. This way, the factory function can just be `np.asarray`, `cupy.asarray`, or `torch.tensor` in most cases.

src/zarr/buffer.py
madsbk: To show that a component that supports `NDBuffer` also supports `Buffer` but not necessarily the other way around. 
A `Buffer` is just a specialized instantiation of a `NDBuffer` with `dim=1, stride=1, dtype="b"`.

normanrz: What components do you have in mind that would support both an `NDBuffer` and a flat `Buffer`?
In zarr-python 3 we are quite deliberate about turning ndarrays into bytes and vice versa. Afaik this only happens in the `ArrayBytesCodec`s. I think that is a big win in terms of clarity. I would prefer to keep the buffer types separate to not make it too easy for components to do implicit conversions on their own.

madsbk: > What components do you have in mind that would support both an `NDBuffer` and a flat `Buffer`?

All components implementing one of:
```
ArrayArrayCodec
ArrayArrayCodec
ArrayBytesCodec
ArrayBytesCodecPartialDecodeMixin
ArrayBytesCodecPartialEncodeMixin
``` 
But I see you point, it is a good idea to separate the classes to avoid implicit conversion. 
However, do we have a good reason to support non-contiguous arrays? If not, I would strongly suggest that we use `NDBuffer` exclusively; components that doesn't use the dimension information can just ignore. This would simplify the code a lot!

normanrz: I don't see a reason to support non-contiguous arrays. But wouldn't using `NDBuffer` exclusively mean that we lose the separation of the classes?

madsbk: > But wouldn't using `NDBuffer` exclusively mean that we lose the separation of the classes?

Yes, if we force `NDBuffer` to be contiguous always, we don’t need to differentiate between a n-dim object and a flat 1-dim object because components that don’t need the n-dim info can just ignore it (no need to convert or copy data). This means that all codecs become `ArrayArrayCodec` and we can remove all the Array-Bytes codec combinations.  

 

Now, if we really do need non-contiguous arrays, then the current design with all the Array-Bytes codec combinations is good because it makes conversion and data copying explicit. But if we are operating on contiguous arrays only, I don’t see the point of this added complexity.

normanrz: I am not sure I follow. I am advocating for separating bytes from ndarrays. That has been one of my design goals for the zarr-python 3 refactor. It brings a lot more clarity and invariants throughout the library. For example, a store will just not return a multi-dimensional array, a `bytes` codec will always decode into an ndarray, a `transpose` codec will always return an ndarray. Also, the separation into `ArrayArrayCodec`, `ArrayArrayCodec` and `ArrayBytesCodec` was deliberate. Turning all codecs into `ArrayArrayCodec` would be a step backwards, in my opinion.
I would even argue that separating bytes from ndarrays reduces complexity because each component can rely on the type interfaces instead of having to call `ensure_bytes` or `ensure_ndarray` everywhere.

d-v-b: > I would even argue that separating bytes from ndarrays reduces complexity because each component can rely on the type interfaces instead of having to call ensure_bytes or ensure_ndarray everywhere.

If we force `NDBuffer` to be contiguous,  why would we need to call `ensure_bytes` or `ensure_ndarray`?

normanrz: > > I would even argue that separating bytes from ndarrays reduces complexity because each component can rely on the type interfaces instead of having to call ensure_bytes or ensure_ndarray everywhere.
> 
> If we force `NDBuffer` to be contiguous, why would we need to call `ensure_bytes` or `ensure_ndarray`?

To validate whether the input is a flat bytestream or a multi-dimenional array. If this information is not available through the type system anymore, there need to runtime checks to do this.

madsbk: I think we are talking past each other. By forcing `NDBuffer` to be contiguous, we are saying that `NDBuffer` is guaranteed to be a flat bytestream. Now, it might also have some additional information such as dtype and strides, but this can always be ignored. E.g., the `LocalStore` doesn’t care about dtype and strides, it just reads or writes the `NDBuffer` as one big flat bytestream.  

In fact, all our stores read and write contiguous blocks of memory, which makes it hard to think of a scenario that needs non-contiguous arrays. But please let me know if such a scenario exists.

normanrz: Should have a brief call to talk about this? Maybe that would be easier to resolve. I am free today afternoon (CEST).

My main point remains the separation and enforcement thereof. I understand that components can ignore dtype, stride etc. I would rather have it that they cannot even access this information or call methods such as `astype`, `transpose`, `reshape`. All of this ideally without runtime checks and through the type system.

normanrz:
I think this looks great! Thanks.

madsbk:
This PR is ready for a new round of reviews. As @normanrz suggested, I have removed the inheritance between the two buffer classes `Buffer` and `NDBuffer`.   

`AsyncArray.getitem()` and `AsyncArray.setitem()` now takes a factory function to create a buffer from a ndarray-like object.  We still need a mechanism to tell codecs and stores what type of buffer they should return but let’s do that in a follow-up PR.

src/zarr/buffer.py
```diff
@@ -0,0 +1,446 @@
...
+            Return
+            ------
+                A new NDBuffer or subclass instance
+            """
+
+    class NDArrayLike(Protocol):
+        def __call__(self, ndarray_like: NDArrayLike) -> NDBuffer:
+            """Factory function to coerce an array into a NDBuffer (or subclass)
+
+            Callables implementing the `Factor.NDArrayLike` protocol must return
```
normanrz, line 73:
```suggestion
            Callables implementing the `Factory.NDArrayLike` protocol must return
```

src/zarr/buffer.py
```diff
@@ -0,0 +1,446 @@
...
+            self,
+            *,
+            shape: Iterable[int],
+            dtype: np.DTypeLike,
+            order: Literal["C", "F"],
+            fill_value: Optional[Any],
+        ) -> NDBuffer:
+            """Factory function to create a new NDBuffer (or subclass)
+
+            Callables implementing the `Factor.Create` protocol must create a new
```
normanrz, line 49:
```suggestion
            Callables implementing the `Factory.Create` protocol must create a new
```

src/zarr/buffer.py
```diff
@@ -0,0 +1,446 @@
...
+        bytes_like
+           bytes-like object
+
+        Return
+        ------
+            New buffer representing `bytes_like`
+        """
+        return cls.from_ndarray_like(np.frombuffer(bytes_like, dtype="b"))
+
+    def as_ndarray_like(self) -> NDArrayLike:
```
normanrz, line 155: Would it be possible to have an `as_memoryview` method that could be used for passing to Blosc, Gzip etc.? Would that also work without copying?

madsbk: An `as_memoryview` would work like `as_numpy_array`. It might involve copying data depending on what the underlying ndarray-like object the buffer represent. E.g., if `self._data` is a cupy array, we have to copy from device to host memory.

normanrz: I just found the method name `as_ndarray_like` confusing because it actually returns a wrapped 1d byte array

normanrz (approved):

src/zarr/buffer.py
normanrz: Sweet!

PR #1838, opened 2024-05-05
Remove zstandard dependency in favor of numcodecs

normanrz:
I added `zstandard` as a dependency because the `zstd` in numcodecs doesn't expose the `checksum` config. This will be introduced in https://github.com/zarr-developers/numcodecs/pull/519. Once that PR is merged, we can use the numcodecs version.


mkitti:
The correct default level should be `0` which will then allow the Zstd C library to use the default setting according to how it was built:
https://github.com/facebook/zstd/blob/97291fc5020a8994019ab76cf0cda83a9824374c/lib/zstd.h#L338

src/zarr/codecs/zstd.py
```diff
@@ -15,6 +16,8 @@
     from zarr.config import RuntimeConfiguration
     from zarr.common import BytesLike, JSON, ArraySpec
 
+DEFAULT_ZSTD_LEVEL = 3
```
mkitti, line 19:
```suggestion
DEFAULT_ZSTD_LEVEL = 0
```

mkitti: Here this should stay `0`. I'm not sure if we need this constant.

mkitti (changes requested):
At this level the default should stay `0` which the Zstandard C library will interpret as use the default that was built with as defined by zstd.h:
https://github.com/facebook/zstd/blob/97291fc5020a8994019ab76cf0cda83a9824374c/lib/zstd.h#L338

normanrz:
Now taht numcodecs 0.13 is out, we can move forward with this.

PR #1851, opened 2024-05-07
Feature/store metadata methods

jhamman:
This PR takes adds the option for stores to store metadata in whatever container makes sense for the store. 

- adds `Store.set_metadata` and `Store.get_metadata` methods
  - moves json serialization/deserialization to Store ABC.
  - these methods can be overridden by stores that don't want to store metadata as JSON. For now, I've illustrated this with the `MemoryStore` that now keeps metadata objects as plain dictionaries.

TODO:

### DIFF
--- a/src/zarr/abc/store.py
+++ b/src/zarr/abc/store.py
@@ -1,7 +1,32 @@
 from abc import abstractmethod, ABC
+import json
+import numbers
+from enum import Enum
 
+import numpy as np
 from collections.abc import AsyncGenerator
-from typing import List, Tuple, Optional
+from typing import Any, List, Tuple, Optional
+
+
+def json_default(o: Any) -> Any:
+    # See json.JSONEncoder.default docstring for explanation
+    # This is necessary to encode numpy dtype
+    if isinstance(o, numbers.Integral):
+        return int(o)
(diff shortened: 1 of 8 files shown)


tests/v3/test_codecs.py
```diff
@@ -966,10 +966,10 @@ async def test_blosc_evolve(store: Store):
         codecs=[BytesCodec(), BloscCodec()],
     )
 
-    zarr_json = json.loads(await (store / "blosc_evolve_u1" / "zarr.json").get())
+    zarr_json = await (store / "blosc_evolve_u1" / "zarr.json").get_metadata()
     blosc_configuration_json = zarr_json["codecs"][1]["configuration"]
     assert blosc_configuration_json["typesize"] == 1
-    assert blosc_configuration_json["shuffle"] == "bitshuffle"
+    assert blosc_configuration_json["shuffle"] == BloscShuffle.bitshuffle
```
jhamman, line 972: I threw this in because it makes the test pass but its really not what we want. We were previously letting `json.dumps` do a lot of work to serialize Python objects. I think this means that we now need to push more of that logic into the `to_dict()` methods on the Metadata objects.

cc @d-v-b

Issue #1772, opened 2024-04-05
[v3] revisit runtime config


jhamman:
This issue tracks a evaluation of the v3 runtime config

### Context

The v3 branch runtime config currently looks like this:

https://github.com/zarr-developers/zarr-python/blob/76c345071db950b2362f7588ad20da4a1af03b85/src/zarr/v3/config.py#L34-L38

This is then attached to Array/Group classes

https://github.com/zarr-developers/zarr-python/blob/76c345071db950b2362f7588ad20da4a1af03b85/src/zarr/v3/array.py#L51-L55

#### A few things are missing here:

1. User experience
    - as a user, I may want to set config settings and forget about them (e.g. order, concurrency)
2. Portability
    - I don't know for sure but I really doubt that putting the AsyncIO loop on the Array class is going to work when it comes to serialization

#### Improvements

So looking for some ideas for how to manage this better. Two ideas:

1. Xarray style `set-options`: https://docs.xarray.dev/en/stable/generated/xarray.set_options.html
    - Pros: allows for validation and is typed
    - Cons: a bit bespoke, doesn't support environment variables or a config file option
2. Dask style config - https://donfig.readthedocs.io/en/latest/
    - Pros: very flexible framework, support for environment variables and config files, nested namespaces, etc.
    - Cons: extra dependency (though we could vendor it), no typing or validation

#### what do we expect to go in the runtime config?

- Order
- Concurrency
- logging settings
- what else?

jhamman:
I spoke with @maxrjones today about this. Our thought for now was to try using donfig and see how it goes. We can continue to evaluate the dependency vs vendoring and typing/validation as needed. 

cc @djhoese

normanrz:
Additional config options:
- Specify alternate implementations for `CodecPipeline` (e.g. for a rust-based codec pipeline)
- Specify alternate implementations for codecs (e.g. for GPU-based batch-aware codecs)
- Batch size in the `HybridCodecPipeline`

maxrjones:
thanks @normanrz! Joe mentioned you asked about this today. I'm working on getting a minimal PR opened now and should have that submitted within the next couple hours.

jhamman:
@normanrz - #1855 is now in the `v3` branch. Should clear the way to add additional config options as needed.

jhamman:
Thanks @maxrjones for getting this moving!

PR #1855, opened 2024-05-08
Use donfig for V3 configuration

maxrjones:
Progress towards the proposal in https://github.com/zarr-developers/zarr-python/issues/1772 to use Dask style config (donfig) for configuration in v3.

TODO:
* [X] Migrate sync configuration to donfig
* [x] Migrate runtime configuration to donfig
* [x] Add tests for setting configuration
* [ ] Add docs for listing and setting configuration


maxrjones:
@jhamman fyi I've gotten as far as I can right now on the configuration and will be offline until wednesday. I started working on the byte order in https://github.com/zarr-developers/zarr-python/commit/5de1db2053ad5ec575a2ea691874efe9d581393e, but while the tests pass I don't think it's fully correct which it's left out of this PR.

You're welcome to take over this PR or close if you think it'd be better to make progress separately. I'm free Wednesday 5/15 from 1-3 PM ET to work on this next if it's still an open issue at that time.

src/zarr/array.py
```diff
@@ -204,7 +194,7 @@ async def getitem(self, selection: Selection) -> np.ndarray:
         out = np.zeros(
             indexer.shape,
             dtype=self.metadata.dtype,
-            order=self.runtime_configuration.order,
+            order=self.order,
```
jhamman, line 197: here is where we define the order of the output array

src/zarr/array.py
```diff
@@ -232,25 +222,21 @@ async def _read_chunk(
         out_selection: SliceSelection,
         out: np.ndarray,
     ) -> None:
-        chunk_spec = self.metadata.get_chunk_spec(chunk_coords)
+        chunk_spec = self.metadata.get_chunk_spec(chunk_coords, self.order)
```
jhamman, line 225: here is where we tell the signal to codec pipeline what the order of the output chunk is.

src/zarr/metadata.py
```diff
@@ -184,14 +181,15 @@ def dtype(self) -> np.dtype[Any]:
...
-    def get_chunk_spec(self, _chunk_coords: ChunkCoords) -> ArraySpec:
+    def get_chunk_spec(self, _chunk_coords: ChunkCoords, order: Literal["C", "F"]) -> ArraySpec:
         assert isinstance(
             self.chunk_grid, RegularChunkGrid
         ), "Currently, only regular chunk grid is supported"
         return ArraySpec(
             shape=self.chunk_grid.chunk_shape,
             dtype=self.dtype,
             fill_value=self.fill_value,
+            order=order,
```
jhamman, line 192: this `ArraySpec` object ends up with the output order attached. And this is what we use throughout the codec pipelines, etc.

jhamman (commented):
@normanrz - I picked up where @maxrjones left off on this one. Things are going well but I have one design question for you. 

The question is what to do with the `order` config option. For now, I've added it to the `ArraySpec` object which works but, to some extent, mixes array metadata w/ array runtime options. The relevant constraints that I see are:

1. We want to decide on an order at the time we create the `AsyncArray` object. From then on, all chunks and output arrays should use this. You'll notice a new attribute `AsyncArray.order` defined in the `__init__` method. This is important because we don't wouldn't want the following to change the order:
    ```python
    array = AsyncArray.open(...)
    config.set({'array.order': 'C'}}
    a1 = array.getitem(...)
    config.set({'array.order': 'F'}}
    a2 = array.getitem(...)
    assert a1.order == a2.order
    ```
2. `order` does not make sense to include in the `ArrayMetadata`. It would be convenient to attach to that object but I don't think its the right choice -- we would have to special case serialization of that object so 👎 .

What do you think about the approach I've taken? Do you have other suggestions?

src/zarr/codecs/sharding.py
```diff
@@ -631,6 +621,7 @@ def _get_index_chunk_spec(self, chunks_per_shard: ChunkCoords) -> ArraySpec:
             shape=chunks_per_shard + (2,),
             dtype=np.dtype("<u8"),
             fill_value=MAX_UINT_64,
+            order="C",  # TODO!!!!
```
normanrz, line 624:
```suggestion
            order="C",
```

A hard-coded `C` is fine here, because it is just the shard index. It doesn't surface into user code.

normanrz (approved):
I like it! I think `ArraySpec` is a good place to put it, because it is the specification of how an array shall be constructed. That makes sense to me!

PR #1857, opened 2024-05-08
Remove ArrayV2

normanrz:
This PR removes the `ArrayV2` class. The `Array` class can now take either a `ArrayMetadata` or `ArrayV2Metadata` object and perform create, open and save accordingly. 
Note, there is still a separate `create_v2` method. We might want to merge that into the `create` method.

The V2 codec pipeline is implemented with the new `V2Compressor` and `V2Filters` codecs. They act as wrappers around the compressor and filters of zarr v2. 

I also did some light refactorings in the Group to unify with the Array.

cc @jhamman 

TODO:


normanrz:
@jhamman This PR quickly turned into the elimination of the ArrayV2 class. Is there someone already working on porting over the v2 array tests (@d-v-b)?

src/zarr/group.py
```diff
@@ -49,13 +50,15 @@ class GroupMetadata(Metadata):
...
+    def to_bytes(self) -> dict[str, bytes | None]:
         if self.zarr_format == 3:
             return {ZARR_JSON: json.dumps(self.to_dict()).encode()}
         else:
             return {
                 ZGROUP_JSON: json.dumps({"zarr_format": 2}).encode(),
-                ZATTRS_JSON: json.dumps(self.attributes).encode(),
+                ZATTRS_JSON: json.dumps(self.attributes).encode()
+                if len(self.attributes) > 0
+                else None,
```
d-v-b, line 61: why make this function nullable?

normanrz: To allow that `.zattrs` can be deleted if the attributes are empty.

d-v-b: can't we do that by checking if the attributes are the empty dict `{}`?

normanrz: Sure. But that would break the abstraction a bit.

d-v-b: which abstraction does it break?

d-v-b: I think a special-casing serializing `{}` to JSON as a deletion is a fine price to pay for a simpler (non-nullable) type signature

normanrz: > which abstraction does it break?

The `GroupMetadata` class provides the bytes to store away without needing to know about the bytes contents or semantics.

d-v-b: what if we just omit ZATTRS_JSON from this dict if the attributes are empty?

normanrz: Then the `Group` needs to know about the file names of the zarr formats (which it already does for opening).

d-v-b: alternatively, we could just write the `{}` to storage and not worry about deleting anything. Readers are going to try and fetch that file anyway, so I'm not sure there's a performance difference between an absent JSON file and one that contains `{}`

normanrz: Fine with me.

src/zarr/__init__.py
```diff
@@ -33,7 +32,7 @@ async def open_auto_async(
 def open_auto(
     store: StoreLike,
     runtime_configuration_: RuntimeConfiguration = RuntimeConfiguration(),
-) -> Union[Array, ArrayV2, Group]:
+) -> Union[Array, Group]:
```
d-v-b, line 35:
```suggestion
) -> Array | Group:
```

src/zarr/abc/store.py
```diff
@@ -88,12 +90,12 @@ def supports_partial_writes(self) -> bool:
         ...
 
     @abstractmethod
-    async def set_partial_values(self, key_start_values: List[Tuple[str, int, bytes]]) -> None:
+    async def set_partial_values(self, key_start_values: List[Tuple[str, int, BytesLike]]) -> None:
```
d-v-b, line 93:
```suggestion
    async def set_partial_values(self, key_start_values: list[tuple[str, int, BytesLike]]) -> None:
```

src/zarr/abc/store.py
```diff
@@ -145,3 +147,28 @@ def list_dir(self, prefix: str) -> AsyncGenerator[str, None]:
...
+@runtime_checkable
+class ByteGetter(Protocol):
+    async def get(
+        self, byte_range: Optional[Tuple[int, Optional[int]]] = None
+    ) -> Optional[BytesLike]: ...
+
+
+@runtime_checkable
+class ByteSetter(Protocol):
+    async def get(
```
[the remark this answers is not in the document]
d-v-b: (given that these protocols are not being introduced in this PR, consider this tangential)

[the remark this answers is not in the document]
normanrz: I did that because Python doesn't have intersection types, does it?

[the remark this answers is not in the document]
d-v-b: I _think_ you can just inherit from both classes, and as long as the method / property names don't conflict it should work.

src/zarr/abc/store.py
```diff
@@ -145,3 +147,28 @@ def list_dir(self, prefix: str) -> AsyncGenerator[str, None]:
...
+        self, byte_range: Optional[Tuple[int, Optional[int]]] = None
+    ) -> Optional[BytesLike]: ...
+
+    async def set(self, value: BytesLike, byte_range: Optional[Tuple[int, int]] = None) -> None: ...
+
+    async def delete(self) -> None: ...
+
+
+async def set_or_delete(byte_setter: ByteSetter, value: BytesLike | None) -> None:
```
d-v-b, line 170: do we still need this if we consider writing `{}` to `.zattrs` functionally equivalent to deleting `.zattrs`?

normanrz: This is still required for chunks.

src/zarr/abc/store.py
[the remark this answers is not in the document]
normanrz: I'd probably have to make a `ByteGetterAndSetter` protocol because I want to pass objects conforming to both protocols to a function.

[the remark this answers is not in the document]
d-v-b: I wonder how it would look to test for instances of the intersection type with `isinstance(obj, ByteGetter) and isinstance(obj, ByteSetter)`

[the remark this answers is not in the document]
d-v-b: but I think I see what you mean -- without an concrete class that implements both, it's not clear how to annotate a function that takes the intersection

jhamman (commented):
This is great @normanrz! I'll give this a deeper look next week but here are some initial reactions.

src/zarr/array.py
```diff
@@ -11,45 +11,53 @@
...
 from zarr.chunk_key_encodings import DefaultChunkKeyEncoding, V2ChunkKeyEncoding
-from zarr.metadata import ArrayMetadata
+from zarr.metadata import ArrayMetadata, ArrayV2Metadata
 from zarr.store import StoreLike, StorePath, make_store_path
 from zarr.sync import sync
 
 
-def parse_array_metadata(data: Any) -> ArrayMetadata:
-    if isinstance(data, ArrayMetadata):
+def parse_array_metadata(data: Any) -> ArrayMetadata | ArrayV2Metadata:
```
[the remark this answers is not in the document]
normanrz: Done. It was quite a ride to get ABC and the frozen dataclasses to work together.

src/zarr/array.py
```diff
@@ -117,19 +124,63 @@ async def create(
...
-            runtime_configuration=runtime_configuration,
+            metadata=metadata, store_path=store_path, runtime_configuration=runtime_configuration
         )
 
-        await array._save_metadata()
+        await array._save_metadata(metadata)
+        return array
+
+    @classmethod
+    async def create_v2(
```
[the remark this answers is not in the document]
normanrz: Also done. It explodes the method signatures and weakens mypy-checkability, though.

src/zarr/metadata.py
```diff
@@ -103,16 +109,58 @@ def from_dtype(cls, dtype: np.dtype[Any]) -> DataType:
         return DataType[dtype_to_data_type[dtype.str]]
 
 
-@dataclass(frozen=True)
-class ArrayMetadata(Metadata):
+@dataclass(frozen=True, kw_only=True)
+class ArrayMetadata(Metadata, ABC):
```
jhamman, line 113: thanks for making this a base class, really makes the interface clear.

jhamman (approved):
Thanks @normanrz!

normanrz:
This PR depends on #1670. I'll leave this one open until the other is ready.

PR #1870, opened 2024-05-13
Add more typing to `zarr.group`

dstansby:
This finishes most of the typing for `zarr.group`. I think (at least for now) typing all the `kwargs` as `Any` is the best we can do?

TODO:


jhamman (approved):
>  I think (at least for now) typing all the kwargs as Any is the best we can do?

I think that is right. Once we have all of these convenience methods implemented, I we can probably do a better job typing the implementations.

PR #1875, opened 2024-05-13
Fix a regression with scalar indexing due to #1800

dcherian:
Fixes a regression in indexing out scalars from size-1 chunks.

TODO:
* [x] Closes #1874, Closes #1883
* [x] Add unit tests and/or doctests in docstrings
* [x] Changes documented in docs/release.rst
* [x] GitHub Actions have all passed
* [x] Test coverage is 100% (Codecov passes)

PR #1882, opened 2024-05-16
Revert "Fix `is_total_slice` for size-1 dimensions (#1800)"

### LINKED ISSUE #1874: Regression in v2.18.0 when setting an item in an object array
### Zarr version

v2.18.0

### Numcodecs version

0.12.1

### Python Version

3.12

### Operating System

Windows

### Installation

via `poetry install`

### Description

There seems to be a regression in v2.18.0 (compared to v2.17.2) concerning setting items in an object array.

### Steps to reproduce

Setup:

```py
import zarr
from numcodecs import MsgPack
import numpy as np

root = zarr.group()
arr = root.create_dataset(
    name="my_dataset",
    shape=0,
    dtype=object,
    object_codec=MsgPack(),
)
new_items = [
    ["A", 1],
    ["B", 2, "hello"],
]
arr_add = np.empty(len(new_items), dtype=object)
arr_add[:] = new_items
arr.append(arr_add)
arr[:] # --> array([list(['A', 1]), list(['B', 2, 'hello'])], dtype=object)
```
Problem:

```py
arr[0] = ["C", 3] # raises AttributeError: 'list' object has no attribute 'astype' (from Array._process_for_setitem)
```

Using Zarr v2.17.2, no such exception is raised, and the first item is successfully modified.

### Additional output

_No response_

dcherian:
This reverts commit 9d046ea.

Closes #1874 by reverting #1800

TODO:

### DIFF
--- a/zarr/util.py
+++ b/zarr/util.py
@@ -234,17 +234,8 @@ def is_total_slice(item, shape: Tuple[int]) -> bool:
     if isinstance(item, tuple):
         return all(
             (
-                (
-                    isinstance(it, slice)
-                    and (
-                        (it == slice(None))
-                        or ((it.stop - it.start == sh) and (it.step in [1, None]))
-                    )
-                )
-                # The only scalar edge case, indexing with int 0 along a size-1 dimension
-                # is identical to a total slice
-                # https://github.com/zarr-developers/zarr-python/issues/1730
-                or (isinstance(it, int) and it == 0 and sh == 1)
+                isinstance(it, slice)
+                and ((it == slice(None)) or ((it.stop - it.start == sh) and (it.step in [1, None])))
             )
             for it, sh in zip(item, shape)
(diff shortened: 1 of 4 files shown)


zarr/util.py
```diff
@@ -234,17 +234,8 @@ def is_total_slice(item, shape: Tuple[int]) -> bool:
     if isinstance(item, tuple):
         return all(
             (
-                (
```
dcherian, line 237: Alternatively I could comment this out and add a TODO

joshmoore: From my side, +1 for reverting. We can also revert the revert.

jhamman:
I lean toward failing forward on this issue. #1875 is looking good so I think we should get that out, rather than revert.

dcherian:
As you wish :)

PR #1884, opened 2024-05-16
Feature: Top level V3 API

### LINKED ISSUE #1019: [v3] consider using position only and keyword only arguments
As a side note we might consider using position only and keyword only arguments in Zarr 3 to avoid these issues around renaming arguments.

_Originally posted by @jakirkham in https://github.com/zarr-developers/zarr-python/issues/1018#issuecomment-1116879414_

### LINKED ISSUE #1598: port top level api to `v3` branch.
Zarr's top level API is the primary entrypoint for users to interact with Zarr objects. Stuff like:

- [x] `zarr.open`
- [x] `zarr.zeros`, `zarr.ones`, `zarr.empty`
- [x] `zarr.load`, `zarr.save`
- [x] `zarr.array`, `zarr.group`

This API needs to be wired up to the new `Array` and `Group` classes in the `v3` dev branch. We should definitely attempt to keep these all close to the same API but `v3` will give us the chance to rework these if there is a strong reason to do so.

jhamman:
Marking as Work In Progress for now. Currently waiting on #1670 to merge. 

Closes #1598
Fixes #1019

Open questions I could use feedback on:

1. I've combined `zarr.creation` and `zarr.convenience` into a common API module. This makes sense to me because they are all basically wrappers that help us construct Group/Array objects. What people think about this?
2. I've created parallel synchronous and asynchronous modules with basically identical APIs. All the synchronous functions simply call their asynchronous cousin. I like this much more than having two versions of each function (e.g. `ones` and `async_ones`). Thoughts?

More soon.

TODO:

cc @d-v-b, @aldenks


jhamman:
This is ready for an initial review. To set expectations, I don't plan to address the following optional arguments in this PR:

- chunk_store
- cache_attrs
- synchronizer
- meta_array
- storage_options
- dimension_separator
- write_empty_chunks

These will need to be handled in later PRs.

src/zarr/common.py
```diff
@@ -35,6 +35,7 @@
 Selection = slice | SliceSelection
 ZarrFormat = Literal[2, 3]
 JSON = None | str | int | float | Enum | dict[str, "JSON"] | list["JSON"] | tuple["JSON", ...]
+MEMORY_ORDER = Literal["C", "F"]
```
normanrz, line 38:
```suggestion
MemoryOrder = Literal["C", "F"]
```

src/zarr/api/asynchronous.py
```diff
@@ -0,0 +1,925 @@
...
+from zarr.chunk_key_encodings import ChunkKeyEncoding
+from zarr.common import JSON, MEMORY_ORDER, ChunkCoords, OpenMode, ZarrFormat
+from zarr.group import AsyncGroup
+from zarr.metadata import ArrayV2Metadata, ArrayV3Metadata
+from zarr.store import (
+    StoreLike,
+    make_store_path,
+)
+
+ShapeLike = tuple[int, ...]  # TODO: support int for shape
```
normanrz, line 23: Can we use `ChunkCoords` here?

src/zarr/api/asynchronous.py
```diff
@@ -0,0 +1,925 @@
...
+            chunks = a.chunks
+
+        elif hasattr(a, "chunklen"):
+            # bcolz carray
+            chunks = (a.chunklen,) + a.shape[1:]
+
+    return shape, chunks
+
+
+def _like_args(a: ArrayLike, kwargs: dict[str, Any]) -> None:
```
d-v-b, line 47: instead of mutating kwargs, can we copy it, and make changes to the copy, and return it?

d-v-b:
I added a few minor comments, none of which are blockers. A few high-level thoughts:

A lot of the routines `zarr.convenience` seem strange to me, e.g. `save` has a very weird function signature that we should probably give a second thought when the dust settles from the v3 release. But changes like that are out of scope for this work.


> I've combined zarr.creation and zarr.convenience into a common API module. This makes sense to me because they are all basically wrappers that help us construct Group/Array objects. What people think about this?

This works for me. 

> I've created parallel synchronous and asynchronous modules with basically identical APIs. All the synchronous functions simply call their asynchronous cousin. I like this much more than having two versions of each function (e.g. ones and async_ones). Thoughts?

Definitely prefer what we have here over `async_ones`. I don't see any drawbacks to this organization, since presumably we want everything to rest on the async foundation, but I don't have much experience with libraries that expose async and sync apis.

src/zarr/api/asynchronous.py
jhamman: yes, but maybe not for long. I went ahead and changed it but ShapeLike was intended to be `int | tuple[int, ...]`... Perhaps that will work for ChunkCoords too?

src/zarr/api/asynchronous.py
```diff
@@ -0,0 +1,925 @@
...
+        The new array.
+    """
+    _like_args(a, kwargs)
+    return await ones(**kwargs)
+
+
+async def open_array(
+    *,  # note: this is a change from v2
+    store: StoreLike | None = None,
+    zarr_version: ZarrFormat | None = None,  # deprecated
```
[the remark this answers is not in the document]
jhamman: Because if you don't know which version you are opening, you would like Zarr to discover this for you.

src/zarr/api/asynchronous.py
jhamman: Good call. This function was coped from v2 but is now refactored along the lines of what you were suggesting.

will-moore:
Hi, thanks for all the work on this.
I'm looking at using this from https://github.com/ome/ome-zarr-py and seeing this error when accessing via Dask:

```
  File "/Users/wmoore/Desktop/ZARR/ome-zarr-py/ome_zarr/io.py", line 177, in load
    return da.from_zarr(self.__store, subpath)
  File "/Users/wmoore/opt/anaconda3/envs/zarr_v3/lib/python3.10/site-packages/dask/array/core.py", line 3597, in from_zarr
    z = zarr.Array(url, read_only=True, path=component, **kwargs)
TypeError: Array.__init__() got an unexpected keyword argument 'read_only'
```

Is `read_only` another argument to add to the list above at https://github.com/zarr-developers/zarr-python/pull/1884#issuecomment-2143309555 ?

jhamman:
Hi @will-moore -- interesting to find out that Dask is using the array constructor directly. I would have expected to see `zarr.array(...)`. This PR is covering the API methods formerly found in `zarr.convenience` and `zarr.creation`. We'll have to do a bit of work to handle the case you've shown here.

btw, `zarr.array(url, read_only_true, path=component)` should work with this PR.

jhamman:
I'm planning to merge this in the morning unless I get further reviews.

PR #1889, opened 2024-05-17
Followup on codecs

normanrz:
- Merges the codec batch mixins into the Codec ABC
- Renames `evolve` to `evolve_from_array_spec`
- Uses `as_ndarray_like` in the _v2 codecs. @madsbk maybe we can do even better?


madsbk (approved):
Looks good, nice with some memory-type agnostic codecs!
I will start work on a PR that introduces a CUDA buffer, which these codecs should be able to handle as-is.

PR #1897, opened 2024-05-21
remove fixture files from src

d-v-b:
Removes autogenerated test fixtures from the source tree.

tracking issue: https://github.com/zarr-developers/zarr-python/issues/1812
see https://github.com/zarr-developers/zarr-python/pull/1813 for a potential fix to the underlying issue


d-v-b:
Since there are still valid uses of the v2 codebase in the v3 test suite, we should hold off on merging this until we are sure that we don't need v2 as a comparison point any more.

normanrz (approved):

Issue #1104, opened 2022-08-02
async zarr


martindurant:

jakirkham:
AIUI PR ( https://github.com/zarr-developers/zarr-python/pull/534 ) was exploring the [`concurrent.futures.Executor`]( https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Executor ) approach

joshmoore:
See https://github.com/martindurant/async-zarr/

martindurant:
I have started writing a blog about my implementation, might be out this afternoon. That won't say anything new to people who are already on this thread, but might gain more general interest. Specifically for pyodide/pyscript, I think it's still fair to say that the IO story is very far from solved for typical pydata libraries.

jhamman:
This was a great discussion. Pointing folks to the continuation of this idea slated for `v3`: https://github.com/zarr-developers/zarr-python/pull/1583

Issue #1018, opened 2022-05-03
Creating ZipStore with file-like object


oeway:
The current ZipStore implementation assumes the input is [always a path](https://github.com/zarr-developers/zarr-python/blob/27cf31555975eaeb34e802bfdd59c11222f91597/zarr/storage.py#L1668) which becomes a bit limiting when dealing with in-memory file or virtual file object (e.g. in Pyodide in the browser). The virtual file object support is crucial for making zarr library useful in the browser environment, since the emscripten file system itself are rather limited at the moment.

It would be nice if we can allow passing a file-like object. I did a test with a small modification in the __init__ function and it seems to be working nicely in Pyodide/JupyterLite:

```python
from threading import Lock, RLock
import zipfile

class InMemoryZipStore(zarr.ZipStore):
    def __init__(self, path, compression=zipfile.ZIP_STORED, allowZip64=True, mode='a',
                 dimension_separator=None):

        # store properties
        self.path = None # TODO: This need to be handled properly for os.PathLike or file-like object
        self.compression = compression
        self.allowZip64 = allowZip64
        self.mode = mode
        self._dimension_separator = dimension_separator

        # Current understanding is that zipfile module in stdlib is not thread-safe,
        # and so locking is required for both read and write. However, this has not
        # been investigated in detail, perhaps no lock is needed if mode='r'.
        self.mutex = RLock()

        # open zip file
        self.zf = zipfile.ZipFile(path, mode=mode, compression=compression,
                                  allowZip64=allowZip64)
```

I will maintain this piece of code somewhere for now, but it would be great if we can support this from upstream and eventually have it in pyodide.

joshmoore:
Hey @oeway. Thanks for the suggestion! In your mind is what we need the addition of an optional argument for `path`? Seems like the only possible downside of that would be if someone was expecting the exception from `os.path.abspath(path)` to detect `None`.

oeway:
> Hey @oeway. Thanks for the suggestion! In your mind is what we need the addition of an optional argument for `path`? Seems like the only possible downside of that would be if someone was expecting the exception from `os.path.abspath(path)` to detect `None`.

How about we rename the first argument `path` to `file_or_path`? Then we try to infer the path based on the types:
```python
if isinstance(file_or_path, (str, os.PathLike)):
    self.path = file_or_path
else: # file-like object
    if hasattr(file_or_path, 'name'):
        # Normal file object has the name property which contains the path
        self.path = file_or_path.name
[...]

PR #1903, opened 2024-05-23
Add Pyodide support and CI jobs for Zarr

agriyakhetarpal:
## Description

This PR adds official support for running Zarr v3 via Pyodide, running in the browser via WebAssembly. I have introduced a CI job and various code changes to ensure compatibility, disable unsupported features, and adjust test infrastructure accordingly. The most important changes are grouped below.

I've provided a short summary below. AI disclosure: it was partially generated using the GitHub Copilot actions button that appears when editing the PR description. I've carefully read it and augmented it by adding some points of my own, and I've removed some points that were not particularly relevant.

### WASM/Pyodide platform support

- Added an `emscripten.yml` workflow to run CI tests on Pyodide/WASM.
- Introduced a new constant `IS_WASM` in `src/zarr/_constants.py` to detect WASM environments, and used it throughout the codebase for platform-specific logic
- Updated `zarr.core.sync` to support WASM: synchronous APIs now check for [JSPI](https://v8.dev/blog/jspi) availability. Thread-based event loops are disabled in WASM.
- `codec_pipeline.max_workers` resolves to 1 in WASM environments as we can't spawn new threads, so `FusedCodecPipeline` stays on its sequential path even if a larger value is set explicitly.
- `_safe_move` in `zarr.storage._local` handles file moves safely on Emscripten, as inode-based hard links are not present in its virtual file systems. I've split it up into three platform-specific methods.

### Testing infrastructure and CI adjustments

- Declared a dedicated Pyodide test Hatch environment in `pyproject.toml`. Its pytest invocation passes `-o faulthandler_timeout=0` because pytest's faulthandler watchdog needs a thread to run its timer and cannot start one under Emscripten.
- Updated test dependencies and markers to skip or adapt tests that are incompatible with WASM (e.g., subprocesses, multiprocessing, or threading).
- Monkeypatched `asyncio.Runner` in WASM environments to integrate with [Pyodide’s native event loop, the WebLoop](https://pyodide.org/en/stable/usage/api/python-api/webloop.html). This fixes two issues: (i) various `ResourceWarning`s appearing with consecutive async tests as a result of interactions between the `pytest-asyncio` private event loop and the webloop, and (ii) the webloop takes over as the running loop and thus making pytest-asyncio run coroutines on it instead of trying to run on its own.
- Unit tests are skipped entirely where certain modules from the Python standard library are not available or limited in functionality, which at the time are `threading` and `multiprocessing` (see also: https://github.com/pyodide/pyodide/issues/237).

Here are some related conversations where Pyodide (and its related discussions) has come up as a topic: #536, #1018, #1019, #1104, fsspec/filesystem_spec#960, and zarr-developers/community#43

It is to be noted that this PR has been accompanied by a corresponding sister PR for Pyodide builds for `numcodecs` at zarr-developers/numcodecs#529, where `numcodecs` builds successfully and the test suite passes. The change was proposed after discussion at one of the community meetings for Zarr on March 6th, 2024 ([meeting notes](https://zarr.dev/community-calls/2024/2024-03-06.html)).

TODO:

* [ ] ~Add docstrings and API docs for any new/modified user-facing classes and functions~ No user-facing features or public API were proposed
* [ ] ~New/modified features documented in `docs/user-guide/*.rst`~ There are no new features in this PR

## Additional context

Here are a few xrefs to some other PRs that have been implemented for packages in the Scientific Python ecosystem circa 2024:

- NumPy's Pyodide job: https://github.com/numpy/numpy/pull/25894, by courtesy of @rgommers
- The same for PyWavelets (which is an optional dependency for `scikit-image`): https://github.com/PyWavelets/pywt/pull/701 and https://github.com/PyWavelets/pywt/pull/714
- For the `pandas` repository https://github.com/pandas-dev/pandas/pull/57896
- https://github.com/scikit-image/scikit-image/pull/7350
- and so on

As of May 2026, [PEP 783](https://discuss.python.org/t/pep-783-emscripten-packaging/86862) has been accepted, and wheels for the Pyodide platform, with a "pyemscripten" platform tag, are being made available via PyPI. This PR and the sister numcodecs PR I mentioned above are efforts made in this light to increase our adoption in the ecosystem. Our next release of Pyodide, v314.0.0, is coming soon, and two alphas have already been released. I can draw on the changes I've made in this PR to re-enable Zarr support after it was disabled due to the v3 upgrade in version 0.28 (pyodide/pyodide-recipes#99).

My previous (and still current, but slightly lower priority than before) long-term goal towards this change, when I opened this PR back in 2024, was to implement Sphinx-based interactive documentation via JupyterLite and WASM-powered in-browser kernels, as referenced in Quansight-Labs/czi-scientific-python-mgmt#19; see also Quansight-Labs/czi-scientific-python-mgmt#18. A pilot of this that can be readily tried out is available for the "API reference" pages under the [PyWavelets documentation](https://pywavelets.readthedocs.io/en/latest/ref/wavelets.html). This can be preceded by integrating Sphinx extensions such as [`jupyterlite-sphinx`](https://github.com/jupyterlite/jupyterlite-sphinx) for hosted documentation. Nightly wheels from the Anaconda Scientific Python Nightly Wheels channel can be used for nightly/dev documentation, and hopefully pure Python wheels from PyPI with this PR included for stable versions.


agriyakhetarpal:
Zarr v3 has `crc32c` and `zstandard` as dependencies, which were not present for v2. These packages contain compiled code and are [not available in Pyodide yet](https://github.com/pyodide/pyodide/tree/main/packages/) (based on the [logs from the workflow run](https://github.com/agriyakhetarpal/zarr-python/actions/runs/9213626773/job/25347955126)). I shall try to add recipes for them Pyodide, but I am afraid I shall not have an answer for how long that can take.

agriyakhetarpal:
**Update**: we added both `crc32c` and `zstandard` (please see the issues/PRs linked above). Pyodide 0.26.0 should be out very soon containing both packages plus with a bump to the Python version (to 3.12.1) and an ABI-breaking change (via Emscripten 3.1.58). I shall revisit this PR in a few days.

Next, when Zarr v3 (stable) comes out on PyPI in the next month or so, I shall bump the version/recipe for the in-tree Pyodide build as well. It would be great to have the out-of-tree build, i.e., this PR, ready for review/merging by then, and I will be committed to putting my best efforts in doing so – thanks!

agriyakhetarpal:
**Update:** we added both `crc32c` and `zstandard` (please see the issues/PRs linked above). Pyodide 0.26.0 should be out very soon containing both packages plus with a bump to the Python version (to 3.12.1) and an ABI-breaking change (via Emscripten 3.1.58). I shall revisit this PR in a few days.

Next, when Zarr v3 (stable) comes out on PyPI in the next month or so, I shall bump the version/recipe for the in-tree Pyodide build as well. It would be great to have the out-of-tree build, i.e., this PR, ready for review/merging by then, and I will be committed to putting my best efforts in doing so – thanks!

agriyakhetarpal:
The [Pyodide 0.26.0](https://github.com/pyodide/pyodide/releases/tag/0.26.0) release is out, and we have been lucky to have got both `crc32c` and `zstandard` in time (but in any case, they would have got into a patch release anyway). Let me update this PR in a short moment.

tests/v3/test_group.py
```diff
@@ -18,6 +19,7 @@
 from zarr.store import StorePath
 
 
+@pytest.mark.skipif(IS_WASM, reason="Can't test async code in WASM")
```
jhamman, line 22: This is going to be a problem for moving this PR forward. The entire design of V3 depends on being able to execute async. If threading and async are a show stopper, I think we should evaluate whether this can work another way.

agriyakhetarpal: From the latest run on my fork:

```pwsh
14 failed, 1606 passed, 1572 skipped, 9 xfailed, 93 warnings in 104.70s (0:01:44)
```

jhamman: I assume 99% of the passing tests are coming from the `v2` test suite. I wouldn't put much weight on that. We need `tests/v3` to be passing for this to work.

agriyakhetarpal: Yes. If there is a way to run asynchronous tests in a manner that is not asynchronous, then there might be a way forward, but if Zarr's underlying functionality is going to rely on this, I am not sure if there is a solution. The first way would be to get `pytest-asyncio` both packaged and running under Pyodide.

hoodmane: It is at this point possible to run async tests in Pyodide using stack switching. Just have to implement it.

hoodmane: Thanks for pushing this forward @agriyakhetarpal.

.github/workflows/emscripten.yml
```diff
@@ -0,0 +1,112 @@
...
+          ref: setup-emscripten-ci
+          path: numcodecs-wasm
+          submodules: recursive
+          fetch-depth: 0
+          fetch-tags: true
+
+      # For some reason fetch-depth: 0 and fetch-tags: true aren't working...
+      - name: Manually fetch tags for numcodecs
+        working-directory: numcodecs-wasm
+        run: git fetch --tags
```
agriyakhetarpal, line 83: For unexplained reasons, even this doesn't help.

src/zarr/codecs/zstd.py
```diff
@@ -44,12 +42,12 @@ class ZstdCodec(BytesBytesCodec):
...
-            raise RuntimeError(
-                "numcodecs version >= 0.13.0 is required to use the zstd codec. "
-                f"Version {_numcodecs_version} is currently installed."
-            )
+        # _numcodecs_version = Version(numcodecs.__version__)
+        # if _numcodecs_version < Version("0.13.0"):
+        #     raise RuntimeError(
+        #         "numcodecs version >= 0.13.0 is required to use the zstd codec. "
+        #         f"Version {_numcodecs_version} is currently installed."
+        #     )
```
agriyakhetarpal, line 50: Temporary change; these lines will be uncommented when we either have a solution for the `numcodecs` version being fetched in the CI job or we create yet another alpha with an updated version of `numcodecs`.

src/zarr/core/config.py
```diff
@@ -106,8 +108,8 @@ def enable_gpu(self) -> ConfigSet:
                     ],
                 },
             },
-            "async": {"concurrency": 10, "timeout": None},
-            "threading": {"max_workers": None},
+            "async": {"concurrency": 1 if IS_WASM else 10, "timeout": None},
+            "threading": {"max_workers": 1 if IS_WASM else None},
```
agriyakhetarpal, line 112: The idea here is to block spawning any new threads. I'm not sure if `concurrency` should be set to `1` or if we can safely use `10`.

src/zarr/core/sync.py
```diff
@@ -36,6 +38,45 @@ class SyncError(Exception):
...
+    hack, but it allows us to run tests that use pytest-asyncio.
+
+    This is necessary because pytest-asyncio tries to clean up async generators
+    when tearing down test event loops, but Pyodide's WebLoop doesn't support
+    this as it integrates with the browser's event loop rather than managing
+    its own lifecycle.
+    """
+    try:
+        if not IS_WASM and "pyodide" not in sys.modules:
+            return
```
agriyakhetarpal, line 58: The missing coverage annotations for this test are because Codecov does not know about this WASM CI job where these code paths are hit. I can mark them with a `# pragma: no cover` comment closer to the time of merging this PR.

tests/test_config.py
```diff
@@ -82,8 +83,8 @@ def test_config_defaults_set() -> None:
                     ],
                 },
             },
-            "async": {"concurrency": 10, "timeout": None},
-            "threading": {"max_workers": None},
+            "async": {"concurrency": 1 if IS_WASM else 10, "timeout": None},
+            "threading": {"max_workers": 1 if IS_WASM else None},
```
agriyakhetarpal, line 87: I have the same question as above around what an appropriate `async.concurrency` value would be

agriyakhetarpal (commented):
Hi! I apologise for the radio silence on this PR. I've recently been exploring Pyodide/WASM support for Zarr again, as we aim to reintroduce Zarr to https://github.com/pyodide/pyodide-recipes. We will utilise that repository in the next release of Pyodide, namely version 0.28, scheduled for release in a few weeks – please see the tracking issue at https://github.com/pyodide/pyodide-recipes/issues/99 for a subset of our remaining work. We currently use Python 3.13, and the first version of Zarr that supports it is v3, which meant that we needed to get this working – we had disabled Zarr due to the use of threads. I've tried to revisit support for Pyodide through the recent commits I've pushed, and I've been able to get it to work.

If I understand correctly, Zarr's inherent threading model in v3 can be bypassed in WebAssembly through the browser's/Node.js's event loop, as demonstrated in my changes to `src/zarr/core/sync.py`. The only route it disables is `zarr.sync`, and tests in `test_sync.py` are skipped as a result. If I understand correctly, `zarr.sync` is going away anyway with a later version of Zarr.

This PR depends on its sister PR that I created at zarr-developers/numcodecs#529 where we're hoping to add Pyodide/WASM support for `numcodecs`. All tests have been passing there and it has been ready for further review. Here, I'm temporarily building `numcodecs` from my PR's branch there, as the version of `numcodecs` in the Pyodide distribution itself is old does not satisfy the constraint here (>=0.16). I still haven't been able to figure out why the VCS system is not picking up the right version in CI – it looks like a bug where `actions/checkout@v4` does not fetch tags and does not perform an unshallow clone when checking out to a directory other than `{{ github.workspace }}`. Either way, I've temporarily skipped the version check for now so that we can let the tests run till the end.

Hence, while this is obviously not in a mergeable state – I've been able to get several tests passing and Zarr seems to be working well. I believe this PR is ready for review and further inspection, and it should enable us to integrate a portion of this PR as a patch for Pyodide, allowing Zarr to be re-enabled once the idea receives peer review here. A full review can proceed in the meantime as I work through this PR.

In other respects, what I'm finding a bit concerning is the slowness of the test suite in CI. The CI job took 5 hours to run, which seems like an anomaly – in a Pyodide virtual environment on my macOS machine locally, I've been able to consistently run the tests until completion in ~11 minutes on multiple runs, rather than hours. `test_set_orthogonal_selection_3d` takes 6034 seconds here, which is more than 150x of what it takes outside of CI. Would it make sense to add an additional `slow_in_wasm` marker so that `pytest` can ignore the tests shown below?

```
============================= slowest 10 durations =============================
6034.84s call     tests/test_indexing.py::test_set_orthogonal_selection_3d
4101.22s call     tests/test_properties.py::test_array_creates_implicit_groups
3878.97s call     tests/test_properties.py::test_array_roundtrip
924.30s call     tests/test_properties.py::test_roundtrip_array_metadata_from_store
714.00s call     tests/test_indexing.py::test_set_orthogonal_selection_2d
397.94s call     tests/test_indexing.py::test_set_coordinate_selection_2d
242.29s call     tests/test_indexing.py::test_set_orthogonal_selection_1d
150.35s call     tests/test_indexing.py::test_set_block_selection_2d
121.05s call     tests/test_indexing.py::test_set_block_selection_1d
95.83s call     tests/test_indexing.py::test_get_orthogonal_selection_3d
```

I've also xfailed some flaky tests arising from Hypothesis, which I haven't been able to figure out the cause for, and marked a Blosc test where I get mismatched arrays (which will need further probing into how Blosc was compiled to WASM) as a known failure case – I hope that is alright!

Out of the two failing tests in the CI job:
- `test_array_roundtrip` xfails on my local setup on several reruns due to Hypothesis having degraded performance when generating test cases, and it looks like it strict-xpassed here. It is likely a flake, too.
- I had introduced a lenience factor for `tests/test_group.py::test_group_members_performance[memory]` to accommodate the fact that it takes a little longer to run, likely due to overhead accumulated by running in WASM and the latency coming from the Node.js event loop. The lenience factor/tolerance of 1.1 is apparently good enough for my machine, but we know CI machines are quite unreliable, and the required lenience factor to make it pass here would be 4. I'm open to receiving suggestions about this test. My suggestion would be to rework the test method to test versus multiple group counts—parametrised—and verify that the time taken doesn't scale linearly with the group count (i.e., times should be roughly similar if $O(1)$, later times should be much larger if $O(n)$; and the time ratio should ideally be much less than the linear scaling ratio).

> [!TIP]
> I've introduced a `slow_wasm` marker and skipped the relevant slow tests completely, and the `test_group_members_performance` test should pass now.

Thank you for your time!

src/zarr/core/sync.py
```diff
@@ -36,6 +38,45 @@ class SyncError(Exception):
     pass
 
 
+# TODO: not sure if there is a better place to put this function, so I've kept it here for now.
+def _make_shutdown_asyncgens_noop_for_pyodide() -> None:
+    """
+    Patch Pyodide's WebLoop to fix interoperability with pytest-asyncio.
+
+    WebLoop.shutdown_asyncgens() raises NotImplementedError, which causes
```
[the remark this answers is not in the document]
agriyakhetarpal: >Maybe we should implement it in Pyodide?

I haven't thought about it, but yes, I think this should be helpful to put somewhere in `pytest-pyodide`, if I understand you correctly. I haven't come across this case in any other downstream package, so I'm not sure.

> If this is only problematic in testing environment, how about moving this into conftest.py?

Yes, makes sense to me! I'll move it to `conftest.py` and make it run only when `PYTEST_CURRENT_TEST` is active.

tests/test_group.py
```diff
@@ -1977,6 +1978,12 @@ def test_group_members_performance(store: Store) -> None:
     """
     get_latency = 0.1
 
+    # Performance for Pyodide is only marginally worse than for Python,
+    # usually in the 1.012~1.018 range for a latency of 0.1). So we add
```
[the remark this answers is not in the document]
agriyakhetarpal: This was resolved by resetting to the default `async.concurrency` value of `10` which bumped up the performance – this test should no longer fail.

agriyakhetarpal:
The tests are now passing in https://github.com/zarr-developers/zarr-python/actions/runs/15322119791/job/43108023246?pr=1903 (except the extremely slow ones – also, they don't exhibit the same level of slowness at all if you run them locally); and I've added a release note highlighting Pyodide/WebAssembly support. Thank you!

src/zarr/core/sync.py
```diff
@@ -133,6 +174,17 @@ def sync(
     --------
     >>> sync(async_function(), existing_loop)
     """
+    # WASM environments (like Pyodide) cannot start new threads, so we need to handle
+    # coroutines differently. We integrate with the existing Pyodide WebLoop which
+    # schedules tasks on the browser's event loop using setTimeout():
+    # https://developer.mozilla.org/en-US/docs/Web/API/setTimeout
+    if IS_WASM:
+        current_loop = asyncio.get_running_loop()
+        return current_loop.run_until_complete(coro)
```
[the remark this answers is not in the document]
ryanking13: Oh, yes. It looks like we are enabling `jspi` in `pyodide venv`. We pass `enableRunUntilComplete` parameter in loadPyodide.

https://github.com/pyodide/pyodide/blob/adfa092b26745a16127ce2ad680f81407f07738d/src/templates/python_cli_entry.mjs#L41

But this parameter might not be enabled for users who are using in zarr in the browser. So I think it is safe to not assume that `run_until_complete` would wait until the async job is finished.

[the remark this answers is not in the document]
agriyakhetarpal: That makes sense to me – I can grab a wheel and test it in the browser as well; it's probably not going to work. It looks like there would be a reasonable amount of time between Phase 4 and Phase 5 for JSPI.

Could you give me a code suggestion here? i.e., I'm thinking that it isn't possible to ship support without JSPI in that case (unless you know of a way for this) and that we can't re-enable Zarr's recipe because of this PR alone – so this PR will benefit only CI testing for now?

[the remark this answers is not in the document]
ryanking13: Is this API crucial for `zarr`? If not, I would just raise an error and say it is not supported in WASM environment yet.

If it is, I think the best effort we can do would be something like

```py
ret = current_loop.run_until_complete(coro)
if isinstance(ret, asyncio.Task) or isinstance(ret, asyncio.Future):
    # raise, or show warning to tell JSPI is not enabled and run_until_complete does not work
else:
    return ret 
```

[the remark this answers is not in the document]
agriyakhetarpal: Ah, I see what you mean now. Yes, this is crucial for Zarr, as most of the functionality goes through here. For synchronous operations, it uses `_get_loop()` which needs access to the `ThreadPoolExecutor`, which we don't support.

I tried the wheel in a browser, and we'll indeed need JSPI for Zarr to work in browsers – it is going through this code path:

```
<PyodideTask pending name='Task-71' coro=<zeros() running at /lib/python3.13/site-packages/zarr/api/asynchronous.py:1301> cb=[WebLoop._decrement_in_progress()]>
```

so `run_until_complete`, as you mentioned, is returning a pending task instead of executing it. There's no way to "wait" for an async operation without blocking the main thread, which the browser won't let us do. There are no theoretical workarounds, unless we were to rewrite all of the async operations here to call `pyodide.ffi` and use JavaScript.

I guess I'll add a message on why this doesn't work, and we can at least provide guidance for unsuspecting users so that they know they need to use stack switching. Either way, this PR is suitable for testing Zarr in CI, even if we don't immediately re-enable it. Thanks!

P.S. Firefox has had JSPI behind a flag, but the WebKit folks don't seem to be too keen on it: https://github.com/WebKit/standards-positions/issues/422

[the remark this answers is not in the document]
agriyakhetarpal: Addressed in 090c62bc467e190bfd2f54ce6d8fc34b5bdc1e0d.

src/zarr/core/sync.py
```diff
@@ -133,6 +134,36 @@ def sync(
...
+            raise RuntimeError(
+                "Cannot use synchronous zarr API in browser environments without JSPI. "
+                "Zarr requires JavaScript Promise Integration (JSPI) to work in browsers "
+                "but JSPI is not enabled in your environment.\n"
+                "Solutions:\n"
+                "1. Use the async API instead, with zarr.api.asynchronous"
+                "2. Enable JSPI in your Pyodide setup with "
+                "`loadPyodide({ enableRunUntilComplete: true })`"
+                "3. Use a JSPI-enabled website or browser configuration"
+                "4. If you are using Node.js, pass the --experimental-wasm-jspi flag (v20+)"
```
ryanking13: To avoid confusion, the solution is 1 or 2+3+4. To make this code work in chrome 
 for example, one should do both 1) set `--enable-features=WebAssemblyExperimentalJSPI` and pass `enableRunUntilComplete` flag in loadPyodide.

ryanking13 (commented):
For Pyodide-wise, it looks good to me.

agriyakhetarpal:
@ryanking13 – I've corrected the error message based on your suggestion, thanks!

@hoodmane, surfacing your conversation from above about solving the issue in `webloop.py` rather than just in the test code if it's possible: I've taken a more in-depth look at this, and I think it makes sense not to do much here. The idea is that `shutdown_asyncgens` closes tracked async generators (or the ones it's able to track with `sys.set_asyncgen_hooks()` – which allows the event loop to be notified of the generators that are created, through callbacks), by calling `aclose()` on them during shutdown. CPython uses `_asyncgen_firstiter_hook()` and `_asyncgen_finalizer_hook()` to manage them. 

Next, what I understand here is that `WebLoop`'s lifecycle management philosophy is fundamentally different from standard `asyncio` loops. We're not letting `pyodide.webloop` manage its lifecycle deliberately because it's delegated to the browser. At most, we could add a no-op `async def shutdown_asyncgens()` to `webloop.py`, but 1. it can't stop `pytest-asyncio` from raising this, and 2. I've realised that `pytest-asyncio` is raising only a `RuntimeWarning` here, which `pytest` has been filtering to an error based on Zarr's `pytest` configuration (!), so there was no need to add a fixture for this – there should be no harm in ignoring the warning. Please let me know if you had something else in mind; thanks!

ryanking13:
> At most, we could add a no-op async def shutdown_asyncgens() to webloop.py, but 1. it can't stop pytest-asyncio from raising this, and 2. I've realised that pytest-asyncio is raising only a RuntimeWarning here, which pytest has been filtering to an error based on Zarr's pytest configuration (!), so there was no need to add a fixture for this – there should be no harm in ignoring the warning. Please let me know if you had something else in mind; thanks!

Thanks for the investigation. If it only raises a runtime warning, I think it is okay to leave it as-is (without the fixture) in this PR.

Apart from that, adding no-op implementation of `shutdown_asyncgens` in Pyodide Webloop makes sense to me, as users (not just pytest-asyncio) who might call `shutdown_asyncgens` will face NotImplementedError. I think no-op implementation that does nothing is better than raising an error.

hoodmane:
I think we could add an actual implementation of `shutdown_asyncgens` which is the same as the one for the normal event loop. We would keep track of the generators that are running, then iterate over that set and call `close()` on all of them.

nenb:
> test_set_orthogonal_selection_3d

Hey @agriyakhetarpal 👋

I had some questions related to your comments. 

**My background** 
I have a moderate amount of experience with the `zarr-python` codebase, but I am not an expert nor a maintainer. Treat my comments as _possibly_ correct, but not definitely correct. 😄 

**My high-level understanding of `zarr-python` v3**
- it has an event loop (from `asyncio`) at its core. This is useful for I/O operations e.g. read from object storage like S3.
- the event loop dispatches to a thread-pool for CPU-intensive calculations. This is useful for de/compression calculations like zstd.
- the event loop/async API has a synchronous wrapper for convenience. The synchronous wrapper runs all operations in a blocking manner in a separate thread to the `asyncio` event loop.

**My questions on this PR**
- Is the idea to try and let the user run `zarr` as normal in `pyodide` for the async API? If so, be aware that the de/compression logic ('codecs', see https://github.com/zarr-developers/zarr-python/tree/main/src/zarr/codecs) rely on `asyncio.to_thread`. This is because these are generally CPU-intensive calculations. I'm a little surprised that the CI is still passing for the current PR - I would have expected `pyodide` to throw an exception here.
- For this initial PR, it might also be useful experimenting with the `concurrency` value in the config. Its default is currently 10, but if everything will be run in a single thread for this initial PR, then my instinct is to set this to 1. But this is only speculation on my part!
- I know very little about the internals of `pyodide`. Is it possible (and if so, how much effort would it be) to create a web worker of some kind to handle the CPU-intensive calculations? My impression is that zarr v3 will run much more smoothly on `pyodide` if it was possible to have > 1 thread (it might require some tweaking of configuration defaults still, but fundamentally I think it has been designed with an assumption that there is at least 2 threads available)

Thanks for all the work here! Exciting to think that it may be possible to run `zarr-python` reading Zarr v3 in the browser in the not-so-distant future!

d-v-b:
hi @agriyakhetarpal, what do you think the outlook is on this PR? Is it something you are going to pick up again in the future? If not, would you be OK closing this?

agriyakhetarpal:
> hi @agriyakhetarpal, what do you think the outlook is on this PR? Is it something you are going to pick up again in the future? If not, would you be OK closing this?

@d-v-b hi! Thank you for the ping! I'd say this will be useful for Zarr to include as a PR, but I admit I haven't been tracking this very closely since I last worked on this branch. We did have some improvements added to our webloop code later on which may resolve the previous concerns on this PR, and we do have a new 314.0 release coming soon where I can add this PR as a patch to re-enable Zarr. This PR can then take its due course to be merged. Having a CI job running WASM code in the test suite will be beneficial.

I'll spend some time on this PR. Some things to do:
- Resolve the previous comments from here
- Fix all the merge conflicts
- Respond to @nenb's comments and see if the new webloop changes this, and also the other questions


Thank you!

agriyakhetarpal:
I've just spent a few hours updating the sister PR in numcodecs to this PR: https://github.com/zarr-developers/numcodecs/pull/529. While it isn't ready for merging right away, it's ready for review, and I'd appreciate it if y'all could take a look! The build passes cleanly, and so does the test suite (minus a few skips).

Doing that revealed a few things ahead of our 314.0.0 stable release, one of them being that https://github.com/pyodide/pyodide/pull/6240: we weren't compiling the new CPython 3.14 `compression.zstd` yet, which `pyzstd` uses on 3.14 and above.

We use that numcodecs PR in our CI here because we need a numcodecs WASM build to install in our `pyodide venv` and run the Zarr test suite. I'm going to come back to this PR later in my evening and do the rest. Thanks!

agriyakhetarpal:
> We use that numcodecs PR in our CI here because we need a numcodecs WASM build to install in our `pyodide venv` and run the Zarr test suite.

As part of my rework of this PR (see the commits above), we no longer need the numcodecs PR, as we have an adequate version of numcodecs now in Pyodide 314.0.0a2 (see [pyodide-lock.json](https://cdn.jsdelivr.net/pyodide/v314.0.0a2/full/pyodide-lock.json)) as a prerequisite of this PR. However, that PR would still be fruitful independently, and my intention is to keep updating it!

I have pushed a bunch of changes here to modernise the test workflow, add some Hatch environments plumbing to make Pyodide testing easier locally (I mostly still recommend on Linux and macOS for now, not Windows), add a monkeypatch for pytest-asyncio to interoperate better with the Pyodide WebLoop, and add some fixes for failing tests that were introduced after I merged the `main` branch. There are still some tests being skipped, but hopefully nothing too consequential. Overall, I can see this locally:

`=== 5429 passed, 423 skipped, 7 xfailed in 148.56s (0:02:28) ===`

And the workflow passed in CI just now as well. The job is much faster, too – last I remember, it took ~17 minutes back then!

I am aware that there are some strands of conversations for me to scroll up above and track and respond to (it's a pretty long thread now :D), but otherwise this is ready for another look. I'll pull the non-test changes into a branch in pyodide-recipes to finally re-enable Zarr v3 in the next version of Pyodide. Thank you!

agriyakhetarpal:
Since we've now [released Pyodide 314.0.0 stable](https://blog.pyodide.org/posts/314-release/), I've bumped to it and removed the TODO note. I've pinned pytest<9.1 due to a regression which is being tracked upstream: https://github.com/pytest-dev/pytest/issues/14591, https://github.com/pytest-dev/pytest/issues/14592.

agriyakhetarpal:
Looks like the `astral-sh/setup-uv` action isn't able to recognise the platform marker I've added to ignore installing it on WASM/Emscripten. This doesn't happen on other PRs or on runs in `main`. I'll open an issue.

PR #1907, opened 2024-05-24
Fix list of packages in mypy pre-commit environment

dstansby:
This makes sure the list of dependencies in the mypy pre-commit environment mirrors the dependencies of the package (+ any dependencies needed for tests). I also had to make a small fix in `zstd`, which I think is correct?

TODO:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,6 +21,7 @@ maintainers = [
     { name = "Ryan Abernathey" }
 ]
 requires-python = ">=3.10"
+# If you add a new dependency here, please also add it to .pre-commit-config.yml
 dependencies = [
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -21,6 +21,7 @@ maintainers = [
     { name = "Ryan Abernathey" }
 ]
 requires-python = ">=3.10"
+# If you add a new dependency here, please also add it to .pre-commit-config.yml
```
jhamman, line 24: Thanks for adding this comment. This is a bit of a bummer but it seems there isn't a better way to keep these in sync.

jhamman (approved):

PR #1910, opened 2024-05-24
Buffer Prototype Argument

madsbk:
In order to pass array information down through the whole stack, codecs and stores now take a `prototype` argument. This ways, an user can control what buffers and arrays are created in the stack.

This replaces the `factory` argument


madsbk:
@normanrz if you can find time to review/merge this, it would be much appreciated.

src/zarr/array_spec.py
```diff
@@ -0,0 +1,41 @@
...
+from zarr.common import ChunkCoords, parse_dtype, parse_fill_value, parse_order, parse_shapelike
+
+
+@dataclass(frozen=True)
+class ArraySpec:
+    shape: ChunkCoords
+    dtype: np.dtype[Any]
+    fill_value: Any
+    order: Literal["C", "F"]
+    prototype: Prototype
```
normanrz, line 18: Is the `Prototype` pickleable?

normanrz:
I wonder if we should use the runtime config to configure the buffer implementation instead of a Prototype object that gets passed through?

madsbk:
> I wonder if we should use the runtime config to configure the buffer implementation instead of a Prototype object that gets passed through?

Is it possible/practical to set the runtime config for each `getitem` call? It might change between calls.

src/zarr/array_spec.py
madsbk: The `default_prototype`, which is the one we use in [`ArrayV3Metadata`](https://github.com/zarr-developers/zarr-python/pull/1910/files#diff-6ad3ea3fee31fe35b9e5c70976ad07c8ade9674b1e60578af98be6ab1cdd7ec2R199),  is pickleable. 
AFAIK, it is not possible to type hint _pickable_ but added the requirement in the docstring.

normanrz:
> > I wonder if we should use the runtime config to configure the buffer implementation instead of a Prototype object that gets passed through?
> 
> Is it possible/practical to set the runtime config for each `getitem` call? It might change between calls.

Runtime config can be changed, but it is a global setting so as you pointed out it might change during chunk processing. Do you think it is useful to change the buffer class for multiple getitem calls?

madsbk:
> Do you think it is useful to change the buffer class for multiple getitem calls?

From the users perspective, using a buffer class per zarr arrays seems fine but mixing zarr arrays that are backed by different buffer classes is common. 
Internally, a zarr array mix buffer classes since metadata is always using host memory.

src/zarr/buffer.py
```diff
@@ -464,9 +417,32 @@ def as_numpy_array_wrapper(func: Callable[[npt.NDArray[Any]], bytes], buf: Buffe
...
+        The prototype of the output buffer.
+
+    Returns
+    -------
+        The result of `func` converted to a `prototype.buffer`
+    """
+    return prototype.buffer.from_bytes(func(buf.as_numpy_array()))
+
+
+class Prototype(NamedTuple):
```
normanrz, line 430: I am unsure whether the name `Prototype` conveys what the purpose of this class is. Is that a common way of calling such a construct? Maybe we could make it less general: `BufferPrototype`. @d-v-b any opinion on this?

normanrz:
I'm afraid my recent PRs introduced some conflicts here. Would you mind cleaning these up?

src/zarr/buffer.py
rabernat: I think `Template` would be a better name.

src/zarr/array_spec.py
rabernat: The thing that has to be pickleable is the class, not the object itself. I'm having a hard time imagining any scenario in which a class is not pickleable.

src/zarr/buffer.py
d-v-b: i think prototype works, but I think the need for this class suggests that we might want to re-consider the hard boundary we have drawn between buffers and ndbuffers.

madsbk: I have no strong opinion, `Template` or  `BufferPrototype` are also good names or maybe `BufferTemplate` ?

normanrz: No strong opinion, either, but if we had to decide, I'd go with `BufferPrototype`.

normanrz (approved):

src/zarr/buffer.py
madsbk: Renamed to `BufferPrototype`

Issue #1755, opened 2024-04-05
[v3] Finalize store API


jhamman:
The discussion in https://github.com/zarr-developers/zarr-python/discussions/1686 highlighted a number of good improvements to the store API. This ticket tracks getting those updated in the design doc and implemented.

- [x] update design doc
- [x] update store base class and current implementations.
  - [x] Async Generators (https://github.com/zarr-developers/zarr-python/pull/1844)
  - [x] Store Open/Close/Mode 
  - ~[ ] Store metadata methods (https://github.com/zarr-developers/zarr-python/pull/1851)~

PR #1911, opened 2024-05-25
wip

jhamman:
This adds a new required attribute/property to the store API (`Store.mode`). Enforcing read/write permissions at the store is a much cleaner way of managing access than at the Group/Array level (as it is done in v2). I haven't pushed any test changes yet (waiting on #1900) but those will come next.

Xref: https://github.com/zarr-developers/zarr-python/discussions/1686#discussioncomment-8620796
Toward #1755 

Design note: Think of this as the MVP, with the main focus on the `Store.mode` property and the most basic functionality. I think we can continue to explore how to help stores enforce permissions (e.g. via a decorator or something else) later.

TODO:

### DIFF
--- a/src/zarr/store/memory.py
+++ b/src/zarr/store/memory.py
@@ -4,7 +4,7 @@ from collections.abc import AsyncGenerator, MutableMapping
 
 from zarr.abc.store import Store
 from zarr.buffer import Buffer
-from zarr.common import concurrent_map
+from zarr.common import OpenMode, concurrent_map
 
 
 # TODO: this store could easily be extended to wrap any MutableMapping store from v2
@@ -16,8 +16,11 @@ class MemoryStore(Store):
 
     _store_dict: MutableMapping[str, Buffer]
 
-    def __init__(self, store_dict: MutableMapping[str, Buffer] | None = None):
+    def __init__(
+        self, store_dict: MutableMapping[str, Buffer] | None = None, *, mode: OpenMode = "r"
+    ):
         self._store_dict = store_dict or {}
+        self._mode = mode
 
     def __str__(self) -> str:
         return f"memory://{id(self._store_dict)}"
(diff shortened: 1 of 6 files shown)


src/zarr/store/memory.py
```diff
@@ -47,6 +50,7 @@ async def exists(self, key: str) -> bool:
         return key in self._store_dict
 
     async def set(self, key: str, value: Buffer, byte_range: tuple[int, int] | None = None) -> None:
+        self._check_writable()
```
d-v-b, line 53: It doesn't seem easy for a method like `set` or `set_partial_values` to handle the exception themselves, since the exception gets raised in `_check_writable()`. Also, if someone wants to check if the store is writeable, but _not_ raise an exception (e.g., when including the writable state in a repr of the store), then that would also be hard with `_check_writable()`.

What if `self.writable` was a boolean, and we did `if not self.writable: raise BlaBlaBlaException()` in methods like `set`? This would solve the two issues I raised above.

jeromekelleher (commented):
Lgtm!

src/zarr/store/memory.py
jhamman: @d-v-b - I'm not sure I follow. We can certainly add a `writable` property but wouldn't the behavior be the same?

d-v-b: Yes, with a `writeable` property, the overall behavior of `set` would be the same (raise an exception if `writeable` is False), but the `writeable` property would also be accessible in contexts when we don't want to raise an exception, e.g. displaying a repr for the store that includes the writeable mode. The current design (`_check_writeable()`) preemptively raises an exception and makes this second use case cumbersome.

d-v-b: ... but we can sort this particular issue out at a later time

d-v-b (approved):

PR #1917, opened 2024-05-27
Support all indexing variants

normanrz:
In this PR I ported over all indexing variants from the v2 codebase into v3. The `Array` class exposes the sames methods as in the v2 code base (e.g. `get_basic_selection`, `set_orthogonal_selection`, `oindex[...]`) whereas the `AsyncArray` only has `_get_selection(indexer, ...)` and `_set_selection(indexer, ...)`.

The current status of this PR is: the tests are green but typing still causes some headaches.

There are a few breaking changes of the v3 code that we should discuss:
- 0-dimensional arrays (i.e. single values) are not supported
- selecting fields (object dtype) is shaky
- The `out` kwarg for getitem doesn't work and can be complicated to implement with the new NDBuffer abstraction

Please let me know your thoughts on these limitations @jhamman @d-v-b.

TODO:


d-v-b:
Thanks for this effort! I will give it a closer look later today.


regarding these two concerns: 

> * 0-dimensional arrays (i.e. single values) are not supported
> 
> * selecting fields (object dtype) is shaky

I don't have a lot of experience with either of these features from v2, so I am probably not the best judge of how we want this to look. Are there any big time 0-dimensional array or object dtype users we can ping to have them look it over?

> The out kwarg for getitem doesn't work and can be complicated to implement with the new NDBuffer abstraction

Similarly, I never use `out`, but maybe @madsbk has some thoughts here?

madsbk:
Regarding `out`, I think it should be all or nothing. That is, either we implement an **_output-by-caller_** policy where all components takes an output argument, or we don't use output arguments at all. 

As @akshaysubr point out in https://github.com/zarr-developers/zarr-python/issues/1751#issuecomment-2083503495, it might be diffecult to support an **_output-by-caller_** policy so I am leaning to remove the `out` argument and make it clear that the Zarr stack will copy data in most cases.

normanrz:
While I agree that providing an out kwarg is pointless if zarr-python is doing copies anyways, we have it in the existing API. I wonder if we should continue to provide the out arg but issue a (deprecation) warning?

madsbk:
Good point, I think a deprecation warning is a good idea :)

src/zarr/codecs/pipeline.py
```diff
@@ -371,12 +377,16 @@ async def _read_key(byte_setter: ByteSetter | None) -> Buffer | None:
 
             def _merge_chunk_array(
```
d-v-b, line 378: Does this function need to be defined inside the `write_batch` function? I noticed that `_merge_chunk_array` has access to `drop_axis`, even though `drop_axis` is not passed as an argument to `_merge_chunk_array`, and it took me a bit to notice that this is because `_merge_chunk_array` is defined in the body of another function where `drop_axis` is in scope. To keep things less surprising, maybe `_merge_chunk_array` should be defined at the same scope as all the other functions?

src/zarr/indexing.py
```diff
@@ -2,84 +2,291 @@
...
+BasicSelector = int | slice | EllipsisType
+BasicSelectorTuple = tuple[BasicSelector, ...]
+BasicSelection = BasicSelector | BasicSelectorTuple
+BasicSelectionNormalized = tuple[int | slice, ...]
+CoordinateSelection = npt.NDArray[np.intp]
+BlockSelector = int | slice
+BlockSelection = BlockSelector | tuple[BlockSelector, ...]
+BlockSelectionNormalized = tuple[BlockSelector, ...]
+MaskSelection = npt.NDArray[np.bool_]
+OrthogonalSelector = int | slice | npt.NDArray[np.intp | np.bool_]
```
d-v-b, line 42: we should look for a way to simplify this as much as we can.`OrthogonalSelector` vs `OrthogonalSelection`, where the latter contains the former, is a recipe for confusion.

src/zarr/indexing.py
```diff
@@ -2,84 +2,291 @@
...
     from zarr.chunk_grids import ChunkGrid
 
+BasicSelector = int | slice | EllipsisType
+BasicSelectorTuple = tuple[BasicSelector, ...]
+BasicSelection = BasicSelector | BasicSelectorTuple
+BasicSelectionNormalized = tuple[int | slice, ...]
+CoordinateSelection = npt.NDArray[np.intp]
+BlockSelector = int | slice
+BlockSelection = BlockSelector | tuple[BlockSelector, ...]
+BlockSelectionNormalized = tuple[BlockSelector, ...]
```
d-v-b, line 40: I think when we use a type alias like `BlockSelectionNormalized` for `tuple[BlockSelector,...]` we lose some readability. To know that `BlockSelectionNormalized` is `tuple[int | slice, ...]` I have to do 2 lookups, and we aren't even saving lines of code because the type aliases are longer than the types. Was there a particular problem caused by the plain type aliases?

src/zarr/indexing.py
```diff
@@ -2,84 +2,291 @@
...
+MaskSelection = npt.NDArray[np.bool_]
+OrthogonalSelector = int | slice | npt.NDArray[np.intp | np.bool_]
+OrthogonalSelection = OrthogonalSelector | tuple[OrthogonalSelector, ...]
+OrthogonalSelectionNormalized = tuple[OrthogonalSelector, ...]
```
d-v-b, line 44: From the name `OrthogonalSelectionNormalized` I would expect that this type takes `OrthogonalSelection`, but it takes `OrthogonalSelecTOR`, which is a bit surprising / confusing

normanrz:
I got the types to a state where mypy doesn't complain anymore. I know that the types are far from ideal. However, it will take more work to iron that out given the variety of indexing methods that zarr-python supports and I don't want to hold this off from the alpha release.

d-v-b:
This is great, thanks @normanrz! Agreed that we can sort out the types in a later effort.

d-v-b (approved):

tests/v3/util.py
```diff
@@ -0,0 +1,64 @@
...
+
+try:
+    import fsspec  # noqa: F401
+
+    have_fsspec = True
+except ImportError:  # pragma: no cover
+    have_fsspec = False
+
+
+def abs_container():
```
jhamman, line 40: Can we remove this? I don't think its related to this PR.

tests/v3/util.py
```diff
@@ -0,0 +1,64 @@
...
+    value = os.environ.get(name, "0")
+    return pytest.mark.skipif(value == "0", reason="Tests not enabled via environment variable")
+
+
+try:
+    import fsspec  # noqa: F401
+
+    have_fsspec = True
+except ImportError:  # pragma: no cover
+    have_fsspec = False
```
jhamman, line 37: Can we remove this? I don't think its related to this PR.

tests/v3/util.py
```diff
@@ -0,0 +1,64 @@
...
+        self.counter["__getitem__", key_suffix] += 1
+        return await super().get(key, byte_range)
+
+    async def set(self, key, value, byte_range=None):
+        key_suffix = "/".join(key.split("/")[1:])
+        self.counter["__setitem__", key_suffix] += 1
+        return await super().set(key, value, byte_range)
+
+
+def skip_test_env_var(name):
```
jhamman, line 26: Can we remove this? I don't think its related to this PR.

jhamman (commented):
Looking great @normanrz -- a few comments to help tidy this up.

PR #1929, opened 2024-05-29
implement .chunks on v3 arrays

### LINKED ISSUE #1925: v3 arrays have no `.chunks` attribute
Similar to: https://github.com/zarr-developers/zarr-python/issues/1924

V2 arrays have a `.chunks` attribute: https://zarr.readthedocs.io/en/stable/_autoapi/zarr.core.Array.html#zarr.core.Array.chunks

V3 arrays don't:

```python
import zarr
arr = zarr.Array.create(store=zarr.store.MemoryStore(), shape=(10,), chunks=(5,), dtype="i4")
arr.chunks
# -> attribute error

# instead it is available at
arr.metadata.chunk_grid.chunk_shape
```

It would be a pretty disruptive API change to remove this. I understand that the ChunkGrid concept makes this more complicated to support. We should reimplement a sane default for this attribute for regular chunk grids. So much code likely relies on this parameter.

rabernat:
Closes #1925

TODO:
* ~Add docstrings and API docs for any new/modified user-facing classes and functions~
* ~New/modified features documented in docs/tutorial.rst~
* ~Changes documented in docs/release.rst~


src/zarr/array.py
```diff
@@ -343,6 +343,16 @@ def ndim(self) -> int:
     def shape(self) -> ChunkCoords:
         return self.metadata.shape
 
+    @property
+    def chunks(self) -> ChunkCoords:
```
normanrz, line 347: Should we add a deprecation warning here? I think users should instead look at the chunk grid directly.

d-v-b: Is the chunk grid object is guaranteed to have a consistent structure, beyond `{name: str, configuration: dict[str, JSON]}`? We might instead want users to look at a normalized (explicit) representation of the chunks, i.e. `tuple[tuple[int, ...], ...]`

rabernat: My opinion is that having a backwards compatible `.chunks` attribute is very important. Deprecating this or changing its behavior will have far-reaching consequences for libraries that depend on Zarr. Given that we only support regular chunk grids so far, this solution feels very reasonable to me. Once other chunk grids are developed this will raise an attribute error, and the user code can handle it some other way (i.e. looking at the chunk grid directly).

d-v-b: @rabernat I think your reasoning makes sense, given that irregular chunks are prospective at this point. 

Thinking about future changes to support irregular chunks, do you think it would make sense to transform the type of `chunks` to `tuple[tuple[int, ...], ...]` (matching the `chunks` attribute of a dask array), assuming we give a long deprecation warning?

normanrz: > My opinion is that having a backwards compatible `.chunks` attribute is very important. Deprecating this or changing its behavior will have far-reaching consequences for libraries that depend on Zarr. Given that we only support regular chunk grids so far, this solution feels very reasonable to me. Once other chunk grids are developed this will raise an attribute error, and the user code can handle it some other way (i.e. looking at the chunk grid directly).

Fair enough, we can revisit this once we have other chunk grids.



> `tuple[tuple[int, ...], ...]`

How would that scale if you have billions of chunks?

src/zarr/array.py
```diff
@@ -343,6 +343,16 @@ def ndim(self) -> int:
...
+        try:
+            chunk_shape = getattr(self.metadata.chunk_grid, "chunk_shape")  # noqa: B009
+            return chunk_shape
```
jhamman, line 350:
```suggestion
            return self.metadata.chunk_grid.chunk_shape
```

B009 is actually useful here. 

> B009: Do not call getattr(x, 'attr'), instead use normal property access: x.attr. Missing a default to getattr will cause an AttributeError to be raised for non-existent properties. There is no additional safety in using getattr if you know the attribute name ahead of time.

jhamman (approved):
This is a nice backward compatible change. We probably want to move users toward something like `Array.get_chunk_grid() -> Generator[tuple[int, ...], ...]` but for now, this is good for supporting existing applications.

One small suggested change.

src/zarr/array.py
rabernat: Ok done.

rabernat:
This is my first PR to the V3 branch.

How are we handling docs? Is there anything to update?

d-v-b:
> This is my first PR to the V3 branch.
> 
> How are we handling docs? Is there anything to update?

Docs are a (big) tbd, for the moment code changes alone are fine.

src/zarr/array.py
```diff
@@ -343,6 +343,15 @@ def ndim(self) -> int:
     def shape(self) -> ChunkCoords:
         return self.metadata.shape
 
+    @property
+    def chunks(self) -> ChunkCoords:
+        try:
+            return self.metadata.chunk_grid.chunk_shape  # type: ignore[attr-defined]
```
rabernat, line 349: I am fighting with mypy here. My local pre-commit env does not like
```
# type: ignore[attr-defined, no-any-return]
```

but that seems to be what the CI-based pre-commit linter wants

jhamman: I took a swing at fixing this @rabernat. Let me know what you think.

PR #1938, opened 2024-06-01
narrow the type of drop_axes to just tuple[int, ...]

d-v-b:
I think an empty tuple conveys "drop no axes" as effectively as `None`, with less code.


src/zarr/codecs/pipeline.py
```diff
@@ -381,6 +381,7 @@ def _merge_chunk_array(
                 out_selection: SelectorTuple,
                 chunk_spec: ArraySpec,
                 chunk_selection: SelectorTuple,
+                drop_axes: tuple[int, ...],
```
d-v-b, line 384: this partially addreses my concern raised here:  https://github.com/zarr-developers/zarr-python/pull/1917/files#r1623186488

normanrz:
Thanks!

Issue #1783, opened 2024-04-10
[v3] Path to 100% typing


dstansby:
I thought I'd open an issue to track the path to adding 100% typing to the v3 branch. The mypy docs have a good [guide to gradually adding typing to existing code](https://mypy.readthedocs.io/en/stable/existing_code.html), and from there I think the best way to get to enabling the `mypy` strict option is work through the sub-options that make up the strict rule:

- [x] warn_unused_configs = True https://github.com/zarr-developers/zarr-python/pull/1781
- [x] warn_redundant_casts = True https://github.com/zarr-developers/zarr-python/pull/1781
- [x] warn_unused_ignores = True https://github.com/zarr-developers/zarr-python/pull/1781
- [x] strict_equality = True https://github.com/zarr-developers/zarr-python/pull/1793
- [x] strict_concatenate = True https://github.com/zarr-developers/zarr-python/pull/1793
- [x] check_untyped_defs = True https://github.com/zarr-developers/zarr-python/pull/1784, https://github.com/zarr-developers/zarr-python/pull/1790
- [x] disallow_subclassing_any = True
- [x] disallow_untyped_decorators = True https://github.com/zarr-developers/zarr-python/pull/1793
- [x] disallow_any_generics = True https://github.com/zarr-developers/zarr-python/pull/1794
- [x] disallow_untyped_calls = True https://github.com/zarr-developers/zarr-python/pull/1811
- [x] disallow_incomplete_defs = True https://github.com/zarr-developers/zarr-python/pull/1814
- [x] disallow_untyped_defs = True https://github.com/zarr-developers/zarr-python/pull/1834
- [x] no_implicit_reexport = True https://github.com/zarr-developers/zarr-python/pull/1908
- [x] warn_return_any = True

From [repo-review](https://scientific-python.github.io/repo-review/?repo=zarr-developers%2Fzarr-python&branch=v3)
- [x] warn_unreachable = True https://github.com/zarr-developers/zarr-python/pull/1937
- [x] enable_error_code = ["ignore-without-code", "redundant-expr", "truthy-bool"] https://github.com/zarr-developers/zarr-python/pull/1909

We can use the list above to track progress, and link pull requests that address each option in turn (from top to bottom).

PR #1939, opened 2024-06-01
Fix final typing errors

dstansby:
This should fix the final typing errors. Fixes https://github.com/zarr-developers/zarr-python/issues/1783

### DIFF
--- a/src/zarr/store/remote.py
+++ b/src/zarr/store/remote.py
@@ -62,7 +62,7 @@ class RemoteStore(Store):
         path = _dereference_path(root, key)
 
         try:
-            value = await (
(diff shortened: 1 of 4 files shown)


jhamman (approved):

src/zarr/store/remote.py
```diff
@@ -62,7 +62,7 @@ async def get(
         path = _dereference_path(root, key)
 
         try:
-            value = await (
```
jhamman, line 65: this is not 100% correct but I think it can be fixed in #1785

d-v-b (approved):

PR #1945, opened 2024-06-02
Add type checking to tests

dstansby:
This adds type checking for the v3 tests.  There's still quite a few errors that need fixing... does anyone want to volunteer to fix them here, or should I just ignore errors in the files that are currently erroroing for now?

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -25,7 +25,7 @@ repos:
     rev: v1.10.0
     hooks:
       - id: mypy
-        files: src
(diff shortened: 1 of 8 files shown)


jhamman:
> does anyone want to volunteer to fix them here, or should I just ignore errors in the files that are currently erroroing for now?

I'm happy to help plug away at these after we get the alpha release out. For now, ignoring the errors seems like the right thing.

dstansby:
Going to close this as I don't think it's worth the effort to fix at the moment (while the tests are still in flux)

PR #1952, opened 2024-06-05
[v3] add json indentation to config

### LINKED ISSUE #704: Compact JSON in .zarray
JSON output is deliberately made human-readable with much whitespace. That produces large `.zarray` files with string arrays and categorize. In one small example with about 150 different strings, the human-readable `.zarray` was 3837 bytes and the compact version was 1284 bytes. With a larger variety of strings, the difference would be larger.

As Zarr is a storage format that isn't intended for human readability, I would like to propose to write JSON with `indent=None, separators=(",", ":")`.

d-v-b:
Level of JSON indentation is set by by the global config object.

resolves #704 

TODO:

### DIFF
--- a/src/zarr/group.py
+++ b/src/zarr/group.py
@@ -25,6 +25,7 @@ from zarr.common import (
     ChunkCoords,
     ZarrFormat,
 )
+from zarr.config import config
 from zarr.store import StoreLike, StorePath, make_store_path
 from zarr.sync import SyncMixin, sync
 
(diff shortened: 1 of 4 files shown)


d-v-b:
The correct tests would compare JSON documents with different levels of `json_indent` values, to ensure that the indentation is working correctly. If we do want those tests, I can add them, but I think I would rather defer that for a big sprint to fill out the metadata tests.

src/zarr/group.py
```diff
@@ -79,14 +80,21 @@ class GroupMetadata(Metadata):
     node_type: Literal["group"] = field(default="group", init=False)
 
     def to_buffer_dict(self) -> dict[str, Buffer]:
+        json_indent = config.get("json_indent")
         if self.zarr_format == 3:
-            return {ZARR_JSON: Buffer.from_bytes(json.dumps(self.to_dict()).encode())}
+            return {
+                ZARR_JSON: Buffer.from_bytes(
```
normanrz, line 86: Maybe we should add a Buffer.from_json method

normanrz (approved):

jhamman:
I like the feature and I'm happy to see this go in. I expect the specific implementation will change a bit with #1851 but that is fine with me.

will-moore:
Looks great, thanks. Working as expected for me. 👍

PR #1956, opened 2024-06-09
[v3] `StoreTests[RemoteStore]`

d-v-b:
Attempts to specialize `StoreTests` with `RemoteStore`. Depends on #1785. 

This is currently broken because I can't figure out the correct way to link up the `s3` fixture with the `StoreTests` instance. Help would be appreciated!

TODO:


jhamman:
Thanks @d-v-b for advancing this!

joshmoore:
A heads up that I am likely in the throws of the same or at least similar issues here:

https://github.com/ome/ome2024-ngff-challenge/pull/37

The mock tests failed but also a production run against a production but non-standard `--endpoint-url` is failing with:

```
botocore.exceptions.HTTPClientError: An HTTP Client raised an unhandled exception: Task <Task pending name='Task-13' coro=<_runner() running at /opt/homebrew/Caskroom/mambaforge/base/envs/challenge4/lib/python3.12/site-packages/zarr/sync.py:51> cb=[_chain_future.<locals>._call_set_state() at /opt/homebrew/Caskroom/mambaforge/base/envs/challenge4/lib/python3.12/asyncio/futures.py:394]> got Future <Future pending cb=[BaseSelectorEventLoop._sock_write_done(10, handle=<Handle BaseS...96.70', 443))>)()]> attached to a different loop
```

I'll read through @d-v-b's history and see if it's the same problem or not.

martindurant:
If the s3fs instance is to be used in async mode (which is the case), then it must be created within a coroutine running on the same event loop and thread as when it will be called. You _can_ assign the loop= kwargs and defer creation of the session object, but it's better to follow the above rule.

> but non-standard --endpoint-url

This will make a new instance, which is probably the difference. fsspec will reuse instances having exactly the same set of init arguments unless otherwise directed.

joshmoore:
Thanks, @martindurant! That makes sense conceptually, but can you give me a pointer on how to do that reliably for the new RemoteStore?

martindurant:
I suppose `RemoteStore(...)` must also be done in a coroutine and either pass asynchronous=True explicitly or ensure that no other s3fs instance was created beforehand. Although I helped make the class, it was far from clear to me how and when it was to be used...

PR #1961, opened 2024-06-12
Clean up typing and docs for indexing

### LINKED ISSUE #1946: Clean up typing and docs for indexing
#1917 ported all indexing variants from the v2 codebase to v3. However, the typing is a mess and the docs are missing. This needs to be addressed before the 3.0 release.

brokkoli71:
fixes #1946

**Changes:**
- Improve **typing** of indexing variants ported from v2 in #1917 
- Add **docstrings** for indexing variants ported from v2 in #1917 
- The interfaces should accept **scalar values**. Therefore, the type of **input interfaces** was changed from `NDArrayLike` to `npt.ArrayLike` which accepts scalars


src/zarr/indexing.py
```diff
@@ -1206,8 +1196,8 @@ def pop_fields(selection: SelectionWithFields) -> tuple[Fields | None, Selection
         return fields, selection
 
 
-def make_slice_selection(selection: Any) -> list[int | slice]:
-    ls: list[int | slice] = []
+def make_slice_selection(selection: Any) -> list[IntOrSlice]:
```
d-v-b, line 1199: question for people who know array indexing better than me: is there any reason for this function to return `x: int`, when we can express the same thing with `slice(x, x + 1, 1)`?

brokkoli71: I think you are right, it should only return `list[slice]`. Is that what you mean?

d-v-b: No, I think the annotation is correct for this function, but I am curious more broadly about _why_ this function returns ints, as opposed to turning ints into slices

d-v-b (approved):
This is a good improvement! The beastiary of types in `indexing.py` is still a problem, but I think we should try solve that at the level of the indexing API itself

normanrz (approved):
Awesome work!

PR #1964, opened 2024-06-12
Update `RemoteStore.__str__` and add UPath tests

d-v-b:
This PR changes the `.__str__` method of `RemoteStore` to make it similar to `LocalStore` and `MemoryStore` (i.e., `RemoteStore("s3://path").__str__` returns `s3://path`). 

~I also add tests for UPath, and tweaked the test suite accordingly. The tests are failing, and I don't know why. @martindurant could you take a look if you have time?~ edit: tests are passing

TODO:

### DIFF
--- a/src/zarr/store/remote.py
+++ b/src/zarr/store/remote.py
@@ -52,8 +52,8 @@ class RemoteStore(Store):
         """
 
         super().__init__(mode=mode)
-        self._url = url
         if isinstance(url, str):
+            self._url = url
             self._fs, self.path = fsspec.url_to_fs(url, **storage_options)
         elif hasattr(url, "protocol") and hasattr(url, "fs"):
             # is UPath-like - but without importing
@@ -62,8 +62,10 @@ class RemoteStore(Store):
                     "If constructed with a UPath object, no additional "
                     "storage_options are allowed"
                 )
+
+            self._url = str(url)
             self.path = url.path
-            self._fs = url._fs
+            self._fs = url.fs
         else:
             raise ValueError("URL not understood, %s", url)
         self.allowed_exceptions = allowed_exceptions
--- a/src/zarr/store/remote.py
+++ b/src/zarr/store/remote.py
@@ -54,7 +54,8 @@ class RemoteStore(Store):
         super().__init__(mode=mode)
         if isinstance(url, str):
             self._url = url.rstrip("/")
-            self._fs, self.path = fsspec.url_to_fs(url, **storage_options)
+            self._fs, _path = fsspec.url_to_fs(url, **storage_options)
+            self.path = _path.rstrip("/")
         elif hasattr(url, "protocol") and hasattr(url, "fs"):
             # is UPath-like - but without importing
             if storage_options:
@@ -62,9 +63,10 @@ class RemoteStore(Store):
                     "If constructed with a UPath object, no additional "
                     "storage_options are allowed"
                 )
-
+            # n.b. UPath returns the url and path attributes with a trailing /, at least for s3
+            # that trailing / must be removed to compose with the store interface
             self._url = str(url).rstrip("/")
-            self.path = url.path
+            self.path = url.path.rstrip("/")
             self._fs = url.fs
         else:
             raise ValueError("URL not understood, %s", url)
--- a/src/zarr/store/remote.py
+++ b/src/zarr/store/remote.py
@@ -53,7 +53,7 @@ class RemoteStore(Store):
 
         super().__init__(mode=mode)
         if isinstance(url, str):
-            self._url = url
+            self._url = url.rstrip("/")
             self._fs, self.path = fsspec.url_to_fs(url, **storage_options)
         elif hasattr(url, "protocol") and hasattr(url, "fs"):
             # is UPath-like - but without importing
@@ -63,7 +63,7 @@ class RemoteStore(Store):
                     "storage_options are allowed"
                 )
 
-            self._url = str(url)
+            self._url = str(url).rstrip("/")
             self.path = url.path
             self._fs = url.fs
         else:
--- a/src/zarr/store/remote.py
+++ b/src/zarr/store/remote.py
@@ -25,6 +25,7 @@ class RemoteStore(Store):
     supports_listing: bool = True
 
     _fs: AsyncFileSystem
+    _url: str
     path: str
     allowed_exceptions: tuple[type[Exception], ...]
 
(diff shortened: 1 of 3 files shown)


d-v-b:
the test failures were due to a trailing "/" character added to the url / path attributes in the "init-from-UPath" code paths, and that's resolved via `str.rstrip('/')`

src/zarr/store/remote.py
```diff
@@ -51,18 +52,22 @@ def __init__(
         """
 
         super().__init__(mode=mode)
-
         if isinstance(url, str):
-            self._fs, self.path = fsspec.url_to_fs(url, **storage_options)
+            self._url = url.rstrip("/")
+            self._fs, _path = fsspec.url_to_fs(url, **storage_options)
```
martindurant, line 57: Is there a case where url has no trailing "/", but the return fom url_to_fs does?

tests/v3/test_store/test_remote.py
```diff
@@ -2,6 +2,7 @@
 
 import fsspec
 import pytest
+from upath import UPath
```
martindurant, line 5: add to test requirements, or use importorskip in affected tests

tests/v3/test_store/test_remote.py
```diff
@@ -40,18 +41,31 @@ def get_boto3_client():
...
     client = get_boto3_client()
     client.create_bucket(Bucket=test_bucket_name, ACL="public-read")
     s3fs.S3FileSystem.clear_instance_cache()
-    s3 = s3fs.S3FileSystem(anon=False, client_kwargs={"endpoint_url": endpoint_uri})
+    s3 = s3fs.S3FileSystem(anon=False, client_kwargs={"endpoint_url": endpoint_url})
     s3.invalidate_cache()
     yield s3
-    requests.post(f"{endpoint_uri}/moto-api/reset")
+    requests.post(f"{endpoint_url}/moto-api/reset")
+    client.close()
```
martindurant, line 68: +1

tests/v3/test_store/test_remote.py
```diff
@@ -81,31 +95,51 @@ async def test_basic():
...
+    def store_kwargs(self, request) -> dict[str, str | bool]:
+        url = f"s3://{test_bucket_name}"
+        anon = False
+        mode = "w"
+        if request.param == "use_upath":
+            return {"mode": mode, "url": UPath(url, endpoint_url=endpoint_url, anon=anon)}
+        elif request.param == "use_str":
+            return {"url": url, "mode": mode, "anon": anon, "endpoint_url": endpoint_url}
+        else:
+            raise AssertionError
```
martindurant, line 108: not hit

tests/v3/test_store/test_remote.py
```diff
@@ -81,31 +95,51 @@ async def test_basic():
...
+            out = self.store_cls(url=url, mode=mode)
+        else:
+            endpoint_url = store_kwargs["endpoint_url"]
+            out = self.store_cls(url=url, asynchronous=True, mode=mode, endpoint_url=endpoint_url)
         return out
 
     def get(self, store: RemoteStore, key: str) -> Buffer:
-        return Buffer.from_bytes(self._fs.cat(f"{store.path}/{key}"))
+        #  make a new, synchronous instance of the filesystem because this test is run in sync code
+        fs, _ = fsspec.url_to_fs(
```
martindurant, line 123: Seems like going around in a circle here: str->UPath(which contains fs)->str->fs. Shouldn't it be the same as `store.fs`?

src/zarr/store/remote.py
```diff
@@ -71,10 +76,10 @@ def __init__(
             raise TypeError("FileSystem needs to support async operations")
 
     def __str__(self) -> str:
-        return f"Remote fsspec store: {type(self._fs).__name__} , {self.path}"
+        return f"{self._url}"
```
martindurant, line 79: We are implying that this store is the only way to access the URL displayed here

tests/v3/test_store/test_remote.py
d-v-b: unless I'm misunderstanding your concern, the issue with `store._fs` is that it's async, but `StoreTests.get` and `StoreTests.set` are sync, and so using `store._fs` in those methods doesn't work

martindurant: Right of course, sorry.

src/zarr/store/remote.py
d-v-b: yes, but this is consistent with our current practice for [memory store](https://github.com/zarr-developers/zarr-python/blob/f648b619e53cde809cd516f308ff40f206a742bd/src/zarr/store/memory.py#L27) and [local store](https://github.com/zarr-developers/zarr-python/blob/f648b619e53cde809cd516f308ff40f206a742bd/src/zarr/store/local.py#L83). Yes, this practice does mean that `RemoteStore` for local files will have the same __str__ output as `LocalStore`, but I figure we should make the stores consistent before making them correct.

tests/v3/test_store/test_remote.py
d-v-b: despite adding this, I'm still seeing output like this at the end of my tests:

> Task was destroyed but it is pending!
task: <Task pending name='Task-10' coro=<StoreTests.test_get() done, defined at /home/bennettd/dev/zarr-python/src/zarr/testing/store.py:83> wait_for=<Future pending cb=[Task.task_wakeup()]>>
Unclosed client session
client_session: <aiohttp.client.ClientSession object at 0x7f7f53484c70>

src/zarr/store/remote.py
d-v-b: I am not aware of any, but I wanted to ensure that the UPath and non-UPath code paths do the exact same thing to the url string.

d-v-b: I might be misunderstanding the question...

tests/v3/test_store/test_remote.py
martindurant: It's probably client in the async filesystem instance. The full workflow should look like
```
s3 = S3FileSystem(..., asynchronous=True)
session = await s3.set_session()
...
await session.close()
```

martindurant:
None of my comments are blockers

tests/v3/test_store/test_remote.py
d-v-b: added to test requirements

tests/v3/test_store/test_remote.py
d-v-b: thanks for the pointers, those warnings are now gone!

tests/v3/test_store/test_remote.py
d-v-b: this is defensive in case someone changes the `params` argument to the fixture decorator. But I can remove the final `elif` block.

PR #1967, opened 2024-06-14
Initial GPU support

akshaysubr:
Adding an initial implementation to support GPU arrays. Currently limited to only arrays that support the `__cuda_array_interface__` (cupy, numba, pytorch). Can extend this to supporting dlpack as well later for JAX and Tensorflow support.

TODO:


d-v-b:
thanks for working on this @akshaysubr. what would we need to change on the CI side to run gpu tests?

tests/v3/test_store/test_memory.py
```diff
@@ -38,3 +38,36 @@ def test_store_supports_partial_writes(self, store: MemoryStore) -> None:
...
+        assert store.supports_writes
+
+    def test_store_supports_listing(self, store: GpuMemoryStore) -> None:
+        assert store.supports_listing
+
+    def test_store_supports_partial_writes(self, store: GpuMemoryStore) -> None:
+        assert store.supports_partial_writes
+
+    def test_list_prefix(self, store: GpuMemoryStore) -> None:
+        assert True
```
rabernat, line 73: Is there something to actually test here?

rabernat (commented):
Very excited about this! Looks really good! I now understand better why we went to the trouble of creating these different buffer protocols.

Here's some random low-priority feedback based on a first read through.

src/zarr/buffer.py
```diff
@@ -436,6 +441,290 @@ def as_numpy_array_wrapper(
...
+            raise ValueError("array_like: only byte dtype allowed")
+
+        # if type(array_like) == cp.ndarray
+        #     self._data = array_like
+        if hasattr(array_like, "__cuda_array_interface__"):
+            self._data = cp.asarray(array_like)
+        else:
+            # raise ArgumentError("GpuBuffer only supports inputs that implement the '__cuda_array_interface__' protocol")
+            # Slow copy based path for arrays that don't support the __cuda_array_interface__
+            # TODO: Add a fast zero-copy path for arrays that support the dlpack protocol
```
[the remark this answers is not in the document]
akshaysubr: Added a user warning here and ignoring it in `pyproject.toml`

tests/v3/test_store/test_memory.py
akshaysubr: AFAIU, this is here just to silence the xfail in the base `StoreTests` class: https://github.com/zarr-developers/zarr-python/blob/v3/src/zarr/testing/store.py#L175

src/zarr/buffer.py
```diff
@@ -436,6 +441,290 @@ def as_numpy_array_wrapper(
...
+
+    Parameters
+    ----------
+    array_like
+        array-like object that must be 1-dim, contiguous, and byte dtype.
+    """
+
+    def __init__(self, array_like: ArrayLike):
+        if cp is None:
+            raise RuntimeError("Cannot use GpuBuffer without cupy")
```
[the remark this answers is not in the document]
akshaysubr: Made this change

src/zarr/buffer.py
```diff
@@ -436,6 +441,290 @@ def as_numpy_array_wrapper(
...
+
+    Parameters
+    ----------
+    ndarray_like
+        ndarray-like object that is convertible to a regular Numpy array.
+    """
+
+    def __init__(self, array: NDArrayLike):
+        if cp is None:
+            raise RuntimeError("Cannot use GpuNDBuffer without cupy")
```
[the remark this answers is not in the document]
akshaysubr: Made this change

akshaysubr:
@d-v-b For CI, a GPU runner would be needed and the pipeline should install the optional `[gpu]` dependencies (just cupy).

rabernat:
I would have thought that this line

https://github.com/zarr-developers/zarr-python/blob/143faeaf9007657cd688ea8afa2cc62e1135ae2b/.github/workflows/test.yml#L26

Would have picked up this new optional dependency

https://github.com/zarr-developers/zarr-python/blob/fe8591a877a7acd636f65d1f1c208254ba2095a6/pyproject.toml#L63-L65

But that didn't happen in this CI run: https://github.com/zarr-developers/zarr-python/actions/runs/9555282123/job/26338638922?pr=1967#step:5:1

src/zarr/store/memory.py
```diff
@@ -101,3 +101,63 @@ async def list_dir(self, prefix: str) -> AsyncGenerator[str, None]:
             for key in self._store_dict:
                 if key.startswith(prefix + "/") and key != prefix:
                     yield key.removeprefix(prefix + "/").split("/")[0]
+
+
+class GpuMemoryStore(MemoryStore):
```
madsbk, line 106: Maybe rename the class to something like `GpuOnlyMemoryStore` to indicate that this stores everything on the GPU no matter the original location. And point out in a docstring to use `MemoryStore` if you are location agnostic.

src/zarr/store/memory.py
```diff
@@ -101,3 +101,63 @@ async def list_dir(self, prefix: str) -> AsyncGenerator[str, None]:
...
+        key_ranges: list[tuple[str, tuple[int | None, int | None]]],
+    ) -> list[Buffer | None]:
+        # All the key-ranges arguments goes with the same prototype
+        async def _get(key: str, byte_range: tuple[int, int | None]) -> Buffer | None:
+            # Q: use prototype here to convert to bespoke buffer class? If so, how?
+            return await self.get(key, prototype=prototype, byte_range=byte_range)
+
+        vals = await concurrent_map(key_ranges, _get, limit=None)
+        return vals
+
```
madsbk, line 149: With #1974, `get` and `get_partial_values` from `MemoryStore` should automatically return the buffer type specified in the `prototype` argument.

```suggestion
```

madsbk (changes requested):
Looks good, I only have minor suggestions.

But I wonder if we should move all of this to its own file: `gpu_buffer.py` ?

d-v-b:
> But I wonder if we should move all of this to its own file: `gpu_buffer.py` ?

I could imagine something like this:
```
src/zarr/buffer 
|- gpu.py
|   |- Buffer
|- cpu.py
|   |- Buffer
```

leading to usage like 
```python
from zarr.buffer import cpu, gpu
...
cpu.Buffer()
...
gpu.Buffer()
```

src/zarr/buffer.py
```diff
@@ -436,6 +442,311 @@ def as_numpy_array_wrapper(
...
+        Parameters
+        ----------
+        array_like
+            array-like object that must be 1-dim, contiguous, and byte dtype.
+
+        Returns
+        -------
+            New buffer representing `array_like`
+        """
+        return cls(array_like)
```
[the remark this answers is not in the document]
akshaysubr: Agree with this for `from_array_like`. However, `from_buffer` is not in the parent `Buffer` class. Maybe good to move this to the parent class then but have it default to go through numpy arrays?
```python
    @classmethod
    def from_buffer(cls, buffer: Buffer) -> Self:
        """Create a Buffer given another arbitrary Buffer
        """
        return cls(buffer.as_numpy_array())
```

src/zarr/buffer.py
```diff
@@ -436,6 +442,311 @@ def as_numpy_array_wrapper(
...
+                "Creating a GpuBuffer with an array that does not support the "
+                "__cuda_array_interface__ for zero-copy transfers, "
+                "falling back to slow copy based path"
+            )
+            warnings.warn(
+                msg,
+                stacklevel=2,
+            )
+            buffer = Buffer(array_like)
+            self._data = cp.asarray(buffer.as_numpy_array())
```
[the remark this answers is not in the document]
akshaysubr: Made this change. But we'd still need a separate path for DLPack right? Can probably deal with that in a later PR since cupy, torch and jax should all be covered by this.

src/zarr/buffer.py
```diff
@@ -436,6 +442,311 @@ def as_numpy_array_wrapper(
...
+
+        Warning
+        -------
+        Might have to copy data, consider using `.as_array_like()` instead.
+
+        Returns
+        -------
+            NumPy array of this buffer (might be a data copy)
+        """
+        from typing import cast
```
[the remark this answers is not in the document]
akshaysubr: Done

src/zarr/store/memory.py
akshaysubr: This is a good point. Maybe worth adding some tests against a non-default prototype to ensure this is properly happening for all stores?

Btw, in #1974, how would the `from_buffer` work if you pass in a `GpuBuffer`? Won't that just set `_data` to the underlying cupy array and cause downstream accessors that are expecting a numpy array to fail?

src/zarr/store/memory.py
akshaysubr: Added docs about this guaranteeing storage in GPU memory and added the location agnostic clause as well. But left the name as is for simplicity.

src/zarr/buffer.py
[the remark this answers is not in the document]
akshaysubr: Sorry, I missed the previous PR that added `from_buffer` to the parent class. However, I feel like the `Buffer` class' `from_buffer` should use `as_numpy_array` instead of `as_array_like`. Otherwise, if you use the `GpuMemoryStore` or any other component that outputs a `GpuBuffer` and you try to coerce it to a regular buffer using `default_prototype.buffer.from_buffer(gpu_buffer)`, that will create a new `Buffer` object with `_data` being a cupy array. A downstream CPU compressor will then call `.as_numpy_array` on this Buffer class and will see an error similar to the snippet below right?
```python
>>> np.asanyarray(cp.random.random(5))
Cell In[3], line 1
----> 1 np.asanyarray(cp.random.random(5))

File cupy/_core/core.pyx:1479, in cupy._core.core._ndarray_base.__array__()

TypeError: Implicit conversion to a NumPy array is not allowed. Please use `.get()` to construct a NumPy array explicitly.

```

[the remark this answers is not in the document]
madsbk: Good point, I guess the underlying problem is that `Buffer` requires that `ArrayLike` supports `np.asanyarray()`.

We might consider adding some some kind of `to-numpy-function/-method` to the `ArrayLike` protocol in a future PR.

src/zarr/store/memory.py
```diff
@@ -101,3 +101,62 @@ async def list_dir(self, prefix: str) -> AsyncGenerator[str, None]:
...
+        prototype: BufferPrototype,
+        key_ranges: list[tuple[str, tuple[int | None, int | None]]],
+    ) -> list[Buffer | None]:
+        # All the key-ranges arguments goes with the same prototype
+        async def _get(key: str, byte_range: tuple[int, int | None]) -> Buffer | None:
+            return await self.get(key, prototype=prototype, byte_range=byte_range)
+
+        vals = await concurrent_map(key_ranges, _get, limit=None)
+        return vals
+
```
madsbk, line 153:
```suggestion
```

madsbk (approved):
Looks good, nice work. 
I have a minor suggestion and I think the gpu buffer should get its own file like @d-v-b suggest: https://github.com/zarr-developers/zarr-python/pull/1967#issuecomment-2175563354

src/zarr/store/memory.py
akshaysubr: Removed this and also `get` now. Left this in until I got your thoughts on the `from_buffer` issue.

src/zarr/buffer.py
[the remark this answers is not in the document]
akshaysubr: Agreed, we can add some functionality like that in the future. For now, I'll switch the base class to use `.as_numpy_array` which is equivalent here but allows interop. Also added a docstring note that subclassed of `Buffer` should implement more optimal `from_buffer` methods that can avoid copying when possible.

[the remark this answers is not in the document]
madsbk: Looks good

akshaysubr:
@d-v-b Thanks for the suggestion to refactor into two separate files. I made those changes and added some more that I think benefit the overall usage of these buffer classes. Essentially now, `zarr.buffer.Buffer` is an abstract class and `zarr.buffer.cpu.Buffer` and `zarr.buffer.gpu.Buffer` are concrete implementations which makes things cleaner. It also requires implementations to be specific about which type of `Buffer` they intend on using which I think is a good thing.

src/zarr/buffer/core.py
```diff
@@ -179,10 +182,16 @@ def from_buffer(cls, buffer: Buffer) -> Self:
...
+        Note
+        ----
+        Subclasses of `Buffer` must override this method to implement
+        more optimal conversions that avoid copies where possible
         """
-        return cls.from_array_like(buffer.as_array_like())
+        ...
 
     @classmethod
+    @abstractmethod
```
d-v-b, line 194: test failures are due to this method becoming abstract, while it still gets called in `to_buffer_dict` in `metadata.py`. Two things should change to fix this: first, I think maybe the order of the decorators here should be flipped to ensure that `Buffer.from_bytes` can't be called without an exception, and second `metadata.py` needs to not be calling any `Buffer` methods.

akshaysubr: I'm not sure we can flip the order of decorators here based on this snippet from the `abstractmethod` [docs](https://docs.python.org/3/library/abc.html#abc.abstractmethod):
> When abstractmethod() is applied in combination with other method descriptors, it should be applied as the innermost decorator

I agree though that `metadata.py` shouldn't be calling any `Buffer` methods and should instead be calling `prototype.buffer` methods or `default_buffer_prototype.buffer` methods. Is there a preference between propagating a `prototype` argument up the call stack or just using `default_buffer_prototype` since these `to_bytes` calls are not in the critical performance path?

akshaysubr: Resolved the CI issues by tracking all calls to abstract classmethods of `Buffer` and `NDBuffer`. Here are the main ones and the current solution:
- `metadata.py`: switch to using `default_buffer_prototype.buffer`
- `group.py`: switch to using `default_buffer_prototype.buffer`
- `sharding.py`: switch to using `default_buffer_prototype.buffer`
- `codecs/_v2.py`: switch to using `cpu.Buffer` and `cpu.NDBuffer` since these are all explicitly CPU only.

.github/workflows/gpu_test.yml
```diff
@@ -0,0 +1,48 @@
...
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  test:
+    name: py=${{ matrix.python-version }}, np=${{ matrix.numpy-version }}, deps=${{ matrix.dependency-set }}
+
+    runs-on: "cirun-gpu-runner--${{ github.run_id }}"
```
jhamman, line 21:
```suggestion
    runs-on: "gpu-runner"
```

I am not 100% sure but I think changing to this will have this action using the github actions GPU runner that @rabernat and I just configured.

akshaysubr: Just reverted the cirun commits and added a github CI workflow for GPU tests. Also tried to make the GPU workflow only run GPU tests. Can certainly optimize the test matrix more for cost.

jhamman:
@akshaysubr and @madsbk - we'd love to get this in. There's now a good list of (hopefully small) conflicts to resolve. Can we get this cleaned up this week?

akshaysubr:
@jhamman Working on fixing these conflicts. Unfortunately, it's a bit messy, but will get this cleaned up this week.

.github/workflows/gpu_test.yml
```diff
@@ -0,0 +1,48 @@
...
+jobs:
+  test:
+    name: py=${{ matrix.python-version }}, np=${{ matrix.numpy-version }}, deps=${{ matrix.dependency-set }}
+
+    runs-on: gpu-runner
+    strategy:
+      matrix:
+        python-version: ['3.10', '3.11', '3.12']
+        numpy-version: ['1.24', '1.26', '2.0']
+        dependency-set: ["minimal", "optional"]
```
akshaysubr, line 26: @jhamman What test matrix do we want to have for GPU testing? This current config seems a bit excessive? Might be worth cutting this down to the bare minimum to keep CI costs down?

.github/workflows/gpu_test.yml
```diff
@@ -0,0 +1,48 @@
...
+        nvidia-smi
+    - name: Set up Python
+      uses: actions/setup-python@v5
+      with:
+        python-version: ${{ matrix.python-version }}
+        cache: 'pip'
+    - name: Install Hatch and CuPy
+      run: |
+        python -m pip install --upgrade pip 
+        pip install hatch cupy-cuda12x
```
akshaysubr, line 41: Manually adding this here for now. Should make this part of the hatch workflow later.

.github/workflows/gpu_test.yml
jhamman, line 26: what if we did something like this for now: 
```suggestion
        python-version: ['3.11']
        numpy-version: ['2.0']
        dependency-set: ["gpu"]
```

akshaysubr: This would be ideal. There is currently a `gpu` dependency set in `pyproject.toml`, but I'm not sure how to get hatch to pick it up.

.github/workflows/gpu_test.yml
jakirkham: Yeah there's probably a better way to handle this. Happy to chat when this is ready for discussion

pyproject.toml
```diff
@@ -74,6 +74,9 @@ jupyter = [
     'ipytree>=0.2.2',
     'ipywidgets>=8.0.0',
 ]
+gpu = [
+    "cupy-cuda12x",
+]
```
jakirkham, line 79:
```suggestion
cuda12 = [
    "cupy-cuda12x",
]
```

pyproject.toml
```diff
@@ -120,9 +123,9 @@ build.hooks.vcs.version-file = "src/zarr/_version.py"
 [tool.hatch.envs.test]
 dependencies = [
     "numpy~={matrix:numpy}",
-    "universal_pathlib"
+    "universal_pathlib",
 ]
-features = ["test", "extra"]
+features = ["test", "extra", "gpu"]
```
jakirkham, line 128:
```suggestion
features = ["test", "extra", "cuda12"]
```

pyproject.toml
```diff
@@ -134,8 +137,14 @@ python = ["3.10", "3.11", "3.12"]
 numpy = ["1.24", "1.26", "2.0"]
 features = ["optional"]
 
+[[tool.hatch.envs.test.matrix]]
+python = ["3.10", "3.11", "3.12"]
+numpy = ["1.24", "1.26", "2.0"]
+features = ["gpu"]
```
jakirkham, line 143:
```suggestion
features = ["cuda12"]
```

.github/workflows/gpu_test.yml
akshaysubr: This is the only thing remaining for this PR, so any solutions here would be great!

pyproject.toml
```diff
@@ -223,4 +232,8 @@ filterwarnings = [
     "error:::zarr.*",
     "ignore:PY_SSIZE_T_CLEAN will be required.*:DeprecationWarning",
     "ignore:The loop argument is deprecated since Python 3.8.*:DeprecationWarning",
+    "ignore:Creating a zarr.buffer.gpu.*:UserWarning",
+]
+markers = [
+    "gpu: mark a test as requiring CuPy and GPU"
```
jakirkham, line 238: Make it is worth using `cuda` here?

jakirkham (commented):
Thanks Akshay! 🙏

Think it would be worthwhile to put these under `cuda12`

Idk if we want CUDA 11 at this stage (probably not given this is new and CUDA 11 is getting older). Though do think putting CUDA 12 dependencies under that name will make it easier to upgrade in the future

If we do use this naming, we may want to reflect something similar (`cuda` for when version doesn't matter and `cuda12` where it does)

.github/workflows/gpu_test.yml
jakirkham: Great, did an initial pass below

We should be able to use [the self-referential extra with hatch]( https://hatch.pypa.io/1.12/config/dependency/#self-referential ) as well

.github/workflows/gpu_test.yml
```diff
@@ -0,0 +1,57 @@
...
+    - uses: actions/checkout@v4
+    - name: cuda-toolkit
+      uses: Jimver/cuda-toolkit@v0.2.16
+      id: cuda-toolkit
+      with:
+        cuda: '12.5.0'
+      run: |
+        echo "Installed cuda version is: ${{steps.cuda-toolkit.outputs.cuda}}"
+        echo "Cuda install location: ${{steps.cuda-toolkit.outputs.CUDA_PATH}}"
+        nvcc -V
```
jhamman, line 38: I'm seeing a github action lint error from this section: 

```
[Error](https://github.com/zarr-developers/zarr-python/actions/runs/10569457903/workflow)
a step cannot have both the `uses` and `run` keys
```

jhamman:
@akshaysubr - I'm keen to get this in today or tomorrow if possible. Seems like we should just ignore the pre-commit errors once the ci issue noted above is addressed.

.github/workflows/gpu_test.yml
akshaysubr: Pushed a change that should hopefully fix this. Moved the run portion to the GPU check step instead.

jhamman (approved):

pyproject.toml
joshmoore: You're thinking, @jakirkham, that then there could be a multiplicity of these.

PR #1968, opened 2024-06-15
Fix indexing with bools

### LINKED ISSUE #1963: [v3] Accessing data with bools in square bracket notation
### Zarr version

v3

### Numcodecs version

0.12.1

### Python Version

3.12

### Operating System

Linux

### Description

```
ix = np.random.binomial(1, 0.6, size=z.shape[0]).astype(bool)
print(z[ix])
```

will result in  `IndexError( "unsupported selection item for basic indexing; expected integer or slice, got <class 'numpy.ndarray'>)`

while `z.get_orthogonal_selection(ix) `  works.

I suppose the problem is within `Array.__getitem__`

### Steps to reproduce

```
data = np.arange(10)
z = Array.create(
            StorePath(MemoryStore(mode="w")),
            shape=data.shape,
            chunk_shape=(2,),
            dtype=data.dtype,
            )
z[:] = data
ix = np.random.binomial(1, 0.6, size=z.shape[0]).astype(bool)
print(z.get_orthogonal_selection(ix)) #works
print(z[ix]) #fails
```

### Additional output

_No response_

brokkoli71:
fixes #1963

**Changes:**
- Mask indexing is available as indexing with square brackets: `a[mask_selection]` 
- Orthogonal indexing with bool lists is available as indexing with square brackets: `a[bool_selection_axis1, bool_selection_axis2]` 

**Example**
For an Array
```
data = np.arange(9, dtype="uint16").reshape(3, 3)
z = Array.create(
        StorePath(MemoryStore(mode="w")),
        shape=data.shape,
        chunk_shape=(data.shape),
        dtype=data.dtype,
    )
z[:] = data
mask = np.tile([True, False, True], (3, 1))

```
Each of the following failed before and works now
```
print(z[mask])
z[mask] = 42
print(z[[True, False, True], [True, False, True]])
z[[True, False, True], [True, False, True]] = 10
```


src/zarr/indexing.py
```diff
@@ -118,6 +120,16 @@ def is_integer_list(x: Any) -> TypeGuard[list[int]]:
     return isinstance(x, list) and len(x) > 0 and is_integer(x[0])
 
 
+def is_bool_list(x: Any) -> TypeGuard[list[int]]:
+    """True if x is a list of boolean.
+
+    This function assumes ie *does not check* that all elements of the list
+    have the same type. Mixed type lists will result in other errors that will
+    bubble up anyway.
```
jhamman, line 128: curious if we need to make this optimization at this point.

src/zarr/indexing.py
```diff
@@ -118,6 +120,16 @@ def is_integer_list(x: Any) -> TypeGuard[list[int]]:
     return isinstance(x, list) and len(x) > 0 and is_integer(x[0])
 
 
+def is_bool_list(x: Any) -> TypeGuard[list[int]]:
```
jhamman, line 123:
```suggestion
def is_bool_list(x: Any) -> TypeGuard[list[bool]]:
```

src/zarr/indexing.py
```diff
@@ -101,11 +101,13 @@ def ceildiv(a: float, b: float) -> int:
...
 def is_integer(x: Any) -> TypeGuard[int]:
-    """True if x is an integer (both pure Python or NumPy).
+    """True if x is an integer (both pure Python or NumPy)."""
+    return isinstance(x, numbers.Integral) and not is_bool(x)
 
-    Note that Python's bool is considered an integer too.
-    """
-    return isinstance(x, numbers.Integral)
+
+def is_bool(x: Any) -> TypeGuard[int]:
```
jhamman, line 108:
```suggestion
def is_bool(x: Any) -> TypeGuard[bool]:
```

jhamman (commented):
A few quick comments.

src/zarr/indexing.py
brokkoli71: probably not. I have tested performance differences of accessing with mask selection using boolean lists and I had a loss of performance only noticable from size 10^5 onwards (50% slower at 10^5 and 400% slower at 10^6)
But at this point you should probably not use numpy arrays instead of python list anyway

tests/v3/test_indexing.py
```diff
@@ -1718,3 +1721,51 @@ def test_accessed_chunks(shape, chunks, ops):
...
+        [np.full(1000, False), np.tile([True, False], 5)],
+        [np.full(1000, True), np.full(10, True)],
+        [np.full(1000, True), [True, False] * 5],
+    ],
+)
+def test_orthogonal_bool_indexing_like_numpy_ix(store, selection):
+    a = np.arange(10000, dtype=int).reshape(1000, 10)
+    z = zarr_array_from_numpy_array(store, a, chunk_shape=(300, 3))
+    expected = a[np.ix_(*selection)]
+    # note: in python 3.10 z[*selection] is no valid unpacking syntax
```
d-v-b, line 1769:
```suggestion
    # note: in python 3.10 z[*selection] is not valid unpacking syntax
```

tests/v3/test_indexing.py
```diff
@@ -1718,3 +1721,51 @@ def test_accessed_chunks(shape, chunks, ops):
...
+        [np.full((1000, 10), False)],
+        # coordinate selection
+        [[1, 2, 3, 4], [5, 6, 7, 8]],
+        [[100, 200, 300], [4, 5, 6]],
+    ],
+)
+def test_indexing_equals_numpy(store, selection):
+    a = np.arange(10000, dtype=int).reshape(1000, 10)
+    z = zarr_array_from_numpy_array(store, a, chunk_shape=(300, 3))
+    # note: in python 3.10 a[*selection] is no valid unpacking syntax
```
d-v-b, line 1750:
```suggestion
    # note: in python 3.10 a[*selection] is not valid unpacking syntax
```

normanrz (approved):

Issue #1819, opened 2024-04-26
7 tests fail


yurivict:
### Zarr version

2.17.2

### Numcodecs version

Numcodecs

### Python Version

3.9

### Operating System

FreeBSD 14

### Installation

port

### Description

```
========================================================================================= FAILURES =========================================================================================
_____________________________________________________________________________ TestNestedFSStore.test_hierarchy _____________________________________________________________________________
[gw2] freebsd14 -- Python 3.9.18 /usr/local/bin/python3.9

self = <zarr.tests.test_storage.TestNestedFSStore object at 0x1453af27310>

    def test_hierarchy(self):
        # setup
        store = self.create_store()
        store[self.root + "a"] = b"aaa"
        store[self.root + "b"] = b"bbb"
        store[self.root + "c/d"] = b"ddd"
        store[self.root + "c/e/f"] = b"fff"
        store[self.root + "c/e/g"] = b"ggg"
    
        # check keys
        assert self.root + "a" in store
        assert self.root + "b" in store
        assert self.root + "c/d" in store
        assert self.root + "c/e/f" in store
        assert self.root + "c/e/g" in store
        assert self.root + "c" not in store
        assert self.root + "c/" not in store
        assert self.root + "c/e" not in store
        assert self.root + "c/e/" not in store
        assert self.root + "c/d/x" not in store
    
        # check __getitem__
        with pytest.raises(KeyError):
            store[self.root + "c"]
        with pytest.raises(KeyError):
            store[self.root + "c/e"]
        with pytest.raises(KeyError):
            store[self.root + "c/d/x"]
    
        # test getsize (optional)
        if hasattr(store, "getsize"):
            # TODO: proper behavior of getsize?
            #       v3 returns size of all nested arrays, not just the
            #       size of the arrays in the current folder.
            if self.version == 2:
                assert 6 == store.getsize()
            else:
                assert 15 == store.getsize()
            assert 3 == store.getsize("a")
            assert 3 == store.getsize("b")
            if self.version == 2:
                assert 3 == store.getsize("c")
            else:
                assert 9 == store.getsize("c")
            assert 3 == store.getsize("c/d")
            assert 6 == store.getsize("c/e")
            assert 3 == store.getsize("c/e/f")
            assert 3 == store.getsize("c/e/g")
            # non-existent paths
            assert 0 == store.getsize("x")
            assert 0 == store.getsize("a/x")
            assert 0 == store.getsize("c/x")
[...]

Issue #1679, opened 2024-02-27
Hierarchy test fail with fsspec 2024.2.0


QuLogic:
### Zarr version

2.17.0

### Numcodecs version

0.12.1

### Python Version

3.12.1

### Operating System

Fedora 39/40/Rawhide

### Installation

pip into virtual environment; also from source

### Description

When testing against fsspec 2024.2.0, some hierarchy tests fail. These tests passed with 2023.12.2 (the previous release).

### Steps to reproduce

```
$ python -m venv venv
$ . venv/bin/activate
$ pip install pytest zarr
$ pip install fsspec==2023.12.2
$ pytest --pyargs zarr.tests.test_storage  # All tests pass.
```
but then with fsspec 2024.2.0:
```
$ pip install fsspec==2024.2.0
$ pytest --pyargs zarr.tests.test_storage
================================================================================================================================ short test summary info =================================================================================================================================
FAILED venv/lib64/python3.12/site-packages/zarr/tests/test_storage.py::TestFSStore::test_hierarchy - AssertionError: assert [] == ['f']
FAILED venv/lib64/python3.12/site-packages/zarr/tests/test_storage.py::TestFSStoreWithKeySeparator::test_hierarchy - AssertionError: assert [] == ['f']
FAILED venv/lib64/python3.12/site-packages/zarr/tests/test_storage.py::TestFSStoreFromFilesystem::test_hierarchy - AssertionError: assert [] == ['f']
FAILED venv/lib64/python3.12/site-packages/zarr/tests/test_storage.py::TestN5FSStore::test_hierarchy - AssertionError: assert [] == ['f']
FAILED venv/lib64/python3.12/site-packages/zarr/tests/test_storage.py::TestNestedFSStore::test_hierarchy - AssertionError: assert [] == ['f']
================================================================================================================= 5 failed, 581 passed, 137 skipped, 1 warning in 1.67s ==================================================================================================================
```

### Additional output

<details>
<summary>Full failure traceback output:</summary>

```
_______________________________________________________________________________________________________________________________ TestFSStore.test_hierarchy _______________________________________________________________________________________________________________________________

self = <zarr.tests.test_storage.TestFSStore object at 0x7f938a6bc8c0>

    def test_hierarchy(self):
        # setup
        store = self.create_store()
        store[self.root + "a"] = b"aaa"
        store[self.root + "b"] = b"bbb"
        store[self.root + "c/d"] = b"ddd"
[...]

PR #1970, opened 2024-06-17
Adapt storage tests for changes in fsspec (#1819) (#1679)

AdamWill:
This is adapted from the fixes that were rolled into https://github.com/zarr-developers/zarr-python/pull/1785 for the v3 branch.

TODO:

Deleted TODO items are not applicable as this changes no code, only tests.


AdamWill:
On v3 the check in `test_hierarchy` was commented out entirely, but that seems excessive to me. The comment indicates the primary point of these checks is to make sure they run without causing an assertion, and we can still do that, and check that the result is one of the two we know are possible.

dstansby:
Thanks for this! To get the tests passing, I think the version of fsspec in `requirements_dev_optional.txt` will need increasing to a version that's compatible with the new tests.

AdamWill:
ugh, it's been a while, but I *think* the change is supposed to work with both old and new fsspec? I'll try and find a minute to remember it...

jhamman:

AdamWill:
yeah, looking at it again, this is intended to work both ways.

AdamWill:
eh, but looking at the test results, it didn't. not sure why not, i'd have to remember all this stuff again. sigh.

PR #1974, opened 2024-06-18
[v3] `Buffer` ensure correct subclass based on the `BufferPrototype` argument

madsbk:
TODO:


src/zarr/buffer.py
```diff
@@ -157,6 +157,29 @@ def from_array_like(cls, array_like: ArrayLike) -> Self:
...
+    @classmethod
+    def from_buffer(cls, buffer: Buffer) -> Self:
+        """Create a new buffer of an existing Buffer
+
+        This is useful if you want to ensure that an existing buffer is
+        of the correct subclass of Buffer. E.g., MemoryStore use this
+        to return a buffer instance of the subclass specified by its
+        BufferPrototype argument.
+
+        Typically, this only copy data if the data has to be moved between
```
d-v-b, line 169:
```suggestion
        Typically, this only copies data if the data has to be moved between
```

src/zarr/buffer.py
```diff
@@ -157,6 +157,29 @@ def from_array_like(cls, array_like: ArrayLike) -> Self:
         """
         return cls(array_like)
 
+    @classmethod
+    def from_buffer(cls, buffer: Buffer) -> Self:
+        """Create a new buffer of an existing Buffer
+
+        This is useful if you want to ensure that an existing buffer is
+        of the correct subclass of Buffer. E.g., MemoryStore use this
```
d-v-b, line 165:
```suggestion
        of the correct subclass of Buffer. E.g., MemoryStore uses this
```

src/zarr/buffer.py
```diff
@@ -157,6 +157,29 @@ def from_array_like(cls, array_like: ArrayLike) -> Self:
...
+        memory types, such as from host to device memory.
+
+        Parameters
+        ----------
+        buffer
+            buffer object.
+
+        Returns
+        -------
+            New buffer representing buffer
```
d-v-b, line 181:
```suggestion
            A new buffer representing the content of the input buffer
```

d-v-b:
I have a broader question about the relationship between the store API and different flavors of buffer (This question is out of scope for this PR, but the answer would help me interpret the effort here).

Taking `MemoryStore.get` as an example, we have 

```python
async def get(
        self,
        key: str,
        prototype: BufferPrototype,
        byte_range: tuple[int | None, int | None] | None = None,
    ) -> Buffer | None:
```
`key` and `byte` range are parametrizing the resource that the store is going to get; we can think of this as basically a query against a `{keys: seekable_bytes}` map-abstraction of a storage backend. But `prototype` is different -- it's determining the concrete return type of `get`. 

Do we expect arrays / groups to dynamically use `prototype` parameter? For example, is the same array instance going to call `store.get(prototype=GPUBuffer)` for some chunks, and `store.get(prototype=CPUBuffer)` for other chunks? Alternatively, we might want to constrain a store instance to always use a fixed buffer type, in which case maybe the `prototype` should be an attribute of the store class? Sorry if I missed a discussion that explains the design here.

madsbk:
> Do we expect arrays / groups to dynamically use `prototype` parameter? For example, is the same array instance going to call `store.get(prototype=GPUBuffer)` for some chunks, and `store.get(prototype=CPUBuffer)` for other chunks? 

Yes, e.g. for GPU backed arrays we want the metadata to use `CPUBuffer` and the data to use `GPUBuffer`. Furthermore, the end-user might also want to load some part of the array into a numpy array and other parts into a cupy array.

d-v-b:
thanks! that's very helpful

madsbk:
Thanks @d-v-b !

PR #1979, opened 2024-06-18
`parse_shapelike` allows 0

d-v-b:
- `parse_shapelike` allows shape values to be 0
- `parse_shapelike` allows the empty tuple `()`
- `parse_shapelike` rejects negative integers correctly (in `main` this is not working)
- expands tests accordingly

response to #1976 and #1977, although I'm not sure it addresses those concerns entirely, since I haven't changed array indexing tests at all

TODO:
* [ ] Add unit tests and/or doctests in docstrings
* [ ] Add docstrings and API docs for any new/modified user-facing classes and functions
* [ ] New/modified features documented in docs/tutorial.rst
* [ ] Changes documented in docs/release.rst
* [ ] GitHub Actions have all passed
* [ ] Test coverage is 100% (Codecov passes)

Issue #1976, opened 2024-06-18
[v3] Support zero-dimensional arrays


tomwhite:
### Zarr version

3.0.0a1

### Numcodecs version

0.12.1

### Python Version

3.11.9

### Operating System

Mac

### Installation

pip

### Description

Zero-dimensional arrays are supported in v2, so they should be in v3 too.

### Steps to reproduce

```python
>>> import zarr.v2 as zarr
>>> z = zarr.open(store='example-v2.zarr', mode='w', shape=())
>>> z[...] = 2
>>> z[...]
array(2.)
>>> import zarr
>>> z = zarr.open(store='example-v3.zarr', mode='w', shape=())
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/tom/workspace/zarr-python/src/zarr/api/synchronous.py", line 45, in open
    obj = sync(
          ^^^^^
  File "/Users/tom/workspace/zarr-python/src/zarr/sync.py", line 92, in sync
    raise return_result
  File "/Users/tom/workspace/zarr-python/src/zarr/sync.py", line 51, in _runner
    return await coro
           ^^^^^^^^^^
  File "/Users/tom/workspace/zarr-python/src/zarr/api/asynchronous.py", line 198, in open
    return await open_array(store=store_path, zarr_format=zarr_format, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/tom/workspace/zarr-python/src/zarr/api/asynchronous.py", line 868, in open_array
    return await create(store=store, path=path, zarr_format=zarr_format, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/tom/workspace/zarr-python/src/zarr/api/asynchronous.py", line 686, in create
    return await AsyncArray.create(
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/tom/workspace/zarr-python/src/zarr/array.py", line 153, in create
    return await cls._create_v3(
           ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/tom/workspace/zarr-python/src/zarr/array.py", line 237, in _create_v3
[...]

PR #1980, opened 2024-06-18
0 dim arrays: indexing

d-v-b:
Support (basic) indexing of 0 dimensional arrays. I parameterized the old v2 test for this w.r.t. the `out` keyword and also the `dtype` of the array, marking the structured dtype case as an xfail. Builds on #1979.

@normanrz git blame tells me that you [marked the 0d indexing tests as xfail](https://github.com/zarr-developers/zarr-python/blame/143faeaf9007657cd688ea8afa2cc62e1135ae2b/tests/v3/test_indexing.py#L469), with the reasoning being that v3 does not support 0d arrays. I couldn't find anything in the v3 spec about 0d arrays (and I actually couldn't find anything in the v2 spec about it either!). Did I miss something in the spec? If not, we might want to make this explicit one way or another.

cc @TomWhite if you have the time to test this I would appreciate it. It's possible that there are other broken things that I haven't found via our test suite.

TODO:


tomwhite:
Thanks for working on this @d-v-b!

I tried running the example from #1976, and it gets further, but doesn't seem to write the value:

```python
>>> import zarr
>>> z = zarr.open(store='example-v3.zarr', mode='w', shape=())
>>> z[...] = 3
>>> z[...]
array(1.5e-323)
```

> Did I miss something in the spec? If not, we might want to make this explicit one way or another.

I agree that the spec should say something about this case. 0-d arrays are still arrays, and should be supported in Zarr v3 IMO.

normanrz:
The spec says that the `shape` needs to be an array of integers. I think this includes empty arrays (ie. 0d arrays). I think it is fine to add support. I would have expected more things to break.

How does chunk key encoding work for 0d arrays?

d-v-b:
> How does chunk key encoding work for 0d arrays?

@normanrz I was wondering the same thing. It turns out that I _did_ miss [something](https://github.com/zarr-developers/zarr-specs/blob/daef2bc91a246d24d895276679b6ac9556d7ebb1/docs/v3/core/v3.0.rst?plain=1#L539-L541) in the v3 spec: in the section describing the default chunk key encoding, it states:
> Arrays may have 0 dimensions (when for example representing scalars), in which case the coordinate of a chunk is the empty tuple, and the chunk key will consist of the string c.  

I wonder if this is a typo, and the intention was to use the v2 behavior with a single chunk at `c/0` or `c.0`, depending on the separator metadata? Seems that there's already an [open issue](https://github.com/zarr-developers/zarr-specs/issues/282) about this, which I had completely forgotten about.

normanrz:
> > Arrays may have 0 dimensions (when for example representing scalars), in which case the coordinate of a chunk is the empty tuple, and the chunk key will consist of the string c.
> 
> I wonder if this is a typo, and the intention was to use the v2 behavior with a single chunk at `c/0` or `c.0`, depending on the separator metadata? Seems that there's already an [open issue](https://github.com/zarr-developers/zarr-specs/issues/282) about this, which I had completely forgotten about.

Reading it now, I think the spec pretty clear. For `default`, the chunk key will be `c`. `v2` is more tricky intuitively. But the [spec explicitly sets](https://zarr-specs.readthedocs.io/en/latest/v3/core/v3.0.html#chunk-key-encoding) it to `0` for all 0d arrays. zarrita also implements it like that: https://github.com/scalableminds/zarrita/blob/async/zarrita/metadata.py#L138-L139

normanrz (approved):

d-v-b:
I ended up needing to implement `__array__` on the `Array` class, and I fixed a bug in `Array._set_basic_selection` wherein scalars were not being cast to the correct dtype.

normanrz (approved):

PR #1982, opened 2024-06-20
Use config to select implementation

### LINKED ISSUE #1878: Use config to select CodecPipeline and Codec implementations
We want users to be able to choose their own codec pipeline and codec implementations. We need to use the config system to make that configurable. For that, the codec registry needs to deal with multiple codec implementations for the same name.

brokkoli71:
fixes #1878

Using the config (https://github.com/pytroll/donfig), the user can specify now the implementation of all **codecs**, the **CodecPipeline**, **Buffer** and **NDBuffer**.
For each of these objects, the codec registry can deal with multiple different implementations and will use the one selected by the config.

Further changes:
- All calls on classes Buffer and NDBuffer now get called on selected Implementation 
- Registry was expanded to register codec-pipelines, buffers and ndbuffers
- Moved registry.py from `zarr.codecs.registry` to `zarr.registry`


normanrz:
@madsbk I was wondering what you think about overriding the default_buffer_prototype via config. Do you think that is a good idea or unneccessary?

madsbk:
> @madsbk I was wondering what you think about overriding the default_buffer_prototype via config. Do you think that is a good idea or unneccessary?

I think it would a good idea, or maybe add a default attribute to each `AsyncArray` instance, which would be using the config value if not specified when creating the array?

brokkoli71:
@madsbk
> I think it would a good idea, or maybe add a default attribute to each `AsyncArray` instance, which would be set to the config value if not specified when creating the array?

One can set the used buffer with the `BufferPrototype` parameter in e.g.  `setitem` . Do you mean to have additionally to that a `BufferPrototype` stored in the `AsyncArray` instance which might get set upon creation of the array? So the decision of which buffer to use would be

prototype in `setitem` &rarr; prototype in `AsyncArray` instance &rarr; config  &rarr; numpy
 (with "&rarr;" being the fallback if previous was not set) 

wouldn't that be a bit confusing with 4 layers of fallbacks?

madsbk:
Yes, to enable the end-user to set default prototypes for each `AsyncArray` instance. But I see your point, it might be a bit too many fall backs :)

In any case, if we allow modification of `default_buffer_prototype`, I think we need an another constant like `numpy_buffer_prototype` that is always backed by a numpy array. E.g.  when reading the shard index, we always want to use numpy: https://github.com/zarr-developers/zarr-python/blob/v3/src/zarr/codecs/sharding.py#L610

brokkoli71:
good point! @madsbk

src/zarr/abc/codec.py
```diff
@@ -11,18 +11,23 @@
...
     from zarr.array_spec import ArraySpec
+    from zarr.common import JSON
     from zarr.indexing import SelectorTuple
 
 CodecInput = TypeVar("CodecInput", bound=NDBuffer | Buffer)
 CodecOutput = TypeVar("CodecOutput", bound=NDBuffer | Buffer)
 
 
+def get_config() -> Config:
+    return config
```
jhamman, line 28: at first glance, I don't understand what this function is doing. Do we need it? If so, would you mind adding a comment explaining what it is for?

src/zarr/config.py
```diff
@@ -2,16 +2,52 @@
...
+    """
+
+    def reset(self) -> None:
+        self.clear()
+        self.refresh()
+
 
 config = Config(
-    "zarr",
+    "zarr_python",
```
jhamman, line 32: I'm curious about the motivation for this change. I think I'd like to keep this as `"zarr"` because that is the name of this python package.

tests/v3/test_buffer.py
```diff
@@ -1,77 +1,18 @@
...
-        prototype: BufferPrototype,
-        byte_range: tuple[int, int | None] | None = None,
-    ) -> Buffer | None:
-        if "json" not in key:
-            assert prototype.buffer is MyBuffer
-        ret = await super().get(key=key, prototype=prototype, byte_range=byte_range)
-        if ret is not None:
-            assert isinstance(ret, prototype.buffer)
-        return ret
+from zarr.testing.buffer import MyBuffer, MyNDArrayLike, MyNDBuffer, StoreExpectingMyBuffer
```
jhamman, line 15: I suggest we work a bit on these names. Something more descriptive would be helpful if we expect developers to use these in their own buffer tests.

tests/v3/test_config.py
```diff
@@ -1,8 +1,38 @@
...
+    register_pipeline,
+)
+from zarr.testing.buffer import MyBuffer, MyNDArrayLike, MyNDBuffer, StoreExpectingMyBuffer
+
+
+@pytest.fixture()
+def reset_config():
+    config.reset()
+    yield
+    config.reset()
```
jhamman, line 35: this seems like a duplicate of the `autouse` fixture above.

src/zarr/config.py
```diff
@@ -2,16 +2,52 @@
...
+                "gzip": {"name": "GzipCodec"},
+                "zstd": {"name": "ZstdCodec"},
+                "bytes": {"name": "BytesCodec"},
+                "endian": {"name": "BytesCodec"},  # compatibility with earlier versions of ZEP1
+                "crc32c": {"name": "Crc32cCodec"},
+                "sharding_indexed": {"name": "ShardingCodec"},
+                "transpose": {"name": "TransposeCodec"},
+            },
+            "buffer": {"name": "Buffer"},
+            "ndbuffer": {"name": "NDBuffer"},
```
jhamman, line 50: I'm a bit confused on how these entries interact with our entrypoints. Can we document this somewhere? Here or in registry.py would be good candidate locations.

jhamman (commented):
@brokkoli71 - thank for pushing this forward. Most everything is looking great! Just a few comments and requests for documentation.

src/zarr/config.py
brokkoli71: Donfig lets you use environment variables to set the configuration starting with the name specified here (e.g. `ZARR_PYTHON_CODEC_PIPELINE__NAME`)
My thoughts were: to avoid confusion about the origin of this environment variable, I wanted to give a hint with "python" 
But I also have no problem with "zarr"

src/zarr/config.py
brokkoli71: I've added some documentation: https://github.com/zarr-developers/zarr-python/pull/1982/commits/96271571c82cc87c09907318dbfc33215a282da0
Not sure about the level of detail, I'm open to feedback and discussion

src/zarr/config.py
normanrz: I'd stick with `zarr`

src/zarr/config.py
normanrz: Entry points are great to register options. The configs here are meant to select of the options.

src/zarr/registry.py
```diff
@@ -0,0 +1,144 @@
...
+if TYPE_CHECKING:
+    from zarr.abc.codec import Codec, CodecPipeline
+    from zarr.buffer import Buffer, NDBuffer
+
+from importlib.metadata import EntryPoint
+from importlib.metadata import entry_points as get_entry_points
+
+from zarr.config import BadConfigError, config
+
+__codec_registry: dict[str, dict[str, type[Codec]]] = {}
```
normanrz, line 15: This code could be a bit DRY-er. Maybe add a generic class (e.g. namedtuple) for `__*_registry` and `__lazy_load_*`?

src/zarr/registry.py
```diff
@@ -0,0 +1,144 @@
...
+
+
+def _collect_entrypoints() -> (
+    tuple[dict[str, EntryPoint], list[EntryPoint], list[EntryPoint], list[EntryPoint]]
+):
+    """Collects codecs, pipelines, buffers and ndbuffers from entrypoints"""
+    entry_points = get_entry_points()
+    for e in entry_points.select(group="zarr.codecs"):
+        __lazy_load_codecs[e.name] = e
+    for e in entry_points.select(group="zarr"):
```
normanrz, line 38: What happens in multiple libraries in my env declare a `buffer` class? How will I be able to pick the right one?

src/zarr/config.py
```diff
@@ -2,16 +2,63 @@
...
+    "zarr_python",
     defaults=[
         {
             "array": {"order": "C"},
             "async": {"concurrency": None, "timeout": None},
-            "codec_pipeline": {"batch_size": 1},
             "json_indent": 2,
+            "codec_pipeline": {"name": "BatchedCodecPipeline", "batch_size": 1},
+            "codecs": {
+                "blosc": {"name": "BloscCodec"},
```
normanrz, line 51: I think these should be fully-qualified class names, e.g.
```suggestion
                "blosc": "zarr.codecs.blosc.BloscCodec",
```
I don't think the value needs to be an object.

src/zarr/registry.py
brokkoli71: all will be registered and the correct one according to `config` will be chosen in `get_buffer_class`. With using the fully qualified name this should be an unambiguous assignment

brokkoli71: do you mean multiple implementations per entrypoint?

brokkoli71: oh, multiple implementations of the same codec would overwrite each other in the __lazy_load_codecs dict

normanrz: I think that needs to change. The idea is that multiple libraries can provide implementations for a specific codec (e.g. CPU- and GPU-based gzip) via entrypoints and the user can select one of these implementations via config.

I was also concerned about the other entrypoints (e.g. buffer, pipeline). It would be nice if libraries could declare more than one of each. Again, the user could then select the class via the config through the fully-qualified name.

brokkoli71: hmm how would one specify that in `entry_points.txt`?
my ideas were 
1. seperate different implementations comma-separated like
```
[zarr.codecs]
gzip = package:EntrypointGzipCodec1, package:EntrypointGzipCodec2
[zarr]
buffer = package:TestEntrypointBuffer, package:AnotherTestEntrypointBuffer
```
but that would deviate from the PEP 508 standard of entry_points.txt 

2. provide names for different implementations (which do not have an effect) and have a group for buffer, ndbuffer, pipeline and every Codec 
```
[zarr.codecs.gzip]
some_name = package:EntrypointGzipCodec1
another = package:EntrypointGzipCodec2
[zarr.buffer]
some_name = package:TestEntrypointBuffer
another = package:AnotherTestEntrypointBuffer
```
I'd prefer the second but I dont like having random names for each implementation that will not be used at parsing
any other ideas?

normanrz: Would it be possible to support both single items (e.g. `zarr.buffer = package:TestEntrypointBuffer`) and groups (e.g. `zarr.buffer = { gpu_buffer = package:TestEntrypointBuffer, cpu_buffer = package:AnotherTestEntrypointBuffer }`)?

brokkoli71: good idea! allowed syntax is now all of the following:
```
[zarr.codecs]
gzip = package:EntrypointGzipCodec1
[zarr.codecs.gzip]
some_name = package:EntrypointGzipCodec2
another = package:EntrypointGzipCodec3

[zarr]
buffer = package:TestBuffer1
[zarr.buffer]
xyz = package:TestBuffer2
abc = package:TestBuffer3
```

src/zarr/config.py
```diff
@@ -2,16 +2,66 @@
...
 config = Config(
     "zarr",
     defaults=[
         {
             "array": {"order": "C"},
             "async": {"concurrency": None, "timeout": None},
-            "codec_pipeline": {"batch_size": 1},
             "json_indent": 2,
+            "codec_pipeline": {
+                "path": "zarr.codecs.pipeline.BatchedCodecPipeline",
```
normanrz, line 50: Seeing this, I wonder if we should use an object with a `path` (or `class`?) for codecs as well. While we don't need it right now, this could be useful in the future for backwards-compat. We could also add that in the future, though. 
cc @jhamman @d-v-b

normanrz: Sorry for the back and forth on this!

src/zarr/registry.py
```diff
@@ -0,0 +1,179 @@
...
+The registry module is responsible for managing implementations of codecs, pipelines, buffers and ndbuffers and 
+collecting them from entrypoints.
+The implementation used is determined by the config
+"""
+
+
+def _collect_entrypoints() -> list[Registry[Any]]:
+    """
+    Collects codecs, pipelines, buffers and ndbuffers from entrypoints.
+    Allowed syntax for entry_points.txt is e.g.
```
normanrz, line 48:
```suggestion
    Entry points can either be single items or groups of items.
    Allowed syntax for entry_points.txt is e.g.
```
Why entry_points.txt? Isn't that something that is declared in a pyproject.toml?

src/zarr/config.py
normanrz: What I am proposing is 
```python
        {
            "array": {"order": "C"},
            "async": {"concurrency": None, "timeout": None},
            "json_indent": 2,
            "codec_pipeline": {
                "path": "zarr.codecs.pipeline.BatchedCodecPipeline",
                "batch_size": 1,
            },
            "codecs": {
                "blosc": { "path": "zarr.codecs.blosc.BloscCodec" },
                "gzip": { "path": "zarr.codecs.gzip.GzipCodec" },
                "zstd": { "path": "zarr.codecs.zstd.ZstdCodec" },
                "bytes": { "path": "zarr.codecs.bytes.BytesCodec" },
                "endian": { "path": "zarr.codecs.bytes.BytesCodec" },  # compatibility with earlier versions of ZEP1
                "crc32c": { "path": "zarr.codecs.crc32c_.Crc32cCodec" },
                "sharding_indexed": { "path": "zarr.codecs.sharding.ShardingCodec" },
                "transpose": { "path": "zarr.codecs.transpose.TransposeCodec" },
            },
            "buffer": "zarr.buffer.Buffer",
            "ndbuffer": "zarr.buffer.NDBuffer",
        }

```

PR #1987, opened 2024-06-24
Fix doc build

dstansby:
Fixes " WARNING: unknown mimetype for _static/donotdelete, ignoring ".  This file came in via https://github.com/zarr-developers/zarr-python/commit/7b741aef6053fb1a94951393ab061c02b646b2f4 with not much explanation, so I think it's probalby safe to delete given it's empty 🤷


jakirkham:
Thanks David! 🙏

Happy to try this 🙂

When this file was created in commit ( https://github.com/zarr-developers/zarr-python/commit/7b741aef6053fb1a94951393ab061c02b646b2f4 ), it notes...

> try to fix sphinx viewcode

Not sure why this would be needed for that functionality (or if it still matters given that was 2016). Though maybe we should double check that we can still view code blocks in RTD

jakirkham:
Ok checked one occurrence using the new API docs and it appears to work ok

* Started with https://zarr--1987.org.readthedocs.build/en/1987/api/zarr.html
* Clicked `[source]` up top
* Went to https://zarr--1987.org.readthedocs.build/en/1987/_modules/zarr/array.html#Array

AFAICT this renders correctly

jakirkham:
Hooray! It [passed]( https://readthedocs.org/projects/zarr/builds/24798393/ ) 🥳

Thanks David! 🙏

Issue #1978, opened 2024-06-18
[v3] Mode 'r+' is not working


tomwhite:
### Zarr version

3.0.0a1

### Numcodecs version

0.12.1

### Python Version

3.11.9

### Operating System

Mac

### Installation

pip

### Description

It looks like 'r+', which should read or write to an existing store, doesn't work in v3. (Some of the other modes don't work either, like 'a', which should create the store for read/write if it doesn't exist.)

### Steps to reproduce

```python
>>> import zarr
>>> zarr.create(store='example-v3.zarr', shape=(3, 3))
<Array file://example-v3.zarr shape=(3, 3) dtype=float64>
>>> z = zarr.open(store='example-v3.zarr', mode='r+')
>>> z[:] = 3
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/tom/workspace/zarr-python/src/zarr/array.py", line 720, in __setitem__
    self.set_basic_selection(cast(BasicSelection, pure_selection), value, fields=fields)
  File "/Users/tom/workspace/zarr-python/src/zarr/array.py", line 751, in set_basic_selection
    sync(self._async_array._set_selection(indexer, value, fields=fields, prototype=prototype))
  File "/Users/tom/workspace/zarr-python/src/zarr/sync.py", line 92, in sync
    raise return_result
  File "/Users/tom/workspace/zarr-python/src/zarr/sync.py", line 51, in _runner
    return await coro
           ^^^^^^^^^^
  File "/Users/tom/workspace/zarr-python/src/zarr/array.py", line 507, in _set_selection
    await self.metadata.codec_pipeline.write(
  File "/Users/tom/workspace/zarr-python/src/zarr/codecs/pipeline.py", line 504, in write
    await concurrent_map(
  File "/Users/tom/workspace/zarr-python/src/zarr/common.py", line 53, in concurrent_map
    return await asyncio.gather(*[func(*item) for item in items])
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/tom/workspace/zarr-python/src/zarr/codecs/pipeline.py", line 454, in write_batch
    await concurrent_map(
  File "/Users/tom/workspace/zarr-python/src/zarr/common.py", line 53, in concurrent_map
    return await asyncio.gather(*[func(*item) for item in items])
[...]

PR #2000, opened 2024-06-27
fix file modes

brokkoli71:
fixes https://github.com/zarr-developers/zarr-python/issues/1978
according to docstrings of `zarr.open` the modes should do:
```
mode : {'r', 'r+', 'a', 'w', 'w-'}, optional
        Persistence mode: 'r' means read only (must exist); 'r+' means
        read/write (must exist); 'a' means read/write (create if doesn't
        exist); 'w' means create (overwrite if exists); 'w-' means create
        (fail if exists).
```
this PR implements and tests this behavior, so https://github.com/zarr-developers/zarr-python/issues/1978 gets fixed


src/zarr/abc/store.py
```diff
@@ -22,8 +22,20 @@ def mode(self) -> OpenMode:
     @property
     def writeable(self) -> bool:
```
d-v-b, line 23: instead of multiple properties that all interpret `self.mode` in different ways, perhaps we should just make `self.mode` more expressive? E.g., suppose `self.mode` was an instance of the following type:

```
WriteMode = {
    create: bool,
    overwrite :bool,
},
write: False | WriteMode
}
```

I'm sure someone can come up with something better. But I think we should consider this approach (making one property more expressive) before we add a lot of properties to the store.

brokkoli71: good idea!

brokkoli71: do you mean to remove the literals` ["r", "r+", "a", "w", "w-"] `completely? or just organize the properties better into `WriteMode`

d-v-b: I think we could accept the literals as parameters in the constructor, but we should consider normalizing them to a more explicit data structure.

src/zarr/abc/store.py
```diff
@@ -1,31 +1,44 @@
...
+    @classmethod
+    def from_str(cls, mode: str) -> Self:
+        if mode not in ("r", "r+", "a", "w", "w-"):
+            raise ValueError("mode must be one of 'r', 'r+', 'w', 'w-', 'a'")
+        return cls(
+            overwrite=mode == "w",
+            can_create=mode in ("a", "w", "w-"),
+            can_open_existing=mode in ("r", "r+", "a"),
+            is_writable=mode in ("r+", "a", "w", "w-"),
+        )
```
brokkoli71, line 27: what do you think? @d-v-b 
(I think the name `write_mode` does not describe it fully, as it also describes the behavior at opening an array, so I called it `OpenMode`, but I am open to discussion)

jhamman:
Thanks @brokkoli71 for picking this up. What is the expected behavior of `mode='w'`? Should we have a `Store.clear()` method that we call in these cases?

brokkoli71:
> Thanks @brokkoli71 for picking this up. What is the expected behavior of `mode='w'`? Should we have a `Store.clear()` method that we call in these cases?

good point, according to the docstrings: `'w' means create (overwrite if exists)`. So yes, would be probably a good idea to have a generalized `Store.clear`. I will have a look at that

d-v-b:
I agree that a method for clearing a store is a good idea. Just bear in mind that this can be a performance bottleneck when deleting arrays with a lot of chunks, especially on remote storage. We should think about performance at some point.

src/zarr/abc/store.py
d-v-b: re: naming, could `AccessMode` also work? But `OpenMode` is also fine with me.

regarding the fields of `OpenMode`:
-  if `overwrite` is `True`, then `can_create` and `is_writable` must also be `True`, and `can_open_existing` must be `False`. This means that `overwrite` is degenerate with respect to the other three attributes, and thus `overwrite` can be removed as a field (perhaps replaced by `@property`-decorated method, if we use a dataclass instead of a named tuple here).
- I would shorten `can_create` to just `create`, `is_writeable` to `write`, and `can_open_existing` to `allow_existing` / `allow_extant`. This is personal preference.


Here's another idea, for inspiration; feel free to ignore it :) 

```python

class AccessMode:
  mutable: bool

@dataclass
class ReadMode(AccessMode):
    mutable: False

@dataclass
class WriteMode(AccessMode):
  mutable: True
  update: bool # instead of can_open_existing
  create: bool # instead of can_create


class Store:
  access_mode: ReadMode | WriteMode
  ...

```
Code consuming this API would look like this:

```python

if (m := store.access_mode).mutable:
    if m.update:
        ...
    if m.create:
       ...
```

brokkoli71:
> I agree that a method for clearing a store is a good idea. Just bear in mind that this can be a performance bottleneck when deleting arrays with a lot of chunks, especially on remote storage. We should think about performance at some point.

I was wondering, if supporting `mode='w'` as `create (overwrite if exists)` is a good idea concerning the performance? what is the usecase of this mode? maybe it would be better to explicitly have a `delete` or `clear` method? or we should have a warning in the docstrings?

d-v-b:
> > I agree that a method for clearing a store is a good idea. Just bear in mind that this can be a performance bottleneck when deleting arrays with a lot of chunks, especially on remote storage. We should think about performance at some point.
> 
> I was wondering, if supporting `mode='w'` as `create (overwrite if exists)` is a good idea concerning the performance? what is the usecase of this mode? maybe it would be better to explicitly have a `delete` or `clear` method? or we should have a warning in the docstrings?

There's a lot of usage for `create + overwrite` -- if users are rapidly iterating on something, they already created an old array, but they want to save a new array with the same name + different datatype, then the easiest way to do this is via something like an overwrite option. Handling this via array the creation flow is more ergonomic than forcing users to go through a separate API.

Don't worry too much about the performance concerns with bulk deletion for now. We should make the API convenient, then worry about performance later. I only brought delete performance up because it's something that zarr python v2 didn't deal with, which caused me some headaches.

src/zarr/abc/store.py
brokkoli71: I like this. also `AccessMode` feels better than `OpenMode`. 

The overwrite attribute is necessary to distinguish between `w` (overwrite) and `w-` (raise error if file exist)

src/zarr/abc/store.py
```diff
@@ -1,31 +1,58 @@
...
+            elif self.mode.overwrite:
+                self.clear()
+            else:
+                raise FileExistsError("Store already exists")
+
+    @abstractmethod
+    def _exists(self) -> bool: ...
+
+    @abstractmethod
+    def clear(self) -> None: ...
```
jhamman, line 47: I feel that these should be async methods. That may mean we need a class method to open a store (instead of using __init__.py).

jhamman: Something like:

```suggestion
   @classmethod
   @abstractmethod
    async def open(cls, mode: AccessModeLiteral = "r"):
        if await self.exists():
            ...
        return cls(mode=mode)

    @abstractmethod
    async def exists(self) -> bool: ...

    @abstractmethod
    async def clear(self) -> None: ...
```

brokkoli71: true! `async open` instead of `__init__ `seems to be the right way to go

tests/v3/test_api.py
```diff
@@ -88,6 +91,65 @@ def test_save_errors() -> None:
...
 
 
+@fixture
+def tmppath(tmpdir):
+    return str(tmpdir / "example.zarr")
+
+
+def test_open_with_mode_r(tmppath) -> None:
+    # 'r' means read only (must exist)
+    import zarr
```
jhamman, line 101: I don't think you need any of these `import zarr` statements.

src/zarr/abc/store.py
```diff
@@ -1,31 +1,72 @@
...
+        if await self.root_exists():
+            if self.mode.update or self.mode.readonly:
+                pass
+            elif self.mode.overwrite:
+                await self.clear()
+            else:
+                raise FileExistsError("Store already exists")
+        self._is_open = True
+
+    async def ensure_open(self) -> None:
```
jhamman, line 53: can we make this `_ensure_open()`

src/zarr/abc/store.py
```diff
@@ -1,31 +1,72 @@
...
 
-    def __init__(self, mode: OpenMode = "r"):
-        if mode not in ("r", "r+", "w", "w-", "a"):
-            raise ValueError("mode must be one of 'r', 'r+', 'w', 'w-', 'a'")
-        self._mode = mode
+    @abstractmethod
+    async def root_exists(self) -> bool: ...
+
+    @abstractmethod
+    async def clear(self) -> None: ...
```
jhamman, line 61: let's add tests for both of these in `testing/store.py`

jhamman (commented):
this is looking great! Just a few final comments.

src/zarr/abc/store.py
brokkoli71: Thanks for the feedback. I'd be interested in why marking it protected, as it's called in `make_store_path`

src/zarr/abc/store.py
brokkoli71: good point! I added tests and changed `root_exists` to `not empty`

src/zarr/abc/store.py
jhamman: I see your point. I was taking the perspective that we probably don't want users to depend on this (or developers to modify its behavior). For this reason, I was thinking it would be better to show that this is an internal API.

src/zarr/abc/store.py
```diff
@@ -1,31 +1,72 @@
...
+        self._is_open = False
+        self._mode = AccessMode.from_literal(mode)
+
+    @classmethod
+    async def open(cls, *args: Any, **kwargs: Any) -> Self:
+        store = cls(*args, **kwargs)
+        await store._open()
+        return store
+
+    async def _open(self) -> None:
```
jhamman, line 43: This seems important to avoid clobbering existing data.
```suggestion
    async def _open(self) -> None:
        if self._is_open:
            raise ValueError('store is already open')
```

jhamman (approved):
One last comment and a suggestion on `ensure_open` that you can take or leave. Thanks again for pushing on this.

src/zarr/abc/store.py
brokkoli71: that's a valid point

jhamman:
@brokkoli71 - happy to merge this if you can resolve the conflicts.

joshmoore:
Sadly still conflicting, @brokkoli71

Issue #1993, opened 2024-06-26
Finish v3 API reference docs


dstansby:
In https://github.com/zarr-developers/zarr-python/pull/1918 I added structure for generating API docs, but only added the `zarr` namespace. We should add the other namespaces that need documenting too.

Does someone more involved in `v3` have a list of which sub-modules should be added to the API docs?

d-v-b:
If we are auto-generating content, then I would say we add everything to start, and then remove things later as needed.

PR #2002, opened 2024-06-28
Autogenerate all API docs

dstansby:
This updates the doc build to automatically generate all API docs. I've set it so any module with a part that starts with "_" is not documented, so that's the way to make a sub-module private.

See https://zarr--2002.org.readthedocs.build/en/2002/_autoapi/zarr/index.html for the API ref in the build of this PR. 

I think it's worth merging this to make sure new additions/removals to the API don't break API docs, and we can bikeshed what's in/out of the public API in a follow up PR.

Fixes https://github.com/zarr-developers/zarr-python/issues/1993.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -178,7 +178,7 @@ def autoapi_skip_modules(app: sphinx.application.Sphinx, what: str, name: str, o
     Return True if a module should be skipped in th API docs.
     """
     parts = name.split(".")
-    if what == "module" and any((part.startswith("_") or part == "v2") for part in parts):
+    if what == "module" and (any(part.startswith("_") for part in parts) or "v2" in name or name.startswith("zarr.core")):
         return True
     return False
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -59,7 +59,7 @@ autoapi_dirs = ['../src/zarr']
 autoapi_add_toctree_entry = False
 autoapi_generate_api_docs = True
 autoapi_member_order = "groupwise"
-autoapi_root = "api"
+autoapi_root = "_autoapi"
 autoapi_keep_files = True
 
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -57,9 +57,10 @@ issues_github_path = "zarr-developers/zarr-python"
 
 autoapi_dirs = ['../src/zarr']
 autoapi_add_toctree_entry = False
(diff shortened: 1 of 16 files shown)


docs/conf.py
```diff
@@ -172,8 +173,19 @@
 html_logo = "_static/logo1.png"
 
 
+def autoapi_skip_modules(app: sphinx.application.Sphinx, what: str, name: str, obj: object, skip: bool, options: dict[str, Any]) -> bool:
+    """
+    Return True if a module should be skipped in th API docs.
+    """
+    parts = name.split(".")
+    if what == "module" and any((part.startswith("_") or part == "v2") for part in parts):
```
jhamman, line 181: can we also skip [`zarr.core`](https://zarr--2002.org.readthedocs.build/en/2002/_autoapi/zarr/core/index.html)

jhamman (commented):
I like this @dstansby. One suggestion but otherwise this looks good.

docs/conf.py
dstansby: Instead of putting in another condition to skip `zarr.core`, could we rename it to `zarr._core` instead? This will automatically skip it in the API docs with the current logic, and also signal to folks using autocomplete in their editor that it's not public API.

dstansby: @jhamman I skipped it for now - I think it's worth renaming to zarr.core but it's a big diff so I can propose that in a follow up PR 😄

dstansby:
pre-commit.ci autofix

dstansby:
pre-commit.ci autofix

dstansby:
@zarr-developers/python-core-devs would be good to get this reviewed and merged soon before more conflicts with other PRs start coming in 😄

d-v-b (approved):

d-v-b:
this looks good to me! thanks for this @dstansby

d-v-b:
it looks like one of the [hypothesis tests](https://github.com/zarr-developers/zarr-python/actions/runs/10758622910/job/29834241753?pr=2002) is failing -- is that related to these changes?

dstansby:
I can reproduce on `v3` branch, I've opened an issue at https://github.com/zarr-developers/zarr-python/issues/2155

d-v-b:
> I can reproduce on `v3` branch, I've opened an issue at #2155

In that case I'm going to hit merge on this.

PR #2003, opened 2024-06-28
Fix low contrast in dark theme h3 headings

dstansby:
The contrast between text and background is bad for `h3` elements:

Screenshot 2024-06-28 at 08 29 26

This is because the colour is manually set in the `zarr` CSS - removing this fixes the low contrast by making the titles white.


d-v-b:
the change seems reasonable, do we know why the manual adjustment was made?

dstansby:
It came in this rather large commit: https://github.com/zarr-developers/zarr-python/commit/4dc6f1f5046708648b5e6b82d0a64a24d1a40566. I think it's worth relying on `pydata-sphinx-theme` to set this though to reduce the amount of custom CSS we have to maintain.

PR #2006, opened 2024-07-02
test: check that store, array, and group classes are serializable 

### LINKED ISSUE #1760: [v3] Test serialization of arrays and groups
Arrays and Groups should be able to be round tripped with Python's pickle protocol. We need a test to make sure this works to be sure downstream applications can work with our new APIs.

jhamman:
This PR adds tests to insure that our store/array/group classes are serializable w/ using pickle and can be reliably roundtripped. This is important for Dask and other libraries that need to be able to share Zarr objects between processes/machines.

A necessary side effect of this work was the creation of a few additional `__eq__` methods. 

closes #1760 

TODO:


src/zarr/store/memory.py
```diff
@@ -29,6 +29,25 @@ def __str__(self) -> str:
...
+        #     f"unpickling {type(self)} may produce unexpected behavior and should only be used for testing and/or debugging"
+        # )
+        self._store_dict, self._mode = state
+
+    def __getstate__(self) -> tuple[MutableMapping[str, Buffer], OpenMode]:
+        # warnings.warn(
+        #     f"pickling {type(self)} may produce unexpected behavior and should only be used for testing and/or debugging"
+        # )
+        return self._store_dict, self._mode
+
```
jhamman, line 50: I'm not settled on what is the correct behavior for serializing the MemoryStore class. Perhaps we should just raise an error? It is useful for testing but ultimately, its not a useful feature beyond that.

d-v-b: +1 to raising an error here. I can't see any use for serializing a memory store directly, vs persisting the contents to more durable storage like `LocalStore`

jhamman: this is done now.

tests/v3/test_group.py
```diff
@@ -391,3 +392,26 @@ def test_group_name_properties(store: LocalStore | MemoryStore, zarr_format: Zar
...
+        store=store, attributes={"foo": 999}, zarr_format=zarr_format
+    )
+    p = pickle.dumps(expected)
+    actual = pickle.loads(p)
+    assert actual == expected
+
+
+@pytest.mark.parametrize("store", ("local",), indirect=["store"])
+@pytest.mark.parametrize("zarr_format", (2, 3))
+def test_serizalizable_sync_group(store: LocalStore | MemoryStore, zarr_format: ZarrFormat) -> None:
```
d-v-b, line 412:
```suggestion
def test_serizalizable_sync_group(store: LocalStore, zarr_format: ZarrFormat) -> None:
```

tests/v3/test_array.py
```diff
@@ -34,3 +37,36 @@ def test_array_name_properties_with_group(
...
+    actual = pickle.loads(p)
+
+    assert actual == expected
+    # np.testing.assert_array_equal(await actual.getitem(slice(None)), await expected.getitem(slice(None)))
+    # TODO: uncomment the parts of this test that will be impacted by the config/prototype changes in flight
+
+
+@pytest.mark.parametrize("store", ("local",), indirect=["store"])
+@pytest.mark.parametrize("zarr_format", (2, 3))
+def test_serizalizable_sync_array(store: LocalStore | MemoryStore, zarr_format: ZarrFormat) -> None:
```
d-v-b, line 62:
```suggestion
def test_serizalizable_sync_array(store: LocalStore, zarr_format: ZarrFormat) -> None:
```

src/zarr/buffer.py
```diff
@@ -261,6 +261,11 @@ def __add__(self, other: Buffer) -> Self:
             np.concatenate((np.asanyarray(self._data), np.asanyarray(other_array)))
         )
 
+    def __eq__(self, other: object) -> bool:
+        # Note: this was needed to support comparing MemoryStore instances with Buffer values in them
+        # if/when we stopped putting buffers in memory stores, this can be removed
+        return isinstance(other, type(self)) and self.to_bytes() == other.to_bytes()
```
d-v-b, line 267: for reference, because I had to look this up: `and` short-circuits, so if the first leg of the operator is `False` we never invoke the potentially expensive second leg, which is good.

src/zarr/testing/store.py
```diff
@@ -42,6 +43,19 @@ def test_store_type(self, store: S) -> None:
         assert isinstance(store, Store)
         assert isinstance(store, self.store_cls)
 
+    def test_store_eq(self, store: S, store_kwargs: dict[str, Any]) -> None:
+        # check self equality
+        assert store == store
+
+        # check store equality with same inputs
+        # asserting this is important for being able to compare (de)serialized stores
+        store2 = self.store_cls(**store_kwargs)
```
d-v-b, line 52: we should also add a test showing that if `**store_kwargs` is different, then the two stores are not equal. But I don't think this test can be defined in the base class, since it depends on the structure of `store_kwargs`. So no action needed here.

tests/v3/test_array.py
```diff
@@ -136,3 +137,36 @@ def test_array_v3_fill_value(store: MemoryStore, fill_value: int, dtype_str: str
...
+    actual = pickle.loads(p)
+
+    assert actual == expected
+    # np.testing.assert_array_equal(await actual.getitem(slice(None)), await expected.getitem(slice(None)))
+    # TODO: uncomment the parts of this test that will be impacted by the config/prototype changes in flight
+
+
+@pytest.mark.parametrize("store", ("local",), indirect=["store"])
+@pytest.mark.parametrize("zarr_format", (2, 3))
+def test_serizalizable_sync_array(store: LocalStore, zarr_format: ZarrFormat) -> None:
```
d-v-b, line 162:
```suggestion
def test_serializable_sync_array(store: LocalStore, zarr_format: ZarrFormat) -> None:
```

d-v-b (approved):
Some typos in the test names but otherwise this looks good to go

tests/v3/test_array.py
```diff
@@ -136,3 +137,36 @@ def test_array_v3_fill_value(store: MemoryStore, fill_value: int, dtype_str: str
 
     assert arr.fill_value == np.dtype(dtype_str).type(fill_value)
     assert arr.fill_value.dtype == arr.dtype
+
+
+@pytest.mark.parametrize("store", ("local",), indirect=["store"])
```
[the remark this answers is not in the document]
jhamman: because the memory store is not pickle-able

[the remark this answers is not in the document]
d-v-b: that's a good reason! :) we could include memorystore, but mark it with xfail, but I don't think that should be a blocker if you want to get this out the door

src/zarr/buffer.py
akshaysubr: Can we make this a separate utility function that compares a buffer with any other arbitrary object?
```python
class Buffer:
    ...
    def __eq__(self, other: Buffer) -> bool:
        # Another Buffer class can override this to choose a more efficient path
        return np.allclose(self._data, other.as_numpy_array())

def buffers_equal(buffer: Buffer, other: object) -> bool:
    return isinstance(other, Buffer) and buffer == other
```

jhamman: @akshaysubr - happy to. Your example makes mypy unhappy though:

```
src/zarr/core/buffer.py:267: error: Argument 1 to "allclose" has incompatible type "ArrayLike"; expected "_SupportsArray[dtype[Any]] | _NestedSequence[_SupportsArray[dtype[Any]]] | bool | int | float | complex | str | bytes | _NestedSequence[bool | int | float | complex | str | bytes]"  [arg-type]
src/zarr/core/buffer.py:267: error: "object" has no attribute "as_numpy_array"  [attr-defined]
```
Would this be okay?

```python
        # Another Buffer class can override this to choose a more efficient path
        try:
            return np.allclose(self._data, other.as_numpy_array())  # type: ignore[arg-type, attr-defined]
        except AttributeError:
            return False
```

akshaysubr: It might be cleaner to just do something like this so both arguments are guaranteed to be numpy arrays?
```python
    def __eq__(self, other: Buffer) -> bool:
        # Another Buffer class can override this to choose a more efficient path
        return np.array_equal(self.as_numpy_array(), other.as_numpy_array())
```

tests/v3/test_group.py
```diff
@@ -681,152 +680,22 @@ async def test_asyncgroup_update_attributes(store: Store, zarr_format: ZarrForma
     assert agroup_new_attributes.attrs == attributes_new
 
 
-async def test_group_members_async(store: LocalStore | MemoryStore) -> None:
```
TomAugspurger, line 684: @jhamman was this `test_group_members_async` test moved somewhere, or deliberately removed?

jhamman: Not deliberately removed. Must have been a bad merge conflict resolution. I'll bring it back today. Sorry!

src/zarr/store/memory.py
TomAugspurger: There's an xarray test relying on memory store being pickle-able.

You can tell stories about about how this might be useful (opening a store in one process and then using multiple processes for reads; with shared memory that might actually cut down on the size of the actual bytes in memory).

I think the serialized representation should be pretty unambiguous (just the default?) so adding this should be low risk.

d-v-b: i take back my original skepticism of the utility of pickling the memorystore :) as long as people know what they are getting into, I guess it's fine (and it's good to avoid breaking dependent projects)

PR #2014, opened 2024-07-04
ensure that fill_value matches dtype

d-v-b:
- ensures that the `fill_value` of a v3 array is an instance of that array's dtype. The behavior here matches numpy type casting rules, so if `dtype` is `bool` and the `fill_value` is provided as `1`, the resulting `fill_value` metadata will be `True`. I'm happy to change this behavior, but for now this is expedient and matches what people might expect from v2 and numpy 
- refactors the metadata classes by partitioning v2 and v3 metadata into `src/zarr/metadata/v2.py` and `src/zarr/metadata/v2.py`, respectively

TODO:


d-v-b:
I got a bit carried away here, i'm closing this and re-opening something more tame

PR #2020, opened 2024-07-09
Cast fill value to array's dtype

d-v-b:
In [v3 right now](https://github.com/zarr-developers/zarr-python/blob/33b158974a55f1818f27dcc9a3bd2135c51450ff/src/zarr/common.py#L161), we don't apply any parsing or validation to the `fill_value` attribute of an array. This PR fixes that.

In this PR, the `fill_value` is cast to an instance of the array's dtype, using normal numpy dtype casting semantics. I also added JSON serialization for complex dtypes, but this is not tested yet. I will add those tests as part of a later refactor of the metadata tests.

For v2 arrays, I am importing the v2 fill value parsing function and wrapping it in a very light wrapper. This preserves all the v2 behavior.

I also added a `fill_value` attribute to the `Array` class. Happy to remove this if it's not supposed to be there.

I put the metadata tests in a new directory structure: `tests/v3/metadata/test_v2.py` for v2 stuff and `tests/v3/metadata/test_v3.py` for v3 stuff. This keeps the v2 and v3 logic separated, and removes the need to name functions `test_foo_v3` and the like. We should ultimately use an analogous layout in the library code.

Longer term we should support the raw bits datatypes defined in the zarr v3 spec, but I don't think we need that now.

TODO:


tests/v3/test_metadata/test_v3.py
```diff
@@ -0,0 +1,158 @@
...
+    )
+    with pytest.raises(ValueError, match=re.escape(match)):
+        parse_fill_value(fill_value=fill_value, dtype=dtype)
+
+
+@pytest.mark.parametrize("fill_value", [{"foo": 10}])
+@pytest.mark.parametrize("dtype_str", [*int_dtypes, *float_dtypes, *complex_dtypes])
+def test_parse_fill_value_invalid_type(fill_value: Any, dtype_str: str) -> None:
+    """
+    Test that parse_fill_value(fill-value, dtype) raises TypeError for invalid non-sequential types.
```
d-v-b, line 131:
```suggestion
    Test that parse_fill_value(fill_value, dtype) raises TypeError for invalid non-sequential types.
```

normanrz (approved):
Great work!

PR #2031, opened 2024-07-12
ArrowRecordBatchCodec and vlen string support

rabernat:
The discussion in https://github.com/zarr-developers/zeps/pull/47 got me thinking: what if, instead of turning numpy arrays into bytes, we turn them into self-describing [Arrow Record Batches](https://arrow.apache.org/docs/python/generated/pyarrow.RecordBatch.html) and serialize them using the Arrow IPC format.

This would be a new type of Array -> Bytes codec. The beautiful thing about this is that it gives us string encoding for free (as well as potentially many other benefits).

This PR is a proof of concept that this is feasible and in fact very easy.

There is a lot more to explore here, but I thought I would just through this up for discussion.

TODO:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -109,6 +109,7 @@ dependencies = [
     "universal_pathlib"
 ]
 extra-dependencies = [
+    "pyarrow",
     "coverage",
     "pytest",
     "pytest-cov",
--- a/src/zarr/codecs/arrow.py
+++ b/src/zarr/codecs/arrow.py
@@ -1,0 +1,81 @@
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import TYPE_CHECKING
+
+import pyarrow as pa
+
+from zarr.abc.codec import ArrayBytesCodec
+from zarr.array_spec import ArraySpec
+from zarr.buffer import Buffer, NDBuffer
+from zarr.codecs.registry import register_codec
+from zarr.common import JSON, parse_named_configuration
+
+if TYPE_CHECKING:
+    from typing_extensions import Self
+
+CHUNK_FIELD_NAME = "zarr_chunk"
+
+
+@dataclass(frozen=True)
+class ArrowRecordBatchCodec(ArrayBytesCodec):
+    def __init__(self) -> None:
+        pass
+
+    @classmethod
+    def from_dict(cls, data: dict[str, JSON]) -> Self:
+        _, configuration_parsed = parse_named_configuration(
+            data, "arrow", require_configuration=False
+        )
+        configuration_parsed = configuration_parsed or {}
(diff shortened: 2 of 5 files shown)


rabernat:
This experiment also suggests another interesting possibility: returning Arrow Arrays and Tables from a Zarr Array or Group. If the Zarr Arrays are all 1D, they can be represented as Arrow Arrays all the way through, and there are potentially opportunities to reduce memory copies. We could have ArrowBuffer / ArrowArrayBuffer types.

jeromekelleher:
This sounds really interesting and potentially very powerful @rabernat!

Would you mind commenting on the implications of a pyarrow dependency?

rabernat:
> Would you mind commenting on the implications of a pyarrow dependency?

I feel like it is becoming as ubiquitous as numpy in the ecosystem, so I don't consider this a major blocker. Or it could be an optional dependency if you want to read data encoded this way. But I'd be curious to hear opinions on that.

TomAugspurger:
There's lots of feedback in https://github.com/pandas-dev/pandas/issues/54466 on pandas adopting pyarrow as a required dependency. The primary concern raised is the size of the package, especially in serverless contexts (though it seems like AWS Lambda [has some built-in support](https://github.com/pandas-dev/pandas/issues/54466#issuecomment-2102599970) to make this not so much of an issue?).

There's some work being done in pyarrow to make core pieces available without having to bring in everything.

pyproject.toml
```diff
@@ -109,6 +109,7 @@ dependencies = [
     "universal_pathlib"
 ]
 extra-dependencies = [
+    "pyarrow",
```
joshmoore, line 112: TIL: https://pypi.org/project/nanoarrow

src/zarr/codecs/arrow.py
```diff
@@ -0,0 +1,81 @@
...
+            )
+        return chunk_array
+
+    async def _encode_single(
+        self,
+        chunk_array: NDBuffer,
+        chunk_spec: ArraySpec,
+    ) -> Buffer | None:
+        assert isinstance(chunk_array, NDBuffer)
+        arrow_array = pa.array(chunk_array.as_ndarray_like().ravel())
```
akshaysubr, line 69: Should probably be `chunk_array.as_numpy_array()` here since `pa.array()` doesn't recognize CuPy arrays? Would be good to add a GPU test here for safety.

In theory, it would be possible to do zero-copy transfers for CuPy arrays too but would need to go from [CuPy -> Numba](https://numba.readthedocs.io/en/stable/cuda/cuda_array_interface.html#numba.cuda.as_cuda_array) first and then [Numba -> Arrow](https://arrow.apache.org/docs/python/integration/cuda.html#numba-to-arrow).

PR #2036, opened 2024-07-14
Add `string` and `bytes` dtypes plus `vlen-utf8` and `vlen-bytes` codecs

rabernat:
This is an alternative approach to #2031 for implementing variable length string encoding, using the legacy [numcodecs vlen-utf8](https://github.com/zarr-developers/numcodecs/blob/main/numcodecs/vlen.pyx) codec.

The codec is very simple. It encodes variable length data like this:

```
| header | item 0 len | item 0 data | item 1 len | item 1 data | ... 
| uint4  | unit4      | variable    | unit4      | variable    | ...
```

where `header` is the number of items and each item is preceded by an int which tells how many bytes long it is.

We could add this as an "official" Zarr V3 codec if we want. The point of this PR is to demonstrate that it is simple to implement. If desirable, we could extend this to the other vlen types (bytes and array).

I am in favor of supporting this "legacy" encoding for backwards compatibility--I'd like existing Zarr V2 data to be convertible to V3 without rewriting the chunks.

How, I would favor using Arrow encoding (#2031) as the preferred default for new data.

xref:
- https://github.com/zarr-developers/zeps/pull/47
- https://github.com/zarr-developers/zarr-specs/issues/83


rabernat:
I compared this encoding + Zstd compression with the arrow-based encoding in #2031 and got some surprising results. The dataset consists of 47868 city names.

| encoding | compression | stored bytes |
| -- | -- | -- |
| vlen-utf8 | None | 642616 |
| vlen-utf8 | Zstd(2) | 366504 |
| vlen-utf8 | Zstd(2) | 358428 |
| arrow | None | 657185 |
| arrow | Zstd(2) | 462208 |
| arrow | Zstd(5) | 455234 |

The surprise is that the legacy format compresses _better_ than the arrow-based one.

<details>

```python
import zarr
from zarr.codecs import ZstdCodec
from zarr.buffer import default_buffer_prototype
from zarr.api.synchronous import array
import asyncio

# swap to compare across PRs
from zarr.codecs import VLenUTF8Codec as BytesCodec
#from zarr.codecs import ArrowRecordBatchCodec as BytesCodec

a = array(df.city.values, chunks=(1000,), fill_value='', codecs=[BytesCodec(), ZstdCodec(level=5)])

# ridiculous workarounds the lack of getsize on V3 store API
store_path = a._async_array.store_path
all_items = [item async for item in store_path.store.list()]
async def get_item_size(store, item):
    return len(await store.get(item, prototype=default_buffer_prototype))
all_sizes = await asyncio.gather(*[get_item_size(store_path.store, item) for item in all_items])
total = sum(all_sizes)
```

</details>



[worldcities.csv](https://github.com/user-attachments/files/16228393/worldcities.csv)

tomwhite:
Thanks for opening this PR @rabernat!

I just tried your implementation of `VLenUTF8Codec` with the VCF to Zarr conversion in https://github.com/sgkit-dev/bio2zarr/issues/254, and the test worked, which shows string encoding and decoding is working correctly in this case.

tomwhite:
> I am in favor of supporting this "legacy" encoding for backwards compatibility--I'd like existing Zarr V2 data to be convertible to V3 without rewriting the chunks.

+1

TomAugspurger:
Thanks. Would it be possible to support object-dtype arrays here too (I think zarr-v2 or xarray assumes object-dtype arrays hold strings)?

Right now this raises an exception.

```python
import zarr
import zarr.store
import numpy as np
import json
# from zarr.codecs import VLenUTF8Codec

store = zarr.store.MemoryStore(mode="a")
data = np.array(['a', 'bc', 'def', 'asdf', 'asdfg'], dtype=np.dtype(object))

arr = zarr.create(shape=data.shape, store=store, zarr_format=2, filters=[{"id": "vlen-utf8"}], fill_value="", dtype=object)
arr[:] = data
arr2 = zarr.open_array(store=store, zarr_format=2)
print(arr2[:])
```

Something like this seems to do the trick for `zarr_format=2`:

```diff
diff --git a/src/zarr/codecs/_v2.py b/src/zarr/codecs/_v2.py
index cc6129e6..c1f4944e 100644
--- a/src/zarr/codecs/_v2.py
+++ b/src/zarr/codecs/_v2.py
@@ -37,7 +37,13 @@ class V2Compressor(ArrayBytesCodec):
 
         # ensure correct dtype
         if str(chunk_numpy_array.dtype) != chunk_spec.dtype:
-            chunk_numpy_array = chunk_numpy_array.view(chunk_spec.dtype)
+            if chunk_spec.dtype.kind == "O":
+                # I think we need to assert something about the codec pipeline here
+                # to ensure that we're followed by something that'll do the cast properly.
+                # chunk_numpy_array = chunk_numpy_array.astype(chunk_spec.dtype)
+                pass
+            else:
+                chunk_numpy_array = chunk_numpy_array.view(chunk_spec.dtype)
 
         return get_ndbuffer_class().from_numpy_array(chunk_numpy_array)
 
diff --git a/src/zarr/core/buffer/core.py b/src/zarr/core/buffer/core.py
index 9a808e08..034711d9 100644
--- a/src/zarr/core/buffer/core.py
+++ b/src/zarr/core/buffer/core.py
@@ -472,7 +472,7 @@ class NDBuffer:
         # use array_equal to obtain equal_nan=True functionality
         _data, other = np.broadcast_arrays(self._data, other)
         return np.array_equal(
-            self._data, other, equal_nan=equal_nan if self._data.dtype.kind not in "US" else False
+            self._data, other, equal_nan=equal_nan if self._data.dtype.kind not in "OUS" else False
         )
 
     def fill(self, value: Any) -> None:
```

rabernat:
We could do that if we assume that the object dtype only ever hold strings. But that seems like a pretty extreme assumption.

Zarr V2 allows arbitrary python objects, but V3 does not. Seems reasonable to force users to use an explicit string dtype for string arrays.

What's the use case you have in mind for object string arrays?

TomAugspurger:
I'll look a bit closer tomorrow, but this was for reading zarr v2 data (an xarray dataset with an object-dtype coordinate array full of strings).

I agree that being explicit about the dtype here is better. Hopefully we can find a way that works well for older v2 datasets without too much hassle.

rabernat:
Here's a little exploration of different types of numpy string arrays

```python
import numpy as np
strings = ["hello", "world", "this", "is", "a", "test"]
# default
a1 = np.array(strings)
assert a1.dtype == '<U5'
# numpy 2.0 vlen string
a2 = np.array(strings, dtype='T')
# object
a3 = np.array(strings, dtype='O')

# the first two are detectable as string arrays
assert np.issubdtype(np.str_, a1.dtype)
assert np.issubdtype(np.str_, a2.dtype)

# the object is not, even though it contains all strings
assert not np.issubdtype(np.str_, a3.dtype)

# we can cast the object dtype to a vlen string or regular string
assert np.issubdtype(np.str_, a3.astype('T').dtype)

# however, we want to guard against casting if the array contains non-string elements
a4 = np.array(strings + [1], dtype='O') 
a4.astype('T')  # works
a4.astype('T', casking="same_kind")  # -> TypeError
```

So my proposal for object arrays would be to try to cast them using `.astype('T', casking="same_kind")` and, if successful, use string encoding.

@TomAugspurger - do you have a test case you are using for object arrays? Is my example enough to cover what Xarray needs? Under what circumstances does Xarray produce object arrays?

---

Also, my code is relying on numpy 2.0 vlen-string arrays. I guess that's not a hard dependency for Zarr. 🤔

TomAugspurger:
> do you have a test case you are using for object arrays?

In xarray, `xarray/tests/test_backends.py::TestZarrDictStore::test_roundtrip_object_dtype` is the failing test: https://github.com/pydata/xarray/blob/095d47fcb036441532bf6f5aed907a6c4cfdfe0d/xarray/tests/test_backends.py#L506


```python
 In [10]: import xarray as xr, numpy as np, zarr

In [11]: ds = xr.Dataset({"a": xr.DataArray(np.array(['a', 'b', 'c'], dtype=object))})

In [12]: store = zarr.store.MemoryStore(mode="a")

In [13]: ds.to_zarr(store)
```

Those are manually set to object dtype. But that same test also creates float arrays, so there's something more sophisticated that "object dtype == stings". I'll do some more looking as I work through those tests, but I think this PR can proceed independently.

TomAugspurger:
OK, I think I understand how v2 did things a bit better now.

1. It's xarray that treats object-dtype arrays as strings, *after validating that's actually true*. If so, then xarray sets the `dtype` passed to Zarr to `str`: https://github.com/pydata/xarray/blob/5c6418219e551cd59f81e3d1c6a828c1c23cd763/xarray/backends/zarr.py#L899-L900
2. Then zarr turns `dtype=str` *back* into `dtype=object` with https://github.com/zarr-developers/zarr-python/blob/6b9ab9e29f526986c5c7519dd4063546d5a81d14/zarr/util.py#L187-L192 (assuming an object codec is provided, which xarray does)

The net result is that in zarr-python 2.x writing zarr v2 data, here's the compressor and filters for various dtypes:

| dtype       | compressor | filters  |
| ----------- | ---------- | -------- |
| strings (U) | Blosc      |          |
| bytes (S)   | Blosc      |          |
| object[str] (O)  | Blosc      | VLenUTF8 |

rabernat:
Thanks Tom. Planning to work on this today.

mkitti:
Could we use the existing [sharding codec](https://zarr-specs.readthedocs.io/en/latest/v3/codecs/sharding-indexed/v1.0.html) for new variable length data? Essentially each variable length element is a just an "inner chunk" with some offset and length (nbytes)?

rabernat:
I'm not planning on doing anything along those lines here. The "variable length" we're talking about here is on the order of bytes. Doesn't make sense to compress each string individually. 

Just trying to get to basic feature parity with Zarr V2, not reinvent anything right now. Plenty of room to experiment with new ideas in the future.

rabernat:
I do not understand why the tests are failing in CI on this

```
tests/v3/test_codecs/test_vlen.py:10: in <module>
    from zarr.store.common import StorePath
E   ModuleNotFoundError: No module named 'zarr.store'
```

This is just boilerplate I copied from the other codec tests. Works fine locally. Help please @d-v-b or @TomAugspurger! 🙏

d-v-b:
> I do not understand why the tests are failing in CI on this
> 
> ```
> tests/v3/test_codecs/test_vlen.py:10: in <module>
>     from zarr.store.common import StorePath
> E   ModuleNotFoundError: No module named 'zarr.store'
> ```
> 
> This is just boilerplate I copied from the other codec tests. Works fine locally. Help please @d-v-b or @TomAugspurger! 🙏

it's [`zarr.storage`](https://github.com/zarr-developers/zarr-python/tree/v3/src/zarr/storage) now!

d-v-b:
I would really love to have https://github.com/zarr-developers/zarr-specs/pull/312 resolved at the same time as this PR

rabernat:
This seems to be working now at a basic level. 

I have questions about the API though.
- How do we "tell" Zarr that an input object array is supposed to be a string? Can we assume that any object array should be `string` dtype?
- Should we automatically set the array-to-bytes codec to `VLenUTF8Codec` when storing string data?

If possible, I'd propose addressing those questions in a separate PR, as this one is already rather long, complicated.

Also, #2278 needs to go in first.

src/zarr/core/buffer/core.py
```diff
@@ -467,9 +466,12 @@ def all_equal(self, other: Any, equal_nan: bool = True) -> bool:
             # Handle None fill_value for Zarr V2
             return False
         # use array_equal to obtain equal_nan=True functionality
+        # Note from Ryan: doesn't this lead to a huge amount of unnecessary memory allocation on every single chunk?
```
TomAugspurger, line 469: I think we're OK.

```python
In [10]: a, b = np.broadcast_arrays(np.ones(10), 2)

In [11]: b.base
Out[11]: array(2)
```

I think that it just does some tricks with the strides or something?

src/zarr/core/metadata/v3.py
```diff
@@ -445,6 +448,7 @@ class DataType(Enum):
     float64 = "float64"
     complex64 = "complex64"
     complex128 = "complex128"
+    string = "string"
```
TomAugspurger, line 451: Is this the value that shows up in `zarr.json`? If so I think we should be a bit careful here to not claim this as *the* string data type. Maybe call this `vlen-utf8`?

As I write this, I realize I don't really understand the relationship between what exactly effect `data_type` has, so maybe this isn't so important.

TomAugspurger: > From my point of view, the dtype says what sort of value the data should be cast to after all the codecs are done decoding.

Thanks, this is the bit I was missing earlier.

One perspective I like is Apache Parquet's distinction between [Logical Types](https://github.com/apache/parquet-format/blob/master/LogicalTypes.md) and physical ("primitive") Types. I think in this case our `data_type` field is logical, and so `string` is appropriate. Does that sound right?

rabernat: I think Parquet's framework is useful. But it feels a bit unclear how it applies in Zarr. From the point of view of a codec pipeline, the only physical datatype is `bytes`, right? We basically have `in_memory_data -> [encoding pipeline] -> bytes_to_store`? At what point is data translated from logical to physical?

Bringing Xarray into the picture makes it even harder, because Xarray does its own coding (e.g. from int to datetime).

I think a "logical type" framework like this would help Zarr a lot, but it feels like a big project to fully implement.

src/zarr/core/buffer/core.py
```diff
@@ -467,9 +466,12 @@ def all_equal(self, other: Any, equal_nan: bool = True) -> bool:
...
+        # Since fill-value is a scalar, isn't there a faster path than allocating a new array for fill value
+        # every single time we have to write data?
         _data, other = np.broadcast_arrays(self._data, other)
         return np.array_equal(
-            self._data, other, equal_nan=equal_nan if self._data.dtype.kind not in "US" else False
+            self._data, other, equal_nan=equal_nan if self._data.dtype.kind not in "USTO" else False
```
TomAugspurger, line 474: Is the `T` in the `dtype.kind`s here a typo? I don't see it at https://numpy.org/doc/2.0/reference/generated/numpy.dtype.kind.html.

TomAugspurger (approved):

src/zarr/core/buffer/core.py
rabernat: `np.dtypes.StringDType().kind` => `'T'`

rabernat: Actually, it's interesting that you flagged that here, because StringDType has some cool support for missing data: https://numpy.org/devdocs/user/basics.strings.html#missing-data-support

Not quite sure how we should be leveraging that in Zarr.

src/zarr/core/metadata/v3.py
LDeakin: @rabernat
> Also aware the we have some ugly coupling between a codec and a type.

Re: https://github.com/zarr-developers/zeps/pull/47#issuecomment-2238480835.

I still think it is worth considering the consolidation of `vlen-utf8`, `vlen-bytes`, `vlen-array` into a new codec name in Zarr V3 metadata to fix this coupling. These codecs all encode the data the same, just interleaved lengths and bytes.

I think `vlen_v2` is a good candidate name to indicate it is legacy and discourage use in Zarr V3[^1], similarly to the `v2` chunk key encoding. In `zarrs`, I use the same codec implementation whether it is a V2 array with `vlen-*` or a V3 array with `vlen_v2`, it is just a [metadata change](https://github.com/LDeakin/zarrs/blob/10d3b80e0fbf98515eb14e4197f011d0301ec196/zarrs_metadata/src/v2_to_v3.rs#L148-L154).

> Lachlan, what would the metadata have to look like to be interoperable with your zarrs implmentation?

It is compatible now with a change in [runtime config](https://github.com/LDeakin/zarrs/blob/10d3b80e0fbf98515eb14e4197f011d0301ec196/zarrs/src/config.rs#L126). I can change the default depending on what gets spec'd.

[^1]: Out of the scope of this PR, but I hope `zarr-python` can get a codec like what @jbms proposed [here](https://github.com/zarr-developers/zeps/pull/47#issuecomment-1710505141) (post v3.0.0?) as the preferred `vlen` codec going forward. Unlike the V2 `vlen-*` codecs, it can support partial decoding. I have experimental support for such a codec in `zarrs`.

rabernat:
To move forward with Xarray compatibility, we also need to do bytestrings. I think that should probably be a separate PR.

tests/v3/test_codecs/test_vlen.py
```diff
@@ -0,0 +1,81 @@
...
+        chunk_shape=data.shape,
+        dtype=data.dtype,
+        fill_value=b"",
+        codecs=[VLenBytesCodec()],
+    )
+    assert isinstance(a.metadata, ArrayV3Metadata)  # needed for mypy
+
+    a[:, :] = data
+    assert np.array_equal(data, a[:, :])
+    # assert a.metadata.data_type == DataType.string
```
TomAugspurger, line 73: Can you expand on why these are commented out? Should this be `DataType.bytes`?

TomAugspurger (commented):
The `bytes` changes also look good, just one small question.

tests/v3/test_codecs/test_vlen.py
rabernat: good catch, will fix

src/zarr/core/metadata/v3.py
```diff
@@ -359,9 +394,18 @@ def parse_fill_value(
...
+        return b""
+    else:
+        return dtype.to_numpy().type(0)
+
+
 def parse_fill_value(
-    fill_value: complex | str | bytes | np.generic | Sequence[Any] | bool | None,
-    dtype: BOOL_DTYPE | INTEGER_DTYPE | FLOAT_DTYPE | COMPLEX_DTYPE | np.dtype[Any],
+    fill_value: complex | str | bytes | np.generic | Sequence[Any] | bool,
+    dtype: DataType,
```
rabernat, line 408: I have changed this function to accept a Zarr v3 `DataType` rather than a numpy dtype.
All the tests are working. However, mypy is not happy.

```
src/zarr/core/metadata/v3.py:219: error: No overload variant of "parse_fill_value" matches argument types "Any", "DataType"  [call-overload]
src/zarr/core/metadata/v3.py:219: note: Possible overload variants:
src/zarr/core/metadata/v3.py:219: note:     def parse_fill_value(fill_value: complex | str | bytes | generic | Sequence[Any] | bool, dtype: BoolDType) -> bool
src/zarr/core/metadata/v3.py:219: note:     def parse_fill_value(fill_value: complex | str | bytes | generic | Sequence[Any] | bool, dtype: Int8DType | Int16DType | Int32DType | Int64DType | UInt8DType | UInt16DType | UInt32DType | UInt64DType) -> signedinteger[_8Bit] | signedinteger[_16Bit] | signedinteger[_32Bit] | signedinteger[_64Bit] | unsignedinteger[_8Bit] | unsignedinteger[_16Bit] | unsignedinteger[_32Bit] | unsignedinteger[_64Bit]
src/zarr/core/metadata/v3.py:219: note:     def parse_fill_value(fill_value: complex | str | bytes | generic | Sequence[Any] | bool | None, dtype: Float16DType | Float32DType | Float64DType) -> floating[_16Bit] | floating[_32Bit] | floating[_64Bit]
src/zarr/core/metadata/v3.py:219: note:     def parse_fill_value(fill_value: complex | str | bytes | generic | Sequence[Any] | bool | None, dtype: Complex64DType | Complex128DType) -> complexfloating[_32Bit, _32Bit] | complexfloating[_64Bit, _64Bit]
src/zarr/core/metadata/v3.py:219: note:     def parse_fill_value(fill_value: complex | str | bytes | generic | Sequence[Any] | bool | None, dtype: dtype[Any]) -> Any
src/zarr/core/metadata/v3.py:354: error: An overloaded function outside a stub file must have an implementation  [no-overload-impl]
src/zarr/core/metadata/v3.py:403: error: Returning Any from function declared to return "str | bytes | generic"  [no-any-return]
src/zarr/core/metadata/v3.py:406: error: Name "parse_fill_value" already defined on line 354  [no-redef]
src/zarr/core/metadata/v3.py:442: error: Incompatible types in assignment (expression has type "Complex64DType | Complex128DType", variable has type "DataType")  [assignment]
src/zarr/core/array.py:321: error: Argument 1 to "get_default_array_bytes_codec" has incompatible type "dtype[Any] | type[Any] | _SupportsDType[dtype[Any]] | str | tuple[Any, int] | tuple[Any, SupportsIndex | Sequence[SupportsIndex]] | list[Any] | _DTypeDict | tuple[Any, Any] | None"; expected "dtype[Any]"  [arg-type]
```

Seeking advice and what is the goal of the this `@overload` stuff and how it might be refactored to work with `DataType` rather than numpy dtypes.

src/zarr/core/metadata/v3.py
```diff
@@ -297,71 +332,99 @@ def update_attributes(self, attributes: dict[str, JSON]) -> Self:
...
+# INTEGER_DTYPE = (
+#     np.dtypes.Int8DType
+#     | np.dtypes.Int16DType
+#     | np.dtypes.Int32DType
+#     | np.dtypes.Int64DType
+#     | np.dtypes.UInt8DType
+#     | np.dtypes.UInt16DType
+#     | np.dtypes.UInt32DType
+#     | np.dtypes.UInt64DType
+# )
```
rabernat, line 351: I am attempting to replace these numpy dtypes with our own internal dtypes using literals (trying to follow this advice - https://github.com/python/typing/issues/781).

It does not seem to be working.

src/zarr/core/metadata/v3.py
```diff
@@ -297,71 +332,99 @@ def update_attributes(self, attributes: dict[str, JSON]) -> Self:
...
+# # If you want the specialized types rather than Any, you need to use `np.dtypes.<dtype>`
+# # rather than np.dtypes(<type>)
+# ...
 
 
 def parse_fill_value(
-    fill_value: complex | str | bytes | np.generic | Sequence[Any] | bool | None,
-    dtype: BOOL_DTYPE | INTEGER_DTYPE | FLOAT_DTYPE | COMPLEX_DTYPE | np.dtype[Any],
+    fill_value: complex | str | bytes | np.generic | Sequence[Any] | bool,
+    dtype_value: ALL_DTYPES,
```
rabernat, line 427: Mypy is not happy here.

```
src/zarr/core/metadata/v3.py:220: error: No overload variant of "parse_fill_value" matches argument types "Any", "Literal['string', 'bytes'] | Literal['int8', 'int16', 'int32', 'int64', 'uint8', 'uint16', 'uint32', 'uint64'] | Literal['float16', 'float32', 'float64'] | Literal['complex64', 'complex128']"  [call-overload]
src/zarr/core/metadata/v3.py:220: note: Possible overload variants:
src/zarr/core/metadata/v3.py:220: note:     def parse_fill_value(fill_value: complex | str | bytes | generic | Sequence[Any] | bool, dtype: Literal['bool']) -> bool
src/zarr/core/metadata/v3.py:220: note:     def parse_fill_value(fill_value: complex | str | bytes | generic | Sequence[Any] | bool, dtype: Literal['int8', 'int16', 'int32', 'int64', 'uint8', 'uint16', 'uint32', 'uint64']) -> signedinteger[_8Bit] | signedinteger[_16Bit] | signedinteger[_32Bit] | signedinteger[_64Bit] | unsignedinteger[_8Bit] | unsignedinteger[_16Bit] | unsignedinteger[_32Bit] | unsignedinteger[_64Bit]
src/zarr/core/metadata/v3.py:220: note:     def parse_fill_value(fill_value: complex | str | bytes | generic | Sequence[Any] | bool, dtype: Literal['float16', 'float32', 'float64']) -> floating[_16Bit] | floating[_32Bit] | floating[_64Bit]
src/zarr/core/metadata/v3.py:220: note:     def parse_fill_value(fill_value: complex | str | bytes | generic | Sequence[Any] | bool, dtype: Literal['complex64', 'complex128']) -> complexfloating[_32Bit, _32Bit] | complexfloating[_64Bit, _64Bit]
src/zarr/core/metadata/v3.py:220: note:     def parse_fill_value(fill_value: complex | str | bytes | generic | Sequence[Any] | bool, dtype: Literal['string']) -> str_
src/zarr/core/metadata/v3.py:220: note:     def parse_fill_value(fill_value: complex | str | bytes | generic | Sequence[Any] | bool, dtype: Literal['bytes']) -> bytes_
src/zarr/core/metadata/v3.py:425: error: Overloaded function implementation does not accept all possible arguments of signature 1  [misc]
src/zarr/core/metadata/v3.py:425: error: Overloaded function implementation does not accept all possible arguments of signature 2  [misc]
src/zarr/core/metadata/v3.py:425: error: Overloaded function implementation does not accept all possible arguments of signature 3  [misc]
src/zarr/core/metadata/v3.py:425: error: Overloaded function implementation does not accept all possible arguments of signature 4  [misc]
src/zarr/core/metadata/v3.py:425: error: Overloaded function implementation does not accept all possible arguments of signature 5  [misc]
src/zarr/core/metadata/v3.py:425: error: Overloaded function implementation does not accept all possible arguments of signature 6  [misc]
src/zarr/core/metadata/v3.py:514: error: Returning Any from function declared to return "str | bytes | generic"  [no-any-return]
```

rabernat (commented):
I have this PR in a place where I am pretty happy with it. I have mostly refactored the internal code to use our own `DataType` enum instead of numpy dtypes. This is necessary because the vlen types (strings, bytestrings) do not have such a simple 1:1 relationship with numpy dtypes.

However, I am thoroughly stuck on typing stuff. I am not able to recreate the sort of overloaded strict typing that was previously used on on `parse_fill_value`.

I would greatly appreciate some help and advice from whoever wrote the `parse_fill_value` type signatures. 🙏

src/zarr/core/metadata/v3.py
TomAugspurger: I think 7d9d89736245c8e547c9cda0b70683de74982dfd fixed this. Two things:

1. Some of the overloads used `dtype` and some used `dtype` and some used `dtype_value`. They need to match
2. `BOOL_DTYPE` wasn't present in `ALL_DTYPES`, and mypy's error message didn't make this clear :/ I ended up running that file through pyright which made it a bit easier to see:

```
  /Users/tom/gh/zarr-developers/zarr-python/src/zarr/core/metadata/v3.py:410:5 - error: Overloaded implementation is not consistent with signature of overload 1
    Type "(fill_value: Any, dtype: ALL_DTYPES) -> Any" is not assignable to type "(fill_value: complex | str | bytes | generic | Sequence[Any] | bool, dtype: BOOL_DTYPE) -> BOOL"
      Parameter 2: type "BOOL_DTYPE" is incompatible with type "ALL_DTYPES"
        Type "BOOL_DTYPE" is not assignable to type "ALL_DTYPES"
          "BOOL_DTYPE" is not assignable to type "Literal['int8']"
          "BOOL_DTYPE" is not assignable to type "Literal['int16']"
          "BOOL_DTYPE" is not assignable to type "Literal['int32']"
          "BOOL_DTYPE" is not assignable to type "Literal['int64']"
          "BOOL_DTYPE" is not assignable to type "Literal['uint8']"
    ... (reportInconsistentOverload)
15 errors, 0 warnings, 0 informations 
```

rabernat: Thanks Tom! Really appreciate your help here!

src/zarr/strings.py
```diff
@@ -0,0 +1,87 @@
...
+from typing import Any, cast
+from warnings import warn
+
+import numpy as np
+
+# STRING_DTYPE is the in-memory datatype that will be used for V3 string arrays
+# when reading data back from Zarr.
+# Any valid string-like datatype should be fine for *setting* data.
+
+STRING_DTYPE: np.dtypes.StringDType | np.dtypes.ObjectDType
```
rabernat, line 14: I'm stuck on a tricky typing issue. However, the `np.dtypes.ObjectDType` is only available in numpy 2.0.

Is there a way to have the typing somehow depend on whether numpy 2.0 is available in the environment?

rabernat:
Do not know why RTD build is failing, but otherwise this is GTG.

jhamman:
> Do not know why RTD build is failing, but otherwise this is GTG.

This warning is causing the failure:

```
/home/docs/checkouts/readthedocs.org/user_builds/zarr/checkouts/2036/docs/_autoapi/zarr/strings/index.rst:73: WARNING: duplicate object description of zarr.strings.STRING_DTYPE, other instance in _autoapi/zarr/strings/index, use :no-index: for one of them
```

src/zarr/core/buffer/core.py
```diff
@@ -313,8 +313,7 @@ class NDBuffer:
     """
 
     def __init__(self, array: NDArrayLike) -> None:
-        # assert array.ndim > 0
-        assert array.dtype != object
+        # assert array.dtype != object
```
jhamman, line 316:
```suggestion
```

src/zarr/codecs/__init__.py
```diff
@@ -21,5 +28,19 @@
     "ShardingCodec",
     "ShardingCodecIndexLocation",
     "TransposeCodec",
+    "VLenUTF8Codec",
+    "VLenBytesCodec",
     "ZstdCodec",
 ]
+
+
+def get_default_array_bytes_codec(
```
jhamman, line 37: Let's mark this as private for now

```suggestion
def _get_default_array_bytes_codec(
```

src/zarr/codecs/vlen_utf8.py
```diff
@@ -0,0 +1,117 @@
...
+
+if TYPE_CHECKING:
+    from typing import Self
+
+    from zarr.core.array_spec import ArraySpec
+
+
+# can use a global because there are no parameters
+vlen_utf8_codec = VLenUTF8()
+vlen_bytes_codec = VLenBytes()
```
jhamman, line 23:
```suggestion
_vlen_utf8_codec = VLenUTF8()
_vlen_bytes_codec = VLenBytes()
```

src/zarr/core/buffer/core.py
jhamman, line 469:
```suggestion
```

src/zarr/core/metadata/v3.py
```diff
@@ -376,29 +410,39 @@ def parse_fill_value(
...
-        return dtype.type(0)
+        raise ValueError("Fill value cannot be None")
+    if data_type == DataType.string:
+        return np.str_(fill_value)
+    if data_type == DataType.bytes:
+        return np.bytes_(fill_value)
+
+    # the rest are numeric types
+    np_dtype = data_type.to_numpy()
+    np_dtype = cast(np.dtype[np.generic], np_dtype)
```
jhamman, line 430:
```suggestion
    np_dtype = cast(np.dtype[np.generic], data_type.to_numpy())
```

src/zarr/core/metadata/v3.py
```diff
@@ -376,29 +410,39 @@ def parse_fill_value(
...
+
     if isinstance(fill_value, Sequence) and not isinstance(fill_value, str):
-        if dtype.type in (np.complex64, np.complex128):
-            dtype = cast(COMPLEX_DTYPE, dtype)
+        if data_type in (DataType.complex64, DataType.complex128):
+            # dtype = cast(np.dtypes.Complex64DType | np.dtypes.Complex128DType, np_dtype)
```
jhamman, line 434:
```suggestion
```

src/zarr/core/metadata/v3.py
```diff
@@ -409,27 +453,38 @@ def parse_fill_value(
...
 
 
+def default_fill_value(dtype: DataType) -> str | bytes | np.generic:
+    if dtype == DataType.string:
+        return ""
+    elif dtype == DataType.bytes:
+        return b""
+    else:
+        np_dtype = dtype.to_numpy()
+        np_dtype = cast(np.dtype[np.generic], np_dtype)
```
jhamman, line 484:
```suggestion
        np_dtype = cast(np.dtype[np.generic], dtype.to_numpy())
```

tests/v3/test_codecs/test_vlen.py
```diff
@@ -0,0 +1,128 @@
...
+
+    # test round trip
+    b = Array.open(sp)
+    assert isinstance(b.metadata, ArrayV3Metadata)  # needed for mypy
+    assert np.array_equal(data, b[:, :])
+    assert b.metadata.data_type == DataType.bytes
+    assert a.dtype == "O"
+
+
+# TODO: move these tests out of codecs and into a more appropriate location
```
jhamman, line 98: 👀 

test_array seems like a good place for these

jhamman (approved):
Really cool @rabernat!

I have a handful of suggestions, mostly cosmetic.

PR #2046, opened 2024-07-19
Bump fsspec version in tests

dstansby:
I'm trying to debug https://github.com/zarr-developers/zarr-python/issues/1679 - if we're lucky, there might just be some bad versions of fsspec we need to exclude.


dstansby:
Just realised these failiures will be fixed by https://github.com/zarr-developers/zarr-python/pull/1970 🤦

PR #2053, opened 2024-07-23
Apply and enforce more ruff rules

DimitriPapadopoulos:
* Apply [`PERF401`](https://docs.astral.sh/ruff/rules/manual-list-comprehension/), but do not enforce for now - see https://github.com/pypa/setuptools/pull/4449#issuecomment-2239482683 and https://github.com/numpy/numpy/pull/26876#issuecomment-2223009885 and make your own opinion :smile:
* Apply [`PT022`](https://docs.astral.sh/ruff/rules/pytest-useless-yield-fixture/). Do not enforce for now, I fear you won't like other [`PT`](https://docs.astral.sh/ruff/rules/#flake8-pytest-style-pt) rules.
* ~~Enforce [`TCH`](https://docs.astral.sh/ruff/rules/#flake8-type-checking-tch) rules.~~
* Enforce [`PGH`](https://docs.astral.sh/ruff/rules/#pygrep-hooks-pgh) rules.
* Apply [`C405`](https://docs.astral.sh/ruff/rules/unnecessary-literal-set/) and [`C417`](https://docs.astral.sh/ruff/rules/unnecessary-map/). Do not enforce for now, some [`C4`](https://docs.astral.sh/ruff/rules/#flake8-comprehensions-c4) rules seem too strict.
* Enforce [`PYI`](https://docs.astral.sh/ruff/rules/#flake8-pyi-pyi) rules.
* ~~Fix pre-commit warnings and errors as they occur~~
* ~~Catch more precise exceptions than `Exception`~~

TODO:

### DIFF
--- a/tests/v3/test_indexing.py
+++ b/tests/v3/test_indexing.py
@@ -1782,9 +1782,9 @@ async def test_accessed_chunks(
 
         # Combine and generate the cartesian product to determine the chunks keys that
         # will be accessed
-        chunks_accessed = []
-        for comb in itertools.product(*chunks_per_dim):
-            chunks_accessed.append(".".join([str(ci) for ci in comb]))
+        chunks_accessed = [
+            ".".join([str(ci) for ci in comb]) for comb in itertools.product(*chunks_per_dim)
+        ]
 
         counts_before = store.counter.copy()
 
--- a/tests/v3/test_indexing.py
+++ b/tests/v3/test_indexing.py
@@ -36,7 +36,7 @@ if TYPE_CHECKING:
 
 @pytest.fixture
 async def store() -> AsyncGenerator[StorePath]:
-    yield StorePath(await MemoryStore.open(mode="w"))
+    return StorePath(await MemoryStore.open(mode="w"))
 
 
 def zarr_array_from_numpy_array(
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -211,6 +211,7 @@ extend-select = [
     "I",   # isort
     "ISC", # flake8-implicit-str-concat
     "PGH", # pygrep-hooks
+    "PYI", # flake8-pyi
     "RSE", # flake8-raise
     "RUF",
     "TCH", # flake8-type-checking
@@ -218,6 +219,7 @@ extend-select = [
     "UP",  # pyupgrade
 ]
 ignore = [
+    "PYI013",
     "RUF005",
     "TRY003",
     # https://docs.astral.sh/ruff/formatter/#conflicting-lint-rules
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -207,14 +207,15 @@ extend-exclude = [
 
 [tool.ruff.lint]
 extend-select = [
-    "B",  # flake8-bugbear
-    "I",  # isort
-    "ISC",
-    "UP",  # pyupgrade
-    "RSE",
+    "B",   # flake8-bugbear
+    "I",   # isort
+    "ISC", # flake8-implicit-str-concat
+    "PGH", # pygrep-hooks
+    "RSE", # flake8-raise
     "RUF",
     "TCH", # flake8-type-checking
     "TRY", # tryceratops
+    "UP",  # pyupgrade
 ]
 ignore = [
     "RUF005",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -219,6 +219,21 @@ extend-select = [
 ignore = [
     "RUF005",
     "TRY003",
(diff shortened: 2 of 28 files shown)


src/zarr/store/core.py
```diff
@@ -54,13 +56,11 @@ def __str__(self) -> str:
     def __repr__(self) -> str:
         return f"StorePath({self.store.__class__.__name__}, {str(self)!r})"
 
-    def __eq__(self, other: Any) -> bool:
+    def __eq__(self, other: object) -> bool:
         try:
-            if self.store == other.store and self.path == other.path:
-                return True
+            return bool(self.store == other.store and self.path == other.path)  # type: ignore[attr-defined]
```
DimitriPapadopoulos, line 61: I had to add the `bool` cast to avoid this mypy error:
```src/zarr/store/core.py:61: error: Returning Any from function declared to return "bool"  [no-any-return]
src/zarr/store/core.py:61: note: Error code "no-any-return" not covered by "type: ignore" comment
```
I must admit I don't understand it. How could `__eq__` return `Any` instead of `bool`?

src/zarr/store/core.py
```diff
@@ -54,13 +56,11 @@ def __str__(self) -> str:
...
         except Exception:
```
DimitriPapadopoulos, line 62: Perhaps catch only `AttributeError` instead of `Exception`?

src/zarr/store/core.py
jhamman: @dstansby - any ideas here?

src/zarr/store/core.py
jhamman: seems like a good improvement

jhamman (approved):
Everything here looks good @DimitriPapadopoulos. 

cc'ing @dstansby for a final review and look at the open mypy question

src/zarr/store/core.py
DimitriPapadopoulos: Applied `Exception` → `(AttributeError, TypeError)`

jhamman:
@DimitriPapadopoulos - lots of merge conflicts here. Mind taking this one from the top?

DimitriPapadopoulos:
It might be easier to restart from scratch. Let me have a look.

DimitriPapadopoulos:
OK, I have re-applied all rules from scratch, leaving `TCH` out for now. A specific PR for `TCH` might be easier to review and merge faster, avoiding new conflicts.

DimitriPapadopoulos:
Rebased.

jhamman:
@DimitriPapadopoulos - is this ready for another review?

DimitriPapadopoulos:
It's all good, except the question above: Why does MyPy think `==` could return `Any` instead of `bool`?

src/zarr/store/core.py
DimitriPapadopoulos: Outdated: That code has disappeared.

jhamman:
@DimitriPapadopoulos - give us a ping when this is ready for a final review.

DimitriPapadopoulos:
Because `other` is an `object`, the `other.store` and `other.path` attributes might be anything — if they exist at all. Yet, according to the [`__eq__` documentation](https://docs.python.org/3/reference/datamodel.html#object.__eq__):
* `self.store == other.store` is equivalent to `self.store.__eq__(other.store)` and `self.store` is a `Store` and `Store.__eq__` returns `bool`.
* `self.path == other.path` is equivalent to `self.path.__eq__(other.path)` and `self.path` is a `str`.

I ended up deciding this is a MyPy issue and ignoring  `no-any-return`.

DimitriPapadopoulos:
@jhamman This merge request seems ready for a review.

tests/v3/test_indexing.py
```diff
@@ -1782,9 +1782,9 @@ async def test_accessed_chunks(
 
         # Combine and generate the cartesian product to determine the chunks keys that
         # will be accessed
-        chunks_accessed = []
-        for comb in itertools.product(*chunks_per_dim):
-            chunks_accessed.append(".".join([str(ci) for ci in comb]))
+        chunks_accessed = [
+            ".".join([str(ci) for ci in comb]) for comb in itertools.product(*chunks_per_dim)
+        ]
```
jakirkham, line 1787: Maybe we can simplify this a bit further?

```suggestion
        chunks_accessed = [
            ".".join(map(str, comb)) for comb in itertools.product(*chunks_per_dim)
        ]
```

Please see PR: https://github.com/zarr-developers/zarr-python/pull/2229

PR #2058, opened 2024-07-29
Handle missing attributes key from metadata, and other fixes

### LINKED ISSUE #2024: allow `null` attributes
We currently error if the `attributes` field of `zarr.json` is missing. We should handle this instead of erroring.

reported by @joshmoore 

Edit: data was originally written by tensorstore

d-v-b:
- handle creating v3 metadata from a dict without an `attributes` key (`attributes` will be set to `None` in the resulting metadata document).
- handle null entries in `dimension_names` correctly at validation
- ensure that collection types become `tuple` during metadata serialization
- make a copy of the input dict in `from_dict` method of array v3 metadata, to avoid mutating the input dict
- allow creating `ChunkKeyEncoding` from dicts without `configuration`, because it's optional.
- add a test case that revealed the problems fixed by all of the above changes.

closes #2024 

TODO:

### DIFF
--- a/src/zarr/chunk_key_encodings.py
+++ b/src/zarr/chunk_key_encodings.py
@@ -39,17 +39,19 @@ class ChunkKeyEncoding(Metadata):
             return data
 
         # configuration is optional for chunk key encodings
-        name_parsed, configuration_parsed = parse_named_configuration(
-            data, require_configuration=False
-        )
-        # normalize missing configuration to the default "/" separator.
-        if configuration_parsed is None:
-            configuration_parsed = {"separator": "/"}
+        name_parsed, config_parsed = parse_named_configuration(data, require_configuration=False)
         if name_parsed == "default":
-            return DefaultChunkKeyEncoding(**configuration_parsed)  # type: ignore[arg-type]
+            if config_parsed is None:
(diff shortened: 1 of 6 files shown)


src/zarr/chunk_key_encodings.py
```diff
@@ -38,7 +38,13 @@ def from_dict(cls, data: dict[str, JSON] | ChunkKeyEncoding) -> ChunkKeyEncoding
         if isinstance(data, ChunkKeyEncoding):
             return data
 
-        name_parsed, configuration_parsed = parse_named_configuration(data)
+        # configuration is optional for chunk key encodings
+        name_parsed, configuration_parsed = parse_named_configuration(
+            data, require_configuration=False
+        )
+        # normalize missing configuration to the default "/" separator.
```
normanrz, line 45: The default for `default` is `/`. For `v2` it is `.`

d-v-b: good catch, I fixed this in [404b60a](https://github.com/zarr-developers/zarr-python/pull/2058/commits/404b60aa65b3c6a07634557d5b8ddce926c2a285)

normanrz (approved):

PR #2059, opened 2024-07-31
refactor metadata into `metadata.v2` and `metadata.v3` modules

d-v-b:
In `main`, we handle v2 and v3 metadata differences by having functions with names like `function_purpose_v2` and `function_purpose_v3`. I think it would be more clear to collect all the v2- and v3-specific routines into separate modules.

This PR creates new `metadata.v2` and `metadata.v3` modules for the metadata classes, as well as a `metadata.common` module, for some routines that have identical v2 and v3 semantics. Functions that previously signified v2-ness or v3-ness in their name now signify it from their location. In this PR The only exports from `zarr.metadata` are `ArrayV2Metadata` and `ArrayV3Metadata`.

I made a few other related changes:

- in `main`, we annotate the `metadata` attribute of `AsyncArray` with the abstract array metadata base class. I changed these annotations to `ArrayV2Metadata | ArrayV3Metadata`, which should be much more useful. In a later effort we could consider making `AsyncArray` generic over the type of array metadata to give even more information to the type system.

- In `main` v2 metadata creation relies on an import from `zarr.v2`. This will not be valid after v2 is gone, so I ported the `v2.normalize_fill_value` routine into `zarr.metadata.v2.parse_fill_value`, and made some minor internal changes to the exceptions in that routine.

TODO:


src/zarr/metadata/v2.py
normanrz, line 1: Should this file also get an `__all__`?

d-v-b: I'm open to this, but I think it would only affect cases where people did `from zarr.metadata.v2 import *`; `from zarr.metadata.v2 import foo` would still work whether or not `foo` is in `__all__`.

src/zarr/metadata/v3.py
```diff
@@ -636,3 +314,83 @@ def parse_fill_value_v3(
...
+    uint8 = "uint8"
+    uint16 = "uint16"
+    uint32 = "uint32"
+    uint64 = "uint64"
+    float32 = "float32"
+    float64 = "float64"
+
+    @property
+    def byte_count(self) -> int:
+        data_type_byte_counts = {
```
jhamman, line 344: I know this wasn't introduced here but we should pull all of these dicts into module level variables. No need to recreate them every time we use the datatype.

src/zarr/array.py
```diff
@@ -92,14 +92,14 @@ def create_codec_pipeline(metadata: ArrayV2Metadata | ArrayV3Metadata) -> CodecP
 
 @dataclass(frozen=True)
 class AsyncArray:
-    metadata: ArrayMetadata
+    metadata: ArrayV3Metadata | ArrayV2Metadata
```
jhamman, line 95: I slightly favor reverting this change as a way to future proof this a bit.

jhamman (approved):
Nice cleanup @d-v-b.

d-v-b:
this was addressed by https://github.com/zarr-developers/zarr-python/pull/2163

PR #2062, opened 2024-07-31
Make zarr.config private (> zarr._config)

dstansby:
Fixes (partially) https://github.com/zarr-developers/zarr-python/issues/2037


jhamman:
See https://github.com/zarr-developers/zarr-python/pull/2072 for an alternative to this.

PR #2064, opened 2024-08-03
implement `store.list_prefix` and `store._set_many`

d-v-b:
- fixes / implements `list_prefix` for stores. `list_prefix(prefix=foo)` now consistently returns keys with the shared prefix `foo` stripped. I'd be fine altering this to return absolute keys instead.
- implements tests for `list_prefix` on the `StoreTests` base class
- adds a `_set_dict(Mapping[str, Buffer])` method to stores that allows declaring a collection of key: value pairs to write to storage. The primary usage is to make store tests simpler via a declarative API. I also suspect that making tests simpler anticipates making other code simpler. The default implementation of `_set_dict` simply wraps `store.set`, but it's easy to imagine fancier batching / transaction implementations.
- alters some of the store tests to use `_set_dict`.
- adds some missing type annotations in the remotestore tests

TODO:


d-v-b:
@martindurant i'm on flaky train internet and the github web API is giving me inconsistent signals about whether you have been requested for review; to be clear, I would appreciate your review :)

martindurant:
> list_prefix(prefix=foo) now consistently returns keys with the shared prefix foo stripped. I'd be fine altering this to return absolute keys instead.

What matters is how we intend to call it! fsspec likes to provide full paths (but has explicit prefix interfaces) - of course it's fine to tailor the situation to what we need, but I don't really know what that is. 

> I would appreciate your review

I may have got a notification or a few.

src/zarr/abc/store.py
```diff
@@ -221,6 +221,14 @@ def close(self) -> None:
         self._is_open = False
         pass
 
+    async def _set_dict(self, dict: Mapping[str, Buffer]) -> None:
+        """
+        Insert objects into storage as defined by a prefix: value mapping.
+        """
+        for key, value in dict.items():
+            await self.set(key, value)
```
martindurant, line 229: would prefer asyncio.gather as the default here, since each coroutine in principle may be latency bound

src/zarr/store/local.py
```diff
@@ -193,14 +193,10 @@ async def list_prefix(self, prefix: str) -> AsyncGenerator[str, None]:
         -------
         AsyncGenerator[str, None]
         """
-        for p in (self.root / prefix).rglob("*"):
-            if p.is_file():
-                yield str(p)
```
martindurant, line 198: Were we getting duplicates?

src/zarr/store/local.py
```diff
@@ -193,14 +193,10 @@ async def list_prefix(self, prefix: str) -> AsyncGenerator[str, None]:
...
-
         to_strip = str(self.root) + "/"
         for p in (self.root / prefix).rglob("*"):
             if p.is_file():
-                yield str(p).replace(to_strip, "")
+                yield str(p).removeprefix(to_strip)
```
martindurant, line 199: +1

src/zarr/store/remote.py
```diff
@@ -205,5 +205,6 @@ async def list_dir(self, prefix: str) -> AsyncGenerator[str, None]:
             yield onefile
 
     async def list_prefix(self, prefix: str) -> AsyncGenerator[str, None]:
-        for onefile in await self._fs._ls(prefix, detail=False):
-            yield onefile
+        find_str = "/".join([self.path, prefix])
+        for onefile in await self._fs._find(find_str):
```
martindurant, line 209: The defaults for find are: maxdepth=None, withdirs=False, detail=False; maybe good to be specific.

Why is find() better than ls()? The former will return all  child files, not just one level deep - is that the intent? If not, ls() ought to be generally more efficient.

src/zarr/sync.py
```diff
@@ -114,6 +114,23 @@ def _get_loop() -> asyncio.AbstractEventLoop:
     return loop[0]
 
 
+async def _collect_aiterator(data: AsyncIterator[T]) -> tuple[T, ...]:
+    """
+    Collect an entire async iterator into a tuple
+    """
+    result = []
+    async for x in data:
+        result.append(x)
```
martindurant, line 123: asyncio.gather? Like above, not much point in having coroutines if we serially wait for them.

src/zarr/testing/store.py
```diff
@@ -103,6 +104,18 @@ async def test_set(self, store: S, key: str, data: bytes) -> None:
...
+        """
+        Test that a dict of key : value pairs can be inserted into the store via the
+        `_set_dict` method.
+        """
+        keys = ["zarr.json", "c/0", "foo/c/0.0", "foo/0/0"]
+        data_buf = [Buffer.from_bytes(k.encode()) for k in keys]
+        store_dict = dict(zip(keys, data_buf, strict=True))
+        await store._set_dict(store_dict)
+        for k, v in store_dict.items():
+            assert self.get(store, k).to_bytes() == v.to_bytes()
```
martindurant, line 117: if x.to_bytes() == y.to_bytes(), does x== y?

Isn't there a multiple get? Maybe not important here.

tests/v3/test_store/test_remote.py
```diff
@@ -113,8 +107,8 @@ def store_kwargs(self, request) -> dict[str, str | bool]:
         raise AssertionError
 
     @pytest.fixture(scope="function")
-    def store(self, store_kwargs: dict[str, str | bool]) -> RemoteStore:
-        url = store_kwargs["url"]
+    async def store(self, store_kwargs: dict[str, str | bool | UPath]) -> RemoteStore:
```
martindurant, line 110: This isn't actually async

src/zarr/store/local.py
d-v-b: i think we were. this code path was not tested until this PR

src/zarr/store/remote.py
d-v-b: using `find` here is merely due to my ignorance of fsspec. I will implement `ls` as you suggest

martindurant: It depends on whether you want one directory level or everything below it. When I wrote the original, I didn't know the intent.

src/zarr/testing/store.py
d-v-b: > if x.to_bytes() == y.to_bytes(), does x== y?

no, and I _suspect_ this might be deliberate since in principle `Buffer` instances can have identical bytes but different devices (e.g., gpu memory vs host memory); thus `x == y` might only be true if two buffers are bytes-equal _and_ device-equal, but I'm speculating here. @madsbk would have a better answer I think.

> Isn't there a multiple get? Maybe not important here.

there is no multiple get (nor a multiple set, nor a multiple delete).

tests/v3/test_store/test_remote.py
d-v-b: correct, but the class we are inheriting from defines this as an async method

src/zarr/store/remote.py
d-v-b: i believe the intent here is to list everything below `prefix` (at least, that's how I'm using it)

src/zarr/sync.py
d-v-b: I don't think we can use `asyncio.gather` here, because `AsyncGenerator` is not iterable. Happy to be corrected, since I don't really know asyncio very well.

d-v-b: and for clarification, `_collect_aiterator` largely exists for convenience in testing, because I need some way to collect async generators when debugging with `pdb`. This function is not intended for use in anything performance sensitive.

src/zarr/abc/store.py
d-v-b: switched to asyncio.gather in e4101b7a

src/zarr/sync.py
martindurant: We should probably re-examine the use of async-iterators, though. If we can't gather() on them (seems to be true?), then they are the wrong abstraction since gather() is probably always what we actually want.

martindurant: On second thoughts, maybe I'm wrong - does `async for` schedule all the coroutines at one?? Should be easy to test.

d-v-b: I don't think it schedules them all at once. in `[x async for x in async_generator]`, `x` is not an awaitable; it's already awaited. since the basic model of the generator is that it's a resumable, stateful iterator, I don't think we can schedule all the tasks at once.

src/zarr/store/remote.py
d-v-b: I misunderstood your first comment. since the intent is to use the behavior of `_find`, I'm keeping it, but adding explicit kwargs as you suggested.

src/zarr/sync.py
jhamman: the idea with the generators is to a) support seamless pagination and b) support pipelining (`del_prefix` will be able to take advantage of this at some point).

src/zarr/abc/store.py
```diff
@@ -221,6 +224,13 @@ def close(self) -> None:
         self._is_open = False
         pass
 
+    async def _set_dict(self, dict: Mapping[str, Buffer]) -> None:
```
jhamman, line 227: `set_many()` (analogous to [`insert_many`](https://motor.readthedocs.io/en/stable/api-tornado/motor_collection.html#motor.motor_tornado.MotorCollection.insert_many))

src/zarr/testing/store.py
jhamman: xref: [src/zarr/buffer.py](https://github.com/zarr-developers/zarr-python/pull/2006/files#diff-3affa2119d5b734fc851307a6ddd275af09dc0650a97a95cc82fa7ae30272a27R258) in https://github.com/zarr-developers/zarr-python/pull/2006

src/zarr/abc/store.py
d-v-b: I moved away from set_dict and switched to `_set_many`

d-v-b:
we should try to get this in, because it fixes problematic `list_prefix` behavior that has been noted elsewhere.

jhamman (approved):

PR #2070, opened 2024-08-09
Stateful store tests

e-marshall:
This is a PR that implements stateful tests for Zarr MemoryStores. 

TODO items below:
- I added docstrings to new classes, functions
- I don't think any changes need to be made to tutorial.rst
- I wasn't sure if these changes should be added in release.rst ? 
- GH actions pass
- I wasn't sure where to find codecov results, but didn't see a notification that it failed

TODO:


src/zarr/strategies.py
```diff
@@ -143,3 +145,19 @@ def basic_indices(draw: st.DrawFn, *, shape: tuple[int], **kwargs):  # type: ign
...
+
+
+def key_ranges(keys: SearchStrategy | None = None) -> st.lists[:]:
+    """fn to generate key_ranges strategy for get_partial_values()
+    returns list strategy w/ form: [(key, (range_start, range_step)),
+                                    (key, (range_start, range_step)),...]
+    """
+    if keys is None:
+        group_st = st.text(alphabet=string.ascii_letters + string.digits, min_size=1, max_size=10)
+        keys = st.lists(group_st, min_size=1, max_size=5).map("/".join)
```
dcherian, line 157:
```suggestion
def key_ranges(keys: SearchStrategy = node_names) -> st.lists[:]:
    """fn to generate key_ranges strategy for get_partial_values()
    returns list strategy w/ form: [(key, (range_start, range_step)),
                                    (key, (range_start, range_step)),...]
    """
```

tests/v3/test_store/test_stateful_store.py
```diff
@@ -0,0 +1,206 @@
...
+    def __init__(self):  # look into using run_machine_as_test()
+        super().__init__()
+        self.model = {}
+        self.store = MemoryStore(mode="w")
+
+    # Unfortunately, hypothesis' stateful testing infra does not support asyncio
+    # So we redefine sync versions of the Store API.
+    # https://github.com/HypothesisWorks/hypothesis/issues/3712#issuecomment-1668999041
+    async def store_set(self, key, data_buffer):
+        await self.store.set(key, data_buffer)
```
dcherian, line 34: This stuff is independent of the test so lets move it out to a separate class.

```python
class SyncStoreWrapper():
    def __init__(self, store):
        self.store = store
	...
```

tests/v3/test_store/test_stateful_store.py
```diff
@@ -0,0 +1,206 @@
...
+        asyncio.run(self.store_set(key, data_buf))
+        self.model[key] = data_buf  # this was data
+
+    @invariant()
+    def check_paths_equal(self) -> None:
+        note("Checking that paths are equal")
+        paths = asyncio.run(self.store_list())
+        # note(f'(check paths equal) {self.model=}, {self.store._store_dict=}')
+        assert list(self.model.keys()) == paths
+        assert len(self.model.keys()) == len(self.store._store_dict)
```
dcherian, line 81:
```suggestion
        assert len(self.model.keys()) == len(self.store._store_dict)
```

We want this to test general Zarr stores, so you cannot use `_store_dict`

tests/v3/test_store/test_stateful_store.py
```diff
@@ -0,0 +1,206 @@
...
+        observed = []
+
+        for obs in obs_maybe:
+            assert obs is not None
+            observed.append(obs.to_bytes())
+
+        model_vals_ls = []
+
+        for idx in range(len(observed)):
+            key, byte_range = key_range[idx]
```
dcherian, line 110:
```suggestion
        for key, byte_range in key_range:
```

tests/v3/test_store/test_stateful_store.py
```diff
@@ -0,0 +1,206 @@
...
+            model_vals = self.model[key]
+            start = byte_range[0] or 0
+            step = byte_range[1]
+            stop = start + step if step is not None else None
+            model_vals_partial = model_vals[start:stop]
+            model_vals_ls.append(model_vals_partial)
```
dcherian, line 116:
```suggestion
            model_vals_ls.append(model_vals[start:stop])
```

tests/v3/test_store/test_stateful_store.py
```diff
@@ -0,0 +1,206 @@
...
+    def delete(self, data) -> None:
+        path_st = data.draw(st.sampled_from(sorted(self.model.keys())))
+        note(f"(delete) Deleting {path_st=}")
+
+        asyncio.run(self.delete_key(path_st))
+
+        del self.model[path_st]
+
+        # property tests
+        assert self.model.keys() == self.store._store_dict.keys()
```
dcherian, line 136:
```suggestion
        assert self.model.keys() == self.store._store_dict.keys()
```

Don't you have a version of this in your invariant already? It's going to be run after every rule...

tests/v3/test_store/test_stateful_store.py
```diff
@@ -0,0 +1,206 @@
...
+        assert path_st not in list(self.model.keys())
+
+    @rule(key=paths, data=st.binary(min_size=0, max_size=100))
+    def clear(self, key: str, data: bytes):
+        """clear() is in zarr/store/memory.py
```
dcherian, line 141: Use the [ABC](https://github.com/zarr-developers/zarr-python/blob/v3/src/zarr/abc/store.py) as reference. That interface guaranteed to be supported by every store. And that's the specification we're trying to test.

tests/v3/test_store/test_stateful_store.py
```diff
@@ -0,0 +1,206 @@
...
+        it calls clear on self._store_dict
+        clear() is dict method that removes all key-val pairs from dict
+        """
+        assert not self.store.mode.readonly
+        note("(clear)")
+        asyncio.run(self.store_clear())
+        self.model.clear()
+
+        assert self.model == self.store._store_dict
```
dcherian, line 150:
```suggestion
```

tests/v3/test_store/test_stateful_store.py
```diff
@@ -0,0 +1,206 @@
...
+        assert len(self.model.keys()) == 0
+
+    def empty(self, data) -> None:
+        """empty checks if a store is empty or not
+        return true if self._store_dict doesn't exist
+        return false if self._store_dict exists"""
+        note("(empty)")
+
+        asyncio.run(self.store_clear())
```
dcherian, line 159: The model version of this test is
- The store should be empty, if the model is empty. 
- The store should not be empty, if the model is not empty.

tests/v3/test_store/test_stateful_store.py
```diff
@@ -0,0 +1,206 @@
...
+        note("Checking that paths are equal")
+        paths = asyncio.run(self.store_list())
+        # note(f'(check paths equal) {self.model=}, {self.store._store_dict=}')
+        assert list(self.model.keys()) == paths
+        assert len(self.model.keys()) == len(self.store._store_dict)
+        assert self.model == self.store._store_dict
+
+    @precondition(lambda self: len(self.model.keys()) > 0)
+    @rule(data=st.data())
+    # @rule(key=keys_bundle)
```
dcherian, line 86: Curious why you got rid of these.

I realized yesterday that by using bundles you're effectively telling Hypothesis which rules are related closely to each other (e.g. you can only `get` or `delete` what you have `set`).

src/zarr/strategies.py
```diff
@@ -143,3 +145,19 @@ def basic_indices(draw: st.DrawFn, *, shape: tuple[int], **kwargs):  # type: ign
             )
         )
     )
+
+
+def key_ranges(keys: SearchStrategy | None = None) -> st.lists[:]:
```
dcherian, line 150:
```suggestion
def key_ranges(keys: SearchStrategy | None = None) -> SearchStrategy[list]:
```
https://hypothesis.readthedocs.io/en/latest/details.html#writing-downstream-type-hints

You can specialize the inner `list` type a lot more but this is the syntax.

src/zarr/strategies.py
```diff
@@ -1,9 +1,11 @@
+import string
```
jhamman, line 1: as part of this PR, can we move this module to `zarr.testing.strategies`?

tests/v3/test_store/test_stateful_store.py
e-marshall: okay, thanks, and then pass `SyncStoreWrapper()` to `ZarrStoreStateMachine` ? I think that means we'll definitely need to switch to using `run_machine_as_test()` (which we'd need to do anyway to handle different stores )

dcherian: For a start you can just use `SyncStoreWrapper(MemoryStore(mode="w"))` . We can generalize later.

tests/v3/test_store/test_stateful_store.py
```diff
@@ -0,0 +1,190 @@
...
+from zarr.buffer import Buffer, default_buffer_prototype
+from zarr.store import LocalStore, MemoryStore, RemoteStore
+
+# from strategies_store import StoreStatefulStrategies, key_ranges
+from zarr.testing.strategies import key_ranges, paths
+
+# zarr spec: https://zarr-specs.readthedocs.io/en/latest/v3/core/v3.0.html
+
+
+class SyncMemStoreWrapper(StoreABC):
```
dcherian, line 24:
```suggestion
class SyncMemStoreWrapper():
```

Sorry I led you down the wrong path with the ABC suggestion, it's an async ABC which doesn't help us. Just make sure the method names match for convenience.

tests/v3/test_store/test_stateful_store.py
```diff
@@ -0,0 +1,190 @@
...
+    def __init__(self, store_type):
+        """Store to hold async functions that map to StoreABC abstract methods"""
+        if store_type in ("Memory", "memory"):
+            self.store = MemoryStore(mode="w")  # store_container.store
+        elif store_type in ("Local", "local"):
+            self.store = LocalStore(mode="w")
+        elif store_type in ("Remote", "remote"):
+            self.store = RemoteStore(mode="w")
+        else:
+            raise ValueError(f"Invalid item_type: {store_type}")
```
dcherian, line 34:
```suggestion
    def __init__(self, store):
        """Store to hold async functions that map to StoreABC abstract methods"""
        self.store = store
```

This let's you do `SyncStoreWrapper(MemoryStore(mode="w"))` or any other store that satisfies `StoreABC`

tests/v3/test_store/test_stateful_store.py
```diff
@@ -0,0 +1,190 @@
...
+        # Unfortunately, hypothesis' stateful testing infra does not support asyncio
+
+    # So we redefine sync versions of the Store API.
+    # https://github.com/HypothesisWorks/hypothesis/issues/3712#issuecomment-1668999041
+    async def set(self, key, data_buffer):  # buffer is value
+        await self.store.set(key, data_buffer)
```
dcherian, line 40:
```suggestion
    def set(self, key, data_buffer):  # buffer is value
        return asyncio.run(self.store.set(key, data_buffer))
```

and so on...

tests/v3/test_store/test_stateful_store.py
```diff
@@ -0,0 +1,190 @@
...
+    async def supports_writes(self):
+        raise NotImplementedError
+
+
+class ZarrStoreStateMachine(RuleBasedStateMachine):
+    def __init__(self):  # look into using run_machine_as_test()
+        super().__init__()
+        self.model = {}
+        self.sync_wrapper = SyncMemStoreWrapper("memory")
+        self.store = self.sync_wrapper.store
```
dcherian, line 96:
```suggestion
        self.store = SyncStoreWrapper(MemoryStore(mode="w"))
```

tests/v3/test_store/test_stateful_store.py
```diff
@@ -0,0 +1,190 @@
...
+        # self.store = MemoryStore(mode='w')
+
+    @rule(key=paths, data=st.binary(min_size=0, max_size=100))
+    def set(self, key: str, data: bytes) -> None:
+        note(f"(set) Setting {key!r} with {data}")
+        assert not self.store.mode.readonly
+        data_buf = Buffer.from_bytes(data)
+        asyncio.run(self.sync_wrapper.set(key, data_buf))
+        self.model[key] = data_buf  # this was data
```
dcherian, line 105:
```suggestion
    def set(self, key: str, data: bytes) -> None:
        note(f"(set) Setting {key!r} with {data}")
        assert not self.store.mode.readonly
        data_buf = Buffer.from_bytes(data)
        self.store.set(key, data_buf) # this is what we want to enable with the wrapper.
        self.model[key] = data_buf
```

tests/v3/test_store/test_stateful_store.py
```diff
@@ -0,0 +1,190 @@
...
+        self.sync_wrapper = SyncMemStoreWrapper("memory")
+        self.store = self.sync_wrapper.store
+        # self.store = MemoryStore(mode='w')
+
+    @rule(key=paths, data=st.binary(min_size=0, max_size=100))
+    def set(self, key: str, data: bytes) -> None:
+        note(f"(set) Setting {key!r} with {data}")
+        assert not self.store.mode.readonly
+        data_buf = Buffer.from_bytes(data)
+        asyncio.run(self.sync_wrapper.set(key, data_buf))
```
dcherian, line 104:
```suggestion
        self.store.set(key, data_buf))
```

This is what we want to enable.

tests/v3/test_store/test_stateful_store.py
```diff
@@ -0,0 +1,190 @@
...
+        self.model[key] = data_buf  # this was data
+
+    @invariant()
+    def check_paths_equal(self) -> None:
+        note("Checking that paths are equal")
+        paths = asyncio.run(self.sync_wrapper.list())
```
dcherian, line 110:
```suggestion
        paths = self.store.list()
```

tests/v3/test_store/test_stateful_store.py
```diff
@@ -0,0 +1,190 @@
...
+        assert list(self.model.keys()) == paths
+        note("Checking values equal")
+        for key, _val in self.model.items():
+            store_item = asyncio.run(self.sync_wrapper.get(key)).to_bytes()
```
dcherian, line 114:
```suggestion
            store_item = self.store.get(key).to_bytes()
```

tests/v3/test_store/test_stateful_store.py
e-marshall: I get an error like this when i try these `E           ValueError: a coroutine was expected, got <async_generator object MemoryStore.list at 0x7f90d0a3ff40>` 
I'm still trying to figure out async things but from what I can tell I think it's asycing for an `async def... await` instead of `asyncio.run()` does that sound right to you ?

e-marshall: ah nevermind, I think i mixed that up

tests/v3/test_store/test_stateful_store.py
```diff
@@ -0,0 +1,182 @@
...
+    @rule(data=st.data())
+    def get_partial_values(self, data) -> None:
+        key_st = st.sampled_from(sorted(self.model.keys()))
+        key_range = data.draw(key_ranges(keys=key_st))
+
+        obs_maybe = asyncio.run(self.sync_wrapper.get_partial_values(key_range))
+        observed = []
+
+        for obs in obs_maybe:
+            assert obs is not None
```
dcherian, line 130: assert that None's are in the same place for the model and the store.

tests/v3/test_store/test_stateful_store.py
```diff
@@ -0,0 +1,182 @@
...
+        raise NotImplementedError
+
+    async def supports_writes(self):
+        raise NotImplementedError
+
+
+class ZarrStoreStateMachine(RuleBasedStateMachine):
+    def __init__(self):  # look into using run_machine_as_test()
+        super().__init__()
+        self.model = {}
```
dcherian, line 87:
```suggestion
        self.model: dict[str, bytes] = {}
```

tests/v3/test_store/test_stateful_store.py
e-marshall: didn't see this earlier but addressed [here](https://github.com/zarr-developers/zarr-python/pull/2070/commits/8e6f03c02c920e4979d70fb154dbb2a33f4ca1ee), I think i did what you meant here but not totally sure...

tests/v3/test_store/test_stateful_store.py
e-marshall: missed this earlier but addressed in https://github.com/zarr-developers/zarr-python/pull/2070/commits/8e6f03c02c920e4979d70fb154dbb2a33f4ca1ee

src/zarr/testing/strategies.py
```diff
@@ -143,3 +145,19 @@ def basic_indices(draw: st.DrawFn, *, shape: tuple[int], **kwargs):  # type: ign
...
+
+
+def key_ranges(keys: SearchStrategy | None = None) -> SearchStrategy[list]:
+    """fn to generate key_ranges strategy for get_partial_values()
+    returns list strategy w/ form: [(key, (range_start, range_step)),
+                                    (key, (range_start, range_step)),...]
+    """
+    if keys is None:
+        group_st = st.text(alphabet=string.ascii_letters + string.digits, min_size=1, max_size=10)
+        keys = st.lists(group_st, min_size=1, max_size=5).map("/".join)
```
dcherian, line 157:
```suggestion
def key_ranges(keys: SearchStrategy = node_names) -> SearchStrategy[list]:
    """fn to generate key_ranges strategy for get_partial_values()
    returns list strategy w/ form: [(key, (range_start, range_step)),
                                    (key, (range_start, range_step)),...]
    """
```

Let's use `numpydoc` conventions for docstrings: https://numpydoc.readthedocs.io/en/latest/

tests/v3/test_store/test_stateful_store.py
```diff
@@ -0,0 +1,223 @@
...
+
+    def set(self, key, data_buffer):  # buffer is value
+        return asyncio.run(self.store.set(key, data_buffer))
+
+    def list(self):
+        async def wrapper(gen):
+            return [i async for i in gen]
+
+        gen = self.store.list()  # async store list
+        return (i for i in asyncio.run(wrapper(gen)))
```
dcherian, line 45:
```suggestion
        yield from asyncio.run(wrapper(gen))
```
should work too

tests/v3/test_store/test_stateful_store.py
```diff
@@ -0,0 +1,223 @@
...
+
+    def get(self, key):
+        obs = asyncio.run(self.store.get(key, prototype=default_buffer_prototype()))
+        return obs
+
+    def get_partial_values(self, key_ranges):
```
dcherian, line 51: Let's preserve the original type signature: i.e. pass `prototype` in.

tests/v3/test_store/test_stateful_store.py
```diff
@@ -0,0 +1,223 @@
...
+
+    def __init__(self):
+        super().__init__()
+        self.model: dict[str, bytes] = {}
+        self.store = SyncStoreWrapper(MemoryStore(mode="w"))
+
+    # NOTE: set this and the next as initialize instead of rule to make sure they run
+    #      and create key, key_ranges before any rules that use key, key_ranges
+    #      not totally sure this is necessary
+    @initialize(key=paths, data=st.binary(min_size=0, max_size=100), target=keys_bundle)
```
dcherian, line 102:
```suggestion
    @rule(key=paths, data=st.binary(min_size=0, max_size=100), target=keys_bundle)
```

yes should not be necessary.

`initialize` only runs once so this is not good for exploring all possibilities.

tests/v3/test_store/test_stateful_store.py
```diff
@@ -0,0 +1,223 @@
...
+        self.model[key] = data_buf
+        note("init key")
+        return key
+
+    @initialize(key=keys_bundle, data=st.data(), target=key_ranges_bundle)
+    def init_key_ranges(self, data, key):
+        note("init key ranges")
+        key_st = st.just(key)
+        key_ranges_st = key_ranges(keys=key_st)
+        return data.draw(key_ranges_st)
```
dcherian, line 117:
```suggestion
```
Why do we need to have a `key_ranges` bundle?

tests/v3/test_store/test_stateful_store.py
```diff
@@ -0,0 +1,223 @@
...
+    def get(self, key) -> None:
+        note(f"(get) {key=}")
+        store_value = self.store.get(key)
+        # to bytes here necessary (on model and store) because data_buf set to model in set()
+        assert self.model[key].to_bytes() == (store_value.to_bytes())
+
+    @precondition(lambda self: len(self.model.keys()) > 0)
+    @rule(key=keys_bundle, key_range=key_ranges_bundle)
+    def get_partial_values(self, key, key_range) -> None:
+        note(f"(get partial) {key_range=}")
```
dcherian, line 130:
```suggestion
    @rule()
    def get_partial_values(self) -> None:
        key_range = data.draw(key_ranges(keys=st.sampled_from(sorted(self.model.keys())))
```

let's chat about this.

tests/v3/test_store/test_stateful_store.py
```diff
@@ -0,0 +1,223 @@
...
+        del self.model[path_st]
+
+    @rule()
+    def clear(self):
+        assert not self.store.mode.readonly
+        note("(clear)")
+        self.store.clear()
+        self.model.clear()
+        # check that model was cleared
+        assert len(self.model.keys()) == 0
```
dcherian, line 171: let's move this assertion to inside `model.clear()` if you need it.

tests/v3/test_store/test_stateful_store.py
```diff
@@ -0,0 +1,223 @@
...
+
+        # make sure they either both are or both aren't (same state)
+        assert self.store.empty() == (not self.model)
+
+    @rule(key=keys_bundle)
+    def exists(self, key) -> None:
+        #
+        note("(exists)")
+        # make sure same state
+        assert self.store.exists(key) == (key in self.model)
```
dcherian, line 185:
```suggestion
        assert self.store.exists(key)
        assert key in self.model
```
If you're drawing from `keys_bundle` it should always exist. If it doesn't, you've made an error in the test.

This is good IMO, you're double checking your test.

tests/v3/test_store/test_stateful_store.py
```diff
@@ -0,0 +1,223 @@
...
+        )
+
+    @precondition(lambda self: len(self.model.keys()) > 0)
+    @rule(data=st.data())
+    def delete(self, data) -> None:
+        path_st = data.draw(st.sampled_from(sorted(self.model.keys())))
+        note(f"(delete) Deleting {path_st=}")
+
+        self.store.delete(path_st)
+        del self.model[path_st]
```
dcherian, line 162:
```suggestion
    @rule(key=consumes(keys_bundle))
    def delete(self, data) -> None:
        note(f"(delete) Deleting {key=}")
        self.store.delete(key)
        del self.model[key]
```

https://hypothesis.readthedocs.io/en/latest/stateful.html#hypothesis.stateful.consumes

tests/v3/test_store/test_stateful_store.py
dcherian: now I'm confused. Does it only return `None` for non-existent keys? Does it return `None` for invalid key ranges too.

if the former, that path isn't tested here since you're drawing from the keys bundle.
If the latter, then we'll need to assert those `None`s are in the right place using the model.

tests/v3/test_store/test_stateful_store.py
e-marshall: okay, that would be great. I'm currently having trouble getting it to work in conjunction with `clear()` , having delete consume the keys_bundle works fine, but i haven't figured out how to delete the entire keys bundle to match the behavior of `.clear()`

tests/v3/test_store/test_stateful_store.py
e-marshall: I think this doesn't apply anymore now that we're not using bundles. going to go back to the old test where its asserting that if store.exists() is true, key in self.model also true and vice versa, rather than that both are true

tests/v3/test_store/test_stateful_store.py
e-marshall: I'm not sure how to handle this; if `prototype` is in the signature, then it would need to be passed to all of the rules that use `get()`, but one of the invariants also uses `get()` so this doesn't seem like a good option. I tried setting `prototype` as a default argument but this fails on precommit due to this error: 
image
The way i got it to work was by setting `default_buffer_prototype()` as an instance attr of the statemachine with `self.prototype = default_buffer_prototype()`, but this doesn't seem great, and it's still hardcoded since we can't pass a `prototype` arg to statemachine without switching to using `run_state_machine_as_test()` , any ideas on the best way to proceed?

dcherian: > but this doesn't seem great, and it's still hardcoded

This is OK. We can address later when someone needs this

tests/v3/test_store/test_stateful_store.py
e-marshall: I added a `get_invalid_keys()` rule to cover this

tests/v3/test_store/test_stateful_store.py
e-marshall: thanks for this link! really helpful

tests/v3/test_store/test_stateful_store.py
```diff
@@ -0,0 +1,219 @@
...
+        model_keys = list(self.model.keys())
+        # model_keys = ['/']
+        # key = '/'
+        note("(get_invalid)")
+
+        # note(f"(get invalid) key: {key}")
+        # note(f"(get invalid) val: {self.store.get(key, self.prototype)}")
+
+        assume(key not in model_keys)
+        assert self.store.get(key, self.prototype) is None
```
dcherian, line 123:
```suggestion
    @rule(key=paths)
    def get_invalid_keys(self, key) -> None:
        note("(get_invalid)")
        assume(key not in self.model.keys())
        assert self.store.get(key, self.prototype) is None
```

tests/v3/test_store/test_stateful_store.py
```diff
@@ -0,0 +1,219 @@
...
+            observed.append(obs.to_bytes())
+
+        model_vals_ls = []
+
+        for key, byte_range in key_range:
+            model_vals = self.model[key]
+            start = byte_range[0] or 0
+            step = byte_range[1]
+            stop = start + step if step is not None else None
+            model_vals_ls.append(model_vals[start:stop])
```
dcherian, line 144:
```suggestion
            start = byte_range[0] or 0
            step = byte_range[1]
            stop = start + step if step is not None else None
            model_vals_ls.append(self.model[key][start:stop])
```

dcherian (approved):
Very nice work, Emma!

👏🏾 👏🏾 👏🏾

tests/v3/test_store/test_stateful_store.py
e-marshall: thanks! adding this and the get invalid change locally

e-marshall:
thanks @dcherian ! thanks for all your help, fun to work with you and Seba on this!

tests/v3/test_store/test_stateful_store.py
```diff
@@ -0,0 +1,209 @@
...
+        then, methods in statemachine class are sync and call sync.
+        Unfortunately, hypothesis' stateful testing infra does not support asyncio
+        So we redefine sync versions of the Store API.
+        https://github.com/HypothesisWorks/hypothesis/issues/3712#issuecomment-1668999041
+        """
+        self.store = store
+        self.mode = store.mode
+
+    def set(self, key, data_buffer):
+        return asyncio.run(self.store.set(key, data_buffer))
```
jhamman, line 32: did we evaluate using [`zarr.core.sync.sync`](https://github.com/zarr-developers/zarr-python/blob/10ae5f3b1e2b67269c2181617302f305ef4b277c/src/zarr/core/sync.py#L55C1-L56C1) here?

tests/v3/test_store/test_stateful_store.py
```diff
@@ -0,0 +1,209 @@
...
+    rule,
+)
+
+from zarr.core.buffer import Buffer, BufferPrototype, default_buffer_prototype
+from zarr.store import MemoryStore
+from zarr.testing.strategies import key_ranges, paths
+
+
+class SyncStoreWrapper:
+    def __init__(self, store):
```
jhamman, line 20:
```suggestion
    def __init__(self, store: Store):
```

this assumes this will work for any store.

tests/v3/test_store/test_stateful_store.py
```diff
@@ -0,0 +1,209 @@
...
+        """Class to hold sync functions that map to async methods of MemoryStore
+        MemoryStore methods are async, this class' methods are sync, so just need to call asyncio.run() in them
+        then, methods in statemachine class are sync and call sync.
+        Unfortunately, hypothesis' stateful testing infra does not support asyncio
+        So we redefine sync versions of the Store API.
+        https://github.com/HypothesisWorks/hypothesis/issues/3712#issuecomment-1668999041
```
jhamman, line 26:
```suggestion
        """Synchronous Store wrapper
        
        This class holds synchronous methods that map to async methods of Store classes.
        The synchronous wrapper is needed because hypothesis' stateful testing infra does 
        not support asyncio so we redefine sync versions of the Store API.
        https://github.com/HypothesisWorks/hypothesis/issues/3712#issuecomment-1668999041
```

tests/v3/test_store/test_stateful_store.py
```diff
@@ -0,0 +1,209 @@
...
+    precondition,
+    rule,
+)
+
+from zarr.core.buffer import Buffer, BufferPrototype, default_buffer_prototype
+from zarr.store import MemoryStore
+from zarr.testing.strategies import key_ranges, paths
+
+
+class SyncStoreWrapper:
```
jhamman, line 19: You might consider changing this to: 
```suggestion
class SyncStoreWrapper(zarr.core.sync.SyncMixin):
```

Which will give you access to the `_sync` and `_sync_iter` methods out of the box.

tests/v3/test_store/test_stateful_store.py
```diff
@@ -0,0 +1,209 @@
...
+        raise NotImplementedError
+
+    def supports_listing(self):
+        raise NotImplementedError
+
+    def supports_partial_writes(self):
+        raise NotImplementedError
+
+    def supports_writes(self):
+        raise NotImplementedError
```
jhamman, line 79:
```suggestion
    @property
    def supports_listing(self):
        return self.store.supports_listing

    @property
    def supports_partial_writes(self):
        return self.supports_partial_writes

    @property
    def supports_writes(self):
        return self.store.supports_writes
```

No harm in implementing these.

tests/v3/test_store/test_stateful_store.py
```diff
@@ -0,0 +1,209 @@
...
+
+
+class ZarrStoreStateMachine(RuleBasedStateMachine):
```
jhamman, line 82: this concept will be new to many zarr contributors. Can we add a doc string that helps explain what the purpose is here?

tests/v3/test_store/test_stateful_store.py
```diff
@@ -0,0 +1,209 @@
...
+    def __init__(self, store):
+        """Class to hold sync functions that map to async methods of MemoryStore
+        MemoryStore methods are async, this class' methods are sync, so just need to call asyncio.run() in them
+        then, methods in statemachine class are sync and call sync.
+        Unfortunately, hypothesis' stateful testing infra does not support asyncio
+        So we redefine sync versions of the Store API.
+        https://github.com/HypothesisWorks/hypothesis/issues/3712#issuecomment-1668999041
+        """
+        self.store = store
+        self.mode = store.mode
```
jhamman, line 29: consider making this a `@property` of the store

```python
@property
def mode(self) -> AccessMode
    return self.store.mode
```

tests/v3/test_store/test_stateful_store.py
```diff
@@ -0,0 +1,209 @@
...
+
+    def set(self, key, data_buffer):
+        return asyncio.run(self.store.set(key, data_buffer))
+
+    def list(self):
```
jhamman, line 34: If possible, it would be nice to add return types to the wrapper methods.

jhamman (commented):
Super work @e-marshall! I have a bunch of comments on the Store wrapper -- none of them major.

tests/v3/test_store/test_stateful_store.py
dcherian: no didn't know that it existed.

tests/v3/test_store/test_stateful_store.py
dcherian: Should we use `_sync` or `zarr.core.sync.sync`

tests/v3/test_store/test_stateful_store.py
e-marshall: ok, thanks ! i'll check this out

e-marshall:
thanks @jhamman ! I made these changes on https://github.com/zarr-developers/zarr-python/pull/2070/commits/d63df09a04503b0f36323b27933597ba60dd5652. happy to tweak/shorten the statemachine docstring if it seems too long

tests/v3/test_store/test_stateful_store.py
e-marshall: added in https://github.com/zarr-developers/zarr-python/pull/2070/commits/d63df09a04503b0f36323b27933597ba60dd5652

tests/v3/test_store/test_stateful_store.py
e-marshall: added in https://github.com/zarr-developers/zarr-python/pull/2070/commits/d63df09a04503b0f36323b27933597ba60dd5652

tests/v3/test_store/test_stateful_store.py
e-marshall: sounds good! I added a docstring but open to suggestions/improvements

tests/v3/test_store/test_stateful_store.py
```diff
@@ -0,0 +1,232 @@
...
+        not support asyncio so we redefine sync versions of the Store API.
+        https://github.com/HypothesisWorks/hypothesis/issues/3712#issuecomment-1668999041
+        """
+        self.store = store
+
+    @property
+    def mode(self) -> AccessMode:
+        return self.store.mode
+
+    def set(self, key, data_buffer) -> None:
```
jhamman, line 36: The return types look great here. Let's also type the input parameters (key, data_buffer). This comment applies to the rest of this class too.

tests/v3/test_store/test_stateful_store.py
```diff
@@ -0,0 +1,232 @@
...
+        return sync(self.store.set(key, data_buffer))
+
+    def list(self) -> list:
+        async def wrapper(gen):
+            return [i async for i in gen]
+
+        gen = self.store.list()
+        yield from sync(wrapper(gen))
```
jhamman, line 44: I think this will work :)

```suggestion
    return self._sync_iter(self.store.list())
```

tests/v3/test_store/test_stateful_store.py
```diff
@@ -0,0 +1,232 @@
...
+
+    def get(self, key, prototype: BufferPrototype) -> zarr.core.buffer.Buffer:
+        obs = sync(self.store.get(key, prototype=prototype))
```
jhamman, line 47:
```suggestion
        obs = self._sync(self.store.get(key, prototype=prototype))
```

tests/v3/test_store/test_stateful_store.py
```diff
@@ -0,0 +1,232 @@
...
+        return obs
+
+    def get_partial_values(self, key_ranges, prototype: BufferPrototype) -> zarr.core.buffer.Buffer:
+        obs_partial = sync(
```
jhamman, line 51:
```suggestion
        obs_partial = self._sync(
```

tests/v3/test_store/test_stateful_store.py
```diff
@@ -0,0 +1,232 @@
...
+        return sync(self.store.delete(path))
+
+    def empty(self) -> bool:
+        return sync(self.store.empty())
+
+    def clear(self) -> None:
+        return sync(self.store.clear())
+
+    def exists(self, key) -> bool:
+        return sync(self.store.exists(key))
```
jhamman, line 66:
```suggestion
        return self._sync(self.store.delete(path))

    def empty(self) -> bool:
        return self._sync(self.store.empty())

    def clear(self) -> None:
        return self._sync(self.store.clear())

    def exists(self, key) -> bool:
        return self._sync(self.store.exists(key))
```

tests/v3/test_store/test_stateful_store.py
```diff
@@ -0,0 +1,232 @@
...
+    RuleBasedStateMachine,
+    invariant,
+    precondition,
+    rule,
+)
+
+import zarr
+from zarr.abc.store import AccessMode, Store
+from zarr.core.buffer import Buffer, BufferPrototype, default_buffer_prototype
+from zarr.core.sync import sync
```
jhamman, line 16:
```suggestion
```

I think with my comments below, this is no longer needed.

jhamman (approved):
One last set of small comments but I'm approving this as is. 🙌

tests/v3/test_store/test_stateful_store.py
e-marshall: sounds good! still getting used to type hinting , should i add them to the input parameters of the state machine methods as well?

dcherian: Yes, it'll be good practice

e-marshall: ok perfect

e-marshall:
thanks @jhamman !

dcherian:
Can we merge so that one might add `list_prefix`? #2064

d-v-b:
> Can we merge so that one might add `list_prefix`? #2064

I am working on getting that PR updated and ready

dcherian:
Right, I meant that we could merge *this* PR so that we can add `list_prefix` in that PR

PR #2073, opened 2024-08-10
[v2] Fix doctests with numpy 2.0

dstansby:
This fixes doctests to run with numpy 2.0, and skips them if numpy is < 2. This isn't a very elegant solution to these values changing with a new version of numpy, but I think it's a pragmatic solution to getting CI working again for v2.


jhamman:
I support the concept here but I do wonder if there is a way to achieve this behavior without injecting `__doctest_skip__` into the namespace of these modules. Can we skip at the CI level or something like that?

dstansby:
Turns out there is a way to specify which files should be skipped with specific versions of dependendcies in the pytest config 👍

jhamman (approved):
Much better! Thanks @dstansby.

PR #2074, opened 2024-08-10
[v2] Drop support for Python 3.9

dstansby:
Python 3.9 has been out of support for a while in [SPEC 0](https://scientific-python.org/specs/spec-0000/#support-window), so drop support in zarr v2.

TODO:


d-v-b (approved):
this looks good to me, thanks @dstansby 

edit: i didn't see that this was a draft. so i should say, what I read so far looks good to me :)

d-v-b:
test failures look like numpy 2 issues

dstansby:
https://github.com/zarr-developers/zarr-python/pull/2073 should fix these tests

zarr/_storage/v3_storage_transformers.py
```diff
@@ -32,7 +32,8 @@ class _ShardIndex(NamedTuple):
 
     def __localize_chunk__(self, chunk: Tuple[int, ...]) -> Tuple[int, ...]:
         return tuple(
-            chunk_i % shard_i for chunk_i, shard_i in zip(chunk, self.store.chunks_per_shard)
+            chunk_i % shard_i
+            for chunk_i, shard_i in zip(chunk, self.store.chunks_per_shard, strict=False)
```
d-v-b, line 36: why is `strict=False`? Are the lengths of the two iterables fixed somewhere else?

d-v-b: (ditto for all the other `zip` invocations)

dstansby: Admittedly I just blindly applied this change that `ruff` was asking for 🙈 . For simplicitly I've reverted the `strict=False` changes, and ignored the rule for now.

d-v-b: is it this rule? https://docs.astral.sh/ruff/rules/zip-without-explicit-strict/ I think we can just comply by using `strict=True` in that case

d-v-b: to clarify, I don't think there's any usage of zip in the codebase where we want the `strict=False` behavior, so `strict=True` should be fine.

dstansby: I thought for simpliclty in this PR it would be better to ignore the rule - happy to add `strict=True` everywhere instead if you think that's a better option?

d-v-b: i think we should follow the rule and, for each invocation of `zip`, we decide if the usage is consistent with `strict=True` or `strict=False`. I'm pretty sure that we always use zip with the assumption that the two iterables have the same number of elements, so any test failures after `strict=False` would probably be real bugs.

dstansby: Do you think this PR should be blocked on that - I'm not sure I have the expertise to go through and decide each case, but I'm happy for someone else to push to this branch.

d-v-b: no need to block, let's do it in a follow-up PR

dstansby: Fab - I can open an issue in a bit so this doesn't get lost

PR #2075, opened 2024-08-10
Enforce Repo-Review rules

DimitriPapadopoulos:
https://learn.scientific-python.org/development/guides/repo-review/?repo=zarr-developers%2Fzarr-python&branch=v3

TODO:

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -39,3 +39,12 @@ repos:
           - pytest
           # Zarr v2
           - types-redis
+  - repo: https://github.com/scientific-python/cookie
+    rev: 2024.04.23
+    hooks:
+      - id: sp-repo-review
+  - repo: https://github.com/pre-commit/pygrep-hooks
+    rev: v1.10.0
+    hooks:
+      - id: rst-directive-colons
+      - id: rst-inline-touching-normal
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -257,3 +257,9 @@ filterwarnings = [
 markers = [
     "gpu: mark a test as requiring CuPy and GPU"
 ]
+
+[tool.repo-review]
+ignore = [
+	"PC111",  # fix Python code in documentation - enable later 
+	"PC180",  # for JavaScript - not interested
+]


.pre-commit-config.yaml
```diff
@@ -40,3 +40,12 @@ repos:
           - pytest
           # Zarr v2
           - types-redis
+  - repo: https://github.com/scientific-python/cookie
+    rev: 2024.04.23
+    hooks:
+      - id: sp-repo-review
```
dstansby, line 46:
```suggestion
  - repo: https://github.com/scientific-python/repo-review
    rev: 0.11
    hooks:
      - id: repo-review
```
I think better to use the repo-review repository itself?

pyproject.toml
```diff
@@ -223,3 +223,6 @@ filterwarnings = [
     "ignore:PY_SSIZE_T_CLEAN will be required.*:DeprecationWarning",
     "ignore:The loop argument is deprecated since Python 3.8.*:DeprecationWarning",
 ]
+
+[tool.repo-review]
+ignore = ["PC111", "PC180"]
```
dstansby, line 228: It would be nice to add some comments explaining what the rules being ignored are (and perhaps why they're being ignored?)

.pre-commit-config.yaml
DimitriPapadopoulos: I understand sp-repo-review/cookie is the tool evaluate repos against the guidelines in the Scientific Python Library Development Guide. The repo-review framework has sp-repo-review and validate-pyproject plyugins.

The [repo-review documentation](https://repo-review.readthedocs.io/en/latest/intro.html#pre-commit) suggests we should directly use the plugin we're interested in:
> ### Pre-commit
> You can also use this from pre-commit:
> ```yaml
> - repo: https://github.com/scientific-python/repo-review
>   rev: <version>
>   hooks:
>     - id: repo-review
>       additional_dependencies: ["repo-review[cli]", "sp-repo-review==<version>"]
> ```
> (Insert the current version above, and ideally pin the plugin version, as long as you have a way to auto-update it.)
> 
> Though check your favorite plugin, which might directly support running from pre-commit, and then pre-commit’s pinning system will pin on your plugin, rather than the framework (repo-review).

pyproject.toml
DimitriPapadopoulos: These are rules for Python code in documentation and for JavaScript.
- I thought I'd leave the documentation as is to start with.
- There only one JavaScript file, in the documentation: `docs/_static/custom.js`.

.pre-commit-config.yaml
dstansby: In that example they use https://github.com/scientific-python/repo-review instead of https://github.com/scientific-python/cookie, so I think it's best to change it from `cookie` to `repo-review`?

dstansby (approved):
Looks 👍  - just another question about which repository we should be running the pre-commit hook from

.pre-commit-config.yaml
DimitriPapadopoulos: It depends on who applied the relevant commit. Let me investigate.

DimitriPapadopoulos: It goes on with:
> Though check your favorite plugin, which might directly support running from pre-commit, and then pre-commit’s pinning system will pin on your plugin, rather than the framework (repo-review).

PR #2078, opened 2024-08-10
feature(store): V3 ZipStore

### LINKED ISSUE #2010: [v3] Develop ZipStore
In Zarr-Python 2, we had the ZipStore. This hasn't been implemented yet in v3 but will need to be before the 3.0 release. 

Fsspec has a ZipStore but I don't think it has an async interface so I think we should implement our own. Some design thought will be needed here to get locking/etc. right.

The intended behavior is something like this:

```python
import zarr

store = zarr.ZipStore('data/example.zip', mode='w')
root = zarr.group(store=store)
z = root.zeros('foo/bar', shape=(1000, 1000), chunks=(100, 100), dtype='i4')
z[:] = 42
store.close()
```

jhamman:
I threw together a super basic implementation of a v3 zip store. I am still testing this but it seems to be working for basic things.

closes #2010 

try it!

```python
In [1]: import zarr
   ...: 
   ...: store = zarr.store.ZipStore('data/example.zip', mode='w')
   ...: root = zarr.group(store=store)
   ...: z = root.create_array(shape=(100, 100), chunks=(10, 10), name="foo")

In [2]: z[:] = 42

In [3]: store.close()

# unzip -l data/example.zip
```

TODO:


src/zarr/store/zip.py
```diff
@@ -0,0 +1,221 @@
...
+        if not isinstance(value, Buffer):
+            raise TypeError("ZipStore.set(): `value` must a Buffer instance")
+        with self._lock:
+            self._set(key, value)
+
+    async def set_partial_values(self, key_start_values: list[tuple[str, int, bytes]]) -> None:
+        raise NotImplementedError
+
+    async def delete(self, key: str) -> None:
+        raise NotImplementedError
```
jhamman, line 152: Today I was reminded that you can't delete anything from inside a ZipFile 😢. This behavior also existed in 2.18.

src/zarr/store/zip.py
```diff
@@ -0,0 +1,221 @@
...
+from zarr.abc.store import Store
+from zarr.buffer import Buffer, BufferPrototype
+
+ZipStoreAccessModeLiteral = Literal["r", "w", "a"]
+
+
+class ZipStore(Store):
+    supports_writes: bool = True
+    supports_partial_writes: bool = False
+    supports_listing: bool = True
```
jhamman, line 20: what do we think about adding `supports_deletes: bool = False` as a class attribute?

d-v-b: that would be a solid extrapolation of the design we are currently using. one downside of that design is that, even if a class has `supports_x` set to `False`, the class will still need an implementation of `x`. Another solution would be to express `supports_x` by having the class inherit from a `DoesX` mixin. But that's maybe out of scope for this PR.

jhamman: I've got the `supports_deletes` route for now.

d-v-b:
I played around with this a bit locally. 
- the zip archive gets completely broken if you forget to call `store.close` after writing (or, presumably, if the program exits before reaching the line where `store.close()` is invoked). I think this is severe enough that this PR should have the ZipStore implement `ContextManager` to defend against data corruption issues.
- attempting to write twice to the array fails, because the zip store doesn't allow deletion. In such a case, the user will see a `NotImplementedError`. We should probably make something more informative for this case. But I don't think this is a blocker.
- We will at some point need to address the possibility of zip file stored on s3. Our current store hierarchy can't express this easily I think. but that's a future problem.

jhamman:
> I played around with this a bit locally.
> 
> * the zip archive gets completely broken if you forget to call `store.close` after writing (or, presumably, if the program exits before reaching the line where `store.close()` is invoked). I think this is severe enough that this PR should have the ZipStore implement `ContextManager` to defend against data corruption issues.

There is now a context manager interface here. I'll note that I'm mildly uncomfortable with this because its not clear if this should be a async context manager or not. While I think this is what you were asking for, its not really the way classes that do async things should work. I've got some ideas for how we can improve this but I think we should hold them for another PR.

Two asides: 
1. not closing the v2 `ZipStore` also led to bad things. So this is "expected behavior" I guess.
2. we could consider an `atexit` cleanup of the file handle but I think we should first decide on the context api.

> * attempting to write twice to the array fails, because the zip store doesn't allow deletion. In such a case, the user will see a `NotImplementedError`. We should probably make something more informative for this case. But I don't think this is a blocker.

Noting that this behavior is also present in v2: https://github.com/zarr-developers/zarr-python/blob/11fd8db8f06372ffc1f6f7e22315d43e4702ec60/zarr/storage.py#L1884-L1885

> * We will at some point need to address the possibility of zip file stored on s3. Our current store hierarchy can't express this easily I think. but that's a future problem.

I think fsspec can handle most of this for us. Some work is likely needed there but it should be possible.

src/zarr/store/zip.py
zoj613: Wouldn't it be nice to have an In-memory version of a Zip store where all the zip data is read in memory. This way it can support deleting and updating entries. Thereafter, a user can persist the data using a method like `write_to_file`. I think it would be very efficient for data sets whose compressed size is small enough to fit entirely in memory.

jhamman: Indeed that would be nice @zoj613. I would like to save that until after the 3.0 release though as the minimal zip store is a release blocker at this point.

joshmoore:
A sidenote that similar to the situation with consolidated metadata, the zip store will bring v3 to parity with v2 (good thing :tm: 👍🏽) but will leave other implementations in the same situation of not having a spec. If there's a chance that the discussion around that spec will lead to changes in the on-disk format, it would be in our best interest to make sure we have a specification to go along with these change so that there's less likelihood of multiple versions we need to support.

d-v-b:
> A sidenote that similar to the situation with consolidated metadata, the zip store will bring v3 to parity with v2 (good thing ™️ 👍🏽) but will leave other implementations in the same situation of not having a spec. If there's a chance that the discussion around that spec will lead to changes in the on-disk format, it would be in our best interest to make sure we have a specification to go along with these change so that there's less likelihood of multiple versions we need to support.

+1, it would be great to define a spec for this so other implementations can safely support it

jhamman:
Draft spec to go along with this feature is now available for review: https://github.com/zarr-developers/zarr-specs/pull/311

src/zarr/abc/store.py
```diff
@@ -11,6 +11,7 @@
 
 
 class AccessMode(NamedTuple):
+    str: AccessModeLiteral
```
d-v-b, line 14: Do we want to shadow the builtin `str`?

tests/v3/test_store/test_zip.py
```diff
@@ -0,0 +1,68 @@
...
+
+import pytest
+
+from zarr.abc.store import AccessMode
+from zarr.core.buffer import Buffer, default_buffer_prototype
+from zarr.store.zip import ZipStore
+from zarr.testing.store import StoreTests
+
+
+class TestZipStore(StoreTests[ZipStore]):
```
d-v-b, line 16: Just a note that we should have nice docstrings for the store classes, and their methods. This is not a blocker, and can be handled by a later PR.

tests/v3/test_store/test_zip.py
```diff
@@ -0,0 +1,68 @@
...
+        kwargs = {**store_kwargs, "mode": "r"}
+        store = await self.store_cls.open(**kwargs)
+        assert store.mode == AccessMode.from_literal("r")
+        assert store.mode.readonly
+
+        # set
+        with pytest.raises(ValueError):
+            await store.set("foo", Buffer.from_bytes(b"bar"))
+
+        # # delete
```
d-v-b, line 50: is this still needed?

src/zarr/store/zip.py
```diff
@@ -0,0 +1,244 @@
...
+from zarr.abc.store import Store
+from zarr.core.buffer import Buffer, BufferPrototype
+
+if TYPE_CHECKING:
+    from typing_extensions import Self
+
+ZipStoreAccessModeLiteral = Literal["r", "w", "a"]
+
+
+class ZipStore(Store):
```
d-v-b, line 20: Just a note that we should have nice docstrings for the store classes, and their methods. This is not a blocker, and can be handled by a later PR.

src/zarr/abc/store.py
jhamman: I don't see an issue here. This simply allows us to extract the mode as a string from the named tuple like this: `mode.str`. I think this is a fairly common pattern.

d-v-b: I didn't see that this was a member of a named tuple. seems legit

d-v-b (approved):
Looks good! Let's get it in

(@jhamman do you think the issue @LDeakin raised is a blocker, or can we sort that out in a future pr?)

jhamman:
Let me see about addressing @LDeakin's finding here before merging.

jhamman:
@LDeakin - your example exposed something beyond the reach of this PR. Setting a chunk equal to the fill value triggers a delete - even if the chunk does not exist. This is handled nicely for all other stores but is hard to manage here. Something we could do in the future is allow `delete` to `pass` if the key does not exist.

PR #2086, opened 2024-08-14
Move BatchedCodecPipeline to zarr.core

normanrz:
Moves the BatchedCodecPipeline to zarr.core to make it easier to evolve it post 3.0.0 release.


jhamman (approved):
Thanks @normanrz!

Issue #1773, opened 2024-04-05
[v3] Typing: use TypedDict for all dictionaries


jhamman:
Small typing improvement here. @d-v-b - can you expand on what needs to be done (feel free to edit this description)?

ahmadjiha:
Hello @jhamman @d-v-b! I am excited for zarr v3 and would be interested in working on this issue.

PR #2099, opened 2024-08-19
Return TypedDict instances from Metadata.to_dict()

ahmadjiha:
Define TypedDict classes for metadata models that have a well-typed dict representation and set the return value of the the model's to_dict() method to the TypedDict

[Description of PR]

Closes zarr-developers/zarr-python#1773

TODO:

### DIFF
--- a/src/zarr/abc/codec.py
+++ b/src/zarr/abc/codec.py
@@ -2,9 +2,10 @@ from __future__ import annotations
 
 from abc import abstractmethod
 from collections.abc import Awaitable, Callable, Iterable
-from typing import TYPE_CHECKING, Any, Generic, TypeVar
+from typing import TYPE_CHECKING, Any, Generic, TypedDict, TypeVar
 
 import numpy as np
+from typing_extensions import NotRequired
 
 from zarr.abc.metadata import Metadata
 from zarr.abc.store import ByteGetter, ByteSetter
@@ -168,6 +169,22 @@ class BytesBytesCodec(_Codec[Buffer, Buffer]):
 
 
 Codec = ArrayArrayCodec | ArrayBytesCodec | BytesBytesCodec
+
+
+class CodecConfigDict(TypedDict):
+    """A dictionary representing a codec configuration."""
+
+    ...
+
+
+T = TypeVar("T", bound=CodecConfigDict)
+
+
+class CodecDict(TypedDict, Generic[T]):
+    """A generic dictionary representing a codec."""
+
+    name: str
+    configuration: NotRequired[T]
 
 
 class ArrayBytesCodecPartialDecodeMixin:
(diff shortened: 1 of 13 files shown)


ahmadjiha:
I experimented with making the `Metadata` base class generic and ran into issues with bounding the generic type with `dict[str, JSON]`. 

Making the `Metadata` class generic was easy enough:
```python3
# src/zarr/abc/metadata.py

...

T = TypeVar("T", bound=dict[str, JSON])

@dataclass(frozen=True)
class Metadata(Generic[T]):
    def to_dict(self) -> T:
        ...
```

However, I ran into issues with the `TypedDict` classes being outside of the bounds of the generic type. For example:
```python3
# src/zarr/core/chunk_key_encodings.py

...

class ChunkKeyEncodingDict(TypedDict):
    """A dictionary representing a chunk key encoding configuration."""

    name: str
    configuration: dict[Literal["separator"], SeparatorLiteral]


@dataclass(frozen=True)
class ChunkKeyEncoding(Metadata[ChunkKeyEncodingDict]):  # Type argument "ChunkKeyEncodingDict" of "Metadata" must be a subtype of "dict[str, JSON]"
    name: str
    separator: SeparatorLiteral = "."
```

It seems like we need a way to bound the `TypedDict` value types at the dictionary level. I've tried quite a few things -- I have yet to come across a solution. 

Is it possible that this is not supported? Would appreciate any guidance @jhamman @d-v-b

d-v-b:
> `T = TypeVar("T", bound=dict[str, JSON])`

I think the problem is specifying the type bound to be `dict[str, JSON]`. First, typeddicts are supposed to be immutable, so they are not a subclass of `dict`. `Mapping` should be used instead.

Second, for some reason `Mapping[str, JSON]` won't work, but `Mapping[str, object]` does work as a type bound. I don't really know the answer to this one, hopefully someone can explain it.

Here's an example that I got working:

```python
from abc import abstractmethod
from dataclasses import dataclass
from typing import Any, Generic, TypeVar, Mapping, TypedDict

T = TypeVar('T', bound=Mapping[str, object])

class Meta(Generic[T]):

    @abstractmethod
    def to_dict(self) -> T:
        ...

class ExampleSpec(TypedDict):
    a: str
    b: str

@dataclass
class Example(Meta[ExampleSpec]):
    a: str
    b: str

    def to_dict(self) -> ExampleSpec:
        return {'a': self.a, 'b': self.b}
    
x = Example(a='10', b='10')
y = x.to_dict()
reveal_type(y)
"""
Revealed type is "TypedDict('tessel.ExampleSpec', {'a': builtins.str, 'b': builtins.str})"
"""
```

```

ahmadjiha:
Hello @d-v-b! 

I rebased and pushed commits to make the `Metadata` abstract base class generic + required changes to the downstream inheriting classes. I updated the PR descriptions w/ implementation notes + additional questions. 

When you have a moment, a review would be super helpful!

Issue #1849, opened 2024-05-07
[v3] remove old v2 code entirely


d-v-b:
This is a more broad alternative to #1791 -- basically we remove all the old v2 code in favor of the v3 efforts. Happy to hear thoughts on this.

jeromekelleher:
What are the implications for backward compatibility?

d-v-b:
@jeromekelleher the proposed change would remove API backwards compatibility, but I don't think we plan to keep the v2 API when we release v3 anyway -- v2 has a very large API surface area, it would be a burden to maintain purely for legacy reasons.

The v3 codebase does support reading zarr v2, so removing the v2 code does not change _functional_ compatibility with zarr v2.

@jni suggested that we publish the final release of `zarr-python` 2.X on pypi as `zarr2` or something similar, to ensure that zarr 3.x and zarr 2.x can co-exist in the same python environment. I think this is a good idea, but there might be alternatives to consider.

jeromekelleher:
I see - if there's good compatibility on v2 stores with the new v3 code, and reasonable backward compatibility with old code bases, then I agree cutting out the old v2 code entirely is the right approach.

jeromekelleher:
Well, discussions about making a new  v2 package and/or maintaining the old v2 code within the Zarr-Python repo raise questions about compatibility. If good compatibility with existing v2 stores and code was a first-order concern, then these wouldn't even be on the table, I would have thought.

> but I don't think we plan to keep the v2 API when we release v3 anyway -- v2 has a very large API surface area, it would be a burden to maintain purely for legacy reasons.

That's quite a big thing, though. I agree there's a lot of sprawl around the v2 API that could be removed without breaking much, but surely there's an idiomatic core of functionality around creating/opening/reading/writing arrays that will be the same? Are there deprecations on the bits of v2 that are going to break when v3 drops?

d-v-b:
@jeromekelleher 
> I think you guys are doing a great job by the way, I just want to get my head around what's being planned here. I would rather not have my code broken unless it's necessary, but I understand you've got to break some eggs sometimes.

[...]

PR #2102, opened 2024-08-19
Doc/v3 migration guide

jhamman:
Over the past few weeks, we've had a number of conversations/questions about the policy for backward compatibility, deprecations, and breaking changes with the upcoming 3.0 release. This doc is meant to help us iterate toward common language. In its initial form, it is not complete.

Goals for the text here:

1. Developers of 2.18 and 3.0 should be able to decide if backward compatibility is a required attribute of a contribution
2. Users of Zarr should be able to understand if their application will be impacted by the upcoming 3.0 release
3. Users of Zarr should be able to make a plan for how they will adapt their usage of Zarr after the release
1. [non-goal] This is not meant to provide a comprehensive listing of the changes to the zarr API 

cc @zarr-developers/python-core-devs

### DIFF
--- a/docs/migration.rst
+++ b/docs/migration.rst
@@ -1,0 +1,86 @@
+Zarr-Python 3 Migration Guide
+=============================
+
+Zarr-Python 3 introduces a number of changes to the API, including a number
+of significant breaking changes and pending deprecations.
+
+This page provides a guide highlighting the most notable changes to help you
+migrate your code from Zarr-Python 2.x to Zarr-Python 3.x.
+
+Compatibility target
+--------------------
+
+Zarr-Python 3 represents a major refactor of the Zarr-Python codebase. Some of the goals motivating this refactor included:
+
+  - adding support for the V3 specification (alongside the V3 specification)
+  - cleaning up internal and user facing APIs
+  - improving performance (particularly in high latency storage environments like cloud object store)
+
+Though these goals necessitated some breaking changes to the API (hence the major version update), we have tried to maintain
+backwards compatibility in the most widely used parts of the API including the `Array` and `Group` classes and the top-level
+API (e.g. `zarr.open_array` and `zarr.open_group`).
+
+Getting ready for 3.0
+---------------------
+
+Ahead of the 3.0 release, we suggest projects that depend on Zarr-Python take the following actions:
+
+1. Pin the supported Zarr-Python version to ``zarr>=2,<3``. This is a best practice and will protect your users from any incompatibilities that may arise during the release of Zarr-Python 3.0.
+2. Limit your imports from the Zarr-Python package. Most of the primary API ``zarr.*`` will be compatible in 3.0. However, the following breaking API changes are planned:
+   
+   - ``numcodecs.*`` will no longer be available in ``zarr.*``. (Suggested action: transition to importing codecs from ``numcodecs`` directly.)
+   - The ``zarr.v3_api_available`` feature flag is being removed. (Suggested action: this experimental feature was deprecated in v2.18.)
+   - The following internal modules are being removed or significant changed:
+   
+    - ``zarr.attrs``
+    - ``zarr.codecs``
+    - ``zarr.context``
+    - ``zarr.core``
+    - ``zarr.hierarchy``
+    - ``zarr.indexing``
+    - ``zarr.meta``
+    - ``zarr.meta_v1``
+    - ``zarr.storage``
+    - ``zarr.sync``
+    - ``zarr.types``
+    - ``zarr.util``
+
+Continue using Zarr-Python 2
+----------------------------
+
+Zarr-Python 2.x is still available, though we recommend migrating to Zarr-Python 3 for its improvements and new features.
+
+If you need to use the latest Zarr-Python 2 release, you can install it with:
+
+    $ pip install "zarr==2.*"
+
+
+Migration Guide
+---------------
+
+The following sections provide details on the most important changes in Zarr-Python 3.
+
+Changes to the Array class
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+TODO
+
+Changes to the Group class
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+TODO
+
+Changes to the Store class
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Some of the biggest changes in Zarr-Python 3 are found in the `Store` class. The most notable changes to the Store API are:
+
+1. Dropped the ``MutableMapping`` base class in favor of a custom abstract base class (``zarr.abc.store.Store``).
+2. Switched to a primarily Async interface.
+
+TODO
+
+Miscellaneous changes
+~~~~~~~~~~~~~~~~~~~~~
+
+TODO
--- a/docs/migration.rst
+++ b/docs/migration.rst
@@ -18,7 +18,8 @@ Zarr-Python 3 represents a major refactor of the Zarr-Python codebase. Some of t
 
 Though these goals necessitated some breaking changes to the API (hence the major version update), we have tried to maintain
 backwards compatibility in the most widely used parts of the API including the `Array` and `Group` classes and the top-level
-API (e.g. `zarr.open_array` and `zarr.open_group`).
+API (e.g. `zarr.open_array` and `zarr.open_group`). It is worth noting that we significantly evolved the internal data model,
+moving away from a model that was tightly coupled to the v2 specification, and to a more generic representation of Zarr objects.
 
 Getting ready for 3.0
 ---------------------
@@ -44,11 +45,13 @@ Ahead of the 3.0 release, we suggest projects that depend on Zarr-Python take th
     - ``zarr.sync``
     - ``zarr.types``
     - ``zarr.util``
+    - ``zarr.n5``
 
 Continue using Zarr-Python 2
 ----------------------------
 
 Zarr-Python 2.x is still available, though we recommend migrating to Zarr-Python 3 for its improvements and new features.
+Security and bug fixes will be made to the 2.x series for at least 6 months following the first Zarr-Python 3 release.
 
 If you need to use the latest Zarr-Python 2 release, you can install it with:
 
@@ -63,12 +66,12 @@ The following sections provide details on the most important changes in Zarr-Pyt
 Changes to the Array class
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-TODO
+1. Disallow direct construction
 
 Changes to the Group class
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-TODO
+1. Disallow direct construction
 
 Changes to the Store class
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/docs/migration.rst
+++ b/docs/migration.rst
@@ -48,6 +48,7 @@ Ahead of the 3.0 release, we suggest projects that depend on Zarr-Python take th
 
 3. Test that your package works with v3. You can start testing against version 3 now (pre-releases are being published to PyPI weekly).
 4. Update the pin to zarr >=3
+
 Continue using Zarr-Python 2
 ----------------------------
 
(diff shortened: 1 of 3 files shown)


TomAugspurger:
One thing that might be helpful: what's the group's tolerance for either compatibility code or deprecations as a way to ease the transition? It sounds like strict backwards compatibility (perhaps with warnings) isn't a goal. Is there tolerance for things like https://github.com/zarr-developers/zarr-python/pull/2098 (e.g. restore some properties to the `Group` object, loosen the keyword-only requirement for some functions). Likewise for things like "cleaning up internal and user facing APIs", which could be done with a  deprecation warning. Even if there isn't tolerance for backwards compatibility shims that really clash with the V3 spec or the current v3 implementation?

docs/migration.rst
```diff
@@ -0,0 +1,89 @@
+Zarr-Python 3 Migration Guide
+=============================
+
+Zarr-Python 3 introduces a number of changes to the API, including a number
+of significant breaking changes and pending deprecations.
+
+This page provides a guide highlighting the most notable changes to help you
+migrate your code from Zarr-Python 2.x to Zarr-Python 3.x.
```
dstansby, line 8:
```suggestion
migrate your code from version 2 to version 3.
```

docs/migration.rst
```diff
@@ -0,0 +1,89 @@
...
+
+Compatibility target
+--------------------
+
+Zarr-Python 3 represents a major refactor of the Zarr-Python codebase. Some of the goals motivating this refactor included:
+
+  - adding support for the V3 specification (alongside the V3 specification)
```
dstansby, line 15:
```suggestion
  - adding support for the Zarr V3 specification (alongside the Zarr V2 specification)
```

docs/migration.rst
```diff
@@ -0,0 +1,89 @@
...
+  - cleaning up internal and user facing APIs
+  - improving performance (particularly in high latency storage environments like cloud object store)
+
+Though these goals necessitated some breaking changes to the API (hence the major version update), we have tried to maintain
```
dstansby, line 19:
```suggestion
These goals necessitated some breaking changes to the API (hence the major version update), but we have tried to maintain
```

docs/migration.rst
```diff
@@ -0,0 +1,89 @@
...
+backwards compatibility in the most widely used parts of the API including the `Array` and `Group` classes and the top-level
+API (e.g. `zarr.open_array` and `zarr.open_group`). It is worth noting that we significantly evolved the internal data model,
+moving away from a model that was tightly coupled to the v2 specification, and to a more generic representation of Zarr objects.
```
dstansby, line 22: This sentence raises more questions (why did we evolve the data model? how did we evolve the data model? what does the new model look like?), and I don't think is needed for a simple compatability guide, so I think best to drop?
```suggestion
API (e.g. `zarr.open_array` and `zarr.open_group`).
```

docs/migration.rst
```diff
@@ -0,0 +1,89 @@
...
+
+Getting ready for 3.0
+---------------------
+
+Ahead of the 3.0 release, we suggest projects that depend on Zarr-Python take the following actions:
+
+1. Pin the supported Zarr-Python version to ``zarr>=2,<3``. This is a best practice and will protect your users from any incompatibilities that may arise during the release of Zarr-Python 3.0.
+2. Limit your imports from the Zarr-Python package. Most of the primary API ``zarr.*`` will be compatible in 3.0. However, the following breaking API changes are planned:
```
dstansby, line 30: Shouldn't this list have

3. Test that your package works with v3
4. Update the pin to zarr >=3

At the end?

docs/migration.rst
```diff
@@ -0,0 +1,89 @@
...
+   
+   - ``numcodecs.*`` will no longer be available in ``zarr.*``. (Suggested action: transition to importing codecs from ``numcodecs`` directly.)
```
dstansby, line 32:
```suggestion
   - ``zarr.numcodecs.*`` will no longer be available. These imports can be replaced by importing ``numcodecs`` directly.
```
We should be doing more than "suggesting", we should be providing concrete fixes/code updates!

docs/migration.rst
```diff
@@ -0,0 +1,89 @@
...
+   - The ``zarr.v3_api_available`` feature flag is being removed. (Suggested action: this experimental feature was deprecated in v2.18.)
```
dstansby, line 33:
```suggestion
   - The ``zarr.v3_api_available`` feature flag is being removed. In zarr-python v3 the v3 API is always available, so you shouldn't need to use this flag.
```

docs/migration.rst
```diff
@@ -0,0 +1,89 @@
...
+    - ``zarr.core``
+    - ``zarr.hierarchy``
+    - ``zarr.indexing``
+    - ``zarr.meta``
+    - ``zarr.meta_v1``
+    - ``zarr.storage``
+    - ``zarr.sync``
+    - ``zarr.types``
+    - ``zarr.util``
+    - ``zarr.n5``
```
dstansby, line 48: This needs a clear list of how to update code to adapt to these removals or changes.

docs/migration.rst
```diff
@@ -0,0 +1,89 @@
...
+
+Continue using Zarr-Python 2
+----------------------------
+
+Zarr-Python 2.x is still available, though we recommend migrating to Zarr-Python 3 for its improvements and new features.
+Security and bug fixes will be made to the 2.x series for at least 6 months following the first Zarr-Python 3 release.
+
+If you need to use the latest Zarr-Python 2 release, you can install it with:
+
+    $ pip install "zarr==2.*"
```
dstansby, line 58: I don't think this needs a whole section - perhaps just mention zarr-python 2 is available for at least 6 months above somewhere?

dstansby (changes requested):
I left some comments - I'll mark this as request changes since it still needs major additions.

docs/migration.rst
jhamman: I wish it were that easy! Unfortunately, the top level namespace is littered with `*` imports

https://github.com/zarr-developers/zarr-python/blob/e1d98cdf3d35e4cb2815912d82a64c1623347c87/zarr/__init__.py#L2
https://github.com/zarr-developers/zarr-python/blob/e1d98cdf3d35e4cb2815912d82a64c1623347c87/zarr/codecs.py#L2

If there is a reasonable way for us to deprecate folks using these imports, I'm all ears.

docs/migration.rst
jhamman: I see what you are saying but this has come up a lot. In this case, I'd like to be clear, even if it is obvious.

docs/migration.rst
```diff
@@ -0,0 +1,89 @@
...
+    - ``zarr.hierarchy``
+    - ``zarr.indexing``
+    - ``zarr.meta``
+    - ``zarr.meta_v1``
+    - ``zarr.storage``
+    - ``zarr.sync``
+    - ``zarr.types``
+    - ``zarr.util``
+    - ``zarr.n5``
+
```
jhamman, line 49:
```suggestion

3. Test that your package works with v3. You can start testing against version 3 now (pre-releases are being published to PyPI weekly).
4. Update the pin to zarr >=3
```

docs/migration.rst
```diff
@@ -0,0 +1,89 @@
...
+---------------------
+
+Ahead of the 3.0 release, we suggest projects that depend on Zarr-Python take the following actions:
+
+1. Pin the supported Zarr-Python version to ``zarr>=2,<3``. This is a best practice and will protect your users from any incompatibilities that may arise during the release of Zarr-Python 3.0.
+2. Limit your imports from the Zarr-Python package. Most of the primary API ``zarr.*`` will be compatible in 3.0. However, the following breaking API changes are planned:
+   
+   - ``numcodecs.*`` will no longer be available in ``zarr.*``. (Suggested action: transition to importing codecs from ``numcodecs`` directly.)
+   - The ``zarr.v3_api_available`` feature flag is being removed. (Suggested action: this experimental feature was deprecated in v2.18.)
+   - The following internal modules are being removed or significant changed:
```
jhamman, line 34:
```suggestion
   - The following internal modules are being removed or significantly changed. If your application relies on imports from any of the below modules, you will need to either a) modify your application to no longer rely on these imports or b) vendor the parts of the specific modules that you need.
```

docs/migration.rst
```diff
@@ -0,0 +1,90 @@
...
+    - ``zarr.meta_v1``
+    - ``zarr.storage``
+    - ``zarr.sync``
+    - ``zarr.types``
+    - ``zarr.util``
+    - ``zarr.n5``
+
+3. Test that your package works with v3. You can start testing against version 3 now (pre-releases are being published to PyPI weekly).
+4. Update the pin to zarr >=3
+Continue using Zarr-Python 2
```
jhamman, line 51:
```suggestion

Continue using Zarr-Python 2
```

jni:
I just came here following the link from #1849. Currently, this guide feels very incomplete, and since #2182 is now in-flight, I think this guide should become very clear.

From the perspective of a maintainer / close user of a bunch of libraries that depend on zarr *but have extremely limited maintainer time*, the most important thing I want to understand is how hard it will be to support *both* zarr 2.18 and zarr 3+ within a single library. I think this is *the* critical question for a smooth transition, because it is hard for libraries to all migrate at the same time, and you want libraries to be installable together in the same environment — you don't want someone depending on both napari and ome-zarr to face napari requiring zarr>=3 and ome-zarr requiring zarr<3. So many libraries would want to support the subset of zarr that is identical in v3 and v2 until everyone can agree to depend on 3+.

docs/migration.rst
dstansby: Even if there isn't a reasonable way to deprecate the imports, as a minimum this migration guide should provide a complete list of all imports that are disappearing and how to replace them.

jhamman:
@jni - thanks for the feedback. I agree this is not ready to ship yet. The main things that we know are changing is the `Store` API and access to internal APIs (e.g. `zarr.core.xxx`). Beyond that, the best way for us to fill out the migration guide is to have projects attempt to support v3 and report back. I'm doing that with Dask right now (https://github.com/dask/dask/pull/11388, https://github.com/zarr-developers/zarr-python/pull/2186) and I understand @TomAugspurger has begun the process for Xarray. So an ask for you and the Napari / ome-zarr devs is to try to do this and report back. Beyond that, specific suggestions to this doc are more than welcome (props to @dstansby for his edits already).

dstansby:
If we want downstream packages to test with version 3.0.0a1, it might be good to do a blog post or add something to the docs explaining how to do that testing, what to look for, and how to provide feedback?

jni:
Well, @Czaki started doing that for us in napari/napari#7215 and @d-v-b has [been helping](https://napari.zulipchat.com/#narrow/stream/212875-general/topic/handling.20zarr.203.2E0.20release/near/445118833) 🙏.

It looks like one of the remaining issues is that `zarr.open` defaults to v3 zarr (it'd be worth considering switching to calver, which I hate, *if only* to avoid the confusion between zarr format v3 and zarr-python v3... 😂), and tensorstore does not yet support v3 zarr files. At least that's my interpretation of [these lines](https://github.com/napari/napari/pull/7215/files#diff-7e3cbb11d497f0a4fe3344e5d99e8453b9037f5a073bc5642e00ebca69af6316R1119). Nor do I see any motions to change this in the tensorstore repo... @jbms?

But I think we can resolve this by explicitly writing a v2 zarr in the test?

> specific suggestions to this doc are more than welcome

Something along the lines of:

> Common functions have switched to keyword-only arguments, so you will need to change any invocation of, for example, `zarr.open(path, 'a')` to `zarr.open(store=path, mode='a')`.

(An exhaustive list of such changes would be useful.)

d-v-b:
> and tensorstore does not yet support v3 zarr files.

Tensorstore has support zarr v3 for a long time: https://google.github.io/tensorstore/driver/zarr3/index.html

jni:
> Tensorstore has supported zarr v3 for a long time

oh, interesting, thanks for pointing that out @d-v-b! I couldn't actually find the relevant PR but did not look exhaustively. 🙏

TomAugspurger:
A few changes I've found while updating xarray. Are all of these intentional?

- `Array.resize` returns a new `Array` object. 2.x mutated the `Array` in place
- `zarr_version` has been renamed to `zarr_format`
- Some exception types have changed (e.g. 2.x raised a `zarr.errors.GroupNotFoundError` while 3.x raises a `ValueError`)
- `write_empty_chunks` has been removed

PR #2106, opened 2024-08-22
Fix Array.__array__ for numpy 2.1

dstansby:
This updates __array__ for numpy 2.1, specifically by implementing the copy keyword argument.

The one test failure is an issue with a new version of numcodecs (0.13.0) which is a simple fix, but easier to make when we bump the numocdecs version in the test dependencies, so I'd advocate for merging this PR and dealing with the numcodecs bump in a follow up PR.

TODO:

### DIFF
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -24,13 +24,22 @@ Release notes
 2.18.3
 ------
 
+Enhancements
+~~~~~~~~~~~~
+* Added support for creating a copy of data when converting a `zarr.Array`
+  to a numpy array.
+  By :user:`David Stansby <dstansby>`
+
 Maintenance
 ~~~~~~~~~~~
 * Removed support for Python 3.9.
   By :user:`David Stansby <dstansby>`
-  
+
 * Fix a regression when using orthogonal indexing with a scalar.
   By :user:`Deepak Cherian <dcherian>` :issue:`1931`
+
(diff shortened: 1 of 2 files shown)


joshmoore:
The only immediate question in my mind is whether this adds any additional copies of the array.

dstansby:
Using the test script below shows that with

- current zarr and `copy` not set, a copy is made
- current zarr and `copy=False` errors
- this PR and `copy` not set, a copy is made
- this PR and `copy=False`, a copy is *not* made

So behaviour with this PR is the same as before.


```python
import numpy as np
import zarr

z = zarr.zeros((100, 100), chunks=(10, 10), dtype='i4')
print("Copy not set")
print("Same memory?", id(z[:]) == id(np.array(z)))
```

jhamman (approved):

jhamman:
@dstansby - have you looked into the test failure in the `minimum_build`. Certainly looks unrelated but something we should probably fix.

dstansby:
Yes, in the top comment here I said:

> The one test failure is an issue with a new version of numcodecs (0.13.0) which is a simple fix, but easier to make when we bump the numocdecs version in the test dependencies, so I'd advocate for merging this PR and dealing with the numcodecs bump in a follow up PR.

joshmoore:
Trusting @dstansby to be Mr. Fixit! :smile:

PR #2113, opened 2024-08-23
Zarr-v3 Consolidated Metadata

### LINKED ISSUE #1161: V3 Consolidated Metadata
follow on to: https://github.com/zarr-developers/zarr-python/pull/1142
background reading: https://github.com/zarr-developers/zarr-specs/issues/136

I would like to propose that we remove the machinery for consolidated metadata in the current V3 implementation in favor of an upcoming specification. This would include:

 - dropping `zarr._storage.v3.ConsolidatedMetadataStoreV3`
 - disallow version=v3 in `zarr.convenience.consolidated_metadata`

TomAugspurger:
Implements the optional Consolidated Metadata feature of zarr-v3: https://github.com/zarr-developers/zarr-specs/pull/309, along with reading zarr v2 metadata.

The implementation defines a new dataclass: `ConsoliatedMetadata`. It's an optional field on the existing `GroupMetadata` object. Opening as a draft until the PR to the zarr-specs repo finishes up.

closes #1161 

TODO:


src/zarr/codecs/crc32c_.py
```diff
@@ -37,7 +37,7 @@ async def _decode_single(
         crc32_bytes = data[-4:]
         inner_bytes = data[:-4]
 
-        computed_checksum = np.uint32(crc32c(inner_bytes)).tobytes()
+        computed_checksum = np.uint32(crc32c(inner_bytes)).tobytes()  # type: ignore[arg-type]
```
TomAugspurger, line 40: Locally, I see this without the type ignore

```
src/zarr/codecs/crc32c_.py:40: error: Argument 1 to "crc32c" has incompatible type "ndarray[Any, dtype[Any]]"; expected "Buffer"  [arg-type]
Found 1 error in 1 file (checked 66 source files)
```

I don't suppose anyone else gets that?

src/zarr/core/array.py
```diff
@@ -372,9 +372,15 @@ async def open(
         else:
             # V3 arrays are comprised of a zarr.json object
             assert zarr_json_bytes is not None
+            zarr_metadata = json.loads(zarr_json_bytes.to_bytes())
```
TomAugspurger, line 375: I'm not too happy with the need for this... Without it, we eventually try to load the metadata as an ArrayMetadata, which throws a `ValueError` IIRC. But the parent does a

```
try:
    open_array(...)
except KeyError:
    open_group(...)
```

Might see if we can clean that up. At least for `v3` we can key off the `node_type` once we've loaded the JSON metadata.

src/zarr/core/common.py
```diff
@@ -167,3 +169,24 @@ def parse_order(data: Any) -> Literal["C", "F"]:
     if data in ("C", "F"):
         return cast(Literal["C", "F"], data)
     raise ValueError(f"Expected one of ('C', 'F'), got {data} instead.")
+
+
+def _json_convert(o: Any) -> Any:
```
TomAugspurger, line 174: Moved here so that I can use it in a couple of spots (but I think I forgot to replace the original usage of this).

All serialization stuff was a bit of a pain. I might look into cleaning that up a bit as a precursor.

src/zarr/core/group.py
```diff
@@ -101,7 +150,17 @@ def to_buffer_dict(self, prototype: BufferPrototype) -> dict[str, Buffer]:
                 ),
             }
 
-    def __init__(self, attributes: dict[str, Any] | None = None, zarr_format: ZarrFormat = 3):
+    # def __post_init__(self):
```
TomAugspurger, line 153: This is leftover from testing and can be removed. FWIW, `__post_init__` seemed to work fine.

src/zarr/codecs/crc32c_.py
jhamman: I get these locally but not in our CI. I haven't bothered to figure out why that is.

TomAugspurger:
This isn't handling nested groups correctly yet. My understanding (and I'll clarify this in the spec) is that given a structure like

```
/root
  /arr-1
  /arr-2
  /child-group
    /child-arr-1
    /child-arr-2
```

we'd expect that the metadata from child-group, child-arr-1, and child-arr-2 should end up in the consolidated metadata.

src/zarr/core/group.py
```diff
@@ -101,20 +150,33 @@ def to_buffer_dict(self, prototype: BufferPrototype) -> dict[str, Buffer]:
...
         attributes_parsed = parse_attributes(attributes)
         zarr_format_parsed = parse_zarr_format(zarr_format)
 
         object.__setattr__(self, "attributes", attributes_parsed)
         object.__setattr__(self, "zarr_format", zarr_format_parsed)
+        object.__setattr__(self, "consolidated_metadata", consolidated_metadata)
 
     @classmethod
     def from_dict(cls, data: dict[str, Any]) -> GroupMetadata:
+        data = dict(data)
```
jhamman, line 168: are we sure this is needed? seems like mypy would scream at us if `data` was not already a dict.

jhamman (commented):
Good start here Tom! I have a few questions about parts you haven't implemented yet that should help explore the design a bit more:

1. How are you thinking about wiring this into read operations (e.g. `getitem`, `listdir`, etc.)? 
2. How should we think about consistency w.r.t. the consolidated metadata field? When should we invalidate the consolidated metadata and when is it safe to keep using it?
3. How can this approach be used to support reading/writing v2 consolidated metadata (`.zmetadata`)?

src/zarr/core/common.py
TomAugspurger: Hmm #1851 looks interesting. I started down a path of a pretty generic serialization / deserialization implementation at https://github.com/TomAugspurger/zarr-python/blob/feature/serde/src/zarr/_serialization.py. That's decoupled stores and mostly decoupled from Metadata objects. Though not necessarily using JSON might complicate things a bit.

This is a whole can of worms, which is why libraries like msgspec, cattrs, etc exist. I do think it's worth trying to make this work without taking a dependency on any of those first. I'll push that branch a little bit further and see how things work.

TomAugspurger:
Quick update here:

- `AsyncGroup` now uses consolidated metadata in some operations (primarily `getitem`. I'm going through some more spots now). This means you can do `AsyncGroup.getitem(key)` to get a child node without any additional I/O.
- https://github.com/zarr-developers/zarr-python/pull/2113/commits/fc901eb5be0e4aa113cd913feb6a135223ba9bb0 added support for reading zarr V2 consolidated metadata. Needs some cleanup and probably more testing, but the basics seem to work.

src/zarr/api/asynchronous.py
```diff
@@ -129,8 +138,38 @@ def _default_zarr_version() -> ZarrFormat:
...
+
+    members_metadata = {k: v.metadata for k, v in members.items()}
+
+    consolidated_metadata = ConsolidatedMetadata(metadata=members_metadata)
+    metadata = dataclasses.replace(group.metadata, consolidated_metadata=consolidated_metadata)
+    group = dataclasses.replace(
+        group,
+        metadata=metadata,
+    )
+    await group._save_metadata()
```
TomAugspurger, line 171: We should probably check that the store is writable as the first thing in this method, in case consolidating metadata takes some time.

src/zarr/core/group.py
```diff
@@ -149,15 +234,26 @@ async def open(
         cls,
         store: StoreLike,
         zarr_format: Literal[2, 3, None] = 3,
+        open_consolidated: bool = False,
```
TomAugspurger, line 237: I'm not sure about this API yet. I only added it for Zarr V2. I think the main question to answer is:

What top-level behavior do we want if a user does `zarr.open_consolidated` on a hierarchy that doesn't have consolidated metadata (v2 or v3)? Silently not use consolidated metadata, or error loudly?

src/zarr/core/group.py
```diff
@@ -149,15 +234,26 @@ async def open(
...
     ) -> AsyncGroup:
         store_path = await make_store_path(store)
 
         if zarr_format == 2:
-            zgroup_bytes, zattrs_bytes = await asyncio.gather(
-                (store_path / ZGROUP_JSON).get(), (store_path / ZATTRS_JSON).get()
+            paths = [store_path / ZGROUP_JSON, store_path / ZATTRS_JSON]
+            if open_consolidated:
+                paths.append(store_path / ".zmetadata")  # todo: configurable
```
TomAugspurger, line 244: The key `.zmetadata` was configurable in zarr V2 I think. Maybe `open_consolidated` needs to accept a string to indicate which key to use? Or another kwarg?

src/zarr/core/group.py
```diff
@@ -77,18 +79,65 @@ def _parse_async_node(node: AsyncArray | AsyncGroup) -> Array | Group:
         raise TypeError(f"Unknown node type, got {type(node)}")
 
 
+@dataclass(frozen=True)
+class ConsolidatedMetadata:
+    metadata: dict[str, ArrayMetadata | GroupMetadata]
+    kind: Literal["inline"] = "inline"
```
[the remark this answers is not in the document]
TomAugspurger: I added support for Zarr V2 reading, and I think it worked out OK... We're able to basically translate from "external" to "inline" metadata similar to how we handle attrs. The V2 reader will read the attributes and then stuff them in the dict used to create the `Metadata` object under the "attributes" key.

This is I think a nice way of doing things. In memory, we only need to worry about the V3 data model. When we're reading V2 data, we can (hopefully robustly) translate from v2 to v3.

src/zarr/core/group.py
```diff
@@ -79,18 +82,87 @@ def _parse_async_node(node: AsyncArray | AsyncGroup) -> Array | Group:
...
+                else:
+                    # We either have V2 metadata, or invalid metadata
+                    if "shape" in v:
+                        # probably ArrayV2Metadata
+                        metadata[k] = ArrayV2Metadata.from_dict(v)
+                    else:
+                        # probably v2 Group metadata
+                        metadata[k] = GroupMetadata.from_dict(v)
+
+        # assert data["kind"] == "inline"
```
TomAugspurger, line 138: Remove this dead code.

src/zarr/core/group.py
```diff
@@ -79,18 +82,87 @@ def _parse_async_node(node: AsyncArray | AsyncGroup) -> Array | Group:
...
+        if data["kind"] != "inline":
+            raise ValueError
```
TomAugspurger, line 140: Add good error messages here and a few lines down.

src/zarr/core/group.py
```diff
@@ -79,18 +82,87 @@ def _parse_async_node(node: AsyncArray | AsyncGroup) -> Array | Group:
...
+
+        if data["must_understand"] is not False:
+            raise ValueError
+        return cls(metadata=metadata)
+
+    def _filter_prefix(self, key: str) -> ConsolidatedMetadata:
```
TomAugspurger, line 146: I think this is unused. I probably inlined it?

src/zarr/core/group.py
TomAugspurger: It's already a dict. This is just to copy the refs so that we don't mutate the `data` passed in by the caller.

src/zarr/core/group.py
```diff
@@ -149,15 +234,26 @@ async def open(
...
         store_path = await make_store_path(store)
 
         if zarr_format == 2:
-            zgroup_bytes, zattrs_bytes = await asyncio.gather(
-                (store_path / ZGROUP_JSON).get(), (store_path / ZATTRS_JSON).get()
+            paths = [store_path / ZGROUP_JSON, store_path / ZATTRS_JSON]
+            if open_consolidated:
+                paths.append(store_path / ".zmetadata")  # todo: configurable
+
+            zgroup_bytes, zattrs_bytes, *rest = await asyncio.gather(
```
TomAugspurger, line 246: The `*rest` business is to concurrently read the V2 consolidated metadata at the same time as the `.zarray` and `.zattrs`. Slightly complicated, but IMO worth it.

src/zarr/core/group.py
```diff
@@ -188,6 +284,37 @@ async def open(
             zgroup = json.loads(zgroup_bytes.to_bytes())
             zattrs = json.loads(zattrs_bytes.to_bytes()) if zattrs_bytes is not None else {}
             group_metadata = {**zgroup, "attributes": zattrs}
+
+            if open_consolidated:
```
TomAugspurger, line 288: This could probably use a refactor. I think starting at the `if zarr_format == 2` we're all done with I/O, so we could have another constructor like `def _from_raw(...)` that doesn't have to be async.

src/zarr/core/group.py
```diff
@@ -214,13 +341,51 @@ async def getitem(
         store_path = self.store_path / key
         logger.debug("key=%s, store_path=%s", key, store_path)
 
+        # Consolidated metadata lets us avoid some I/O operations so try that first.
+        if self.metadata.consolidated_metadata is not None:
```
TomAugspurger, line 345: Consider splitting this into a `_getitem_consolidated` and `_getitem_unconsolidated`?

src/zarr/core/group.py
```diff
@@ -214,13 +341,51 @@ async def getitem(
...
+                            },
+                        ),
+                    )
+                    return AsyncGroup(metadata=metadata, store_path=store_path)
+                else:
+                    # TODO: order?
+                    return AsyncArray(metadata=metadata, store_path=store_path)
+            else:
+                # v2
+                raise NotImplementedError
```
TomAugspurger, line 381: Do this.

src/zarr/core/group.py
```diff
@@ -214,13 +341,51 @@ async def getitem(
...
+                            metadata={
+                                k.split(key, 1)[-1].lstrip("/"): v
+                                for k, v in self.metadata.consolidated_metadata.metadata.items()
+                                if k.startswith(key) and k != key
+                            },
+                        ),
+                    )
+                    return AsyncGroup(metadata=metadata, store_path=store_path)
+                else:
+                    # TODO: order?
```
TomAugspurger, line 377: I think order is fine to ignore here...

src/zarr/core/group.py
```diff
@@ -628,6 +799,7 @@ async def members(
     async def _members(
         self, max_depth: int | None, current_depth: int
     ) -> AsyncGenerator[tuple[str, AsyncArray | AsyncGroup], None]:
+        # todo!
```
TomAugspurger, line 802: Do this.

TomAugspurger:
As I start to use this a bit, I'm rethinking the in-memory representation of consolidated metadata for a nested hierarchy. Specifically the `consolidated_metadata.metadata` dictionary which maps keys to `ArrayMetadata | GroupMetadata` objects. Our options are:

1. A flat structure, where the keys include all path segments from the root group and the root `ConsolidatedMetadata.metadata` dict is equal to the length of *all* child nodes (not just immediate children). This matches the representation on disk.
2. A nested structure, where the keys include *just* the name (so not the leading segments of the path). The root `ConsolidatedMetadata.metadata` dict holds just immediate children nodes. Metadata for nested groups can still be accessed, but through the `.consolidated_metadata` on the children. This matches the logical representation.

The motivation for this rethink is from the current implementation having to be careful about just using `consolidated_metdata.metadata` directly. If you want to *propagate* consolidated metadata, you need to use the `Group.getitem` method. See `Group.members` where this becomes relevant.

Here's an example:

Given a hierarchy like

```
root/
  g0/
   c0/
    array0
    array1
   c1/
     array0
     array1
  g1/
    c0/
      array0
      array1
    c1/
      array0
      array1
```

We'll represent the consolidated metadata as a flat mapping of (store) keys to values.

```
"g0": {"attributes": ..., "node_type": "group"},
"g1": {"attributes": ..., "node_type": "group"},
"g0/c0": {"attributes": ..., "node_type": "group"},
...
"g0/c0/array0": {"shape": [...], "node_type": "array"},
...
"g1/c1/array1": {"shape": [...], "node_type": "array"}
```

But *in memory*, what is the `Group.metadata.consolidated_metadata` for each of theses groups?

Should it match the flat structure on disk, where the keys imply the structure?

```
# g0
GroupMetadata(
    attributes=...,
    consolidated_metadata=ConsolidatedMetadata(
	    metadata={
		    "c0": GroupMetadata(...),
		    "c0/array0": ArrayMetadata(...),
		    ...,
		    "c1/array1": ArrayMetadata(...),
		},
)
```

Or should it have a nested / tree-like structure, where just immediate children appear in `group.metadata.consolidated_metadata.metadata`, and nested members can be accessed through that dictionary?

```
# g0
GroupMetadata(
    attributes=...,
    consolidated_metadata=ConsolidatedMetadata(
	    metadata={
		    "c0": GroupMetadata(
			    attributes=...,
			    consoliated_metadata=ConsolidatedMetadata(
				    metadata={
					    "array0": ArrayMetadata(...),
					    "array1": ArrayMetadata(...),
				    }
			    )
			),
			"c1": GroupMetadata(
				attributes=...,
				consolidated_metadata=ConsolidatedMetadata(
					metadata={
					    "array0": ArrayMetadata(...),
					    "array1": ArrayMetadata(...),
					}
				)
			)
		},
)
```

Right now I've implemented option 1. I'll out option 2 today.

d-v-b:
do we have to pick just 1 in-memory representation? Over in `pydantic-zarr` I do something similar to metadata consolidation, and I use a tree representation or a flat `dict[str, Array | Group]` representation situationally (see the [`to_flat`](https://janelia-cellmap.github.io/pydantic-zarr/api/v2/#pydantic_zarr.v2.to_flat) and [`from_flat`](https://janelia-cellmap.github.io/pydantic-zarr/api/v2/#pydantic_zarr.v2.from_flat) functions).

TomAugspurger:
I think we'll be able to translate between the two pretty easily. I think we'll want methods like you shared to convert between the two.

I think the main choice is what to if we want a "default" representation that's exposed through `GroupMetadata.consolidated_metadata.metadata`. I think we do, and I'm leaning towards nested being the default.

src/zarr/core/group.py
```diff
@@ -264,19 +559,51 @@ async def getitem(
...
                 (store_path / ZARRAY_JSON).delete(),  # TODO: missing_ok=False
                 (store_path / ZATTRS_JSON).delete(),  # TODO: missing_ok=True
             )
+
         else:
             raise ValueError(f"unexpected zarr_format: {self.metadata.zarr_format}")
 
+        if self.metadata.consolidated_metadata:
+            self.metadata.consolidated_metadata.metadata.pop(key, None)
+            # FIXME: This should probably rewrite all the consolidated metadata?
```
TomAugspurger, line 604: How should we think about mutating operations (`setitem`, `delitem`) on `Group` (and Array) objects, and how they relate to the state in the backing `store`? And how does consolidated metadata factor in here?

My understanding is that mutating operations should be "immediately" reflected in the backing store, as part of the function call.

I *think* things are currently written such that the in-memory `Group` will always reflect the status of the *non*-consolidated metadata. If you do a `setitem` or `delitem`, it'll be set / deleted from both the in-memory Group and the backing store synchronously. Should we also try to write out the consolidated metadata on each operation?

Notably, we *shouldn't need to list each member* when *re*-consolidating the metadata. We should already know all the members thanks to the starting consolidated metadata and all the operations we've done on it (setting aside the case of multiple clients / Groups concurrently mutating the same store... that's a problem; we'll need a way to "force" reconsolidating).

I'm pretty conflicted here. I don't think there's an obviously correct answer here. The safest is *probably* to automatically re-consolidate though. Maybe the best option is a flag on the Group or Zarr config to control the behavior?

TomAugspurger:
What behavior do we want by default when users do `zarr.open_group`, and should that behavior depend on whether or not the hierarchy being opened is v2 or v3?

As currently written, this PR uses consolidated metadata by default for v3 data but not for v2 data. Initially my thought was to match the behavior of zarr-python for v2 data, but IMO the default behavior should depend just on the library version, not on the data being read.

With that in mind, do we have a preference for using consolidated metadata by default or not in `zarr.open_group`? I don't have a strong preference.

src/zarr/core/group.py
jhamman: Sorry @TomAugspurger this question didn't get answered sooner. Here's my perspective:

1. Mutating a consolidated store is inherently unsafe. In the case of multiple writers, there is not way ensure consistency without a lock/transaction layer in the store.
2. Readers should be able to opt in / out of trusting consolidated metadata. (some stores, like http, may require it)
3. Hierarchy and metadata mutating operations should probably invalidate the consolidated metadata.

TomAugspurger: Thanks. I think I'm mostly aligned on that. Some clarification on 2 and 3:

1. If I open a Group, bypassing consolidated metadata with `use_consolidated=False`, I would expect mutating operations on that group to *not* touch the consolidated metadata (if any).
2. If I open a Group with consolidated metadata, should mutating operations *invalidate* the consolidated metadata (by deleting that entire key from the Group metadata) or *reconsolidate*?

Assuming we don't have concurrent writers mutating the same consolidated metadata (which we'll need to document), then invalidation and re-consolidation are both just a single atomic operation on the store (write the in-memory Group's metadata (including consolidated metadata) to the store).

jhamman: > If I open a Group with consolidated metadata, should mutating operations invalidate the consolidated metadata (by deleting that entire key from the Group metadata) or reconsolidate?

There is a 3rd option following a mutation -- do nothing! This is what we do in 2.x. I personally favor the `invalidate` option but we could also follow the behavior from 2.x and simply tell the user (through documentation or a warning) that they need to reconsolidate.

TomAugspurger: > I personally favor the invalidate option

I'm coming around to that too. Initially, I favored "reconsolidate" since we can do that *without re-listing* so it would be as expensive as invalidating (which will just remove the `consolidated_metadata` key from the root zarr.json. But that is only safe to do when there's a single concurrent writer mutating the group.

The following would lead to inconsistent consolidated metadata:

```python
store = ...
await consolidate_metadata(store)

root = await open_group(store)
root2 = await open_group(store)

await asyncio.gather([root1.delitem("array-0"), root2.delitem("array-1")])
```

Whichever one ran last would "win", but the consolidated metadata would still contain the metadata for the other array.

So for now I think this is best solved through documentation. In the future, I hope that we can implement a better interface between the Store and Group that will allow for unit-of-work style updates to the metadata in-memory, and the ability to safely persist those in the store.

TomAugspurger:
I think the last task here is to implement writing V2 consolidated metadata (that's causing the test failures).

TomAugspurger:
There's a behavior change here that's causing some xarray tests to fail. It comes down to whether we interpret `zarr.open_consolidated(store, path="path/to/group")` as

1. Open the consolidated metadata at the root of the store and then select the (nested) group `path/to/group`.
2. Open the consolidated metadata at `"path/to/group"`

Zarr 2.x implemented the first one. This branch currently implements the second version.

The rough flow is

1. create a Group / arrays at some level down in the store: e.g. `a/b/group` with arrays `a/b/group/x`, `a/b/group/y`, ...
2. Consolidate the store *at the root level* with `zarr.consolidate_metadata(store)`.
3. open the (nested) group with `zarr.open_consolidated(store=store, path="a/b/group")`

I'll see how hard it is to implement what 2.x was doing, but I want to confirm that's the behavior we want.

src/zarr/core/group.py
TomAugspurger: The behavior is documented as

1. `open` uses consolidated metadata if present, doesn't if not.
2. Users can set `use_consolidated=True` to error if it's not found, `False` to bypass it if it's not present.
3. `open_consolidated` sets `use_consolidated=True` which will error if consolidated metadata is not found.

src/zarr/core/group.py
```diff
@@ -199,15 +524,87 @@ async def open(
...
+
+            return cls._from_bytes_v2(
+                store_path, zgroup_bytes, zattrs_bytes, maybe_consolidated_metadata_bytes
+            )
         else:
             # V3 groups are comprised of a zarr.json object
             assert zarr_json_bytes is not None
-            group_metadata = json.loads(zarr_json_bytes.to_bytes())
+            if not isinstance(use_consolidated, bool | None):
+                raise TypeError("use_consolidated must be a bool for Zarr V3.")
```
jhamman, line 546:
```suggestion
                raise TypeError("use_consolidated must be a bool or None for Zarr V3.")
```

src/zarr/core/group.py
```diff
@@ -199,15 +524,87 @@ async def open(
...
+        zgroup = json.loads(zgroup_bytes.to_bytes())
+        zattrs = json.loads(zattrs_bytes.to_bytes()) if zattrs_bytes is not None else {}
+        group_metadata = {**zgroup, "attributes": zattrs}
+
+        if consolidated_metadata_bytes is not None:
+            v2_consolidated_metadata = json.loads(consolidated_metadata_bytes.to_bytes())
+            v2_consolidated_metadata = v2_consolidated_metadata["metadata"]
+            # We already read zattrs and zgroup. Should we ignore these?
+            v2_consolidated_metadata.pop(".zattrs")
+            v2_consolidated_metadata.pop(".zgroup")
```
jhamman, line 572: In the future, we may choose to inform the caller that the consolidated metadata does not match the root group metadata but for now, this seems fine.

src/zarr/core/group.py
```diff
@@ -706,6 +1157,11 @@ async def _members(
         # and scoped to specific zarr versions
         _skip_keys = ("zarr.json", ".zgroup", ".zattrs")
 
+        # hmm lots of I/O and logic interleaved here.
+        # We *could* have an async gen over self.metadata.consolidated_metadata.metadata.keys()
+        # and plug in here.l `getitem` will skip I/O.
+        # Kinda a shape to have all the asyncio task overhead though, when it isn't needed.
```
jhamman, line 1163:
```suggestion
        # Kinda a shame to have all the asyncio task overhead though, when it isn't needed.
```

😜

src/zarr/core/group.py
```diff
@@ -747,7 +1223,9 @@ async def contains(self, member: str) -> bool:
             return True
 
     async def groups(self) -> AsyncGenerator[tuple[str, AsyncGroup], None]:
-        async for name, value in self.members():
+        members = [x async for x in self.members()]
+        # async for name, value in self.members():
+        for name, value in members:
```
jhamman, line 1228: Why was this needed?

jhamman (commented):
@TomAugspurger -- this came together so nicely! Way to go 😄 

I left a few comments of substance but I think this can go in very soon (tomorrow?)

src/zarr/api/asynchronous.py
```diff
@@ -251,8 +312,11 @@ async def open(
         return await open_group(store=store_path, zarr_format=zarr_format, **kwargs)
 
 
-async def open_consolidated(*args: Any, **kwargs: Any) -> AsyncGroup:
-    raise NotImplementedError
+async def open_consolidated(*args: Any, use_consolidated: bool = True, **kwargs: Any) -> AsyncGroup:
+    """
+    Alias for :func:`open_group` with ``use_consolidated=True``.
+    """
+    return await open_group(*args, use_consolidated=use_consolidated, **kwargs)
```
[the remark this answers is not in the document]
TomAugspurger: My goal here is to avoid a user accidentally passing `use_consolidated` in `**kwargs` and us silently overwriting it. I'll update this to raise if `use_consolidated` isn't True.

src/zarr/core/group.py
```diff
@@ -82,42 +89,310 @@ def _parse_async_node(node: AsyncArray | AsyncGroup) -> Array | Group:
         raise TypeError(f"Unknown node type, got {type(node)}")
 
 
+def _json_convert(o: object) -> Any:
```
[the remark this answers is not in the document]
TomAugspurger: Seems like they matched exactly, so I think we can just use it here.

src/zarr/core/group.py
```diff
@@ -82,42 +89,310 @@ def _parse_async_node(node: AsyncArray | AsyncGroup) -> Array | Group:
...
+                elif node_type == "array":
+                    metadata[k] = ArrayV2Metadata.from_dict(v)
+                else:
+                    # We either have V2 metadata, or invalid metadata
+                    if "shape" in v:
+                        # probably ArrayV2Metadata
+                        metadata[k] = ArrayV2Metadata.from_dict(v)
+                    else:
+                        # probably v2 Group metadata
+                        metadata[k] = GroupMetadata.from_dict(v)
```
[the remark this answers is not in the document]
TomAugspurger: b5fb72153dfde336e269362fc1dc0a70d573768e cleaned this up quite a bit. I'll call out this use of Python 3.11's `typing.assert_never()` which pairs nicely with some of the runtime value checking we do to ensure we have proper values / types for `zarr_format` and now `node_type`.

src/zarr/core/group.py
TomAugspurger: Leftover from debugging I think. Removing.

src/zarr/api/asynchronous.py
```diff
@@ -132,8 +141,61 @@ def _default_zarr_version() -> ZarrFormat:
...
+    ----------
+    store: StoreLike
+        The store-like object whose metadata you wish to consolidate.
+    path: str, optional
+        A path to a group in the store to consolidate at. Only children
+        below that group will be consolidated.
+
+        By default, the root node is used so all the metadata in the
+        store is consolidated.
+
```
TomAugspurger, line 165: document `zarr_format`

src/zarr/api/asynchronous.py
```diff
@@ -589,6 +662,22 @@ async def open_group(
         to users. Use `numpy.empty(())` by default.
     attributes : dict
         A dictionary of JSON-serializable values with user-defined attributes.
+        use_consolidated : bool or str, default None
```
TomAugspurger, line 665: fix indent.

src/zarr/core/array.py
```diff
@@ -165,6 +165,13 @@ async def get_array_metadata(
         # V3 arrays are comprised of a zarr.json object
         assert zarr_json_bytes is not None
         metadata_dict = json.loads(zarr_json_bytes.to_bytes())
+
+        node_type = metadata_dict.get("node_type")
```
TomAugspurger, line 169: Use `parse_node_type_array`

src/zarr/core/group.py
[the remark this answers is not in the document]
TomAugspurger: Forgot to actually delete this block.

jhamman:
Huge @TomAugspurger 🎉 🎉 🎉 !!!

d-v-b:
awesome work!

PR #2117, opened 2024-08-25
Fixed MemoryStore.list_dir

### LINKED ISSUE #2116: MemoryStore.list_dir incorrectly includes non-immediate children
### Zarr version

v3

### Numcodecs version

n/a

### Python Version

n/a

### Operating System

n/a

### Installation

n/a

### Description

MemoryStore.list_dir incorrectly includes an entry for each nested child of a group. This differs from the behavior of LocalStore.list_dir

### Steps to reproduce

This program sets up a store with some nested objects:

```
import zarr.store
import zarr.core.buffer


async def main():
    local = zarr.store.LocalStore("/tmp/foo", mode="w")
    mem = zarr.store.MemoryStore({}, mode="w")

    paths = [
        'root/a0/zarr.json',
        'root/g0/zarr.json',
        'root/g0/a1/zarr.json',
        'root/g0/g1/zarr.json',
        'root/g0/g1/a2/zarr.json',
        'root/g0/g1/g2/zarr.json',
    ]

    for store in [local, mem]:
        for path in paths:
            await store.set(path, zarr.core.buffer.Buffer.from_bytes(b""))

        print(type(store).__name__)
        print([x async for x in store.list_dir("root")])
        print("-" * 80)


if __name__ == "__main__":
    import asyncio
    asyncio.run(main())

```

Here's the output:

```
LocalStore
['g0', 'a0']
--------------------------------------------------------------------------------
MemoryStore
['a0', 'g0', 'g0', 'g0', 'g0', 'g0']
--------------------------------------------------------------------------------
```

The `g0` group is repeated once per child.

### Additional output

I'll have a PR shortly.

TomAugspurger:
Ensures that nested children are listed properly.

Closes #2116 

### DIFF
--- a/src/zarr/store/memory.py
+++ b/src/zarr/store/memory.py
@@ -117,6 +117,14 @@ class MemoryStore(Store):
             for key in keys_unique:
                 yield key
         else:
-            for key in self._store_dict:
-                if key.startswith(prefix + "/") and key != prefix:
-                    yield key.removeprefix(prefix + "/").split("/")[0]
+            # Our dictionary doesn't contain directory markers, but we want to include
+            # a pseudo directory when there's a nested item and we're listing an
+            # intermediate level.
+            n = prefix.count("/") + 2
+            keys_unique = {
+                "/".join(k.split("/", n)[:n])
+                for k in self._store_dict
+                if k.startswith(prefix + "/")
+            }
+            for key in keys_unique:
+                yield key.removeprefix(prefix + "/").split("/")[0]
--- a/src/zarr/store/memory.py
+++ b/src/zarr/store/memory.py
@@ -114,17 +114,15 @@ class MemoryStore(Store):
 
         if prefix == "":
             keys_unique = set(k.split("/")[0] for k in self._store_dict.keys())
-            for key in keys_unique:
-                yield key
         else:
             # Our dictionary doesn't contain directory markers, but we want to include
             # a pseudo directory when there's a nested item and we're listing an
             # intermediate level.
-            n = prefix.count("/") + 2
             keys_unique = {
-                "/".join(k.split("/", n)[:n])
-                for k in self._store_dict
-                if k.startswith(prefix + "/")
+                key.removeprefix(prefix + "/").split("/")[0]
+                for key in self._store_dict
+                if key.startswith(prefix + "/") and key != prefix
             }
-            for key in keys_unique:
-                yield key.removeprefix(prefix + "/").split("/")[0]
+
+        for key in keys_unique:
+            yield key
--- a/src/zarr/testing/store.py
+++ b/src/zarr/testing/store.py
@@ -192,6 +192,13 @@ class StoreTests(Generic[S]):
         assert [k async for k in store.list_dir("foo")] == []
         await store.set("foo/zarr.json", Buffer.from_bytes(b"bar"))
         await store.set("foo/c/1", Buffer.from_bytes(b"\x01"))
+        await store.set("foo/c/d/1", Buffer.from_bytes(b"\x01"))
+        await store.set("foo/c/d/2", Buffer.from_bytes(b"\x01"))
+        await store.set("foo/c/d/3", Buffer.from_bytes(b"\x01"))
+
+        keys_expected = ["foo"]
+        keys_observed = [k async for k in store.list_dir("")]
+        assert set(keys_observed) == set(keys_expected), keys_observed
 
         keys_expected = ["zarr.json", "c"]
         keys_observed = [k async for k in store.list_dir("foo")]
--- a/src/zarr/testing/store.py
+++ b/src/zarr/testing/store.py
@@ -198,7 +198,7 @@ class StoreTests(Generic[S]):
 
         keys_expected = ["foo"]
         keys_observed = [k async for k in store.list_dir("")]
-        assert set(keys_observed) == set(keys_expected), keys_observed
+        assert set(keys_observed) == set(keys_expected)
 
         keys_expected = ["zarr.json", "c"]
         keys_observed = [k async for k in store.list_dir("foo")]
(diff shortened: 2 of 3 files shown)


src/zarr/testing/store.py
```diff
@@ -192,6 +192,13 @@ async def test_list_dir(self, store: S) -> None:
         assert [k async for k in store.list_dir("foo")] == []
         await store.set("foo/zarr.json", Buffer.from_bytes(b"bar"))
         await store.set("foo/c/1", Buffer.from_bytes(b"\x01"))
+        await store.set("foo/c/d/1", Buffer.from_bytes(b"\x01"))
+        await store.set("foo/c/d/2", Buffer.from_bytes(b"\x01"))
+        await store.set("foo/c/d/3", Buffer.from_bytes(b"\x01"))
+
+        keys_expected = ["foo"]
```
TomAugspurger, line 199: Ah, this is failing on the S3 tests.

```pytb
        keys_expected = ["foo"]
        keys_observed = [k async for k in store.list_dir("")]
>       assert set(keys_observed) == set(keys_expected), keys_observed
E       AssertionError: ['test/foo']
```

It's including `self.path` (the bucket name?) in listing. Based on my reading of the tests, that should *not* be included, but I'm not 100% sure. 90940a0e01366a7339bf4abae5caca3fdcb73e30 has that fix

dstansby (commented):
I'm struggling to get my head around the new logic. Reading https://github.com/zarr-developers/zarr-python/issues/2116, is the goal here to just not return any duplicate values? If so, I wonder if it would be easier to keep track of the values that have been returned, and only yield a value if it hasn't been returned yet (otherwise, check the next value). This way it would also truly generate the values one at a time, instead of the new approach that computes all the keys before yielding them.

TomAugspurger:
Thanks for the review!

> is the goal here to just not return any duplicate values?

Effectively yes. Duplicate keys were a consequence of the previous implementation not handling arrays nested in child groups properly.

>  If so, I wonder if it would be easier to keep track of the values that have been returned, and only yield a value if it hasn't been returned yet (otherwise, check the next value). This way it would also truly generate the values one at a time, instead of the new approach that computes all the keys before yielding them.

3c845d93c5fd28b18867c4a0a6e6c4ffb089fc8e has an approach that's a middle ground between my previous fix and your suggestion. I think it's a lot simpler. It does still generate all the keys before yield them, but I'm not too worried about that. This is an in-memory store, so things should be pretty small and fast, and this matches the implementation of the non-prefix case.

src/zarr/testing/store.py
```diff
@@ -192,6 +192,13 @@ async def test_list_dir(self, store: S) -> None:
...
+        keys_observed = [k async for k in store.list_dir("")]
+        assert set(keys_observed) == set(keys_expected), keys_observed
```
dstansby, line 201:
```suggestion
        assert set(keys_observed) == set(keys_expected)
```

dstansby (approved):
👍 looks good to me. I think if you drop the last bit of the assert statement, `pytest` is clever and will report the difference between the two sets in it's output when the test fails.

src/zarr/store/memory.py
```diff
@@ -114,9 +114,15 @@ async def list_dir(self, prefix: str) -> AsyncGenerator[str, None]:
...
+            # a pseudo directory when there's a nested item and we're listing an
+            # intermediate level.
+            keys_unique = {
+                key.removeprefix(prefix + "/").split("/")[0]
+                for key in self._store_dict
+                if key.startswith(prefix + "/") and key != prefix
+            }
+
+        for key in keys_unique:
+            yield key
```
jhamman, line 128:
```suggestion
        yield from keys_unique
```

src/zarr/testing/store.py
```diff
@@ -192,6 +192,13 @@ async def test_list_dir(self, store: S) -> None:
         assert [k async for k in store.list_dir("foo")] == []
         await store.set("foo/zarr.json", Buffer.from_bytes(b"bar"))
         await store.set("foo/c/1", Buffer.from_bytes(b"\x01"))
+        await store.set("foo/c/d/1", Buffer.from_bytes(b"\x01"))
+        await store.set("foo/c/d/2", Buffer.from_bytes(b"\x01"))
+        await store.set("foo/c/d/3", Buffer.from_bytes(b"\x01"))
```
jhamman, line 197: I don't think its super material to the test but I believe the intent of this test was to have `foo` be an array `c` be the subdirectory containing chunks. Perhaps it would be better to change this to something more like: 
```suggestion
        await store.set("foo/bar/c/1", Buffer.from_bytes(b"\x01"))
        await store.set("foo/bar/c/2", Buffer.from_bytes(b"\x01"))
        await store.set("foo/spam/c/1", Buffer.from_bytes(b"\x01"))
        await store.set("foo/spam/c/2", Buffer.from_bytes(b"\x01"))
```

jhamman (commented):
Nice catch @TomAugspurger!

src/zarr/store/memory.py
TomAugspurger: Surprisingly(?), you can't use `yield from` in async generators: https://peps.python.org/pep-0525/#asynchronous-yield-from

jhamman: doh! I knew that.

src/zarr/testing/store.py
TomAugspurger: Right... IIUC, if you have a key at `foo/zarr.json` with a `node_type="array"`, that implies you *won't* have any keys at `foo/bar/zarr.json` (a group or a node), since arrays can't have children?

I tried to update this in 7414b1064311fd17f41a40c97fcf96b2299ac29c, to make sure that the test store we construct is at least compatible with how a Zarr hierarchy would be laid out.

jhamman: Looks great!

jhamman (approved):

PR #2123, opened 2024-08-26
fix: numpy 1.24 compat for Array.__array__

jhamman:
This PR slightly modifies #2106 to allow for any args/kwargs to be passed to `__array__`.  Without this, earlier versions of NumPy error with the following traceback:

```
    def __array__(self, dtype=None, copy=None):
>       return np.array(self[...], dtype=dtype, copy=copy)
E       ValueError: NoneType copy mode not allowed.
```

@dstansby - mind giving this a look

TODO:

### DIFF
--- a/zarr/core.py
+++ b/zarr/core.py
@@ -575,8 +575,8 @@ class Array:
             # store comparison
         )
 
-    def __array__(self, dtype=None, copy=None):
-        return np.array(self[...], dtype=dtype, copy=copy)
+    def __array__(self, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


dstansby:
Thanks and sorry for breaking this! It would be good to update one of the CI builds to run with our oldest supported version of numpy so stuff like this is caught in the future. Want to do that here, or I can do it in a follow-up PR?

jhamman:
> Want to do that here, or I can do it in a follow-up PR?

Let's do it in a separate PR.

dstansby (approved):

PR #781, opened 2021-06-17
Extract ABSStore to zarr._storage.absstore

joshmoore:
First step towards being able to assign CODEOWNERS
to simplify maintenance. This does not yet attempt
the rename to zarr.storage.absstore in order lower
conflicts with open PRs. Imports should remain un-
changed for the moment.

see: #764

PR #2126, opened 2024-08-26
Debug azure-storage-blob test failures

dstansby:
Hopefully get to the bottom of what's going on in https://github.com/zarr-developers/zarr-python/pull/2122...

I'm going to start by bisecting what version of `azure-storage-blob` first had an issue. For reference, changelog here: https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/storage/azure-storage-blob/CHANGELOG.md

The highest version that still works is 2.21.0, so I've set that as the pin here.

TODO:

### DIFF
--- a/requirements_dev_optional.txt
+++ b/requirements_dev_optional.txt
@@ -5,9 +5,7 @@ lmdb==1.5.1; sys_platform != 'win32'
 ipytree==0.2.2
 ipywidgets==8.1.3
 # optional library requirements for services
-# don't let pyup change pinning for azure-storage-blob, need to pin to older
-# version to get compatibility with azure storage emulator on appveyor (FIXME)
-azure-storage-blob==12.16.0 # pyup: ignore
+azure-storage-blob==12.21.0
 redis==5.0.8
 types-redis
 types-setuptools


requirements_dev_optional.txt
```diff
@@ -5,9 +5,7 @@ lmdb==1.5.1; sys_platform != 'win32'
 ipytree==0.2.2
 ipywidgets==8.1.3
 # optional library requirements for services
-# don't let pyup change pinning for azure-storage-blob, need to pin to older
-# version to get compatibility with azure storage emulator on appveyor (FIXME)
-azure-storage-blob==12.16.0 # pyup: ignore
+azure-storage-blob==12.21.0
```
dstansby, line 8:
```suggestion
azure-storage-blob==12.22.0
```

dstansby:
Well it's 1.22.0 that the error first shows up in. I suspect it will end up being something in this huge commit that's causing the issue... https://github.com/Azure/azure-sdk-for-python/commit/cb47de53f3716757448343a2603e2f1e3feddc6d

jhamman:
How strongly do we feel about fixing this? There is also support for azure blob storage via fsspec (https://github.com/fsspec/adlfs) so this class is now duplicative. Its also been deprecated and will be removed in 3.0. So maybe we just keep the pin in place?

joshmoore:
I could definitely live with a pin. Going back to https://github.com/zarr-developers/zarr-python/pull/781 having a group that champions these individual stores would be my overall preference (knowing I definitely don't do enough testing or have the experience needed)

dstansby:
`azure-storage-blob` isn't specified as an optional dependency anywhere, so there's nowhere to pin it apart from in the test requirements file. I've updated this PR to have the highest pin that still works. Is it worth adding it as an optional dependency to the package with an upper pin?

dstansby:
The actual version bump to 2.21.0 happened in https://github.com/zarr-developers/zarr-python/pull/2129 - I'll close this, and open an issue if anyone wants to try and debug azure-storage-blob 2.22.0...

PR #2128, opened 2024-08-27
[v3] h5py compat methods on Group

### LINKED ISSUE #1597: h5py compatibility in v3
Zarr-Python currently supports most of the h5py API in groups/arrays. As part of #1583, we've decided to break this to clean up the core API going forward.

Going forward, h5py compatibility could be provided in a set of wrapper classes.

Marking this as _help-wanted_!

jhamman:
This PR adds the h5py compat methods to the Zarr Group API from v2. 

The plan described in #1583 was to remove these but after looking into it, a bit more, I came to the conclusion that `require_group` and `require_dataset` are actually pretty useful. I'm open to adding deprecation warnings to `create_dataset` and `require_dataset` (perhaps replacing the latter with `require_array`) if we want to phase out the h5py specific language. 

closes #1597

TODO:

### DIFF
--- a/src/zarr/core/common.py
+++ b/src/zarr/core/common.py
@@ -28,6 +28,7 @@ ZGROUP_JSON = ".zgroup"
 ZATTRS_JSON = ".zattrs"
 
 BytesLike = bytes | bytearray | memoryview
+ShapeLike = tuple[int, ...] | int
 ChunkCoords = tuple[int, ...]
 ChunkCoordsLike = Iterable[int]
 ZarrFormat = Literal[2, 3]
--- a/src/zarr/core/group.py
+++ b/src/zarr/core/group.py
@@ -26,7 +26,6 @@ from zarr.core.common import (
     ZGROUP_JSON,
     ChunkCoords,
     ZarrFormat,
-    concurrent_map,
     parse_shapelike,
 )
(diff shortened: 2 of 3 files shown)


joshmoore:
In general, big :+1: for the compatibility. cc: @clbarnes of https://github.com/clbarnes/h5py_like fame in case there are any lessons learned, etc. that we should take into account.

d-v-b (approved):
> I'm open to adding deprecation warnings to create_dataset and require_dataset (perhaps replacing the latter with require_array) if we want to phase out the h5py specific language.

+1 to this, ultimately we shouldn't use the term "dataset". But for short-term hdf5 compatibility this PR looks good.

jhamman:
> In general, big 👍 for the compatibility. cc: @clbarnes of https://github.com/clbarnes/h5py_like fame in case there are any lessons learned, etc. that we should take into account.

I'm going to go ahead and merge this but happy to take feedback from @clbarnes here or in #1597.

clbarnes:
Nothing specific to say, just happy if h5py_like was of any use! Definitely agree on throwing PendingDeprecationWarnings whenever these short-term compatibility measures are used, otherwise you'll never be rid of them.

Issue #2131, opened 2024-08-28
[v3] New Mypy errors now on v3 branch


jhamman:
We've started seeing the following errors in our pre-commit action. Would be good if we can give these some attention this week.

```
src/zarr/core/metadata.py:586: error: Overloaded function signature 2 will never be matched: signature 1's parameter type(s) are the same or broader  [misc]
src/zarr/core/metadata.py:586: error: Variable "zarr.core.metadata.INTEGER_DTYPE" is not valid as a type  [valid-type]
src/zarr/core/metadata.py:586: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
src/zarr/core/metadata.py:598: error: Variable "zarr.core.metadata.INTEGER_DTYPE" is not valid as a type  [valid-type]
src/zarr/core/metadata.py:598: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
src/zarr/core/metadata.py:621: error: Item INTEGER_DTYPE? of "BoolDType | INTEGER_DTYPE? | Float16DType | Float32DType | Float64DType | Complex64DType | Complex128DType" has no attribute "type"  [union-attr]
src/zarr/core/metadata.py:637: error: Item INTEGER_DTYPE? of "BoolDType | INTEGER_DTYPE? | Float16DType | Float32DType | Float64DType | Complex64DType | Complex128DType" has no attribute "type"  [union-attr]
src/zarr/codecs/crc32c_.py:40: error: Argument 1 to "crc32c" has incompatible type "ndarray[Any, dtype[Any]]"; expected "Buffer"  [arg-type]
src/zarr/codecs/crc32c_.py:55: error: Argument 1 to "crc32c" has incompatible type "ndarray[Any, dtype[Any]]"; expected "Buffer"  [arg-type]
src/zarr/codecs/sharding.py:104: error: Returning Any from function declared to return "tuple[int, ...]"  [no-any-return]
Found 8 errors in 3 files (checked 66 source files)
```

TomAugspurger:
I think that https://github.com/numpy/numpy/pull/27211 is the root cause of `src/zarr/codecs/sharding.py:104: error: Returning Any from function declared to return "tuple[int, ...]"  [no-any-return]`.

I think we'll need to work around it for now.

TomAugspurger:
I'm seeing a slightly different error for the dtype stuff in `core.metadata`:

```
[...]

PR #2139, opened 2024-08-29
mypy fixes for NumPy 2.1.0

TomAugspurger:
This includes some fixes to get mypy type checks passing.

Closes https://github.com/zarr-developers/zarr-python/issues/2131

### DIFF
--- a/src/zarr/codecs/crc32c_.py
+++ b/src/zarr/codecs/crc32c_.py
@@ -1,9 +1,10 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, cast
 
 import numpy as np
+import typing_extensions
 from crc32c import crc32c
 
 from zarr.abc.codec import BytesBytesCodec
@@ -37,7 +38,7 @@ class Crc32cCodec(BytesBytesCodec):
         crc32_bytes = data[-4:]
         inner_bytes = data[:-4]
 
-        computed_checksum = np.uint32(crc32c(inner_bytes)).tobytes()
+        computed_checksum = np.uint32(crc32c(cast(typing_extensions.Buffer, inner_bytes))).tobytes()
         stored_checksum = bytes(crc32_bytes)
         if computed_checksum != stored_checksum:
             raise ValueError(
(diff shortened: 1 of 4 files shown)


src/zarr/core/metadata.py
```diff
@@ -602,9 +602,21 @@ def parse_fill_value_v3(fill_value: Any, dtype: FLOAT_DTYPE) -> FLOAT: ...
 def parse_fill_value_v3(fill_value: Any, dtype: COMPLEX_DTYPE) -> COMPLEX: ...
 
 
+@overload
+def parse_fill_value_v3(fill_value: Any, dtype: np.dtype[Any]) -> Any:
```
dstansby, line 606:
```suggestion
def parse_fill_value_v3(fill_value: Any, dtype: np.dtype[Any]) -> np.generic:
```
To indicate that a scalar is always returned? (like the other narrower types, which are e.g., np.uint8 which represents a single scalar I think?)

src/zarr/codecs/crc32c_.py
```diff
@@ -37,7 +38,7 @@ async def _decode_single(
         crc32_bytes = data[-4:]
         inner_bytes = data[:-4]
 
-        computed_checksum = np.uint32(crc32c(inner_bytes)).tobytes()
+        computed_checksum = np.uint32(crc32c(cast(typing_extensions.Buffer, inner_bytes))).tobytes()
```
dstansby, line 41: Isn't inner_bytes a numpy array, not a buffer here? So it needs to be converted to a buffer, not just have it's type cast? Maybe I'm misunderstanding something here...

TomAugspurger: Correct, it is an ndarray. At least as far as my copy of mypy is concerned, that's not sufficient for this type check to pass. This might have changed in newer versions of python though.

src/zarr/core/metadata.py
TomAugspurger: That passes. The revealed type of

```
reveal_type(parse_fill_value_v3(1, np.dtype("uint64")))
```

is `np.generic`, which sounds better than `Any`.

src/zarr/codecs/crc32c_.py
dstansby: Are ndarrays also Buffers in reality then?

TomAugspurger: Here's the definition of Buffer: https://github.com/python/typing_extensions/blob/d9509f902c20e3d51c62b8abe522809b4760c0ff/src/typing_extensions.py#L3316-L3342

In python 3.12, a new ABC was added, to cover the addition of a Python API to the buffer protocol: https://peps.python.org/pep-0688/

For 3.11 and earlier, ndarrays are compatible with Buffer (they implement the buffer protocol) but apparently mypy doesn't know that.

dstansby, line 41:
```suggestion
        # Need to do a manual cast until https://github.com/numpy/numpy/issues/26783 is resolved
        computed_checksum = np.uint32(crc32c(cast(typing_extensions.Buffer, inner_bytes))).tobytes()
```

dstansby (approved):

PR #2158, opened 2024-09-08
Default zarr.open to open_group if shape is not provided

### LINKED ISSUE #2143: `zarr.open` do not allow creating group file [v3]
### Zarr version

v3.0.0a2

### Numcodecs version

0.13.0

### Python Version

3.x

### Operating System

all

### Installation

virtualenv in tox

### Description

In napari project we have `--pre` test, ant they start failing. As zarr 3 requires positional only arguments in `open` I have opened a PR to prepare napari on zarr 3. however I found that when I want to create zarr group file using `grp = zarr.open(store=str(root_path), mode='a')` it fails with `TypeError: create() missing 1 required positional argument: 'shape'` https://github.com/napari/napari/actions/runs/10643480724/job/29507082501?pr=7215#step:8:728

I have checked zarr code and found:

https://github.com/zarr-developers/zarr-python/blob/0b5483a7958e2ae5512a14eb424a84b2a75dd727/src/zarr/api/asynchronous.py#L225-L228 

So it looks like older implementation was raising `KeyError`, not `TypeError` (as this lines comes from older PR than 
https://github.com/zarr-developers/zarr-python/blob/0b5483a7958e2ae5512a14eb424a84b2a75dd727/src/zarr/api/asynchronous.py#L895-L901 

### Steps to reproduce

Call `zarr.open(store=str(root_path), mode='a')` with path to a non-existing zarr file. 

### Additional output

_No response_

agoodm:
[Description of PR] Closes #2143 

This change makes the v3 `zarr.open` more consistent with v2 by defaulting to `open_group` if `shape` isn't provided, which now allows for initializing a group in a non-existent path if `mode="a"`.

TODO:


tests/v3/test_api.py
```diff
@@ -117,6 +117,10 @@ def test_open_with_mode_r_plus(tmp_path: pathlib.Path) -> None:
 
 
 def test_open_with_mode_a(tmp_path: pathlib.Path) -> None:
+    # Open without shape argument should default to group
+    g = zarr.open(store=tmp_path, mode="a")
+    assert isinstance(g, Group)
+
     # 'a' means read/write (create if doesn't exist)
     arr = zarr.open(store=tmp_path, mode="a", shape=(3, 3))
     assert isinstance(arr, Array)
```
jhamman, line 126: I'm surprised this does not fail after the previous statement. Shouldn't we be raising an error if a Group already exists at this path?

agoodm: You are right, I had some issues with my local testing config and missed this, but have fixed it and now can see the failures.

src/zarr/api/asynchronous.py
```diff
@@ -225,6 +225,11 @@ async def open(
     if path is not None:
         store_path = store_path / path
 
+    if "shape" not in kwargs and mode in {"a", "w", "w-"}:
+        try:
+            return await open_group(store=store_path, zarr_format=zarr_format, mode=mode, **kwargs)
+        except AssertionError:
```
agoodm, line 231: So one issue with v3 is that we can't trivially determine if `store_path` corresponds to a group or array by checking if it contains `.zarray` or `.zgroup` like we can with v2, so the `zarr.json` needs to be decoded one way or another. Would it be better to explicitly do this and check the JSON rather than catching an `AssertionError`?

tests/v3/test_api.py
```diff
@@ -116,7 +116,12 @@ def test_open_with_mode_r_plus(tmp_path: pathlib.Path) -> None:
     z2[:] = 3
 
 
-def test_open_with_mode_a(tmp_path: pathlib.Path) -> None:
+async def test_open_with_mode_a(tmp_path: pathlib.Path) -> None:
+    # Open without shape argument should default to group
+    g = zarr.open(store=tmp_path, mode="a")
+    assert isinstance(g, Group)
+    await g.store_path.delete()
```
agoodm, line 123: I had to convert this test to async to delete this, is there a better way here?

src/zarr/api/asynchronous.py
dstansby: I guess either way `open_group` will try and decode `zarr.json`, so opening an array with the current code will result in two attempts at decoding (once in open_group, then again in open_array)? In that case I think it's worth just checking the JSON here. A helper `is_group_or_array(store, path)` function would be nice to use here.

agoodm: How does this look? I moved the code for reading the array metadata into a separate function (`get_array_metadata`) and then used it to check if the path contains an array.

src/zarr/api/asynchronous.py
```diff
@@ -225,6 +225,21 @@ async def open(
...
+                is_array = metadata_dict.get("node_type") == "array"
+            else:
+                # for v2, the above would already have raised an exception if not an array
+                is_array = True
+        except (AssertionError, FileNotFoundError):
+            is_array = False
+        if not is_array:
+            return await open_group(store=store_path, zarr_format=zarr_format, mode=mode, **kwargs)
+        else:
+            return await open_array(store=store_path, zarr_format=zarr_format, mode=mode, **kwargs)
```
jhamman, line 242: it seems like we should be able to use the `metadata_dict` to construct the array without calling `open_array` and repeating the IO step, no?

jhamman (approved):
This looks good @agoodm -- thanks for working out the logic in a clean way.

src/zarr/core/array.py
```diff
@@ -115,10 +162,17 @@ class AsyncArray:
 
     def __init__(
         self,
-        metadata: ArrayMetadata,
+        metadata: ArrayMetadata | dict[str, Any],
         store_path: StorePath,
         order: Literal["C", "F"] | None = None,
-    ) -> None:
+    ): -> None:
```
jhamman, line 168:
```suggestion
    ) -> None:
```

PR #2163, opened 2024-09-09
refactor: split metadata into v2 and v3 modules

d-v-b:
A clean re-implementation of #2059. 
- Splits `metadata.py` into `metadata/common.py`, `metadata/v2.py`, and `metadata/v3.py`
- removes the `_v2` or `_v3` suffix that adorned quite a few parsing functions. 
- exports `ArrayV3Metadata` and `ArrayV2Metadata` from `metadata`. This might also be a good time to reconsider the names of those two classes, if anyone is bothered by them (`ArrayMetadataV2`, or `V2ArrayMetadata` could seem a bit more logical than `ArrayV2Metadata`).

TODO:


src/zarr/core/metadata/v2.py
```diff
@@ -0,0 +1,205 @@
...
+        # make a copy to protect the original from modification
+        _data = data.copy()
+        # check that the zarr_format attribute is correct
+        _ = parse_zarr_format_v2(_data.pop("zarr_format"))
+        return cls(**_data)
+
+    def to_dict(self) -> JSON:
+        zarray_dict = super().to_dict()
+
+        assert isinstance(zarray_dict, dict)
```
jhamman, line 131: is this check needed?

src/zarr/core/metadata/v2.py
```diff
@@ -0,0 +1,205 @@
...
+    ----------
+    fill_value: Any
+        A potential fill value.
+    dtype: np.dtype[Any]
+        A numpy dtype.
+
+    Returns
+        An instance of `dtype`, or `None`, or any python object (in the case of an object dtype)
+    """
+    from zarr.v2.util import normalize_fill_value
```
jhamman, line 203: eeek -- let's just copy this function over to the main namespace.

src/zarr/core/metadata/v2.py
```diff
@@ -0,0 +1,205 @@
...
+            if np.isscalar(o):
+                # convert numpy scalar to python type, and pass
+                # python types through
+                return getattr(o, "item", lambda: o)()
+            raise TypeError
+
+        zarray_dict = self.to_dict()
+        assert isinstance(zarray_dict, dict)
+        zattrs_dict = zarray_dict.pop("attributes", {})
+        assert isinstance(zattrs_dict, dict)
```
jhamman, line 109: I don't understand why we need all these `isinstance` checks? Are these to make mypy happy?

jhamman (approved):
A few nits but this is a really nice improvement. Thanks @d-v-b!

src/zarr/core/metadata/v2.py
d-v-b: I doubt it, and if we do need type guarding here we should do it formally and not with assert statements

src/zarr/core/metadata/v2.py
d-v-b: good question, as with the previous ones I will excise them and see if anything breaks

d-v-b: we need to explicitly check that the output of `to_dict` is a dict because we have mangled the return type of the `to_dict` method so that it doens't always return dicts. that's a problem we should solve in another PR.

src/zarr/core/metadata/v2.py
d-v-b: done

src/zarr/core/metadata/v2.py
d-v-b: I removed these assert statements and added explicit type checks, that we can remove when we fix the return type of `to_dict`

PR #2177, opened 2024-09-11
Bump verison of autoapi used to build docs

dstansby:
- Bumps the version of autoapi we use to build the docs
- Pins specific versions of sphinx and autoapi, which can then be updated by dependabot. I think it's best to pin the doc build to specific versions?
- Simplifies the logic of excluding private members from the docs. This is a workaround to https://github.com/readthedocs/sphinx-autoapi/issues/481, but I think worth having regardless as a simplification of `conf.py`.

On the last point, this does mean that `zarr.v2` and `zarr.core` are now in the API docs, but I figure we can think about making them private by appending a `_` to the beginning in follow up PRs where we can decided where we want the various bits of API to actually live.


jhamman:
The plan is to remove the `v2` namespace here soon so I wouldn't worry too much about that one. I'm also fine renaming `core` to `_core` if that helps here.

dstansby:
👍  - I think maybe lets get this in, then rename `core` > `_core` in another PR (it will be a big diff with all the changed imports!). Or perhaps once this is merged, we can review the public API and as part of that decide what to do with `core`.

jhamman:
@dstansby - is this ready for a merge?

dstansby:
Yep

PR #2179, opened 2024-09-12
Narrow JSON type, ensure that `to_dict` always returns a dict, and v2 filter / compressor parsing  

d-v-b:
A typing / metadata cleanup now that #2163 is in.

### narrowing the JSON type

In `v3`, the JSON type looks like this:

`JSON = None | str | int | float | Enum | dict[str, "JSON"] | list["JSON"] | tuple["JSON", ...]`

I narrowed this type to

`JSON = None | str | int | float | Mapping[str, "JSON"] | tuple["JSON", ...]`

- I don't think we need to support `list` AND `tuple`; as `tuple` is immutable, I got rid of the `list` variant. 
- I switched `dict[str, JSON]` to `Mapping[str, JSON]` out of the same preference for immutability.
- Allowing `Enum` is a back door around type safety, so I removed it; instead of serializing enums, the relevant classes serialize the enum values. Maybe we could put `Enum` back in `JSON` if there's a way to tell the type checker that our enums only wrap JSON values. 

### `to_dict` methods
Along with narrowing the JSON type, I ensured that all instances of `Metadata` return `dict[str, JSON]` from their `to_dict` methods. 

### type hints

There were a lot of functions in `metadata` that accepted `Any`, when `object` is a better type hint. I adjusted these accordingly.

### v2 filters and codecs

We weren't doing any validation of filters and compressors for v2 metadata. This PR adds that missing validation, and I found a buggy test as a result of adding these checks. I also changed `ArrayV2Metadata.filters` and `ArrayV2Metadata.compressor` to be instances of `numcodecs.abc.Codec` instead of dicts, bringing `ArrayV2Metadata` closer to `ArrayV3Metadata` and simplifying some codec parsing code.

TODO:

### DIFF
--- a/src/zarr/codecs/pipeline.py
+++ b/src/zarr/codecs/pipeline.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-from collections.abc import Iterable, Iterator
 from dataclasses import dataclass
 from itertools import islice, pairwise
 from typing import TYPE_CHECKING, Any, TypeVar
@@ -15,12 +14,14 @@ from zarr.abc.codec import (
     Codec,
     CodecPipeline,
 )
-from zarr.core.common import JSON, ChunkCoords, concurrent_map, parse_named_configuration
+from zarr.core.common import ChunkCoords, concurrent_map
 from zarr.core.config import config
 from zarr.core.indexing import SelectorTuple, is_scalar, is_total_slice
-from zarr.registry import get_codec_class, register_pipeline
+from zarr.registry import register_pipeline
 
 if TYPE_CHECKING:
+    from collections.abc import Iterable, Iterator
+
     import numpy as np
     from typing_extensions import Self
 
@@ -67,30 +68,6 @@ class BatchedCodecPipeline(CodecPipeline):
     array_bytes_codec: ArrayBytesCodec
     bytes_bytes_codecs: tuple[BytesBytesCodec, ...]
     batch_size: int
-
-    @classmethod
-    def from_dict(cls, data: Iterable[JSON | Codec], *, batch_size: int | None = None) -> Self:
-        out: list[Codec] = []
-        if not isinstance(data, Iterable):
-            raise TypeError(f"Expected iterable, got {type(data)}")
-
-        for c in data:
-            if isinstance(
-                c, ArrayArrayCodec | ArrayBytesCodec | BytesBytesCodec
-            ):  # Can't use Codec here because of mypy limitation
-                out.append(c)
-            else:
-                name_parsed, _ = parse_named_configuration(c, require_configuration=False)
-                out.append(get_codec_class(name_parsed).from_dict(c))  # type: ignore[arg-type]
-        return cls.from_list(out, batch_size=batch_size)
-
-    def to_dict(self) -> dict[str, JSON]:
-        return {
-            "array_array_codecs": tuple(c.to_dict() for c in self.array_array_codecs),
-            "array_bytes_codec": self.array_bytes_codec.to_dict(),
-            "bytes_bytes_codec": tuple(c.to_dict() for c in self.bytes_bytes_codecs),
-            "batch_size": self.batch_size,
-        }
 
     def evolve_from_array_spec(self, array_spec: ArraySpec) -> Self:
         return type(self).from_list([c.evolve_from_array_spec(array_spec=array_spec) for c in self])
--- a/src/zarr/codecs/pipeline.py
+++ b/src/zarr/codecs/pipeline.py
@@ -70,10 +70,10 @@ class BatchedCodecPipeline(CodecPipeline):
     batch_size: int
 
     def evolve_from_array_spec(self, array_spec: ArraySpec) -> Self:
-        return type(self).from_list([c.evolve_from_array_spec(array_spec=array_spec) for c in self])
+        return type(self).from_codecs(c.evolve_from_array_spec(array_spec=array_spec) for c in self)
 
     @classmethod
-    def from_list(cls, codecs: Iterable[Codec], *, batch_size: int | None = None) -> Self:
+    def from_codecs(cls, codecs: Iterable[Codec], *, batch_size: int | None = None) -> Self:
         array_array_codecs, array_bytes_codec, bytes_bytes_codecs = codecs_from_list(codecs)
 
         return cls(
--- a/src/zarr/codecs/pipeline.py
+++ b/src/zarr/codecs/pipeline.py
@@ -84,8 +84,13 @@ class BatchedCodecPipeline(CodecPipeline):
                 out.append(get_codec_class(name_parsed).from_dict(c))  # type: ignore[arg-type]
         return cls.from_list(out, batch_size=batch_size)
 
-    def to_dict(self) -> JSON:
-        return [c.to_dict() for c in self]
+    def to_dict(self) -> dict[str, JSON]:
+        return {
+            "array_array_codecs": tuple(c.to_dict() for c in self.array_array_codecs),
+            "array_bytes_codec": self.array_bytes_codec.to_dict(),
+            "bytes_bytes_codec": tuple(c.to_dict() for c in self.bytes_bytes_codecs),
+            "batch_size": self.batch_size,
+        }
 
     def evolve_from_array_spec(self, array_spec: ArraySpec) -> Self:
         return type(self).from_list([c.evolve_from_array_spec(array_spec=array_spec) for c in self])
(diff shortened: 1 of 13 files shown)


dstansby:
> I don't think we need to support list AND tuple; as tuple is immutable, I got rid of the list variant.

Given dictionaries are mutable, is there another reason not to support lists? If you want to keep the typing simple, perhaps  `typing.Sequence[JSON]` could be used?

src/zarr/codecs/pipeline.py
```diff
@@ -84,8 +84,13 @@ def from_dict(cls, data: Iterable[JSON | Codec], *, batch_size: int | None = Non
                 out.append(get_codec_class(name_parsed).from_dict(c))  # type: ignore[arg-type]
         return cls.from_list(out, batch_size=batch_size)
 
-    def to_dict(self) -> JSON:
-        return [c.to_dict() for c in self]
+    def to_dict(self) -> dict[str, JSON]:
+        return {
```
dstansby, line 88: Would it not be better to do `{'codecs': [c.to_dict() for c in self]}` here? It's a bit simpler, and matches the way the sharding codec stores multiple codecs.

src/zarr/core/common.py
```diff
@@ -80,7 +80,7 @@ def enum_names(enum: type[E]) -> Iterator[str]:
         yield item.name
 
 
-def parse_enum(data: JSON, cls: type[E]) -> E:
+def parse_enum(data: object, cls: type[E]) -> E:
```
dstansby, line 83:
```suggestion
def parse_enum(data: E | str, cls: type[E]) -> E:
```

src/zarr/core/metadata/v2.py
```diff
@@ -165,18 +157,44 @@ def update_attributes(self, attributes: dict[str, JSON]) -> Self:
         return replace(self, attributes=attributes)
 
 
-def parse_zarr_format(data: Literal[2]) -> Literal[2]:
+def parse_zarr_format(data: object) -> Literal[2]:
```
dstansby, line 160: I think this should stay as `Literal[2]`? Even though it's possible to input any type here, giving it a narrow type allow errors to be spotted at type checking time, ie before runtime.

src/zarr/core/metadata/v2.py
```diff
@@ -165,18 +157,44 @@ def update_attributes(self, attributes: dict[str, JSON]) -> Self:
...
     if data == 2:
-        return data
+        return 2
     raise ValueError(f"Invalid value. Expected 2. Got {data}.")
 
 
-def parse_filters(data: list[dict[str, JSON]] | None) -> list[dict[str, JSON]] | None:
-    return data
+def parse_filters(data: object) -> tuple[numcodecs.abc.Codec, ...] | None:
```
dstansby, line 166:
```suggestion
def parse_filters(data: None, Iterable[numcodecs.abc.Codec, dict]) -> tuple[numcodecs.abc.Codec, ...] | None:
```

dstansby (commented):
Looks good overall to me - I'm not sure about widening the input types on the parsing functions though, as then the benefit of typing is lost. What was the motivation for setting these all to `object` instead of the narrower types that the functions are designed to parse?

d-v-b:
> Looks good overall to me - I'm not sure about widening the input types on the parsing functions though, as then the benefit of typing is lost. What was the motivation for setting these all to `object` instead of the narrower types that the functions are designed to parse?

the current model is that the parsing functions are designed to handle entirely unknown objects and parse those objects into something with the correct value and type. So these functions should accept any python object, but only return a value when parsing is successful, and otherwise raise an exception. From that POV, placing any constraints on the input type is stealing work from the body of the parsing function :)

src/zarr/core/metadata/v2.py
d-v-b: `parse_zarr_format` is designed for use with unknown data. If we annotate the input type as `Literal[2]`, then mypy will complain if you try to call this function on unknown data, and we have defeated the point of the function. 

If you already know that a variable has type `Literal[2]`, then there is no need to call `parse_zarr_format`. I do think we should find a way to construct metadata documents from pre-validated data, but that's out of scope for this function.

src/zarr/core/metadata/v2.py
d-v-b: similar answer to https://github.com/zarr-developers/zarr-python/pull/2179#discussion_r1758393646: these parsing functions take arbitrary python objects by design. type checking errors for the inputs to these functions don't help with that.

dstansby:
I'm not sure I understand what the downside is of "stealing work from the function"? The advantage of including the narrower types is finding errors before running code. That's kind of the point of typing - you *can* pass whatever you want in Python, but by adding type hints it's easier to see where you're passing the wrong objects.

dstansby:
And where these parsing functions are used, they are called with objects that already have a narrow type, e.g., here `cname` already has a narrow type:
https://github.com/zarr-developers/zarr-python/blob/52d68490b308ad861ed9d128b320628ce690afea/src/zarr/codecs/blosc.py#L95-L105

So given these aren't being used with objects that have a general type, I think it is better to give the parsing functions narrower types as input, to make sure we're defining the narrow types consistently throughout the codebase?

d-v-b:
> I'm not sure I understand what the downside is of "stealing work from the function"? The advantage of including the narrower types is finding errors before running code. That's kind of the point of typing - you _can_ pass whatever you want in Python, but by adding type hints it's easier to see where you're passing the wrong objects.

For context, these parsing functions are designed for handling the contents of JSON documents that we are reading from external sources. We have _no idea_ what is inside those JSON documents. We can be pretty confident about that the output of python's JSON deserialization will give us dicts, lists, etc, None, etc but the particular structure is unknown. 

The job of the `parse_x` functions is to take such data, about which we know basically nothing, parse it into something correct (i.e., an instance of the return type, with possible value checks) or raise an exception. If `parse_x` is opinionated about the type of it's input, then we need to check the input before feeding it to `parse_x`. But checking the input is exactly the job of `parse_x`! That's what I mean by "stealing work". 

Suppose we do `parse_zarr_format(data: Literal[2]) -> Literal[2]`. This is the signature of a meaningless function, because if we know that the data is `2` then we don't need any parsing. But if we tried to use this function to deal with unknown data, where would our assurance that `data == 2` come from? We would need a function to take that unknown data and check that it's 2, which is what `parse_zarr_format` is for in the first place.

dstansby:
Ah that makes sense, thanks for the patient explanation 👍 I didn't realise these were used to parse arbitrary JSON.

d-v-b:
> Ah that makes sense, thanks for the patient explanation 👍 I didn't realise these were used to parse arbitrary JSON.

no worries! thanks for looking things over carefully.

src/zarr/codecs/pipeline.py
d-v-b: My thinking here was that the dict returned by `to_dict` should match the structure of the class itself, which has attributes like `array_array_codecs`, `array_bytes_codecs`, etc. But your comment made me notice two things:
- I didn't update the `from_dict` method, which assumes the old list representation
- The above mistake did not cause any tests to fail.

So I wonder if there is any need for `CodecPipeline` to implement `to_dict` and `from_dict` at all. [Despite my best efforts](https://github.com/zarr-developers/zarr-specs/issues/298), this data structure is not part of the zarr metadata, and we clearly are not using these serialization methods. I will experiment with removing inheritance from `Metadata` and see if anything breaks.

d-v-b:
continuing [this conversation](https://github.com/zarr-developers/zarr-python/pull/2179#discussion_r1758368047) in the main thread:

I removed the inheritance from `Metadata`, as well as the `to_dict` and `from_dict` methods for the codec pipeline classes, and everything worked :shrug: so those methods were not load bearing. I also renamed the `CodecPipeline.from_list` method to `from_codecs`, because a) it doesn't just take `list` (it takes `iterable`), and b) it's more informative to name the method after the items in the collection, rather than the collection itself.

jhamman (approved):
👏 @d-v-b - fantastic set of simplifications here!

TomAugspurger:
This seems to be causing `tests/v3/test_array.py::test_serializable_sync_array` to fail. I'm taking a look.

src/zarr/core/array.py
```diff
@@ -315,15 +313,9 @@ async def _create_v2(
             chunks=chunks,
             order=order,
             dimension_separator=dimension_separator,
-            fill_value=0 if fill_value is None else fill_value,
```
TomAugspurger, line 318: Restoring the `0 if fill_value is None else fill_value` does fix the failure in tests/v3/test_array.py::test_serializable_sync_array.

But that said, I like the change here... Setting a default fill value *here* seems wrong.

TomAugspurger: Oh, but this is just for `v2` metadata, which is allowed to be None. Maybe this isn't the wrong spot to do that.

PR #2180, opened 2024-09-12
feat: metadata-only support for storage transformers metadata

### LINKED ISSUE #2178: `storage_transformers` are not an accepted keyword for ArrayV3Metadata
### Zarr version

3.0.0a3

### Numcodecs version

0.13.0

### Python Version

3.11.9

### Operating System

Mac

### Installation

pip install

### Description

The zarr v3 Array metadata spec specfies the `storage_transformers` key can exist, but zarr crashes when reading a buffer that includes the `storage_transformers` key, even as an empty list: 

```TypeError: ArrayV3Metadata.__init__() got an unexpected keyword argument 'storage_transformers'```

### Steps to reproduce

```python
import json
import zarr

metadata = json.loads("""{
  "zarr_format": 3,
  "node_type": "array",
  "shape": [
    10
  ],
  "data_type": "int64",
  "chunk_grid": {
    "name": "regular",
    "configuration": {
      "chunk_shape": [
        10
      ]
    }
  },
  "chunk_key_encoding": {
    "name": "default",
    "configuration": {
      "separator": "/"
    }
  },
  "fill_value": 0,
  "codecs": [
    {
      "name": "bytes",
      "configuration": {
        "endian": "little"
      }
    }
  ],
  "storage_transformers": null,
  "dimension_names": null,
  "attributes": {
    "foo": 101
  }
}""")

metadata = zarr.core.array.parse_array_metadata(metadata)
```

### Additional output

_No response_

d-v-b:
Addresses #2178

This adds metadata-only support for the `storage_transformers` key in array metadata. By "metadata only" I mean that creating an array metadata document from JSON or a dict with a `storage_transformers` keyword will not error, and some very mild validation will be run (just ensuring that the value is an iterable or None), but the value is not used for any storage transforming, because we don't have any code for that yet. 

But at least the metadata can be constructed and the `storage_transformer` value should round-trip through metadata properly. This should resolve #2178. 

TODO:


zoj613:
If creating a metadata object from a JSON file that specifies a storage transformer list, then this implies reading from or writing to that array hinges on applying the transformer pipeline since it may modify the keys and/or values of that array node. I think ignoring it and issuing a warning is not enough since reading from said array using the metadata would potentially result in the wrong byte sequence being passed to its codec pipeline (as a result of skipping the transformation pipeline).

To me this appears as a programming error. I think parsing a metadata document with a specified `storage_transformers` field should raise an exception to indicate that `zarr-python` cannot reliably read such an array node since it was clearly created by an implementation that used a storage transformer pipeline.

The case where a `storage_transformer` field could be safely ignored when parsing a JSON is when it is an empty list, as [indicated by the spec](https://zarr-specs.readthedocs.io/en/latest/v3/core/v3.0.html#id13)

d-v-b:
> To me this appears as a programming error. I think parsing a metadata document with a specified `storage_transformers` field should raise an exception to indicate that `zarr-python` cannot reliably read such an array node since it was clearly created by an implementation that used a storage transformer pipeline.

I think this is valid, but we should raise when constructing an `Array` from metadata that references a storage transformer, rather than when constructing the array metadata itself. I will implement this.

src/zarr/core/metadata/v3.py
```diff
@@ -70,6 +71,27 @@ def parse_dimension_names(data: None | Iterable[str | None]) -> tuple[str | None
...
+        return ()
+    if isinstance(data, Iterable):
+        if len(tuple(data)) >= 1:
+            msg = (
+                "Got a non-null storage_transformers keyword argument. "
+                "Storage transformers are not supported by zarr-python at this time. "
+                "The storage transformer(s) will be retained in array metadata but will not "
+                "influence storage routines"
+            )
+            warnings.warn(msg, UserWarning, stacklevel=1)
```
jhamman, line 85: after thinking about this some more, I believe we should raise an error here instead of warning. 

We allow:
- metadata w/o the storage_transformers key
- metadata with `storage_transformers == [] or None`

We error for:
- metadata w/ `len(storage_transformers) > 0`

d-v-b: why not error when a user tries to create an array with the metadata? I'm thinking about how someone would develop a storage transformer with zarr-python as a dependency. If the metadata parsing functionality supports storage transformers, then they can use that immediately, and then subclass / implement their own array class that can use the storage transformer. Seems like a better workflow than forcing them to also subclass the metadata class?

d-v-b:
The changes in this PR now result in errors if an `AsyncArray` is created from `ArrayV3Metadata` that contains a non-zero number of storage transformers. However, you can create metadata with storage transformers without any error or warnings.

jhamman (approved):
Thanks @d-v-b

PR #2186, opened 2024-09-14
[v3] fix: zarr v2 compatibility fixes for Dask

jhamman:
- port `normalize_chunks` from v2
- add `array.store` property
- default to append in `create`
- rename `zarr.store` to `zarr.storage`

closes https://github.com/zarr-developers/zarr-python/issues/1953
xref: https://github.com/dask/dask/pull/11388

TODO:
* [ ] Add unit tests and/or doctests in docstrings
* [ ] Add docstrings and API docs for any new/modified user-facing classes and functions
* [ ] New/modified features documented in docs/tutorial.rst
* [ ] Changes documented in docs/release.rst
* [ ] GitHub Actions have all passed
* [ ] Test coverage is 100% (Codecov passes)

PR #2196, opened 2024-09-17
Deprecate zarr.storage

TomAugspurger:
This deprecates the usage of `zarr.storage`. For classes where we have a (roughly) equivalent `Store` in v3, a warning is emitted and the class from the new location is returned:

```python
In [1]: from zarr.storage import ZipStore
<ipython-input-1-f60ad0b873ae>:1: FutureWarning: 'zarr.storage' is deprecated. Import from 'zarr.store' instead.
```

### DIFF
--- a/src/zarr/storage.py
+++ b/src/zarr/storage.py
@@ -1,0 +1,40 @@
+import importlib
+import warnings
+from typing import Any
+
+_names = [
+    "DictStore",
+    "KVStore",
+    "DirectoryStore",
+    "ZipStore",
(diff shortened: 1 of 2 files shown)


jhamman:
@TomAugspurger - this is very cool. However in #2186, I've renamed `zarr.store -> storage`. I honestly don't think we had good reason to use a different module name.

TomAugspurger:
Even better :) I was wondering that myself but don't want to rock the boat *too* much.

PR #2198, opened 2024-09-17
Default to RemoteStore for fsspec URIs

### LINKED ISSUE #2195: Default to a RemoteStore for (some) string `StoreLike`
### Zarr version

v3

### Numcodecs version

n/a

### Python Version

n/a

### Operating System

n/a

### Installation

n/a

### Description

Zarr v2 users (either direct, or indirect through libraries like xarray) probably relied on the ability to pass an fsspec URI to `zarr.open` and have it work with a remote file system. Is there any tolerance for trying to interpret some string `store` arguments as an argument to `RemoteStore` rather than `LocalFile`? For example:

```python
zarr.open_group("s3://hrrrzarr")
```

would be interpreted as

```python
zarr.open_group(zarr.store.RemoteStore("s3://hrrrzarr"))
```

We'd also need to expose `storage_options` in all the API endpoints that create a `Store` / `StorePath` from user input, to allow customizing the filesystem that's created.

### Steps to reproduce

```python
import zarr

g = zarr.open_group(store="s3://hrrrzarr", storage_options={"anon": False})
```

### Additional output

_No response_

TomAugspurger:
This allows for specifying a RemoteStore by an fsspec URI + storage options, as a convenience (and backwards compatibility shim) for creating a `RemoteStore`:

```python

In [1]: import zarr.store

In [2]: g = zarr.open_group(store="az://daymet-zarr/annual/hi.zarr", zarr_format=2, storage_options={"account_name": "daymeteuwest"})

In [3]: list(g.arrays())
Object at .zmetadata is not recognized as a component of a Zarr hierarchy.
Out[3]: 
[<Array az://daymet-zarr/annual/hi.zarr/lambert_conformal_conic shape=() dtype=int16>,
 <Array az://daymet-zarr/annual/hi.zarr/lat shape=(584, 284) dtype=float32>,
...
```

Closes #2195

### DIFF
--- a/src/zarr/api/asynchronous.py
+++ b/src/zarr/api/asynchronous.py
@@ -194,6 +194,7 @@ async def open(
     zarr_version: ZarrFormat | None = None,  # deprecated
     zarr_format: ZarrFormat | None = None,
     path: str | None = None,
+    storage_options: dict[str, Any] | None = None,
     **kwargs: Any,  # TODO: type kwargs as valid args to open_array
 ) -> AsyncArray | AsyncGroup:
     """Convenience function to open a group or array using file-mode-like semantics.
@@ -211,6 +212,9 @@ async def open(
         The zarr format to use when saving.
     path : str or None, optional
         The path within the store to open.
+    storage_options : dict
+        If using an fsspec URL to create the store, these will be passed to
+        the backend implementation. Ignored otherwise.
     **kwargs
         Additional parameters are passed through to :func:`zarr.creation.open_array` or
         :func:`zarr.hierarchy.open_group`.
@@ -221,7 +225,7 @@ async def open(
         Return type depends on what exists in the given store.
     """
     zarr_format = _handle_zarr_version_or_format(zarr_version=zarr_version, zarr_format=zarr_format)
-    store_path = await make_store_path(store, mode=mode)
+    store_path = await make_store_path(store, mode=mode, storage_options=storage_options)
 
     if path is not None:
         store_path = store_path / path
@@ -276,6 +280,7 @@ async def save_array(
     zarr_version: ZarrFormat | None = None,  # deprecated
     zarr_format: ZarrFormat | None = None,
     path: str | None = None,
+    storage_options: dict[str, Any] | None = None,
     **kwargs: Any,  # TODO: type kwargs as valid args to create
 ) -> None:
     """Convenience function to save a NumPy array to the local file system, following a
@@ -291,6 +296,9 @@ async def save_array(
         The zarr format to use when saving.
     path : str or None, optional
         The path within the store where the array will be saved.
+    storage_options : dict
+        If using an fsspec URL to create the store, these will be passed to
+        the backend implementation. Ignored otherwise.
     kwargs
         Passed through to :func:`create`, e.g., compressor.
     """
@@ -299,7 +307,7 @@ async def save_array(
         or _default_zarr_version()
     )
 
-    store_path = await make_store_path(store, mode="w")
+    store_path = await make_store_path(store, mode="w", storage_options=storage_options)
     if path is not None:
         store_path = store_path / path
     new = await AsyncArray.create(
@@ -319,6 +327,7 @@ async def save_group(
     zarr_version: ZarrFormat | None = None,  # deprecated
     zarr_format: ZarrFormat | None = None,
     path: str | None = None,
+    storage_options: dict[str, Any] | None = None,
     **kwargs: NDArrayLike,
 ) -> None:
     """Convenience function to save several NumPy arrays to the local file system, following a
@@ -334,11 +343,17 @@ async def save_group(
         The zarr format to use when saving.
     path : str or None, optional
         Path within the store where the group will be saved.
+    storage_options : dict
+        If using an fsspec URL to create the store, these will be passed to
+        the backend implementation. Ignored otherwise.
     kwargs
         NumPy arrays with data to save.
     """
     zarr_format = (
-        _handle_zarr_version_or_format(zarr_version=zarr_version, zarr_format=zarr_format)
+        _handle_zarr_version_or_format(
+            zarr_version=zarr_version,
+            zarr_format=zarr_format,
+        )
         or _default_zarr_version()
     )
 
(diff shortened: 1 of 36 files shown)


src/zarr/store/common.py
```diff
@@ -75,30 +80,50 @@ def __eq__(self, other: Any) -> bool:
...
             mode = "w"  # exception to the default mode = 'r'
-        return StorePath(await MemoryStore.open(mode=mode))
+        result = StorePath(await MemoryStore.open(mode=mode))
     elif isinstance(store_like, Path):
-        return StorePath(await LocalStore.open(root=store_like, mode=mode or "r"))
+        result = StorePath(await LocalStore.open(root=store_like, mode=mode or "r"))
     elif isinstance(store_like, str):
-        return StorePath(await LocalStore.open(root=Path(store_like), mode=mode or "r"))
+        try:
+            fs, path = fsspec.url_to_fs(store_like, **storage_options)
```
TomAugspurger, line 108: Nitpick: this constructs the actual filesystem class instance, but we have to go back to a URL (and reconstruct the filesystem instance) for the `RemoteStore` interface. It'd be nice to avoid that.

src/zarr/store/common.py
```diff
@@ -75,30 +80,50 @@ def __eq__(self, other: Any) -> bool:
...
+        except Exception:
+            # not sure what to do here, but I don't want this to fail...
```
TomAugspurger, line 110: This is bad, but in my experience `fsspec.url_to_fs` can run pretty much arbitrary code, depending on what packages you have installed and what URI you pass. I really don't want this to fail and cause issues for people who just want a local path. *maybe* we'd always get an fsspec LocalFileSystem for a plain, non-fsspec URI string, but I'm not sure.

src/zarr/store/common.py
```diff
@@ -75,30 +80,50 @@ def __eq__(self, other: Any) -> bool:
...
     elif isinstance(store_like, dict):
         # We deliberate only consider dict[str, Buffer] here, and not arbitrary mutable mappings.
         # By only allowing dictionaries, which are in-memory, we know that MemoryStore appropriate.
-        return StorePath(await MemoryStore.open(store_dict=store_like, mode=mode))
-    raise TypeError
+        result = StorePath(await MemoryStore.open(store_dict=store_like, mode=mode))
+    else:
+        raise TypeError
+
+    if path is not None:
```
TomAugspurger, line 124: Oops, I forgot to revert this. It's not used yet.

We call `make_store_path` multiple times in a single call by the user (e.g. `save_array` calls `make_store_path` and passes the result to `AsyncArray.create` which also calls `make_store_path`). I was thinking about consolidating all the store / path stuff in a single spot, but it ended up being a bit much and I wasn't sure what the tolerance was for expanding the API of methods like `AsyncArray.create` even more. Do we want users to do

```python
AsyncArray.create(store=StorePath(RemoteStore(..., **storage_options), path))
```

or

```python
AsyncArray.create(store="s3://bucket/path.zarr", storage_options=...)
```

src/zarr/api/asynchronous.py
```diff
@@ -194,6 +194,7 @@ async def open(
     zarr_version: ZarrFormat | None = None,  # deprecated
     zarr_format: ZarrFormat | None = None,
     path: str | None = None,
+    storage_options: dict[str, Any] | None = None,
```
dstansby, line 197: Worth adding a warning if `storage_opitions is not None`, but an fsspec URL isn't passed (ie. they aren't used)?

dstansby (commented):
I don't have much opinion on the actual change, but is it worth adding warnings everywhere if `storage_options` is passed by the user (ie not None), but they haven't passed an fsspec URL so they aren't used?

jhamman (approved):

src/zarr/store/common.py
```diff
@@ -75,30 +80,44 @@ def __eq__(self, other: Any) -> bool:
...
             mode = "w"  # exception to the default mode = 'r'
-        return StorePath(await MemoryStore.open(mode=mode))
+        result = StorePath(await MemoryStore.open(mode=mode))
     elif isinstance(store_like, Path):
-        return StorePath(await LocalStore.open(root=store_like, mode=mode or "r"))
+        result = StorePath(await LocalStore.open(root=store_like, mode=mode or "r"))
     elif isinstance(store_like, str):
-        return StorePath(await LocalStore.open(root=Path(store_like), mode=mode or "r"))
+        storage_options = storage_options or {}
+        fs, _ = fsspec.url_to_fs(store_like, **storage_options)
```
jhamman, line 107: A few thoughts here: 

1. It would be nice to avoid creating the `FileSystem` if possible. A lighter approach could attempt to parse the string and look for `://` (and not `file://`).
2. If we do have to create filesystem, it would be nice to reuse it when creating the `RemoteStore`. Something like `RemoteStore.from_filesystem(...)` could be a nice pattern.

jhamman: See below for the way we used to do this. May be worth replicating parts of this

https://github.com/zarr-developers/zarr-python/blob/e1d98cdf3d35e4cb2815912d82a64c1623347c87/zarr/storage.py#L170

src/zarr/store/common.py
```diff
@@ -75,30 +80,44 @@ def __eq__(self, other: Any) -> bool:
...
-        return StorePath(store_like)
+        result = StorePath(store_like)
     elif store_like is None:
         if mode is None:
             mode = "w"  # exception to the default mode = 'r'
-        return StorePath(await MemoryStore.open(mode=mode))
+        result = StorePath(await MemoryStore.open(mode=mode))
     elif isinstance(store_like, Path):
-        return StorePath(await LocalStore.open(root=store_like, mode=mode or "r"))
+        result = StorePath(await LocalStore.open(root=store_like, mode=mode or "r"))
```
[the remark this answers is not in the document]
TomAugspurger: could be missing something, but I don't think that'll work. `LocalStore.open` will call `LocalStore.__init__`, which just takes `root` and `mode`, which are passed as regular args here.

src/zarr/api/asynchronous.py
TomAugspurger: IMO, an exception in this case is probably best. I'll do that.

src/zarr/store/common.py
TomAugspurger: Just pushed that refactor. It's a bit bigger than I'd like, but the basic version is that `RemoteStores`'s `__init__` matches what it uses internally: you can only given it an async filesystem.

There's a `from_url` and `from_path` that handles the other stuff that was previously in `__init__`.

I removed `RemoteStore._url`. I *think* that that's safe to remove, but I need to actually think through how it interacted with `store.path`

src/zarr/store/common.py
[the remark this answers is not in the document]
jhamman: I see what you mean. I was thinking `auto_mkdir` would be passed through but if that's not the case, let's not get distracted here.

jhamman (approved):

PR #2189, opened 2024-09-16
Stateful tests for array/group manipulation

dcherian:
Port over some stateful tests from [Arraylake](https://docs.earthmover.io/).

[Description of PR]
- [x] blocked by https://github.com/zarr-developers/zarr-python/issues/2191
- [x] check for https://github.com/zarr-developers/zarr-python/issues/2228
- [x] check for https://github.com/zarr-developers/zarr-python/pull/2262

TODO:
* [ ] Add unit tests and/or doctests in docstrings
* [ ] Add docstrings and API docs for any new/modified user-facing classes and functions
* [ ] New/modified features documented in docs/tutorial.rst
* [ ] Changes documented in docs/release.rst
* [ ] GitHub Actions have all passed
* [ ] Test coverage is 100% (Codecov passes)

PR #2200, opened 2024-09-17
Fix fill_value handling for complex dtypes

dcherian:
Broken out of #2189 

1. Property test now tests arbitrary fill_values of the right dtype, as well as None
2. All handle v2, v3 dtypes

TODO:


src/zarr/core/metadata/v3.py
```diff
@@ -304,17 +314,17 @@ def parse_fill_value(
     ----------
     fill_value: Any
         A potential fill value.
-    dtype: BOOL_DTYPE | INTEGER_DTYPE | FLOAT_DTYPE | COMPLEX_DTYPE
+    dtype: BOOL_DTYPE | INTEGER_DTYPE | FLOAT_DTYPE | COMPLEX_DTYPE | DATETIME_DTYPE
```
jhamman, line 317: I'm not sure we can/should support the datetime type in the v3 metadata yet. This is intended to be an extension type (https://github.com/zarr-developers/zarr-specs/issues/110, https://github.com/zarr-developers/zarr-specs/issues/273) but has not been finalized (or even proposed in sufficient detail). 

@d-v-b / @normanrz / others -- what do we want to do here? This is an example of a regression from v2 that we could handle in a few ways. I think the best thing is to try to replicate support for datetime types in v2 metadata but raise an error if a user creates this type in v3 for now.

dcherian:
FWIW the zarr.json is
```
{
  "shape": [
    1
  ],
  "fill_value": 0,
  "chunk_grid": {
    "name": "regular",
    "configuration": {
      "chunk_shape": [
        1
      ]
    }
  },
  "attributes": {},
  "zarr_format": 3,
  "data_type": "datetime64[Y]",
  "chunk_key_encoding": {
    "name": "default",
    "configuration": {
      "separator": "/"
    }
  },
  "codecs": [
    {
      "name": "bytes",
      "configuration": {
        "endian": "little"
      }
    }
  ],
  "node_type": "array"
}
```

src/zarr/core/metadata/v3.py
jhamman: I spoke with @d-v-b about this today. The action item for the `datetime` dtype is to error/warn in `parse_dtype` is not supported by the spec.

d-v-b: +1 to warning / erroring on dtypes that are not defined in the spec. 

We should 100% aim to support datetime types (because people use them), and so I think we should look at what language we can change in the spec to make this viable. I think the simplest approach would be twofold:
- introduce language in the spec distinguishing between core dtypes and community dtypes. core dtypes are part of the core spec, community dtypes come from anyone. the datetime dtype would be a community dtype.
- write up a simple spec for the datetime dtype sufficient for non-python implementation to implement it. submit that document to the collection of community dtypes. 

with that in place, i think we can safely support the dtype with minimal risk of friction across zarr implementations

d-v-b: > I think datetime is a particular complex dtype, because there are many different ways to encode datetime in binary form (e.g. precision, timezones, reference date). We should aim on defining an dtype encoding that is either very explicit in its name or has broad community support.

Definitely. I'm hoping that whatever was the norm in zarr v2 can be documented quickly enough to make the transition smooth, but longer term this particular dtype might not be a winner for cross-platform support.

jhamman: see https://github.com/zarr-developers/zarr-python/pull/2200

src/zarr/core/buffer/core.py
```diff
@@ -464,9 +464,14 @@ def __repr__(self) -> str:
...
     def all_equal(self, other: Any, equal_nan: bool = True) -> bool:
         """Compare to `other` using np.array_equal."""
+        if other is None:
+            # Handle None fill_value for Zarr V2
+            return False
         # use array_equal to obtain equal_nan=True functionality
         data, other = np.broadcast_arrays(self._data, other)
-        result = np.array_equal(self._data, other, equal_nan=equal_nan)
+        result = np.array_equal(
+            self._data, other, equal_nan=equal_nan if self._data.dtype.kind not in "US" else False
```
dcherian, line 473: also needed for V2

src/zarr/core/array.py
```diff
@@ -252,12 +252,6 @@ async def _create_v3(
         shape = parse_shapelike(shape)
         codecs = list(codecs) if codecs is not None else [BytesCodec()]
 
-        if fill_value is None:
```
normanrz, line 255: I don't understand this. What happens if I don't supply a fill_value for a uint8 array?

dcherian: for v3, that gets handled in `parse_fill_value`. That's why I removed it, didn't seem right to set a default in two places.

normanrz: Cool. Thanks for explaining!

dcherian:
pre-commit.ci autofix

jhamman (approved):

PR #2202, opened 2024-09-18
Generalize stateful store test

dcherian:
[Description of PR]

TODO:

cc @e-marshall


tests/v3/test_store/test_stateful_store.py
```diff
@@ -216,6 +226,8 @@ def check_num_keys_equal(self) -> None:
     def check_keys(self) -> None:
         keys = list(self.store.list())
 
+        # NOTE: A local store can be non-empty if there are no files,
+        # but there are subfolders that weren't cleared.
```
dcherian, line 230: hah this is wild. What does it mean for a `LocalStore` to be empty?

jhamman: this seems like a bug. The expected behavior is to ignore all assets (including empty directories) that are not recognized parts of the hierarchy.

dcherian: Opened #2218 .

tests/v3/conftest.py
```diff
@@ -77,6 +78,14 @@ async def store(request: pytest.FixtureRequest, tmpdir: LEGACY_PATH) -> Store:
     return await parse_store(param, str(tmpdir))
 
 
+@pytest.fixture(params=["local", "memory", "zip"])
+def sync_store(request: pytest.FixtureRequest, tmpdir: LEGACY_PATH) -> Store:
```
dstansby, line 82: While this is being touched, I think it's worth replacing `tmpdir` with `tmp_path` which is the modern way to do temporary directories with pytest

tests/v3/test_store/test_stateful_store.py
```diff
@@ -103,13 +105,17 @@ class ZarrStoreStateMachine(RuleBasedStateMachine):
         https://hypothesis.readthedocs.io/en/latest/stateful.html
     """
 
-    def __init__(self) -> None:
+    def __init__(self, store) -> None:
```
dstansby, line 108: Could this get a type annotation?

tests/v3/test_store/test_stateful_store.py
```diff
@@ -1,21 +1,23 @@
...
 )
 from hypothesis.strategies import DataObject
 
 import zarr
 from zarr.abc.store import AccessMode, Store
 from zarr.core.buffer import BufferPrototype, cpu, default_buffer_prototype
-from zarr.store import MemoryStore
-from zarr.testing.strategies import key_ranges, paths
+from zarr.store import LocalStore, ZipStore
+from zarr.testing.strategies import key_ranges, keys
```
dstansby, line 20: Using a vague variable name from an import seems a bit risky, since I imagine someone might use `keys` as a local variable too. Can this change to `from zarr.testing import strategies`, and then use `strategies.keys` instead lower down?

dstansby (commented):
Looks good -  a few minor suggestions for slightly clearer code

jhamman (approved):

dcherian:
Still fails for LocalStore, will debug later.

PR #2208, opened 2024-09-18
Handle fill_value for bytes dtype

TomAugspurger:
This *partly* fixes how we handle `bytes` dtype arrays. Pushing up early for a bit of a discussion around the spec:

AFAICT, `bytes` aren't a "required"(?) dtype for zarr v3. From https://zarr-specs.readthedocs.io/en/latest/v3/core/v3.0.html#fill-value:

> Extensions to the spec that define new data types must also define the JSON fill value representation.

So the proper way to do this would be to

1. Define a zarr extension for bytes dtype (in this case, fixed-width null terminated bytes). Probably need to do the same for unicode
2. In the extension, define how the fill value should be encoded in JSON. Presumably as base64 encoded.

### DIFF
--- a/src/zarr/abc/metadata.py
+++ b/src/zarr/abc/metadata.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import base64
 from collections.abc import Sequence
 from typing import TYPE_CHECKING
 
@@ -29,6 +30,8 @@ class Metadata:
             value = getattr(self, key)
             if isinstance(value, Metadata):
                 out_dict[field.name] = getattr(self, field.name).to_dict()
+            elif isinstance(value, bytes):
+                out_dict[key] = base64.b64encode(value)
             elif isinstance(value, str):
                 out_dict[key] = value
             elif isinstance(value, Sequence):
--- a/tests/v3/test_metadata/test_v3.py
+++ b/tests/v3/test_metadata/test_v3.py
@@ -1,11 +1,13 @@
 from __future__ import annotations
 
+import dataclasses
 import json
 import re
 from typing import TYPE_CHECKING, Literal
 
 from zarr.codecs.bytes import BytesCodec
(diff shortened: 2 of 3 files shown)


src/zarr/abc/metadata.py
```diff
@@ -29,6 +30,8 @@ def to_dict(self) -> dict[str, JSON]:
             value = getattr(self, key)
             if isinstance(value, Metadata):
                 out_dict[field.name] = getattr(self, field.name).to_dict()
+            elif isinstance(value, bytes):
+                out_dict[key] = base64.b64encode(value)
```
TomAugspurger, line 34: In general, we can't assume bytes are valid JSON so they must be transformed somehow. base64 encoding seems as good as anything else.

We'll need to update the `from_dict` to do the decoding. We probably need to see that we're decoding to a bytes dtype and use that when parsing the fill value.

tests/v3/test_metadata/test_v3.py
```diff
@@ -165,6 +167,31 @@ def test_parse_fill_value_invalid_type_sequence(fill_value: Any, dtype_str: str)
...
+def test_parse_fill_value_bytes():
+    result = parse_fill_value("", dtype=np.dtype("S6"))
+    assert result == np.bytes_("")
+
+
+@pytest.mark.parametrize("fill_value", [None, np.bytes_(b"")])
+def test_fill_value_bytes(fill_value: Any) -> None:
+    md = ArrayV3Metadata(
+        shape=(4,),
+        data_type=np.dtype("S6"),
```
jhamman, line 179: 🤦 - we still don't have a string dtype in the spec! In https://github.com/zarr-developers/zarr-python/pull/2209, I implement a dtype validator that would raise an error if this was passed in.

TomAugspurger:
Closing this for now. I'm hoping to do this properly with some proposed extension dtypes to the zarr-specs repo soonish.

PR #2209, opened 2024-09-18
fix: validate v3 dtypes when loading/creating v3 metadata

jhamman:
This PR validates data types in v3 metadata. The behavior implemented raises an error if a data type outside of the current v3 spec is discovered in either a `zarr.json` metadata document or in the `Array` constructor. 

xref: https://github.com/zarr-developers/zarr-python/pull/2200

TODO:

### DIFF
--- a/src/zarr/core/array_spec.py
+++ b/src/zarr/core/array_spec.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any, Literal
 
-from zarr.core.common import parse_dtype, parse_fill_value, parse_order, parse_shapelike
+from zarr.core.common import parse_fill_value, parse_order, parse_shapelike
 
 if TYPE_CHECKING:
     import numpy as np
@@ -29,7 +29,7 @@ class ArraySpec:
         prototype: BufferPrototype,
     ) -> None:
         shape_parsed = parse_shapelike(shape)
-        dtype_parsed = parse_dtype(dtype)
+        dtype_parsed = dtype  # parsing is likely not needed here
         fill_value_parsed = parse_fill_value(fill_value)
         order_parsed = parse_order(order)
 
--- a/src/zarr/core/array_spec.py
+++ b/src/zarr/core/array_spec.py
@@ -29,12 +29,11 @@ class ArraySpec:
         prototype: BufferPrototype,
     ) -> None:
         shape_parsed = parse_shapelike(shape)
-        dtype_parsed = dtype  # parsing is likely not needed here
         fill_value_parsed = parse_fill_value(fill_value)
         order_parsed = parse_order(order)
 
         object.__setattr__(self, "shape", shape_parsed)
-        object.__setattr__(self, "dtype", dtype_parsed)
+        object.__setattr__(self, "dtype", dtype)
         object.__setattr__(self, "fill_value", fill_value_parsed)
         object.__setattr__(self, "order", order_parsed)
         object.__setattr__(self, "prototype", prototype)
--- a/src/zarr/core/metadata/v2.py
+++ b/src/zarr/core/metadata/v2.py
@@ -21,7 +21,7 @@ import numpy as np
 from zarr.core.array_spec import ArraySpec
 from zarr.core.chunk_grids import RegularChunkGrid
 from zarr.core.chunk_key_encodings import parse_separator
-from zarr.core.common import ZARRAY_JSON, ZATTRS_JSON, parse_dtype, parse_shapelike
+from zarr.core.common import ZARRAY_JSON, ZATTRS_JSON, parse_shapelike
 from zarr.core.config import config, parse_indexing_order
 from zarr.core.metadata.common import ArrayMetadata, parse_attributes
 
@@ -155,6 +155,11 @@ class ArrayV2Metadata(ArrayMetadata):
 
     def update_attributes(self, attributes: dict[str, JSON]) -> Self:
         return replace(self, attributes=attributes)
+
+
+def parse_dtype(data: npt.DTypeLike) -> np.dtype[Any]:
+    # todo: real validation
+    return np.dtype(data)
 
 
 def parse_zarr_format(data: object) -> Literal[2]:
--- a/src/zarr/core/metadata/v2.py
+++ b/src/zarr/core/metadata/v2.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 from collections.abc import Iterable
+from enum import Enum
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
@@ -100,9 +101,24 @@ class ArrayV2Metadata(ArrayMetadata):
                 else:
                     return o.descr
             if np.isscalar(o):
-                # convert numpy scalar to python type, and pass
-                # python types through
-                return getattr(o, "item", lambda: o)()
+                out: Any
+                if hasattr(o, "dtype") and o.dtype.kind == "M" and hasattr(o, "view"):
+                    # https://github.com/zarr-developers/zarr-python/issues/2119
+                    # `.item()` on a datetime type might or might not return an
+                    # integer, depending on the value.
+                    # Explicitly cast to an int first, and then grab .item()
+                    out = o.view("i8").item()
+                else:
+                    # convert numpy scalar to python type, and pass
+                    # python types through
+                    out = getattr(o, "item", lambda: o)()
+                    if isinstance(out, complex):
+                        # python complex types are not JSON serializable, so we use the
+                        # serialization defined in the zarr v3 spec
+                        return [out.real, out.imag]
+                return out
+            if isinstance(o, Enum):
+                return o.name
             raise TypeError
 
         zarray_dict = self.to_dict()
--- a/src/zarr/core/metadata/v2.py
+++ b/src/zarr/core/metadata/v2.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from collections.abc import Iterable
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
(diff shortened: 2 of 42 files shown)


src/zarr/core/array_spec.py
```diff
@@ -29,7 +29,7 @@ def __init__(
         prototype: BufferPrototype,
     ) -> None:
         shape_parsed = parse_shapelike(shape)
-        dtype_parsed = parse_dtype(dtype)
+        dtype_parsed = dtype  # parsing is likely not needed here
```
jhamman, line 32:
```suggestion
        dtype_parsed = dtype
```

The input type is already `np.dtype[Any]` so I don't think we need to parse this.

src/zarr/core/metadata/v2.py
```diff
@@ -157,6 +157,11 @@ def update_attributes(self, attributes: dict[str, JSON]) -> Self:
         return replace(self, attributes=attributes)
 
 
+def parse_dtype(data: npt.DTypeLike) -> np.dtype[Any]:
+    # todo: real validation
+    return np.dtype(data)
```
jhamman, line 162: we now parse v2 and v3 dtypes differently.

src/zarr/testing/strategies.py
```diff
@@ -35,7 +35,9 @@
 paths = st.lists(node_names, min_size=1).map(lambda x: "/".join(x)) | st.just("/")
 np_arrays = npst.arrays(
     # TODO: re-enable timedeltas once they are supported
-    dtype=npst.scalar_dtypes().filter(lambda x: x.kind != "m"),
+    dtype=npst.scalar_dtypes().filter(
+        lambda x: (x.kind not in ["m", "M"]) and (x.byteorder not in [">"])
+    ),
```
jhamman, line 40: It is not clear to me how endianness works in v3. As far as I can tell, we are not handling this in a meaningful way today. 

cc @dcherian and @d-v-b

tests/v3/test_metadata/test_v3.py
```diff
@@ -234,22 +237,43 @@ def test_metadata_to_dict(
     assert observed == expected
 
 
-@pytest.mark.parametrize("fill_value", [-1, 0, 1, 2932897])
-@pytest.mark.parametrize("precision", ["ns", "D"])
-async def test_datetime_metadata(fill_value: int, precision: str) -> None:
+# @pytest.mark.parametrize("fill_value", [-1, 0, 1, 2932897])
+# @pytest.mark.parametrize("precision", ["ns", "D"])
+# async def test_datetime_metadata(fill_value: int, precision: str) -> None:
```
jhamman, line 242: this test was creating an invalid dtype (datetime). We can leave it in place and bring it back when there is support for this dtype in v3.

src/zarr/testing/strategies.py
dcherian: I concur. It's not being handled today

tests/v3/test_metadata/test_v3.py
```diff
@@ -76,8 +79,19 @@ def test_parse_auto_fill_value(dtype_str: str) -> None:
...
+@pytest.mark.parametrize(
+    "fill_value,dtype_str",
+    [
+        (True, "bool"),
+        (False, "bool"),
+        (-8, "int8"),
+        (0, "int16"),
+        (1e10, "uint64"),
+        (-999, "float32"),
+        (1e32, "float64"),
```
jhamman, line 91: also add a test for NaN fill values

src/zarr/core/metadata/v3.py
```diff
@@ -399,7 +482,24 @@ def from_dtype(cls, dtype: np.dtype[Any]) -> DataType:
...
+            "<f2": "float16",
             "<f4": "float32",
             "<f8": "float64",
+            "<c8": "complex64",
+            "<c16": "complex128",
         }
         return DataType[dtype_to_data_type[dtype.str]]
+
+
+def parse_dtype(data: npt.DTypeLike) -> np.dtype[Any]:
```
d-v-b, line 494: do we need a narrow annotation for the input? the idea for `parse_dtype` is to accept _anything_ and error if it's invalid. if we use a narrow annotation for the input (e.g., `npt.DTypeLike`) then, for inputs with unknown types (like stuff we read from JSON), then we need some type checking / casting in advance to appease mypy

jhamman: Reading the docs, this basically is an object type:

```
A Union representing objects that can be coerced into a dtype.

Among others this includes the likes of:

type objects.
Character codes or the names of type objects.
Objects with the .dtype attribute.

```

d-v-b: I just checked the output of `json.loads` and it's typed as `Any`, so I guess we won't need any type narrowing for inputs to this function then, but if we start needing that then we should consider widening the annotation

src/zarr/core/array_spec.py
```diff
@@ -29,12 +29,11 @@ def __init__(
         prototype: BufferPrototype,
     ) -> None:
         shape_parsed = parse_shapelike(shape)
-        dtype_parsed = parse_dtype(dtype)
         fill_value_parsed = parse_fill_value(fill_value)
         order_parsed = parse_order(order)
 
         object.__setattr__(self, "shape", shape_parsed)
-        object.__setattr__(self, "dtype", dtype_parsed)
+        object.__setattr__(self, "dtype", dtype)
```
d-v-b, line 36: i'm not sure about this -- the dtype is _annotated_ as `np.dtype[Any]` but we don't know if that's what we will get at runtime. I feel like parsing is safer, and more consistent with the rest of the` __init__` routine

src/zarr/core/metadata/v2.py
```diff
@@ -100,9 +101,24 @@ def _json_convert(
...
+                else:
+                    # convert numpy scalar to python type, and pass
+                    # python types through
+                    out = getattr(o, "item", lambda: o)()
+                    if isinstance(out, complex):
+                        # python complex types are not JSON serializable, so we use the
+                        # serialization defined in the zarr v3 spec
+                        return [out.real, out.imag]
+                return out
+            if isinstance(o, Enum):
```
d-v-b, line 120: Should we serialize the enum `name` or `value`?

src/zarr/core/metadata/v3.py
```diff
@@ -70,6 +71,67 @@ def parse_dimension_names(data: object) -> tuple[str | None, ...] | None:
...
+                if isinstance(out, complex):
+                    # python complex types are not JSON serializable, so we use the
+                    # serialization defined in the zarr v3 spec
+                    return [out.real, out.imag]
+                elif np.isnan(out):
+                    return "NaN"
+                elif np.isinf(out):
+                    return "Infinity" if out > 0 else "-Infinity"
+            return out
+        elif isinstance(o, Enum):
```
d-v-b, line 103: see https://github.com/zarr-developers/zarr-python/pull/2209/files#r1771911740

src/zarr/core/metadata/v3.py
```diff
@@ -70,6 +71,67 @@ def parse_dimension_names(data: object) -> tuple[str | None, ...] | None:
...
+            return super().default(o)
+
+
+def _replace_special_floats(obj: object) -> Any:
+    """Helper function to replace NaN/Inf/-Inf values with special strings
+
+    Note: this cannot be done in the V3JsonEncoder because Python's `json.dumps` optimistically
+    converts NaN/Inf values to special types outside of the encoding step.
+    """
+    print(obj)
```
d-v-b, line 120:
```suggestion
```

src/zarr/core/array_spec.py
jhamman: fair point. This PR introduces separate dtype parsers for v2 and v3. Maybe we can simply call `np.dtype` in this `__init__`?

d-v-b: yeah, that also seems fine. we only use this class internally, and long term I envision a future in which this class disappears, but that might take a while.

src/zarr/core/metadata/v3.py
```diff
@@ -215,6 +243,9 @@ def from_dict(cls, data: dict[str, JSON]) -> Self:
         # check that the node_type attribute is correct
         _ = parse_node_type_array(_data.pop("node_type"))
 
+        # check that the data_type attribute is valid
+        _ = DataType(_data["data_type"])
```
d-v-b, line 247: why not use `parse_dtype` here?

d-v-b: alternatively, why don't we rely on the invocation of `parse_dtype` that will occur when we call `cls(**_data)`

src/zarr/core/metadata/v2.py
jhamman: I believe we want `name`. Also noting that this part of the PR is not new, it was just moved from common.

src/zarr/core/metadata/v3.py
jhamman: Good question.

`parse_dtype` is used to parse data coming from either user code (e.g. `f8`) or from array metadata (where the value must be `float64` for example). `from_dict` is the one place we can intercept data coming from JSON.

src/zarr/core/metadata/v2.py
d-v-b: ah ok, and this I guess is specifically for the `DataType` enum.

src/zarr/core/metadata/v3.py
jhamman: Without this like, we would silently handle a dtype like `f8` coming from 
`zarr.json`.

d-v-b: I see, so we need separate "parse this dtype-like thing that a user provided" logic from "validate the `dtype` field in this zarr.json document" logic. I feel like the best way to do this is with a function called `validate_dtype` or something that ensures that the input is from the finite set of dtype symbols allowed in the spec. but I think that's a project for another PR, and doesn't block this one

d-v-b (approved):

PR #2210, opened 2024-09-18
[v3] Implement Group methods for empty, full, ones, and zeros

### LINKED ISSUE #2125: [v3] implement `Group.{empty,zeros,ones,full,empty_like, zeros_like,onest_like,full_like}`
Many of these are already defined as top level functions and should be easy to port over to the [Group class](https://github.com/zarr-developers/zarr-python/blob/61683be6f13775c3b9fa78b433812ec1fac41786/src/zarr/core/group.py#L504-L526). These will all provide backward compatibility with zarr-python 3

- [ ] `Group.empty`
- [ ] `Group.zeros`
- [ ] `Group.ones`
- [ ] `Group.full`
- [ ] `Group.empty_like`
- [ ] `Group.zeros_like`
- [ ] `Group.ones_like`
- [ ] `Group.full_like`

lindseynield:
This PR implements `Group.{empty,zeros,ones,full,empty_like, zeros_like,onest_like,full_like}` for V3. Here we simply call the top-level functions from `zarr.api.asynchronous` to accomplish this. 

Closes #2125 

TODO:

### DIFF
--- a/src/zarr/core/group.py
+++ b/src/zarr/core/group.py
@@ -706,45 +706,45 @@ class AsyncGroup:
         raise NotImplementedError
 
     async def empty(self, *, name: str, shape: ChunkCoords, **kwargs: Any) -> AsyncArray:
-        path = self.store_path / name
+        path = str(self.store_path / name)
         return await async_api.empty(shape=shape, path=path, **kwargs)
 
     async def zeros(self, *, name: str, shape: ChunkCoords, **kwargs: Any) -> AsyncArray:
-        path = self.store_path / name
+        path = str(self.store_path / name)
         return await async_api.zeros(shape=shape, path=path, **kwargs)
 
     async def ones(self, *, name: str, shape: ChunkCoords, **kwargs: Any) -> AsyncArray:
-        path = self.store_path / name
+        path = str(self.store_path / name)
         return await async_api.ones(shape=shape, path=path, **kwargs)
 
     async def full(
         self, *, name: str, shape: ChunkCoords, fill_value: Any | None, **kwargs: Any
     ) -> AsyncArray:
-        path = self.store_path / name
+        path = str(self.store_path / name)
         return await async_api.full(shape=shape, fill_value=fill_value, path=path, **kwargs)
 
     async def empty_like(
         self, *, name: str, prototype: async_api.ArrayLike, **kwargs: Any
     ) -> AsyncArray:
-        path = self.store_path / name
+        path = str(self.store_path / name)
         return await async_api.empty_like(a=prototype, path=path, **kwargs)
 
     async def zeros_like(
         self, *, name: str, prototype: async_api.ArrayLike, **kwargs: Any
     ) -> AsyncArray:
-        path = self.store_path / name
+        path = str(self.store_path / name)
         return await async_api.zeros_like(a=prototype, path=path, **kwargs)
 
     async def ones_like(
         self, *, name: str, prototype: async_api.ArrayLike, **kwargs: Any
     ) -> AsyncArray:
-        path = self.store_path / name
+        path = str(self.store_path / name)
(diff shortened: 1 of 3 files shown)


tests/v3/test_group.py
```diff
@@ -390,6 +390,64 @@ def test_group_create_array(
...
+    store: Store,
+    zarr_format: ZarrFormat,
+):
+    group = Group.create(store, zarr_format=zarr_format)
+    shape = (10, 10)
+
+    empty_array = group.empty(shape=shape)
+    assert isinstance(empty_array, Array)
+    assert empty_array.fill_value == 0
+    assert empty_array.shape == shape
```
lindseynield, line 403: Fill value expected to be 0: https://github.com/zarr-developers/zarr-python/blob/v3/src/zarr/core/array.py#L253-L257

src/zarr/core/group.py
```diff
@@ -704,29 +705,29 @@ async def arrays(self) -> AsyncGenerator[AsyncArray, None]:
     async def tree(self, expand: bool = False, level: int | None = None) -> Any:
         raise NotImplementedError
 
-    async def empty(self, **kwargs: Any) -> AsyncArray:
-        raise NotImplementedError
+    async def empty(self, shape: ChunkCoords, **kwargs: Any) -> AsyncArray:
+        return await async_api.empty(shape=shape, **kwargs)
```
jhamman, line 709: I'm realizing that whoever wrote these stub method signatures was not really thinking (it was me 🤦 ).

The 2.x signature was this: 

```python
def empty(self, name: str, **kwargs) -> Array:
```
Numpy's equivalent is:
```python
numpy.empty(shape, dtype=float, order='C', *, device=None, like=None)
```

Here's a proposed signature with some pseudo code:

```suggestion
    async def empty(self, *, shape: ChunkCoords, name: str, **kwargs: Any) -> AsyncArray:
        path = self.store_path / name
        return await async_api.empty(shape=shape, path=path**kwargs)
```

tests/v3/test_group.py
```diff
@@ -390,6 +390,64 @@ def test_group_create_array(
...
+    # TODO: fix tests with path
+    empty_array = group.empty(name="empty", shape=shape)
+    assert isinstance(empty_array, Array)
+    assert empty_array.fill_value == 0
+    assert empty_array.shape == shape
+
+    empty_like_array = group.empty_like(name="empty_like", prototype=empty_array)
+    assert isinstance(empty_like_array, Array)
+    assert empty_like_array.fill_value == 0
+    assert empty_like_array.shape == shape
```
jhamman, line 407: Let's add some test coverage around the store. The top level API has som "auto" store discovery and since we're passing the `path` (type: `StorePath`) instead of the `store` parameter, I'm concerned that we aren't using the store we think we are.

Not 100% sure this is what you want but something like this:

```suggestion
    assert empty_like_array.store is store
```

src/zarr/core/group.py
lindseynield: I am wondering if we want to set `path=self.store_path / name` here. It seems like in `create()` we append the `path` to the `store`: https://github.com/zarr-developers/zarr-python/blob/v3/src/zarr/api/asynchronous.py#L729-L730. Would we want the `path` to just be the `name` then?

tests/v3/test_group.py
lindseynield: Instead of passing the store into the `path` (which would set the array's `store_path` to `make_store_path(store, mode=mode) / self.store_path / name`, I passed the store and the path separately; this sets the array's `store_path` to `self.store_path / name`. Let me know if this is not the behavior that you're looking for!

jhamman: Yeah! This looks right. Thanks :)

jhamman (approved):

jhamman:
Thanks @lindseynield and welcome to the Zarr project!

PR #2224, opened 2024-09-23
testing/store.py: Fix typo in chunk paths

dcherian:
I think this is what's intended.


dcherian:
ah this is allowed in v2. sorry for the noise.

PR #2242, opened 2024-09-25
Abstract method should raise NotImplementedError

DimitriPapadopoulos:
[Description of PR]

TODO:

### DIFF
--- a/src/zarr/abc/codec.py
+++ b/src/zarr/abc/codec.py
@@ -276,15 +276,17 @@ class CodecPipeline:
         -------
         Self
         """
-        ...
+        raise NotImplementedError
 
     @property
     @abstractmethod
-    def supports_partial_decode(self) -> bool: ...
+    def supports_partial_decode(self) -> bool:
+        raise NotImplementedError
 
     @property
     @abstractmethod
-    def supports_partial_encode(self) -> bool: ...
+    def supports_partial_encode(self) -> bool:
+        raise NotImplementedError
 
     @abstractmethod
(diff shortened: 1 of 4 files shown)


TomAugspurger:
Thanks. Out of curiosity, did you hit a problem with this? https://docs.python.org/3/library/abc.html#abc.abstractmethod notes that the abstractmethod decorator will prevent you from even creating an instance of the class, so this code should be unreachable.

DimitriPapadopoulos:
No real problem. I am merely trying to silence linter rules ([deepsource](https://deepsource.com/)). However, some of these rules don't really make sense. Rule `PTC-W0053` is described as follows:
> Abstract methods should raise `NotImplementeError` when they require derived classes to override the method, or while the class is being developed to indicate that the real implementation still needs to be added.
> 
> Refer to the [docs](https://docs.python.org/3/library/exceptions.html#NotImplementedError) to read more about `NotImplementedError`.

The link to the Python docs reads:
> _`exception`_ **`NotImplementedError`**
> This exception is derived from [`RuntimeError`](https://docs.python.org/3/library/exceptions.html#RuntimeError). In user defined base classes, abstract methods should raise this exception when they require derived classes to override the method, or while the class is being developed to indicate that the real implementation still needs to be added.

I guess the implementers of the linter just read and implemented that paragraph naively, without realising that `@abstractmethod` is equivalent. Perhaps we should just disregard the linker and close this issue.

jhamman:
> Perhaps we should just disregard the linker and close this issue.

+1

Thanks for all your help on the code quality front @DimitriPapadopoulos!

PR #2249, opened 2024-09-25
Allow mode casting for `Store`s

TomAugspurger:
Following the discussion in https://github.com/zarr-developers/zarr-python/pull/2186#discussion_r1763478659, this is a POC for opening "reopening" a store with a new mode.

The semantics are that reopening a store with a new mode should return a new Store that's backed by the same underlying data.

AFAICT, our current stores use / enforce `mode` "logically". There's nothing about the actual underlying store that's readonly when `mode="r"`. You could imagine some cases where it's not possible to cast between modes (e.g. you have a readonly database connection). I think it's fine for those Stores to handle that.

By default, the method is abstract meaning this is a new bit of code required for stores to implement. We could also raise `NotImplementedError` by default. 

TODO:


src/zarr/store/zip.py
```diff
@@ -115,6 +117,11 @@ async def empty(self) -> bool:
             else:
                 return True
 
+    def with_mode(self, mode: ZipStoreAccessModeLiteral) -> Self:  # type: ignore[override]
+        return type(self)(
```
TomAugspurger, line 121: This probably needs some work... would we want the new Store to share the same Lock?

src/zarr/abc/store.py
```diff
@@ -81,6 +81,13 @@ async def empty(self) -> bool: ...
     @abstractmethod
     async def clear(self) -> None: ...
 
+    @abstractmethod
+    def with_mode(self, mode: AccessModeLiteral) -> Self:
+        """
+        Return a new store pointing to the same location with a new mode.
```
TomAugspurger, line 87: This should document the semantics:

1. same backing data
2. require that `_open` be called? We currently just call it from `make_store_path`, which immediately calls `ensure_open`.

src/zarr/store/zip.py
TomAugspurger: This wasn't straightforward to implement, because of a need to close the old ZipStore (and writing the ending records) before opening the new store.

For now I've opted to raise NotImplementedError for ZipStore.

TomAugspurger:
The GPU test failed. I _think_ that's because at https://github.com/zarr-developers/zarr-python/blob/692593bb32bcd8f526f1b611c8c6ff0bf758ceff/src/zarr/store/memory.py#L170-L171 we create a new dictionary where, so that `store.with_mode('r')` no longer views the same dict as the first store.

@akshaysubr I suspect that those lines are there to ensure that all of the values in the dictionary are GPU buffers. What do you think about documenting that as a requirement to use the store (and maybe narrowing the type of the dict values in `__init__` to a `gpu.Buffer`?

src/zarr/store/memory.py
```diff
@@ -15,6 +17,12 @@
     from zarr.core.common import AccessModeLiteral
 
 
+# T = TypeVar("T", bound=Buffer | gpu.Buffer)
+
+
+# class _MemoryStore
+
+
```
jhamman, line 25:
```suggestion
```

src/zarr/testing/store.py
```diff
@@ -251,3 +252,28 @@ async def test_list_dir(self, store: S) -> None:
...
+            assert self.get(clone, "key").to_bytes() == data
+
+            # writes to original after with_mode is visible
+            self.set(store, "key-2", self.buffer_cls.from_bytes(data))
+            assert self.get(clone, "key-2").to_bytes() == data
+
+            if mode == "w":
+                # writes to clone is visible in the original
+                self.set(store, "key-3", self.buffer_cls.from_bytes(data))
+                assert self.get(clone, "key-3").to_bytes() == data
```
jhamman, line 279: I think this block is unreachable

src/zarr/testing/store.py
```diff
@@ -251,3 +252,28 @@ async def test_list_dir(self, store: S) -> None:
 
         keys_observed = await _collect_aiterator(store.list_dir(root + "/"))
         assert sorted(keys_expected) == sorted(keys_observed)
+
+    async def test_with_mode(self, store: S) -> None:
+        data = b"0000"
+        self.set(store, "key", self.buffer_cls.from_bytes(data))
+        assert self.get(store, "key").to_bytes() == data
+
+        for mode in ["r", "a"]:
```
jhamman, line 261: I would like to see a test that verifies that a store rejects writes if it has been `with_mode('r')`'ed

jhamman (commented):
Thanks @TomAugspurger -- a few suggestions but this is great.

cc @d-v-b who was also looking at this problem today.

src/zarr/testing/store.py
d-v-b: ditto

d-v-b:
> Thanks @TomAugspurger -- a few suggestions but this is great.
> 
> cc @d-v-b who was also looking at this problem today.

second this, once those suggestions are integrated then I think this will be a great addition.

src/zarr/testing/store.py
TomAugspurger: Thanks. I switched from `w` to `a` and forgot to update this. Fixed now.

src/zarr/testing/store.py
TomAugspurger: Fixed.

src/zarr/testing/store.py
```diff
@@ -251,3 +252,38 @@ async def test_list_dir(self, store: S) -> None:
...
+            await clone._ensure_open()
+            assert clone.mode == AccessMode.from_literal(mode)
+            assert isinstance(clone, type(store))
+
+            # earlier writes are visible
+            result = await clone.get("key", default_buffer_prototype())
+            assert result is not None
+            assert result.to_bytes() == data
+
+            # # writes to original after with_mode is visible
```
d-v-b, line 274:
```suggestion
            # writes to original after with_mode is visible
```

TomAugspurger:
It's tangentially related, but I wanted to confirm that people are OK with changing the `GpuMemoryStore.__init__` method to *not* automatically copy the data from an arbitrary `Buffer` into a `gpu.Buffer`. I added a `from_dict` method for people who want that, but I imagine that the common case is a user who knows where the data is and doesn't want to pay the cost of walking that dictionary and potentially moving data in the `__init__`.

jhamman:
> It's tangentially related, but I wanted to confirm that people are OK with changing the `GpuMemoryStore.__init__` method to _not_ automatically copy the data from an arbitrary `Buffer` into a `gpu.Buffer`. I added a `from_dict` method for people who want that, but I imagine that the common case is a user who knows where the data is and doesn't want to pay the cost of walking that dictionary and potentially moving data in the `__init__`.

cc @akshaysubr

src/zarr/store/memory.py
```diff
@@ -152,29 +155,58 @@ async def list_dir(self, prefix: str) -> AsyncGenerator[str, None]:
...
+        super().__init__(store_dict=store_dict, mode=mode)  # type: ignore[arg-type]
 
     def __str__(self) -> str:
         return f"gpumemory://{id(self._store_dict)}"
 
     def __repr__(self) -> str:
         return f"GpuMemoryStore({str(self)!r})"
 
+    @classmethod
+    def from_dict(cls, store_dict: MutableMapping[str, Buffer]) -> Self:
```
jhamman, line 190: I like this API!

jhamman (approved):
This looks good to me @TomAugspurger - some failing tests in GPU land but otherwise looks good to me.

src/zarr/store/logging.py
```diff
@@ -95,6 +98,12 @@ def _is_open(self) -> bool:  # type: ignore[override]
         with self.log():
             return self._store._is_open
 
+    async def _open(self) -> None:
+        return await self._store._open()
```
jhamman, line 102:
```suggestion
        with self.log():
            return await self._store._open()
```

src/zarr/store/logging.py
```diff
@@ -95,6 +98,12 @@ def _is_open(self) -> bool:  # type: ignore[override]
...
+
+    async def _ensure_open(self) -> None:
+        return await self._store._ensure_open()
```
jhamman, line 105:
```suggestion
        with self.log():
            return await self._store._ensure_open()
```

src/zarr/store/logging.py
```diff
@@ -162,3 +171,10 @@ async def list_dir(self, prefix: str) -> AsyncGenerator[str, None]:
         with self.log():
             async for key in self._store.list_dir(prefix=prefix):
                 yield key
+
+    def with_mode(self, mode: AccessModeLiteral) -> Self:
+        return type(self)(
+            self._store.with_mode(mode),
+            log_level=self.log_level,
+            log_handler=self.log_handler,
+        )
```
jhamman, line 180:
```suggestion
        with self.log():
            return type(self)(
                self._store.with_mode(mode),
                log_level=self.log_level,
                log_handler=self.log_handler,
            )
```

jhamman (approved):

jhamman:
pre-commit.ci autofix

PR #2256, opened 2024-09-25
fix(async): set default concurrency to 10 tasks

jhamman:
Unbounded concurrency is not a good idea. This sets the default to 10. 

TODO:

### DIFF
--- a/src/zarr/core/config.py
+++ b/src/zarr/core/config.py
@@ -43,7 +43,7 @@ config = Config(
         {
             "default_zarr_version": 3,
             "array": {"order": "C"},
-            "async": {"concurrency": None, "timeout": None},
+            "async": {"concurrency": 10, "timeout": None},
             "json_indent": 2,
             "codec_pipeline": {
                 "path": "zarr.codecs.pipeline.BatchedCodecPipeline",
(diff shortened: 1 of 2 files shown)


dcherian:
I don't have an opinion but this is what the threading module [does](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor)

> Default value of max_workers is changed to min(32, os.cpu_count() + 4). This default value preserves at least 5 workers for I/O bound tasks. It utilizes at most 32 CPU cores for CPU bound tasks which release the GIL. And it avoids using very large resources implicitly on many-core machines.

jhamman:
There are really two issues here. 

1) how many threads we try to use at once, and
2) how many concurrent IO operations we attempt

Right now, we are using a single config for both of these but in the future, we may consider allowing users to configure the size of the thread pool and/or multi-threading behavior separately.

dcherian:
But perhaps the default should scale with cpu_count like what `threading` does: `min(32, os.cpu_count() + 4)`. If we are making arbitrary choices, perhaps we just go with what the standard library does.

dcherian (approved):

Issue #2269, opened 2024-09-29
`TypeError: ArrayV2Metadata.__init__() got an unexpected keyword argument 'chunk_grid'` when updating attrs on a v2 array


TomAugspurger:
### Zarr version

v3

### Numcodecs version

na

### Python Version

na

### Operating System

na

### Installation

na

### Description

`data_type` is also affected. IIUC, these are called `chunks` and `dtype` in v2. We need to mark them as `dataclasses.field(init=False)` or somehow handle renaming them on the way in.

### Steps to reproduce

```python
import numpy as np
from zarr.storage import MemoryStore, StorePath
from zarr import Array


store = StorePath(MemoryStore(mode="w"), path="/")
data = np.arange(0, 8, dtype="uint16")
a = Array.create(
    store / "simple_v2",
    zarr_format=2,
    shape=data.shape,
    chunks=(4,),
    dtype=data.dtype,
    fill_value=0,
)
a.attrs.put({"key": 0})
```

raises with

```pytb
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[10], line 10
      1 data = np.arange(0, 8, dtype="uint16")
      2 a = Array.create(
      3     store,
      4     zarr_format=2,
   (...)
      8     fill_value=0,
      9 )
---> 10 a.attrs.put({"key": 0})

File ~/gh/zarr-developers/zarr-python/src/zarr/core/attributes.py:53, in Attributes.put(self, d)
     39 def put(self, d: dict[str, JSON]) -> None:
     40     """
     41     Overwrite all attributes with the values from `d`.
     42 
   (...)
     51        {'a': 3, 'c': 4}
     52     """
---> 53     self._obj = self._obj.update_attributes(d)

File ~/gh/zarr-developers/zarr-python/src/zarr/core/array.py:2318, in Array.update_attributes(self, new_attributes)
   2316 def update_attributes(self, new_attributes: dict[str, JSON]) -> Array:
   2317     return type(self)(
-> 2318         sync(
   2319             self._async_array.update_attributes(new_attributes),
   2320         )
   2321     )

File ~/gh/zarr-developers/zarr-python/src/zarr/core/sync.py:91, in sync(coro, loop, timeout)
     88 return_result = next(iter(finished)).result()
     90 if isinstance(return_result, BaseException):
---> 91     raise return_result
     92 else:
     93     return return_result

[...]

PR #2270, opened 2024-09-30
Fixed ArrayV2Metadata parameter names

TomAugspurger:
For things like dataclasses.repalce to work, we need the parameter names to match the attribute names.

All the name change between what we have in memory and what the specs requires should happen during serialization / deserialization.

Closes https://github.com/zarr-developers/zarr-python/issues/2269

TODO:


d-v-b:
Can you show the problem that this PR solves? From my POV it's a good thing that the metadata classes have the same structure as the underlying zarr metadata document(s).

TomAugspurger:
https://github.com/zarr-developers/zarr-python/issues/2269 has the issue. Doing something like `a.attrs.put({"key": 0})` or anything that uses `dataclasses.replace` will fail with

```python
TypeError: ArrayV2Metadata.__init__() got an unexpected keyword argument 'chunk_grid'
```

> From my POV it's a good thing that the metadata classes have the same structure as the underlying zarr metadata document

Would that mean changing the dataclass fields on `ArrayV2Metadata` to `chunks` and `dtype`? I'm guessing we have it this way currently is to support writing code that's generic over `ArrayV2Metadata` and `ArrayMetadata`. IMO, that's more important than having the Python data model exactly match the on-disk representation, since the translation should be isolated to just `to_dict` and `from_dict`.

d-v-b:
I think a design goal for `ArrayV2Metadata` and `ArrayV3Metadata` is that they should be as independent as possible from whatever array API we layer on top of them. This ensures that `zarr-python` can evolve that array API in just one place (the array classes), and also ensures that `zarr-python` is more useful to libraries that want to work with the Zarr data model, but don't want to use the `zarr-python` array runtime classes.  IMO both of these are weakened if our in-memory array abstraction leaks onto the metadata classes.

TomAugspurger:
> I think a design goal for ArrayV2Metadata and ArrayV3Metadata is that they should be as independent as possible from whatever array API we layer on top of them

Mmm, I'm not sure where I stand on that :) I think I agree that it's nice to have something that directly models the stored representation of the data, or is at least trivially convertible to it (and I'd count renaming the key as trivial). But IMO the most important thing is the ability to write code that's generic over v2 and v3. I don't have a strong preference for whether that's achieved by having subtypes with mostly the same interface (ArrayV2Metadata and ArrayV3Metadata using the same field names) or some wrapper class.

I'll leave this on hold for now.

TomAugspurger:
Converting this to draft while https://github.com/zarr-developers/zarr-python/pull/2301 is worked on. But we'll want to confirm that this original issue here is fixed.

TomAugspurger:
This was superseded by #2301.

PR #2272, opened 2024-09-30
Feat/store paths

d-v-b:
Adds a `path` attribute to all the store classes. Key operations like `get`, `set`, `list`, `list_prefix` etc are now scoped to the subtree rooted by the `path` attribute of the store, which is how `RemoteStore` works in the v3 branch. This change is motivated by the desire for a consistent, scalable store API. 

Key changes:
- renamed `LocalStore.root` to `LocalStore.path`, and changed the type of the attribute from `pathlib.Path` to `str`
- added a `path` attribute to `MemoryStore`
- changed the semantics of the `path` attribute of `ZipStore`, and introduced a new attribute `file_path` that does what `ZipStore.path` used to do (point to the location of the Zipfile in the parent file system).

TODO:


TomAugspurger:
Just a note, this fixes a failure in some xarray tests that write an array at the top level of the store, open a group at the root and then get the array:

```python
In [1]: import zarr

In [2]: store = zarr.store.MemoryStore(store_dict={}, mode="w")

In [3]: zarr.open_array(store=store, shape=(4,), path="foo")
Out[3]: <Array memory://4509886080/foo shape=(4,) dtype=float64>

In [4]: zarr.open_group(store)["foo"]  # raises a KeyError on v3 currently
Out[4]: <Array memory://4509886080/foo shape=(4,) dtype=float64>
```

src/zarr/abc/store.py
```diff
@@ -320,6 +336,13 @@ async def _get_many(
         for req in requests:
             yield (req[0], await self.get(*req))
 
+    def with_path(self, path: str) -> Self:
```
jhamman, line 339:
```suggestion
    @abstractmethod
    def with_path(self, path: str) -> Self:
```

Or do you think think we can do this in a generic way that applies to all stores?

src/zarr/store/local.py
```diff
@@ -187,20 +183,20 @@ async def set_partial_values(
...
-        path = self.root / key
+        path = Path(self.path) / key
         if path.is_dir():  # TODO: support deleting directories? shutil.rmtree?
             shutil.rmtree(path)
         else:
             await to_thread(path.unlink, True)  # Q: we may want to raise if path is missing
 
     async def exists(self, key: str) -> bool:
-        path = self.root / key
+        path = Path(self.path) / key
```
jhamman, line 199: You do this a lot. I wonder if the store could have a private attr like `self._ppath = Path(self.path)` that does this once.

src/zarr/store/memory.py
```diff
@@ -28,12 +29,15 @@ def __init__(
         self,
         store_dict: MutableMapping[str, Buffer] | None = None,
         *,
+        path: str = "",
```
jhamman, line 32: this store highlights the fact that not all stores have a meaningful `path`.

src/zarr/store/zip.py
```diff
@@ -51,18 +52,18 @@ class ZipStore(Store):
 
     def __init__(
         self,
-        path: Path | str,
+        file_path: Path | str,
         *,
+        path: str = "",
```
jhamman, line 57: This feels quite confusing. IIUC, you are allowing the caller to specify a path within the zipfile as opposed to putting the root of the store at the root of the zip, is that right?

jhamman (commented):
Thanks for working on this @d-v-b. I'm not 100% convinced the `with_mode` approach is going to be substantially cleaner than our `StorePath` right now but I'm curious to see how this plays out.

src/zarr/abc/store.py
d-v-b: I don't think we can make it generic unfortunately, because each store class might have its own extra attributes. _maybe_ we could use `__getstate__` to get around this?

src/zarr/store/local.py
d-v-b: I had a similar thought when I found myself doing this over and over again 😅

src/zarr/store/memory.py
d-v-b: I'm not sure I agree -- if `store_dict` really is an arbitrary mutable mapping, there's no guarantee that we want all IO operations to be scoped to the top-level key space, and that's exactly what the `path` attribute is for.

src/zarr/store/zip.py
d-v-b: exactly! this is required to make `ZipStore` congruent to `RemoteStore`

ghost (approved):

src/zarr/store/zip.py
joshmoore: though this may have some impact (*simplifying*) https://github.com/zarr-developers/zarr-specs/pull/311

src/zarr/store/local.py
d-v-b: done in https://github.com/zarr-developers/zarr-python/pull/2272/commits/c6c88435e8967c34f2bba2c529053897cae51798

d-v-b:
worth mentioning some changes to the reprs:

- In main, `MemoryStore` and `GPUMemoryStore` reprs are `<store class name>://<id of ._store_dict>`. This PR changes that to `<store class name>://<id of ._store_dict>/<store.path>`
- in main, `LocalStore` repr is `file://<store.root>`, which is allegedly [wrong for a file url](https://en.wikipedia.org/wiki/File_URI_scheme). In this PR, it's `file:///<store.path>`
- `ZipStore` is a bit weird, because it's a nested store: we need a wrapping file system to get to the zip file, and then inside the zip file we have another file system. In this PR i'm using the syntax introduced by @jbms for this purpose, which is basically a sequence of URLs separated by the `|` character, e.g. `file:///<store.file_path>|zip://<store.path>` (Jeremy please correct me if I'm wrong here).

jbms:
> worth mentioning some changes to the reprs:
> 
> * In main, `MemoryStore` and `GPUMemoryStore` reprs are `<store class name>://<id of ._store_dict>`. This PR changes that to `<store class name>://<id of ._store_dict>/<store.path>`
> * in main, `LocalStore` repr is `file://<store.root>`, which is allegedly [wrong for a file url](https://en.wikipedia.org/wiki/File_URI_scheme). In this PR, it's `file:///<store.path>`
> * `ZipStore` is a bit weird, because it's a nested store: we need a wrapping file system to get to the zip file, and then inside the zip file we have another file system. In this PR i'm using the syntax introduced by @jbms for this purpose, which is basically a sequence of URLs separated by the `|` character, e.g. `file:///<store.file_path>|zip://<store.path>` (Jeremy please correct me if I'm wrong here).

I didn't see this before, but note that neuroglancer (and soon tensorstore) use the syntax:

`base-url|zip:path/within/zip`

not

`base-url|zip://path/within/zip`


Regarding `file:` --- currently tensorstore always uses `file://PATH` where `PATH` is typically an absolute path (i.e. beginning with `/`, such that you have `file:///tmp/dataset/`) but tensorstore also allows a relative path as a non-standard extension.

According to https://en.wikipedia.org/wiki/File_URI_scheme it is supposed to be either `file:///tmp/dataset/` or `file:/tmp/dataset/` but not `file://tmp/dataset/` --- the optional extra two slashes at the beginning are for indicating a "hostname", which as far as I am aware is always empty, i.e. unused.

I should probably fix tensorstore to accept either form.  For both tensorstore and zarr-python allowing `file:RELATIVE` or `file://RELATIVE` also makes sense, though I suppose with `file://RELATIVE` there is a risk that the user intended to specify an absolute path but forgot a slash.

src/zarr/storage/zip.py
```diff
@@ -131,13 +132,19 @@ def with_mode(self, mode: ZipStoreAccessModeLiteral) -> Self:  # type: ignore[ov
         raise NotImplementedError("ZipStore cannot be reopened with a new mode.")
 
     def __str__(self) -> str:
-        return f"zip://{self.path}"
+        # lets try https://github.com/zarr-developers/zeps/pull/48/files
+        return f"file:///{self.file_path}|zip://{self.path}"
```
jbms, line 136: I think you have too many slashes after `file:` because `file_path` also starts with a slash, right?

Also you should percent-encode both `file_path` and `path` for consistency with my ZEP proposal and tensorstore, and to avoid ambiguity since `|` is a legal path character except on Windows.

d-v-b: thanks for catching this!

d-v-b:
I'm closing this PR as stale. Expect to see this functionality appear in an upcoming refresh of the store api. Thanks for the discussions all!

Issue #2271, opened 2024-09-30
Don't infer a default `fill_value` for `zarr_format=2`?


TomAugspurger:
### Zarr version

v3

### Numcodecs version

na

### Python Version

na

### Operating System

na

### Installation

na

### Description

Zarr v2 didn't require a fill value to be set when writing data. To make it easier for users to transition to zarr-python v3 (ref https://github.com/pydata/xarray/issues/5475), what do people think about changing the default `fill_value` for array metadata? `ArrayV2Metadata.fill_value` would be allowed to be None in memory and on disk.

We'd still have a fill value implicitly defined using the v2 rules for whatever fill value is actually used at read time.

### Steps to reproduce

```
In [1]: import zarr

In [2]: arr = zarr.open_array(store = zarr.store.MemoryStore(store_dict={}, mode="w"), shape=(4,), zarr_format=2, fill_value=None)  # arr.metadata.fill_value will be None

In [3]: arr[:]  # at read time, find the v2 default and use that.
Out[3]: array([0.0, 0.0, 0.0, 0.0)
```

PR #2274, opened 2024-09-30
Use implicit fill values for zarr v2

TomAugspurger:
Closes https://github.com/zarr-developers/zarr-python/issues/2271

I'm not sure if this is a good idea, but this, consolidated metadata, and #2270 lets xarray round trip a dataset. The basic idea is to let `ArrayV2Metadata` represent the v2 metadata a bit more faithfully. Currently, we "eagerly" interpret the lack of `fill_value` in a V2 document as the default fill value for some dtype (e.g. `0` for int64). This PR changes that, to allow `ArrayV2Metadata.fill_value` to be None if it's unspecified in the metadata. But we *do* need some fill value when reading data. So that's where we finally substitute in the default.

This doesn't really affect Zarr directly, but it is useful for downstream libraries that want to inspect the array metadata before reading and make some decision based on whether or not `fill_value` is specified.

TODO:


src/zarr/codecs/pipeline.py
```diff
@@ -247,7 +248,13 @@ async def read_batch(
                 if chunk_array is not None:
                     out[out_selection] = chunk_array
                 else:
-                    out[out_selection] = chunk_spec.fill_value
+                    fill_value = chunk_spec.fill_value
+
+                    # this should maybe be version specific?
```
jhamman, line 253: given that fill_value is required in `v3`, this path should only be reachable from `v2` arrays.

src/zarr/core/metadata/v2.py
```diff
@@ -62,7 +62,10 @@ def __init__(
         order_parsed = parse_indexing_order(order)
         dimension_separator_parsed = parse_separator(dimension_separator)
         filters_parsed = parse_filters(filters)
-        fill_value_parsed = parse_fill_value(fill_value, dtype=data_type_parsed)
+        if fill_value is None:
+            fill_value_parsed = None
+        else:
+            fill_value_parsed = parse_fill_value(fill_value, dtype=data_type_parsed)
```
jhamman, line 68: rather than putting this check here, can we instead allow `parse_fill_value` to return `None`?

jhamman (commented):
A few comments. Seems reasonable though.

src/zarr/codecs/pipeline.py
TomAugspurger: Currently, the `zarr_format` of the array isn't available to this code. Are you comfortable relying on the fact that `zarr_format=3` implies `fill_value is not None` being validated before we get here?

Personally, I am OK with that and we can just assume that `fill_value` being `None` here implies it is `v2`, rather than having to pass that information around and re-assert it.

TomAugspurger: One potential downside here, this will make it impossible to read object-dtype arrays where the fill value is literally `None`. Now it would return an object-dtype array full of 0s.

IMO, that's an OK limitation to resolve the issue xarray is hitting.

rabernat: > object-dtype arrays where the fill value is literally None

Generic object arrays are not allowed in V3. Are you referring to V2?

In #2036, I am implementing vlen strings and vlen bytes. Both of these come out of the decoding pipeline as object dtype numpy arrays. But `None` is not a valid value for either dtype. Does that collide with this in any way?

TomAugspurger: Nope I think we're clear of both of those. It really would just be the case of object-dtype arrays storing arbitrary objects rather than a specialized type like strings (pretty uncommon in v2, I would guess; and not allowed in v3 yet as you say).

src/zarr/core/metadata/v2.py
TomAugspurger: That already worked.

src/zarr/core/metadata/v2.py
```diff
@@ -284,3 +284,19 @@ def parse_fill_value(fill_value: object, dtype: np.dtype[Any]) -> Any:
             raise ValueError(msg) from e
 
     return fill_value
+
+
+def default_fill_value(dtype: np.dtype[Any]) -> Any:
```
rabernat, line 289: The equivalent method in v3 was made private in #2036.
```suggestion
def _default_fill_value(dtype: np.dtype[Any]) -> Any:
```

src/zarr/core/metadata/v2.py
```diff
@@ -284,3 +284,19 @@ def parse_fill_value(fill_value: object, dtype: np.dtype[Any]) -> Any:
...
+    Notes
+    -----
+    This differs from :func:`parse_fill_value`, which parses a fill value
+    stored in the Array metadata into an in-memory value. This only gives
+    the default fill value for some type.
+
+    This is useful for reading Zarr V2 arrays, which allow the fill
+    value to be unspecified.
+    """
+    return dtype.type(0)
```
rabernat, line 302: What about string or object dtypes? Does 0 really make sense there? In #2036 I used `'' and b''` for strings and bytes.

tests/v3/test_v2.py
```diff
@@ -31,6 +31,15 @@ def test_simple(store: StorePath) -> None:
     assert np.array_equal(data, a[:, :])
 
 
+def test_implicit_fill_value(store: StorePath) -> None:
+    arr = zarr.open_array(store=store, shape=(4,), fill_value=None, zarr_format=2)
```
rabernat, line 35: Consider testing some other dtypes.

rabernat (commented):
This looks good. My one suggestion would be to consider a special case for object / string dtypes.

src/zarr/core/metadata/v2.py
TomAugspurger: Here's the zarr v2 behavior

```
In [13]: zarr.create(shape=(1,), store={}, path="a", dtype="|U1")[:]
Out[13]: array([''], dtype='<U1')

In [14]: zarr.create(shape=(1,), store={}, path="a", dtype="|S1")[:]
Out[14]: array([b''], dtype='|S1')

In [15]: zarr.create(shape=(1,), store={}, path="a", dtype=str)[:]
Out[15]: array([0], dtype=object)
```

The integer `0` there for `str` dtype (which zarr-python 2.x v2 converted to an `object` dtype) doesn't really make sense.

TomAugspurger: ff530c36e41fd725f5a4cdc5e7b35a92a9668e2b restores the 2.x behavior aside from `dtype=str`. In 2.x, that was somehow special cased to use object dtype. Right now on main, that returns a `<U`-dtype array, similar to NumPy:

```
In [1]: import numpy as np

In [2]: np.array([1, 2], dtype=str)
Out[2]: array(['1', '2'], dtype='<U1')
```

I think the treatment of str / object-dtype arrays is mostly orthogonal to this PR, but we might need to update the test I added if we want to change the behavior there.

rabernat (approved):
Awesome, nice work Tom!

src/zarr/core/metadata/v2.py
rabernat: I seriously doubt that anyone is relying on that [weird] behavior. 😆

src/zarr/codecs/pipeline.py
```diff
@@ -247,7 +248,17 @@ async def read_batch(
                 if chunk_array is not None:
                     out[out_selection] = chunk_array
                 else:
-                    out[out_selection] = chunk_spec.fill_value
+                    fill_value = chunk_spec.fill_value
+
+                    if fill_value is None:
```
jhamman, line 253: Something we may consider in the future is to parameterize the `CodecPipeline` class with the `zarr_format` of the calling Array. This would allow us feel more confident that workarounds like this one only apply to v2 data.

jhamman (approved):
Thanks @TomAugspurger!

PR #2278, opened 2024-10-01
Change ArrayV3Metadata.data_type to DataType

rabernat:
In working on #2036, I realized that the current treatment of the `data_type` attribute in V3 metadata was somewhat inconsistent. This class is meant to represent the Zarr metadata in an abstract way, in alignment with the spec. However, it was just using numpy dtypes internally.

This becomes a problem when we want to support dtypes that are not numpy dtypes, like `string`.

This PR changes the `ArrayV3Metadata` class such that the type of `data_type` is the existing `DataType`  enum. However, I've kept the `.dtype` property returning a numpy dtype. This allows this change to be minimally disruptive to other parts of the code.

### DIFF
--- a/src/zarr/core/metadata/v3.py
+++ b/src/zarr/core/metadata/v3.py
@@ -30,6 +30,8 @@ from zarr.core.config import config
 from zarr.core.metadata.common import ArrayMetadata, parse_attributes
 from zarr.registry import get_codec_class
 
+DEFAULT_DTYPE = "float64"
+
 
 def parse_zarr_format(data: object) -> Literal[3]:
     if data == 3:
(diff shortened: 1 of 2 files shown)


src/zarr/core/metadata/v3.py
```diff
@@ -490,8 +492,11 @@ def to_numpy_shortname(self) -> str:
         }
         return data_type_to_numpy[self]
 
+    def to_numpy_dtype(self) -> np.dtype[Any]:
```
d-v-b, line 495: I could see `to_numpy` as a slightly briefer name for this method

src/zarr/core/metadata/v3.py
```diff
@@ -490,8 +492,11 @@ def to_numpy_shortname(self) -> str:
...
+        return np.dtype(self.to_numpy_shortname())
+
     @classmethod
-    def from_dtype(cls, dtype: np.dtype[Any]) -> DataType:
+    def from_numpy_dtype(cls, dtype: np.dtype[Any]) -> DataType:
```
d-v-b, line 499: ditto https://github.com/zarr-developers/zarr-python/pull/2278/files#r1782030539

src/zarr/core/metadata/v3.py
```diff
@@ -511,16 +516,21 @@ def from_dtype(cls, dtype: np.dtype[Any]) -> DataType:
...
-        _ = DataType.from_dtype(dtype)
-    except KeyError as e:
-        raise ValueError(f"Invalid V3 data_type: {dtype}") from e
-
-    return dtype
+    @classmethod
+    def parse(cls, dtype: None | DataType | Any) -> DataType:
+        if dtype is None:
+            # the default dtype
+            return DataType.float64
```
d-v-b, line 523: should we put the default dtype in a module-scoped variable, e.g. `DEFAULT_DTYPE`?

src/zarr/core/metadata/v3.py
d-v-b: modulo to/from of course

d-v-b (approved):

PR #2279, opened 2024-10-01
document storage classes and some developer apis

### LINKED ISSUE #2250: Add documentation for `zarr.storage`
xref: #2215

This issue tracks the documentation progress for the `zarr.storage` module. We'll be focusing on adding docstrings and code samples to the functions.

Progress:

- [ ] docstrings
- [ ] code samples

jhamman:
closes #2250 

TODO:


dstansby:
Would be good to get this up to date with the `v3` branch before reviewing

jhamman:
I can rebase this today.

jhamman:
Okay, @dstansby (and others), this is now ready for a review.

src/zarr/abc/store.py
```diff
@@ -28,6 +30,24 @@ class AccessMode(NamedTuple):
 
     @classmethod
     def from_literal(cls, mode: AccessModeLiteral) -> Self:
+        """
+        Create an AccessMode instance from a literal.
+
+        Parameters
+        ----------
+        mode : AccessModeLiteral
+            One of 'r', 'r+', 'w', 'w-', 'a'.
```
dstansby, line 39: Could we define these as a string somewhere, and then re-use it here and lower down by making the docstring a format string? I worry about lists like this getting out of sync if they're duplicated across docstrings.

src/zarr/abc/store.py
```diff
@@ -40,6 +60,17 @@ def from_literal(cls, mode: AccessModeLiteral) -> Self:
...
 class Store(ABC):
+    """
+    Abstract base class for Zarr stores.
+
+    Attributes
+    ----------
+    _mode : AccessMode
+        Access mode flags.
+    _is_open : bool
+        Whether the store is open.
```
dstansby, line 71: I don't think we should be documenting private attributes, so either the attributes docstring should go, or the attributes should become public.

src/zarr/abc/store.py
```diff
@@ -49,6 +80,21 @@ def __init__(self, *args: Any, mode: AccessModeLiteral = "r", **kwargs: Any) ->
...
+        Parameters
+        ----------
+        *args : Any
+            Positional arguments to pass to the store constructor.
+        **kwargs : Any
+            Keyword arguments to pass to the store constructor.
+
+        Returns
+        -------
+        store
```
dstansby, line 95:
```suggestion
        Store
```
If there's a single entry here, it's interpreted as the return type.

src/zarr/abc/store.py
```diff
@@ -67,6 +113,20 @@ def __exit__(
         self.close()
 
     async def _open(self) -> None:
+        """
+        Open the store.
+
+        Notes
+        -----
+        * When `mode='w'` and the store already exists, it will be cleared.
```
dstansby, line 121:
```suggestion
        * When ``mode='w'`` and the store already exists, it will be cleared.
```
Double backticks for code (otherwise sphinx will try and create a link from it)

src/zarr/storage/common.py
```diff
@@ -39,25 +50,80 @@ async def get(
...
+        """
         await self.store.set_if_not_exists(self.path, default)
 
     async def exists(self) -> bool:
+        """
+        Check if the key exists in the store.
+
+        Returns
+        -------
+        exists : bool
```
dstansby, line 120:
```suggestion
        bool
```

src/zarr/storage/common.py
```diff
@@ -67,6 +133,19 @@ def __repr__(self) -> str:
         return f"StorePath({self.store.__class__.__name__}, {str(self)!r})"
 
     def __eq__(self, other: object) -> bool:
+        """
+        Check if two StorePath objects are equal.
+
+        Returns
+        -------
+        equals : bool
```
dstansby, line 141:
```suggestion
        bool
```

src/zarr/storage/local.py
```diff
@@ -143,17 +149,6 @@ async def get_partial_values(
         prototype: BufferPrototype,
         key_ranges: Iterable[tuple[str, ByteRangeRequest]],
     ) -> list[Buffer | None]:
-        """
-        Read byte ranges from multiple keys.
```
dstansby, line 147: Did you mean to delete all these docstrings?

src/zarr/storage/logging.py
```diff
@@ -18,6 +18,26 @@
...
+    store: Store
+        Store to wrap
+    log_level: str
+        Log level
+    log_handler: logging.Handler
+        Log handler
+
+    Attributes
+    ----------
+    _store: Store
```
dstansby, line 35: Private attribute, either don't document or make non-private

src/zarr/storage/logging.py
```diff
@@ -58,6 +78,11 @@ def _default_handler(self) -> logging.Handler:
 
     @contextmanager
     def log(self) -> Generator[None, None, None]:
+        """context manager to log method calls
```
dstansby, line 81:
```suggestion
        """Context manager to log method calls
```

src/zarr/storage/memory.py
```diff
@@ -14,9 +14,25 @@
     from zarr.core.common import AccessModeLiteral
 
 
-# TODO: this store could easily be extended to wrap any MutableMapping store from v2
```
dstansby, line 17: Should this TODO stay?

src/zarr/storage/memory.py
```diff
@@ -126,18 +142,6 @@ async def list_prefix(self, prefix: str) -> AsyncGenerator[str, None]:
                 yield key.removeprefix(prefix)
 
     async def list_dir(self, prefix: str) -> AsyncGenerator[str, None]:
-        """
```
dstansby, line 129: Why delete this docstring?

src/zarr/storage/remote.py
```diff
@@ -25,6 +25,30 @@
 
 
 class RemoteStore(Store):
+    """
+    A remote Store based on FSSpec
+
+    Parameters
+    ----------
+    fs: AsyncFileSystem
```
dstansby, line 33:
```suggestion
    fs : AsyncFileSystem
```

src/zarr/storage/remote.py
```diff
@@ -25,6 +25,30 @@
...
+        The Async FSSpec filesystem to use with this store.
+    mode: AccessModeLiteral
```
dstansby, line 35:
```suggestion
    mode : AccessModeLiteral
```

src/zarr/storage/remote.py
```diff
@@ -25,6 +25,30 @@
...
+        The access mode to use.
+    path: str
```
dstansby, line 37:
```suggestion
    path : str
```

src/zarr/storage/remote.py
```diff
@@ -25,6 +25,30 @@
...
+        The root path of the store.
+    allowed_exceptions: tuple[type[Exception], ...]
```
dstansby, line 39:
```suggestion
    allowed_exceptions : tuple[type[Exception], ...]
```

src/zarr/storage/remote.py
```diff
@@ -25,6 +25,30 @@
...
+        When fetching data, these cases will be deemed to correspond to missing
```
dstansby, line 40: This sentence finished halfway 😄

dstansby (changes requested):
I've left some comments - one big one, is why so many docstrings deleted? Is it because they're inherited? If so, would be good to include a `# docstring inherited` comment where they are inherited so others don't get confused in the future.

src/zarr/storage/local.py
jhamman: Yes I did :)

These are now inherited from the base class. The only reason to redefine them is if extra context or notes on specific behavior is needed.

src/zarr/storage/memory.py
jhamman: In a recent conversation, we decided not to do this. So this removal is reflecting that decision.

src/zarr/abc/store.py
jhamman: I decided not to do that here because its just this one method. If we have a docstring template tool in the future, I think it would be great to bring that to bear here (and even more so on the Group/Array classes).

jhamman:
@dstansby - ready for another review here.

dstansby (changes requested):
I think you can get rid of the custom docstring inheritance code (see my inline comment). Where a docstring is inherited though, please could you leave a `# docstring inherited` comment under the signature, to signal to others that come by later that a fresh docstring doesn't need writing.

dstansby (approved):
Good for me - might be worth getting someone more familiar with v3 to look over the docstring contents?

dstansby:
:shipit:

PR #2286, opened 2024-10-01
base64 encode fill value for some dtypes with `zarr_format=2`

### LINKED ISSUE #2282: Fix fill_value encoding for fixed-width string / binary v2 array
### Zarr version

v3

### Numcodecs version

na

### Python Version

na

### Operating System

na

### Installation

na

### Description

In https://zarr-specs.readthedocs.io/en/latest/v2/v2.0.html#fill-value-encoding, the zarr-v2 spec says that `fill_value` must be base64 encoded for some dtypes:

> If an array has a fixed length byte string data type (e.g., "|S12"), or a structured data type, and if the fill value is not null, then the fill value MUST be encoded as an ASCII string using the standard Base64 alphabet.

### Steps to reproduce

Given this

```
    store = zarr.storage.MemoryStore(mode="w")
    g = zarr.group(store=store, zarr_format=2)
    g.create_array(
        name="foo",
        shape=(3,),
        dtype="|S4",
        fill_value=b"X",
    )
```

we produce metadata like this currently:

```
 {'chunks': [3],
 'compressor': None,
 'dimension_separator': '.',
 'dtype': '|S4',
 'fill_value': [88],
 'filters': None,
 'order': 'C',
 'shape': [3],
 'zarr_format': 2}
```

It should be something like

```
{'chunks': [3],
 'compressor': {'blocksize': 0,
                'clevel': 5,
                'cname': 'lz4',
                'id': 'blosc',
                'shuffle': 1},
 'dtype': '|S4',
 'fill_value': 'WA==',
 'filters': None,
 'order': 'C',
 'shape': [3],
 'zarr_format': 2}
```

Note that `WA==` is the ascii representation of the base64encoded value `X`.

### Additional output

_No response_

TomAugspurger:
Closes #2282

Needs more tests.


src/zarr/core/metadata/v2.py
```diff
@@ -31,7 +32,7 @@ class ArrayV2Metadata(ArrayMetadata):
     shape: ChunkCoords
     chunk_grid: RegularChunkGrid
     data_type: np.dtype[Any]
-    fill_value: None | int | float = 0
+    fill_value: None | int | float | str | bytes = 0
```
TomAugspurger, line 35: Need to confirm the expectation here: is `ArrayV2Metadata` supposed to be the in-memory version of `fill_value` (e.g. `b"X"`) or the on-disk version (e.g. `"WA=="`). Check other dtypes.

TomAugspurger: Based on the handling of `complex`-type `fill_value`, I think this is supposed to be the "in-memory" version rather than the on-disk version. So I think `bytes` should be allowed here.

TomAugspurger:
This should be ready to go.

d-v-b:
were we not doing this in the v2 codebase? I couldn't find code for it.

TomAugspurger:
I found it at https://github.com/zarr-developers/zarr-python/blob/6b9ab9e29f526986c5c7519dd4063546d5a81d14/zarr/meta.py#L215 (and borrowed the implementation).

src/zarr/core/metadata/v2.py
```diff
@@ -140,6 +141,22 @@ def from_dict(cls, data: dict[str, Any]) -> ArrayV2Metadata:
...
+        dtype = parse_dtype(_data["dtype"])
+
+        if dtype.kind in "SV":
+            fill_value_encoded = _data.get("fill_value")
+            if fill_value_encoded is not None:
+                if dtype.kind == "S":
+                    try:
+                        fill_value = base64.standard_b64decode(fill_value_encoded)
+                        _data["fill_value"] = fill_value
+                    except Exception:
```
d-v-b, line 153: is it possible to limit the exceptions we catch here to only those that could be raised by `standard_b64decode`?

TomAugspurger: Yes, this came from the old implementation.

While writing a test, I found that we just raise an exception later on in `parse_fill_value` so AFAICT there's no value in trying to suppress exceptions here.

jakirkham:
Thanks Tom! 🙏

JFYI we have [a `Base64` codec in Numcodecs]( https://github.com/zarr-developers/numcodecs/blob/71fcab086ba3b8efb4c886fbaba22d9dcc55df62/numcodecs/base64.py#L7 ), perhaps it is worth using to align on behavior

rabernat (approved):

TomAugspurger:
Thanks John. I think this version (rather than numcodec's) is OK. It ends up using the same actual implementation of `base64.standard_b64encode` but it's slightly more direct with the calls directly in the serialization code in this PR.

PR #2290, opened 2024-10-02
Add BaseCodec to the docs

dstansby:
This renames `_Codec` to `CodecBase`, so that it gets documented and linked from e.g. [`ArrayArrayCodec`](https://zarr.readthedocs.io/en/v3/_autoapi/zarr/abc/codec/index.html#zarr.abc.codec.ArrayArrayCodec), which is currently missing a link to it.

I think in general it's nicer to document the full class heirarchy, but put warnings where we don't expect users to use specific classes. This fixes some broken links, a small part of https://github.com/zarr-developers/zarr-python/pull/2291

### DIFF
--- a/src/zarr/abc/codec.py
+++ b/src/zarr/abc/codec.py
@@ -20,6 +20,7 @@ if TYPE_CHECKING:
     from zarr.core.indexing import SelectorTuple
 
 __all__ = [
+    "BaseCodec",
     "ArrayArrayCodec",
     "ArrayBytesCodec",
     "ArrayBytesCodecPartialDecodeMixin",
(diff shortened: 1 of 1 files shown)


src/zarr/abc/codec.py
```diff
@@ -34,11 +35,15 @@
 CodecOutput = TypeVar("CodecOutput", bound=NDBuffer | Buffer)
 
 
-class _Codec(Metadata, Generic[CodecInput, CodecOutput]):
+class CodecBase(Metadata, Generic[CodecInput, CodecOutput]):
```
d-v-b, line 38: Feel free to ignore this suggestion, but somehow the name `BaseCodec` seems a bit more natural than `CodecBase`. Maybe it seems a bit more consistent with names like `RemoteStore` which follow the `<Qualification><QualifiedEntity>` schema?

dstansby: 👍 I think that's a good suggestion, changed

d-v-b (approved):

Issue #2289, opened 2024-10-02
Links in function signatures not being resolved


dstansby:
In the v3 docs, there are some internal references that aren't being linked in function signatures. For example, in the [listing for `zarr.Array`](https://zarr.readthedocs.io/en/v3/_autoapi/zarr/index.html#zarr.Array), there should be a link for `zarr.storage.StoreLike` but there isn't. This is despite there being an [entry for `zarr.storage.StoreLike`](https://zarr.readthedocs.io/en/v3/_autoapi/zarr/storage/index.html#zarr.storage.StoreLike).

dstansby:
`StoreLike` is showing up fine in the inventory, so I think the issue must be sphinx(-autoapi??) not inserting a cross link properly.

```bash
% python -m sphinx.ext.intersphinx _build/html/objects.inv | grep "StoreLike"
    zarr.storage.StoreLike                                                           : _autoapi/zarr/storage/index.html#zarr.storage.StoreLike
    zarr.storage.common.StoreLike                                                    : _autoapi/zarr/storage/common/index.html#zarr.storage.common.StoreLike
```

dstansby:
In `docs/_autoapi/zarr/storage/index.rst` `StoreLike` is being documented as

```rst
.. py:data:: StoreLike
```

dstansby:
I think we're running into https://github.com/sphinx-doc/sphinx/issues/10974 here

jhamman:
@dstansby - you are probably the expert here. Do we see a path to fixing this?

dstansby:
Not without https://github.com/sphinx-doc/sphinx/issues/10785 being fixed, which doesn't seem like it's going to happen soon. So definitely don't block v3 on this!

d-v-b:
IMO the lack of this feature is a massive reason to move to some kind of documentation solution that works for type-annotated python projects. Since the linked sphinx issue is very stalled, I am wondering if we can get this functionality from mkdocs-material, which gives extra urgency to #3118

d-v-b:
to be more specific: the autodoc plugin is the piece of the toolchain that's missing support for types. sphinx itself supports them.

dstansby:
Would using `mkdocs` fix this? I seem to remember that I couldn't get this to work with `mkdocs` over at https://ome-zarr-models-py.readthedocs.io.

d-v-b:
[...]

PR #2291, opened 2024-10-02
Error on broken links in the docs

dstansby:
This errors on any broken links in the documentation, which currently is quite a lot! It will take a while to fix all these, but I think it's worth having this PR open in draft so we can see how we're doing.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -53,6 +53,8 @@ extensions = [
     "sphinx_design",
 ]
 
+nitpicky = True
+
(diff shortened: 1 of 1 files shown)


dstansby:
If anyone's reading this, it's currently blocked on https://github.com/zarr-developers/zarr-python/issues/2289 which as more info on why we have broken links.

dstansby:
I'm guessing after the doc refactor this is no longer relevant

PR #2293, opened 2024-10-02
Remove acknowledgements page from docs

dstansby:
This page is out of date - unless someone wants to commit to updating it regularly, I think it's better to remove it than have an out of date page that doesn't recognise everyone who has contributed to zarr-python.


d-v-b (approved):

d-v-b:
IMO it would make more sense to acknowledge the organizations that have funded zarr development.

jhamman (approved):

joshmoore:
:+1: for this being out of date (and *especially* for the v3 branch which is essentially a rewrite). A half-way house might be to make use of https://allcontributors.org/ to help manage a file outside of the docs/ hierarchy.

Issue #2296, opened 2024-10-02
`GroupMetadata` and `ArrayV2Metadata` should ignore extra keys for zarr v2 metadata


TomAugspurger:
### Zarr version

v3

### Numcodecs version

na

### Python Version

na

### Operating System

na

### Installation

na

### Description

Some Zarr v2 implementations store additional fields at the top level of the metadata files, e.g. [nczarr](https://www.google.com/url?sa=t&source=web&rct=j&opi=89978449&url=https://docs.unidata.ucar.edu/nug/current/nczarr_head.html&ved=2ahUKEwjK2eaL-u-IAxXktokEHSqUMcoQFnoECAgQAQ&usg=AOvVaw01ivrxQR6t1Bv6dmAlw8cb). zarr-python 3.x currently can't read these files because the deserialization methods pass all the keys to the `GroupMetadata` and `ArrayV2Metadata` constructors.

### Steps to reproduce

```python
In [10]: GroupMetadata.from_dict({"zarr_format": 2, "attributes": {"key": "value"}, "extra": "value"})
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[10], line 1
----> 1 GroupMetadata.from_dict({"zarr_format": 2, "attributes": {"key": "value"}, "extra": "value"})

File ~/gh/zarr-developers/zarr-python/src/zarr/core/group.py:119, in GroupMetadata.from_dict(cls, data)
    116 @classmethod
    117 def from_dict(cls, data: dict[str, Any]) -> GroupMetadata:
    118     assert data.pop("node_type", None) in ("group", None)
--> 119     return cls(**data)

TypeError: GroupMetadata.__init__() got an unexpected keyword argument 'extra'
```

### Additional output

We can filter out unexpected keys before calling the constructors. We should decide what action to take when we encounter an unknown key (warning with the python logging module, a Python warning, and maybe some sort of config options to control the behavior).
 
For Zarr v3, I think the expectation is that extra keys are allowed, but they must be objects with a `must_understand` field.

PR #2297, opened 2024-10-02
Ignore extra keys in v2 metadata

TomAugspurger:
Ignore unexpected keys in Zarr V2 metadata, to enable reading zarr files written by other systems, which might store additional data in the top level of the `.zgroup` and `.zarray` files`

Closes https://github.com/zarr-developers/zarr-python/issues/2296

### DIFF
--- a/src/zarr/core/group.py
+++ b/src/zarr/core/group.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import asyncio
 import json
 import logging
-from dataclasses import asdict, dataclass, field, replace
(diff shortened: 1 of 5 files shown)


d-v-b:
I checked and the [spec](https://github.com/zarr-developers/zarr-specs/blob/c2f2966c79a4ec81e6c37a0d7c88d503df681ee7/docs/v2/v2.0.rst?plain=1#L91-L92) says we SHOULD be doing this :)

jhamman (approved):
Looks good to me!

Issue #2309, opened 2024-10-08
Zarr v3.0.0a7 fail to open group dataset


Czaki:
### Zarr version

3.0.0a7

### Numcodecs version

0.13.0

### Python Version

3.12

### Operating System

All

### Installation

pip install --pre

### Description

When open zarr dataset using `image = zarr.open(store=path)` ends with error `ValueError: Invalid value. Expected 'array'. Got group.` 

This guard is expecting key error, not value error:

https://github.com/zarr-developers/zarr-python/blob/c258b275d717106be6cdbe403e585fe03f67961e/src/zarr/api/asynchronous.py#L248-L251

```python
napari_builtins/_tests/test_io.py:121:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
napari_builtins/io/_read.py:225: in magic_imread
    image, zarr_shape = read_zarr_dataset(filename)
napari_builtins/io/_read.py:140: in read_zarr_dataset
    image = zarr.open(store=path)
/tmp/.tox/py312-linux-pyqt5-no_cov/lib/python3.12/site-packages/zarr/_compat.py:43: in inner_f
    return f(*args, **kwargs)
/tmp/.tox/py312-linux-pyqt5-no_cov/lib/python3.12/site-packages/zarr/api/synchronous.py:77: in open
    obj = sync(
/tmp/.tox/py312-linux-pyqt5-no_cov/lib/python3.12/site-packages/zarr/core/sync.py:91: in sync
    raise return_result
/tmp/.tox/py312-linux-pyqt5-no_cov/lib/python3.12/site-packages/zarr/core/sync.py:50: in _runner
    return await coro
/tmp/.tox/py312-linux-pyqt5-no_cov/lib/python3.12/site-packages/zarr/api/asynchronous.py:249: in open
    return await open_array(store=store_path, zarr_format=zarr_format, **kwargs)
/tmp/.tox/py312-linux-pyqt5-no_cov/lib/python3.12/site-packages/zarr/api/asynchronous.py:966: in open_array
    return await AsyncArray.open(store_path, zarr_format=zarr_format)
[...]

PR #2310, opened 2024-10-08
zarr.open should fall back to opening a group

TomAugspurger:
Closes https://github.com/zarr-developers/zarr-python/issues/2309

I'm not happy with this implementation. I broke / fixed this in my consolidated metadata PR as well.

Relying on the exception type from `open_array` to decide whether to fall back to opening a group seems pretty fragile. I might spend a few minutes trying out an alternative way of doing things that'll be a bit more robust (potentially at the cost of some unnecessary I/O; but I think that's OK for `open` since `open_group` and `open_array` offer more direct ways of doing things).

TODO:


src/zarr/api/asynchronous.py
```diff
@@ -247,7 +247,10 @@ async def open(
 
     try:
         return await open_array(store=store_path, zarr_format=zarr_format, **kwargs)
-    except KeyError:
+    except (KeyError, ValueError):
+        # KeyError for a missing key
+        # ValueError for failing to parse node metadata as an array when it's
```
d-v-b, line 252: maybe we should create a specific exception for this case, since "valid document but the wrong node type" is different from "invalid document"

TomAugspurger:
>  I might spend a few minutes trying out an alternative way of doing things that'll be a bit more robust 

https://github.com/TomAugspurger/zarr-python/blob/tom/fix/batched-io/src/zarr/core/metadata/_io.py has the start of that. I like the idea of using `match` statements and an explicit list of files to read. There's still a decent chunk of work to do (figure out which cases we know we files to read and error if any are missing, then discover the actual `zarr_format` and `node_type` based on what we read and parse into the correct objects, and finally updating `open` and hopefully `open_array` and `open_group` to use that or something similar). I don't think I'll pursue this now.

tests/v3/test_api.py
```diff
@@ -921,3 +923,23 @@ def test_open_group_positional_args_deprecated() -> None:
...
+
+
+async def test_open_falls_back_to_open_group_async() -> None:
+    # https://github.com/zarr-developers/zarr-python/issues/2309
+    store = MemoryStore(mode="w")
+    await zarr.api.asynchronous.open_group(store, attributes={"key": "value"})
+
+    group = await zarr.api.asynchronous.open(store=store)
+    assert isinstance(group, zarr.core.group.AsyncGroup)
+    assert group.attrs == {"key": "value"}
```
jhamman, line 945: Do we have test coverage for the `MetadataValidationError` path? If not, could be good to add that here with by trying `open` with a different `zarr_format`

jhamman (approved):
One test suggestion but otherwise this looks good to me.

tests/v3/test_api.py
TomAugspurger: Added a test for that. Looks like `Array.open` / `Group.open` don't use the `parse_zar_format` (which are version specific now) so updated their implementations to raise the `MetadataValidationError`.

PR #2311, opened 2024-10-08
fix: add get(key, default) method to Group APIs

jhamman:
Some downstream applications (xarray.DataTree) were using `Group.get`. This PR adds that to the Group API for backward compatibility.  

TODO:

### DIFF
--- a/src/zarr/core/group.py
+++ b/src/zarr/core/group.py
@@ -4,7 +4,7 @@ import asyncio
 import json
 import logging
 from dataclasses import asdict, dataclass, field, fields, replace
-from typing import TYPE_CHECKING, Literal, cast, overload
+from typing import TYPE_CHECKING, Literal, TypeVar, cast, overload
 
 import numpy as np
 import numpy.typing as npt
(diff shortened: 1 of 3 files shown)


src/zarr/core/group.py
```diff
@@ -290,6 +292,28 @@ async def delitem(self, key: str) -> None:
         else:
             raise ValueError(f"unexpected zarr_format: {self.metadata.zarr_format}")
 
+    async def get(
+        self, key: str, default: DefaultT | None = None
```
TomAugspurger, line 296: I don't really know how to type this properly, but IMO as a user I'd appreciate the ability to know that `.get(key, default)` is still `AsyncArray | AsyncGroup` (or None I guess). So maybe we have `DefaultT` also be bound to `AsyncArray, AsyncGroup`?

That said

```python

foo : dict[str, int] = {}
reveal_type(foo.get("a", "b"))
```

gives

```
src/zarr/core/metadata/v3.py:555: note: Revealed type is "Union[builtins.int, builtins.str]"
```

so nevermind :/

PR #2312, opened 2024-10-08
fix(RemoteStore): avoid listing all objects in remote store in empty() method

jhamman:
Some initial exploration with the RemoteStore found that the `empty()` check was taking a very long time. This sped it up considerably. 

TODO:

### DIFF
--- a/src/zarr/storage/remote.py
+++ b/src/zarr/storage/remote.py
@@ -94,7 +94,11 @@ class RemoteStore(Store):
             pass
 
     async def empty(self) -> bool:
-        return not await self.fs._find(self.path, withdirs=True)
+        async for _path, _dirs, files in self.fs._walk(self.path):
+            # stop once a file is found
+            if files:
+                return False
+        return True
 
     def with_mode(self, mode: AccessModeLiteral) -> Self:
         return type(self)(
--- a/src/zarr/storage/remote.py
+++ b/src/zarr/storage/remote.py
@@ -94,11 +94,9 @@ class RemoteStore(Store):
             pass
 
     async def empty(self) -> bool:
-        async for _path, _dirs, files in self.fs._walk(self.path):
-            # stop once a file is found
-            if files:
-                return False
-        return True
+        # TODO: it would be nice if we didn't have to list all keys here
+        # it should be possible to stop after the first key is discovered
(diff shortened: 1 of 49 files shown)


src/zarr/storage/remote.py
```diff
@@ -94,7 +94,11 @@ async def clear(self) -> None:
             pass
 
     async def empty(self) -> bool:
-        return not await self.fs._find(self.path, withdirs=True)
+        async for _path, _dirs, files in self.fs._walk(self.path):
+            # stop once a file is found
+            if files:
```
martindurant, line 99: Do included directories count as content? The original implied yes.

src/zarr/storage/remote.py
```diff
@@ -94,7 +94,11 @@ async def clear(self) -> None:
             pass
 
     async def empty(self) -> bool:
-        return not await self.fs._find(self.path, withdirs=True)
+        async for _path, _dirs, files in self.fs._walk(self.path):
```
martindurant, line 97: It cannot be guaranteed that this doesn't do a find() too... Would `_ls()` be enough?

There was talk of making `ls()` and friends in fsspec be generators, but not yet.

src/zarr/storage/remote.py
jhamman: We only check for files in the `LocalStore`:

https://github.com/zarr-developers/zarr-python/blob/5a134bf2c1c5e6655edd429ca554aa9699327463/src/zarr/storage/local.py#L101-L111

src/zarr/storage/remote.py
jhamman: gotcha... This was an improvement using gcsfs but I can see how it wouldn't be for others. I'll take your lead on the best path forward here. All we need to know is if there are any files in the store (ignoring empty directories).

martindurant: Given the very many files that can in theory be in a zarr hierarchy, maybe we can keep with this strategy for now. Having said that, in the zarr case, we do know which subdir ought to have the fewest items in it, right?

d-v-b (approved):
thanks for this. perennial question: what kind of performance monitoring should we have in place to flag a regression here (or anywhere)?

PR #2318, opened 2024-10-09
Unnecessary lambda expression

DimitriPapadopoulos:
A lambda that calls a function without modifying any of its parameters is unnecessary. Python functions are first-class objects and can be passed around in the same way as the resulting lambda. It is recommended to remove the lambda and use the function directly.

TODO:

### DIFF
--- a/src/zarr/testing/strategies.py
+++ b/src/zarr/testing/strategies.py
@@ -60,7 +60,7 @@ node_names = st.text(zarr_key_chars, min_size=1).filter(
 )
 array_names = node_names
 attrs = st.none() | st.dictionaries(_attr_keys, _attr_values)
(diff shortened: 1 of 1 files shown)


src/zarr/testing/strategies.py
```diff
@@ -60,7 +60,7 @@ def v2_dtypes() -> st.SearchStrategy[np.dtype]:
 )
 array_names = node_names
 attrs = st.none() | st.dictionaries(_attr_keys, _attr_values)
-keys = st.lists(node_names, min_size=1).map(lambda x: "/".join(x))
+keys = st.lists(node_names, min_size=1).map("/".join)
```
jakirkham, line 63: Cool! 🤩

Did not even realize this was valid in Python. Other languages certainly have this concept. Will keep this in mind for the future

PR #2332, opened 2024-10-11
Add guide section to top level of docs

dstansby:
I'm not sure a section called "Consolidated metadata" belongs at the top level:

Screenshot 2024-10-11 at 08 59 17

So I created a new "Guide" section for it:

Screenshot 2024-10-11 at 09 08 08

 Open to suggestions for other names for this section, but I think we do need this section as the content is distinct to other sections of the docs.


joshmoore:
Big :+1: (I'm a general fan of [the 4 kinds of docs](https://www.writethedocs.org/videos/eu/2017/the-four-kinds-of-documentation-and-why-you-need-to-understand-what-they-are-daniele-procida/))

jhamman (approved):

Issue #2302, opened 2024-10-07
[v3] `zarr-python` fails to decode sharded array written by other implementations.


zoj613:
### Zarr version

3.0.0a5

### Numcodecs version

0.13.0

### Python Version

3.10.14

### Operating System

Linux

### Installation

pip install zarr==3.0.0a5

### Description

It appears as though the implementation for partial decoding chunks written using the sharding indexed codec is buggy. As in, the implementation fails to read a sharding-encoded array created by another implementation. The same dataset can be read without problem by the [`zarrita`](https://github.com/scalableminds/zarrita) project.

### Steps to reproduce

[shardingdata.zip](https://github.com/user-attachments/files/17300875/shardingdata.zip)

I have attached a zip file containing the filesystem store `testdata.zarr` used to reproduce the bug.

```python
import numpy
import zarr

store = zarr.store.LocalStore("./testdata.zarr", mode="r")
a = zarr.Array.open(store.root / "some/group/another") # reads the sharded array with path /some/group/another.
a[:20, 10:11, :].shape  # this fails with a long stacktrace.
# ValueError: When changing to a larger dtype, its size must be a divisor of the total size in bytes of the last axis of the array.

a2 = zarr.Array.open(store.root / "some/group/name") # reads a non-sharded array from same store with path /some/group/name.
a2[:20, 10:11, :].shape  # this works
```

If I use `zarrita` to read the problematic array, I get no errors:
```python
import zarrita

store2 = zarrita.LocalStore('./testdata.zarr')
a3 = zarrita.Array.open(store2 / "some/group/another")
a3[:20, 10:11, :].shape  # decodes just fine and prints (20, 1, 50)
```

### Additional output

Here is the metadata of the "problematic" array:
```json
{
"zarr_format":3,
"shape":[100,100,50],
"node_type":"array",
"data_type":"float64",
"codecs":
	[
	{"name":"sharding_indexed",
	 "configuration":
		{"chunk_shape":[5,3,5],
		 "index_location":"start",
		 "index_codecs":[{"name":"bytes","configuration":{"endian":"big"}},{"name":"crc32c"}],
		 "codecs":[{"name":"transpose","configuration":{"order":[2,0,1]}},
	{"name":"bytes","configuration":{"endian":"little"}}]}}],
"fill_value":"-Infinity",
"chunk_grid":{"name":"regular","configuration":{"chunk_shape":[10,15,20]}},
[...]

PR #2336, opened 2024-10-11
sharding with empty inner chunk and index location start

brokkoli71:
In sharding, empty inner chunks get marked by setting `nbytes` and `offset` to `-1` in the index array. When using `index_location="start"` the size of the index array will be added to the offset.

This PR will disable incrementing the `offset` for empty inner chunks (thus, it still contains `-1`)

may contribute to #2302


jhamman:
@brokkoli71 / @normanrz - can we close this now.

normanrz (approved):

PR #2335, opened 2024-10-11
V3 main sync w/ merge

jhamman:
This was not as bad as I thought it would be:

```
git checkout v3
git merge upstream/main
... # resolve conflicts
git commit
```

Supersedes https://github.com/zarr-developers/zarr-python/pull/2330 
Toward https://github.com/zarr-developers/zarr-python/issues/2308

PR #2339, opened 2024-10-11
Revert "V3 main sync w/ merge (#2335)"

jhamman:
This reverts commit ce5f53e3482727b7133def725af24576057aa164.

The above merge was supposed to be done without a squash. Reverting and will start again 🤦


joshmoore:
Sounds like that might be worth a force push, no?

jhamman:
I don't think that is needed. Really no harm in keeping the v3 history linear (with a revert commit).

Issue #2308, opened 2024-10-08
Moving v3 development to `main`


jhamman:
Over the past 9 months, we have been working hard on a major update to Zarr-Python (props to everyone who has been working hard on this, its really coming together nicely). This work has been developed on the [`v3`](https://github.com/zarr-developers/zarr-python/tree/v3) branch. Now that we are in the last mile of the 3.0 release (#1777), I propose we transition this work to the `main` branch. Along with this move, I also propose we create a support branch for 2.x so that we can continue to make bug-fix releases for at least the next 6 months. 

@zarr-developers/python-core-devs -- looking for feedback on the proposal and a champion to manage the git ops to make this happen. 

- [x] create support branch ([`support/2.x`](https://github.com/zarr-developers/zarr-python/tree/support/2.x))
- [x] merge `main` into `v3` https://github.com/zarr-developers/zarr-python/pull/2335
- [ ] merge `v3` into `main`

d-v-b:
I like this idea! But I'm far from a git pro, so perhaps I'm not the champion we need.

joshmoore:
* C: Kudos all around :tada:
 * C: :+1: for the move
 * C: :+1: for the `2.x` branch
 * Q: Are you proposing a rename or a merge? Or is that left to be determined?

jhamman:
> Q: Are you proposing a rename or a merge? Or is that left to be determined?

Left to be determined. Open to opinions.

dstansby:
For at least 6 months we're going to need parallel branches, so I don't think we can do a merge

jhamman:
> For at least 6 months we're going to need parallel branches, so I don't think we can do a merge

👍  - what we could do is merge current `main` into `v3` with a merge strategy of `--ours`. I believe that would help us maintain the history that has accumulated on `main` since we started work on `v3`.

dstansby:
👍 sounds good, for anyone brave enough to try 😄

joshmoore:
> for anyone brave enough to try 😄

:smile:

```
    git clone git@github.com:zarr-developers/zarr-python
    cd zarr-python/
    git merge -s ours origin/main
```
runs cleanly cleanly at least. *shrug*

> I believe that would help us maintain the history that has accumulated on `main` since we started work on `v3`.

*nods* though I didn't check what the `follow`-ablility of the files will be after the `v2` module etc. are removed.

jhamman:
[...]

PR #2341, opened 2024-10-11
Move V3 development to main

jhamman:
This PR moves the current `v3` branch to `main`. Following this merge, the following actions should be taken:

1. reset the default branch in this repo to `main`
2. remove the `v3` branch (this is an important signal to downstream testers that are currently using `@v3`
3. update our github actions to trigger off `main` instead of `v3`
4. update the target branch for PRs pointing at the `support/2.x` branch

closes https://github.com/zarr-developers/zarr-python/issues/2308

@zarr-developers/python-core-devs -- looking for two 👍 here.


d-v-b (approved):
image

🚀

dstansby:
I'd like to advocate for branch named `v2` instead of `support/2.x` - `v2` is shorter to type, and mirrors the `v3` branch we've had going on for a while.

jhamman:
@dstansby - the name of the support branch is somewhat independent of this PR but the reason I like having the `support` name in there is that it clearly defines the purpose of the branch -- supporting v2.

dstansby:
I don't think it's orthogonal, because this PR proposes to switch `main` from a development branch for v2 to v3, and I think we should have consensus on the details of what is happening to v2 development (including branch name) before this gets merged. So please can we hold off this until https://github.com/zarr-developers/zarr-python/issues/2345 is resolved?

Issue #2317, opened 2024-10-09
Installation fails on Python 3.13


veenstrajelmer:
### Zarr version

latest

### Numcodecs version

latest

### Python Version

3.13

### Operating System

Windows

### Installation

pip install zarr

### Description

Installation fails

### Steps to reproduce

`pip install zarr` in a Python 3.13 environment

### Additional output

Also reported at numcodecs: https://github.com/zarr-developers/numcodecs/issues/593

veenstrajelmer:
The entire traceback is included in the linked numcodecs issue. I did not test beyond numcodecs, so I am not sure whether there will be a zarr-specific problem after this is resolved. However, since zarr cannot be installed because of this I thought it would be useful to have a zarr issue also. But feel free to close as won't do if there is indeed nothing zarr-specific in this installation failure.

veenstrajelmer:
@TomAugspurger After the release of numcodecs 0.13.1, `pip install zarr` works again. I would like to suggest to add python 3.13 to the ci and supported versions, but feel free to close this issue you think otherwise.

TomAugspurger:
Thanks. Let's keep this open for those tasks.

jhamman:
I think we can close this following #2344.

PR #2344, opened 2024-10-12
Add Python 3.13, drop Python 3.10

dstansby:
As per SPEC 0 we can drop Python 3.10, and this also adds tests for Python 3.13. Fixes https://github.com/zarr-developers/zarr-python/issues/2317

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -27,7 +27,7 @@ body:
     attributes:
       label: Python Version
       description: Version of Python interpreter
-      placeholder: 3.10, 3.11, 3.12 etc.
+      placeholder: 3.11, 3.12, 3.13 etc.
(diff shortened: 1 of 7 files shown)


jhamman:
would you mind switching the base of this PR to the `support/2.x` branch.

jhamman (approved):
Looks good to me. I'm impressed things worked on 3.13 without any changes.

joshmoore:
A bit late to the party but perhaps more for future cases of applying SPEC 0, I'd vote for always bumping the minor version when dropping a version of Python, i.e. I would have been for this being 2.19.0. The primary reason is that *if* something else occurs (critical data loss issue, etc.) it gives us the chance to fix that bug without forcing developers to bump their Python version.

joshmoore:
Very fair point. Thoughts on having a third type of compatibility since it stretches the definition of "API compatibility"? Alternatively, it could be a note there.

PR #2348, opened 2024-10-13
fix(remotestore): raise error if path includes scheme

### LINKED ISSUE #2342: RemoteStore doesn't strip protocol correctly, breaking list operations
I've discovered _what I thought was_ a pretty major gap in our support for remote storage. Instead it turns out to be a weird behavior around prefix and protocol stripping with path urls.

Unfortunately there is no way to reproduce this afaict without write access to an actual s3 bucket because RemoteStore requires an async filesystem... ~Which makes me think we are currently not testing RemoteStore at all? 🤔~ Edit: not true. 

```python
import s3fs
import zarr
s3 = s3fs.S3FileSystem()

# replace with a bucket you can write to
target_url = "s3://icechunk-test/ryan/zarr3-tests/groups/1"
store = zarr.storage.RemoteStore(s3, mode="w", path=target_url)
# create a group
g = zarr.group(store=store, zarr_version=3)
# create a child
a = g.create("foo", shape=10, dtype='i4')

# try to discover children
print(list(g))
print(g.members())
[i for i in g.arrays()]
```

All of these return an empty list, along with the warning
```
Object at icechunk-test/ryan/zarr3-tests/groups/1/foo is not recognized as a component of a Zarr hierarchy.
Object at icechunk-test/ryan/zarr3-tests/groups/1/zarr.json is not recognized as a component of a Zarr hierarchy.
```

Unfortunately, this means that Xarray can't discover the arrays in a group and so can't open any RemoteStore datasets with zarr version >= 3.

https://github.com/pydata/xarray/blob/70a2a55afb4a73a4d5027506ea87f918949e2b7c/xarray/backends/zarr.py#L627C61-L627C85

I'm on version 3.0.0a8.dev12+gff530c36

jhamman:
closes #2342

This PR makes a few important changes to the `RemoteStore`:

1. Raise an error if a the provided path has a scheme. 
2. Raise a warning if the Filesystem was not created with `asynchronous=True`

TODO:


src/zarr/storage/remote.py
```diff
@@ -59,6 +90,14 @@ def __init__(
 
         if not self.fs.async_impl:
             raise TypeError("Filesystem needs to support async operations.")
+        if not self.fs.asynchronous:
+            warnings.warn(
+                f"fs ({fs}) was not created with `asynchronous=True`, this may lead to surprising behavior",
+                stacklevel=2,
+            )
```
jhamman, line 97: @martindurant - I'm not sure we actually care about this so I'm open to removing it. Without fully understanding how the internals of fsspec work, I'm curious if you think we should encourage folks to give us filesystems with `asynchronous=True`.

tests/v3/test_store/test_core.py
```diff
@@ -39,10 +39,7 @@ async def test_make_store_path(tmpdir: str) -> None:
 
 
 async def test_make_store_path_fsspec(monkeypatch) -> None:
-    import fsspec.implementations.memory
-
-    monkeypatch.setattr(fsspec.implementations.memory.MemoryFileSystem, "async_impl", True)
-    store_path = await make_store_path("memory://")
+    store_path = await make_store_path("http://foo.com/bar")
```
jhamman, line 42: using a real path here turned up a real bug in fsspec (reported in https://github.com/fsspec/filesystem_spec/issues/1722 and handled in `from_url` for now).

jhamman:
pre-commit.ci autofix

src/zarr/storage/remote.py
martindurant: - not all FSs are async-capable.
- *if* async stores are to be called on awaitable methods, asynchronous=True is the right thing. For some implementations, if makes less difference than it used to (or None).
- I don't know if there's a use case for having fsspec use a separate thread and loop and calling the FS with blocking methods.

jhamman: > * not all FSs are async-capable.

👍 

> * _if_ async stores are to be called on awaitable methods, asynchronous=True is the right thing. For some implementations, if makes less difference than it used to (or None).

By design, we only use awaitable methods. The assumption is that we will never need to use Fsspec's synchronization. 

> * I don't know if there's a use case for having fsspec use a separate thread and loop and calling the FS with blocking methods.

My operating assumption is that we want synchronization to be handled by the highest part of the stack (typically zarr or even something higher). Since Zarr has its own thread/loop in these cases, I hope we can avoid mixing with Fsspec's.

src/zarr/storage/remote.py
```diff
@@ -131,13 +159,23 @@ def from_url(
         -------
         RemoteStore
         """
-        fs, path = fsspec.url_to_fs(url, **storage_options)
+        opts = storage_options or {}
+        opts = {"asynchronous": True, "use_listings_cache": False, **opts}
```
jhamman, line 163: these are two new defaults.

martindurant: Why use_listings_cache=False?

jhamman: I'll be posting an issue here shortly. We could remove this in favor of `refresh=True` in the `ls/find` calls.

jhamman: After looking at this more, I think we can revert this change in favor of #2356.

tests/v3/test_store/test_core.py
jhamman: Noting that this bug is the "expected" behavior in fsspec. Just not what we wanted.

jhamman:
@martindurant - this is ready for another review.

src/zarr/storage/remote.py
```diff
@@ -71,6 +90,15 @@ def __init__(
 
         if not self.fs.async_impl:
             raise TypeError("Filesystem needs to support async operations.")
+        if not self.fs.asynchronous:
+            warnings.warn(
+                f"fs ({fs}) was not created with `asynchronous=True`, this may lead to surprising behavior",
+                stacklevel=2,
+            )
+        if "://" in path and not path.startswith("http"):
```
martindurant, line 98: Could consider also "::", since that is used in chained fsspec URLs, but should not normally end up passed to the FS. However, it is not as special as "://", so I'm ok not to make a special case for it.

src/zarr/storage/remote.py
```diff
@@ -130,13 +158,23 @@ def from_url(
...
+        opts = storage_options or {}
+        opts = {"asynchronous": True, **opts}
+
+        fs, path = fsspec.url_to_fs(url, **opts)
+
+        # fsspec is not consistent about removing the scheme from the path, so check and strip it here
+        # https://github.com/fsspec/filesystem_spec/issues/1722
+        if "://" in path and not path.startswith("http"):
+            # `not path.startswith("http")` is a special case for the http filesystem (¯\_(ツ)_/¯)
+            _, path = path.split("://", maxsplit=1)
```
martindurant, line 170:
```suggestion
        path = fs._strip_protocol(path)
```
?

src/zarr/storage/remote.py
```diff
@@ -130,13 +158,23 @@ def from_url(
...
+
         return cls(fs=fs, path=path, mode=mode, allowed_exceptions=allowed_exceptions)
 
     async def clear(self) -> None:
         # docstring inherited
         try:
-            for subpath in await self.fs._find(self.path, withdirs=True):
+            for subpath in await self.fs._find(self.path, withdirs=True, refresh=True):
```
martindurant, line 177: refresh not required following https://github.com/fsspec/s3fs/pull/904 ?

jhamman: thanks!

martindurant (approved):

jhamman:
pre-commit.ci autofix

Issue #2358, opened 2024-10-14
Accessing nested children  via consolidated metadata fails


jhamman:
### Zarr version

3.0.0.beta

### Numcodecs version

0.13

### Python Version

3.11

### Operating System

Mac

### Installation

pip

### Description

In https://github.com/pydata/xarray/pull/9552, I noticed that accessing nested children fails when using consolidated metadata.

### Steps to reproduce

```python
import zarr

store = zarr.storage.MemoryStore(mode='w')

# create hierarchy root + foo/bar
root = zarr.open_group(store=store, attributes={'a': 'b'}, mode='w')
root.create_array('foo/bar', shape=(2, 2), attributes={'d': 4})

# consolidate metadata
out = zarr.consolidate_metadata(store)

out['foo/bar']
```

```python-traceback
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
File [~/Library/CloudStorage/Dropbox/src/zarr-python/src/zarr/core/group.py:670](http://localhost:8888/lab/tree/~/Library/CloudStorage/Dropbox/src/zarr-python/src/zarr/core/group.py#line=669), in AsyncGroup._getitem_consolidated(self, store_path, key, prefix)
    669 try:
--> 670     metadata = self.metadata.consolidated_metadata.metadata[key]
    671 except KeyError as e:
    672     # The Group Metadata has consolidated metadata, but the key
    673     # isn't present. We trust this to mean that the key isn't in
    674     # the hierarchy, and *don't* fall back to checking the store.

KeyError: 'foo[/bar](http://localhost:8888/bar)'

The above exception was the direct cause of the following exception:

KeyError                                  Traceback (most recent call last)
Cell In[20], line 12
      9 # consolidate metadata
     10 out = zarr.consolidate_metadata(store)
---> 12 out['foo[/bar](http://localhost:8888/bar)']

File [~/Library/CloudStorage/Dropbox/src/zarr-python/src/zarr/core/group.py:1330](http://localhost:8888/lab/tree/~/Library/CloudStorage/Dropbox/src/zarr-python/src/zarr/core/group.py#line=1329), in Group.__getitem__(self, path)
   1329 def __getitem__(self, path: str) -> Array | Group:
-> 1330     obj = self._sync(self._async_group.getitem(path))
   1331     if isinstance(obj, AsyncArray):
   1332         return Array(obj)

[...]

PR #2363, opened 2024-10-14
Fixed consolidated Group getitem with multi-part key

TomAugspurger:
This fixes `Group.__getitem__` when indexing with a key
like 'subgroup/array'. The basic idea is to rewrite the indexing
operation as `group['subgroup']['array']` by splitting the key
and doing each operation independently. This is fine for consolidated
metadata which doesn't need to do IO.

There's a complication around unconsolidated metadata, though. What
if we encounter a node where `Group.getitem` returns a sub Group
without consolidated metadata. Then we need to fall back to
non-consolidated metadata. We've written _getitem_consolidated
as a regular (non-async) function so we need to pop back up to
the async caller and have *it* fall back.

Closes https://github.com/zarr-developers/zarr-python/issues/2358


src/zarr/core/group.py
```diff
@@ -97,6 +97,24 @@ def _parse_async_node(
         raise TypeError(f"Unknown node type, got {type(node)}")
 
 
+class _MixedConsolidatedMetadataException(Exception):
```
TomAugspurger, line 100: Let me know if this `MixConsolidatedMetadataException` stuff makes sense. The crux of the issue is that `Group.__getitem__` on a Group with consolidated metadata *can* be IO-free. To make that explicit, we define it as a plain (non-async) function.

But this `group['subgroup']['subarray']` presents a new challenge. What if `group` has consolidated metadata, but `subgroup` *doesn't*? Its consolidated metadata might be `None`, meaning we need to fall back to the non-consolidated metadata. We discover this fact in `AsyncGroup._getitem_consolidated`, which isn't an async function, so it can't call the async non-consolidated getitem.

To break through that tangle, I've added this custom exception. The expectation is that users never see it; we require (through docs & being careful?) that all the callers of `_getitem_consolidated` handle this case by catching the error and falling back if needed.

TomAugspurger: I forgot to mention: this "mixed" case isn't going to be common. Under normal operation, where users call `consolidate_metadata` and don't mutate the group afterwards we consolidate everything and so won't hit this. I believe the only time this can happen through the public API is when users add a new group to group with consolidated metadata, and don't then re-consolidate.

jhamman (approved):
Nice work here @TomAugspurger - The (somewhat messy) handling of mixed consolidated / non-consolidated metadata makes me wonder if we want to revisit supporting it at all. It would be pretty reasonable to ask callers to switch to `consolidated=False` if they go looking for a key outside of the range of the consolidated metadata.

But we can have that conversation another day.

TomAugspurger:
Makes sense. I think it depends on exactly how transparent a cache we want consolidated metadata to be. If it's really supposed to be transparent, then falling back is probably the right thing to do.

Really, the complexity comes from my stubbornness around keeping `_getitem_consolidated` non-async. We *could* make that async and just do `await replace(self, consolidated_metadata=None).getitem(key)` if we detect non-consolidated metadata.

But overall, I think the guarantee we get that `_getitem_consolidated` is actually I/O free makes jumping through hoops worth it.

An alternative to using Exceptions to signal an issue with mixed consolidated metadata, we could return a `Result` or `Option` type from `_getitem_consolidated` that is either the actual result or some object indicating we need to fall back to non-consolidated metadata.  I think that'd be about the same level of complexity.

src/zarr/core/group.py
```diff
@@ -662,18 +691,39 @@ def _getitem_consolidated(
...
+                # we've indexed into an array with group["array/subarray"]. Invalid.
+                raise KeyError(key)
+            try:
+                if metadata.consolidated_metadata is None:
+                    # we've indexed into a group without consolidated metadata.
+                    # Note that the `None` case is different from `metadata={}`
+                    # where we explicitly know we have no children. In the None
+                    # case we have to fall back to non-consolidated metadata.
+                    raise _MixedConsolidatedMetadataException(key)
+                assert metadata.consolidated_metadata is not None
```
d-v-b, line 718: (minor) would it be possible to raise a specific exception here instead of using an assertion?

src/zarr/core/group.py
```diff
@@ -662,18 +691,39 @@ def _getitem_consolidated(
...
-            # isn't present. We trust this to mean that the key isn't in
-            # the hierarchy, and *don't* fall back to checking the store.
-            msg = f"'{key}' not found in consolidated metadata."
-            raise KeyError(msg) from e
+        # we support nested getitems like group/subgroup/array
+        indexers = key.split("/")
+        indexers.reverse()
+        metadata: ArrayV2Metadata | ArrayV3Metadata | GroupMetadata = self.metadata
+
+        while indexers:
```
d-v-b, line 706: wouldn't the [flattened form](https://github.com/zarr-developers/zarr-python/blob/11312534ebe683d73cbbcc2da9e88933cb00cc14/src/zarr/core/group.py#L239) of the consolidated metadata make this a lot simpler?

src/zarr/core/group.py
d-v-b: actually since the overall structure here is `if x: raise; assert not x` then I think we need neither an assertion nor an exception here

src/zarr/core/group.py
d-v-b: maybe the problem with that is that it requires that the consolidated metadata be complete? whereas the iterative approach can handle a group with 'live' metadata.

src/zarr/core/group.py
d-v-b: > I believe the only time this can happen through the public API is when users add a new group to group with consolidated metadata, and don't then re-consolidate.

Do we need to support this case? I would think that users could expect that, when using consolidated metadata, then metadata lookups (beyond the first) will _never_ trigger IO.

src/zarr/core/group.py
TomAugspurger: This was just for mypy. I'm not sure why its wasn't able to infer that metadata.consolidated_metadata` wasn't None.

I can try removing it, but my local pre-commit seems to have broken so I'll test it in CI.

src/zarr/core/group.py
TomAugspurger: Some discussion on this here: https://github.com/zarr-developers/zarr-python/pull/2363#issuecomment-2414491502. I think there isn't really a clear answer. I'm sympathetic to users who want to know for certain that they're done with I/O after loading a Group with consolidated metadata.

If we decide not to support this, we do still have a decision to make: when we try to index into a nested Group and reach a node with non-consolidated metadata, do we raise a `KeyError` (saying we know this key isn't there) or some public version of this MixedConsolidatedMetadataError (saying we don't know whether or not this key is there)?

d-v-b: I'm not a user of consolidated metadata, so take this with a grain of salt, but I would default to being strict: if people open a group in consolidated metadata mode, then IMO the contents of the consolidated metadata should purport to provide the _complete_ model of the hierarchy, even if that model happens to be wrong :) But I'd prefer to temper my strict default with whatever expectations consolidated metadata users have.

src/zarr/core/group.py
TomAugspurger: That passed on CI. Not sure what was going on earlier.

src/zarr/core/group.py
TomAugspurger: > but I would default to being strict

Just to be clear: what does "strict" mean to you here?

I actually am having trouble constructing a hierarchy that hits this using the public API:

```python
In [1]: import zarr

In [2]: store = zarr.storage.MemoryStore(mode="w")

In [3]: group = zarr.open_group(store=store, path="a/b/c/d")

In [4]: g = zarr.consolidate_metadata(store)

In [5]: g['a/b/c/d'].create_group("x")
Out[5]: <Group memory://4480609344/a/b/c/d/x>

In [6]: g['a/b/c/d'].metadata.consolidated_metadata
Out[6]: ConsolidatedMetadata(metadata={}, kind='inline', must_understand=False)
```

I think the semantics around what exact state in after mutating a consolidated hierarchy, but before reconsolidating, that we can do whatever we want.

d-v-b: > > but I would default to being strict
> 
> Just to be clear: what does "strict" mean to you here?

for me, "strict" would mean that if a user opens a group with `use_consolidated=True, then the consolidated metadata for that group will be the single source of truth about the hierarchy, and so attempting to access a sub-node that actually exists, but isn't in that document, would raise a `KeyError`. Not sure if this is actually a reasonable position

jhamman (approved):
Reapproving -- this is much cleaner and, I think, in line with user expectations for the scope of consolidated metadata

TomAugspurger:
Thanks for the reviews!

src/zarr/core/group.py
TomAugspurger: Sorry I missed this earlier. Flattened metadata would make this specific section simpler, but I think would complicate things a later since we'd when we need to "unflatten" it to put all of *its* children in its `consolidated_metadata`. Doable, but not obviously simpler in the end.

PR #1844, opened 2024-05-07
feature(store): list_* -> AsyncGenerators

jhamman:
This PR peels a few bits off from #1743, primarily focused on moving the output of the store list_* methods to `AsyncGenerators`. 

A few other nice things that ended up here:
- add zarr.testing.store module to support downstream use cases
- set pytest-asyncio mode to auto

PR #2381, opened 2024-10-15
AsyncGenerator → AsyncIterator

### LINKED ISSUE #2377: Inconsistent overridden method
### Zarr version

v3

### Numcodecs version

v0.13.1

### Python Version

3.13

### Operating System

Linux

### Installation

sources

### Description

These methods of abstract base class [`Store` ](https://github.com/zarr-developers/zarr-python/blob/fe752a2a2b0e91dbe566ab9874da5bb56957a688/src/zarr/abc/store.py#L61)are non-`async`:
* [`Store.list`](https://github.com/zarr-developers/zarr-python/blob/fe752a2a2b0e91dbe566ab9874da5bb56957a688/src/zarr/abc/store.py#L333)
* [`Store.list_prefix`](https://github.com/zarr-developers/zarr-python/blob/fe752a2a2b0e91dbe566ab9874da5bb56957a688/src/zarr/abc/store.py#L343)
* [`Store.list_dir`](https://github.com/zarr-developers/zarr-python/blob/fe752a2a2b0e91dbe566ab9874da5bb56957a688/src/zarr/abc/store.py#L359)

Child classes [`LocalStore`](https://github.com/zarr-developers/zarr-python/blob/fe752a2a2b0e91dbe566ab9874da5bb56957a688/src/zarr/storage/local.py#L69), [`LoggingStore`](https://github.com/zarr-developers/zarr-python/blob/fe752a2a2b0e91dbe566ab9874da5bb56957a688/src/zarr/storage/logging.py#L19), [`RemoteStore`](https://github.com/zarr-developers/zarr-python/blob/fe752a2a2b0e91dbe566ab9874da5bb56957a688/src/zarr/storage/remote.py#L27), [`MemoryStore`](https://github.com/zarr-developers/zarr-python/blob/fe752a2a2b0e91dbe566ab9874da5bb56957a688/src/zarr/storage/memory.py#L17), [`ZipStore`](https://github.com/zarr-developers/zarr-python/blob/fe752a2a2b0e91dbe566ab9874da5bb56957a688/src/zarr/storage/zip.py#L19) override these functions as `async`. Take the example of [`ZipStore`](https://github.com/zarr-developers/zarr-python/blob/fe752a2a2b0e91dbe566ab9874da5bb56957a688/src/zarr/storage/zip.py#L19):
* [`ZipStore.list`](https://github.com/zarr-developers/zarr-python/blob/840a3f7bd7a4952cce1281981aac555959e66e18/src/zarr/storage/zip.py#L234)
* [`ZipStore.list_prefix`](https://github.com/zarr-developers/zarr-python/blob/840a3f7bd7a4952cce1281981aac555959e66e18/src/zarr/storage/zip.py#L240)
* [`ZipStore.list_dir`](https://github.com/zarr-developers/zarr-python/blob/840a3f7bd7a4952cce1281981aac555959e66e18/src/zarr/storage/zip.py#L246)

Is that on purpose? I think `async` is part of the method signature and shoudl be added to the base class.

### Steps to reproduce

-

### Additional output

_No response_

DimitriPapadopoulos:
Fixes #2377.

See https://github.com/microsoft/pyright/discussions/4741.

TODO:

### DIFF
--- a/src/zarr/abc/store.py
+++ b/src/zarr/abc/store.py
@@ -6,7 +6,7 @@ from itertools import starmap
 from typing import TYPE_CHECKING, NamedTuple, Protocol, runtime_checkable
 
 if TYPE_CHECKING:
-    from collections.abc import AsyncGenerator, Iterable
+    from collections.abc import AsyncGenerator, AsyncIterator, Iterable
     from types import TracebackType
     from typing import Any, Self, TypeAlias
 
@@ -329,16 +329,19 @@ class Store(ABC):
         ...
 
     @abstractmethod
-    def list(self) -> AsyncGenerator[str]:
+    def list(self) -> AsyncIterator[str]:
         """Retrieve all keys in the store.
 
         Returns
         -------
-        AsyncGenerator[str, None]
+        AsyncIterator[str]
         """
+        # This method should be async, like overridden methods in child classes.
+        # However, that's not straightforward:
+        # https://stackoverflow.com/questions/68905848
 
     @abstractmethod
-    def list_prefix(self, prefix: str) -> AsyncGenerator[str]:
+    def list_prefix(self, prefix: str) -> AsyncIterator[str]:
         """
         Retrieve all keys in the store that begin with a given prefix. Keys are returned relative
         to the root of the store.
@@ -349,11 +352,14 @@ class Store(ABC):
 
         Returns
         -------
-        AsyncGenerator[str, None]
+        AsyncIterator[str]
         """
+        # This method should be async, like overridden methods in child classes.
+        # However, that's not straightforward:
+        # https://stackoverflow.com/questions/68905848
 
     @abstractmethod
-    def list_dir(self, prefix: str) -> AsyncGenerator[str]:
+    def list_dir(self, prefix: str) -> AsyncIterator[str]:
         """
         Retrieve all keys and prefixes with a given prefix and which do not contain the character
         “/” after the given prefix.
@@ -364,8 +370,11 @@ class Store(ABC):
 
         Returns
         -------
-        AsyncGenerator[str, None]
+        AsyncIterator[str]
         """
+        # This method should be async, like overridden methods in child classes.
+        # However, that's not straightforward:
+        # https://stackoverflow.com/questions/68905848
 
     async def delete_dir(self, prefix: str) -> None:
         """
--- a/src/zarr/storage/local.py
+++ b/src/zarr/storage/local.py
@@ -12,7 +12,7 @@ from zarr.core.buffer import Buffer
 from zarr.core.common import concurrent_map
 
 if TYPE_CHECKING:
-    from collections.abc import AsyncGenerator, Iterable
+    from collections.abc import AsyncIterator, Iterable
 
     from zarr.core.buffer import BufferPrototype
     from zarr.core.common import AccessModeLiteral
@@ -217,14 +217,14 @@ class LocalStore(Store):
         path = self.root / key
         return await asyncio.to_thread(path.is_file)
 
-    async def list(self) -> AsyncGenerator[str]:
+    async def list(self) -> AsyncIterator[str]:
         # docstring inherited
         to_strip = self.root.as_posix() + "/"
         for p in list(self.root.rglob("*")):
             if p.is_file():
                 yield p.as_posix().replace(to_strip, "")
 
-    async def list_prefix(self, prefix: str) -> AsyncGenerator[str]:
+    async def list_prefix(self, prefix: str) -> AsyncIterator[str]:
         # docstring inherited
         to_strip = self.root.as_posix() + "/"
         prefix = prefix.rstrip("/")
@@ -232,7 +232,7 @@ class LocalStore(Store):
             if p.is_file():
                 yield p.as_posix().replace(to_strip, "")
 
-    async def list_dir(self, prefix: str) -> AsyncGenerator[str]:
+    async def list_dir(self, prefix: str) -> AsyncIterator[str]:
         # docstring inherited
         base = self.root / prefix
         try:
--- a/src/zarr/storage/logging.py
+++ b/src/zarr/storage/logging.py
@@ -10,7 +10,7 @@ from typing import TYPE_CHECKING, Any, Self
 from zarr.abc.store import AccessMode, ByteRangeRequest, Store
 
 if TYPE_CHECKING:
-    from collections.abc import AsyncGenerator, Generator, Iterable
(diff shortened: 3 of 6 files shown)


src/zarr/abc/store.py
```diff
@@ -329,17 +329,17 @@ def supports_listing(self) -> bool:
...
-    def list(self) -> AsyncGenerator[str, None]:
+    async def list(self) -> AsyncGenerator[str, None]:
         """Retrieve all keys in the store.
 
         Returns
         -------
         AsyncGenerator[str, None]
         """
-        ...
+        yield  # type: ignore[misc]
```
paraseba, line 339: Would it be better to yield an empty string to avoid having to silence mypy?

DimitriPapadopoulos: Sure, that should work.

DimitriPapadopoulos:
> I think a better type for these functions would be:
> 
> ```python
> async def list(self) -> AsyncIterator[str]:
> ```

The roadmap defines the [Store API](https://github.com/zarr-developers/zarr-python/blob/main/docs/roadmap.rst#store-api) as follows:
```python
    async def list(self) -> List[str]:
        ...  # required for listable stores

    async def list_prefix(self, prefix: str) -> List[str]:
        ...  # required for listable stores

    async def list_dir(self, prefix: str) -> List[str]:
        ...  # required for listable stores
```

All current child classes use the `prefix` argument. I don't think it should be removed.

As for the return value, not sure why it changed from `List[str]`/`AsyncGenerator[str]` to `AsyncGenerator[str, None]` in #1844. Switch back to `AsyncGenerator[str]`?

paraseba:
> ~All current child classes use the `prefix` argument. I don't think it should be removed.~

Sorry, I just missed the argument in the example, definitely we shouldn't drop it. My point was about returning `AsyncIterator` instead of `AsyncGenerator`, the later is more of an implementation detail. We definitely don't want `list` because that would force the whole thing to live in memory at once.

DimitriPapadopoulos:
Meanwhile, I tried to revert the return type from `AsyncGenerator[str, None]` to `AsyncGenerator[str]`, to see what happens.

As for the `AsyncGenerator` → `AsyncIterator` change, how about dealing with in an issue and a separate PR?

paraseba:
> As for the AsyncGenerator → AsyncIterator change, how about addressing it in an issue and a separate PR?

It's definitely not a big deal, `AsyncGenerator[str]` also works. But, doing it in separate PR would require changing the abc and every implementation twice.

DimitriPapadopoulos:
OK, let me try then. It's just that's I'll have to modify all child classes.

src/zarr/abc/store.py
```diff
@@ -329,17 +329,17 @@ def supports_listing(self) -> bool:
...
+    async def list(self) -> AsyncIterator[str]:
         """Retrieve all keys in the store.
 
         Returns
         -------
-        AsyncGenerator[str, None]
+        AsyncIterator[str, None]
         """
-        ...
+        return iter([])
```
paraseba, line 339: weird, you shouldn't need the implementation if you annotate with `abstractmethod`

src/zarr/storage/local.py
```diff
@@ -217,28 +217,26 @@ async def exists(self, key: str) -> bool:
...
 
-    async def list(self) -> AsyncGenerator[str, None]:
+    async def list(self) -> AsyncIterator[str]:
         # docstring inherited
         to_strip = str(self.root) + "/"
-        for p in list(self.root.rglob("*")):
-            if p.is_file():
-                yield str(p).replace(to_strip, "")
+        allfiles = [str(p).replace(to_strip, "") for p in self.root.rglob("*") if p.is_file()]
+        return (onefile for onefile in allfiles)
```
paraseba, line 224: we don't want to materialize the full `allfiles` list in memory. Can you use `async for` instead? Here is an example of how you could do this:

```python
from collections.abc import AsyncIterator
from abc import ABC, abstractmethod

class Foo(ABC):
    @abstractmethod
    async def foo(self) -> AsyncIterator[str]:
        ...


class Bar(Foo):

    async def helper(self) -> AsyncIterator[str]:
        raise ValueError("")

    async def foo(self) -> AsyncIterator[str]:
        return (x async for x in await self.helper())
```

This code should work and mypy successfully, and doesn't materialize the results of calling `helper`.

src/zarr/abc/store.py
DimitriPapadopoulos: Indeed. I had to `yield` with `AsyncGenerator`. `AsyncIterator` has the additional benefit that an empty method is good enough.

src/zarr/storage/local.py
DimitriPapadopoulos: I'm afraid I don't have enough experience with `async` and no time to learn it in the short term. I am unable to find a solution that pleases MyPy.

jhamman:
@DimitriPapadopoulos - are you planning to return to this?

DimitriPapadopoulos:
I tried briefly, but I cannot get typing to work after rebasing. I need to take a thorough look.

src/zarr/abc/store.py
```diff
@@ -329,16 +329,22 @@ def supports_listing(self) -> bool:
...
         Returns
         -------
-        AsyncGenerator[str, None]
+        AsyncIterator[str]
+
+        Notes
+        -----
+        This method should be async,
+        `How to correctly specify type hints with AsyncGenerator and AsyncContextManager <https://stackoverflow.com/questions/68905848/how-to-correctly-specify-type-hints-with-asyncgenerator-and-asynccontextmanager>`_
+        explains why it is not.
```
jhamman, line 343: I'd like to move this to a comment rather than the public docstring. Most implementations will inherit this and end users don't need to know about this implementation detail.

jhamman (approved):
I'm happy here (except for the requested docstring changes).

src/zarr/abc/store.py
DimitriPapadopoulos: Yes, I wasn't sure about it. I put it in the documentation because it documents an inconsistency in function signatures.

DimitriPapadopoulos: Ideally, I would document this using something similar to disabling Pylint warning [W0236](https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/invalid-overridden-method.html):
```
# pylint: disable=invalid-overridden-method
```
Unfortunately there's no equivalent rule in ruff, as far as I can see.

DimitriPapadopoulos: Fixed.

jhamman: Thanks.

Issue #2357, opened 2024-10-14
Paths with leading slashes do bad things


jhamman:
### Zarr version

3.0.0-beta

### Numcodecs version

0.13

### Python Version

3.11

### Operating System

Mac

### Installation

pip

### Description

In https://github.com/pydata/xarray/pull/9552, we have noticed that setting keys with leading slashes ends poorly for some stores (namely the local store).

We should probably be normalizing keys in the Array/Group APIs before the store even sees them.

### Steps to reproduce

```python
import zarr
group = zarr.open_group('/Users/jhamman/workdir/zarr/foo', mode='a')
group.create_group('/foo', attributes={'a': 'b'})
```

```python-traceback
---------------------------------------------------------------------------
OSError                                   Traceback (most recent call last)
Cell In[4], line 1
----> 1 group.create_group('[/foo](http://localhost:8888/foo)', attributes={'a': 'b'})

File [~/Library/CloudStorage/Dropbox/src/zarr-python/src/zarr/core/group.py:1480](http://localhost:8888/lab/tree/~/Library/CloudStorage/Dropbox/src/zarr-python/src/zarr/core/group.py#line=1479), in Group.create_group(self, name, **kwargs)
   1479 def create_group(self, name: str, **kwargs: Any) -> Group:
-> 1480     return Group(self._sync(self._async_group.create_group(name, **kwargs)))

File [~/Library/CloudStorage/Dropbox/src/zarr-python/src/zarr/core/sync.py:185](http://localhost:8888/lab/tree/~/Library/CloudStorage/Dropbox/src/zarr-python/src/zarr/core/sync.py#line=184), in SyncMixin._sync(self, coroutine)
    182 def _sync(self, coroutine: Coroutine[Any, Any, T]) -> T:
    183     # TODO: refactor this to to take *args and **kwargs and pass those to the method
    184     # this should allow us to better type the sync wrapper
--> 185     return sync(
    186         coroutine,
    187         timeout=config.get("async.timeout"),
    188     )

File [~/Library/CloudStorage/Dropbox/src/zarr-python/src/zarr/core/sync.py:141](http://localhost:8888/lab/tree/~/Library/CloudStorage/Dropbox/src/zarr-python/src/zarr/core/sync.py#line=140), in sync(coro, loop, timeout)
    138 return_result = next(iter(finished)).result()
    140 if isinstance(return_result, BaseException):
--> 141     raise return_result
    142 else:
    143     return return_result

[...]

PR #2384, opened 2024-10-15
fix/normalize storage paths

d-v-b:
This PR adds some of the path normalization logic from zarr v2 into v3. I largely copied [this function](https://github.com/zarr-developers/zarr-python/blob/ed94877ff69d859f2e9d412e775e2887bc6c71bf/zarr/util.py#L340-L380) as-is with a few tweaks. That function is actually not terribly strict as far as path parsing is concerned, so I'm open to requests that we make it stricter. The main thing to note is that `normalize_path` (the new name for that function) strips leading `/` characters from strings, which ensures that paths are relative.

I wired this function up to `make_store_path`, which previously didn't take `path` as a parameter (now it does). This led to some nice code deletion opportunities in our various creation routines.

Something that always bugged me in zarr v2 was use of `None` as a default value for types where there's already a default value. In the case of strings, the empty string `''` is a perfectly good default value, and so allowing `None` is just noise IMO. I was _tempted_ to try and remove `None` as a valid path argument in this PR, but I held back. If people agree that `None` is a silly default for a stringy value like `path` when we can just use `''` instead, then I can make that change as well.

Fixes https://github.com/zarr-developers/zarr-python/issues/2357

TODO:

### DIFF
--- a/src/zarr/storage/_utils.py
+++ b/src/zarr/storage/_utils.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import re
 from pathlib import Path
 from typing import TYPE_CHECKING
 
@@ -8,13 +9,12 @@ if TYPE_CHECKING:
 
 
 def normalize_path(path: str | bytes | Path | None) -> str:
-    # handle bytes
     if path is None:
         result = ""
     elif isinstance(path, bytes):
         result = str(path, "ascii")
-    # ensure str
-    # handle pathlib.Path and upath.Path
+
+    # handle pathlib.Path
     elif isinstance(path, Path):
         result = str(path)
 
@@ -27,24 +27,12 @@ def normalize_path(path: str | bytes | Path | None) -> str:
     # convert backslash to forward slash
     result = result.replace("\\", "/")
 
-    # ensure no leading slash
-    while len(result) > 0 and result[0] == "/":
-        result = result[1:]
-
-    # ensure no trailing slash
-    while len(result) > 0 and result[-1] == "/":
-        result = result[:-1]
+    # remove leading and trailing slashes
+    result = result.strip("/")
 
     # collapse any repeated slashes
-    previous_char = None
-    collapsed = ""
-    for char in result:
-        if char == "/" and previous_char == "/":
-            pass
-        else:
-            collapsed += char
-        previous_char = char
-    result = collapsed
+    pat = re.compile(r"//+")
+    result = pat.sub("/", result)
 
     # disallow path segments with just '.' or '..'
     segments = result.split("/")


src/zarr/storage/common.py
```diff
@@ -209,37 +214,44 @@ async def make_store_path(
...
-
+    path_normalized = normalize_path(path)
     if isinstance(store_like, StorePath):
         if mode is not None and mode != store_like.store.mode.str:
             _store = store_like.store.with_mode(mode)
             await _store._ensure_open()
-            store_like = StorePath(_store)
-        result = store_like
+            store_like = StorePath(_store, path=store_like.path) / path_normalized
+        result = store_like / path_normalized
```
jhamman, line 223: looks like `/ path_normalized` is being applied twice here.

tests/v3/test_api.py
```diff
@@ -37,6 +48,20 @@ def test_create_array(memory_store: Store) -> None:
     assert z.chunks == (40,)
 
 
+@pytest.mark.parametrize("path", ["/", "///foo/bar"])
```
jhamman, line 51:
```suggestion
@pytest.mark.parametrize("path", ["foo", "/", "/foo", "///foo/bar"])
```

jhamman (approved):
Good stuff @d-v-b - I'm glad we're moving `path` into `make_store_path`.

src/zarr/storage/common.py
d-v-b: good catch. I noticed that this wasn't failing any tests, so in addition to fixing this, I added tests for it in https://github.com/zarr-developers/zarr-python/pull/2384/commits/e6ecf99e67ef8e046214eeec297c014e367e59e0

src/zarr/storage/_utils.py
```diff
@@ -1,11 +1,61 @@
 from __future__ import annotations
 
+from pathlib import Path
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
     from zarr.core.buffer import Buffer
 
 
+def normalize_path(path: str | bytes | Path | None) -> str:
```
TomAugspurger, line 10: Does this need to take `| None` still? Or have we replaced all those with `""` as the default?

src/zarr/storage/_utils.py
```diff
@@ -1,11 +1,61 @@
...
+
+    # convert backslash to forward slash
+    result = result.replace("\\", "/")
+
+    # ensure no leading slash
+    while len(result) > 0 and result[0] == "/":
+        result = result[1:]
+
+    # ensure no trailing slash
+    while len(result) > 0 and result[-1] == "/":
```
TomAugspurger, line 35: Same question, but for `.rstrip`.

src/zarr/storage/_utils.py
```diff
@@ -1,11 +1,61 @@
...
+        result = result[:-1]
+
+    # collapse any repeated slashes
```
TomAugspurger, line 38: Maybe a regex is clearer?

```
pat = re.compile(r"//+")
result = pat.sub("/", result)
```

src/zarr/storage/_utils.py
d-v-b: I have not changed any of the top-level defaults, so this still needs to handle `None`.

src/zarr/storage/_utils.py
d-v-b: `strip` took care of both of them

src/zarr/storage/_utils.py
d-v-b: done

TomAugspurger (approved):

Issue #508, opened 2019-11-14
Document Group.__setitem__ or remove it


alimanfoo:
Behaviour of ``Group.__setitem__`` is not always intuitive, e.g. #502. Either document this behaviour (e.g., in tutorial and API reference) or remove support for it altogether.

Issue #251, opened 2018-03-28
Storing Dask Arrays directly to Zarr Group fails


jakirkham:
The gist of the issue is that when trying to store an array-like object to a Zarr Group, it looks for the `chunks` attribute to advise on how to chunk. When it finds `chunks` on Dask Arrays, the chunk sizes are not global and uniform necessarily, but specific sizes are given for each chunk, which may not be uniform. Zarr understandably stumbles over this as the format is not what it expects.

Even if Zarr could handle the Dask chunking format somehow, there is a question of what to do with non-uniform chunk sizes. There are two main options to consider: support non-uniform chunking in Zarr ( https://github.com/zarr-developers/zarr/issues/245 ) and/or rechunk Dask Arrays to be uniform ( https://github.com/dask/dask/issues/3302 ). So some things to think about on both fronts. This should help provide both of those issues more context.

cc @mrocklin

<br>

#### Minimal, reproducible code sample, a copy-pastable example if possible

<details>

```python
In [1]: import zarr

In [2]: import dask.array as da

In [3]: z = zarr.open_group("test.zarr")

In [4]: a = da.random.random((100, 110), chunks=10)

In [5]: z["a"] = a
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-5-142459babd67> in <module>()
----> 1 z["a"] = a

/zopt/conda2/envs/test/lib/python3.6/site-packages/zarr/hierarchy.py in __setitem__(self, item, value)
    335 
    336     def __setitem__(self, item, value):
--> 337         self.array(item, value, overwrite=True)
    338 
    339     def __delitem__(self, item):

/zopt/conda2/envs/test/lib/python3.6/site-packages/zarr/hierarchy.py in array(self, name, data, **kwargs)
    908         """Create an array. Keyword arguments as per
    909         :func:`zarr.creation.array`."""
--> 910         return self._write_op(self._array_nosync, name, data, **kwargs)
    911 
    912     def _array_nosync(self, name, data, **kwargs):

/zopt/conda2/envs/test/lib/python3.6/site-packages/zarr/hierarchy.py in _write_op(self, f, *args, **kwargs)
    628 
    629         with lock:
--> 630             return f(*args, **kwargs)
    631 
    632     def create_group(self, name, overwrite=False):

/zopt/conda2/envs/test/lib/python3.6/site-packages/zarr/hierarchy.py in _array_nosync(self, name, data, **kwargs)
[...]

Issue #2083, opened 2024-08-13
How easy can it be to create a zarr array


d-v-b:
Users come to Zarr with a variety of array-like objects -- numpy arrays, or dask arrays, or xarray DataArrays, zarr v2 arrays, zarr v3 arrays, etc. Imagine a venn diagram of attributes / methods for these objects: `shape`, `__getitem__`, `dtype` would be in the shared middle, and `chunks`, `chunksize`, `attrs`, `dims`, `codecs`, `filters` in the disjoint periphery. How can we conveniently model an arbitrary array-like object as a Zarr array? In particular, how can we ensure that you can create a _complete_ Zarr array from an existing array-like object (which might be already a zarr array) with a _single function call_?

If we agree on that objective, then here is a rough outline of what that function could look like:
 
- we should have a top-level `from_array` method that creates a Zarr array from an existing array-like object. 
```python
# numpy
np_arr = np.zeros(10)
zarr.from_array(np_arr) # memorystore-backed zarr v3 array with shape 10 and dtype float64, and default parameters for everything else
zarr.from_array(np_arr, zarr_format=2, compressor=Gzip(), attributes={'foo': 10}) # same as above, but v2, with gzip, and attributes

# dask
da_arr = da.zeros((10,), chunks=(1,))
zarr.from_array(da_arr) # inherits the `chunks` attribute from the array
zarr.from_array(da_arr, chunking_bikeshed=(2,)) # overrides the chunks attribute, kwarg name tbd 🙃 

# xarray
xr_arr = xarray.DataArray(np.zeros(10), attrs={'foo': 10}, dims=('dim_0',))
zarr.from_array(xr_arr) # zarr v3 array with dimension names inherited xr_arr.dims, attrs from xr_arr.attrs)

# zarr
zarr.from_array(zarr.zeros(10)) # makes a copy of the array
``` 

some open questions: 
- should we copy data? over in `pydantic-zarr` I implemented a [`from_array`](https://github.com/janelia-cellmap/pydantic-zarr/blob/main/src/pydantic_zarr/v2.py#L181) function that only creates array metadata, because users might not want to eagerly move 10 TB of data at array definition time. Perhaps this could be controlled with a keyword argument.
- should we support creating v2 arrays through this API, or use a `v2.from_array` function for that? I'm fine either way.
- How much work is required to implicitly model the different array-like libraries enough for the above functionality to be useful?
[...]

PR #2393, opened 2024-10-17
feature(group): add group setitem api

jhamman:
I think https://github.com/zarr-developers/zarr-python/issues/251 was fixed in 2.x at some point but it this PR also makes it work with v3.

closes https://github.com/zarr-developers/zarr-python/issues/251
closes https://github.com/zarr-developers/zarr-python/issues/508

TODO:


jni:
Are you aiming to get this merged before 3.0 @jhamman? And if not, what is the recommended way to do this with the current beta? My naive attempt in napari/napari#7215 ([here](https://github.com/napari/napari/pull/7215/files/50a4d60d6888b910bf8f9d384ed4562eeb418460#r1804409843) to translate:

```python
grp[key] = arr
```

is:

```python
grp.create_array(name=key, shape=arr.shape, dtype=arr.dtype, data=arr)
```

(which is a little clunky, maybe shape and dtype should be optional if data is provided?)

d-v-b:
putting aside the value of v2 compatibility, I'm not a fan of this API for a few reasons:
- a zarr group is two mappings: its attributes and the members. Using `__setitem__` for just one of those two mappings is arbitrary and unfair :judge: 
- using `__setitem__` to create an array hides the full configuration options of array creation and can lead users to over-rely on zarr defaults
- using `__getitem__` to retrieve an array similarly offers no place to configure the runtime array-access properties like `write_empty_chunks`, caching, etc. 

for the first point, we could scope all the hierarchy mutation stuff to a `group.members` attribute, e.g. `group.members['array'] = array`. but I would still want a way for users to configure the default array parameters, e.g. via our new config object.

> ```python
> grp.create_array(name=key, shape=arr.shape, dtype=arr.dtype, data=arr)
> ```  
> (which is a little clunky, maybe shape and dtype should be optional if data is provided?)

@jni I agree that this is clunky, and it's silly to expose `data` and `dtype`. I think it would be better if we separated "create a brand new array" from "create an array like this one" into two different methods.

jni:
Regardless of Citizens United 😂, I don't think attributes have human rights @d-v-b, so I don't think "unfair" is really valid here. Which leaves "arbitrary". I think generally in hierarchical APIs like these it's very common for attributes to be under a .attrs, um, attribute? I think "commonly known API" is not arbitrary, but being a nice player in an ecosystem.

As to defaults, I am a big fan of them, let's have more of those. (See also #2083 😉)

Either way, I am looking for guidance as to what I should do in napari for now, and also, that guidance should probably go into #2102. 🙏

d-v-b:
> Regardless of Citizens United 😂, I don't think attributes have human rights @d-v-b, so I don't think "unfair" is really valid here. Which leaves "arbitrary". I think generally in hierarchical APIs like these it's very common for attributes to be under a .attrs, um, attribute? I think "commonly known API" is not arbitrary, but being a nice player in an ecosystem.

I'm not aware of the hierarchical API ecosystem; I'm coming from a more basic "how do we model the data" perspective, and from that POV, arbitrary symmetry breaking is almost always a mistake. For example, if you had to make a JSON serializable model of a zarr group, you have three options. (Recall that a zarr group is a dict of attributes and a dict of arrays / groups)

- option 1: 
  ```
  group: {"array_1": {... model of array_1}, "group_1": {...model of group_1}, "attrs": {... the attributes}
  ```
  This sucks because we can have name collisions between arrays / groups and attribute keys.

- option 2: 
  ```
  group: {"attr_key_1": {... attrs under attribute key 1}, "members": {...the arrays and groups}}`
  ```

  This sucks for the same reason as above. We are mixing our attribute namespace with our group model namespace.

- option 3: 
  ```
  group: {"attrs": {... the attributes}, "members": {...the arrays and groups}}`
  ```
  This separates the attributes and members namespaces. This is the only one of the three that's a usable JSON data model.

We could of course ignore this in our `Group` API and pretend that a group can be modeled as `dict[str, Array | Group]`, but that's a decision motivated by either laziness or deference to bad decisions made in the past. The motivation is fairness :) which should lead us to explicitly separate the members and the attributes IF we want a dict-like data model for a group.

jni:
> that's a decision motivated by either laziness or deference to bad decisions made in the past.

ok, fair. ;)

d-v-b:
> (but, sometimes, deference to the past can be important. I won't make that judgement call here. But hdf5, and (consequently?) NetCDF, both use the .attrs-style API, if I'm not mistaken. And zarr v2??)

and I would argue it's a mistake in all three cases. FWIW I didn't think about this much until I did `pydantic-zarr`, i.e. I had to embed the zarr data model into JSON, at which point I realized these problems with the Zarr v2 API (and the APIs that it was inspired from). since then I can't unsee it :see_no_evil:

jhamman:
My opinion here is that we should include this in 3.0 but consider the following options:

- rename `AsyncGroup.setitem` to something more in line with where we want the GroupAPI to go
- on the sync `Group`
- turn `Group.arrays` into a property 
- use the `arrays.setter` as a fast path 
- deprecate `Group.__setitem__` in favor of `Group.arrays['a'] = arr`

@d-v-b - how would this API feel to you?

d-v-b:
> My opinion here is that we should include this in 3.0 but consider the following options:
> 
>     * rename `AsyncGroup.setitem` to something more in line with where we want the GroupAPI to go
> 
>     * on the sync `Group`
> 
>     * turn `Group.arrays` into a property
> 
>     * use the `arrays.setter` as a fast path
> 
>     * deprecate `Group.__setitem__` in favor of `Group.arrays['a'] = arr`
> 
> 
> @d-v-b - how would this API feel to you?

This is definitely an improvement, and I don't think I'm opposed to it. I'm most excited about the deprecation part :) 

Should we put default array codec settings + chunking in the config, and use that for invocations like `Group.arrays['foo'] = arr`? And similarly for `__getitem__`, although I don't think we have as much configuration surface are there. That way at least this stuff is configurable.

src/zarr/core/group.py
```diff
@@ -294,6 +294,36 @@ def flatten(
...
+        Set an array in the group.
+        """
+        self._group._sync(self._group._async_group.set_array(key, value))
+
+    def __iter__(self) -> Generator[tuple[str, Array], None]:
+        for name, async_array in self._group._sync_iter(self._group._async_group.arrays()):
+            yield name, Array(async_array)
+
+    def __call__(self) -> Generator[tuple[str, Array], None]:
+        return iter(self)
```
jhamman, line 324: I could use some opinions here!

On one hand, `iter(group.arrays)` is nice and tidy. But it also doesn't match the existing API (`group.arrays()`) and there are other methods where symetry seems useful (`array_keys`, `array_values`, etc.). 

@d-v-b - you'll have an opinion here :)

d-v-b: what breaks if we remove `__call__`? I feel like that's redundant if you can just iterate. And then we could add `keys` and `values` methods to the `ArraysProxy` object that lossily call `__iter__`, thereby replacing `array_keys` and `array_values`? 

We end up with a kind of a weird data structure -- something that's _asymptotically_ a dict? But I think that also matches the semantics of "lots of stuff stored on disk with unique names".

jhamman: > what breaks if we remove __call__ 

lots of stuff that looks like this (from our test suite):

```python
>       assert sorted(group.arrays(), key=lambda x: x[0]) == expected_arrays
E       TypeError: 'ArraysProxy' object is not callable
```

Looking at 2.18, I'm realizing we also are missing the `recurse` option to `arrays`... 

All of this makes me think it may be asking too much to overload `.arrays` with setiem/getitem/iter/etc.

jhamman:
I just ran across this blast from the past: https://github.com/zarr-developers/zarr-python/issues/508

jhamman:
This is ready for a review. I've got taken the least invasive route. I think we may want to deprecate this but I see no harm in keeping it now. We can debate its removal after the 3.0 release.

tests/test_group.py
```diff
@@ -424,11 +424,28 @@ def test_group_len(store: Store, zarr_format: ZarrFormat) -> None:
 
 def test_group_setitem(store: Store, zarr_format: ZarrFormat) -> None:
     """
-    Test the `Group.__setitem__` method.
+    Test the `Group.__setitem__` method. (Deprecated)
```
jni, line 427: Only comment: typically it's nice for users to say, "Deprecated: use group.create_array instead" (for example)

jni (approved):
Thanks @jhamman! Personally I quite like soft deprecations as a way to gently move a community towards other APIs.

tests/test_group.py
jhamman, line 427:
```suggestion
    Test the `Group.__setitem__` method.
```

jhamman:
@d-v-b - I'm planning to merge this in its current state. If you would like to come behind with a set of deprecations and/or features that add a separate namespace for `Group.arrays`, go for it!

Issue #735, opened 2021-05-04
Update fsspec minimal version?


joshmoore:
I've just gotten this user report:

```  File "c:\users\c\napari\lib\site-packages\zarr\core.py", line 1836, in _chunk_getitems
    cdatas = self.chunk_store.getitems(ckeys, on_error="omit")
  File "c:\users\c\napari\lib\site-packages\zarr\storage.py", line 1085, in getitems
    return self.map.getitems(keys, on_error="omit")
AttributeError: 'FSMap' object has no attribute 'getitems'
```

```
(napari) PS C:\Users\callan> pip show zarr
Name: zarr
Version: 2.8.1
Summary: An implementation of chunked, compressed, N-dimensional arrays for Python.
Home-page: https://github.com/zarr-developers/zarr-python
Author: None
Author-email: None
License: MIT
Location: c:\users\callan\napari\lib\site-packages
Requires: numcodecs, fasteners, asciitree, numpy
Required-by: ome-zarr, omero-napari
```

Is it possible that we need to (A) update the fsspec pin and (B) perhaps include it as a requirement?

```python
requirements_dev_optional.txt:fsspec==0.8.4; python_version > '3.6'
```

cc: @martindurant

martindurant:
fsspec is a pretty small requirement, since it has no deps of its own, but many zarr users won't need it. 
For an optional requirement, yes I would update. 0.9.0 includes `getitems`, but why not just keep up with latest.

joshmoore:
> but why not just keep up with latest.

Sorry, I didn't mean pin, but rather set a minimal version. :+1: for going `>= 0.9.0`.  If there are other votes for dropping optional, speak up.

joshmoore:
Oh dear. This may end up being something of a can-of-worms. All of the versions in `requirements_dev_optional.txt` are pinned, and I imagine they haven't been updated since appveyor was dropped in gh-706.  The activation of dependabot (gh-734) should kick off the process again.

MatthewBM:
I'm having stability issues with both fsspec<=0.8.0` and `fsspec>=0.9.0`. Only fsspec == 0.8.7 is working for me currently using Zarr and S3

fsspec<=0.8.0 cannot import maybe_sync:
https://www.gitmemory.com/issue/intake/filesystem_spec/597/829392517
```
 from fsspec.asyn import AsyncFileSystem, sync, sync_wrapper, maybe_sync
ImportError: cannot import name 'maybe_sync'
```
[...]

PR #2395, opened 2024-10-17
Slim dependencies

### LINKED ISSUE #2391: Rethinking Zarr's core dependencies
I'd like to open the conversation about what Zarr's core dependencies are for 3.0. Currently, this looks like: 

https://github.com/zarr-developers/zarr-python/blob/11312534ebe683d73cbbcc2da9e88933cb00cc14/pyproject.toml#L25-L34

Some of these are not used anymore (`asciitree` and `fasteners`) so those can safely go.

Then there is `fsspec` and `crc32c`. These are only needed for the `RemoteStore` and `ShardingCodec`, respectively. What do we think about making these optional?

One proposed diff in our dependencies would look something like:

```diff
 dependencies = [
-    'asciitree',
     'numpy>=1.25',
-    'fasteners',
-    'numcodecs>=0.10.2',
-    'fsspec>2024',
-    'crc32c',
+    'numcodecs>=0.12',
     'typing_extensions',
     'donfig',
 ]

 [project.optional-dependencies]
+remote = [
+    "fsspec",
+]
+sharding = [
+    "crc32c",
+]
```

Notes:
- fsspec is pure python with no dependencies so is not a particular heavy dependency.
- crc32c could potentially move into numcodecs, right?

jhamman:
This PR implements the main ideas from #2391, removing `crc32c` and `fsspec` as core dependencies and adding them to the `sharding` and `remote` groups respectively. Imports are adjusted in minor ways and we now take advantage of `pytest.importorskip` in a few new places.

TODO: if we decide to go this route, we should have a separate part of the test matrix that includes optional dependencies.

Closes #2391 
Fixes https://github.com/zarr-developers/zarr-python/issues/735

TODO:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,6 +30,7 @@ dependencies = [
     'numcodecs>=0.12',
     'typing_extensions>=4.9',
     'donfig>=0.8.1',
+    'crc32c>=2.4',
 ]
 dynamic = [
   "version",
@@ -53,9 +54,6 @@ keywords = ["Python", "compressed", "ndimensional-arrays", "zarr"]
 [project.optional-dependencies]
 fsspec = [
     "fsspec>=2023.10.0",
-]
-sharding = [
-    "crc32c>=2.4",
 ]
 test = [
     "coverage",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,12 +23,8 @@ maintainers = [
 requires-python = ">=3.11"
 # If you add a new dependency here, please also add it to .pre-commit-config.yml
 dependencies = [
-    'asciitree',
     'numpy>=1.25',
-    'fasteners',
-    'numcodecs>=0.10.2',
-    'fsspec>2024',
-    'crc32c',
+    'numcodecs>=0.12',
     'typing_extensions',
     'donfig',
 ]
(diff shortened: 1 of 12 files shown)


dstansby (changes requested):
I think we're coming to a consensus in https://github.com/zarr-developers/zarr-python/issues/2391 (?) of
- Move dependency of `crc32c` to numcodecs (https://github.com/zarr-developers/numcodecs/issues/610)
- Make `fsspec` optional

For the second one, would be good to add some docs explaining how to install the `remote` extra, and what functionality it enables.

pyproject.toml
```diff
@@ -54,17 +51,19 @@ license = {text = "MIT License"}
 keywords = ["Python", "compressed", "ndimensional-arrays", "zarr"]
 
 [project.optional-dependencies]
+remote = [
```
jhamman, line 54:
```suggestion
fsspec = [
```

jhamman:
Something I learned over the weekend is that `crc32c` is LGPL Licensed. Regardless of whether we get `crc32c` into `numcodecs`, we may want to pick a different implementation ([googleapis/python-crc32c](https://github.com/googleapis/python-crc32c) is Apache 2.0 licenced).

pyproject.toml
```diff
@@ -26,13 +26,10 @@ maintainers = [
...
-    'asciitree',
     'numpy>=1.25',
-    'numcodecs>=0.10.2',
-    'fsspec>2024',
-    'crc32c',
-    'typing_extensions',
-    'donfig',
+    'numcodecs>=0.12',
+    'typing_extensions>=4.9',
+    'donfig>=0.8.1',
```
jhamman, line 32:
```suggestion
    'donfig>=0.8.1',
    'crc32c',
```

pyproject.toml
```diff
@@ -54,17 +51,19 @@ license = {text = "MIT License"}
 keywords = ["Python", "compressed", "ndimensional-arrays", "zarr"]
 
 [project.optional-dependencies]
+fsspec = [
+    "fsspec>=2023.10.0",
+]
+sharding = [
+    "crc32c>=2.4",
+]
```
jhamman, line 59:
```suggestion
```

PR #2413, opened 2024-10-19
[v3] Array.append

### LINKED ISSUE #2404: [v3] AttributeError: 'Array' object has no attribute 'append'
### Zarr version

3.0.0.beta

### Numcodecs version

0.13

### Python Version

3.11

### Operating System

Mac

### Installation

pip

### Description

We need to implement `Array.append` in 3.0

### Steps to reproduce

```python
In [11]: a = np.arange(10000000, dtype='i4').reshape(10000, 1000)
    ...: z = zarr.array(a, chunks=(1000, 100))
    ...: z.shape
    ...: z.append(a)
    ...: z.append(np.vstack([a, a]), axis=1)
    ...: z.shape
--------------------------------------------------------------------------
AttributeError                           Traceback (most recent call last)
Cell In[11], line 4
      2 z = zarr.array(a, chunks=(1000, 100))
      3 z.shape
----> 4 z.append(a)
      5 z.append(np.vstack([a, a]), axis=1)
      6 z.shape

AttributeError: 'Array' object has no attribute 'append'
```

### Additional output

_No response_

jhamman:
Ports the `Array.append` method over from 2.18 (plus tests for resize/append). Mostly a copy-and-paste but does come with a bit of extra behavior changes:

- `Array` class is now a non-Frozen dataclass. This is needed to support property setters. I've added a note describing when we can go back to `frozen=True`
- `Array.resize` is an in-place op again. As we discussed at the meeting yesterday, it really didn't make sense to have this any other way (what would you do with the old array object?)

closes #2404 

TODO:


src/zarr/core/array.py
```diff
@@ -1130,7 +1131,61 @@ async def _delete_key(key: str) -> None:
 
         # Write new metadata
         await self._save_metadata(new_metadata)
-        return replace(self, metadata=new_metadata)
+
+        # Update metadata (in place)
+        object.__setattr__(self, "metadata", new_metadata)
```
jhamman, line 1136: @TomAugspurger - calling this out because it will impact the xarray use of resize... now in-place again :)

src/zarr/core/array.py
```diff
@@ -1130,7 +1131,61 @@ async def _delete_key(key: str) -> None:
...
+            data = np.asanyarray(data)
+
+        self_shape_preserved = tuple(s for i, s in enumerate(self.shape) if i != axis)
+        data_shape_preserved = tuple(s for i, s in enumerate(data.shape) if i != axis)
+        if self_shape_preserved != data_shape_preserved:
+            raise ValueError(
+                "shape of data to append is not compatible with the array; "
+                "all dimensions must match except for the dimension being "
+                "appended"
+            )
```
d-v-b, line 1168:
```suggestion
            raise ValueError(
                f"shape of data to append is not compatible with the array. "
                f"The shape of the data is ({data_shape_preserved})"
                "and the shape of the array is ({self_shape_preserved})."
                "All dimensions must match except for the dimension being "
                "appended"
            )
```

tests/test_array.py
```diff
@@ -417,3 +417,193 @@ def test_update_attrs(zarr_format: int) -> None:
 
     arr2 = zarr.open_array(store=store, zarr_format=zarr_format)
     assert arr2.attrs["foo"] == "bar"
+
+
+@pytest.mark.parametrize("store", ["memory"], indirect=True)
+@pytest.mark.parametrize("zarr_format", [2, 3])
+def test_resize_1d(store: MemoryStore, zarr_format: int) -> None:
```
d-v-b, line 424: to fight test bloat we should find a way to parameterize over dimensionality instead of making separate 1d, 2d, etc tests. but that's not a blocker here.

d-v-b (changes requested):
looks good! I think we just need a more informative exception [here](https://github.com/zarr-developers/zarr-python/pull/2413/files#r1807781098).

tests/test_array.py
jhamman: agree! These tests were copied over from v2 so I'm going to leave them for now.

jhamman:
pre-commit.ci autofix

d-v-b (approved):
lgtm!

PR #2417, opened 2024-10-19
fix(array): thread order parameter through to Array.__init__

### LINKED ISSUE #2408: [v3] `zarr.array` silently ignores `order` kwarg
### Zarr version

3.0.0.beta

### Numcodecs version

0.13

### Python Version

3.11

### Operating System

Mac

### Installation

pip

### Description

Passing `order=F` to `zarr.array` silently ignores the parameter, producing a `C` ordered array.

### Steps to reproduce

```python
In [20]: a = np.arange(100000000, dtype='i4').reshape(10000, 10000).T

In [21]: f = zarr.array(a, chunks=(1000, 1000), order='F')

In [22]: f.order
Out[22]: 'C'
```

### Additional output

_No response_

jhamman:
The goal here is to allow configuring the runtime memory order of arrays. We had all the pieces in place but we stopped short of passing order through to `Array.__init__`.

closes #2408

TODO:


src/zarr/core/array.py
```diff
@@ -611,7 +609,7 @@ async def _create_v2(
         if not exists_ok:
             await ensure_no_existing_node(store_path, zarr_format=2)
         if order is None:
-            order = "C"
+            order = config.get("array.order", "C")
```
d-v-b, line 612: Suppose we change the key `array.order` to `array_indexing.order` in the config, this will happily return "C", when the correct behavior should be an error. Therefore I think we should define the default value higher up than this function, and this function should have a hard dependency on the schema of the config.

tests/test_api.py
```diff
@@ -206,6 +206,22 @@ def test_open_with_mode_w_minus(tmp_path: pathlib.Path) -> None:
         zarr.open(store=tmp_path, mode="w-")
 
 
+@pytest.mark.parametrize("order", ["C", "F", None])
+@pytest.mark.parametrize("zarr_format", [2, 3])
+def test_array_order(order: str | None, zarr_format: int) -> None:
```
d-v-b, line 211:
```suggestion
def test_array_order(order: Literal["C", "F"] | None, zarr_format: ZarrFormat) -> None:
```
This will require importing the `ZarrFormat` type

src/zarr/core/array.py
jhamman: I'm not sure I follow. What's the ask here?

jhamman: How much higher? We could require passing  `order` into `Array.create` rather than allowing this to be a nullable field. Or we could do the `config` check there.

d-v-b: would it be possible to put the default in the config? I'm guessing we will have a lot of defaults (e.g., codecs), so would probably make sense to put them all in the same place

d-v-b: looks like the [defaults are already in the config](https://github.com/zarr-developers/zarr-python/blob/37fde7defadf13b6894487987e99e098f027dcf6/src/zarr/core/config.py#L45), in which case I think we can just remove any "default to C" behavior in `_create_v2`

jhamman: okay. I think the final state should be what you are after. Because the v2 array metadata requires `order`, we do have to check the config setting if no `order` was passed, but we no longer apply a default on top of that.

d-v-b (approved):
lgtm!

PR #2418, opened 2024-10-20
test(ci): add test environment for upstream dependencies

jhamman:
Adds a new CI job to test against upstream dependencies dev versions (nightly wheels or HEAD). 

Closes https://github.com/zarr-developers/zarr-python/issues/1496

TODO:


pyproject.toml
```diff
@@ -183,6 +183,35 @@ features = ['docs']
...
+    'pytest',
+    'pytest-cov',
+    'pytest-asyncio',
+    'moto[s3]',
+]
+
+# help! this is not working for numpy :(
+[tool.hatch.envs.upstream.env-vars]
+PIP_EXTRA_INDEX_URL = "https://pypi.anaconda.org/scientific-python-nightly-wheels/simple"
+PIP_PRE = "1"
```
jhamman, line 207: I could use some help here. Maybe @d-v-b, @dcherian, or @jakirkham have ideas.

This index is what we want to be using for NumPy (and eventually numcodecs) but something is not right. I can only get this to work if I also a lower bound (`>=2.2.0.dev0`) to the numpy dependency.

pyproject.toml
```diff
@@ -183,6 +183,35 @@ features = ['docs']
 build = "cd docs && make html"
 serve = "sphinx-autobuild docs docs/_build --host 0.0.0.0"
 
+[tool.hatch.envs.upstream]
+dependencies = [
+    'numpy>=2.2.0.dev0',
+    'numcodecs @ git+https://github.com/zarr-developers/numcodecs',
```
jhamman, line 189: It would be great if we could get `numcodecs` up on the scientific-python-nightly-wheels index...

https://github.com/zarr-developers/numcodecs/issues/313

pyproject.toml
dstansby: https://scientific-python.org/specs/spec-0004/#test-with-nightly-wheels suggests explicitly installing the development dependencies like

```
python -m pip install --pre --upgrade --extra-index-url https://pypi.anaconda.org/scientific-python-nightly-wheels/simple numpy scipy
```

jhamman: Yep. I read that. My goal here is to put configure everything in our hatch env configuration but if we can't get that to work, we may have to do some extra installs in the CI action itself.

jhamman: Actually, [4c9619b](https://github.com/zarr-developers/zarr-python/pull/2418/commits/4c9619b29a95adf5fafcfefcaa8d48cb11ddf267) seems to have worked!

jhamman:
Just tagged a few folks for review here. I've got some good momentum on testing/CI so would appreciate getting this in so I can move on to other related tasks.

d-v-b (approved):

PR #2423, opened 2024-10-21
test(ci): Add CI for minimum supported dependency versions

jhamman:
This builds on #2418, adding another env for the minimum supported version of all required dependencies.

If we follow [spec 0000](https://scientific-python.org/specs/spec-0000/), we would support older versions of dependencies going back 2 years (3 years for Python). 

Interestingly, trying this for all our dependencies turned up a few things that are worth looking at:

1. We need numcodecs 0.13 (instead of 0.11) for the zstd codec
2. We need typing extensions 4.5 for `@deprecated` and 4.6 for `Buffer`

At this point, I don't see any major issue with bumping these dependencies but its nice to know what we actually support.


d-v-b (approved):
Thanks for doing this!

Issue #1888, opened 2024-05-17
2.18.1: sphinx warnings `reference target not found`


kloczek:
### Zarr version

2.18.1

### Numcodecs version

0.12.1

### Python Version

3.10.14

### Operating System

Linux

### Installation

From autogenerated from git tag tar ball.

### Description

First of all currently it is not possible to use straight `sphinx-build` command to build documentation out of source tree
<details>

```console
+ /usr/bin/sphinx-build -n -T -b man docs build/sphinx/man
Running Sphinx v7.3.7

Traceback (most recent call last):
  File "/usr/lib/python3.10/site-packages/sphinx/config.py", line 509, in eval_config_file
    exec(code, namespace)  # NoQA: S102
  File "/home/tkloczko/rpmbuild/BUILD/zarr-python-2.18.1/docs/conf.py", line 24, in <module>
    import zarr
ModuleNotFoundError: No module named 'zarr'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/usr/lib/python3.10/site-packages/sphinx/cmd/build.py", line 332, in build_main
    app = Sphinx(args.sourcedir, args.confdir, args.outputdir,
  File "/usr/lib/python3.10/site-packages/sphinx/application.py", line 211, in __init__
    self.config = Config.read(self.confdir, confoverrides or {}, self.tags)
  File "/usr/lib/python3.10/site-packages/sphinx/config.py", line 309, in read
    namespace = eval_config_file(filename, tags)
  File "/usr/lib/python3.10/site-packages/sphinx/config.py", line 522, in eval_config_file
    raise ConfigError(msg % traceback.format_exc()) from exc
sphinx.errors.ConfigError: There is a programmable error in your configuration file:

Traceback (most recent call last):
  File "/usr/lib/python3.10/site-packages/sphinx/config.py", line 509, in eval_config_file
    exec(code, namespace)  # NoQA: S102
  File "/home/tkloczko/rpmbuild/BUILD/zarr-python-2.18.1/docs/conf.py", line 24, in <module>
    import zarr
ModuleNotFoundError: No module named 'zarr'


Configuration error:
There is a programmable error in your configuration file:

Traceback (most recent call last):
  File "/usr/lib/python3.10/site-packages/sphinx/config.py", line 509, in eval_config_file
    exec(code, namespace)  # NoQA: S102
[...]

PR #2424, opened 2024-10-21
Build the docs with nitpicky = True

dstansby:
xref https://github.com/zarr-developers/zarr-python/issues/1888. This should fail the doc build on broken links. Perhaps it's not worth fixing these on the `v2` branch, but I thought we could at least see how bad it is.


dstansby:
Going to close this since I don't think it's high priority to fix for v2. If someone else wants to fix any of the broken links, PRs welcome.

Issue #2420, opened 2024-10-20
[v3]: Add Store method for getting the size of an item.


TomAugspurger:
### Zarr version

v3

### Numcodecs version

n/a

### Python Version

n/a

### Operating System

n/a

### Installation

n/a

### Description

In https://github.com/zarr-developers/zarr-python/pull/2400, we need a method for getting the size of some object from a Store to support `.info`. The cleanest way to do this is a new method in the Store API.

We could implement an inefficient version that basically does

```python
def size(self, key: str) -> int:
    return len(await self.get(key))
```

Many backends will have more efficient methods for getting the size of a file / object without actually reading it.

### Steps to reproduce

.

PR #2426, opened 2024-10-21
Added Store.getsize

TomAugspurger:
Closes https://github.com/zarr-developers/zarr-python/issues/2420

One difference from [Zarr v2](https://zarr.readthedocs.io/en/stable/api/storage.html), its `getsize` seemed to return `-1` if the concrete backend didn't provide a `getsize` method. I think returning a "bad" integer like from a function that returns integers is dangerous. I've implemented a slow but correct default that just reads the object and calls `len` on the bytes.

[Description of PR]

TODO:


src/zarr/abc/store.py
```diff
@@ -386,6 +388,32 @@ async def _get_many(
...
+
+        Raises
+        ------
+        FileNotFoundError
+            When the given key does not exist in the store.
+        """
+        # Note to implementers: this default implementation is very inefficient since
+        # it requires reading the entire object. Many systems will have ways to get the
+        # size of an object without reading it.
+        value = await self.get(key, prototype=default_buffer_prototype())
```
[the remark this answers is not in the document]
d-v-b: I think there's a good reason to avoid dynamic default arguments, e.g. function invocations (and I think there's a linting rule that complains about such things). But that doesn't change the fact that `default_buffer_prototype()` _is_ the default, which makes it annoying to specify. One idea that came up in the past was to statically associate a store with a buffer prototype, e.g. by making the store class generic, but this would reduce the flexibility of the store... although whether we realistically need the same store instance to return multiple different types of buffers is yet to be seen.

d-v-b (approved):
this looks good, and I like the safe, slow default over returning -1

src/zarr/abc/store.py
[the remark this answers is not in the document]
TomAugspurger: Would you be OK with `defualt_buffer_prototype` being an optional argument (`| None`, with a default of None, which means use `default_buffer_prototype()`?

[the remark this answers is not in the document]
d-v-b: yeah, some value like `None` or the literal string `"default"` would be perfect for this, at least in my opinion.

src/zarr/storage/remote.py
```diff
@@ -301,3 +301,16 @@ async def list_prefix(self, prefix: str) -> AsyncGenerator[str, None]:
...
+        info = await self.fs._info(path)
+
+        size = info.get("size")
+
+        if size is None:
+            # Not all filesystems support size. Fall back to reading the entire object
+            return await super().getsize(key)
+        else:
+            # fsspec doesn't have typing. We'll need to assume this is correct.
+            return cast(int, size)
```
jhamman, line 316: would this be better?

```suggestion
            return int(size)
```

src/zarr/abc/store.py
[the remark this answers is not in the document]
TomAugspurger: Two small bits of awkwardness:

1. We'll now require Store implementors to *read the documentation* to understand what the value of `None / "default"` for `prototype` means when they're implementing `get`. I think that's fine, but merits documenting.
2. I might have missed it, but IIUC `zarr.core` is private, but `zarr.core.buffer` isn't re-exported from anywhere else. I'll add it to `zarr.buffer` unless anyone objects.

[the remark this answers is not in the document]
kylebarron: > One idea that came up in the past was to statically associate a store with a buffer prototype

This thread came up in #1661, where I was struggling to understand the purpose of the prototype argument.

I think it would be nice to either statically associate stores with buffer prototypes or remove the prototypes altogether and let a store return a protocol instead of a specific class.

TomAugspurger:
I might be confusing myself, but I think this implementation might not be what we want... I think what users want (like us in https://github.com/zarr-developers/zarr-python/pull/2400) is the size of an Array in storage, not the size of a particular key.  I guess we could do something like generate all the keys for a given array and then call `store.getsize` with each of those keys...

So maybe we do need this, since the store is knows (or can figure out) what bytes are actually stored for a given array. But we also need a bit on top of it to bring it to the array level.

d-v-b:
> I guess we could do something like generate all the keys for a given array and then call `store.getsize` with each of those keys...

In case you want to go this direction, [this method](https://github.com/zarr-developers/zarr-python/blob/bc588a760a804f783c4242d4435863a43a5f3f9f/src/zarr/core/array.py#L870) is designed for exactly such a use case

jhamman:
I'll throw an idea into the mix. We probably want two things:

- `Store.get_size(key: str) ->  int`
- `Store.get_size_dir(path: str) -> int`

Of course, `list_dir` + `get_size` would be the same as `get_size_dir` but some stores will be able to provide a fast path for the dir size.

TomAugspurger:
Thanks. Looking at how Icechunk would implement `getsize` is what prompted my question so I can se how a `getsize_dir` makes sense there.

Would you expect the size of metadata documents to show up in total for `getsize_dir`?

TomAugspurger:
I've pushed an update that adds a `getsize_prefix`, but am having second thoughts about whether this is worth adding to the API. It's not clear to me that a Store will always have a well-defined size for an array (things like references or store-level sharding complicate things), and so maybe it doesn't make sense to add it to the store interface.

src/zarr/abc/store.py
```diff
@@ -386,6 +391,60 @@ async def _get_many(
...
+        return len(value)
+
+    async def getsize_prefix(self, prefix: str) -> int:
+        """
+        Return the size, in bytes, of all values under a prefix.
+
+        Parameters
+        ----------
+        prefix : str
+            The prefix of the directory to measure.
```
paraseba, line 427: Can we offer implementers the following in documentation?:

This function will be called by zarr using a `prefix` that is the path of a group, an array, or the root. Implementations can choose to do undefined behavior when that is not the case.

src/zarr/abc/store.py
```diff
@@ -386,6 +391,60 @@ async def _get_many(
...
+        Store.getsize
+
+        Notes
+        -----
+        ``getsize_prefix`` is just provided as a potentially faster alternative to
+        listing all the keys under a prefix calling :meth:`Store.getsize` on each.
+        """
+        keys = [x async for x in self.list_prefix(prefix)]
+        sizes = await gather(*[self.getsize(key) for key in keys])
+        return sum(sizes)
```
paraseba, line 446: This materializes the full list of keys in memory, can we maintain the generator longer to avoid that?

Also, this has unlimited concurrency, for a potentially very large number of keys. It could easily create millions of async tasks. We should probably run in chunks limited by the value of the `concurrency` setting.

paraseba: See [`concurrent_map`](https://github.com/zarr-developers/zarr-python/blob/a82d0476656cd2de12145abd899c81eb8c726ed3/src/zarr/core/common.py#L52) for an example

src/zarr/abc/store.py
TomAugspurger: Sure... I was hoping we could somehow ensure that we don't call it with anything other than a group / array / root path, but users can directly use `Store.getsize_prefix` and they can do whatever. 

LMK if you want any more specific guidance on what to do (e.g. raise a `ValueError`). I'm hesitant about trying to force required exceptions into an ABC / interface.

TomAugspurger: > I'm hesitant about trying to force required exceptions into an ABC / interface.

And now I'm noticing that I've done exactly that in `getsize` :)

TomNicholas:
> am having second thoughts about whether this is worth adding to the API.

Something like this interface

> `Store.get_size(key: str) -> int`

would be very useful for virtualizarr, as then we can easily and efficiently learn the byte range lengths of all objects in a store, in order to ingest existing zarr as virtual zarr.

src/zarr/abc/store.py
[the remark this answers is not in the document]
TomAugspurger: I've removed the buffer related changes from this PR (xref https://github.com/zarr-developers/zarr-python/issues/2473, which might bear on this parameter).

src/zarr/abc/store.py
TomAugspurger: 5f1d0360d3b985c1cb6f0678fbdaa4e4991f0742 removed support for `AsyncIterable` in `concurrent_map`, replacing it with a TODO.

I think there's some discussion around improving our use of asyncio to handle cases like this (using queues to mediate task producers like list_prefix and consumers like `getsize`) that will address this.

jhamman (approved):
LGTM. Thanks @TomAugspurger :)

src/zarr/abc/store.py
```diff
@@ -396,6 +400,70 @@ async def _get_many(
         for req in requests:
             yield (req[0], await self.get(*req))
 
+    async def getsize(self, key: str) -> int:
```
jhamman, line 403: @TomAugspurger - can we add these two new methods to the `LoggingStore`?

jhamman (changes requested):

TomAugspurger:
Should be all set now.

jhamman (approved):

Issue #2015, opened 2024-07-05
write behavior for empty chunks


d-v-b:
In v2, at array access time it is possible to set whether empty chunks (defined as chunks that are entirely `fill_value`) should be written to storage or skipped. This is an extremely useful feature for high-latency storage backends, or in any context where too many objects in storage is burdensome. 

We don't support this in v3 yet, but we should. How should we do it? I will throw out a few options in order of practicality:

- emulate v2: provide a keyword argument like `write_empty_chunks` when accessing an array. All chunk writes from that array will be affected. 
- put the `write_empty_chunks` setting in a global config. All chunk writes from all arrays in a session will be affected by the config parameter.
- design an API for array IO wherein IO is wrapped in a context that can be parametrized, e.g. with a context manager, and one of those parameters is the write_empty_chunks-ness of the write transaction. Highly speculative.

The first option seems pretty expedient, and I don't think we had a lot of problems with this approach in v2. The only drawback is that if people want the same array to exhibit conditional write_empty_chunks behavior, then they might need something like the second approach, which has its own drawbacks IMO (i'm not a big fan of mutable global state). 

I would propose that we emulate v2 for now (i.e., make `write_empty_chunks` a keyword argument to array access) and note any friction this causes, and consider ways to alleviate that in a subsequent design refresh if the friction is severe.

cc @constantinpape

constantinpape:
My 2cents: I personally think this should be the default behavior. We have use-cases where writing empty chunks is extremely bad (as in bringing us over IO node quota immediately and killing everyone's jobs) and I am hesitant to give a library to students which has this as a default behavior. 

Besides this, I think that both option 1 and 2 sound ok to me.

d-v-b:
> My 2cents: I personally think this should be the default behavior.

I tend to agree, the one counter-point is that for dense arrays, i.e. those arrays where all chunks have useful data in them, inspecting each chunk adds some overhead to writing. But I think our design should err on the side of using up some extra CPU cycles over potentially generating massive numbers of useless files.

rabernat:
> inspecting each chunk adds some overhead to writing

Here is some data on that from my laptop

| dtype | bytes | time (s) | throughput (MB/s) |
|--|--|--|--|
| i2 | 2000 | 1.40e-04 | 14.31 |
| i2 | 2000000 | 2.53e-04 | 7897.34 |
| i2 | 2000000000 | 6.26e-01 | 3195.49 |
[...]

PR #2429, opened 2024-10-22
Feat/write empty chunks

### LINKED ISSUE #2409: [v3] support write_empty_chunks
### Zarr version

3.0.0.beta

### Numcodecs version

0.13

### Python Version

3.11

### Operating System

Mac

### Installation

pip

### Description

`write_empty_chunks` was a very useful write-time optimization that we should bring back to 3.0

https://zarr.readthedocs.io/en/stable/tutorial.html#empty-chunks

### Steps to reproduce

```python
In [23]: a = zarr.create(shape=(10, 10), chunks=(5, 5), write_empty_chunks
    ...: =True)
/Users/jhamman/miniforge3/envs/icechunk-demo/lib/python3.12/site-packages/zarr/core/sync.py:100: RuntimeWarning: write_empty_chunks is not yet implemented
```

### Additional output

_No response_

d-v-b:
This PR adds a boolean `array.write_empty_chunks` value to the global config, and uses this value to control whether chunks that are "empty", i.e. filled with values equivalent to the array's fill value, are written to storage.

In [`zarr-python` 2.x](https://github.com/zarr-developers/zarr-python/pull/738), `write_empty_chunks` was a property of an `Array` that users specified when creating the `Array` object. This had pros and cons which I'm happy to discuss if people are interested, but the tl;dr is that the _cons_ of that approach are driving my decision in this PR to make `write_empty_chunks` a global runtime property accessible via the config API. 

Usage looks something like this (`donfig` experts please correct me if there's a better way): 
```python
with config.set({"array.write_empty_chunks": write_empty_chunks}):
    arr[:] = fill_value
```

If people hate this, then we can definitely change this API. I'm very open to discussion here.

Also worth noting:

Our [check for whether a chunk is equal to the fill value](https://github.com/zarr-developers/zarr-python/blob/a9d6d74c4a4d4efcdb64015919d0a5685a7e0239/src/zarr/core/buffer/core.py#L462-L473) is pretty inefficient -- it's allocating a new array for every check invocation. This can definitely be made more efficient, in a stupid way by caching an all-fill-value chunk on the array instance and using that for the comparison, or a smarter way by doing the `(chunk, fill_value)` comparison without allocating a new array. But I think this is an effort for a separate PR.

closes #2409 

TODO:


normanrz (approved):
I think this is a good approach. Should we add some backwards compatibility thing for the `write_empty_chunks` kwarg in zarr.open?

TomAugspurger:
> Also worth noting:

Our [check for whether a chunk is equal to the fill value](https://github.com/zarr-developers/zarr-python/blob/a9d6d74c4a4d4efcdb64015919d0a5685a7e0239/src/zarr/core/buffer/core.py#L462-L473) is pretty inefficient

I think that `broadcast_arrays` is smart in a couple of ways when it's an array, scalar operation.

```python
In [41]: x = np.random.randn(10, 10, 10)

In [42]: x2, y = np.broadcast_arrays(x, 0)

In [43]: x is x2  # No copy of the array is created
Out[43]: True

In [44]: y.base  # Only a single value is allocated for the fill value array data.
```

 
It'd be nice to avoid the equality check when writing, at least under some circumstances, but I haven't thought of an easy way to do that.

d-v-b:
> Should we add some backwards compatibility thing for the `write_empty_chunks` kwarg in zarr.open?

Are you thinking of something like a warning to guide people to use the configuration approach, if they pass in `write_empty_chunks`?

normanrz:
Yes, a warning and maybe even setting the config for them?

dcherian:
> It'd be nice to avoid the equality check when writing

I've forgotten the code path now, but if zarr created the empty chunk using `np.broadcast_to(self.fill_value, chunk_shape)` when we might just check equality of `.base` or something like that.

jhamman (changes requested):

src/zarr/codecs/pipeline.py
```diff
@@ -331,6 +331,7 @@ async def write_batch(
         value: NDBuffer,
         drop_axes: tuple[int, ...] = (),
     ) -> None:
+        write_empty_chunks = config.get("array.write_empty_chunks") == True  # noqa: E712
```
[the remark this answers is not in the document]
d-v-b: > I'd rather make this a property of the Array so that we can guarantee consistent write behavior after an Array has been initialized.

If we do that, then we also require that users create a brand new array if they want to write just some parts of the data with different empty chunks handling (or we introduce `write_empty_chunks` as a mutable attribute, which I would rather avoid)

src/zarr/api/asynchronous.py
```diff
@@ -850,14 +855,12 @@ async def create(
             raise ValueError(
                 "dimension_separator is not supported for zarr format 3, use chunk_key_encoding instead"
             )
-        else:
-            warnings.warn(
-                "dimension_separator is not yet implemented",
```
normanrz, line 855: This is actually not true, because it is implemented for v2 arrays.

src/zarr/core/array.py
```diff
@@ -465,6 +473,8 @@ async def create(
         else:
             _chunks = normalize_chunks(chunk_shape, shape, dtype_parsed.itemsize)
 
+        config_parsed = ArrayConfig(order=order, write_empty_chunks=write_empty_chunks)
```
normanrz, line 476: Should we warn here as well?

normanrz:
I added an `ArrayConfig` class that holds `order` and `write_empty_chunks`. Both are initialized when the Array object is initialized. By default, the values from the global config are used.

I added deprecation warnings for both the `order` and `write_empty_chunks` kwargs.

d-v-b:
@normanrz do you think we should surface the `.config` attribute on the `Array` class?

src/zarr/codecs/pipeline.py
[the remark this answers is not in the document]
d-v-b: i think this is resolved by making `write_empty_chunks` an attribute of the array config

normanrz:
> @normanrz do you think we should surface the `.config` attribute on the `Array` class?

I was thinking the same thing. We could expose it as `_config`.

normanrz:
I renamed `AsyncArray.config` to `AsyncArray._config` in https://github.com/zarr-developers/zarr-python/pull/2429/commits/08a2d529a0f0e12e905206e1d5d95d517253a1fb

d-v-b:
@normanrz do you think we should add `config` as a keyword argument to array creation routines? The type of this parameter would be `ArrayConfig | TypedDictThat'sLikeAnArrayConfig` This gives people who are currently controlling `write_empty_chunks` via a keyword argument a simple way to handle the deprecation warning

d-v-b:
a catch-all `config` kwarg also lets us potentially grow the space of runtime array configuration without needing to add a lot of weight to the array constructors

src/zarr/core/array_spec.py
```diff
@@ -5,40 +5,68 @@
...
     from zarr.core.buffer import BufferPrototype
     from zarr.core.common import ChunkCoords
 
 
+@dataclass(frozen=True)
+class ArrayConfig:
+    order: MemoryOrder
+    write_empty_chunks: bool
+
+    def __init__(
```
d-v-b, line 27: this design prevents us from having `None` as a valid value for a configuration parameter, because it will not possible to distinguish "parameter was unset" from "parameter was set to `None`". A solution is to use a classmethod that takes a dictionary where some keys may not be present, which unambiguously express a missing parameter. I am working on this.

normanrz:
> @normanrz do you think we should add `config` as a keyword argument to array creation routines? The type of this parameter would be `ArrayConfig | TypedDictThat'sLikeAnArrayConfig` This gives people who are currently controlling `write_empty_chunks` via a keyword argument a simple way to handle the deprecation warning

Yeah. That could work

d-v-b:
> > @normanrz do you think we should add `config` as a keyword argument to array creation routines? The type of this parameter would be `ArrayConfig | TypedDictThat'sLikeAnArrayConfig` This gives people who are currently controlling `write_empty_chunks` via a keyword argument a simple way to handle the deprecation warning
> 
> Yeah. That could work

nice, i'm hacking on this rn

d-v-b:
After working on this a bit, I think we can't deprecate the `order` parameter, because it's part of the v2 array metadata document, and that means it's orthogonal to the runtime memory order parameter. this means there is no contradiction if a user creates a v2 array with `order=F` and `config={'order': 'C'}`, although we will need to ensure that this works properly (the memory order would need to be inspected prior to writing to ensure that memory order conversions work)

by contrast, `write_empty_chunks` is purely runtime.

normanrz:
Ok. Are you working on this or do you want help?

d-v-b:
i'm working on it!

d-v-b:
- I removed `write_empty_chunks` from the `Array.create` signature. Instead, pass `config={'write_empty_chunks': <value>}`. The higher-level API creation routines still take `write_empty_chunks`, but it's deprecated, and those values are used to create an `ArrayConfig` with the user-provided right `write_empty_chunks` value.
- `ArrayConfig` has a `from_dict` method that takes a dict; any fields of `ArrayConfig` that are missing from this dict will be pulled from the global config object
- we cannot deprecate the `order` keyword argument, because it's a core part of v2 metadata. But we emit a warning if someone specifies it for a v3 array, because the only degree of freedom there is the `config.order` parameter.

src/zarr/core/array.py
```diff
@@ -439,7 +446,11 @@ async def create(
             The dimension separator (default is ".").
             V2 only. V3 arrays should use ``chunk_key_encoding`` instead.
         order : Literal["C", "F"], optional
-            The order of the array (default is specified by ``array.order`` in :mod:`zarr.core.config`).
+            The memory of the array (default is "C").
+            If ``zarr_format`` is 2, this parameter sets the memory order of the array.
+            If `zarr_format`` is 3, then this parameter is deprecated, because memory order
+            is a runtime parameter for zarr 3 arrays. The recommended way to specify the memory
+            order for zarr v3 arrays is via the ``config`` parameter, e.g. ``{'config': 'C'}``.
```
normanrz, line 453:
```suggestion
            is a runtime parameter for Zarr 3 arrays. The recommended way to specify the memory
            order for Zarr 3 arrays is via the ``config`` parameter, e.g. ``{'config': 'C'}``.
```

src/zarr/core/array.py
```diff
@@ -1545,7 +1580,11 @@ def create(
             The dimension separator (default is ".").
             V2 only. V3 arrays should use ``chunk_key_encoding`` instead.
         order : Literal["C", "F"], optional
-            The order of the array (default is specified by ``array.order`` in :mod:`zarr.core.config`).
+            The memory of the array (default is "C").
+            If ``zarr_format`` is 2, this parameter sets the memory order of the array.
+            If `zarr_format`` is 3, then this parameter is deprecated, because memory order
+            is a runtime parameter for zarr 3 arrays. The recommended way to specify the memory
+            order for zarr v3 arrays is via the ``config`` parameter, e.g. ``{'order': 'C'}``.
```
normanrz, line 1587:
```suggestion
            is a runtime parameter for Zarr 3 arrays. The recommended way to specify the memory
            order for Zarr 3 arrays is via the ``config`` parameter, e.g. ``{'order': 'C'}``.
```

normanrz (approved):

jni:
Is there a quick way (documentation/docstrings) for users to know that the new default for `write_empty_chunks` is false? I just read a lot of code before I was able to figure that out. 😂 

@d-v-b I'm also not entirely sure why you went with option 2 from #2015 (re: mutable global state), though it's alleviated by passing config to the write functions + context managers. I actually quite like the current implementation despite having the same reservation, I'd just love to see a fully-worked-out rationale for the chosen option, maybe in an update to the top-level description of this PR. (Which could also show the other ways of setting config other than the context manager.)

d-v-b:
we definitely need an easy way to document that the default value is `False`. Tying the default value to the config, which users can change, makes it a bit risky to write "the default value is `False`" in a docstring, because a user could change the config and thus invalidate that docstring. We could say "unless the global configuration is modified, the default value is `False`", and ensure that if we ever change the default global config values, we also change that docstring accordingly. I think we could also express the `config` in a rich repr of the array, so that it's easy to quickly check without hunting through array attributes.

As for the explanation of the merged PR, I'd rather not change the top-level description of the PR at this point but I can add my thoughts here:
We create a new `config` attribute for arrays, and put `write_empty_chunks` in the `config` attribute, and the array config can be transiently altered in a context manager. I'm pretty confident that this is a good fit, because even for a single array,  a user might want to skip the "is this chunk empty" check for a select subset of chunks. Put another way, in abstract terms the "is this chunk empty" check is scoped to a single chunk. We don't have a per-chunk configuration, so `write_empty_chunks` has to live at a higher level of abstraction -- at the array level -- even though it doesn't exactly belong there. Allowing the config to be transiently overridden without mutating the array object itself lets users exert fine-grained control over how their chunks are written -- they can ensure that a subset of the chunks of a given array are written with `write_empty_chunks=False`, and another subset with `True`. Even if nobody uses this "override the config for a bit" functionality specifically for `write_empty_chunks`, I think it could be generally useful, and it's technically a fit for `write_empty_chunks`.

The default value of the `write_empty_chunks` parameter comes from the global zarr config. I'm less confident about this decision, but I think it's worth trying out for now. The reason to make the default values context-sensitive is so that users could define an array config in a _group_, and then the arrays accessed via that group would inherit the group's array config. As long as we support APIs for accessing arrays by name only via `group.__getitem__`, we need a way to define the config that those arrays use, and this "config inheritance" addresses that. This PR does not add this functionality, but sets the stage for it.

Of course another solution would be to deprecate `group.__getitem__` for accessing arrays / groups, and instead recommend a method like `get_array(name: str, config: ArrayConfig)`. but I figured removing `__getitem__` would break a lot of code.

Issue #2191, opened 2024-09-17
[v3] deleting a group does not delete subgroups


dcherian:
### Zarr version

3

### Numcodecs version

?

### Python Version

?

### Operating System

?

### Installation

?

### Description

deleting a group does not delete subgroups.

Discovered in #2189 

### Steps to reproduce

```python
import zarr
from zarr.store import MemoryStore

store = MemoryStore(mode="w")
g1 = zarr.group(store=store)
g1.create_group("0")
g1.create_group("0/0")
del g1["0"]
g1["0/0"] # succeeds
```

zoj613:
Looks like the `MemoryStore` implementation is missing a `erase_prefix` function to allow proper deletion of a group and its children.

jhamman:
I have a fix for this in the works.

Issue #2359, opened 2024-10-14
[v3] Re-opening an existing RemoteStore V3 group causes the first child to be deleted


rabernat:
This is a weird and unexpected behavior 

```python
import xarray as xr
import zarr
import s3fs
print(xr.__version__, zarr.__version__)
# -> ('0.9.7.dev3734+g26081d4f', '3.0.0b1.dev8+g9bbfd88')

s3 = s3fs.S3FileSystem()
# replace with your favorite bucket
target_path = "icechunk-test/ryan/zarr-v3/test-92"
store = zarr.storage.RemoteStore(s3, mode="w", path=target_path)

# create a group with two children
group = zarr.open_group(store=store, mode="w", zarr_format=3)
group.create("A", shape=1)
group.create("B", shape=1)
print(group.members())
#  (('A',
#  <Array <RemoteStore(S3FileSystem, icechunk-test/ryan/zarr-v3/test-era5-v3-92)>/A shape=(1,) dtype=float64>),
#  ('B',
#   <Array <RemoteStore(S3FileSystem, icechunk-test/ryan/zarr-v3/test-era5-v3-92)>/B shape=(1,) dtype=float64>))

# re-open the group
store2 = zarr.storage.RemoteStore(s3, mode="w", path=target_path)
group = zarr.open_group(store=store2, mode="w", zarr_format=3)
print(group.members())
# (('B',
#  <Array <RemoteStore(S3FileSystem, icechunk-test/ryan/zarr-v3/test-era5-v3-92)>/B shape=(1,) dtype=float64>),)
```

Array `A` has disappeared! That's bad!

This might have something to do with #2338. Really not sure. I haven't gotten to the bottom of it, but I thought I would post a report anyway.

jhamman:
I understand what is going on here. I believe this is the intended behavior for our new stores but am open to the idea that we should change it!

In Zarr-Python 2, stores were just a mapping interface on top of a storage backend. No access mode was applied at the store level - _except_ for stores that required an access mode (`ZipStore`, `DBMStore`). Ironically, the `FSStore` also took a `mode` argument but `w` there just meant "for writing" -- not overwrite.

[...]

Issue #2108, opened 2024-08-22
[v3] Implement store methods for erase and rename operations


jhamman:
We have not yet implemented bulk erase (dir or prefix) or rename methods for the V3 store interface. These are both needed to support efficient evolution of existing zarr hierarchies.  Marking this as a blocker for 3.0.

jhamman:
We implemented erase (`del_prefix`) but we have not done anything for `rename` yet. That remains a TODO here.

ziw-liu:
The lack of `move` has broken my code and it would be nice to not have to re-implement this feature downstream.

PR #2430, opened 2024-10-22
Feature: store learns to delete prefixes when overwriting/creating hierarchy nodes

jhamman:
This PR implements `delete_dir` teaches the Group and Array classes to use it when overwriting nodes in a hierarchy. 

Also included here:
~- `delete_prefix` - not used but perhaps we should just use it instead of `delete_dir`~
- changes `list_prefix` to return absolute paths instead of stripping `prefix`. Given that `list_prefix` was unused before, this wasn't very disruptive but calling it out so others can help decide if this is what we want.

Still needs:
- [x] new store tests for `delete_dir`
- [x] new group tests for edge cases around `delete_dir`

closes https://github.com/zarr-developers/zarr-python/issues/2191
first steps toward https://github.com/zarr-developers/zarr-python/issues/2108 and #2359 

TODO:

### DIFF
--- a/src/zarr/abc/store.py
+++ b/src/zarr/abc/store.py
@@ -168,7 +168,7 @@ class Store(ABC):
 
         Returns
         -------
-        store:
+        store
             A new store of the same type with the new mode.
 
         Examples
--- a/src/zarr/abc/store.py
+++ b/src/zarr/abc/store.py
@@ -384,7 +384,7 @@ class Store(ABC):
             if not prefix.endswith("/"):
                 prefix += "/"
             async for key in self.list_prefix(prefix):
-                await self.delete(f"{key}")
+                await self.delete(key)
         else:
             async for key in self.list_dir(prefix):
                 await self.delete(f"{prefix}/{key}")
--- a/src/zarr/abc/store.py
+++ b/src/zarr/abc/store.py
@@ -371,6 +371,36 @@ class Store(ABC):
         """
         ...
 
+    async def delete_dir(self, prefix: str, recursive: bool = True) -> None:
+        """
+        Remove all keys and prefixes in the store that begin with a given prefix.
+        """
+        if not self.supports_deletes:
+            raise NotImplementedError
+        if not self.supports_listing:
+            raise NotImplementedError
+        self._check_writable()
+        if recursive:
+            if not prefix.endswith("/"):
+                prefix += "/"
+            async for key in self.list_prefix(prefix):
+                await self.delete(f"{key}")
+        else:
+            async for key in self.list_dir(prefix):
+                await self.delete(f"{prefix}/{key}")
+
+    async def delete_prefix(self, prefix: str) -> None:
+        """
+        Remove all keys in the store that begin with a given prefix.
+        """
+        if not self.supports_deletes:
+            raise NotImplementedError
+        if not self.supports_listing:
+            raise NotImplementedError
+        self._check_writable()
+        async for key in self.list_prefix(prefix):
+            await self.delete(f"{key}")
+
     def close(self) -> None:
         """Close the store."""
         self._is_open = False
--- a/src/zarr/abc/store.py
+++ b/src/zarr/abc/store.py
@@ -342,8 +342,8 @@ class Store(ABC):
     @abstractmethod
     def list_prefix(self, prefix: str) -> AsyncGenerator[str, None]:
         """
-        Retrieve all keys in the store that begin with a given prefix. Keys are returned as
-        absolute paths (i.e. including the prefix).
+        Retrieve all keys in the store that begin with a given prefix. Keys are returned relative 
+        to the root of the store.
 
         Parameters
         ----------
--- a/src/zarr/abc/store.py
+++ b/src/zarr/abc/store.py
@@ -342,7 +342,7 @@ class Store(ABC):
     @abstractmethod
     def list_prefix(self, prefix: str) -> AsyncGenerator[str, None]:
         """
-        Retrieve all keys in the store that begin with a given prefix. Keys are returned relative 
+        Retrieve all keys in the store that begin with a given prefix. Keys are returned relative
         to the root of the store.
 
         Parameters
--- a/src/zarr/abc/store.py
+++ b/src/zarr/abc/store.py
@@ -342,8 +342,8 @@ class Store(ABC):
     @abstractmethod
     def list_prefix(self, prefix: str) -> AsyncGenerator[str, None]:
         """
-        Retrieve all keys in the store that begin with a given prefix. Keys are returned with the
-        common leading prefix removed.
+        Retrieve all keys in the store that begin with a given prefix. Keys are returned as
+        absolute paths (i.e. including the prefix).
 
         Parameters
         ----------
@@ -371,7 +371,7 @@ class Store(ABC):
         """
         ...
 
-    async def delete_dir(self, prefix: str, recursive: bool = True) -> None:
+    async def delete_dir(self, prefix: str) -> None:
         """
         Remove all keys and prefixes in the store that begin with a given prefix.
         """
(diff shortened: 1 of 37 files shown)


src/zarr/abc/store.py
```diff
@@ -371,6 +371,36 @@ def list_dir(self, prefix: str) -> AsyncGenerator[str, None]:
...
+        if not self.supports_deletes:
+            raise NotImplementedError
+        if not self.supports_listing:
+            raise NotImplementedError
+        self._check_writable()
+        if recursive:
+            if not prefix.endswith("/"):
+                prefix += "/"
+            async for key in self.list_prefix(prefix):
+                await self.delete(f"{key}")
```
d-v-b, line 387: do we need an fstring here?

src/zarr/api/asynchronous.py
```diff
@@ -712,7 +712,7 @@ async def create(
     dtype: npt.DTypeLike | None = None,
     compressor: dict[str, JSON] | None = None,  # TODO: default and type change
     fill_value: Any | None = 0,  # TODO: need type
-    order: MemoryOrder | None = None,  # TODO: default change
+    order: MemoryOrder | None = None,
```
d-v-b, line 715: weird that these changes from #2417 are appearing here.

src/zarr/core/array.py
```diff
@@ -554,7 +554,12 @@ async def _create_v3(
         attributes: dict[str, JSON] | None = None,
         exists_ok: bool = False,
     ) -> AsyncArray[ArrayV3Metadata]:
-        if not exists_ok:
+        if exists_ok:
+            if store_path.store.supports_deletes:
+                await store_path.delete_dir(recursive=True)
+            else:
+                await ensure_no_existing_node(store_path, zarr_format=3)
+        else:
```
d-v-b, line 562: [not a blocker for this PR but just an idea triggered by this code]

deleting an entire array can take `O(num_chunks)` time, and I know this has been a pain point for for zarr users including myself. To alleviate this, we should expose two settings:
- one setting for array creation that avoids a recursive delete, and instead only creates / modifies metadata documents
- one setting for chunks that causes zarr to ignore existing invalid chunks and simply overwrite them.

together, these two settings would make creating re-creating arrays in-place very cheap, albeit more likely to contain garbage from old arrays / groups, so if we do this we should also create a routine for cleaning up that garbage.

src/zarr/storage/local.py
```diff
@@ -226,7 +226,9 @@ async def list(self) -> AsyncGenerator[str, None]:
 
     async def list_prefix(self, prefix: str) -> AsyncGenerator[str, None]:
         # docstring inherited
-        to_strip = os.path.join(str(self.root / prefix))
+        to_strip = str(self.root)
+        if prefix.endswith("/"):
+            prefix = prefix[:-1]
```
d-v-b, line 231:
```suggestion
            prefix = prefix.rstrip('/')
```

src/zarr/core/array.py
jhamman: Happy to discuss this idea in a separate issue by my initial reaction is that this sounds like a very dangerous option!

> To alleviate this, we should expose two settings:

The big change in 3.0 is that the store is async. So we can optimize this by more aggressively utilizing concurrency.

d-v-b: > Happy to discuss this idea in a separate issue by my initial reaction is that this sounds like a very dangerous option!

I will open an issue for the idea. I agree that it's opens up a "hard hat zone" in the API. 👷

src/zarr/abc/store.py
```diff
@@ -371,6 +371,36 @@ def list_dir(self, prefix: str) -> AsyncGenerator[str, None]:
...
+        if recursive:
+            if not prefix.endswith("/"):
+                prefix += "/"
+            async for key in self.list_prefix(prefix):
+                await self.delete(key)
+        else:
+            async for key in self.list_dir(prefix):
+                await self.delete(f"{prefix}/{key}")
+
+    async def delete_prefix(self, prefix: str) -> None:
```
jhamman, line 392: Open to input on this one but I do not think this is needed and could be removed.

src/zarr/abc/store.py
```diff
@@ -371,6 +371,36 @@ def list_dir(self, prefix: str) -> AsyncGenerator[str, None]:
         """
         ...
 
+    async def delete_dir(self, prefix: str, recursive: bool = True) -> None:
+        """
+        Remove all keys and prefixes in the store that begin with a given prefix.
```
paraseba, line 376: I think it would make life much easier for `Store` implementers if we declare that `prefix` will always be the path of a group or a array. Otherwise, all implementations need to deal with unnecessary edge cases like `delete_dir("path/to/array/c/0")` or worse.

[Icechunk](https://icechunk.io/) is an example of a store that could very cleanly and efficiently implement the case for a group/array prefix, but it would struggle to handle other prefixes.

src/zarr/abc/store.py
paraseba: I'd delete if we don't need it.

src/zarr/abc/store.py
paraseba: What is the use case for `recursive = False` ? If zarr itself doesn't have one, I'd remove the argument.

src/zarr/abc/store.py
```diff
@@ -371,6 +371,36 @@ def list_dir(self, prefix: str) -> AsyncGenerator[str, None]:
...
+    async def delete_prefix(self, prefix: str) -> None:
+        """
+        Remove all keys in the store that begin with a given prefix.
+        """
+        if not self.supports_deletes:
+            raise NotImplementedError
+        if not self.supports_listing:
+            raise NotImplementedError
+        self._check_writable()
+        async for key in self.list_prefix(prefix):
```
paraseba, line 401: Can we document what's the expected return value of `list_prefix`? Absolute vs. relative, final "/" or not, etc.

src/zarr/abc/store.py
jhamman: I can see a use case for `delete_dir("path/to/array/c/")` (remove all chunks from an array) so I think we should allow that -- even if it makes life a little harder in icechunk.

src/zarr/abc/store.py
jhamman: this is done in the list_prefix docstring

TomAugspurger:
Looks good overall. I've been struggling a bit with whether Stores should be a bit more aware about the context the operation is occurring in (e.g. deleting a group vs a specific chunk of an array; I think this relates to the conversation at https://github.com/zarr-developers/zarr-python/pull/2430#discussion_r1813941831). But I think this is a good improvement on its own.

src/zarr/abc/store.py
paraseba: That's totally fine. Let's document the three possible prefix cases, and say that stores are allowed to do undefined behavior for other prefixes

d-v-b (approved):
looks good. regarding the change to `list_prefix`, I will note that the definition of "absolute paths" is a bit shaky across different stores, since `RemoteStore` will return keys relative to its `path` attribute, and similarly `LocalStore` with its `root` attribute.

jhamman:
> regarding the change to list_prefix, I will note that the definition of "absolute paths" is a bit shaky across different stores

Would defining these as "relative to the root of the store" be clearer?

src/zarr/abc/store.py
```diff
@@ -342,8 +342,8 @@ def list(self) -> AsyncGenerator[str, None]:
     @abstractmethod
     def list_prefix(self, prefix: str) -> AsyncGenerator[str, None]:
         """
-        Retrieve all keys in the store that begin with a given prefix. Keys are returned with the
-        common leading prefix removed.
+        Retrieve all keys in the store that begin with a given prefix. Keys are returned as
+        absolute paths (i.e. including the prefix).
```
jhamman, line 346:
```suggestion
        Retrieve all keys in the store that begin with a given prefix. Keys are returned relative 
        to the root of the store.
```

jhamman:
pre-commit.ci autofix

PR #2431, opened 2024-10-22
test(ci): test on windows and osx

### LINKED ISSUE #1648: [v3] Test environments and CI

jhamman:
This PR adds CI matrix entries for OSX and Windows. 

Closes #1648 

Notes:

There are some interesting failures here.

~- [ ] path handling in local store~
- [x] async timeout error in osx

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,7 +21,7 @@ jobs:
     strategy:
       matrix:
         python-version: ['3.11', '3.12', '3.13']
-        numpy-version: ['1.25', '2']
+        numpy-version: ['1.25', '2.1']
         dependency-set: ["minimal", "optional"]
         os: ["ubuntu-latest"]
         include:
@@ -30,7 +30,7 @@ jobs:
             dependency-set: 'optional'
             os: 'macos-latest'
           - python-version: '3.13'
-            numpy-version: '2'
+            numpy-version: '2.1'
             dependency-set: 'optional'
             os: 'macos-latest'
           - python-version: '3.11'
@@ -38,7 +38,7 @@ jobs:
             dependency-set: 'optional'
             os: 'windows-latest'
           - python-version: '3.13'
-            numpy-version: '2'
+            numpy-version: '2.1'
             dependency-set: 'optional'
             os: 'windows-latest'
     runs-on: ${{ matrix.os }}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -33,14 +33,14 @@ jobs:
             numpy-version: '2.1'
             dependency-set: 'optional'
             os: 'macos-latest'
-          - python-version: '3.11'
-            numpy-version: '1.25'
-            dependency-set: 'optional'
-            os: 'windows-latest'
-          - python-version: '3.13'
-            numpy-version: '2.1'
-            dependency-set: 'optional'
-            os: 'windows-latest'
+          # - python-version: '3.11'
+          #   numpy-version: '1.25'
+          #   dependency-set: 'optional'
+          #   os: 'windows-latest'
+          # - python-version: '3.13'
+          #   numpy-version: '2.1'
+          #   dependency-set: 'optional'
+          #   os: 'windows-latest'
     runs-on: ${{ matrix.os }}
 
     steps:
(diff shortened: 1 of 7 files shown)


martindurant:
> tweak path handling in the LocalStore

fsspec has a decent amount of work for this which you can reuse. Actually, could RemoteStore write to local too to avoid the duplication, or would that be unnecessarily complicated (given the async/sync difference in particular).

jhamman:
> fsspec has a decent amount of work for this which you can reuse.

Can you provide some pointers?

> Actually, could RemoteStore write to local too to avoid the duplication, or would that be unnecessarily complicated (given the async/sync difference in particular).

I don't think this will help us because the `RemoteStore` requires an Async Filesystem which the fsspec local filesystem is not.

martindurant:
fsspec calls https://github.com/fsspec/filesystem_spec/blob/master/fsspec/implementations/local.py#L284 as early as possible in localFS, since posix paths always work on windows in python (but not the other way around).

src/zarr/storage/local.py
```diff
@@ -219,17 +219,17 @@ async def exists(self, key: str) -> bool:
...
+        to_strip = self.root.as_posix() + "/"
         for p in list(self.root.rglob("*")):
             if p.is_file():
-                yield str(p).replace(to_strip, "")
+                yield p.as_posix().replace(to_strip, "")
 
     async def list_prefix(self, prefix: str) -> AsyncGenerator[str, None]:
         # docstring inherited
-        to_strip = os.path.join(str(self.root / prefix))
+        to_strip = (self.root / prefix).as_posix() + "/"  # TODO: fixme in 2430
```
jhamman, line 229: see https://github.com/zarr-developers/zarr-python/issues/2438

jhamman, line 229:
```suggestion
        to_strip = (self.root / prefix).as_posix() + "/"  # TODO: fixme in https://github.com/zarr-developers/zarr-python/issues/2438
```

.github/workflows/test.yml
```diff
@@ -16,14 +16,32 @@ concurrency:
...
+            dependency-set: 'optional'
+            os: 'macos-latest'
+          # - python-version: '3.11'
+          #   numpy-version: '1.25'
+          #   dependency-set: 'optional'
+          #   os: 'windows-latest'
+          # - python-version: '3.13'
+          #   numpy-version: '2.1'
+          #   dependency-set: 'optional'
+          #   os: 'windows-latest'
```
jhamman, line 43: I've opened a dedicated ticket for this: https://github.com/zarr-developers/zarr-python/issues/2438

.github/workflows/test.yml
```diff
@@ -16,14 +16,32 @@ concurrency:
...
+        include:
+          - python-version: '3.11'
+            numpy-version: '1.25'
+            dependency-set: 'optional'
+            os: 'macos-latest'
+          - python-version: '3.13'
+            numpy-version: '2.1'
+            dependency-set: 'optional'
+            os: 'macos-latest'
+          # - python-version: '3.11'
```
jhamman, line 36:
```suggestion
          # https://github.com/zarr-developers/zarr-python/issues/2438
          # - python-version: '3.11'
```

dcherian (approved):

jhamman:
pre-commit.ci autofix

PR #2432, opened 2024-10-22
Fix JSON encoding of complex fill values

dcherian:
We were not replacing NaNs and Infs with the string versions.

[Description of PR]

TODO:

### DIFF
--- a/src/zarr/core/metadata/v3.py
+++ b/src/zarr/core/metadata/v3.py
@@ -43,6 +43,13 @@ from zarr.registry import get_codec_class
 
 DEFAULT_DTYPE = "float64"
 
+# Keep in sync with _replace_special_floats
+SPECIAL_FLOATS_ENCODED = {
+    "Infinity": np.inf,
+    "-Infinity": -np.inf,
+    "NaN": np.nan,
+}
+
 
 def parse_zarr_format(data: object) -> Literal[3]:
     if data == 3:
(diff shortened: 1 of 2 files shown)


tests/test_array.py
```diff
@@ -436,3 +439,23 @@ def test_array_create_order(
...
+        (np.inf * 1j, ["NaN", "Infinity"]),
+        (-np.inf, ["-Infinity", 0.0]),
+        (math.inf, ["Infinity", 0.0]),
+    ],
+)
+async def test_special_complex_fill_values_roundtrip(fill_value: Any, expected: list[Any]) -> None:
+    store = MemoryStore({}, mode="w")
+    Array.create(store=store, shape=(1,), dtype=np.complex64, fill_value=fill_value)
+    content = await store.get("zarr.json", prototype=default_buffer_prototype())
+    assert content is not None
```
d-v-b, line 459: won't we get an error later in the test if `content` is `None`? or is this just for mypy (in which case I'd prefer explicitly ignore the type error)?

d-v-b: that `store.get` can return `None` is an unfortunate part of our store API that we need to fix IMO...

src/zarr/core/metadata/v3.py
```diff
@@ -475,12 +485,20 @@ def parse_fill_value(
...
-    elif np_dtype.kind in "cf":
+    elif np_dtype.kind == "f":
         # float comparison is not exact, especially when dtype <float64
-        # so we us np.isclose for this comparison.
+        # so we use np.isclose for this comparison.
         # this also allows us to compare nan fill_values
         if not np.isclose(fill_value, casted_value, equal_nan=True):
             raise ValueError(f"fill value {fill_value!r} is not valid for dtype {data_type}")
+    elif np_dtype.kind == "c":
+        # confusingly np.isclose(np.inf, np.inf + 0j) is False, so compare real and imag parts
```
dcherian, line 495:
```suggestion
        # confusingly np.isclose(np.inf, np.inf + 0j) is False on numpy<2, so compare real and imag parts
```

jhamman (approved):

Issue #2406, opened 2024-10-18
[v3] Implement `Group.tree`


jhamman:
### Zarr version

3.0.0.beta

### Numcodecs version

0.13

### Python Version

3.11

### Operating System

Mac

### Installation

pip

### Description

We are missing the `Group.tree` method from 2.x. We should add it for backward compatibility. 

### Steps to reproduce

```python
In [18]: root = zarr.group()

In [19]: root.tree()
--------------------------------------------------------------------------
NotImplementedError                      Traceback (most recent call last)
Cell In[19], line 1
----> 1 root.tree()

File ~/miniforge3/envs/icechunk-demo/lib/python3.12/site-packages/zarr/core/group.py:1474, in Group.tree(self, expand, level)
   1473 def tree(self, expand: bool = False, level: int | None = None) -> Any:
-> 1474     return self._sync(self._async_group.tree(expand=expand, level=level))

File ~/miniforge3/envs/icechunk-demo/lib/python3.12/site-packages/zarr/core/sync.py:185, in SyncMixin._sync(self, coroutine)
    182 def _sync(self, coroutine: Coroutine[Any, Any, T]) -> T:
    183     # TODO: refactor this to to take *args and **kwargs and pass those to the method
    184     # this should allow us to better type the sync wrapper
--> 185     return sync(
    186         coroutine,
    187         timeout=config.get("async.timeout"),
    188     )

File ~/miniforge3/envs/icechunk-demo/lib/python3.12/site-packages/zarr/core/sync.py:141, in sync(coro, loop, timeout)
    138 return_result = next(iter(finished)).result()
    140 if isinstance(return_result, BaseException):
--> 141     raise return_result
    142 else:
    143     return return_result

File ~/miniforge3/envs/icechunk-demo/lib/python3.12/site-packages/zarr/core/sync.py:100, in _runner(coro)
     95 """
     96 Await a coroutine and return the result of running it. If awaiting the coroutine raises an
     97 exception, the exception will be returned.
     98 """
     99 try:
--> 100     return await coro
    101 except Exception as ex:
    102     return ex

File ~/miniforge3/envs/icechunk-demo/lib/python3.12/site-packages/zarr/core/group.py:1244, in AsyncGroup.tree(self, expand, level)
   1243 async def tree(self, expand: bool = False, level: int | None = None) -> Any:
-> 1244     raise NotImplementedError

NotImplementedError:
```

TomAugspurger:
[...]

PR #2433, opened 2024-10-23
Added Group.tree method

TomAugspurger:
Adds a `Group.tree` method for pretty printing a hierarchy.

Screenshot 2024-10-24 at 8 05 59 AM

For now the old `expand` keyword, which I think controlled whether the tree was expanded by default in the HTML output, is not implemented.

We have one design choice to make: should `.tree` return an object with a custom `__repr__`, or should it return the string itself? My preference, implemented here, is returning the object. That makes it easier to pipe into other applications other than the Python REPL. The cost is a bit more expansion of the public API.

Closes https://github.com/zarr-developers/zarr-python/issues/2406


TomAugspurger:
There was a `notebooks/repr_tree.ipynb` notebook that I just ran through. This PR does *not* implement

- A plain ASCII tree repr with `print(bytes(g1.tree()).decode())`. We kind of get that for free through rich, which will automatically detect whether the console being rendered to supports unicode or just ASCII. We won't have an API for doing that (we can probably document how to configure rich to get out an ascii repr if someone really wants that)
- An interactive HTML repr (I think for collapsing nodes). I've toyed around with the idea of a textual-based TUI for Zarr groups that would be a better spot for this kind of interactive explorer IMO.

src/zarr/core/_tree.py
```diff
@@ -0,0 +1,64 @@
...
+    ) -> dict[str, str]:
+        # For jupyter support.
+        # Unsure why mypy infers the return type to by Any
+        return self._tree._repr_mimebundle_(include=include, exclude=exclude, **kwargs)  # type: ignore[no-any-return]
+
+
+async def group_tree_async(group: AsyncGroup, max_depth: int | None = None) -> TreeRepr:
+    tree = rich.tree.Tree(label=f"[bold]{group.name}[/bold]")
+    nodes = {"": tree}
+    members = sorted([x async for x in group.members(max_depth=max_depth)])
```
jhamman, line 46: cc @d-v-b who is looking at making this faster 🚀

jhamman (approved):
Looks great @TomAugspurger!

src/zarr/core/_tree.py
```diff
@@ -0,0 +1,64 @@
...
+except ImportError as e:
+    raise ImportError("'rich' is required for Group.tree") from e
+
+
+class TreeRepr:
+    """
+    A simple object with a tree-like repr for the Zarr Group.
+
+    Note that this object and it's implementation isn't considered part
+    of Zarr's public API.
```
jhamman, line 20: 👍

PR #2442, opened 2024-10-28
Refactor: store mode and initialization

### LINKED ISSUE #2359: [v3] Re-opening an existing RemoteStore V3 group causes the first child to be deleted

jhamman:
Previously, opening a store in `w` mode, had a side effect (`Store.clear`). That was leading to unexpected behavior when a `path` argument was also supplied. This refactor takes the following approach:

- ~keeps `Store.mode` but allows this to be only `StoreAccessMode = Literal["r", "w"]`~
- replaces `Store.mode` with `Store.read_only`
- removes any side effect when opening a store.
- adds `StorePath.open(..., mode: AccessModeLiteral)` where side effects can be applied at the path level

builds on #2430 
closes #2359

TODO:


src/zarr/abc/store.py
```diff
@@ -3,72 +3,34 @@
...
-    _mode: AccessMode
+    _mode: StoreAccessMode
     _is_open: bool
 
-    def __init__(self, *args: Any, mode: AccessModeLiteral = "r", **kwargs: Any) -> None:
+    def __init__(self, *args: Any, mode: StoreAccessMode = "r", **kwargs: Any) -> None:
         self._is_open = False
-        self._mode = AccessMode.from_literal(mode)
+        assert mode in ("r", "w")
+        self._mode = mode
```
jhamman, line 33: we may consider a different setting here (e.g. `readonly`)

src/zarr/abc/store.py
```diff
@@ -112,49 +74,48 @@ async def _open(self) -> None:
...
 
         Remove all keys and values from the store.
         """
-        ...
+        async for key in self.list():
+            await self.delete(key)
 
     @abstractmethod
-    def with_mode(self, mode: AccessModeLiteral) -> Self:
+    def with_mode(self, mode: StoreAccessMode) -> Self:
```
jhamman, line 118: To limit the scope here, I decided to keep this. But I'm not convinced we need it anymore.

src/zarr/api/asynchronous.py
```diff
@@ -576,9 +607,17 @@ async def group(
...
-    mode = None if isinstance(store, Store) else cast(AccessModeLiteral, "a")
-
-    store_path = await make_store_path(store, path=path, mode=mode, storage_options=storage_options)
+    mode: AccessModeLiteral
+    if overwrite:
+        mode = "w"
+    else:
+        mode = "r+"
+    store_mode = _handle_store_mode(mode)
+    print(f"store_mode: {store_mode}")
```
jhamman, line 616:
```suggestion
```

jhamman (commented):
A few notes to guide review

src/zarr/api/asynchronous.py
```diff
@@ -66,6 +65,31 @@
...
+    "w": "w",
+    "w-": "w",
+}
+
+
+def _handle_store_mode(mode: AccessModeLiteral | None) -> StoreAccessMode:
+    if mode is None:
+        return "r"
+    else:
+        return persistence_to_store_modes[mode]
```
d-v-b, line 85: What do we have to change higher up in the codebase if we narrow the function signature to `mode: AccessModeLiteral`? My gut feeling is that we should not need to handle `None` here, and the `None` -> `r` normalization can happen earlier.

src/zarr/api/asynchronous.py
```diff
@@ -691,22 +730,31 @@ async def open_group(
...
-        return await AsyncGroup.open(
-            store_path, zarr_format=zarr_format, use_consolidated=use_consolidated
-        )
+        if mode in {"r", "r+", "a"}:
+            return await AsyncGroup.open(
+                store_path, zarr_format=zarr_format, use_consolidated=use_consolidated
+            )
     except (KeyError, FileNotFoundError):
+        pass
+    if mode in {"a", "w", "w-"}:
```
d-v-b, line 750: should we bind this set to a variable, and similarly for `{"r", "r+", "a"}`? that would improve the literacy a bit.

src/zarr/storage/common.py
```diff
@@ -46,6 +45,36 @@ def __init__(self, store: Store, path: str = "") -> None:
...
+
+        Raises
+        ------
+        FileExistsError
+            If the mode is 'w-' and the store path already exists.
+        """
+        match mode:
+            case "w-":
+                if not await self.empty():
+                    raise FileExistsError(f"{self} must be empty. Mode is 'w-'")
```
d-v-b, line 69:
```suggestion
                    msg = (
                      "f"{self} is not empty, but `mode` is set to 'w-'."
                      "Either remove the existing objects in storage,"
                      "or set `mode` to a value that handles pre-existing objects"
                      "in storage, like `a` or `w`."
                    raise FileExistsError(msg)
```

src/zarr/storage/common.py
```diff
@@ -46,6 +45,36 @@ def __init__(self, store: Store, path: str = "") -> None:
...
+        ----------
+        mode : AccessModeLiteral
+            The mode to use when initializing the store path.
+
+        Raises
+        ------
+        FileExistsError
+            If the mode is 'w-' and the store path already exists.
+        """
+        match mode:
```
d-v-b, line 66: is this our first `match` statement in the codebase?

src/zarr/testing/strategies.py
```diff
@@ -134,7 +136,12 @@ def arrays(
     expected_attrs = {} if attributes is None else attributes
 
     array_path = path + ("/" if not path.endswith("/") else "") + name
-    root = Group.from_store(store, zarr_format=zarr_format)
+    root = zarr.open_group(store, mode="w")
+
+    # try:
+    #     del root[array_path]
+    # except KeyError:
+    #     pass
```
d-v-b, line 144:
```suggestion
```

src/zarr/abc/store.py
```diff
@@ -3,72 +3,34 @@
...
 
-    _mode: AccessMode
+    _mode: StoreAccessMode
     _is_open: bool
 
-    def __init__(self, *args: Any, mode: AccessModeLiteral = "r", **kwargs: Any) -> None:
+    def __init__(self, *args: Any, mode: StoreAccessMode = "r", **kwargs: Any) -> None:
         self._is_open = False
-        self._mode = AccessMode.from_literal(mode)
+        assert mode in ("r", "w")
```
d-v-b, line 32: what's this assertion for?

src/zarr/abc/store.py
```diff
@@ -112,49 +74,48 @@ async def _open(self) -> None:
...
         """Open the store if it is not already open."""
         if not self._is_open:
             await self._open()
 
-    @abstractmethod
-    async def empty(self) -> bool:
+    async def empty(self, prefix: str = "") -> bool:
         """
-        Check if the store is empty.
+        Check if the directory is empty.
```
d-v-b, line 89: are we going with object storage vocab or file system vocab? If the former, then maybe we should say something like "check if the store contains any keys with the given prefix"

src/zarr/abc/store.py
jhamman: I'll remove it. I think I threw it in during testing.

src/zarr/abc/store.py
jhamman: Good question. This highlights something we've been inconsistent about. I'll argue that we should use key-value store nomenclature for our current stores. Later on, it may be very useful to introduce something like a `HighLevelStore` that knows about directories, arrays, groups, etc.

src/zarr/api/asynchronous.py
jhamman: `READ_MODES` and `WRITE_MODES`?

src/zarr/storage/common.py
jhamman: perhaps! Do you like it?

src/zarr/api/asynchronous.py
d-v-b: works for me

src/zarr/storage/common.py
d-v-b: i do!

src/zarr/storage/common.py
```diff
@@ -66,7 +66,13 @@ async def _init(self, mode: AccessModeLiteral) -> None:
         match mode:
             case "w-":
                 if not await self.empty():
-                    raise FileExistsError(f"{self} must be empty. Mode is 'w-'")
+                    msg = (
+                      f"{self} is not empty, but `mode` is set to 'w-'."
+                      "Either remove the existing objects in storage,"
```
d-v-b, line 71: I think we need some whitespace: 

```suggestion
                      "Either remove the existing objects in storage, "
```

src/zarr/storage/common.py
```diff
@@ -66,7 +66,13 @@ async def _init(self, mode: AccessModeLiteral) -> None:
...
+                      "or set `mode` to a value that handles pre-existing objects"
```
d-v-b, line 72:
```suggestion
                      "or set `mode` to a value that handles pre-existing objects "
```

src/zarr/abc/store.py
TomAugspurger: I generally prefer the attribute matching the parameter name.

We can have a property

```
@property
def readonly(self): return self.mode == "r"
```

Edit: which I see you have down below.

src/zarr/abc/store.py
```diff
@@ -3,72 +3,34 @@
...
     """
     Abstract base class for Zarr stores.
     """
 
-    _mode: AccessMode
+    _mode: StoreAccessMode
     _is_open: bool
 
-    def __init__(self, *args: Any, mode: AccessModeLiteral = "r", **kwargs: Any) -> None:
+    def __init__(self, *args: Any, mode: StoreAccessMode = "r", **kwargs: Any) -> None:
```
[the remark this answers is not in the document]
jhamman: `Store.__init__` is currently called by all subclasses (e.g. `super().__init__(readonly=readonly)`). I did take away the `*args` and `**kwargs` though. Subclasses should feel free to use their own `__init__` and can optionally call the base class one.

TomAugspurger (approved):

src/zarr/abc/store.py
```diff
@@ -112,81 +72,55 @@ async def _open(self) -> None:
...
 
         Returns
         -------
         bool
             True if the store is empty, False otherwise.
         """
-        ...
 
-    @abstractmethod
+        if prefix and not prefix.endswith("/"):
```
d-v-b, line 100:
```suggestion
        if prefix != '' and not prefix.endswith("/"):
```
In case you want to be more explicit

src/zarr/api/asynchronous.py
```diff
@@ -67,6 +65,15 @@
 ]
 
 
+_READ_MODES: tuple[AccessModeLiteral, ...] = ("r", "r+", "a")
+_CREATE_MODES: tuple[AccessModeLiteral, ...] = ("a", "w", "w-")
+_OVERWRITE_MODES: tuple[AccessModeLiteral, ...] = ("a", "r+", "w")
+
+
+def _infer_exists_ok(mode: AccessModeLiteral) -> bool:
```
d-v-b, line 73:
```suggestion
def _infer_exists_ok(mode: AccessModeLiteral) -> bool:
    """
    Check that an ``AccessModeLiteral`` is compatible with overwriting an existing Zarr node.
    """
```

src/zarr/api/asynchronous.py
```diff
@@ -408,6 +416,7 @@ async def save_array(
         shape=shape,
         dtype=arr.dtype,
         chunks=chunks,
+        exists_ok=kwargs.pop("exists_ok", None) or _infer_exists_ok(mode),
```
d-v-b, line 419: maybe this is a taste thing but I would rather not have conditionals getting evaluated inside the function invocation like this. It saves some lines of code at the cost of readability / debuggability, not worth it IMO Could we instead do something like `_exists_ok = kwargs.pop("exists_ok", None) or _infer_exists_ok(mode)` followed by `...exists_ok=_exists_ok`?

src/zarr/abc/store.py
```diff
@@ -112,81 +72,55 @@ async def _open(self) -> None:
...
         self._is_open = True
 
     async def _ensure_open(self) -> None:
         """Open the store if it is not already open."""
         if not self._is_open:
             await self._open()
 
-    @abstractmethod
-    async def empty(self) -> bool:
+    async def empty_dir(self, prefix: str = "") -> bool:
```
d-v-b, line 85: Do we want to make the _default_ behavior of `empty_dir` to wipe the entire store at its root? Seems a bit dangerous to me. A safer alternative might be to make the `prefix` a required keyword argument.

jhamman: `empty_dir` only checks if a directory is empty (`delete_dir` does the deleting). But I'm happy to remove the default.

d-v-b: huh, the name `empty_dir` made me think it was a method that would empty a directory, not check if one was empty. Do we think this has enough confusion potential to warrant a slightly different name?

jhamman: Changed to `is_empty(prefix: str)`

jhamman:
Thanks folks for the reviews here!

Issue #2444, opened 2024-10-29
open() fails for path to a v2 group in a RemoteStore


agoodm:
### Zarr version

v3.0.0b

### Numcodecs version

n/a

### Python Version

3.11

### Operating System

Mac

### Installation

pip install -e .

### Description

Using the standard `open()` to open a v2 group inside a `RemoteStore` fails.

### Steps to reproduce

```python
In [1]: import zarr

In [2]: zarr.open("s3://mur-sst/zarr/", zarr_version=2, storage_options=dict(anon=True))
---------------------------------------------------------------------------
FileNotFoundError                         Traceback (most recent call last)
Cell In[2], line 1
----> 1 zarr.open("s3://mur-sst/zarr/", zarr_version=2, storage_options=dict(anon=True))

File ~/dev/zarr-python/src/zarr/_compat.py:43, in _deprecate_positional_args.<locals>._inner_deprecate_positional_args.<locals>.inner_f(*args, **kwargs)
     41 extra_args = len(args) - len(all_args)
     42 if extra_args <= 0:
---> 43     return f(*args, **kwargs)
     45 # extra_args > 0
     46 args_msg = [
     47     f"{name}={arg}"
     48     for name, arg in zip(kwonly_args[:extra_args], args[-extra_args:], strict=False)
     49 ]

File ~/dev/zarr-python/src/zarr/api/synchronous.py:77, in open(store, mode, zarr_version, zarr_format, path, **kwargs)
     67 @_deprecate_positional_args
     68 def open(
     69     store: StoreLike | None = None,
   (...)
     75     **kwargs: Any,  # TODO: type kwargs as valid args to async_api.open
     76 ) -> Array | Group:
---> 77     obj = sync(
     78         async_api.open(
     79             store=store,
     80             mode=mode,
     81             zarr_version=zarr_version,
     82             zarr_format=zarr_format,
     83             path=path,
     84             **kwargs,
     85         )
     86     )
     87     if isinstance(obj, AsyncArray):
     88         return Array(obj)

File ~/dev/zarr-python/src/zarr/core/sync.py:141, in sync(coro, loop, timeout)
    138 return_result = next(iter(finished)).result()
    140 if isinstance(return_result, BaseException):
--> 141     raise return_result
    142 else:
    143     return return_result

File ~/dev/zarr-python/src/zarr/core/sync.py:100, in _runner(coro)
     95 """
     96 Await a coroutine and return the result of running it. If awaiting the coroutine raises an
[...]

PR #2447, opened 2024-10-30
more extensive `api` x `store` testing

d-v-b:
The `api` module contains a lot of our user-facing code, but we are essentially only testing it against `LocalStore` and `MemoryStore`. A better situation would be to test the functions in `api` against all the stores. That's what this PR adds. Note that there are many test failures. I will try to get them all fixed.

This PR also contains a fix for #2444, but the central addition is the (currently failing) test for that scenario.

TODO:


d-v-b:
this became a bit of a saga. 

### S3 test fixtures

I moved the s3 test fixture to the main `conftest.py` so that any test can depend on a `RemoteStore` instance. I was getting strange errors from our s3 test fixture. they went away after I made the following changes:
- set the region to `us-east-1`
- set the moto port automatically instead of hard-coding it. this makes the moto server URL dynamic, so I wired up the test fixtures that need that url to consume the `s3_base` fixture. 

### `store` test fixture

I also refactored the `store` test fixture. It's now possible to request a store class and a store mode with the appropriately formatted string, e.g. `@pytest.mark.parametrize('store', ['local_a', 'local_w'], indirect=True)` will request a `LocalStore` with `mode="a"`, and a `LocalStore` with `mode="w"`. This test fixture was completely broken for `RemoteStore`, so I fixed that. I also removed a redundant `store` test fixture from the group tests, and I removed the individual `memory_store`, `local_store`, etc test fixtures. 

### we should default to testing all the stores

IMO our default behavior should be to test relevant APIs against as many stores as possible, at least until slow tests becomes an issue. That being said, I'm happy to narrow the scope of certain tests, e.g. consolidated metadata tests, to only run on `MemoryStore` if we think that's OK.

### `ZipStore` problems

We have a lot of tests that involve creating an array or group with a store, then opening a new array or group with the same store but with a different mode. This is impossible today because `ZipStore` doesn't do `with_mode`, so tests where I couldn't hack around this limitation are xfailed (there are several in `test_api`).

### `open` logic

I had test failures relating to some of the logic in `zarr.api.asynchronous.open`. The version of this function in  `main` is very hard to understand, the version in this PR is marginally simpler. I think it's possible in `main` to run `open(...shape=(10,))` and get back a group, which should not be possible (and is not possible after my changes). 

### Failing tests

Here's the currently failing tests. I think these are all real problems exposed by adding more tests, but I don't really have any idea how to solve them. Help would be appreciated here!

```
FAILED tests/test_group.py::test_group_update_attributes_async[zarr2-remote] - botocore.exceptions.HTTPClientError: An HTTP Client raised an unhandled exception: Task <Task pending name='Task-147164' coro=<set_or_delete() running at /Users/runner/work/zarr-python/zarr-python/src/zarr/abc/store.py:425> cb=[gather.<locals>._done_callback() at /Library/Frameworks/Python.framework/Versions/3.13/lib/python3.13/asyncio/tasks.py:820]> got Future <Future pending cb=[BaseSelectorEventLoop._sock_write_done()()]> attached to a different loop
FAILED tests/test_group.py::test_group_update_attributes_async[zarr3-remote] - botocore.exceptions.HTTPClientError: An HTTP Client raised an unhandled exception: Task <Task pending name='Task-147207' coro=<set_or_delete() running at /Users/runner/work/zarr-python/zarr-python/src/zarr/abc/store.py:425> cb=[gather.<locals>._done_callback() at /Library/Frameworks/Python.framework/Versions/3.13/lib/python3.13/asyncio/tasks.py:820]> got Future <Future pending cb=[BaseSelectorEventLoop._sock_write_done()()]> attached to a different loop
FAILED tests/test_group.py::test_members_name[zarr2-remote-True] - DeprecationWarning: datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.now(datetime.UTC).
FAILED tests/test_group.py::test_members_name[zarr2-remote-False] - DeprecationWarning: datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.now(datetime.UTC).
FAILED tests/test_group.py::test_members_name[zarr3-remote-True] - DeprecationWarning: datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.now(datetime.UTC).
FAILED tests/test_group.py::test_members_name[zarr3-remote-False] - DeprecationWarning: datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.now(datetime.UTC).
```

d-v-b:
The remaining test falures are in the python 3.13 tests where `botocore.auth` uses `datetime` in a deprecated way, resulting in  `DeprecationWarning`, which `test_members_name` treats as an error. See https://github.com/zarr-developers/zarr-python/actions/runs/11620549084/job/32362570100?pr=2447. Since this is purely a `botocore` x `py3.13` issue, this test qua Zarr test should probably not fail. @jhamman any suggestions for changes to the `test_members_name` that would avoid this error?

tests/test_api.py
```diff
@@ -175,7 +224,28 @@ def test_save_errors() -> None:
...
+
+    z1 = zarr.ones(store=store, shape=(3, 3))
+    assert z1.fill_value == 1
+
+    z2 = zarr.open(store=store, mode="r")
+    assert isinstance(z2, Array)
+    assert z2.fill_value == 1
+    assert (z2[:] == 1).all()
+    with pytest.raises(ValueError):
+        z2[:] = 3
```
d-v-b, line 245: @jhamman this test is failing in this PR. What's the expected behavior here, given the new semantics of `read_only` for stores? Should we raise an error in `zarr.open` if the mode is `r` but the store is not read only? 

Note the test below, which is copied from `main`, passes, but in that case `zarr.open` takes a path / string and thus has the opportunity to create the store in read only mode.

d-v-b:
getting the lasts tests to pass required adding two new methods to the stores: `_as_mutable` and `_as_immutable`, that return a non-read-only or a read-only copy of a store class, respectively. This is necessary because without some means of creating read-only copies of stores, there's no way for `zarr.open(mutable_store, mode='r')` to correctly produce a read-only store class.

I think this PR is pretty important insofar as it opens up the possibility of testing the store classes much more uniformly, it would be great to get this looked at soon!

tests/test_api.py
d-v-b: see https://github.com/zarr-developers/zarr-python/pull/2447#issuecomment-2500205184

d-v-b:
A major downside to this PR is the additional time required for tests -- with these changes, tests in CI take nearly 20m to complete! On the other hand, I have found many bugs by testing extensively. Not sure how to find balance here.

dcherian:
> Not sure how to find balance here.

Hypothesis property tests!

src/zarr/abc/store.py
```diff
@@ -412,6 +412,18 @@ async def getsize_prefix(self, prefix: str) -> int:
         sizes = await concurrent_map(keys, self.getsize, limit=limit)
         return sum(sizes)
 
+    def _as_immutable(self: Self) -> Self:
```
dcherian, line 415: in icechunk we have these as `set_read_only` and `set_writeable` which I mildly prefer. Simple words are nice.

tests/test_api.py
```diff
@@ -22,14 +22,15 @@
...
 from zarr.storage._utils import normalize_path
 from zarr.storage.memory import MemoryStore
+from zarr.storage.zip import ZipStore
 
 
-def test_create_array(memory_store: Store) -> None:
-    store = memory_store
-
+@pytest.mark.parametrize("store", ["local", "memory", "remote"], indirect=True)
+def test_create_array(store: Store) -> None:
```
dcherian, line 33:
```suggestion
from hypothesis import given
import hypothesis.strategies as st

stores = st.sampled_from(["local", "memory", "remote"])

@given(store=st)
def test_create_array(store: Store) -> None:
```

src/zarr/abc/store.py
```diff
@@ -412,6 +412,18 @@ async def getsize_prefix(self, prefix: str) -> int:
...
+        """
+        Return a mutable copy of the store.
```
chuckwondo, line 417: It looks like perhaps the words "mutable" and "immutable" should be swapped between this method and the `_as_mutable` method, no?

d-v-b: oops, good catch!

d-v-b:
This is so old and stale, that I'm confident any bugs this would real have already been fixed (hopefully)

PR #2463, opened 2024-11-04
top-level functions for reading, creating data

d-v-b:
adds async and sync `read_array`, `read_group`, and `read` functions. `read_array` will read an array from storage. `read_group` does the same, but for groups, and `read` will try to read arrays or groups at the storage path. These functions wrap their respective creation functions, passing `mode=r` in each case.

I still need to add tests.

TODO:


d-v-b:
in writing this PR I noticed that we don't have a `create_array` function, but we do have `create_group`. I don't like the asymmetry. We should fix that (by adding a `create_array` func).

d-v-b:
> in writing this PR I noticed that we don't have a `create_array` function, but we do have `create_group`. I don't like the asymmetry. We should fix that (by adding a `create_array` func).

I'm wrong, we don't have `create_group`! But we should! I think we should actually deprecate file-mode semantics across the board in favor of `create_x`, `read_x` functions.

d-v-b:
This is ready for review. To summarize:

I created 5 new top-level exports: 
- `read_array`
- `read_group`
- `read`
- `create_array`
- `create_group`

The goal for these functions is to directly support common access patterns without forcing users to worry about the semantics of the `mode` keyword, and to guard against accessing zarr data with an unintended access mode (e.g., getting mutable access to an array you only want to read).

src/zarr/core/array.py
```diff
@@ -144,9 +145,9 @@ async def get_array_metadata(
             (store_path / ZATTRS_JSON).get(),
         )
         if zarr_json_bytes is not None and zarray_bytes is not None:
-            # TODO: revisit this exception type
-            # alternatively, we could warn and favor v3
-            raise ValueError("Both zarr.json and .zarray objects exist")
+            # wwarn and favor v3
```
d-v-b, line 148:
```suggestion
            # warn and favor v3
```

src/zarr/core/group.py
```diff
@@ -492,9 +492,9 @@ async def open(
                 (store_path / str(consolidated_key)).get(),
             )
             if zarr_json_bytes is not None and zgroup_bytes is not None:
-                # TODO: revisit this exception type
-                # alternatively, we could warn and favor v3
-                raise ValueError("Both zarr.json and .zgroup objects exist")
+                # we could warn and favor v3
```
d-v-b, line 495:
```suggestion
                # warn and favor v3
```

tests/test_api.py
```diff
@@ -48,6 +53,69 @@ def test_create_array(memory_store: Store) -> None:
     assert z.chunks == (40,)
 
 
+@pytest.mark.parametrize("store", ["memory"], indirect=True)
+def test_read(store: Store) -> None:
+    """
+    Test that the polymorphic read function works.
```
d-v-b, line 59:
```suggestion
    Test that the polymorphic read function can return an Array or a Group, depending on the path argument.
```

d-v-b:
Some feedback from the community meeting last week:
- One point of contention in this PR was the polymorphic `read` function, which returns either an array or group, depending on what's in the hierarchy. Personally I think this is invaluable for navigating unknown Zarr hierarchies, but others thought it was unnecessary or confusing. Does anyone else have thoughts? 
- Should this PR contain deprecation warnings for functions that are being made less essential by this PR (e.g., `open` and `create`)?

d-v-b:
from the community meeting:
- `create_array` should alias `create`
- we should use these new API functions to compactly express sharding, e.g. `create_array(shards=..., chunks=...)`

src/zarr/api/asynchronous.py
```diff
@@ -764,6 +854,173 @@ async def open_group(
...
+        store=store,
+        mode="r",
+        path=path,
+        storage_options=storage_options,
+        zarr_format=zarr_format,
+        use_consolidated=use_consolidated,
+    )
+
+
+async def create_array(
```
normanrz, line 918: What do you think, should we use the same API signature in `AsyncGroup.create_array` (except for the first arg)?

d-v-b: i think the two signatures should be the same, but that might entail coming up with a brand new signature here, and changing `AsyncGroup.create_array` accordingly

d-v-b:
this is ready for people to look at. there are not enough tests yet for my tastes, and there's some mypy issues that need to be squashed, but feedback would be appreciated

src/zarr/core/chunk_grids.py
```diff
@@ -144,6 +146,44 @@ def normalize_chunks(chunks: Any, shape: tuple[int, ...], typesize: int) -> tupl
     return tuple(int(c) for c in chunks)
 
 
+def _auto_partition(
```
normanrz, line 149: Does this need updating, now that `auto` is not part of the signature anymore?

d-v-b:
feedback from meeting:
- drop `read_array`, `read_group`, and `read` (keep it simple)

d-v-b:
this has evolved a bit. In particular, ensuring that `Group.create_array` matches the free-floating `create_array` function drove a lot of churn. 

it's too late for me to continue right now, so I will summarize the current state of `create_array`:
- I renamed `compressors` to `compression`. the name `compressors` was confusing in the context of zarr v2, where there can only be a single compressor. `compression` is ambiguous about the number of elements, which is useful.
- I had to introduce automatic sharding and compression, because the previous `create_array` function on groups had auto chunking, and specifying chunk shape for each invocation of `create_array` in the test suite would be incredibly painful. If `shard_shape='auto'`, then we use the following heuristic for generating a shard size is: for each axis, if there are more than 16 chunks along that axis, then the shard shape will be 2 x the chunk shape along that axis, otherwise the shard shape == chunk_shape for that axis. The default value of `shard_shape` is `None`, which doesn't use sharding at all.
- The specification of the default codecs in the config needs to be improved IMO. We are only specifying the names of the codecs, but what I would find much more helpful would be for the config to contain more than just a name, e.g. full parameterization of codecs. So I had to implement some functions for actually creating the default codecs / compressors, should the go for the default auto-codec route.

one substantial change that required touching some tests: I made `dtype` a required parameter for `create_array`. This is a personal thing -- I think a default dtype like `float64` is a recipe for accidentally writing the wrong data -- but if people disagree I can go back to the default of `float64`. I do think we need to deviate from numpy a bit, because writing data can be expensive in ways that creating in-memory arrays is not.

tests are not all passing (but the array and group tests pass) and mypy is very mad. I don't think it's great that we have to deal with numcodecs codecs and `zarr-python` codecs. It would be a lot simpler from a typing POV to have have 1 codec type. Not clear how to make this happen as long as we want to allow people to provide numcodecs codecs when they are creating v2 arrays.

what we need:
- review of the design
- passing tests
- passing mypy
- tests for `create_array` that check auto shard shape, auto chunk shape combinations

jhamman:
> I renamed compressors to compression. the name compressors was confusing in the context of zarr v2, where there can only be a single compressor. compression is ambiguous about the number of elements, which is useful.

In zarr-python 2, we have `compressor`. Should we consider handling the case where a user passes `compressor`? Not sure I like the idea but we could make the `compressor` argument handle multiple compressors.

No strong opinions here.  I guess we're still feeling low conviction on the API we want to surface.

> one substantial change that required touching some tests: I made dtype a required parameter for create_array. This is a personal thing -- I think a default dtype like float64 is a recipe for accidentally writing the wrong data -- but if people disagree I can go back to the default of float64. I do think we need to deviate from numpy a bit, because writing data can be expensive in ways that creating in-memory arrays is not.

I'm worried this will break a lot of stuff for folks. What if we change the default to `None` and then issue a warning. 

```python
if dtype is None:
    warnings.warn("dtype was not set when creating array, defaulting to float64")
    dtype = "f8"
...
```

> mypy is very mad

Revert your change with `dtype[generic]` and you should resolve some of the issues. Numpy 2.2 broke some things so we have since pinned the numpy version for our mypy config.

normanrz:
Thanks for your work. I'll work on changing the default codecs to full config objects tonight.

d-v-b:
> > I renamed compressors to compression. the name compressors was confusing in the context of zarr v2, where there can only be a single compressor. compression is ambiguous about the number of elements, which is useful.
> 
> In zarr-python 2, we have `compressor`. Should we consider handling the case where a user passes `compressor`? Not sure I like the idea but we could make the `compressor` argument handle multiple compressors.
> 
> No strong opinions here. I guess we're still feeling low conviction on the API we want to surface.

One of my goals with `create_array` is to make the specification of chunk encoding (i.e., how the chunks are written to disk) as consistent as possible between zarr v2 and v3. This is hard, because v2 and v3 don't use the same model, but in the interest of a unified API, I think we should strive to avoid zarr-2-only parameters, or zarr-3-only parameters, wherever we can, so I would lean against a zarr-v2-only `compressor` parameter. I think seeing `compressor` and `compression` in the same function signature will be very confusing to users, and we would need to handle a lot of annoying edge cases (the user provides `compression`, `compressor`, and `zarr_format=3`, ...).

my goals are:
- use the _exact_ same keyword arguments to specify chunk encoding for zarr v2 and v3
- give users intuitive control over sharding for zarr v3 arrays

satisfying the second goal requires that we make the choice of `ArrayBytesCodec` implicit, because users shouldn't have to say "I want to use the sharding codec; I want my shard shape to be bla bla bla"  in two separate statements. So we are down to 2 free parameters for v3 (array array codecs, bytes bytes codecs), and 2 parameters for v2 (filters, compressor). I think `(filters, compression)` is a solid compromise between the two, but we could also choose two completely new words, so that v2 users can't rely on any familiarity? 

FWIW, I'm much more settled on the "we should the same small number of free parameters for both v2 and v3 chunk encoding" question than on what those parameters should be named. I think the names are important, but secondary to the semantics of array creation.


> I'm worried this will break a lot of stuff for folks. What if we change the default to `None` and then issue a warning.
> 
> ```python
> if dtype is None:
>     warnings.warn("dtype was not set when creating array, defaulting to float64")
>     dtype = "f8"
> ...
> ```

If we aren't deprecating the default dtype, then I don't think users will appreciate the default behavior of a function emitting a warning. I think it's simpler to just copy the old (bad, IMO) behavior of v2 without a warning.

src/zarr/core/array.py
```diff
@@ -3449,3 +3456,305 @@ def _get_default_codecs(
...
+            order=order_parsed,
+            filters=filters_parsed,
+            compressor=compressor_parsed,
+            attributes=attributes,
+            overwrite=overwrite,
+            config=config_parsed,
+        )
+    else:
+        array_array, array_bytes, bytes_bytes = _get_default_encoding_v3(dtype_parsed)
+        sub_codecs = (*array_array, array_bytes, *bytes_bytes)
```
jhamman, line 3598: this seems to ignore any compression/filter that is passed

d-v-b: oops, that's definitely not intended!

src/zarr/core/array.py
```diff
@@ -3449,3 +3456,305 @@ def _get_default_codecs(
...
+    dtype : npt.DTypeLike
+        Data type of the array.
+    chunk_shape : ChunkCoords
+        Chunk shape of the array.
+    shard_shape : ChunkCoords, optional
+        Shard shape of the array. The default value of ``None`` results in no sharding at all.
+    filters : Iterable[Codec], optional
+        List of filters to apply to the array.
+    compression : Iterable[Codec], optional
+        List of compressors to apply to the array.
```
jhamman, line 3510: These definitions could use some tuning. Should we advertise `filters` as `array->array` codecs and `compressors` as `bytes->bytes` codecs? Some further constraints here seem important to helping us decide on behavior and users to understand how to configure them.

jhamman (commented):
A few more notes...

src/zarr/core/chunk_grids.py
```diff
@@ -194,3 +194,48 @@ def get_nchunks(self, array_shape: ChunkCoords) -> int:
...
+            # aim for a 1MiB chunk
+            _chunks_out = _guess_chunks(array_shape, item_size, max_bytes=1024)
+        else:
+            _chunks_out = chunk_shape
+
+        if shard_shape == "auto":
+            _shards_out = ()
+            for a_shape, c_shape in zip(array_shape, _chunks_out, strict=True):
+                # TODO: make a better heuristic than this.
+                # for each axis, if there are more than 8 chunks along that axis, then put
```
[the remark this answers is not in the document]
normanrz: Probably for another PR?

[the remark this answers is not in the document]
d-v-b: +1 to an explicit size being better than "auto", and also to it being in-scope for another PR

src/zarr/api/synchronous.py
```diff
@@ -523,6 +533,47 @@ def open_group(
     )
 
 
+def create_group(
```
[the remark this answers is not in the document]
d-v-b: the downside of this is that the return types are different (async array vs array). I think I can just copy + paste with modifications at the very end of this PR

src/zarr/core/array.py
d-v-b: agreed, I will add some clarifying language

src/zarr/api/asynchronous.py
```diff
@@ -1015,6 +1058,41 @@ async def create(
     )
 
 
+async def read_array(
```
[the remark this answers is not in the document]
d-v-b: removed it

src/zarr/core/array.py
d-v-b: this is now fixed

src/zarr/core/group.py
```diff
@@ -998,115 +1000,83 @@ async def create_array(
         name: str,
         *,
         shape: ShapeLike,
-        dtype: npt.DTypeLike = "float64",
-        fill_value: Any | None = None,
+        dtype: npt.DTypeLike,
+        chunks: ChunkCoords | Literal["auto"] = "auto",
+        shards: ChunkCoords | Literal["auto"] | None = None,
+        filters: Iterable[dict[str, JSON] | Codec] = (),
+        compressors: Iterable[dict[str, JSON] | Codec] = (),
```
normanrz, line 1007: Are these supposed to be `"auto"`? Otherwise the default codecs will not be used. I found out because `tests/test_metadata/test_consolidated.py::TestConsolidated::test_consolidated` fails

normanrz: I changed that in https://github.com/zarr-developers/zarr-python/pull/2463/commits/ea3ed0e499d9ed6826c2f61d97e86c60084d285d. Feel free to revert.

src/zarr/core/array.py
```diff
@@ -3448,4 +3492,333 @@ def _get_default_codecs(
...
+    fill_value: Any | None = 0,
+    order: MemoryOrder | None = "C",
+    zarr_format: ZarrFormat | None = 3,
+    attributes: dict[str, JSON] | None = None,
+    chunk_key_encoding: ChunkKeyEncoding | ChunkKeyEncodingParams | None = None,
+    dimension_names: Iterable[str] | None = None,
+    storage_options: dict[str, Any] | None = None,
+    overwrite: bool = False,
+    config: ArrayConfig | ArrayConfigParams | None = None,
+    data: npt.ArrayLike | None = None,
```
jhamman, line 3525: how important is it that we include the `data` kwarg. Previously, if `data` was provided, shape/dtype were inferred from the input. That's now how things work here (which is fine) but it draws into question the value of this parameter.

I suggest we remove this in favor of a `zarr.from_array` function (to come later).

src/zarr/core/group.py
```diff
@@ -2594,120 +2535,84 @@ def array(
...
         *,
-        shape: ChunkCoords,
-        dtype: npt.DTypeLike = "float64",
-        fill_value: Any | None = None,
+        shape: ShapeLike,
+        dtype: npt.DTypeLike,
+        chunks: ChunkCoords | Literal["auto"] = "auto",
+        shards: ChunkCoords | Literal["auto"] | None = None,
+        filters: Iterable[dict[str, JSON] | Codec] = (),
+        compressors: Iterable[dict[str, JSON] | Codec] = (),
```
normanrz, line 2543: Could we use auto here?
```suggestion
        filters: Iterable[dict[str, JSON] | Codec] = "auto",
        compressors: Iterable[dict[str, JSON] | Codec] = "auto",
```

src/zarr/core/array.py
d-v-b: +1, i think data doesn't fit here, and `from_array` is a better spot for it

d-v-b: `data` is removed as a kwarg in https://github.com/zarr-developers/zarr-python/pull/2463/commits/77f40a5de0422ff2f14ae55d54e8b94637f870a5

normanrz (commented):
I had another review pass. It is looking really good. 

Can you have a look at https://github.com/zarr-developers/zarr-python/pull/2463#discussion_r1896361722 ?

We should do another pass to make sure that the docstrings are consistent for all variants of the the create_array methods.

normanrz:
I was wondering, should we replace `(Async)Array.create` with `create_array`? It seems a bit confusing to have both.

normanrz:
> I was wondering, should we replace `(Async)Array.create` with `create_array`? It seems a bit confusing to have both. 

Maybe that is a bit too big for a last minute change. But we could remove `create` from the public API (prefix with `_`) and change the uses in the tests to `create_array`. That would dramatically increase the test coverage.

src/zarr/api/asynchronous.py
```diff
@@ -666,6 +661,54 @@ async def group(
...
+    -------
+    g : group
+        The new group.
+    """
+
+    if zarr_format is None:
+        zarr_format = _default_zarr_version()
+
+    # TODO: fix this when modes make sense. It should be `w` for overwriting, `w-` otherwise
+    mode: Literal["a"] = "a"
```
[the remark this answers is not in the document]
d-v-b: I removed the TODO as its from an era before some store mode refactoring.

d-v-b:
> I was wondering, should we replace `(Async)Array.create` with `create_array`? It seems a bit confusing to have both.

I totally agree that it's confusing to have both, and I would prefer to just have `create_array`, but maybe that's for a separate PR?

d-v-b:
I think this is ready for review.

src/zarr/api/asynchronous.py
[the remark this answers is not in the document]
normanrz: @d-v-b Did you push your change?

src/zarr/core/array.py
```diff
@@ -3448,4 +3500,367 @@ def _get_default_codecs(
     else:
         dtype_key = "numeric"
 
-    return [{"name": codec_id, "configuration": {}} for codec_id in default_codecs[dtype_key]]
+    return cast(list[dict[str, JSON]], default_codecs[dtype_key])
+
+
+FiltersParam: TypeAlias = (
```
normanrz, line 3506: What about allowing just a single filter (without wrapping in a list/tuple)?

src/zarr/core/array.py
```diff
@@ -3448,4 +3500,367 @@ def _get_default_codecs(
...
+async def create_array(
+    store: str | StoreLike,
+    *,
+    name: str | None = None,
+    shape: ShapeLike,
+    dtype: npt.DTypeLike,
+    chunks: ChunkCoords | Literal["auto"] = "auto",
+    shards: ChunkCoords | Literal["auto"] | None = None,
+    filters: FiltersParam = "auto",
+    compressors: CompressorsParam = "auto",
```
normanrz, line 3529: How can a user specify that they don't want any filters or compressors? Overriding the defaults. Would that be `None` or `()`? I think we should allow both.

normanrz: See https://github.com/zarr-developers/zarr-python/pull/2463/commits/c4330ef3778270e7384d0e14e5402af21426f203#diff-d1499eae500b1036f9aa3e3c6266bc516b9f03efcd65644acac5418e406822baR968-R1003

d-v-b: I _think_ an empty iterable should work with the current code, and I'm fine with taking `None` as an alias for an empty iterable here.

src/zarr/core/array.py
d-v-b: we should allow this!

src/zarr/core/array.py
d-v-b: tests are failing for v2 with the `compressors` kwarg -- I don't think we should allow an empty tuple for v2

normanrz: Why not? Shouldn't it be as uniform as possible?

d-v-b: we should be as uniform as we can be, but the underlying metadata documents are simply not uniform: in v2, `compressor` maps on to single entity, in v3 it maps on to a collection. If an empty tuple is an valid `compressor` for zarr v2, people might also expect that a tuple with codecs in it is also valid, but it's not. I introduced the term "compression" instead of "compressors" to try and mitigate the "singleton in v2" / "collection in v3" confusion...

d-v-b: but maybe a uniform API is worth the possibility of confusing users with an empty tuple? I guess I'm not too certain here. The fundamental problem is that the semantics of v2 metadata and v3 metadata are similar, but different enough to make this design a compromise one way or another :/

d-v-b: what about accepting either an empty tuple, or a tuple with 1 element, and erroring otherwise for `compressors` when `zarr_format=2` ? And we have a nice fat docstring to explain this

normanrz: We should also accept a numcodecs codec/dict without wrapping in a tuple.

normanrz: Ok! I added something to the docstring for clarification.

normanrz (approved):

d-v-b:
one big question we should answer before merging: this PR adds `compressors` and `filters` as keyword arguments to `create_array`, but `Array.filters` and `Array.compressors` are not properties of the `Array` class. I don't think this is very ergonomic -- I think users reasonably expect an API like `x = create_array(filters=foo)... x.filters`. 

Should plan to add those properties (e.g., in a separate PR)? 

And if we do, what would the type of `Array.compressors` be for zarr v2 arrays?  A tuple containing 0 or 1 numcodecs codecs, or a numcodecs codec? If we do the latter and `Array.compressors` is _not_ an iterable collection, then I think this will be very confusing. So I would suggest that we plan to add `Array.compressors` / `Array.filters` properties to the `Array` class, and they are both tuples, but the length of `compressors` is clamped to 0 or 1 for zarr v2 arrays. Thoughts?

src/zarr/api/asynchronous.py
```diff
@@ -9,7 +9,7 @@
 import numpy.typing as npt
 from typing_extensions import deprecated
 
-from zarr.core.array import Array, AsyncArray, get_array_metadata
+from zarr.core.array import Array, AsyncArray, create_array, get_array_metadata
```
jhamman, line 12: Question: why did we put `create_array` in `zarr.core.array` and not in this module?

src/zarr/api/synchronous.py
```diff
@@ -523,6 +525,29 @@ def open_group(
...
+    store: StoreLike,
+    *,
+    path: str | None = None,
+    zarr_format: ZarrFormat | None = None,
+    overwrite: bool = False,
+    attributes: dict[str, Any] | None = None,
+    storage_options: dict[str, Any] | None = None,
+) -> Group:
```
jhamman, line 536: Please add a docstring for this function.

src/zarr/api/synchronous.py
```diff
@@ -675,6 +700,10 @@ def create(
     )
 
 
+def create_array(*args: Any, **kwargs: Any) -> Array:
+    return Array(sync(zarr.core.array.create_array(*args, **kwargs)))
```
jhamman, line 704: Please add a docstring for this function.

src/zarr/api/asynchronous.py
```diff
@@ -666,6 +662,54 @@ async def group(
...
+    Returns
+    -------
+    g : group
+        The new group.
+    """
+
+    if zarr_format is None:
+        zarr_format = _default_zarr_version()
+
+    # TODO: fix this when modes make sense. It should be `w` for overwriting, `w-` otherwise
```
jhamman, line 700:
```suggestion
```

src/zarr/core/group.py
```diff
@@ -2225,120 +2199,82 @@ def create_array(
         name: str,
         *,
         shape: ShapeLike,
-        dtype: npt.DTypeLike = "float64",
-        fill_value: Any | None = None,
+        dtype: npt.DTypeLike,
+        chunks: ChunkCoords | Literal["auto"] = "auto",
+        shards: ChunkCoords | None = None,
+        filters: FiltersParam = "auto",
+        compressors: CompressorsParam = "auto",
```
jhamman, line 2206: For backward compatibility, we should also support the `compressor` argument. The logic should be simple though:

```python
if compressor:
    warnings.warn("compressor is deprecated ...")
    compressors = (compressor, )
```

jhamman (commented):
This looks great. Just one comment of substance. Let's get this in and start using it.

src/zarr/api/asynchronous.py
d-v-b: if `create_array` is defined in `api.asynchronous`, then the `Group.create_array` method has to import `create_array` from `api.asynchronous`, but that module imports the `Group` class, which leads to a circular import. So I moved `create_array` lower in the hierarchy, which makes it easy for `group.py` to import.

src/zarr/api/synchronous.py
normanrz: I added one in #2607

src/zarr/core/group.py
normanrz: Should we do that just in `(Async)Group.create_array` or also in the new top-level `zarr.api.(a)synchronous.create_array` methods?

d-v-b: my feeling is that the brand new functions shouldn't have to worry about deprecated APIs

normanrz: Added in https://github.com/zarr-developers/zarr-python/pull/2463/commits/be60d73ce00f0ed9a23cf8bd4e3e80fd96f58056

src/zarr/api/asynchronous.py
jhamman: got it.

jhamman (approved):

jhamman:
Thanks @d-v-b and @normanrz - what a lift 🏋️‍♂️

Issue #2438, opened 2024-10-24
[v3] Windows support


jhamman:
### Zarr version

3.0.0b1

### Numcodecs version

0.13.1

### Python Version

3.13

### Operating System

Windows

### Installation

hatch

### Description

CI is currently failing on Windows due to some path issues in the `LocalStore`. I don't have access to a Windows development environment so  I'm splitting the Windows issues out of https://github.com/zarr-developers/zarr-python/pull/2431

This is likely a light lift for someone with access to a Windows development environment:

1. See the comments in https://github.com/zarr-developers/zarr-python/pull/2431
2. Review the failures in the CI logs below
3. Uncomment the Windows section in our test CI job
https://github.com/zarr-developers/zarr-python/blob/4a3bbf1cbb89e90ea9ca4d6d75dae23ed4b957c9/.github/workflows/test.yml#L36-L44

### Steps to reproduce

See https://github.com/zarr-developers/zarr-python/actions/runs/11472585956/job/31925406759?pr=2431

PR #2467, opened 2024-11-05
use as_posix correctly for str and for listdir

d-v-b:
Fixes a bug in `LocalStore.list_dir` caused by not using a posixpath representation when removing the prefix in `list_dir`. This was only apparent on windows, I also turned on Windows tests.


dstansby (approved):
Nice!

dstansby:
Fixes https://github.com/zarr-developers/zarr-python/issues/2438 I think?

jhamman:
Whoop!!! 🙌

jhamman (approved):

dcherian (approved):

PR #2470, opened 2024-11-06
Add default compressors to config

### LINKED ISSUE #2267: [v3] default compressor / codec pipeline
### Zarr version

3.0.0.alpha6

### Numcodecs version

N/A

### Python Version

N/A

### Operating System

N/A

### Installation

N/A

### Description

In Zarr-Python 2.x, Zarr provided default compressors for most (all?) datatypes. As of now, in 3.0, we don't provide any defaults.

### Steps to reproduce

In 2.18:

```python
In [1]: import zarr

In [2]: arr = zarr.create(shape=(10, ), path='foo', store={})

In [3]: arr.compressor
Out[3]: Blosc(cname='lz4', clevel=5, shuffle=SHUFFLE, blocksize=0)
```

In 3.0.0.alpha6
```python
In [4]: arr = zarr.create(shape=(10, ), path='foo', store={}, zarr_format=2)

In [5]: arr.metadata
Out[5]: ArrayV2Metadata(shape=(10,), fill_value=0.0, chunk_grid=RegularChunkGrid(chunk_shape=(10,)), attributes={}, zarr_format=2, data_type=dtype('float64'), order='C', filters=None, dimension_separator='.', compressor=None)
```

### Additional output

_No response_

brokkoli71:
This PR adds:
 - default compressor  for  `zarr_format=2` 
 - specification of default compressors dependent on dtype in `zarr.config`

fixes #2267

Should `_get_default_array_bytes_codec` for  `zarr_format=3`  also be configurable  in `zarr.config`?

TODO:


brokkoli71:
I am a little confused: in which cases should we use VLenBytes and in which cases VLenUTF8?

brokkoli71:
Also, do we need a default compressor or are default filters sufficient?

src/zarr/codecs/_v2.py
```diff
@@ -43,10 +44,16 @@ async def _decode_single(
 
         # view as numpy array with correct dtype
         chunk = ensure_ndarray_like(chunk)
+        print(chunk)
+        print(chunk.dtype)
```
jhamman, line 48:
```suggestion
```

src/zarr/core/config.py
```diff
@@ -64,6 +64,11 @@ def reset(self) -> None:
             },
             "buffer": "zarr.core.buffer.cpu.Buffer",
             "ndbuffer": "zarr.core.buffer.cpu.NDBuffer",
+            "v2_dtype_kind_to_default_filters_and_compressor": {
+                "biufcmM": ["zstd"],
+                "U": ["vlen-utf8"],
+                "OSV": ["vlen-bytes"],
+            },
```
jhamman, line 71: This is quite a mouthful. Can we iterate a bit on the name of the config value and the keys?

src/zarr/core/metadata/v2.py
```diff
@@ -71,6 +71,14 @@ def __init__(
...
         dtype_parsed = parse_dtype(dtype)
         chunks_parsed = parse_shapelike(chunks)
+        if not filters and not compressor:
+            filters, compressor = _default_filters_and_compressor(dtype_parsed)
+        if dtype is str or dtype == "str":
+            vlen_codec: dict[str, JSON] = {"id": "vlen-utf8"}
+            if filters and not any(x["id"] == "vlen-utf8" for x in filters):
+                filters = list(filters) + [vlen_codec]
+            else:
+                filters = [vlen_codec]
```
jhamman, line 81: I'm wondering if this is really the right place for this. Or if it would be better higher up in the API (in the creation routines). My concern is that we may end up applying defaults to arrays that were intended to not use a filter/compressor.

jhamman (approved):
Thanks for working on this @brokkoli71. I think @normanrz is also going to review this but I wanted to also bring up an additional point. 

In Zarr2 we told people to set `zarr.storage.default_compressor = SomeCompressor()` This was simple but also an odd way to manage config. What we have now is much better. However, I wonder if we should do something to catch folks trying do set the `default_compressor` variable. Thoughts?

src/zarr/codecs/_v2.py
```diff
@@ -43,10 +44,16 @@ async def _decode_single(
...
         # special case object dtype, because incorrect handling can lead to
         # segfaults and other bad things happening
         if chunk_spec.dtype != object:
-            chunk = chunk.view(chunk_spec.dtype)
+            try:
+                chunk = chunk.view(chunk_spec.dtype)
+            except TypeError:
```
dstansby, line 54: If possible, could you add a comment explaining what this type error is catching? Is it perhaps compatibility for older versions of numpy?

src/zarr/core/config.py
dstansby: 👍 to changing these names. I'm afraid I don't really have enough context to suggest better names, but I think `v2_dtype_kind_to_default_filters_and_compressor` should be shorter, and `bitfuncM`, `U` and `OSV` should be a bit more descriptive (maybe one or two words instead of acroynms)

dstansby (changes requested):
Thanks! Couple of things:

- Print statements need removing
- The config naming could probably be improved

This would also be a good oppurtunity to update or add user facing documentation on what default compressors are used for what types of data - is that something you could add in this PR?

src/zarr/core/config.py
normanrz: What about?
```
            "v2_default_compressor": {
                "numeric": ["zstd"],
                "str": ["vlen-utf8"],
                "bytes": ["vlen-bytes"],
            },
```

src/zarr/core/metadata/v2.py
normanrz: I agree this should be higher up. Either in the (Async)Array.create or even in the top-level zarr.open

normanrz:
I think we should also have defaults for v3:
```
          "v3_default_codecs": {
                "numeric": ["bytes", "zstd"],
                "string": ["vlen-utf8"],
                "bytes": ["vlen-bytes"],
            },
```

LDeakin:
> I think we should also have defaults for v3:
> 
> ```
>           "v3_default_codecs": {
>                 "numeric": ["bytes", "zstd"],
>                 "string": ["vlen-utf8"],
>                 "bytes": ["vlen-bytes"],
>             },
> ```

zstd isn't in the spec yet: https://github.com/zarr-developers/zarr-specs/pull/256

brokkoli71:
thanks for your feedback, i will integrate it next week 👍🏼

src/zarr/core/config.py
brokkoli71: I agree with your suggestion. What do you think about `v2_default_compressors` (plural) as one could want multiple default filters and compressors?

normanrz: I think we should keep it as it is in v2. So 
```
            "v2_default_compressor": {
                "numeric": "zstd",
                "string": "vlen-utf8",
                "bytes": "vlen-bytes",
            },
```

brokkoli71:
> In Zarr2 we told people to set `zarr.storage.default_compressor = SomeCompressor()` This was simple but also an odd way to manage config. What we have now is much better. However, I wonder if we should do something to catch folks trying do set the `default_compressor` variable. Thoughts?

Maybe we can add a property `default_compressor` in `zarr.storage` that will give a warning if someone tries to change it? Not sure if this can be achieved with the `@property' decorator.

tests/test_config.py
```diff
@@ -63,6 +63,11 @@ def test_config_defaults_set() -> None:
                 "vlen-utf8": "zarr.codecs.vlen_utf8.VLenUTF8Codec",
                 "vlen-bytes": "zarr.codecs.vlen_utf8.VLenBytesCodec",
             },
+            "v2_default_compressor": {
```
normanrz, line 66: I would put this under `array`.

normanrz (commented):
> Should _get_default_array_bytes_codec for zarr_format=3 also be configurable in zarr.config?

Yes, I think it should.

normanrz:
> > In Zarr2 we told people to set `zarr.storage.default_compressor = SomeCompressor()` This was simple but also an odd way to manage config. What we have now is much better. However, I wonder if we should do something to catch folks trying do set the `default_compressor` variable. Thoughts?
> 
> Maybe we can add a property `default_compressor` in `zarr.storage` that will give a warning if someone tries to change it? Not sure if this can be achieved with the `@property' decorator.

I would lean towards raising an exception here with a message to use `zarr.config` instead. It is possible to override `__setattr__` and `__getattr__` on the module level: https://docs.python.org/3/reference/datamodel.html#customizing-module-attribute-access

src/zarr/storage/__init__.py
```diff
@@ -17,3 +22,20 @@
...
+
+class VerboseModule(ModuleType):
+    def __setattr__(self, attr: str, value: Any) -> None:
+        if attr == "default_compressor":
+            warnings.warn(
+                "setting zarr.storage.default_compressor is deprecated, use "
+                "zarr.config to configure array.v2_default_compressor "
+                "e.g. config.set({'codecs.zstd':'your.module.Zstd', 'array.v2_default_compressor.numeric': 'zstd'})",
+                DeprecationWarning,
+                stacklevel=1,
```
brokkoli71, line 35: not sure if this is the correct stacklevel

src/zarr/core/metadata/v3.py
```diff
@@ -606,6 +607,10 @@ def from_numpy(cls, dtype: np.dtype[Any]) -> DataType:
             return DataType.string
         elif dtype.kind == "S":
             return DataType.bytes
+        elif not _NUMPY_SUPPORTS_VLEN_STRING and dtype.kind == "O":
+            # numpy < 2.0 does not support vlen string dtype
+            # so we fall back on object array of strings
+            return DataType.string
```
brokkoli71, line 613: Do we support objects other than strings in v3?

normanrz: No, only `string` and `bytes`. But they seem to have another `kind`

src/zarr/storage/__init__.py
normanrz: I think this is fine.

normanrz (approved):
Thanks! Could you please go through the docstrings again and add some info about the default compressor, filters, codecs? Then, I think this is good to go.

src/zarr/api/asynchronous.py
```diff
@@ -401,7 +403,7 @@ async def save_array(
     arr : ndarray
         NumPy array with data to save.
     zarr_format : {2, 3, None}, optional
-        The zarr format to use when saving.
+        The zarr format to use when saving (default is 3).
```
dstansby, line 406:
```suggestion
        The zarr format to use when saving (default is 3 if not specified).
```

src/zarr/api/asynchronous.py
```diff
@@ -817,19 +819,40 @@ async def create(
     shape : int or tuple of ints
         Array shape.
     chunks : int or tuple of ints, optional
-        Chunk shape. If True, will be guessed from `shape` and `dtype`. If
-        False, will be set to `shape`, i.e., single chunk for the whole array.
-        If an int, the chunk size in each dimension will be given by the value
-        of `chunks`. Default is True.
+        The shape of the array's chunks.
+        V2 only. V3 arrays should use `chunk_shape` instead.
+        Default values are guessed based on the shape and dtype.
```
dstansby, line 824:
```suggestion
        If not specified, default values are guessed based on the shape and dtype.
```

src/zarr/api/asynchronous.py
```diff
@@ -817,19 +819,40 @@ async def create(
...
     dtype : str or dtype, optional
         NumPy dtype.
+    chunk_shape : int or tuple of ints, optional
+        The shape of the Array's chunks (default is None).
+        V3 only. V2 arrays should use `chunks` instead.
+    chunk_key_encoding : ChunkKeyEncoding, optional
+        A specification of how the chunk keys are represented in storage.
+        V3 only. V2 arrays should use `dimension_separator` instead.
+        Default is ("default", "/").
```
dstansby, line 833:
```suggestion
        Default is ``("default", "/")``.
```

src/zarr/api/asynchronous.py
```diff
@@ -817,19 +819,40 @@ async def create(
...
+    codecs : Sequence of Codecs or dicts, optional
+        An iterable of Codec or dict serializations thereof. The elements of
```
dstansby, line 835:
```suggestion
        An iterable of Codec or dict serializations of Codecs. The elements of
```

src/zarr/api/asynchronous.py
```diff
@@ -817,19 +819,40 @@ async def create(
...
+        this collection specify the transformation from array values to stored bytes.
+        V3 only. V2 arrays should use `filters` and `compressor` instead.
+        If no codecs are provided, default codecs will be used:
+        - For numeric arrays, the default is `BytesCodec` and `ZstdCodec`.
```
dstansby, line 839:
```suggestion
        - For numeric arrays, the default is `BytesCodec` and `ZstdCodec`.
```
Can we also document the default compression level (and any other parameters) here?

src/zarr/api/asynchronous.py
```diff
@@ -817,19 +819,40 @@ async def create(
...
+        - For Unicode strings, the default is `VLenUTF8Codec`.
```
dstansby, line 840: Are the codecs here meant to be links to their API reference in numcodecs? If so, they are not working.

src/zarr/api/asynchronous.py
```diff
@@ -817,19 +819,40 @@ async def create(
...
+        - For bytes or objects, the default is `VLenBytesCodec`.
+        These defaults can be changed using the `array.v3_default_codecs` variable in the Zarr config.
```
dstansby, line 842: This link is not working. It should either be fixed, or some other way of liking to docs in the zarr config should be added.

src/zarr/api/asynchronous.py
```diff
@@ -817,19 +819,40 @@ async def create(
...
     compressor : Codec, optional
-        Primary compressor.
+        Primary compressor to compress chunk data.
+        V2 only. V3 arrays should use `codecs` instead.
```
dstansby, line 845:
```suggestion
        V2 only. V3 arrays should use ``codecs`` instead.
```

src/zarr/api/asynchronous.py
```diff
@@ -817,19 +819,40 @@ async def create(
...
+        If neither `compressor` nor `filters` are provided, a default compressor will be used:
```
dstansby, line 846:
```suggestion
        If neither ``compressor`` nor ``filters`` are provided, a default compressor will be used:
```
Single ticks are sphinx links, double backticks are code (and I think these should be code instead of attempting to link somewhere)

src/zarr/api/asynchronous.py
```diff
@@ -817,19 +819,40 @@ async def create(
...
+        - For numeric arrays, the default is `ZstdCodec`.
+        - For Unicode strings, the default is `VLenUTF8Codec`.
+        - For bytes or objects, the default is `VLenBytesCodec`.
+        These defaults can be changed using the `array.v2_default_compressor` variable in the Zarr config.
```
dstansby, line 850: Same comments as above re.:
- Broken links to codecs
- Adding a working link to config docs

src/zarr/api/asynchronous.py
```diff
@@ -817,19 +819,40 @@ async def create(
...
     fill_value : object
         Default value to use for uninitialized portions of the array.
     order : {'C', 'F'}, optional
         Memory layout to be used within each chunk.
-        Default is set in Zarr's config (`array.order`).
+        Default is specified in the Zarr config `array.order`.
```
dstansby, line 855:
```suggestion
        If not specified, default is taken from the Zarr config ``array.order``.
```

dstansby: Should also have a link to the config docs

src/zarr/api/asynchronous.py
```diff
@@ -844,6 +867,8 @@ async def create(
         for storage of both chunks and metadata.
     filters : sequence of Codecs, optional
         Sequence of filters to use to encode chunk data prior to compression.
+        V2 only. If neither `compressor` nor `filters` are provided, a default
```
dstansby, line 870:
```suggestion
        V2 only. If neither ``compressor`` nor ``filters`` are provided, a default
```

src/zarr/api/asynchronous.py
```diff
@@ -844,6 +867,8 @@ async def create(
...
+        compressor will be used. (see `compressor` for details)
```
dstansby, line 871:
```suggestion
        compressor will be used. (see ``compressor`` for details).
```

src/zarr/api/asynchronous.py
```diff
@@ -859,7 +884,8 @@ async def create(
         A codec to encode object arrays, only needed if dtype=object.
     dimension_separator : {'.', '/'}, optional
         Separator placed between the dimensions of a chunk.
-
+        V2 only. V3 arrays should use `chunk_key_encoding` instead.
```
dstansby, line 887:
```suggestion
        V2 only. V3 arrays should use ``chunk_key_encoding`` instead.
```

src/zarr/api/asynchronous.py
```diff
@@ -894,9 +921,13 @@ async def create(
...
-    if zarr_format == 2 and chunks is None:
-        chunks = shape
-    elif zarr_format == 3 and chunk_shape is None:
+    if zarr_format == 2:
+        if chunks is None:
+            chunks = shape
+        dtype = parse_dtype(dtype, zarr_format)
+        if not filters and not compressor:
+            filters, compressor = _default_filters_and_compressor(dtype)
+    elif zarr_format == 3 and chunk_shape is None:  # type: ignore[redundant-expr]
```
dstansby, line 930:
```suggestion
    elif chunk_shape is None:
```
Since by this point zarr_format is definitely either 2 or 3. My suggestion is a bit less readable though, so feel free to take or leave.

src/zarr/core/array.py
```diff
@@ -634,6 +642,14 @@ async def _create_v2(
         if dimension_separator is None:
             dimension_separator = "."
 
+        dtype = parse_dtype(dtype, 2)
```
dstansby, line 645:
```suggestion
        dtype = parse_dtype(dtype, zarr_format=2)
```
It wasn't clear to me what 2 was doing without the argument name

tests/test_v2.py
```diff
@@ -177,3 +200,27 @@ def test_v2_non_contiguous(array_order: Literal["C", "F"], data_order: Literal["
         assert a.flags.c_contiguous
     arr[slice(6, 9, None), slice(3, 6, None)] = a
     np.testing.assert_array_equal(arr[slice(6, 9, None), slice(3, 6, None)], a)
+
+
+def test_default_compressor_deprecation_warning():
+    with pytest.warns(DeprecationWarning):
```
dstansby, line 206: This should have a `match` argument to check the warning string

src/zarr/storage/__init__.py
```diff
@@ -17,3 +22,20 @@
...
 ]
+
+
+class VerboseModule(ModuleType):
+    def __setattr__(self, attr: str, value: Any) -> None:
+        if attr == "default_compressor":
+            warnings.warn(
+                "setting zarr.storage.default_compressor is deprecated, use "
+                "zarr.config to configure array.v2_default_compressor "
+                "e.g. config.set({'codecs.zstd':'your.module.Zstd', 'array.v2_default_compressor.numeric': 'zstd'})",
```
dstansby, line 33:
```suggestion
                "e.g. config.set({'codecs.zstd':'numcodecs.Zstd', 'array.v2_default_compressor.numeric': 'zstd'})",
```
I think better to use a numcodecs example here?

dstansby (changes requested):
I've left lots of small requests for changes, mainly around the docstrings - in general they are great and clear, but I think worth fixing a lot of the little issues while we're here.

I left most of the docstring comments in asynchronous.py, but they also apply to the other files that have updated docstrintgs.

src/zarr/api/asynchronous.py
brokkoli71: I did not intend to create links. How would you link the API reference? Is there another way but to manually add markdown links like ```[`VLenUTF8Codec`](https://numcodecs.readthedocs.io/en/stable/other/vlen.html#vlenutf8)```
Do you think we should add them?

src/zarr/api/asynchronous.py
brokkoli71: I had the same thought and preferred having `zarr_format == 3` for readability

src/zarr/api/asynchronous.py
dstansby: For ease here, maybe lets leave the links for now and just put these in double backticks. We can add the links later in another PR

dstansby (approved):

d-v-b:
is there a reason why none of the default compressors / codecs have a configuration?

d-v-b:
and a second question, why aren't strings / bytes compressed with `zstd` by default, like numeric dtypes?

PR #2471, opened 2024-11-07
Remove useless `{...!s}`

DimitriPapadopoulos:
Remove useless `!s` in `f"...{...!s}..."`.

### DIFF
--- a/src/zarr/storage/logging.py
+++ b/src/zarr/storage/logging.py
@@ -55,7 +55,7 @@ class LoggingStore(Store):
         self, log_level: str = "DEBUG", log_handler: logging.Handler | None = None
     ) -> None:
         self.log_level = log_level
-        self.logger = logging.getLogger(f"LoggingStore({self._store!s})")
+        self.logger = logging.getLogger(f"LoggingStore({self._store})")
         self.logger.setLevel(log_level)
 
         if not self.logger.hasHandlers():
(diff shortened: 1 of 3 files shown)


d-v-b:
I don't actually know, but I do know that officially `!s` is redundant: https://peps.python.org/pep-0498/#s-r-and-a-are-redundant, so if we do want to be sure we are calling `str()` in the fstring we should directly call `str`.

DimitriPapadopoulos:
From [PEP 3101](https://peps.python.org/pep-3101/):
> ### [Explicit Conversion Flag](https://peps.python.org/pep-3101/#explicit-conversion-flag)
> 
> The explicit conversion flag is used to transform the format field value before it is formatted. This can be used to override the type-specific formatting behavior, and format the value as if it were a more generic type.

In the absence of an explicit conversion flag, the value is simply formatted, and the default formatting converts to a `str` first:
> ### [Controlling Formatting on a Per-Type Basis](https://peps.python.org/pep-3101/#controlling-formatting-on-a-per-type-basis)
> [...]
> The `object.__format__` method is the simplest: It simply converts the object to a string, and then calls format again:
> ```python
> class object:
>     def __format__(self, format_spec):
>         return format(str(self), format_spec)
> ```

I don't think `!s` was added for a reason here.

d-v-b (approved):
lgtm

PR #2511, opened 2024-11-21
Remove asciitree dependency

### LINKED ISSUE #1370: asciitree has no wheels
### Zarr version

all

### Numcodecs version

NA

### Python Version

all

### Operating System

all

### Installation

pip install --only-binary ':all:' zarr

### Description

Zarr depends on [asciitree](https://github.com/mbr/asciitree), which does not publish wheel.

When trying to install and avoiding tar.gz for various reasons this make it harder to install zarr.

The `--only-binary ':all:' ` is useful in CI to make sure that binary artifact are available for all dependencies and usually to make sure the end user does not need to have a compiler. In some cases, you may want to ensure all your dependencies have wheels as you do not want code execution at install time. Plus, even for python, wheel install is usually faster, and dependency resolution does not need to run `setup.py`.

Because if depends on asciitree, zarr is one package cannot be installed with `--only-binary ':all:' ` flag, and contaminate all the dependees. (you can use --no-binary asciitree to work around this specific dependency so not critical)

It would be great zarr would only rely on packages that have wheels.

As asciitree seem unmaintained, as author does not resond to [issue asking for wheels](https://github.com/mbr/asciitree/issues/18), nor have seen a commit in 3years.

This is likely because its feature complete?

So solution might be:
 - Vendor asciitree
 - Repackage with wheel under a different package name (but same import name).
 - Custom code.

### Steps to reproduce

```
$ pip install --only-binary ':all:' zarr
Collecting zarr
  Using cached zarr-2.14.2-py3-none-any.whl (203 kB)
  ...
  Using cached zarr-2.5.0-py3-none-any.whl (131 kB)
ERROR: Cannot install zarr==2.10.0, ... zarr==2.9.4 and zarr==2.9.5 because these package versions have conflicting dependencies.

The conflict is caused by:
    zarr 2.14.2 depends on asciitree
    ...
    zarr 2.5.0 depends on asciitree

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict

ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/topics/dependency-resolution/#dealing-with-dependency-conflicts
```

### Additional output

_No response_

normanrz:
Remove asciitree dependency. I couldn't find any use of it in the codebase.

Fixes #1370
Contributes to #2391

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,9 +9,9 @@ repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
     rev: v0.7.4
     hooks:
-      - id: ruff
-        args: ["--fix", "--show-fixes"]
-      - id: ruff-format
+    - id: ruff
+      args: ["--fix", "--show-fixes"]
+    - id: ruff-format
   - repo: https://github.com/codespell-project/codespell
     rev: v2.3.0
     hooks:
@@ -20,7 +20,7 @@ repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v5.0.0
     hooks:
-      - id: check-yaml
+    - id: check-yaml
   - repo: https://github.com/pre-commit/mirrors-mypy
     rev: v1.13.0
     hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,9 +9,9 @@ repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
     rev: v0.7.4
     hooks:
-    - id: ruff
-      args: ["--fix", "--show-fixes"]
-    - id: ruff-format
+      - id: ruff
+        args: ["--fix", "--show-fixes"]
+      - id: ruff-format
   - repo: https://github.com/codespell-project/codespell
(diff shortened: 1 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -9,9 +9,9 @@ repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
     rev: v0.7.4
     hooks:
-    - id: ruff
-      args: ["--fix", "--show-fixes"]
-    - id: ruff-format
+      - id: ruff
+        args: ["--fix", "--show-fixes"]
+      - id: ruff-format
```
d-v-b, line 14: Is this indentation expected?

normanrz: No, that was just my auto-formatter.

d-v-b (approved):
thanks for cleaning this up!

normanrz:
cc @oanegros

d-v-b:
I think that statement is scoped to the time when it was written, which was a time when those were in fact the minimum dependencies. So removing `asciitree` would actually make the document inaccurate.

PR #2512, opened 2024-11-22
Error when attempting to set with an array of incompatible shape.

### LINKED ISSUE #2469: Unexpected Vectorized Indexing Setting Behavior
### Zarr version

v3.0.0-beta.1 (main)

### Numcodecs version

0.13.0

### Python Version

3.12

### Operating System

Mac

### Installation

`uv` from repo

### Description

I think this is vectorized indexing...in any case I think that the following example should error since `arr[:][[np.array([1, 2]), np.array([1, 2])]]` is 1d but it allows storing a 2d

### Steps to reproduce

```python
import tempfile
import numpy as np
import zarr

shape = (4, 4)
chunks = (2, 2)
fill_value = 32767

tmp = tempfile.TemporaryDirectory()
arr = zarr.create(
    shape,
    store=LocalStore(root=tmp.name, mode="w"),
    chunks=chunks,
    dtype=np.int16,
    fill_value=fill_value,
    codecs=[zarr.codecs.BytesCodec(), zarr.codecs.BloscCodec()],
)

arr[np.array([1, 2]), np.array([1, 2])] = np.array([[-1, -2], [-3, -4]])
print(arr[:])
```
gives
```
[[32767 32767 32767 32767]
 [32767    -1 32767 32767]
 [32767 32767    -2 32767]
 [32767 32767 32767 32767]]
```
which I would not expect - I would probably expect this to error.

### Additional output

_No response_

dcherian:
Closes #2469

[Description of PR]

TODO:

### DIFF
--- a/src/zarr/core/array.py
+++ b/src/zarr/core/array.py
@@ -2882,7 +2882,7 @@ class Array:
             value = np.array(value).reshape(-1)
 
         if not is_scalar(value, self.dtype) and (
-            isinstance(value, np.ndarray) and indexer.shape != value.shape
+            isinstance(value, NDArrayLike) and indexer.shape != value.shape
         ):
             raise ValueError(
                 f"Attempting to set a selection of {indexer.sel_shape[0]} "
--- a/src/zarr/core/array.py
+++ b/src/zarr/core/array.py
@@ -2842,24 +2842,22 @@ class Array:
             try:
                 from numcodecs.compat import ensure_ndarray_like
 
-                value_array = ensure_ndarray_like(value)  # TODO replace with agnostic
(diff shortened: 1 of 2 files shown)


src/zarr/core/array.py
```diff
@@ -2881,6 +2881,14 @@ def set_coordinate_selection(
         if hasattr(value, "shape") and len(value.shape) > 1:
             value = np.array(value).reshape(-1)
 
+        if not is_scalar(value, self.dtype) and (
+            isinstance(value, np.ndarray) and indexer.shape != value.shape
```
normanrz, line 2885: I guess this would need to check for `is_ndarray_like`

dcherian: done, thanks!

PR #2514, opened 2024-11-22
Added Array.info_complete

TomAugspurger:
Now that Store.getsize is a thing, we can do info_complete which includes the number of chunks written and the size of those bytes.

The current implementation unfortunately does two `list_prefix`es on the same prefix. The first to get the count of chunks initialized and the second to get the bytes stored under a prefix. Unfortunately, these down compose well. I haven't thought of a nice way to eliminate that yet. We can do this naively by doing a single `list_prefix` and then counting the number of keys as we call `getsize` on each. But we also have a `Store.getsize_prefix` for those stores that have fastpaths for getting the total number of bytes under a prefix. I don't really want a `Store.getsize_and_count_prefix`, but maybe some kind of `Store.statistics_prefix`? Probably not worth worrying about today.

### DIFF
--- a/src/zarr/core/array.py
+++ b/src/zarr/core/array.py
@@ -1346,18 +1346,53 @@ class AsyncArray(Generic[T_ArrayMetadata]):
         AsyncArray.info_complete
             All information about a group, including dynamic information
             like the number of bytes and chunks written.
+
+        Examples
+        --------
+
+        >>> arr = await zarr.api.asynchronous.create(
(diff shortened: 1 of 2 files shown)


rabernat (approved):
Thanks so much for this Tom!

The double listing will definitely add some latency. But we can work on that in a future PR.

I could imagine an API like

```python
nitems, total_size = store.statistics(prefix, ("count", "size"))
```

PR #2515, opened 2024-11-24
[v3] Makes data contiguous in v2 codec

### LINKED ISSUE #2501: bug in v2 codec pipeline
### Zarr version

main

### Description

There is a bug in these lines:

`cdata` is not of type bytes in the `else` clause:

https://github.com/zarr-developers/zarr-python/blob/eb85ee7ba32df4983b80ae9e60686e7d2f35349e/src/zarr/codecs/_v2.py#L80-L86

cc @normanrz 

### Steps to reproduce

This is triggering failures in xarray's test suite. I haven't succeeded at making a minimal reproducer.

### Additional output

_No response_

normanrz:
fixes #2501

### DIFF
--- a/src/zarr/codecs/_v2.py
+++ b/src/zarr/codecs/_v2.py
@@ -69,7 +69,7 @@ class V2Codec(ArrayBytesCodec):
         chunk = chunk_array.as_ndarray_like()
 
         # ensure contiguous and correct order
-        chunk = chunk.astype(chunk_spec.dtype, order=chunk_spec.order)
+        chunk = chunk.astype(chunk_spec.dtype, order=chunk_spec.order, copy=False)
 
         # apply filters
         if self.filters:
--- a/src/zarr/codecs/_v2.py
+++ b/src/zarr/codecs/_v2.py
@@ -5,6 +5,7 @@ from dataclasses import dataclass
 from typing import TYPE_CHECKING
 
 import numcodecs
+import numpy as np
 from numcodecs.compat import ensure_ndarray_like
 
 from zarr.abc.codec import ArrayBytesCodec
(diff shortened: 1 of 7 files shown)


tests/test_v2.py
```diff
@@ -132,3 +132,19 @@ def test_v2_filters_codecs(filters: Any) -> None:
...
+        MemoryStore({}),
+        shape=(10, 8),
+        chunks=(3, 3),
+        fill_value=np.nan,
+        dtype="float64",
+        zarr_format=2,
+        exists_ok=True,
+    )
+    a = np.ones(arr.shape)
+    arr[slice(6, 9, None), slice(3, 6, None)] = a[
```
dcherian, line 148: can we also test assigning F- and C-contiguous arrays please?

src/zarr/codecs/_v2.py
```diff
@@ -83,6 +92,7 @@ async def _encode_single(
         else:
             cdata = chunk
 
+        cdata = ensure_contiguous(ensure_ndarray_like(cdata))
```
dcherian, line 95: an alternative is https://numpy.org/doc/stable/reference/generated/numpy.ndarray.tobytes.html but I'm not sure if that always copies.

tests/test_v2.py
normanrz: Good catch! I added more tests and fixed another issue.

src/zarr/codecs/_v2.py
normanrz: `ensure_bytes` essentially calls `arr.tobytes` but adds a few checks.

src/zarr/codecs/_v2.py
```diff
@@ -68,6 +68,9 @@ async def _encode_single(
     ) -> Buffer | None:
         chunk = chunk_array.as_ndarray_like()
 
+        # ensure contiguous and correct order
+        chunk = chunk.astype(chunk_spec.dtype, order=chunk_spec.order)
```
dcherian, line 72:
```suggestion
        chunk = chunk.astype(chunk_spec.dtype, order=chunk_spec.order, copy=False)
```

This should only copy if needed

dcherian (approved):

PR #2474, opened 2024-11-08
Feat/latency store

d-v-b:
This PR adds a new store class, `WrapperStore`, that implements `Store` but wraps an existing store instance and delegates all of its methods to those of the wrapped store. I  then use `WrapperStore` to define a store class that exists for performance testing, called `LatencyStore`, which (at the moment) adds parametric latency to the `set` and `get` methods.

TODO:
* [ ] Add unit tests and/or doctests in docstrings
* [ ] Add docstrings and API docs for any new/modified user-facing classes and functions
* [ ] New/modified features documented in docs/tutorial.rst
* [ ] Changes documented in docs/release.rst
* [ ] GitHub Actions have all passed
* [ ] Test coverage is 100% (Codecov passes)

PR #2519, opened 2024-11-26
Feat/concurrent members

d-v-b:
Makes `AsyncGroup.members()` fetch keys and fetch metadata concurrently, which provides a big performance win for high-latency storage backends. The number of concurrent operations is limited by the zarr-wide configuration setting. 

in `main`, `Group.members()` requires `~O(num_members)` time to complete, because it does not perform IO concurrently. In this PR, `Group.members` runs in constant time (until the number of concurrent requests exceeds the concurrency limit).


d-v-b:
note: this PR depends on #2474

src/zarr/core/group.py
```diff
@@ -686,21 +684,18 @@ async def getitem(
...
             else:
-                zgroup = (
-                    json.loads(zgroup_bytes.to_bytes())
-                    if zgroup_bytes is not None
-                    else {"zarr_format": self.metadata.zarr_format}
-                )
-                zarr_json = {**zgroup, "attributes": zattrs}
-                return type(self).from_dict(store_path, zarr_json)
+                # this is just for mypy
+                assert zgroup is not None
```
dcherian, line 696:
```suggestion
                if TYPE_CHECKING:
                    assert zgroup is not None
```

src/zarr/core/group.py
```diff
@@ -2627,3 +2600,191 @@ def array(
                 )
             )
         )
+
+
+async def _getitem_semaphore(
```
dcherian, line 2605: Is there a reason we need to passing a semaphore around?


FWIW we have this utility function in Arraylake that gets decently wide use.

```python
async def async_gather(*coros, concurrency: Optional[int] = None, return_exceptions: bool = False) -> list[Any]:
    """Execute a gather while limiting the number of concurrent tasks.

    Args:
        coros: coroutines
            list of coroutines to execute
        concurrency: int
            concurrency limit
            if None, defaults to config_obj.get('async.concurrency', 4)
            if <= 0, no concurrency limit

    """
    if concurrency is None:
        concurrency = int(config_obj.get("async.concurrency", 4))

    if concurrency > 0:
        # if concurrency > 0, we use a semaphore to limit the number of concurrent coroutines
        semaphore = asyncio.Semaphore(concurrency)

        async def sem_coro(coro):
            async with semaphore:
                return await coro

        results = await asyncio.gather(*(sem_coro(c) for c in coros), return_exceptions=return_exceptions)
    else:
        results = await asyncio.gather(*coros, return_exceptions=return_exceptions)

    return results
```

src/zarr/core/group.py
```diff
@@ -2627,3 +2600,191 @@ def array(
...
+        An optional semaphore to use for concurrency control.
+
+    Yields
+    ------
+    tuple[str, AsyncArray[ArrayV3Metadata] | AsyncArray[ArrayV2Metadata] | AsyncGroup]
+    """
+
+    # retrieve keys from storage
+    keys = [key async for key in node.store.list_dir(node.path)]
+    keys_filtered = tuple(filter(lambda v: v not in skip_keys, keys))
```
dcherian, line 2647:
```suggestion
    keys_filtered = [key async for key in node.store.list_dir(node.path) if key not in skip_keys]
```

tests/test_group.py
```diff
@@ -1420,3 +1486,67 @@ def test_delitem_removes_children(store: Store, zarr_format: ZarrFormat) -> None
     del g1["0"]
     with pytest.raises(KeyError):
         g1["0/0"]
+
+
+@pytest.mark.parametrize("store", ["memory"], indirect=True)
+def test_group_members_performance(store: MemoryStore) -> None:
```
dcherian, line 1492: I like these two tests!

src/zarr/core/group.py
d-v-b: for debugging purposes I actually like having `keys` and `keys_filtered` as separate variables, since they represent logically distinct operations that could either (or both) fail, so I'd prefer to keep that unless there's a major objection.

src/zarr/core/group.py
d-v-b: I'm up for implementing an `as_completed_with_sempahore` function, as I'm sure we will need that in more places. But i'm not sure if that should be a blocker for this pr

src/zarr/core/group.py
dcherian: No I dont feel strongly about this

src/zarr/core/group.py
```diff
@@ -2712,3 +2686,191 @@ def array(
...
+
+    for fetched_node_coro in asyncio.as_completed(node_tasks):
+        try:
+            fetched_node = await fetched_node_coro
+        except KeyError as e:
+            # keyerror is raised when `key` names an object (in the object storage sense),
+            # as opposed to a prefix, in the store under the prefix associated with this group
+            # in which case `key` cannot be the name of a sub-array or sub-group.
+            warnings.warn(
+                f"Object at {e.args[0]} is not recognized as a component of a Zarr hierarchy.",
```
TomAugspurger, line 2748: It's *technically* possible for `KeyError.args` to be empty, so this `[0]` would raise an IndexError:

```python
IndexError                                Traceback (most recent call last)
Cell In[10], line 1
----> 1 KeyError().args[0]
```

If we're comfortable assuming / requiring that things raising a KeyError here will populate that with the key, then I think this is fine to ignore. Otherwise, we might want to catch that IndexError.

src/zarr/core/group.py
```diff
@@ -2712,3 +2686,191 @@ def array(
...
+async def _getitem_semaphore(
+    node: AsyncGroup, key: str, semaphore: asyncio.Semaphore | None
+) -> AsyncArray[ArrayV3Metadata] | AsyncArray[ArrayV2Metadata] | AsyncGroup:
+    """
+    Combine node.getitem with an optional semaphore. If the semaphore parameter is an
+    asyncio.Semaphore instance, then the getitem operation is performed inside an async context
+    manager provided by that semaphore. If the semaphore parameter is None, then getitem is invoked
+    without a context manager.
+    """
+    if semaphore is not None:
```
TomAugspurger, line 2700: Style nitpick: I like `contextlib.nullcontext` for cases where you may or may not have a context manager: https://docs.python.org/3/library/contextlib.html#contextlib.nullcontext

3.10 added support for asynchronous context managers, so this should be usable here:

```python
semaphore = semaphore or contextlib.nullcontext()
async with semaphore:
    ....
```

And I wonder whether `getitem_semaphore` should be the one to look up `async.concurrency` from the zarr config? Maybe we want that to be the default so that we don't miss it anywhere (and have some other way to indicate unbounded concurrency?)

d-v-b: I tried `contextlib.nullcontext` but I didn't feel like the LOC saved was worth the added indirection in this case. I _assume_ that `contextlib.nullcontext` doesn't add notable performance overhead, by contrast, I find the current `if... else` construction easier to reason about, and it requires 1 fewer import. 

> And I wonder whether getitem_semaphore should be the one to look up async.concurrency from the zarr config? Maybe we want that to be the default so that we don't miss it anywhere (and have some other way to indicate unbounded concurrency?)

As I understand it, we only query the concurrency limit when the semaphore is created, which is necessarily before `getitem_semaphore` gets called. All invocations of `getitem_semaphore` have to use the same `Semaphore` instance, because otherwise there would be no coordination mechanism for rate-limiting. So I think that means `getitem_semaphore` shouldn't know anything about the config, or even concurrency limits.

src/zarr/core/group.py
d-v-b: I'm _pretty_ sure that `args` will always be populated, because we are handling exceptions coming downstream of a function that necessarily takes a concrete key to query from storage. For that reason, even if I did add code to handle the case when `e.args` is not populated, I have no idea how we would test this case with our current `Group` api, since there's no way that I know of to reach this line without some concrete keys.

TomAugspurger (approved):

dcherian (approved):

d-v-b:
@dstansby how can I re-rerun the docs build? and do you have any idea why it failed?

dstansby:
Do you have a readtheodcs account? If you send me the email or username for that, I can add you as a maintainer to the project on readthedocs.org. Then you'll get a "Rebuild this build" button:

Screenshot 2025-01-07 at 14 22 48

I'm not sure why this one failed... I will restart!

d-v-b:
thanks for fixing it @dstansby!

PR #2521, opened 2024-11-29
add a complete step to the test CI

normanrz:
Adds a "test complete" step to the test CI pipeline so that we can use that as a requirement for auto-merging.


dstansby:
Ooh I like this, is the idea not to have to update the required GH actions jobs every time a test name changes (e.g., when a Python version is updated)?

normanrz:
Exactly. This should make it easier to define required CI runs for Github's auto-merge functionality without having to select all runs in all matrix configurations.

jhamman:
Let's give this a try.

dstansby (approved):

PR #2533, opened 2024-12-04
Wrap sync fs for xarray.to_zarr

moradology:
This PR automatically wraps synchronous filesystems in the `RemoteStore.from_url` function. This is necessary unless we provide a `force_async` argument on `fsspec`s `url_to_fs` function.

I'm not sure adding more flags (even if we sugar them with default values) is the best move. Another option might be to implement a `url_to_asyncfs` method that simply wraps `url_to_fs` and wraps any synchronous filesystems that come out automatically but that, too, would require changes here.

This PR is in service of downstream kerchunk compatibility: https://github.com/fsspec/kerchunk/pull/516

### DIFF
--- a/src/zarr/storage/remote.py
+++ b/src/zarr/storage/remote.py
@@ -166,6 +166,9 @@ class RemoteStore(Store):
         opts = {"asynchronous": True, **opts}
 
         fs, path = url_to_fs(url, **opts)
+        if not fs.async_impl:
+            from fsspec.implementations.asyn_wrapper import AsyncFileSystemWrapper
+            fs = AsyncFileSystemWrapper(fs)
 
         # fsspec is not consistent about removing the scheme from the path, so check and strip it here
         # https://github.com/fsspec/filesystem_spec/issues/1722


moradology:
This fix is implicated in the issue here: https://github.com/zarr-developers/zarr-python/issues/2554

jhamman:
@moradology - is there interest in finishing this up? The proposed changes look good but we'll want a test to make sure this is working as expected.

moradology:
Yeah, definite interest in getting back to this - just at the tail end of a long period of vacation/travel and will have time in the coming week!

src/zarr/storage/_fsspec.py
```diff
@@ -3,6 +3,8 @@
 import warnings
 from typing import TYPE_CHECKING, Any
 
+from fsspec.implementations.asyn_wrapper import AsyncFileSystemWrapper
```
martindurant, line 6: This makes the whole module depend on fsspec. I'm OK with that.

martindurant (approved):

src/zarr/storage/_fsspec.py
moradology: Yeah, this is definitely unfortunate but a bit difficult to avoid

jhamman: why can't this import be done in or around line 162?

moradology: Happy to do that (and add a bit of caution around the possibly failing import depending on fsspec versions)

martindurant:
The kerchunk test suite makes extensive use of xr.to_zarr("memory://..."), which currently doesn't work without this change.

- I am happy to move the import as requested, to get that done
- I can fix the test to skip for mindeps

OK to prodeed

cc @jhamman

martindurant:
OK, I have no idea what labeler is or why it's not happy. I added one label ("bug") to this PR to be certain.

moradology:
@martindurant I think I figured the labeler's strange behavior out. Should be fixed once I get eyes+approval on #2747

dstansby:
I wouldn't worry too much about the broken labeller - this (and other PRs) can definitely be merged before it's fixed (although we should definitely fix it!)

PR #2534, opened 2024-12-05
chore(deps): make fsspec and upath optional dependencies

### LINKED ISSUE #735: Update fsspec minimal version?

### LINKED ISSUE #2391: Rethinking Zarr's core dependencies

jhamman:
This PR makes a few minor changes to our project dependencies and test imports. 

closes #2391
Fixes #735
Supersedes #2395

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,8 +28,7 @@ requires-python = ">=3.11"
 dependencies = [
     'numpy>=1.25',
     'numcodecs[crc32c]>=0.14',
-    'fsspec>=2022.10.0',
-    'typing_extensions>=4.6',
+    'typing_extensions>=4.9',
     'donfig>=0.8',
 ]
 
@@ -54,16 +53,16 @@ license = {text = "MIT License"}
 keywords = ["Python", "compressed", "ndimensional-arrays", "zarr"]
 
 [project.optional-dependencies]
+fsspec = [
+    "fsspec>=2023.10.0",
+]
 test = [
     "coverage",
     "pytest",
     "pytest-cov",
-    "msgpack",
     "s3fs",
     "pytest-asyncio",
     "moto[s3]",
-    "flask-cors",
-    "flask",
     "requests",
     "mypy",
     "hypothesis",
--- a/tests/test_store/test_core.py
+++ b/tests/test_store/test_core.py
@@ -3,7 +3,6 @@ from pathlib import Path
 
 import pytest
 from _pytest.compat import LEGACY_PATH
-from upath import UPath
 
 from zarr.core.common import AccessModeLiteral
 from zarr.storage._utils import normalize_path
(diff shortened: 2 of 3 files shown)


dstansby (changes requested):
This is a big change in user facing behaviour when the package is installed, so should definitely have a prominent entry in the what's new/changed for 3.0 docs!

This probably also needs a change to the installation docs to explain how to install packages for remote data access.

d-v-b (approved):

dstansby:
> This is not quite right. Fsspec is an optional dependency for Zarr-Python 2 as well. I'm planning to do a rev on the docs next week so will include this then.

Ah my bad. I think it's still worth adding a quick changelog note somewhere that this is a change in behaviour from the current beta release - shouldn't take much time to add, and I think adds a lot of value for users if we have a user-facing log of changes like these, even if from beta to beta to release.

dstansby (approved):
👍 thanks!

Issue #2530, opened 2024-12-03
Shape and chunks accept float values when creating an array


faymanns:
### Zarr version

v2.18.3

### Numcodecs version

v0.12.1

### Python Version

v3.11.2

### Operating System

Linux

### Installation

Using pip into virtual environment

### Description

When you create a new array (e.g. using `zarr.zeros`), the tuples for shape and chunks can contain float values.
The values after the decimal point are trimmed without a warning.
Is this the intended behavior?
I would expect `zarr-python` to raise an exception or at least a warning.

### Steps to reproduce

```
import zarr

z = zarr.zeros((100.4, 100.9), chunks=(10.5, 10.1), dtype="i4")
print(z.shape)
print(z.chunks)
```
As you can see from the output the float values are trimmed without warning:
```
(100, 100)
(10, 10)
```

faymanns:
Thanks for the reply @d-v-b! I'd be happy to submit a pull request for the chunks following the way the shape input is parsed.
I guess parsing the input for `chunks` and `chunk_shape` with [`parse_shapelike`](https://github.com/zarr-developers/zarr-python/blob/bda158a600592c1eb90706980f8c76f27e6a10a8/src/zarr/core/common.py#L132) inside [`create`](https://github.com/zarr-developers/zarr-python/blob/bda158a600592c1eb90706980f8c76f27e6a10a8/src/zarr/core/array.py#L363) should be sufficient?

d-v-b:
yes that sounds like the right direction.

dstansby:
> Thanks for the bug report; we are working on version 3 of zarr-python so I don't think we are going to make any more releases in the 2.x series

@d-v-b I thought we were supporting verseion 2 for (at least) six months after release of version 3? So I think it would be reasonable to fix this in the version 2 branch, and I'd be happy to review a PR to version 2 and do a bugfix release.

d-v-b:
@dstansby that makes sense! In that case we should fix 2.x and 3.x

faymanns:
My PR for the main branch has been merged. Should I submit another PR for the v2 branch before closing this issue?

dstansby:
A PR to the v2 branch would be great - I would say that instead of raising an error, we should raise a warning on the v2 branch because an error would technically be a breaking change.

faymanns:
Sorry that it took me so long to do this. I created a pull request (#2579) for v2 that raises warnings.
[...]

PR #2535, opened 2024-12-05
Parse chunk shape to check for float values

faymanns:
This PR solves the fact that specifying a chunk shape containing one or several float values does not raise an error.
Before the PR the values after the decimal point are trimmed without a warning.
After the PR a `TypeError` is raised when the chunk shape contains float values.
It solves issue #2530.

TODO:


d-v-b (approved):
looks good, thank you

src/zarr/core/array.py
```diff
@@ -461,8 +461,11 @@ async def create(
             raise ValueError("Only one of chunk_shape or chunks can be provided.")
 
         if chunks:
+            chunks = parse_shapelike(chunks)
             _chunks = normalize_chunks(chunks, shape, dtype_parsed.itemsize)
         else:
+            if chunk_shape:
+                chunk_shape = parse_shapelike(chunk_shape)
```
jhamman, line 468: I suggest we move this check into `normalize_chunks`. Here's where the behavior you're looking to change is:

https://github.com/zarr-developers/zarr-python/blob/2fe12a7f424ac78c18a688d0a75049a3df67c98f/src/zarr/core/chunk_grids.py#L116-L117

If you look at `normalize_chunks`, you'll see why Zarr is more permissive on chunk shape than array shape. I'm not opposed to us disallowing floats but we need to be careful keep some of the other behaviors in `normalize_chunks`.

jhamman (changes requested):
Thanks @faymanns for working on this. I left one comment to direct your attention to `normalize_chunks`

src/zarr/core/array.py
faymanns: That seems like a better place indeed! I've updated the PR.

jhamman (approved):

PR #2537, opened 2024-12-06
[v3] implement / deprecate zarr.tree

jhamman:
We probably should have deprecated this in 2.18 but alas, here we are. This PR uses `Group.tree` for `zarr.tree` but adds a deprecation warning stating that we'll remove this after 3.0.

TODO:

### DIFF
--- a/src/zarr/api/asynchronous.py
+++ b/src/zarr/api/asynchronous.py
@@ -7,6 +7,7 @@ from typing import TYPE_CHECKING, Any, Literal, cast
 
 import numpy as np
 import numpy.typing as npt
+from typing_extensions import deprecated
 
 from zarr.core.array import Array, AsyncArray, get_array_metadata
 from zarr.core.buffer import NDArrayLike
@@ -493,6 +494,7 @@ async def save_group(
     await asyncio.gather(*aws)
 
 
+@deprecated("Use AsyncGroup.tree instead.")
 async def tree(grp: AsyncGroup, expand: bool | None = None, level: int | None = None) -> Any:
     """Provide a rich display of the hierarchy.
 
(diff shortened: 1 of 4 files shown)


src/zarr/api/asynchronous.py
```diff
@@ -493,8 +493,33 @@ async def save_group(
     await asyncio.gather(*aws)
 
 
-async def tree(*args: Any, **kwargs: Any) -> None:
-    raise NotImplementedError
+async def tree(grp: AsyncGroup, expand: bool | None = None, level: int | None = None) -> Any:
```
dstansby, line 496: It's probably worth decorating this instead of putting the warning inside, like here:

https://github.com/zarr-developers/zarr-python/blob/0cca6b061e1f98594b206c69ecfc9f1c8972b4ea/src/zarr/core/group.py#L1085-L1088

I think this way IDEs and type checkers will pick up the deprecation if you try and use the method.

src/zarr/api/synchronous.py
```diff
@@ -155,8 +157,9 @@ def save_group(
     )
 
 
-def tree(*args: Any, **kwargs: Any) -> None:
-    return sync(async_api.tree(*args, **kwargs))
+@deprecated("Use Group.tree instead.")
+def tree(grp: Group, expand: bool | None = None, level: int | None = None) -> Any:
+    return sync(async_api.tree(grp._async_group, expand=expand, level=level))
```
dstansby, line 163:
```suggestion
tree.__doc__ = async_api.tree.__doc__
```
I *think* this should fix the missing docstring at https://zarr--2537.org.readthedocs.build/en/2537/_autoapi/zarr/convenience/index.html#zarr.convenience.tree

jhamman: Hmmmm, we already have that! 

https://github.com/zarr-developers/zarr-python/blob/0cca6b061e1f98594b206c69ecfc9f1c8972b4ea/src/zarr/api/synchronous.py#L302

dstansby: Huh, odd

dstansby: Oh I think it might be becauase `autoapi` doesn't actualy import code to scrape docstrings. Anyway, not worth fixing in this PR

dstansby (approved):

Issue #2539, opened 2024-12-06
Clean up optional dependencies


dstansby:
On the current main branch, there are "extra" and "optional" optional dependecy groups. These names aren't very descriptive, so over in https://github.com/zarr-developers/zarr-python/pull/2538 I wasn't sure how to documente them. It would be good if we moved or renamed these dependencies. Does anyone have any suggestions?

```toml
[project.optional-dependencies]
...
extra = [
    'msgpack',
]
optional = [
    'universal-pathlib>=0.0.22',
    'rich'
]
...
```

d-v-b:
Do `msgpack`, `rich` or `universal-pathlib` interact in the codebase? I am guessing that they are pretty independent, which means that they probably shouldn't be grouped. Maybe one dependency group for each one is the way to address this?

dstansby:
It looks like:
- `msgpack` is only used in the docs (?), so we can remove it from "extra" (and remove "extra")
- `universal-pathlib` is used once, so we can move to a "universal-pathlib" extra
- `rich` is only used for tree, so we can get rid of it from "optional"

How does that sound?

PR #2541, opened 2024-12-06
Clean up optional dependency groups

dstansby:
Fixes https://github.com/zarr-developers/zarr-python/issues/2539.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -53,9 +53,25 @@ license = {text = "MIT License"}
 keywords = ["Python", "compressed", "ndimensional-arrays", "zarr"]
 
 [project.optional-dependencies]
+# User extras
 fsspec = [
     "fsspec>=2023.10.0",
 ]
+jupyter = [
+    'notebook',
+    'ipytree>=0.2.2',
+    'ipywidgets>=8.0.0',
+]
+gpu = [
+    "cupy-cuda12x",
+]
+universal-pathlib = [
+    'universal-pathlib>=0.0.22'
+]
+tree = [
+    'rich',
+]
+# Development extras
 test = [
     "coverage",
     "pytest",
@@ -68,15 +84,7 @@ test = [
     "hypothesis",
     "universal-pathlib",
 ]
-
-jupyter = [
-    'notebook',
-    'ipytree>=0.2.2',
-    'ipywidgets>=8.0.0',
-]
-gpu = [
-    "cupy-cuda12x",
-]
+optional = ["zarr[universal-pathlib, tree]"]
 docs = [
     'sphinx==8.1.3',
     'sphinx-autobuild>=2021.3.14',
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -53,9 +53,25 @@ license = {text = "MIT License"}
...
+]
+gpu = [
+    "cupy-cuda12x",
+]
+universal-pathlib = [
+    'universal-pathlib>=0.0.22'
+]
+tree = [
+    'rich',
+]
```
jhamman, line 73:
```suggestion
rich = [
    'rich',
]
```

pyproject.toml
```diff
@@ -68,15 +84,7 @@ test = [
...
-
-jupyter = [
-    'notebook',
-    'ipytree>=0.2.2',
-    'ipywidgets>=8.0.0',
-]
-gpu = [
-    "cupy-cuda12x",
-]
+optional = ["zarr[universal-pathlib, tree]"]
```
jhamman, line 87:
```suggestion
optional = ["zarr[universal-pathlib, rich]"]
```

jhamman (commented):
Thanks. Just one suggestion. 

For groups with only one optional dependency, I think its best if we just use the dependency as the name.

dstansby:
Is there any benefit in having a dependency group where the name is the same as the package? I'd suggest we either
- leave it as `tree`, to indicate "install this extra if you want to use `tree()`
-  remove it altogether and just document that folks need to install `rich` to use `tree()`

pyproject.toml
```diff
@@ -53,9 +53,25 @@ license = {text = "MIT License"}
...
 [project.optional-dependencies]
+# User extras
 fsspec = [
     "fsspec>=2023.10.0",
 ]
+jupyter = [
+    'notebook',
+    'ipytree>=0.2.2',
+    'ipywidgets>=8.0.0',
+]
```
jhamman, line 64: I believe these can be removed.

jhamman (commented):
> Is there any benefit in having a dependency group where the name is the same as the package? I'd suggest we either
> 
> * leave it as `tree`, to indicate "install this extra if you want to use `tree()`

If we do this, we should adopt the "feature-oriented groups" approach for the other groups as well. That would apply to fsspec for `RemoteStore`.

> * remove it altogether and just document that folks need to install `rich` to use `tree()`

I favor this option actually for `rich` and `universal-pathlib`.

pyproject.toml
dstansby: 👍 done

dstansby:
pre-commit.ci autofix

dstansby:
> If we do this, we should adopt the "feature-oriented groups" approach for the other groups as well

👍 done. I also removed universal-pathlib, tree, and started a v3 what's new (I don't think this already exists?) to communicate what's changed from v2.

jhamman (approved):

PR #2546, opened 2024-12-09
Fix json indent

will-moore:
This fixes the default json_indent added in #1952.
It was getting ignored in the `V3JsonEncoder` because `kwargs` has `"indent": None` so that `kwargs.pop("indent", config.get("json_indent"))` always returns `None`.

TODO:


src/zarr/core/metadata/v3.py
```diff
@@ -135,8 +135,8 @@ def parse_storage_transformers(data: object) -> tuple[dict[str, JSON], ...]:
 
 class V3JsonEncoder(json.JSONEncoder):
     def __init__(self, *args: Any, **kwargs: Any) -> None:
```
d-v-b, line 137: We should aim for something a bit more explicit than this function signature :sweat_smile: . I'm pretty sure a better signature here would have made this bug easier to catch. (not a blocker for this PR)

tests/test_metadata/test_v3.py
```diff
@@ -304,6 +304,19 @@ def test_metadata_to_dict(
...
+def test_json_indent():
+    m = GroupMetadata()
+    d = m.to_buffer_dict(default_buffer_prototype())["zarr.json"].to_bytes()
+    class TestIndentEncoder(json.JSONEncoder):
+        def __init__(self, *args: Any, **kwargs: Any) -> None:
+            super().__init__(*args, **kwargs)
+            self.indent = 2
+
+    # expected has extra ' ' on each line compared with json.dumps( indent=2)
+    expected = json.dumps(json.loads(d), cls=TestIndentEncoder).encode()
```
d-v-b, line 316: `json.dumps` takes an `indent` keyword argument, so I don't think we need the custom encoder class for this test

will-moore: Yeah, I tried that first!
But `json.dumps(data, indent=2)` gives a different output than a class derived from `json.JSONEncoder`. The `json.JSONEncoder` has an extra whitespace at the end of each line (see my comment on line 315 above).
I don't know why this is (I guess it would be nice to know)?!

d-v-b, line 316: Would it be possible to refactor this test so that it tests that the `json_indent` can be changed at runtime to produce different JSON-formatted strings? e.g., something like 
```python
@pytest.mark.parametrize('indent', (2,4))
def test_json_indent(indent: int):
    with config.set({"json_indent": indent}):
     #    ... write the metadata document to string 
    # compare with expected  
```

d-v-b: a separate test could check that the default `json_indent` value is actually 2

d-v-b: weird, i would also like to know why this is the case. is the extra whitespace more or less correct?

will-moore: Just confirmed this doing a test writing of some zarr data. All the `zarr.json` have whitespace at the end of rows that end with a comma.

will-moore: I think you have that in test_config.py https://github.com/zarr-developers/zarr-python/blob/ed0f199e8542b410ae655c8168a49a0be6237cd4/tests/test_config.py#L47 which you added in #1952 ?

d-v-b: quite right! no need for anything new here

will-moore: > weird, i would also like to know why this is the case. is the extra whitespace more or less correct?

I would say it's not really correct to have the extra whitespace at the end of lines.
Here, there shouldn't be whitespace after `"foo",`

```
>>> json.dumps({"test": "foo", "bar": 2}, cls=TestIndentEncoder)
'{\n  "test": "foo", \n  "bar": 2\n}'
```

src/zarr/core/metadata/v3.py
```diff
@@ -135,8 +135,8 @@ def parse_storage_transformers(data: object) -> tuple[dict[str, JSON], ...]:
...
-        self.indent = kwargs.pop("indent", config.get("json_indent"))
         super().__init__(*args, **kwargs)
+        self.indent = config.get("json_indent")
```
jhamman, line 139: What do you think about this instead?

```suggestion
        self.indent = kwargs.pop("indent") or config.get("json_indent")
```

alternatively, we could do this:

```python
if self.indent is None:
    self.indent = config.get("json_indent")

Comment (d-v-b) [reply src/zarr/core/metadata/v3.py:139]: the [signature](https://docs.python.org/3/library/json.html#json.JSONEncoder) for `JSONEncoder.__init__` isn't so bad, why don't we just use it explicitly?

Comment (d-v-b) [reply src/zarr/core/metadata/v3.py:139]: e.g., 
```python
class V3JsonEncoder(json.JSONEncoder):
    def __init__(self, *, skipkeys=False, ensure_ascii=True, check_circular=True, allow_nan=True, sort_keys=False, indent=None, separators=None, default=None) -> None
```

will-moore: Like this? [7a442e1](https://github.com/zarr-developers/zarr-python/pull/2546/commits/7a442e1b700e0d908732e21be89250ce48202560)
Strangely, that also fixes the strange whitespace problem, so the test is now much simpler!

will-moore:
There are 4 test failures just now:

```
FAILED tests/test_array.py::test_nbytes_stored - assert 502 == 366
FAILED tests/test_array.py::test_nbytes_stored_async - assert 502 == 366
FAILED tests/test_array.py::TestInfo::test_info_complete - AssertionError: assert Type
FAILED tests/test_array.py::TestInfo::test_info_complete_async - AssertionError: assert Type
```

I wouldn't expect those to be caused by this PR? Are they passing elsewhere?

jhamman:
Looking at the failing tests, I do think these are related to this PR:

```python
    async def test_nbytes_stored_async() -> None:
        arr = await zarr.api.asynchronous.create(shape=(100,), chunks=(10,), dtype="i4")
        result = await arr.nbytes_stored()
>       assert result == 366  # the size of the metadata document. This is a fragile test.
E       assert 502 == 366
```

So it seems the change to the JSON encoding is impacting the size of the metadata document. I agree with the comment in the test "this is a fragile test". I think we should be open to changing this test or set the json encoding for these specific tests.

d-v-b:
I think the info test should not be hard-coding the metadata size but rather comparing the reported metadata size to the actual metadata size, that would make the test much less brittle.

d-v-b:
similarly, `test_nbytes_stored` should test that the actual size of the metadata document (measured directly) matches what comes out of `nbytes_stored()`, instead of hard-coding things. I can clean these up in a PR unless someone beats me to it.

will-moore:
@d-v-b I wasn't sure how to implement your suggestion from https://github.com/zarr-developers/zarr-python/pull/2546#issuecomment-2539860395 - Maybe you could give some more pointers?
But for now I've updated the byte counts so that the tests are green (but still fragile as before).

d-v-b:
> @d-v-b I wasn't sure how to implement your suggestion from [#2546 (comment)](https://github.com/zarr-developers/zarr-python/pull/2546#issuecomment-2539860395) - Maybe you could give some more pointers? But for now I've updated the byte counts so that the tests are green (but still fragile as before).

I was thinking that instead of comparing the reported size of the metadata document to a hard-coded value, we could just get the size of the metadata document directly, and ensure that the reported size matches the observed size. We would get the actual size of the metadata document by invoking `store.get(zarr.json)` (for v3) and measuring the length of the bytes coming out. A similar process would give you the size of the metadata document for v2. But this can be done in a separate PR.

will-moore:
@d-v-b Anything else I need to do here?

dstansby (approved):

dstansby:
I fixed the docstests in the latest commit - 🤞 this works, and thanks for the PR!

will-moore:
Great, thanks @dstansby

PR #2549, opened 2024-12-10
docs/add docstrings to synchronous API

### LINKED ISSUE #2475: Comprehensive docstrings for all top-level exports
Very few of the [top-level functions](https://github.com/zarr-developers/zarr-python/blob/498cb788b29900183dbcfd7633c48afd5cee8a95/src/zarr/__init__.py#L37-L67) have doctrings. This is bad and we should fix it, either by writing good docstrings from scratch or dynamically copying docstrings from other functions where appropriate.

d-v-b:
I add docstrings to the exported functions defined in `api/synchronous.py`, copy exactly the same functions in `api/asynchronous.py`, except for a few places where the sync / async functions diverge. 

I also add a test to check that docstrings are identical, and another test to check that function signatures are identical, for said paired functions. These tests also get some exceptions for functions that are not quite identical.

closes #2475 

TODO:


d-v-b:
note that some lower-level routines needed their function signatures updated -- it seems like `Array.create` wasn't fully aware that it takes `int` for `chunks` / `chunk_shape`

dstansby (commented):
Nice, and great that it caught some dodgy signatures. I left a couple of suggestions/questions

tests/test_api.py
```diff
@@ -1055,3 +1056,38 @@ def test_open_array_with_mode_r_plus(store: Store) -> None:
...
+    Test that functions in the API module have signatures that are derived from those in the asynchronous API module
+    """
+    if func_name in ("tree", "load", "open", "open_array", "group"):
+        msg = (
+            f"{func_name} gets xfailed because the async version and the sync "
+            "version have meaningful differences in their signatures."
+        )
+        pytest.xfail(reason=msg)
+
+    sync_sig_params = inspect.signature(getattr(zarr.api.synchronous, func_name)).parameters
```
dstansby, line 1091: Is it worth checking [`Signature.return_annotation`](https://docs.python.org/3/library/inspect.html#inspect.Signature.return_annotation) here too?

src/zarr/api/asynchronous.py
```diff
@@ -765,9 +765,9 @@ async def open_group(
 
 
 async def create(
-    shape: ChunkCoords,
+    shape: ChunkCoords | int,
```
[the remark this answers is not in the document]
d-v-b: tbh I don't like `ChunkCoords` (that type is a tuple of ints, and it's valid for more than chunk coordinates, so the name is inaccurate) and I think we should stick with raw `tuple[int, ...]` wherever possible, but I think we already have `ShapeLike` which is `ChunkCoords | int`. I can drop that in here.

tests/test_api.py
```diff
@@ -1055,3 +1056,38 @@ def test_open_array_with_mode_r_plus(store: Store) -> None:
...
+        == getattr(zarr.api.asynchronous, func_name).__doc__
+    )
+
+
+@pytest.mark.parametrize("func_name", zarr.api.synchronous.__all__)
+def test_derived_signatures(func_name: str) -> None:
+    """
+    Test that functions in the API module have signatures that are derived from those in the asynchronous API module
+    """
+    if func_name in ("tree", "load", "open", "open_array", "group"):
```
dstansby, line 1084: If the signature of two functions is different, shouldn't the docstring be different too. ie., this list and the same list above in `test_derived_docstrings` should be the same?

In at least `tree()` this seems to be correct - the docstring for `zarr.api.asynchronous.tree` needs correcting to say it takes an `AsyncGroup` as input, at which point it will have a different docstring to `zarr.api.synchronous.tree`

d-v-b: yes i think you are right -- by construction _all_ of the docstrings should be different in their return type annotation, and so this test is currently incorrect (fortunately I guarded for something like this outcome in the test docstring by stating "derived from" rather than "equal to")

src/zarr/api/synchronous.py
```diff
@@ -44,8 +51,38 @@
...
+
+        By default, the root node is used so all the metadata in the
+        store is consolidated.
+    zarr_format : {2, 3, None}, optional
+        The zarr format of the hierarchy. By default the zarr format
+        is inferred.
+
+    Returns
+    -------
+    group: AsyncGroup
```
jhamman, line 81:
```suggestion
    group: Group
```

src/zarr/api/asynchronous.py
```diff
@@ -765,9 +765,9 @@ async def open_group(
...
     *,  # Note: this is a change from v2
-    chunks: ChunkCoords | None = None,  # TODO: v2 allowed chunks=True
+    chunks: ChunkCoords | int | None = None,  # TODO: v2 allowed chunks=True
```
jhamman, line 770: we must not have test coverage here but you could also throw in `bool` and remove this TODO.

src/zarr/api/synchronous.py
```diff
@@ -74,15 +141,44 @@ def open(
...
+    zarr_format : {2, 3, None}, optional
+        The zarr format to use when saving.
+    path : str or None, optional
+        The path within the store to open.
+    storage_options : dict
+        If using an fsspec URL to create the store, these will be passed to
+        the backend implementation. Ignored otherwise.
+    **kwargs
+        Additional parameters are passed through to :func:`zarr.creation.open_array` or
+        :func:`zarr.hierarchy.open_group`.
```
jhamman, line 167:
```suggestion
        Additional parameters are passed through to :func:`zarr.api.asynchronous.open_array` or
        :func:`zarr.api.asynchronous.open_group`.
```

src/zarr/api/synchronous.py
```diff
@@ -159,28 +314,98 @@ def save_group(
...
+    Parameters
+    ----------
+    data : array_like
+        The data to fill the array with.
+    **kwargs
+        Passed through to :func:`create`.
+
+    Returns
+    -------
+    array : array
```
jhamman, line 353:
```suggestion
    array : Array
```

src/zarr/api/synchronous.py
```diff
@@ -159,28 +314,98 @@ def save_group(
...
+        to users. Use `numpy.empty(())` by default.
+    zarr_format : {2, 3, None}, optional
+        The zarr format to use when saving.
+    storage_options : dict
+        If using an fsspec URL to create the store, these will be passed to
+        the backend implementation. Ignored otherwise.
+
+    Returns
+    -------
+    g : group
```
jhamman, line 406:
```suggestion
    g : Group
```

src/zarr/api/synchronous.py
```diff
@@ -215,6 +441,67 @@ def open_group(
...
+        To explicitly *not* use consolidated metadata, set ``use_consolidated=False``,
+        which will fall back to using the regular, non consolidated metadata.
+
+        Zarr v2 allowed configuring the key storing the consolidated metadata
+        (``.zmetadata`` by default). Specify the custom key as ``use_consolidated``
+        to load consolidated metadata from a non-default key.
+
+    Returns
+    -------
+    g : group
```
jhamman, line 502:
```suggestion
    g : Group
```

src/zarr/api/synchronous.py
```diff
@@ -236,84 +523,377 @@ def open_group(
...
+        to users. Use `numpy.empty(())` by default.
+
+        .. versionadded:: 2.13
+    storage_options : dict
+        If using an fsspec URL to create the store, these will be passed to
+        the backend implementation. Ignored otherwise.
+
+    Returns
+    -------
+    z : array
```
jhamman, line 639:
```suggestion
    z : Array
```

src/zarr/api/synchronous.py
```diff
@@ -236,84 +523,377 @@ def open_group(
...
 # TODO: add type annotations for kwargs
 def empty(shape: ChunkCoords, **kwargs: Any) -> Array:
+    """Create an empty array.
+
+    Parameters
+    ----------
+    shape : int or tuple of int
+        Shape of the empty array.
+    **kwargs
+        Keyword arguments passed to :func:`zarr.api.asynchronous.create`.
```
jhamman, line 687:
```suggestion
        Keyword arguments passed to :func:`zarr.api.asynchronous.create`.
        
    Returns
    -------
    Array
        The new array.
```

jhamman (approved):
:clap:

dstansby (changes requested):
Marking as requested changes so we don't forget the test needs fixing (see https://github.com/zarr-developers/zarr-python/pull/2549#discussion_r1883625220)

d-v-b:
I removed the docstring / function signature tests, because I think they are too complex to implement properly right now. I think this is basically ready.

dstansby (approved):

PR #2552, opened 2024-12-11
Un-pin numpy

dstansby:
I pinned numpy in a previous PR because it was breaking tests 😭 . This un-pins it to see what's going wrong, and then hopefully fix it.


d-v-b:
numpy 2.2 changes the string repr for arrays (shape info is now displayed), maybe that's the culprit?

dstansby:
From memory that's exactly it

dcherian:
If you have the env, then running [`pytest-accept`](https://github.com/max-sixty/pytest-accept) should quickly fix it

dstansby:
Ah amazing, thanks for sharing! https://github.com/scientific-python/pytest-doctestplus does a similar thing, but I'll checkout pytest-accept too

dcherian (approved):

PR #2556, opened 2024-12-13
add warnings when using non-spec features with v3

### LINKED ISSUE #2411: [v3] It is possible to create v3 data with features not currently written into the spec (dtypes, codecs, etc.)
### Zarr version

3.0.0.beta

### Numcodecs version

0.13

### Python Version

3.11

### Operating System

Mac

### Installation

pip

### Description

We have discussed adding a warning when writing data to `zarr_format=3` that is outside the current spec. Currently, this mostly applies to data types and codecs but also includes stores (e.g. `ZipStore`).

normanrz:
Adds warnings when using the following non-spec features with v3 groups and arrays:

- vlen string and bytes dtypes
- vlen-utf8 and vlen-bytes codecs
- consolidated metadata

Fixes #2411


normanrz:
The warnings are currently `xx is currently not part in the Zarr version 3 specification and may not be supported by other zarr implementations and may change in the future.` I'm open to suggestions how to change the wording.

src/zarr/api/asynchronous.py
```diff
@@ -195,6 +195,14 @@ async def consolidate_metadata(
             v = dataclasses.replace(v, consolidated_metadata=ConsolidatedMetadata(metadata={}))
             members_metadata[k] = v
 
+    if any(m.zarr_format == 3 for m in members_metadata.values()):
+        warnings.warn(
+            "Consolidated metadata is currently not part in the Zarr version 3 specification and "
+            "may not be supported by other zarr implementations and may change in the future.",
```
jhamman, line 201:
```suggestion
            "Consolidated metadata is currently not part in the Zarr version 3 specification. It "
            "may not be supported by other zarr implementations and may change in the future.",
```

jhamman (approved):
Thanks @normanrz - I like it! I left one small comment on the wording of the warning.

PR #2558, opened 2024-12-14
Raise an exception for previously set key

maxrjones:
I think this tests was not correctly checking the behavior because the data at `k` was previously set, so calling `set_if_not_exists` on that same key should actually raise an exception. The behavior for obstore in https://github.com/zarr-developers/zarr-python/pull/1661 and https://github.com/kylebarron/zarr-python/pull/3 seem correct, in contrast to the fsspec-based stores which now fail on this test.

### DIFF
--- a/src/zarr/testing/store.py
+++ b/src/zarr/testing/store.py
@@ -319,8 +319,9 @@ class StoreTests(Generic[S, B]):
         data_buf = self.buffer_cls.from_bytes(b"0000")
         await self.set(store, key, data_buf)
 
-        new = self.buffer_cls.from_bytes(b"1111")
-        await store.set_if_not_exists("k", new)  # no error
+        with pytest.raises(Exception):
+            new = self.buffer_cls.from_bytes(b"1111")
+            await store.set_if_not_exists("k", new)
 
         result = await store.get(key, default_buffer_prototype())
         assert result == data_buf
--- a/src/zarr/testing/store.py
+++ b/src/zarr/testing/store.py
@@ -321,7 +321,7 @@ class StoreTests(Generic[S, B]):
 
         with pytest.raises(Exception):
             new = self.buffer_cls.from_bytes(b"1111")
-            await store.set_if_not_exists("k", new)
+            await store.set_if_not_exists(key, new)
 
         result = await store.get(key, default_buffer_prototype())
         assert result == data_buf


src/zarr/testing/store.py
```diff
@@ -319,8 +319,9 @@ async def test_set_if_not_exists(self, store: S) -> None:
         data_buf = self.buffer_cls.from_bytes(b"0000")
         await self.set(store, key, data_buf)
 
-        new = self.buffer_cls.from_bytes(b"1111")
-        await store.set_if_not_exists("k", new)  # no error
+        with pytest.raises(Exception):
+            new = self.buffer_cls.from_bytes(b"1111")
+            await store.set_if_not_exists("k", new)
```
d-v-b, line 324:
```suggestion
            await store.set_if_not_exists(key, new)
```

src/zarr/testing/store.py
```diff
@@ -319,8 +319,9 @@ async def test_set_if_not_exists(self, store: S) -> None:
         data_buf = self.buffer_cls.from_bytes(b"0000")
         await self.set(store, key, data_buf)
 
-        new = self.buffer_cls.from_bytes(b"1111")
-        await store.set_if_not_exists("k", new)  # no error
+        with pytest.raises(Exception):
```
d-v-b, line 322: can we be more specific about the exception? Ideally we would match the type and the error message

maxrjones:
@d-v-b when looking into what exceptions different fsspec interfaces raise, I noticed that [fsspec's internal code comments](https://github.com/fsspec/gcsfs/blob/5c2b70c7859ae9beca6e4a08e1f0fb90eaf09899/gcsfs/core.py#L225-L227) and [the docs](https://github.com/fsspec/filesystem_spec/blob/c36066ce002fbc45efecab87bd260f6562e75bbf/docs/source/features.rst#exclusive-write) lists exclusive write as experimental and risky. Should we warn about this in the Zarr docs?

d-v-b:
@martindurant what do you think?

src/zarr/testing/store.py
maxrjones: I'm not quite sure how to manage matching the type with optional dependencies. Obstore raises its `AlreadyExistsError` and fsspec raises its `FileExistsError`. One option is to put this as `NotImplemented` in the base test class and require the individual stores to implement it in their tests, but that seems non-ideal. Do you have advice?

Matching the error message would be easier because we wouldn't need to import those errors.

d-v-b: if it's not simple to get more specific than `Exception` than I think it's OK to leave it as-is, and i can open an issue about how we could make this particular test more specific 😄

jhamman:
> so calling set_if_not_exists on that same key should actually raise an exception.

I don't know if this is the behavior we want actually. The application in Zarr that relies on this method is to set an empty group metadata object if it doesn't exist yet. If it does, the desired behavior is to move on.

One way to move forward is to catch this error in the Obstore Store and then move on.

maxrjones:
> > so calling set_if_not_exists on that same key should actually raise an exception.
> 
> I don't know if this is the behavior we want actually. The application in Zarr that relies on this method is to set an empty group metadata object if it doesn't exist yet. If it does, the desired behavior is to move on.
> 
> One way to move forward is to catch this error in the Obstore Store and then move on.

Thanks Joe, that makes sense now. I'll follow your recommendation to suggest a change within ObjectStore instead

PR #2568, opened 2024-12-18
Doc updates for 3.0 release

jhamman:
This PR includes a fairly large reorganization of the Zarr-Python documentation in preparation for the 3.0 release. Major changes include:

- new top level `quickstart` page
- break up `tutorial.rst`. See `user-guide/{parts}.rst`
  - use ipython directives to run examples at doc build time
  - new user guide sections for advanced topics (async, extensions, etc.)
- add new site section for developers (contributors guide, roadmap, etc.)

#### TODOs
- [x] Get feedback and decide on what to do with the docs in `todo.rst`. These are mostly features that have not been ported to v3.
- [x] I would like to migrate the content from #2102 into `user-guide/whatsnew_v3.rst`.

Closes: #1765, #1767, #1770, #1798, #2550

PR #2570, opened 2024-12-18
Remove license page from docs

dstansby:
I don't usually see licenses in other docs, so I think worth removing from ours? Also part of https://github.com/zarr-developers/zarr-python/pull/2568, but I think worth pulling out for ease of review.


d-v-b (approved):

dstansby:
I relalised it would be a good idea to redirect this if we're deleting it, so redirect added

jhamman (approved):

PR #2576, opened 2024-12-19
correct array.nbytes, and add tests

### LINKED ISSUE #2575: Array.nbytes logic is wrong
### Zarr version

3.0.0.b2

### Numcodecs version

!

### Python Version

!

### Operating System

!

### Installation

!

### Description

This logic is obviously and embarrassingly wrong:

https://github.com/zarr-developers/zarr-python/blob/f035d453894ec83234f69f7627b540390ee2b6eb/src/zarr/core/array.py#L977-L982

### Steps to reproduce

Read the code

### Additional output

_No response_

d-v-b:
`AsyncArray.nbytes` is very wrong, this PR corrects it (and adds tests)

closes #2575 

TODO:

### DIFF
--- a/src/zarr/core/array.py
+++ b/src/zarr/core/array.py
@@ -980,7 +980,7 @@ class AsyncArray(Generic[T_ArrayMetadata]):
         The number of bytes that can be stored in the chunks of this array.
         """
         # TODO: how can this be meaningful for variable-length types?
-        return int(np.prod(self.shape) * self.dtype.itemsize)
+        return self.size * self.dtype.itemsize
 
     async def _get_selection(
         self,
@@ -1430,7 +1430,7 @@ class AsyncArray(Generic[T_ArrayMetadata]):
             _order=self.order,
             _read_only=self.read_only,
             _store_type=type(self.store_path.store).__name__,
-            _count_bytes=self.dtype.itemsize * self.size,
+            _count_bytes=self.nbytes,
             _count_bytes_stored=count_bytes_stored,
             _count_chunks_initialized=count_chunks_initialized,
             **kwargs,
--- a/src/zarr/core/array.py
+++ b/src/zarr/core/array.py
@@ -977,9 +977,16 @@ class AsyncArray(Generic[T_ArrayMetadata]):
     @property
     def nbytes(self) -> int:
         """
-        The number of bytes that can be stored in the chunks of this array.
+        The total number of bytes that can be stored in the chunks of this array.
+
+        Notes
+        -----
+        This value is calculated by multiplying the number of elements in the array and the size
+        of each element, the latter of which is determined by the dtype of the array.
+        For this reason, ``nbytes`` will likely be inaccurate for arrays with variable-length
+        dtypes. It is not possible to determine the size of an array with variable-length elements
+        from the shape and dtype alone.
         """
-        # TODO: how can this be meaningful for variable-length types?
         return self.size * self.dtype.itemsize
 
     async def _get_selection(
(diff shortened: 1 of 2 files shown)


src/zarr/core/array.py
```diff
@@ -977,9 +977,10 @@ def _iter_chunk_regions(
     @property
     def nbytes(self) -> int:
         """
-        The number of bytes that can be stored in this array.
+        The number of bytes that can be stored in the chunks of this array.
         """
-        return self.nchunks * self.dtype.itemsize
+        # TODO: how can this be meaningful for variable-length types?
+        return int(np.prod(self.shape) * self.dtype.itemsize)
```
rabernat, line 983: I'm wondering if it would be worth using the exact same logic we use in `info`:

https://github.com/zarr-developers/zarr-python/blob/5bf7bcfcfe6426bea37df5d0691211f6a71daa39/src/zarr/core/array.py#L1432

rabernat (approved):
One minor suggestion to be self consistent in our implementation.

We should reuse this function within info (where it is currently correct) rather than duplicating the same logic twice.

src/zarr/core/array.py
d-v-b: we should definitely only calculate this value in one place. I will look into changing `nbytes` to use the simpler formula above, and then `info` will use `nbytes`

src/zarr/core/array.py
```diff
@@ -977,9 +977,10 @@ def _iter_chunk_regions(
     @property
     def nbytes(self) -> int:
         """
-        The number of bytes that can be stored in this array.
+        The number of bytes that can be stored in the chunks of this array.
```
dstansby, line 980: Worth adding a note or warning that this only gives the correct answer for fixed-length types? (if I'm interpreting your TODO correctly?)

tests/test_array.py
```diff
@@ -776,3 +776,21 @@ async def test_special_complex_fill_values_roundtrip(fill_value: Any, expected:
...
+    shape: tuple[int, ...], dtype: str, array_type: Literal["async", "sync"]
+) -> None:
+    """
+    Test that the ``nbytes`` attribute of an Array or AsyncArray correctly reports the capacity of
+    the chunks of that array.
+    """
+    store = MemoryStore()
+    arr = Array.create(store=store, shape=shape, dtype=dtype, fill_value=0)
+    if array_type == "async":
+        assert arr._async_array.nbytes == np.prod(arr.shape) * arr.dtype.itemsize
```
dstansby, line 794: Since this is kind of duplicating the logic in the actual function, is it worth hard coding some expected values in the test (instead of calculating them from the array itself)?

d-v-b: if we take the definition of `nbytes` to be `size of each element * number of elements`, then I think it's OK to generate the expected values via that calculation  -- we would do that same calculation anyway to generate our expected values. Also, I personally find magic values in tests to be a bit of code smell, because it's not clear where they came from.

src/zarr/core/array.py
d-v-b: I could also just fix the implementation ... what kind of behavior would people expect when calling `nbytes` on an array with variable length types? maybe raise an exception?

d-v-b: second question, is there any reliable way to check if a dtype is variable length, other than a big lookup table?

cc @rabernat

d-v-b: i'm feeling like this out of scope for this PR; numpy itself reports a (misleading) integer itemsize for variable length dtypes already, so we can lean on that bad precedent until we come up with a proper fix

dstansby: 👍 - can you add a quick note to the docstring here though noting that it's incorrect for variable length dtypes?

d-v-b: have a look at the docstrings I added in https://github.com/zarr-developers/zarr-python/pull/2576/commits/254e2764c1b9d64cc906e1cd9cb1dd62762a4a33

dstansby (approved):

Issue #2437, opened 2024-10-24
Make the `ByteRangeRequest` more literate


d-v-b:
Continuing a conversation from #1661 

We have IO operations that can fetch ranges of bytes. That byte range parameter is parameterized by an (optional) 2-tuple of nullable ints, i.e. `tuple[int | None, int | None] | None`. The tuple part is ambiguous between two similar specifications of a range of integers -- `(start, step)` and `(start, stop)`; this ambiguity is bad.

We should rework this type so that it's harder to be confused about these two possibilities. One option suggested by @kylebarron would be a dataclass, e.g. :

```python
@dataclasses.dataclass
class ByteRangeRequest:
  start: int | None = 0
  step: int | None = None
```

It would also be nice if there was an instance of this class (e.g., the default) that conveyed "fetch the whole thing", and then `byte_range` wouldn't need to be optional in all the store methods. I'm also not sure we get much from `None` in this type, since there are integers that can express "the end" of a range, e.g. `-1`. Curious to hear other ideas here.

paraseba:
This would be a good opportunity to think if we should also consider a type that allows doing things like "last N bytes" (without knowing the size beforehand). Most (all?) object stores, for example, allow that type of query.

d-v-b:
> This would be a good opportunity to think if we should also consider a type that allows doing things like "last N bytes" (without knowing the size beforehand). Most (all?) object stores, for example, allow that type of query.

Today the elements of `ByteRangeRequest` can be negative, so I think `(-1, -100)` would express "read the the last 100 bytes". Maybe @normanrz can correct me if this is wrong.

kylebarron:
@paraseba Suffix requests are already used (https://github.com/zarr-developers/zarr-python/pull/1661#issuecomment-2435452928), we just want to ensure the type is well defined so that semantics are clear.

> Most (all?) object stores, for example, allow that type of query.

All except Azure.

> Today the elements of `ByteRangeRequest` can be negative, so I think `(-1, -100)` would express "read the the last 100 bytes".

From here it's `(-100, None)` that would express the last 100 bytes:
https://github.com/zarr-developers/zarr-python/blob/bc588a760a804f783c4242d4435863a43a5f3f9f/src/zarr/codecs/sharding.py#L700-L702

normanrz:
Yes, it is `(-100, None)`

paraseba:
That is great, then making this type more clear is even more important.

normanrz:
[...]

Issue #2412, opened 2024-10-18
Tracking 3.0 Release Blockers


jhamman:
We are now in the last mile of the 3.0 refactor. This issue is tracking what we are considering release blockers: There is also a milestone for [After 3.0.0](https://github.com/zarr-developers/zarr-python/milestone/17) where we are putting tickets that can come later.

### Docs
- [x] https://github.com/zarr-developers/zarr-python/issues/1765
- [x] https://github.com/zarr-developers/zarr-python/issues/1766
- [x] https://github.com/zarr-developers/zarr-python/issues/1767
- [x] https://github.com/zarr-developers/zarr-python/issues/1770
- [x] https://github.com/zarr-developers/zarr-python/issues/2245
- [x] https://github.com/zarr-developers/zarr-python/issues/2246
- [x] https://github.com/zarr-developers/zarr-python/issues/2247
- [x] https://github.com/zarr-developers/zarr-python/issues/2248
- [x] https://github.com/zarr-developers/zarr-python/issues/2252
- [x] https://github.com/zarr-developers/zarr-python/issues/2253

### Array
- [x] https://github.com/zarr-developers/zarr-python/issues/1750
- [x] https://github.com/zarr-developers/zarr-python/issues/1977
- [x] https://github.com/zarr-developers/zarr-python/issues/2029
- [x] https://github.com/zarr-developers/zarr-python/issues/2267
- [x] https://github.com/zarr-developers/zarr-python/issues/2325
- [x] https://github.com/zarr-developers/zarr-python/issues/2404
- [x] https://github.com/zarr-developers/zarr-python/issues/2409

### Group
- [x] https://github.com/zarr-developers/zarr-python/issues/2135
- [x] https://github.com/zarr-developers/zarr-python/issues/2402
- [x] https://github.com/zarr-developers/zarr-python/issues/2406
- [x] https://github.com/zarr-developers/zarr-python/issues/2527

### Top level API
- [x] https://github.com/zarr-developers/zarr-python/issues/1943
- [x] https://github.com/zarr-developers/zarr-python/issues/2403
- [x] https://github.com/zarr-developers/zarr-python/issues/2408
- [x] https://github.com/zarr-developers/zarr-python/issues/2411

### Store API
- [x] https://github.com/zarr-developers/zarr-python/issues/1755
- [x] https://github.com/zarr-developers/zarr-python/issues/2191
- [x] https://github.com/zarr-developers/zarr-python/issues/2342
[...]

PR #2585, opened 2024-12-22
Use dataclasses for ByteRangeRequests

maxrjones:
This draft PR modifies the ByteRangeRequest type to follow the same approach as `obstore`, as suggested [here](https://github.com/zarr-developers/zarr-python/issues/2437#issuecomment-2452528308). It is a larger change than the decision made in https://github.com/zarr-developers/zarr-python/pull/1900#discussion_r1610449768 to specify `ByteRangeRequest` tuples as `start, stop` rather than `start, length` which has not yet been implemented in `main`.

As for performance impacts, please see these simple timings - https://github.com/maxrjones/zarr-byterangerequest-microbenchmarks/blob/main/plot-results.ipynb


normanrz:
I added the changes for sharding in https://github.com/zarr-developers/zarr-python/pull/2585/commits/4a73b708372a6c2ab98f84926ccaa706ef8d131d

maxrjones:
> Why not use literate instead of a tuple?
> 
> ```python
> class ExplicitRange(TypedDict):
>     start: int
>     end: int
> ```

My understanding from https://github.com/zarr-developers/zarr-python/issues/2437#issuecomment-2452695134 was that there were concerns over performance for dataclasses, which I thought could extend to TypedDict over tuple. If you'd rather literate instead of a tuple, I could make that change and run some simple tests to check performance.

normanrz:
> If you'd rather literate instead of a tuple, I could make that change and run some simple tests to check performance.

Yeah. I think that would be great. If the performance is ok, I would prefer a dict (or dataclass).
Btw, have you done some performance testing how `@dataclass(slots=True)` compares to a dict?

maxrjones:
> > If you'd rather literate instead of a tuple, I could make that change and run some simple tests to check performance.
> 
> Yeah. I think that would be great. If the performance is ok, I would prefer a dict (or dataclass). Btw, have you done some performance testing how `@dataclass(slots=True)` or `NamedTuple` compares to a dict?

Thanks @normanrz, I haven't done those tests but can try that out along with your earlier suggestion. For timing, do you think this PR would be considered for the V3 release? I could prioritize this if so, but if not I won't rush to work on it.

normanrz:
We are closing 3.0 right now, so I guess this will be a post-3.0 issue. So, no rush here.

maxrjones:
> We are closing 3.0 right now, so I guess this will be a post-3.0 issue. So, no rush here.

Sounds good, I was pushing for pre v3.0 in https://github.com/zarr-developers/zarr-python/issues/2412#issuecomment-2555580955 under the assumption that using `(start, end)` or TypedDict/NamedTuple/Dataclass instead would not be accepted afterwards because it is a breaking change. But I am happy to wait you are open to later changes in behavior.

normanrz:
My thinking is that this is not a big enough blocker for the 3.0 release. We can change this afterwards, but we will probably need a backwards-compat layer.
@jhamman what do you think here?

jhamman:
I agree. This can come after 3.0.0. But the sooner the better. The interface isn't really used by anyone external except for icechunk so if we do this quickly, we can probably get by without a compat layer.

src/zarr/storage/_fsspec.py
```diff
@@ -274,21 +276,34 @@ async def get_partial_values(
...
+                    if "offset" in byte_range:
+                        starts.append(byte_range["offset"])  # type: ignore[typeddict-item]
+                        stops.append(None)
+                    elif "suffix" in byte_range:
+                        starts.append(-byte_range["suffix"])
+                        stops.append(None)
+                    else:
+                        raise ValueError("Invalid format for ByteRangeRequest")
+                else:
+                    raise ValueError("Invalid format for ByteRangeRequest")
```
maxrjones, line 302: this is where the PR could most likely cause a performance hit by iterating over the inputs rather than using zip.

src/zarr/abc/store.py
```diff
@@ -19,7 +19,22 @@
...
+
+
+class SuffixRange(TypedDict):
+    """Request up to the last `n` bytes"""
+
+    suffix: int
+    """The number of bytes from the suffix to request."""
+
+
+ByteRangeRequest: TypeAlias = None | tuple[int, int] | OffsetRange | SuffixRange
```
maxrjones, line 37: Is there a likely future case where `ByteRangeRequest = None` would not be supported, such that we should exclude `None` from the TypeAlias options?

kylebarron: What does `None` mean? Can the implication of each type be documented in a docstring on this type?

normanrz: I think `None` would mean that the entire object is requested.

kylebarron: IMO it's slightly clearer to have `None` outside this type alias. I like how it was before https://github.com/zarr-developers/zarr-python/blob/12f601258d7af950a853a9e7fbbdc32feae73901/src/zarr/abc/store.py#L144
where `ByteRangeRequest` defines a _valid_ byte range request, and then the type is used as `ByteRangeRequest | None`.

maxrjones: Good call thanks

maxrjones:
@normanrz @jhamman I didn't want to spam the zarr-python repo with PRs so the requested alternative implementations for [dataclasses](https://github.com/maxrjones/zarr-python/pull/2/files) and [typeddicts with ExplicitRangeRequest](https://github.com/maxrjones/zarr-python/pull/3/files) are opened as PRs to my fork. I didn't try out NamedTuple because it seems potentially confusing to have multiple NamedTuples in a single type alias, but would be glad to hear suggestions around that.

Please see https://github.com/maxrjones/zarr-byterangerequest-microbenchmarks/blob/main/plot-results.ipynb for microbenchmark results. Here's a summary for a range request:
- No significant difference between this PR and main (only tuples)
- Dataclasses are significantly different from only tuples (average ~0.007s slower)
- TypedDicts with ExplicitRangeRequest is significantly different from only tupes (average ~0.011s slower)

I'd like to expand this to include other stores (Memory, Zip, and Fsspec) and operations (get_partial_values, suffix request, offset request) but wanted to share this update before heading out for the evening. At any point, please feel welcome to take over if I'm holding you back.

kylebarron:
> * No significant difference between this PR (tuple or TypedDict) and main (only tuples)

Presumably that's because the vast majority of requests still use the `tuple` code path?

> * Dataclasses are significantly different from only tuples (average ~0.007s slower)

Is that 0.007s slower _per request_? That's not _huge_, but an extra 7s per 1000 requests isn't nothing either.

maxrjones:
> > * No significant difference between this PR (tuple or TypedDict) and main (only tuples)
> 
> Presumably that's because the vast majority of requests still use the `tuple` code path?
> 

Yes these tests are just for an explicit range which uses the same design as obstore (tuples with `(start, end)`).

> > * Dataclasses are significantly different from only tuples (average ~0.007s slower)
> 
> Is that 0.007s slower _per request_? That's not _huge_, but an extra 7s per 1000 requests isn't nothing either.

Good question and apologies for not specifying this up-front. The [tests use `timeit.timeit`](https://github.com/maxrjones/zarr-byterangerequest-microbenchmarks/blob/df9c74cda057be2ffd4cdf06c47099fb63a95de2/time_get_range_startlength.py#L10) which [returns the time it takes to execute the main statement a number of times](https://docs.python.org/3/library/timeit.html#timeit.Timer.timeit).  In this case, it is on average 0.007s slower to execute `get` on a byte range 10000 times when using dataclasses relative to using tuples.

jhamman:
My read is that the performance implications are likely to be negligible relative to the network noise for most stores -- so I'm not concerned by what I see here. When we're to the point where squeezing 0.007s out of chunk requests seems necessary, we'll be in a very good place!

@normanrz - are you happy with what you see here?

d-v-b:
thanks for this demo

> My opinion is that a tuple representing `(start, end)` with end exclusive and dataclasses for suffix and offset requests provides the nicest balance of code readability while not requiring importing a dataclass for the simplest form of range requests. To help people form opinions on the options, here's a quick summary of how the user facing API works for each of these options:

👍 I agree. although I do find the typed dicts rather cute for some reason I can imagine they could end up rather tedious to write in aggregate.

normanrz:
I also think the performance is ok.

My problem with tuples is that it is not clear what the second component is. Without further docs, it could be either the exclusive end or a byte length. That is why I like the dataclass (with kwargs: `values = asyncio.run(store.get('c/0', byte_range=ExplicitRange(start=1, end=5)))`) or typed dict approach better.

Have we considered using a `slice` (with disallowed `step != 1`)? It has clear semantics for start and end and supports suffix and offset.

d-v-b:
i think `slice` is a bit too expressive, and it doesn't play nicely with type checking, so we would need to examine each one at runtime to ensure that its a valid specification of a dense interval.

normanrz:
We could really use rust-style enums here :D

normanrz:
Well, after 5min of googling I learned about the `match` statement (new in Python 3.10), which kind of enables rust-style enums.

```python
@dataclass
class ExplicitByteRequest:
    start: int
    end: int

@dataclass
class SuffixByteRequest:
    suffix: int

@dataclass
class OffsetByteRequest:
    offset: int

ByteRequest = ExplicitByteRequest | SuffixByteRequest | OffsetByteRequest

async def get(self, path: str, byte_range: ByteRequest | None):
    match byte_range:
        case None:
            return await self.fs._cat_file(path)
        case ExplicitByteRequest(start, end):
            return await self.fs._cat_file(path, start=start, end=end)
        case SuffixByteRequest(suffix):
            return await self.fs._cat_file(path, start=-suffix, end=None)
        case OffsetByteRequest(offset):
            return await self.fs._cat_file(path, start=offset, end=None)
        case _:
            raise Exception(...)
```

maxrjones:
> Well, after 5min of googling I learned about the match statement (new in Python 3.10), which kind of enables rust-style enums.

Time flies 😅 , I didn't realize Zarr-Python was already at 3.11+. This will be nice to use regardless of the argument syntax.

I share some of Norman's concerns over Tuple. While it's easy to document the behavior inside Zarr-Python, there is a risk of differences if/when people implement new storage classes externally. Also, while I think `(start, stop)` makes more sense than `(start, length)` due to its similarity to an HTTP Range, I am a bit concerned that the sudden change of behavior will cause issues for super early adopters.

I'd like to leave it to @normanrz @jhamman @d-v-b and any other zarr-python devs to decide on the path forward given the factors about simplicity and expressiveness noted in past comments 🙏 I'm happy to contribute additional prototypes or benchmarks if it helps your decision-making.

d-v-b:
> I am a bit concerned that the sudden change of behavior will cause issues for super early adopters.

IMO the least disruptive thing for adopters at any stage is to use semantics that are familiar. both python indexing and http range requests use `(start, stop)` semantics, which makes it more familiar than `(start, length)` so we should switch to the former as soon as we can. How about we use the explicit dataclasses until someone complains that they take too long to write, and then we can consider widening the type to accept tuples?

normanrz:
I think changing at this point is ok.
Again, I think the dataclasses are very expressive, here. We might even have a class like this:
```python
@dataclass
class ExplicitByteRequest:
    start: int
    end: int
    
    def __init__(self, start: int, *, end: int | None = None, length: int | None = None):
        if end is None and length is None:
             raise Exception("Neither end nor length is specified")
        if end is not None and length is not None:
             raise Exception("Only end OR length must be specified")
             
        object.__setattr__(self, "start", start)
        if end is not None:
            object.__setattr__(self, "end", start)
        else:
            object.__setattr__(self, "end", start + length)
            
assert ExplicitByteRequest(start=5, end=10) == ExplicitByteRequest(start=5, length=5)
```

Also, I wouldn't really consider this an end-user-facing API because it will be used by store (or other extensions) developers. Complaining about typing too much would be a bit funny.

src/zarr/abc/store.py
```diff
@@ -19,7 +20,34 @@
 
 __all__ = ["ByteGetter", "ByteSetter", "Store", "set_or_delete"]
 
-ByteRangeRequest: TypeAlias = tuple[int | None, int | None]
+
+@dataclass
+class ExplicitByteRequest:
```
normanrz, line 25:
```suggestion
class ExplicitByteRangeRequest:
```
I think we should go for the longer name for consistency.

normanrz (approved):
This is great! I really like how the API turned out.
Just one comment on the naming of the dataclasses.

src/zarr/abc/store.py
d-v-b: strictly speaking this specifies an interval that doesn't have anything to do with bytes or requests so maybe we can use the name `ExplicitInterval`

maxrjones: What about `RangeByteRequest` to match `<Suffix|Offset>ByteRequest`? Specifically, I'm not really sure how much interpretability `Explicit` adds.

normanrz: All of these are fine with me. I was just thinking that the name of the union type should be consistent with the invidiual types.
1. `ByteRangeRequest: TypeAlias = ExplicitByteRangeRequest | OffsetByteRangeRequest | SuffixByteRangeRequest`
2. `ByteRequest: TypeAlias = RangeByteRequest | OffsetByteRequest | SuffixByteRequest`
3. `Interval: TypeAlias = ClosedInterval | LeftOpenInterval | RightOpenInterval`

I prefer 1 or 2 because they are only used for the`byte_range` argument. But no strong opinion.

maxrjones: I also prefer 1 or 2 and will chose 2 for now given that it's more concise. I'm fine with other options though if someone else has a strong opinion on the name.

maxrjones:
FYI @normanrz https://github.com/maxrjones/zarr-byterangerequest-microbenchmarks/blob/main/plot-results.ipynb suggests that using match; case is actually noticeably slower than if; elif for explicit byte ranges. I know we're not optimizing around performance right now, but IMO the improved readabiliity in https://github.com/zarr-developers/zarr-python/pull/2585/commits/a7d35f876b1b628b3216da61ee26ba0f3a9d9cf8 is not worth any performance hit.

src/zarr/core/common.py
```diff
@@ -31,7 +31,6 @@
 ZATTRS_JSON = ".zattrs"
 ZMETADATA_V2_JSON = ".zmetadata"
 
-ByteRangeRequest = tuple[int | None, int | None]
```
maxrjones, line 34: I'm not 100% confident in this change, but couldn't otherwise quickly find a way to avoid circular imports. @d-v-b can you please confirm that this removal won't cause any issues?

maxrjones:
Thanks all for your help with this PR!

normanrz:
Thanks @maxrjones for pushing this through!

PR #2586, opened 2024-12-25
Adds a shards property to Array

normanrz:
Builds on #2463 

- Adds a `shards` property to Array that returns the shard shape and `None` if sharding isn't used
- `chunks` now returns the inner chunk shape, if sharding is used
- Moves the `chunks` and `shards` properties into the ArrayMetadata classes to handle v2 and v3 cases

I was wondering if `shards` should equal `chunks` if sharding isn't used. Let me know what you think.

TODO:


d-v-b:
> I was wondering if `shards` should equal `chunks` if sharding isn't used. Let me know what you think.

I think we should only report a tuple of ints for `shards` if and only if a sharding codec is being used -- otherwise it's not easy to tell the difference between an array that uses the sharding codec w/ shard shape=inner chunk shape, and an array that is not using the sharding codec. `shards = None` seems like a sane choice for the latter case.

d-v-b (approved):

normanrz:
Merged into #2463

Issue #1767, opened 2024-04-05
[v3] Documentation: update tutorial docs


jhamman:
We are not expecting many changes to the user experience as a part of v3 but a few things do come to mind as worth updating in the tutorial:

- Advanced async usage
- Describe usage w/ v2 and v3 specs
- Sharding
- Codec pipelines
- ? (suggestions welcome)

jhamman:
FYI - I am working on this now and will have a PR up for review by tomorrow.

PR #2589, opened 2024-12-29
docs: split tutorial into multiple user guide sections

jhamman:
This PR updates Zarr's documentation to include a new multi-page user-guide, in-place of the prior single page tutorial.

Notes for reviewers:

- This was split off from https://github.com/zarr-developers/zarr-python/pull/2568
- Most content remains unchanged from the tutorial, despite the large diff line count. Some new sections were added in 2568. The only net new section added was config.rst.
- Examples now use the Ipython sphinx directive

closes https://github.com/zarr-developers/zarr-python/issues/1767


docs/user-guide/arrays.rst
```diff
@@ -0,0 +1,528 @@
+
+Working with Arrays
```
dstansby, line 2:
```suggestion
Working with arrays
```
To match other titles of guide pages

docs/user-guide/storage.rst
```diff
@@ -1,8 +1,10 @@
+.. _tutorial_storage:
```
dstansby, line 1:
```suggestion
.. _guide_storage:
```

docs/user-guide/storage.rst
```diff
@@ -16,22 +18,29 @@ Implicit Store Creation
 In most cases, it is not required to create a ``Store`` object explicitly. Passing a string
 to Zarr's top level API will result in the store being created automatically.
 
-.. code-block:: python
+.. ipython:: python
+
+   import zarr
+
+   # implicitly create a writable LocalStore
```
dstansby, line 25:
```suggestion
   # Implicitly create a writable LocalStore
```

docs/user-guide/storage.rst
```diff
@@ -16,22 +18,29 @@ Implicit Store Creation
...
+   zarr.open_group("data/foo/bar", mode="w")
+
+   # implicitly create a read-only FsspecStore
```
dstansby, line 28:
```suggestion
   # Implicitly create a read-only FsspecStore
```

docs/user-guide/storage.rst
```diff
@@ -16,22 +18,29 @@ Implicit Store Creation
...
 
-   >>> import zarr
-   >>> zarr.open("data/foo/bar", mode="r")  # implicitly creates a read-only LocalStore
-   <Group file://data/foo/bar>
-   >>> zarr.open("s3://foo/bar", mode="r")  # implicitly creates a read-only RemoteStore
-   <Group s3://foo/bar>
-   >>> data = {}
-   >>> zarr.open(data, mode="w")  # implicitly creates a MemoryStore
-   <Group memory://4791444288>
+   # implicitly creates a MemoryStore
```
dstansby, line 35:
```suggestion
   # Implicitly creates a MemoryStore
```

docs/user-guide/storage.rst
```diff
@@ -40,55 +49,52 @@ Local Store
...
 ~~~~~~~~~~~~
 
-The :class:`zarr.storage.RemoteStore` stores the contents of a Zarr hierarchy in following the same
+The :class:`zarr.storage.FsspecStore` stores the contents of a Zarr hierarchy in following the same
 logical layout as the ``LocalStore``, except the store is assumed to be on a remote storage system
 such as cloud object storage (e.g. AWS S3, Google Cloud Storage, Azure Blob Store). The
-:class:`zarr.storage.RemoteStore` is backed by `Fsspec_` and can support any Fsspec backend
-that implements the `AbstractFileSystem` API,
+:class:`zarr.storage.FsspecStore` is backed by `fsspec`_ and can support any backend
+that implements the `AbstractFileSystem` API. ``storage_options`` can be used to configure
```
dstansby, line 75:
```suggestion
that implements the `AbstractFileSystem` API. ``storage_options`` can be used to configure
```
The AbstractFileSystem link here isn't working

docs/user-guide/consolidated_metadata.rst
```diff
@@ -1,7 +1,7 @@
 Consolidated Metadata
 =====================
 
-Zarr-Python implements the `Consolidated Metadata_` extension to the Zarr Spec.
+Zarr-Python implements the `Consolidated Metadata`_ extension to the Zarr Spec.
```
dstansby, line 4: If I search fo "consolidated" on this link it doesn't turn up with anything - can this be fixed to link to a more suitable place than the zarr spec itself, which doesn't mention consolidated metadata?

docs/user-guide/consolidated_metadata.rst
```diff
@@ -13,50 +13,49 @@ Usage
...
 
-   >>> import zarr
-   >>> store = zarr.storage.MemoryStore({}, mode="w")
-   >>> group = zarr.open_group(store=store)
-   >>> group.create_array(shape=(1,), name="a")
-   >>> group.create_array(shape=(2, 2), name="b")
-   >>> group.create_array(shape=(3, 3, 3), name="c")
-   >>> zarr.consolidate_metadata(store)
+.. ipython:: python
+   :okwarning:
```
dstansby, line 25: There aren't any warnings showing up in the built docs, can this be removed?

docs/user-guide/v3_todos.rst
dstansby, line 1: I don't think there's much value in having code for features that don't work in the docs. I'd suggest delete all the examples, have a short bullet point list of the missing features, and then link out to the relevant tutorial section in the v2 docs. This should also make it easier to see what's missing at a glance.

docs/user-guide/performance.rst
```diff
@@ -0,0 +1,332 @@
+
+Optimizing Performance
+======================
+
+.. ipython::
+   :suppress:
+
+   In [144]: rm -r data/
```
dstansby, line 8:
```suggestion
   rm -r data/
```

docs/user-guide/performance.rst
```diff
@@ -0,0 +1,332 @@
...
+
+Sometimes you are not free to choose the initial chunking of your input data, or
+you might have data saved with chunking which is not optimal for the analysis you
+have planned. In such cases it can be advantageous to re-chunk the data. For small
+datasets, or when the mismatch between input and output chunks is small
+such that only a few chunks of the input dataset need to be read to create each
+chunk in the output array, it is sufficient to simply copy the data to a new array
+with the desired chunking, e.g.:
+
+.. .. ipython:: python
```
dstansby, line 169: Is there a reason these blocks of code aren't being run?

docs/user-guide/performance.rst
```diff
@@ -0,0 +1,332 @@
...
+   z1 = zarr.array(store="data/example-2", data=np.arange(100000))
+   s = pickle.dumps(z1)
+   z2 = pickle.loads(s)
+   z1 == z2
+   np.all(z1[:] == z2[:])
+
+.. _tutorial_tips_blosc:
+
+Configuring Blosc
+~~~~~~~~~~~~~~~~~
```
dstansby, line 304: The heading level is wrong here, I don't think this is meant to be under "Pickle support"

pyproject.toml
```diff
@@ -86,6 +86,11 @@ docs = [
     'pydata-sphinx-theme',
     'numpydoc',
     'numcodecs[msgpack]',
+    'msgpack',
+    'ipython',
+    'pickleshare',
```
dstansby, line 91:
```suggestion
```
This doesn't seem to be used

pyproject.toml
```diff
@@ -86,6 +86,11 @@ docs = [
     'pydata-sphinx-theme',
     'numpydoc',
     'numcodecs[msgpack]',
+    'msgpack',
```
dstansby, line 89:
```suggestion
``` 
This is already included in the line above

docs/user-guide/index.rst
```diff
@@ -0,0 +1,33 @@
...
+
+Advanced Topics
+---------------
+
+.. toctree::
+    :maxdepth: 1
+
+    performance
+    consolidated_metadata
+    whatsnew_v3
```
dstansby, line 27: It doesn't feel right for a what's new page to be under Advanced topics - perhaps under a new "Zarr v3" section instead, which can also have the migration guide and TODOs pages?

docs/user-guide/config.rst
```diff
@@ -0,0 +1,43 @@
...
+
+Runtime configuration
+=====================
+
+The :mod:`zarr.core.config` module is responsible for managing the configuration of zarr
+and is based on the `donfig <https://github.com/pytroll/donfig>`_ Python library.
+
+Configuration values can be set using code like the following:
+
+.. code-block:: python
```
dstansby, line 11: This should be an IPython block to be consistent

dstansby (commented):
Liking the new sections, makes it much easier to read than the old tutorial 👍 

A couple of major questions:

The "copying/migrating data" tutorial section wasn't migrated from the old tutorial, is this intentional?

For IPython, what was the motivation for switching to this from doctests? As it stands it has a few major drawbacks:
1. Trying to copy/paste a block of code doesn't work, because it selects the `Out[...]` text too:
Screenshot 2024-12-30 at 10 07 54
2. The expected output isn't embedded any more. This gave us another layer of regression tests when we ran the tutorial as a doc test, and on top of making sure the code runs, made sure the output was as expected.
3. I find having a line break between each line of code makes it much less readable - old vs new:
Screenshot 2024-12-30 at 10 52 17
Screenshot 2024-12-30 at 10 51 24

Is it possible to fix these with new IPython plugin?

Finally, what's the plan for all the TODOs (especially the numcodecs and Sharding bits, and configuring blosc)? To do them in another PR before 3.0? If so would be good to open issues for these so they're not lost and released as TODOs in 3.0

docs/user-guide/arrays.rst
```diff
@@ -0,0 +1,528 @@
...
+Creating an array
+-----------------
+
+Zarr has several functions for creating arrays. For example:
+
+.. ipython:: python
+
+   import zarr
+
+   store = {}
```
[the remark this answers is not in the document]
dstansby: Never mind - this seems like the best way to do this.

docs/user-guide/index.rst
jhamman: I'd like to do away with this page in favor a single change log page. Can we do that in a follow up PR?

docs/user-guide/consolidated_metadata.rst
jhamman: I believe this silences warnings. I'll try removing it but if we should expect a warning when `consolidate_metadata` is used on v3 stores.

jhamman:
Thanks for the review @dstansby! 

> Liking the new sections, makes it much easier to read than the old tutorial 👍
> 
> A couple of major questions:
> 
> The "copying/migrating data" tutorial section wasn't migrated from the old tutorial, is this intentional?

We have not implemented `copy`, `copy_all`, and `copy_store` yet. I'll add these to the list of TODOs in the migration guide if they are not there already.

> For IPython, what was the motivation for switching to this from doctests? 

For one, we weren't using the doc tests! Beyond that, this isn't something I feel super strong about. We've had a lot of success using the Ipython directive in the Xarray project.

As it stands it has a few major drawbacks:
> 1. Trying to copy/paste a block of code doesn't work, because it selects the `Out[...]` text too:

Not true if you use the copy button, example below: 

```python
z1[:] = 42

z1[0, :] = np.arange(10000)

z1[:, 0] = np.arange(10000)
```

I think there is probably a way to further configure the style.

> 2. The expected output isn't embedded any more. This gave us another layer of regression tests when we ran the tutorial as a doc test, and on top of making sure the code runs, made sure the output was as expected.

We can insert assertions in hidden cells as desired. This would achieve the same thing.

> 3. I find having a line break between each line of code makes it much less readable - old vs new:
>
> Is it possible to fix these with new IPython plugin?

Likely yes. Can we iterate on the styling after this first PR goes in?
> 
> Finally, what's the plan for all the TODOs (especially the numcodecs and Sharding bits, and configuring blosc)? To do them in another PR before 3.0? If so would be good to open issues for these so they're not lost and released as TODOs in 3.0

I'm waiting #2463. I have a branch combining these but I'll need to start it over now that I've split up the docs refactor. So yes, let's do them in another PR.

docs/user-guide/arrays.rst
```diff
@@ -0,0 +1,528 @@
...
+useful. E.g.:
+
+.. ipython:: python
+
+   a = np.arange(10)
+   zarr.save('data/example-3.zarr', a)
+   zarr.load('data/example-3.zarr')
+
+Please note that there are a number of other options for persistent array
+storage, see the :ref:`tutorial_storage` guide for more details.
```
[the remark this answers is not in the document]
jhamman: done.

docs/user-guide/arrays.rst
```diff
@@ -0,0 +1,528 @@
...
+-----------------
+
+In the examples above, compressed data for each chunk of the array was stored in
+main memory. Zarr arrays can also be stored on a file system, enabling
+persistence of data between sessions. For example:
+
+.. ipython:: python
+
+   z1 = zarr.open(
+       store='data/example-2.zarr',
```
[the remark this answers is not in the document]
jhamman: changed the text below

docs/user-guide/arrays.rst
```diff
@@ -0,0 +1,528 @@
...
+requested region into memory as a NumPy array, e.g.:
+
+.. ipython:: python
+
+   z[0, 0]
+   z[-1, -1]
+   z[0, :]
+   z[:, 0]
+   z[:]
+
```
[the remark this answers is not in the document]
jhamman: done.

docs/user-guide/arrays.rst
```diff
@@ -0,0 +1,528 @@
...
+For a complete list of array creation routines see the :mod:`zarr.api.synchronous`
+module documentation.
+
+.. _tutorial_array:
+
+Reading and writing data
+------------------------
+
+Zarr arrays support a similar interface to `NumPy <https://numpy.org/doc/stable/>`
```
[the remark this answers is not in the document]
jhamman: fixed (missing the trailing _)

docs/user-guide/arrays.rst
```diff
@@ -0,0 +1,528 @@
...
+   )
+   z
+
+The code above creates a 2-dimensional array of 32-bit integers with 10000 rows
+and 10000 columns, divided into chunks where each chunk has 1000 rows and 1000
+columns (and so there will be 100 chunks in total). The data is written to a
+:class:`zarr.storage.MemoryStore` (e.g. an in-memory dict). See
+:ref:`tutorial_persist` for details on storing arrays in other stores.
+
+For a complete list of array creation routines see the :mod:`zarr.api.synchronous`
```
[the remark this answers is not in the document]
jhamman: done.

docs/user-guide/arrays.rst
```diff
@@ -0,0 +1,528 @@
...
+A Zarr array can be resized, which means that any of its dimensions can be
+increased or decreased in length. For example:
+
+.. ipython:: python
+
+   z = zarr.zeros(
+      store="data/example-4.zarr",
+      shape=(10000, 10000),
+      chunks=(1000, 1000)
+   )
```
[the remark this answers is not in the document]
jhamman: Possible but I'm going to leave this for later. We'd have to go back to an array from a ways up the guide.

docs/user-guide/arrays.rst
```diff
@@ -0,0 +1,528 @@
...
+.. ipython:: python
+
+   z.info
+
+The :func:`zarr.Array.info_complete` method inspects the underlying store and
+prints additional diagnostics, e.g.:
+
+.. ipython:: python
+
+   z.info_complete()
```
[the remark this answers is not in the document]
jhamman: done.

docs/user-guide/arrays.rst
jhamman: done (and for a few other pages)

docs/user-guide/storage.rst
jhamman: fixed.

docs/user-guide/consolidated_metadata.rst
jhamman: Fixed. + added a warning about the lack of an accepted extension to the spec.

docs/user-guide/performance.rst
jhamman: fixed.

docs/user-guide/config.rst
jhamman: done.

docs/user-guide/index.rst
dstansby: 👍 can definitely save this for a follow up PR (maybe one of the other open ones). I think it's worth having a slighlty more fleshed out "what's new in v3" guide that's more explain-y than a changelog would be though (which is what this page currently does very well)

dstansby:
>>  Trying to copy/paste a block of code doesn't work, because it selects the Out[...] text too:

> Not true if you use the copy button

This isn't a solution if you want to select two lines of code from a longer block.

>> This gave us another layer of regression tests when we ran the tutorial as a doc test, and on top of making sure the code runs, made sure the output was as expected.

> We can insert assertions in hidden cells as desired. This would achieve the same thing.

But with a much higher maintenance burden I think? We would have to manually concoct a hidden assertion for every code block in this case, whereas with the doctests we can just generate all the output automatically.


So given there don't seem to be any benefits to using the IPython plugin, but there still seem to be drawbacks to me, I'd say we should keep existing code block style for now.

re. the TODOs, I'm a bit tired to do it now but tomorrow I can open issues to keep track of them (if you haven't already).

jhamman:
My view is that the most important thing is to have documentation that actually runs! I'm happy to hand this off to someone with sufficient experience with doctests (maybe @dstansby) but that isn't me. If no one is volunteering to do it right now, I'm going to press forward with this solution and we can revisit when someone else has time/energy for another doctest effort. I'll also note that these aren't mutually exclusive concepts (reading up on the ipython extension is [worthwhile](https://ipython.readthedocs.io/en/stable/sphinxext.html))

docs/user-guide/arrays.rst
```diff
@@ -0,0 +1,528 @@
...
+compression algorithms internally to compress the data. Blosc also provides
+highly optimized implementations of byte- and bit-shuffle filters, which can
+improve compression ratios for some data. A list of the internal compression
+libraries available within Blosc can be obtained via:
+
+.. ipython:: python
+
+   from numcodecs import blosc
+
+   blosc.list_compressors()
```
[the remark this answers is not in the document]
jhamman: this is relevant for Zarr v2 arrays. We will rev this section after #2463 goes in.

dstansby:
In order to keep this moving, lets drop the switch to using the IPython plugin, since that shouldn't block a v3 release.

dstansby (changes requested):
Marking as request changes to revert the IPython change. @jhamman I'd be happy to change the code snippets in code back, but don't want to push to your PR without permission - let me know either way.

jhamman:
If you can update the code snippets today, feel free to push to this branch. I'm all checked in. Like I said above, I'm not planning on reverting the change myself though.

normanrz:
> I think it's a reasonable ask to revert the change though, and not merge before it's reverted.

To be honest, I think it is more important to merge this PR because of the dependent PRs. I feel this is blocking our progress right now. We can change away from IPython at any later point in time.

d-v-b:
> > I think it's a reasonable ask to revert the change though, and not merge before it's reverted.
> 
> To be honest, I think it is more important to merge this PR quickly because of the dependent PRs. I feel this is blocking our progress right now. We can change away from IPython at any later point in time.

+1 to this, lets get this PR in and sort out doctest / ipython later

dcherian:
I concur. Working doc examples are very valuable!

dstansby:
I'm sorry, but I really want to insist we don't merge this with IPython in it, because it's a change has demonstratable downsides. On a point of principle, just opening a large PR shouldn't come with an assumption that it's going to get merged without engaging with review (if nothing else, it's not particularly respectable of the reviewers time). On a practical point, I spent about an hour sorting this out at https://github.com/zarr-developers/zarr-python/pull/2623 today. So 🙏 can folks review and merge https://github.com/zarr-developers/zarr-python/pull/2623 into this PR first.

d-v-b:
> So 🙏 can folks review and merge #2623 into this PR first.

If the main value-add from the ipython plugin is that it ensures that the docs contain the output of current code, then an alternative to that plugin should achieve the same goal. It seems like #2623 doesn't have the doctests running for the current codebase, so that PR isn't really a replacement for the ipython plugin.

d-v-b:
to be clear, I don't like the style of the ipython output either, and if we can't make lines of code copy + pasteable with the ipython plugin then we should _asymptotically_ move to doctests. But we kind of need docs _today_ (or yesterday :) ) that reflect what the code does _today_, and the ipython plugin seems expedient for that.

dstansby:
I'm not arguing that IPython isn't valuable - just that introduces downsides, and because it seems to need more discussion we should stick with the status quo which is doctests to avoid holding up v3.

d-v-b:
> the status quo which is doctests, to avoid holding up v3.

admittedly I'm late to this discussion, but my impression was that the status quo was that the doctests were not actually getting run, resulting in stale example code in the docs. I don't think anyone wants that. 

So the next question is, how can we ensure that the docs contain real examples? Either we make the doctests work, or we use another strategy, e.g. running code when we build the docs (the ipython directive). Given the current context, we should do whichever of these two is most expedient, in the interest of moving things forward -- the most important thing is that the generated docs are _correct_. This allows everyone to hack on the docs productively. IMO, the only thing that should block that part of this PR is a functionally equivalent solution to the problem solved with the ipython directive. Does that make sense?

dstansby:
> admittedly I'm late to this discussion, but my impression was that the status quo was that the doctests were not actually getting run, resulting in stale example code in the docs. I don't think anyone wants that.

👍 , which is why I fixed them in https://github.com/zarr-developers/zarr-python/pull/2623...

jhamman:
Let's deescalate this a bit.

If we can't get the doctests pr passing with ci today, we can merge this as is (with ipython) and go back to doctests as soon as it's ready.

I'm hearing general agreement that this PR is directionally aligned with where we want to go. Let's continue to iterate but in a non blocking way. 

@dstansby - I know this isn't your ideal but I think it's what is best for the project overall.

dstansby:
Okay, quick update, I'm trialling enabling doctests at https://github.com/zarr-developers/zarr-python/pull/2626, which I *think* I have working. If it works, that PR should fail, and then I'll close and merge into https://github.com/zarr-developers/zarr-python/pull/2623. Sorry it's a bit complex, but avoids me messing around with GH actions config in https://github.com/zarr-developers/zarr-python/pull/2623 to get the tests running...

dstansby:
Quick update, I've:

- added a new doctests CI run (was much simpler than running doctests in all the existing test runs)
- added a new hatch environment that can be used to run the doctests (hatch run doctest:run) or automatically fix the output if it changes (hatch run doctest:fix)

Some of the doctests need fixing because of https://github.com/zarr-developers/zarr-python/pull/2463 being merged, but hopefully the new CI job and hatch env make that a lot easier for someone else to do.

dstansby (changes requested):
Doctests need fixing - otherwise I think this is 👍

normanrz:
> Doctests need fixing - otherwise I think this is 👍

Working on it.

docs/user-guide/consolidated_metadata.rst
```diff
@@ -0,0 +1,119 @@
+.. only:: doctest
+
+   >>> from pprint import pprint
```
dstansby, line 3: I think worth showing this and importing it where it's used instead of hiding it away at the top, so users trying to run the code know where the `pprint` command is coming from?

tests/test_tree.py
```diff
@@ -10,6 +11,8 @@
 
 @pytest.mark.parametrize("root_name", [None, "root"])
 def test_tree(root_name: Any) -> None:
+    os.environ["OVERRIDE_COLOR_SYSTEM"] = "truecolor"
```
dstansby, line 14: Just out of interest, why's this needed?

dstansby (approved):
LGTM 🚢  - a couple of optional comments. I haven't reviewed any of the actual content this time.

tests/test_tree.py
normanrz: `test_tree.py` expects the tree to be rendered with the control sequences for text formatting (e.g. bold font). However, by default `rich` doesn't include that when running within pytest, because it detects that it is not running in a real terminal as the output is captured by pytest. This overrides the behavior in `rich`. This issue only surfaced now, because I added `rich` to the test dependencies and therefore `test_tree.py` wasn't skipped anymore.
https://github.com/zarr-developers/zarr-python/blob/main/tests/test_tree.py#L31-L54

docs/user-guide/consolidated_metadata.rst
normanrz: See #2632

PR #2590, opened 2024-12-29
docs: replace getting_started page with quickstart

jhamman:
This PR replaces Zarr's Getting Started page with a proper Quickstart. 

Notes for reviewers:

- This was split out from #2568 and builds on #2589.
- Examples use the Ipython directive
- Cloud storage examples point at real data. This is a bit of a maintenance hazard but in my view, is worth it to have a real working example that anyone can try.

PR #2591, opened 2024-12-29
docs: update installation docs

jhamman:
This PR replaces moves the installation doc page into the new user-guide and makes some minor updates to the installation docs.

Notes for reviewers:
- This was split out from https://github.com/zarr-developers/zarr-python/pull/2568 and builds on https://github.com/zarr-developers/zarr-python/pull/2590.


docs/user-guide/installation.rst
```diff
@@ -12,12 +26,15 @@ There are a number of optional dependency groups you can install for extra funct
 These can be installed using ``pip install "zarr[<extra>]"``, e.g. ``pip install "zarr[gpu]"``
 
 - ``gpu``: support for GPUs
-- ``fsspec``: support for reading/writing to remote data stores
-- ``tree``: support for pretty printing of directory trees
+- ``remote``: support for reading/writing to remote data stores
+
+Additional option dependencies include ``rich``, ``universal_pathlib``. These must be installed separately.
```
normanrz, line 31:
```suggestion
Additional optional dependencies include ``rich``, ``universal_pathlib``. These must be installed separately.
```

normanrz (approved):

docs/user-guide/installation.rst
```diff
@@ -1,9 +1,23 @@
...
+---------------------
+
+Required dependencies include:
+
+- `Python <https://docs.python.org/3/>`_ (3.11 or later)
+- `packaging <https://packaging.pypa.io>`_ (22.0 or later)
+- `numpy <https://numpy.org>`_ (1.25 or later)
+- `numcodecs[crc32c] <https://numcodecs.readthedocs.io>`_ (0.14 or later)
+- `typing_extensions <https://typing-extensions.readthedocs.io>`_ (4.9 or later)
+- `donfig <https://donfig.readthedocs.io>`_ (0.8 or later)
```
dstansby, line 14: Listing these here is one more place they need updating when we update any of these versions or dependencies. Since pip and conda automatically handle dependencies, I'd be pro getting rid of this section entirely?

docs/user-guide/installation.rst
dstansby, line 31: We do have the `optional` extra that installs these two - I think either we should mention that here in the docs, or get rid of it (I have no strong feelings either way)

docs/user-guide/installation.rst
```diff
@@ -34,4 +51,4 @@ Zarr has endorsed `Scientific-Python SPEC 0 <https://scientific-python.org/specs
 
 Development
 -----------
-To install the latest development version of Zarr, see `the contributing guide <contributing.html>`_.
+To install the latest development version of Zarr, see `the contributing guide <../developers/contributing.html>`_.
```
dstansby, line 54: I think this link is broken.

dstansby (approved):
Broken link needs fixing, and two other suggestions. Otherwise looks 👍

dstansby (changes requested):
Woops, previous review should have been request changes

docs/user-guide/installation.rst
jhamman: We use the `optional` group as a dependency set in our test suite at the moment. I agree that we should remove it but to keep the scope of this PR down, I'm going to leave this as is.

docs/user-guide/installation.rst
jhamman: I regularly reference Xarray's installation page for this information so I'm reasonably certain folks will find this valuable. If it turns out to be a maintanence burden, then we can revisit it in the future.

docs/user-guide/installation.rst
jhamman: fixed.

docs/user-guide/installation.rst
dstansby: 👍  For my own interest, I'm intriuged what you use the info for as a user? (not as a developer)

docs/user-guide/installation.rst
dstansby: Ah of course, I forgot it was used for the tests

dstansby (dismissed):
Looks like the doc build is still broken, otherwise 👍

PR #2592, opened 2024-12-29
docs: add new top level about page

jhamman:
This PR adds a new `about` page to Zarr's documentation. 

Notes for reviewers:

- This was split out from https://github.com/zarr-developers/zarr-python/pull/2568 and builds on https://github.com/zarr-developers/zarr-python/pull/2589.
- This content was previously part of the `getting-started` page which is turning into the `quickstart` page in #2590.

### DIFF
--- a/docs/about.rst
+++ b/docs/about.rst
@@ -1,0 +1,22 @@
+About
+=====
+
+Zarr is a format for the storage of chunked, compressed, N-dimensional arrays
+inspired by `HDF5 <https://www.hdfgroup.org/HDF5/>`_, `h5py
+<https://www.h5py.org/>`_ and `bcolz <https://bcolz.readthedocs.io/>`_.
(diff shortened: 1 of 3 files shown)


normanrz (approved):

dstansby (changes requested):
Shouldn't this content be removed from the getting started page at the same time?

jhamman:
> Shouldn't this content be removed from the getting started page at the same time?

I'm trying to avoid merge conflicts with #2590 which replaces the getting started page with a new quickstart page.

dstansby:
Given the whole of getting started is deleted in https://github.com/zarr-developers/zarr-python/pull/2590, that would be a pretty simple merge/rebase conflict to address if it came up no?

dstansby (approved):

Issue #2602, opened 2024-12-30
Duplicate documentation of stores


dstansby:
Searching e.g., memorystore in the v3 docs shows that stores seem to be documented twice in the API docs: https://zarr.readthedocs.io/en/latest/search.html?q=memorystore

I think the fix here is to clean up the store namespacing, so all the public API is in zarr.store instead of the sub-modules.

PR #2603, opened 2024-12-30
Clean up public store API

dstansby:
This cleans up the public API surface for stores, by making all the zarr.store submodules private, and exposing the API in the `zarr.store` namespace. I think this is what was always intended, but changing the sub-module names to start with a _ makes sure they're not documented.

Tagged as 3.0, because this is an API change it would be good to get in before final release.

Fixes https://github.com/zarr-developers/zarr-python/issues/2602


jhamman:
What do we think about also making the following objects private:

- `StoreLike`
- `StorePath`
- `WrapperStore`
- `make_store_path`

We can always make them public later but these four seem like good candidates to keep private for now.

dstansby:
StoreLike and StorePath should part of the public API, because they're types of arguments of public functions (e.g., `create()`, `from_dict()`), `WrapperStore` should be public because `LoggingStore` inherits from it.

I guess `make_store_path` could be private, but I think out of scope for this PR to give that potential change more visibility to other devs instead of hiding it here.

jhamman:
I opened https://github.com/zarr-developers/zarr-python/pull/2628 for the `make_store_path` issue which can go in after this.

jhamman (approved):

Issue #2604, opened 2024-12-30
Duplicate codec documentation


dstansby:
[Searching e.g., blosccodec in the v3 docs](https://zarr--2596.org.readthedocs.build/en/2596/search.html?q=blosccodec#) shows that codecs are be documented twice in the API docs.

I think the fix here is to clean up the codec namespacing, so all the public API is in zarr.codecs instead of the sub-modules. @normanrz do you think this is the right fix? I did try that in https://github.com/zarr-developers/zarr-python/pull/2176, but I'm not sure why I closed it... I can have another go.

normanrz:
Go for it!

PR #2605, opened 2024-12-30
Don't document zarr.codec submodules

dstansby:
Fixes https://github.com/zarr-developers/zarr-python/issues/2604. I tried renaming the sub-modules, but that messed around with the codec registration, so this seems like the easiest fix.


normanrz (approved):
Makes sense!

PR #2607, opened 2024-12-31
Deprecate (Async)Array.create

normanrz:
Builds upon #2463

- Deprecates `AsyncArray.create` and `Array.create` to `Array._create` including deprecation warnings in favor of `zarr.create_array`.
- Adds `AsyncArray._create` and `Array._create` which now contain the code of the unprefixed methods. We should not use them for new code anymore and remove once other APIs have been deprecated and removed.

I added 2 new features to make `create_array` usable for all codec configurations I could think of: 
- Adds a dict-notation to the `shards` kwarg in `create_array` to specify the index location, e.g. `shards={"shape": (32, 32), "index_location": "start"}`
- Adds an optional `array_bytes_codec` kwarg to `create_array` to override the array-to-bytes codec for v3 arrays, e.g. to specify the endianness.


src/zarr/api/asynchronous.py
```diff
@@ -666,6 +662,54 @@ async def group(
...
+    ----------
+    store : Store or str
+        Store or path to directory in file system.
+    path : str, optional
+        Group path within store.
+    overwrite : bool, optional
+        If True, pre-existing data at ``path`` will be deleted before
+        creating the group.
+    zarr_format : {2, 3, None}, optional
+        The zarr format to use when saving.
```
d-v-b, line 686:
```suggestion
        The zarr format to use when saving. If ``None`` (the default), the zarr format will be retrieved from the current state of the zarr config object.
```

d-v-b: or something like this. there should be a clear path for people to follow to figure out what the default version will be.

src/zarr/api/asynchronous.py
```diff
@@ -666,6 +662,54 @@ async def group(
...
+    Returns
+    -------
+    g : group
+        The new group.
+    """
+
+    if zarr_format is None:
+        zarr_format = _default_zarr_version()
+
+    # TODO: fix this when modes make sense. It should be `w` for overwriting, `w-` otherwise
```
d-v-b, line 700:
```suggestion
```

src/zarr/api/synchronous.py
```diff
@@ -523,6 +525,29 @@ def open_group(
...
 
+def create_group(
+    store: StoreLike,
+    *,
+    path: str | None = None,
+    zarr_format: ZarrFormat | None = None,
+    overwrite: bool = False,
+    attributes: dict[str, Any] | None = None,
+    storage_options: dict[str, Any] | None = None,
+) -> Group:
```
d-v-b, line 536: we should paste in a docstring based on the async version once we are happy with that one

tests/test_codecs/test_codecs.py
```diff
@@ -75,26 +77,17 @@ async def test_order(
...
-            ShardingCodec(
-                chunk_shape=(16, 8),
-                codecs=[TransposeCodec(order=order_from_dim(store_order, data.ndim)), BytesCodec()],
-            )
-        ]
-        if with_sharding
-        else [TransposeCodec(order=order_from_dim(store_order, data.ndim)), BytesCodec()]
-    )
 
     with config.set({"array.order": runtime_write_order}):
```
d-v-b, line 81: should we alter this test to take a `config` kwarg? happy to deem that out of scope for this PR

tests/test_codecs/test_codecs.py
```diff
@@ -370,17 +362,66 @@ def test_invalid_metadata(store: Store) -> None:
...
+    spath3 = StorePath(store, "invalid_order")
+    with pytest.raises(TypeError):
+        zarr.create_array(
+            spath3,
+            shape=(16, 16),
+            chunks=(16, 16),
+            dtype=np.dtype("uint8"),
+            fill_value=0,
+            filters=[
+                TransposeCodec(order="F"),  # type: ignore[arg-type]
```
d-v-b, line 376: Is this testing anything more than `TransposeCodec`'s ability to correctly type-check its inputs? In which case maybe this should be covered by the codec tests?

tests/test_codecs/test_codecs.py
```diff
@@ -370,17 +362,66 @@ def test_invalid_metadata(store: Store) -> None:
...
+    with pytest.raises(ValueError):
+        zarr.create_array(
+            spath5,
+            shape=(16, 16),
+            shards=(16, 16),
+            chunks=(8,),
+            dtype=np.dtype("uint8"),
+            fill_value=0,
+        )
+    spath6 = StorePath(store, "invalid_inner_chunk_shape")
```
d-v-b, line 389: it's fine to keep this here but I think this case should be handled by the sharding codec test suite

src/zarr/core/array.py
```diff
@@ -3448,4 +3625,442 @@ def _get_default_codecs(
...
+    store: str | StoreLike,
+    *,
+    name: str | None = None,
+    shape: ShapeLike,
+    dtype: npt.DTypeLike,
+    chunks: ChunkCoords | Literal["auto"] = "auto",
+    shards: ShardsParam | None = None,
+    filters: FiltersParam | None = "auto",
+    compressors: CompressorsParam = "auto",
+    array_bytes_codec: ArrayBytesCodecParam = "auto",
```
d-v-b, line 3667: I think this parameter is important, but I think we can do better with the name to express a bit more what it does, and detach the name from the type (which might change). Unfortunately the only ideas I have are `array_serializer` `chunk_serializer` but I'm sure there are better options.

src/zarr/core/array.py
```diff
@@ -3448,4 +3625,442 @@ def _get_default_codecs(
...
+)
+ArrayBytesCodecParam: TypeAlias = dict[str, JSON] | ArrayBytesCodec | Literal["auto"]
+
+
+class ShardsConfigParam(TypedDict):
+    shape: ChunkCoords
+    index_location: ShardingCodecIndexLocation | None
+
+
+ShardsParam: TypeAlias = ChunkCoords | ShardsConfigParam | Literal["auto"]
```
d-v-b, line 3654: 👍

d-v-b (changes requested):
This looks great! But I do think we should try to improve the name of the `array_bytes_codec` kwarg. see https://github.com/zarr-developers/zarr-python/pull/2607#discussion_r1900827551

src/zarr/core/array.py
normanrz: What about just `codec`?

d-v-b: to me, `codec` is a category that includes `ArrayBytesCodec` and `BytesBytesCodec`; I think the ideal name would somehow convey that this entity is tasked with unpacking (or packing?) an array into bytes. `chunk_<verb>+er` seems like a good template, but I'm not sure what verbs fit well here. 🤔  `chunk_serializer`, `chunk_packer`, `chunk_streamer`... `chunk_serializer` seems the best to me but it's far from great.

normanrz: I think `chunk_serializer` is fine.
Maybe even just `serializer`. That would be more in line with `filters` and `compressors.`

d-v-b: +1 to dropping `chunk_` as long as we don't think people will be confused about the scope

jhamman (approved):
lgtm, thanks @normanrz

normanrz:
Merged into #2463.

PR #2609, opened 2024-12-31
Improve deprecation of zarr.creation and zarr.convenience

dstansby:
- Add a better deprecation warning (I think these are intended to be used from `zarr.`?)
- Add a warning to the API docs page


jhamman (approved):

jhamman:
Would you do the same for `convenience.py`?

normanrz (approved):

PR #2622, opened 2025-01-02
`zarr.array` from from an existing `zarr.Array`

### LINKED ISSUE #2410: [v3] zarr.array from from an existing `zarr.Array`
### Zarr version

3.0.0.beta

### Numcodecs version

0.13

### Python Version

3.11

### Operating System

Mac

### Installation

pip

### Description

It used to be possible to create a one zarr array from another. This is currently broken but would be interested to bring back in a more optimized form using the underlying async array for streaming.

### Steps to reproduce

```python
In [24]: a = zarr.zeros((10000, 10000), chunks=(100,100), dtype='uint16',
    ...: store='a.zarr')

In [25]: b = zarr.array(a, chunks=(100, 200), store='b.zarr')
--------------------------------------------------------------------------
SyncError                                Traceback (most recent call last)
Cell In[25], line 1
----> 1 b = zarr.array(a, chunks=(100, 200), store='b.zarr')

File ~/miniforge3/envs/icechunk-demo/lib/python3.12/site-packages/zarr/api/synchronous.py:164, in array(data, **kwargs)
    163 def array(data: NDArrayLike, **kwargs: Any) -> Array:
--> 164     return Array(sync(async_api.array(data=data, **kwargs)))

File ~/miniforge3/envs/icechunk-demo/lib/python3.12/site-packages/zarr/core/sync.py:141, in sync(coro, loop, timeout)
    138 return_result = next(iter(finished)).result()
    140 if isinstance(return_result, BaseException):
--> 141     raise return_result
    142 else:
    143     return return_result

File ~/miniforge3/envs/icechunk-demo/lib/python3.12/site-packages/zarr/core/sync.py:100, in _runner(coro)
     95 """
     96 Await a coroutine and return the result of running it. If awaiting the coroutine raises an
     97 exception, the exception will be returned.
     98 """
     99 try:
--> 100     return await coro
    101 except Exception as ex:
    102     return ex

File ~/miniforge3/envs/icechunk-demo/lib/python3.12/site-packages/zarr/api/asynchronous.py:529, in array(data, **kwargs)
    526 z = await create(**kwargs)
    528 # fill with data
--> 529 await z.setitem(slice(None), data)
    531 return z

File ~/miniforge3/envs/icechunk-demo/lib/python3.12/site-packages/zarr/core/array.py:880, in AsyncArray.setitem(self, selection, value, prototype)
    874     prototype = default_buffer_prototype()
    875 indexer = BasicIndexer(
    876     selection,
    877     shape=self.metadata.shape,
    878     chunk_grid=self.metadata.chunk_grid,
    879 )
--> 880 return await self._set_selection(indexer, value, prototype=prototype)

File ~/miniforge3/envs/icechunk-demo/lib/python3.12/site-packages/zarr/core/array.py:853, in AsyncArray._set_selection(self, indexer, value, prototype, fields)
    850 value_buffer = prototype.nd_buffer.from_ndarray_like(value)
    852 # merging with existing data and encoding chunks
--> 853 await self.codec_pipeline.write(
    854     [
    855         (
    856             self.store_path / self.metadata.encode_chunk_key(chunk_coords),
    857             self.metadata.get_chunk_spec(chunk_coords, self.order, prototype),
    858             chunk_selection,
    859             out_selection,
    860         )
    861         for chunk_coords, chunk_selection, out_selection in indexer
    862     ],
    863     value_buffer,
    864     drop_axes=indexer.drop_axes,
    865 )

File ~/miniforge3/envs/icechunk-demo/lib/python3.12/site-packages/zarr/codecs/pipeline.py:456, in BatchedCodecPipeline.write(self, batch_info, value, drop_axes)
    450 async def write(
    451     self,
    452     batch_info: Iterable[tuple[ByteSetter, ArraySpec, SelectorTuple, SelectorTuple]],
    453     value: NDBuffer,
    454     drop_axes: tuple[int, ...] = (),
    455 ) -> None:
--> 456     await concurrent_map(
    457         [
    458             (single_batch_info, value, drop_axes)
    459             for single_batch_info in batched(batch_info, self.batch_size)
    460         ],
    461         self.write_batch,
    462         config.get("async.concurrency"),
    463     )

File ~/miniforge3/envs/icechunk-demo/lib/python3.12/site-packages/zarr/core/common.py:64, in concurrent_map(items, func, limit)
     61     async with sem:
     62         return await func(*item)
---> 64 return await asyncio.gather(*[asyncio.ensure_future(run(item)) for item in items])

File ~/miniforge3/envs/icechunk-demo/lib/python3.12/site-packages/zarr/core/common.py:62, in concurrent_map.<locals>.run(item)
     60 async def run(item: tuple[Any]) -> V:
     61     async with sem:
---> 62         return await func(*item)

File ~/miniforge3/envs/icechunk-demo/lib/python3.12/site-packages/zarr/codecs/pipeline.py:374, in BatchedCodecPipeline.write_batch(self, batch_info, value, drop_axes)
    353 chunk_bytes_batch = await concurrent_map(
    354     [
    355         (
   (...)
    362     config.get("async.concurrency"),
    363 )
    364 chunk_array_batch = await self.decode_batch(
    365     [
    366         (chunk_bytes, chunk_spec)
   (...)
    370     ],
    371 )
    373 chunk_array_batch = [
--> 374     self._merge_chunk_array(
    375         chunk_array, value, out_selection, chunk_spec, chunk_selection, drop_axes
    376     )
    377     for chunk_array, (_, chunk_spec, chunk_selection, out_selection) in zip(
    378         chunk_array_batch, batch_info, strict=False
    379     )
    380 ]
    382 chunk_array_batch = [
    383     None
    384     if chunk_array is None or chunk_array.all_equal(chunk_spec.fill_value)
   (...)
    388     )
    389 ]
    391 chunk_bytes_batch = await self.encode_batch(
    392     [
    393         (chunk_array, chunk_spec)
   (...)
    397     ],
    398 )

File ~/miniforge3/envs/icechunk-demo/lib/python3.12/site-packages/zarr/codecs/pipeline.py:316, in BatchedCodecPipeline._merge_chunk_array(self, existing_chunk_array, value, out_selection, chunk_spec, chunk_selection, drop_axes)
    314     chunk_value = value
    315 else:
--> 316     chunk_value = value[out_selection]
    317     # handle missing singleton dimensions
    318     if drop_axes != ():

File ~/miniforge3/envs/icechunk-demo/lib/python3.12/site-packages/zarr/core/buffer/cpu.py:180, in NDBuffer.__getitem__(self, key)
    179 def __getitem__(self, key: Any) -> Self:
--> 180     return self.__class__(np.asanyarray(self._data.__getitem__(key)))

File ~/miniforge3/envs/icechunk-demo/lib/python3.12/site-packages/zarr/core/array.py:1327, in Array.__getitem__(self, selection)
   1325     return self.vindex[cast(CoordinateSelection | MaskSelection, selection)]
   1326 elif is_pure_orthogonal_indexing(pure_selection, self.ndim):
-> 1327     return self.get_orthogonal_selection(pure_selection, fields=fields)
   1328 else:
   1329     return self.get_basic_selection(cast(BasicSelection, pure_selection), fields=fields)

File ~/miniforge3/envs/icechunk-demo/lib/python3.12/site-packages/zarr/_compat.py:43, in _deprecate_positional_args.<locals>._inner_deprecate_positional_args.<locals>.inner_f(*args, **kwargs)
     41 extra_args = len(args) - len(all_args)
     42 if extra_args <= 0:
---> 43     return f(*args, **kwargs)
     45 # extra_args > 0
     46 args_msg = [
     47     f"{name}={arg}"
     48     for name, arg in zip(kwonly_args[:extra_args], args[-extra_args:], strict=False)
     49 ]

File ~/miniforge3/envs/icechunk-demo/lib/python3.12/site-packages/zarr/core/array.py:1769, in Array.get_orthogonal_selection(self, selection, out, fields, prototype)
   1767     prototype = default_buffer_prototype()
   1768 indexer = OrthogonalIndexer(selection, self.shape, self.metadata.chunk_grid)
-> 1769 return sync(
   1770     self._async_array._get_selection(
   1771         indexer=indexer, out=out, fields=fields, prototype=prototype
   1772     )
   1773 )

File ~/miniforge3/envs/icechunk-demo/lib/python3.12/site-packages/zarr/core/sync.py:128, in sync(coro, loop, timeout)
    126     loop0 = asyncio.events.get_running_loop()
    127     if loop0 is loop:
--> 128         raise SyncError("Calling sync() from within a running loop")
    129 except RuntimeError:
    130     pass

SyncError: Calling sync() from within a running loop
```

### Additional output

_No response_

brokkoli71:
- fixes #2410

added concurrent streaming of source array into new array

TODO:

### DIFF
--- a/src/zarr/api/synchronous.py
+++ b/src/zarr/api/synchronous.py
@@ -339,7 +339,7 @@ def tree(grp: Group, expand: bool | None = None, level: int | None = None) -> An
 
 
 # TODO: add type annotations for kwargs
-def array(data: npt.ArrayLike, **kwargs: Any) -> Array:
+def array(data: npt.ArrayLike | Array, **kwargs: Any) -> Array:
     """Create an array filled with `data`.
 
     Parameters
--- a/src/zarr/api/asynchronous.py
+++ b/src/zarr/api/asynchronous.py
@@ -18,6 +18,8 @@ from zarr.core.common import (
     ChunkCoords,
     MemoryOrder,
     ZarrFormat,
+    parse_dtype,
+    concurrent_map,
     _warn_order_kwarg,
     _warn_write_empty_chunks_kwarg,
     parse_dtype,
@@ -550,6 +552,21 @@ async def array(
     array : array
         The new array.
     """
+
+    if isinstance(data, Array):
+        chunks = kwargs.pop("chunks", None) or data.chunks
+        new_array = await create(shape=data.shape, chunks=chunks, dtype=data.dtype, **kwargs)
+
+        async def _copy_chunk(chunk_coords: ChunkCoords) -> None:
+            await new_array.setitem(chunk_coords, await data._async_array.getitem(chunk_coords))
+
+        # Stream data from the source array to the new array
+        await concurrent_map(
+            [(region,) for region in data._iter_chunk_regions()],
+            _copy_chunk,
+            config.get("async.concurrency"),
+        )
+        return new_array
 
     # ensure data is array-like
     if not hasattr(data, "shape") or not hasattr(data, "dtype"):
--- a/tests/test_array.py
+++ b/tests/test_array.py
@@ -2,6 +2,7 @@ import dataclasses
 import json
 import math
 import pickle
+import time
 from itertools import accumulate
 from typing import Any, Literal
 
@@ -863,6 +864,52 @@ async def test_special_complex_fill_values_roundtrip(fill_value: Any, expected: 
     assert content is not None
     actual = json.loads(content.to_bytes())
     assert actual["fill_value"] == expected
+
+
+async def test_creation_from_other_zarr(tmpdir):
+    src = zarr.zeros(
+        (2000, 20000), chunks=(1000, 1000), dtype="uint8", store=LocalStore(str(tmpdir))
+    )
+    src[:] = 1
+    for _i in range(10):
+        start_time = time.time()
+        c = zarr.array(src, store=MemoryStore())
+        end_time = time.time()
+        print(f"Time fast: {end_time - start_time} seconds")
+
+        start_time = time.time()
+        b = zarr.zeros(src.shape, chunks=src.chunks, store=MemoryStore())
+        b[:] = src[:]
+        end_time = time.time()
+        print(f"Time slow: {end_time - start_time} seconds")
+
+    assert b[123, 123] == 1
+    assert c[123, 123] == 1
+
+
+@pytest.mark.parametrize("shape", [(1,), (2, 3), (4, 5, 6)])
+@pytest.mark.parametrize("dtype", ["uint8", "float32"])
+@pytest.mark.parametrize("array_type", ["async", "sync"])
+async def test_nbytes(
+    shape: tuple[int, ...], dtype: str, array_type: Literal["async", "sync"]
+) -> None:
+    """
+    Test that the ``nbytes`` attribute of an Array or AsyncArray correctly reports the capacity of
+    the chunks of that array.
+    """
+    store = MemoryStore()
+    arr = Array.create(store=store, shape=shape, dtype=dtype, fill_value=0)
+    if array_type == "async":
+        assert arr._async_array.nbytes == np.prod(arr.shape) * arr.dtype.itemsize
+    else:
+        assert arr.nbytes == np.prod(arr.shape) * arr.dtype.itemsize
+
+
+async def test_scalar_array() -> None:
+    arr = zarr.array(1.5)
+    assert arr[...] == 1.5
+    assert arr[()] == 1.5
+    assert arr.shape == ()
 
 
 @pytest.mark.parametrize("shape", [(1,), (2, 3), (4, 5, 6)])


normanrz:
> Do we also want concurrency for different chunk sizes?

That would be nice, if the chunk sizes are somewhat compatible, i.e. one is a multiple of the other.

src/zarr/core/array.py
```diff
@@ -3734,6 +3735,174 @@ class ShardsConfigParam(TypedDict):
...
+        The array.
+
+    Examples
+    --------
+    #TODO
+    """
+
+    # fill missing arguments with metadata of data Array
+    if chunks == "auto":
+        chunks = data.chunks
```
d-v-b, line 3848: is the intention for this to work with numpy arrays? because they don't have a `chunks` attribute. by contrast, dask arrays _do_ have a chunks attribute, but it's a tuple of tuples of ints (because dask chunks can be irregularly sized). So maybe a bit more parsing is needed here.

brokkoli71: I was only thinking of zarr arrays as data. But I can generalize it for numpy and dask

d-v-b:
> * (Is there some measure to prevent this that I am not aware of?)

if you are trying to write `K` input chunks into `M` output chunks, you can partition your `K` chunks into sets, where within each set elements can be written independently from all the other elements. then you write each set one after another. in the worst case scenario there will be 1 set per chunk, but you are guaranteed to avoid write collisions this way.

src/zarr/core/array.py
normanrz: > But I can generalize it for numpy and dask

I think that would be great! In the tests, we could use something like `zarr.from_array(store=store, data=np.arange(10))` in many places!

d-v-b: a lot of my inspiration comes from trying to make tests easier to write :)

it might also make sense to put in a keyword argument that controls whether data is written or not. some users might want to only create the array, and write data to it later with a different method. I think the default should be to avoid IO (i.e., don't write).

normanrz: I would do it the other way around: add an optional metadata_only kwarg.
In any case, there would be IO to write the zarr.json.

d-v-b: yes creating an array will always involve writing metadata, but users with TB scale datasets generally create the array first, then schedule writing to the array in a separate step. IMO attempting to write TB of data by default is not scalable for large datasets, and we should design around that case.

normanrz: I know it is not scalable for large arrays. But from an `from_array` method, I would expect it to actually create the array with the data from the array. Opt out is fine. Maybe what you are proposing should be called something else. No strong opinion here, though.

d-v-b: I would expect `from_array` to create a zarr array. I would not expect it to fill the array with data, because I routinely use zarr with huge arrays. write-on-default would also be problematic for people with dask arrays, because zarr would not be in a position to decide which dask scheduler to use.

normanrz: Ok.

src/zarr/core/array.py
```diff
@@ -3734,6 +3735,174 @@ class ShardsConfigParam(TypedDict):
 ShardsLike: TypeAlias = ChunkCoords | ShardsConfigParam | Literal["auto"]
 
 
+async def from_array(
+    data: Array,
```
normanrz, line 3739:
```suggestion
    data: Array | npt.ArrayLike,
```
As discussed, this function should also work with numpy arrays.

src/zarr/core/array.py
d-v-b: to revisit this, it seems like everyone else does assume that the array should be written by default. so feel free to ignore my objections here :)

d-v-b:
one question to answer here is what "auto" means for chunks if the user passes in a chunked array, but they want to use `zarr-python's` auto-chunking instead of the chunks that came with the array. 

We might want to use a separate value that means "copy the chunks this object already has", which is distinct from "generate some chunks using the chunking heuristics". maybe something like `ChunksLike: Literal['auto'] | Literal['keep'] | ShapeLike`?

brokkoli71:
> one question to answer here is what "auto" means for chunks if the user passes in a chunked array, but they want to use `zarr-python's` auto-chunking instead of the chunks that came with the array.

Good point! I like the idea of distinguishing between `keep' and `auto'.

src/zarr/core/array.py
brokkoli71: Okay:) So, do we want an optional `metadata_only` argument or leave it out?

d-v-b: how about `write_data: bool = True`?

d-v-b:
now that https://github.com/zarr-developers/zarr-python/pull/2761 is in, could we use `from_array` inside `create_array` (after the data / dtype / shape validation)?

brokkoli71:
> now that #2761 is in, could we use `from_array` inside `create_array` (after the data / dtype / shape validation)?

@d-v-b Currently, `from_array` calls `create_array`. Is it redundant to have both `from_array` and `create_array` to create an array from another array? Or is there a benefit in having both?

src/zarr/api/synchronous.py
```diff
@@ -895,6 +902,212 @@ def create_array(
...
+        )
+    )
+
+
+def from_array(
+    data: Array | npt.ArrayLike,
+    store: str | StoreLike,
+    write_data: bool = True,
+    *,
+    name: str | None = None,
```
d-v-b, line 916:
```suggestion
def from_array(
    store: str | StoreLike,
    *,
    data: Array | npt.ArrayLike,
    write_data: bool = True,
    name: str | None = None,
```

lets make everything except `store` keyword-only, to match `create_array`

src/zarr/api/synchronous.py
```diff
@@ -895,6 +902,212 @@ def create_array(
...
+    chunks: Literal["auto", "keep"] | ChunkCoords = "keep",
+    shards: ShardsLike | None = None,
+    filters: FiltersLike | Literal["keep"] = "keep",
+    compressors: CompressorsLike | Literal["keep"] = "keep",
+    serializer: SerializerLike | Literal["keep"] = "keep",
+    fill_value: Any | None = None,
+    order: MemoryOrder | None = None,
+    zarr_format: ZarrFormat | None = None,
+    attributes: dict[str, JSON] | None = None,
+    chunk_key_encoding: ChunkKeyEncoding | ChunkKeyEncodingLike | None = None,
```
d-v-b, line 926:
```suggestion
    chunk_key_encoding: ChunkKeyEncodingLike | None = None,
```
`ChunkKeyEncodingLike` includes `ChunkKeyEncoding`

src/zarr/api/synchronous.py
```diff
@@ -895,6 +902,212 @@ def create_array(
...
+
+
+def from_array(
+    data: Array | npt.ArrayLike,
+    store: str | StoreLike,
+    write_data: bool = True,
+    *,
+    name: str | None = None,
+    chunks: Literal["auto", "keep"] | ChunkCoords = "keep",
+    shards: ShardsLike | None = None,
```
d-v-b, line 918:
```suggestion
    shards: ShardsLike | None = None,
```

should this also be `"keep"`?

src/zarr/core/array.py
```diff
@@ -3793,6 +3794,279 @@ class ShardsConfigParam(TypedDict):
 ShardsLike: TypeAlias = ChunkCoords | ShardsConfigParam | Literal["auto"]
 
 
+async def from_array(
+    data: Array | npt.ArrayLike,
+    store: str | StoreLike,
+    write_data: bool = True,
+    *,
```
d-v-b, line 3801: function signature should match `create_array`

src/zarr/core/array.py
```diff
@@ -3793,6 +3794,279 @@ class ShardsConfigParam(TypedDict):
...
+    chunks: Literal["auto", "keep"] | ChunkCoords = "keep",
+    shards: ShardsLike | None = None,
+    filters: FiltersLike | Literal["keep"] = "keep",
+    compressors: CompressorsLike | Literal["keep"] = "keep",
+    serializer: SerializerLike | Literal["keep"] = "keep",
+    fill_value: Any | None = None,
+    order: MemoryOrder | None = None,
+    zarr_format: ZarrFormat | None = None,
+    attributes: dict[str, JSON] | None = None,
+    chunk_key_encoding: ChunkKeyEncoding | ChunkKeyEncodingLike | None = None,
```
d-v-b, line 3812:
```suggestion
    chunk_key_encoding: ChunkKeyEncodingLike | None = None,
```

src/zarr/core/array.py
```diff
@@ -3793,6 +3794,279 @@ class ShardsConfigParam(TypedDict):
...
+        if chunks == "keep":
+            chunks = "auto"
+        if zarr_format is None:
+            zarr_format = 3
+        if filters == "keep":
+            filters = "auto"
+        if compressors == "keep":
+            compressors = "auto"
+        if serializer == "keep":
+            serializer = "auto"
```
d-v-b, line 4020: can we put all of this logic into a stand-alone function, that can be tested without creating a new array?

src/zarr/core/array.py
```diff
@@ -3793,6 +3794,279 @@ class ShardsConfigParam(TypedDict):
...
+    if not hasattr(data, "dtype") or not hasattr(data, "shape"):
+        data = np.array(data)
```
d-v-b, line 4022: if `data` can be turned into a numpy array, then shouldn't we use its dtype or shape?

tests/conftest.py
```diff
@@ -75,6 +75,13 @@ async def store(request: pytest.FixtureRequest, tmpdir: LEGACY_PATH) -> Store:
     return await parse_store(param, str(tmpdir))
 
 
+@pytest.fixture
+async def store2(request: pytest.FixtureRequest, tmpdir: LEGACY_PATH) -> Store:
```
d-v-b, line 79: what is this fixture for?

d-v-b: maybe add a docstring explaining why it exists

brokkoli71: for creating tests with two arrays with different types of stores. I use it in ``test_from_array``. Maybe this can be moved to ``test_array.py``

src/zarr/core/array.py
brokkoli71: @d-v-b I am not sure if I understand what you mean. we will use its `dtype` and `shape` afterwards in 
``` python
    meta = await init_array(
        store_path=store_path,
        shape=data.shape,
        dtype=data.dtype,
```
Or do you mean use `dtype` or `shape` even if only one of `dtype` or `shape` is an attribute of `data`, such as
```python
    if not hasattr(data, "shape"):
        if not hasattr(data, "dtype"):
            data = np.array(data)
        else:
            data = np.array(data, dtype=data.dtype)
```

d-v-b: no I was simply confused, I forgot that dtype and shape would be used later.

src/zarr/core/array.py
brokkoli71: would you put this all into one stand-alone function such as
```python
    (
        data,
        chunks,
        shards,
        filters,
        compressors,
        serializer,
        fill_value,
        order,
        zarr_format,
        chunk_key_encoding,
        dimension_names,
    ) = _parse_keep_array_attr(
        data=data,
        chunks=chunks,
        shards=shards,
        filters=filters,
        compressors=compressors,
        serializer=serializer,
        fill_value=fill_value,
        order=order,
        zarr_format=zarr_format,
        chunk_key_encoding=chunk_key_encoding,
        dimension_names=dimension_names,
    )
```
or split it into separate functions e.g. `_parse_keep_shards` ?

d-v-b: I think it can be 1 function that takes an existing array + other keyword arguments, and returns the same  keyword arguments but parsed to concrete values based on that array.

normanrz (approved):
I think this is ready. @d-v-b what do you think?

d-v-b (approved):
looks good, thanks @brokkoli71

PR #2623, opened 2025-01-02
Use doctests for guide

dstansby:
This reverts using IPython for code examples in https://github.com/zarr-developers/zarr-python/pull/2589, and is targeted to that branch (not main). See https://github.com/zarr-developers/zarr-python/pull/2589#pullrequestreview-2525360742 for the motivation for not switching to IPython, but sticking with the doctest status-quo

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -49,8 +49,6 @@ extensions = [
     "sphinx_copybutton",
     "sphinx_design",
     'sphinx_reredirects',
-    "IPython.sphinxext.ipython_directive",
-    "IPython.sphinxext.ipython_console_highlighting",
 ]
 
 issues_github_path = "zarr-developers/zarr-python"
--- a/docs/user-guide/arrays.rst
+++ b/docs/user-guide/arrays.rst
@@ -3,30 +3,22 @@
 Working with arrays
 ===================
 
-.. ipython:: python
-   :suppress:
-
-   rm -r data/
+   >>> import shutil
+   >>>
+   >>> shutil.rmtree("./data")
 
 Creating an array
 -----------------
 
-Zarr has several functions for creating arrays. For example:
+Zarr has several functions for creating arrays. For example::
 
-.. ipython:: python
-
-   import zarr
-
-   store = {}
-   # TODO: replace with `create_array` after #2463
-   z = zarr.create(
-       store=store,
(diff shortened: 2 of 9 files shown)


jhamman:
Thanks for doing this @dstansby - how can we run the doc tests? Are they run automatically when you build the docs or in CI?

dstansby:
They can be turned on to run with pytest in the normal test runs. Do you think setting that up should block v3?

dstansby:
doctests are being run on v2 (see https://github.com/zarr-developers/zarr-python/pull/2552 where I had to fix some recently), so there should be a pattern to copy there to enable them for v3.

jhamman:
okay, I think we should do that here before merging. The main reason I went with Ipython was to guarantee that the docs were reflecting exactly what Zarr produces. So we don't want to loose that.

dstansby:
I had a quick try, but it's not trivial (but probably not complicated either) to enable doctests. To keep things simple here I'd advocate enabling doctests as out of scope for this PR, and we can do it in a follow up to all the other PRs. FWIW I have run all the doctests locally to generate the output, so it should be correct at the moment.

jhamman:
@dstansby - can you describe the issues with getting the doc tests passing here? Since we're reverting work that ensures the doc can be run, I'd like to, at a minimum, understand why this route isn't working.

dstansby:
> Since we're reverting work that ensures the doc can be run

I don't think this is stopping the doc build from working? The readthedocs failures in this PR are the same as in the branch it's based on (ie https://github.com/zarr-developers/zarr-python/pull/2589)

I'm happy to either enable doctests in this PR, or in a follow up PR to https://github.com/zarr-developers/zarr-python/pull/2589 as you prefer. I thought the preference was to merge https://github.com/zarr-developers/zarr-python/pull/2589 ASAP, so I didn't want to hold up this (and therefore https://github.com/zarr-developers/zarr-python/pull/2589) by working out the doctests, but I can do that (but maybe not today) if you think that should block https://github.com/zarr-developers/zarr-python/pull/2589?

jhamman:
Let's enable doctests in this PR. Otherwise, we're going to immediately break things in the docs and not know it.

dstansby:
👍 doctests enabled in this PR now

PR #2626, opened 2025-01-02
Enable doctests

dstansby:
This should enable doctests to be run as part of the normal test suite. They will almost certainly fail until https://github.com/zarr-developers/zarr-python/pull/2623 is merged, but I'm opening a new PR targeted to `main` to make sure it works without having to alter the GH actions config.

If this works, I'll add it to https://github.com/zarr-developers/zarr-python/pull/2623, so opening as draft because this shouldn't be merged.


dstansby:
Woop, this is failing as expected!

PR #2634, opened 2025-01-03
Clean up release notes in preparation for v3

dstansby:
This cleans out the release notes page, in preparation for v3. I decided to:

- Remove all the previous release notes for v1 and v2. These will get out of sync with new minor releases of v2, so I thought better to remove them and link out to the v2 branch for these.
- Remove the existing v3 release notes. I think these dried up at some point, so instead of having an imcomplete set I thought better to remove them and link out to the v3 migration guide, and then start a normal release note service once v3 is released.

Thoughts?


normanrz:
Sounds like a good idea. I also stumbled across this issue because some old release notes link to sections in the docs that now describe very different behavior.

jhamman:
+1 on doing this. Can we get this in tomorrow?

docs/release-notes.rst
```diff
@@ -0,0 +1,16 @@
+Release notes
+=============
+
```
jhamman, line 3: Let's seed the release note section for 3.0.1 

```
.. release_3.0.1:

3.0.1 (unreleased)
------------------

...
```

jhamman (approved):
One small request but otherwise looks good.

docs/release-notes.rst
dstansby: We shouldn't have the 3.0.1 title in the 3.0.0 release, so shouldn't we hold off here until after release?

jhamman: We typically seed the release notes with the next release so contributors have a spot to start adding release notes. We can do this right after the 3.0.0 release I suppose.

normanrz: In numcodecs, we also seed a section but without version number. https://github.com/zarr-developers/numcodecs/blob/main/docs/release.rst?plain=1#L16

dstansby: That's after a release is made, the notes for the current release don't have an unreleased section: https://numcodecs.readthedocs.io/en/stable/release.html

jhamman (approved):

PR #2635, opened 2025-01-03
Improve API reference doc structure

dstansby:
This improves the structure of the API reference by:

- Generating the pages under `api/`, instead of redirecting from an index page under `api/ `to `_autoapi/`
- Adding a redirect from the v2 API page to the new page
- Before you had to click on "API reference", then "zarr" on the following (mostly empty) page to get to the ref. Now "API reference" links directly to the API reference page.


docs/index.rst
```diff
@@ -11,7 +11,7 @@ Zarr-Python
     getting_started
     about
     user-guide/index
-    api/index
+    API Reference <api/zarr/index>
```
jhamman, line 14:
```suggestion
    API reference <api/zarr/index>
```
To match the other titles

jhamman:
Great improvement. Happy to approve once this is ready for review.

docs/index.rst
dstansby: 👍 I updated this while rebasing

jhamman (approved):

Issue #2621, opened 2025-01-02
Make zarr.core private


jhamman:
We seem to be converging on using a leading underscore to indicate private modules. Should we do this for `zarr.core` or can we simply document the intent of the module?

dstansby:
Is it intended to be private? If so, I'd definitely add a leading underscore to make it as private.

jhamman:
Yes, the intent is to make it private.

dstansby:
I had a quick go, and there seem to be bits that need to be public (at least `zarr.core.config`, `zarr.core.buffer.Buffer`). Making the whole of core private is going to need someone who knows what should be public (definitely not me 😄) going through carefully and decided what should be public and what should be private.

jhamman:
I can take this on.

dstansby:
As some small first steps I cleaned up `zarr.core.buffer` (https://github.com/zarr-developers/zarr-python/pull/2641) and `zarr.core.metadata` (https://github.com/zarr-developers/zarr-python/pull/2642)

jhamman:
@dstansby - what makes you think `zarr.core.config` needs to be public? The intent, as I see it, is just that `zarr.core.config.config` is public which is why it is available as `zarr.config`.

@normanrz - do you think we could move `zarr.core.buffer` to `zarr.buffer`? The [`Buffer` ABC](https://github.com/zarr-developers/zarr-python/blob/ec014f6e7e184f4ce0ebfb712c50b0b592bf9057/src/zarr/core/buffer/core.py#L118) seems like it actually belongs in the `abc` module.

dstansby:
>  what makes you think zarr.core.config needs to be public? 

There's user facing documentation in the docstring, but maybe that should now be deleted in favour of the user guide? It looks like `BadConfigError` isn't used so can be removed, but I think `Config` should be user facing as it's the type of `zarr.config`, so maybe that could move to `zarr.Config`?

normanrz:
> [@normanrz](https://github.com/normanrz) - do you think we could move `zarr.core.buffer` to `zarr.buffer`? The [`Buffer` ABC](https://github.com/zarr-developers/zarr-python/blob/ec014f6e7e184f4ce0ebfb712c50b0b592bf9057/src/zarr/core/buffer/core.py#L118) seems like it actually belongs in the `abc` module.

Yes, we could do that. And yes, the `Buffer` and `NDBuffer` make sense in the `abc` module.

jhamman:
[...]

PR #2641, opened 2025-01-04
Clean up `zarr.core.buffer` API surface

dstansby:
This cleans up the `zarr.core.buffer` API, by making `zarr.core.buffer._core` private. Everything was already imported into `zarr.core.buffer`, but imports elsewhere in the library just needed changing.

### DIFF
--- a/src/zarr/abc/store.py
+++ b/src/zarr/abc/store.py
@@ -5,7 +5,7 @@ from asyncio import gather
 from itertools import starmap
 from typing import TYPE_CHECKING, Protocol, runtime_checkable
 
-from zarr.core.buffer.core import default_buffer_prototype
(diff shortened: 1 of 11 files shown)


normanrz (approved):

jhamman:
We should make sure we don't end up with two levels of underscores. My preference would be we do the full `zarr._core` switch rather than the sub-sub modules. Before moving forward with these, let's address the broader question of which parts of the `zarr.core` module need to be available as public API (#2621).

jhamman:
@dstansby - do you have thoughts on how to proceed here given #2669?

dstansby:
Yeah, doesn't need to block 3.0.0, and needs updating, but would be good to keep it open to remind me to update it at some point

PR #2642, opened 2025-01-04
Clean up `zarr.core.metadata` API

dstansby:
Similar to other sub-modules in `zarr.core`, `zarr.core.metadata` imports some stuff into it's top level namespace. To stay consistent, I've:

- Imported `DataType` here, because it's used in the code reprs.
- Made `zarr.core.metadata.{common, v2, v3}` private

An alternative here would be to keep `zarr.core.metadata.{common, v2, v3}`, and move the documentation of [classes currently defined in zarr.core.codecs](https://zarr.readthedocs.io/en/latest/_autoapi/zarr/core/metadata/index.html) back to their submodules. Thoughts?


normanrz (approved):
I like it!

jhamman:
@dstansby - do you have thoughts on how to proceed here given https://github.com/zarr-developers/zarr-python/pull/2669?

dstansby:
👍 doesn't need to block 3.0.0, and needs updating, but would be good to keep it open to remind me to update it at some point

PR #2644, opened 2025-01-04
Adds documentation for sharding and sharding in Array.info

### LINKED ISSUE #2643: Array.info should print shard sizes
The `Array.info(_complete)` method should print `Shard shape` for sharded arrays. The current output (see below), where the shard shape is printed as "Chunk shape", may be confusing to users.

```python
  >>> a = zarr.create_array('data/example-20.zarr', shape=(10000, 10000), shards=(1000, 1000), chunks=(100, 100), dtype='uint8')
  >>> a[:] = (np.arange(10000 * 10000) % 256).astype('uint8').reshape(10000, 10000)
  >>> a.info_complete()
   Type               : Array
   Zarr format        : 3
   Data type          : DataType.uint8
   Shape              : (10000, 10000)
   Chunk shape        : (1000, 1000)
   Order              : C
   Read-only          : False
   Store type         : LocalStore
   Codecs             : [{'chunk_shape': (100, 100), 'codecs': ({'endian': <Endian.little: 'little'>}, {'level': 0, 'checksum': False}), 'index_codecs': ({'endian': <Endian.little: 'little'>}, {}), 'index_location': <ShardingCodecIndexLocation.end: 'end'>}]
   No. bytes          : 100000000 (95.4M)
   No. bytes stored   : 3981060
   Storage ratio      : 25.1
   Chunks Initialized : 100
```

normanrz:
- Adds a section about sharding to the array guide
- Adds "shard shape" to `Array.info`, fixes #2643
- Fixed a bug where writing scalars to sharded arrays didn't work


docs/user-guide/arrays.rst
```diff
@@ -574,8 +574,40 @@ Any combination of integer and slice can be used for block indexing::
...
 
-Coming soon.
-
+Using small chunk shapes in very large arrays can lead to a very large number of chunks.
+This can become a performance issue for file systems and object storage.
+With Zarr format 3, a new sharding feature has been added to address this issue.
+
+With sharding, multiple chunks can be stored in a single storage object (e.g. a file).
+Within a shard, chunks are compressed and serialized separately.
+This allows to read individual chunks independently.
```
dstansby, line 583:
```suggestion
This allows individual chunks to be read independently.
```

docs/user-guide/arrays.rst
```diff
@@ -574,8 +574,40 @@ Any combination of integer and slice can be used for block indexing::
...
+However, when writing data, a full shard must be written for optimal performance and to
```
dstansby, line 584:
```suggestion
However, when writing data, a full shard must be written in one go for optimal performance and to
```

docs/user-guide/arrays.rst
```diff
@@ -574,8 +574,40 @@ Any combination of integer and slice can be used for block indexing::
...
+  Read-only          : False
+  Store type         : LocalStore
+  Codecs             : [{'chunk_shape': (100, 100), 'codecs': ({'endian': <Endian.little: 'little'>}, {'level': 0, 'checksum': False}), 'index_codecs': ({'endian': <Endian.little: 'little'>}, {}), 'index_location': <ShardingCodecIndexLocation.end: 'end'>}]
+  No. bytes          : 100000000 (95.4M)
+  No. bytes stored   : 3981060
+  Storage ratio      : 25.1
+  Chunks Initialized : 100
+
+In this example a shard shape of (1000, 1000) and a chunk shape of (100, 100) is used.
+This means that 10*10 chunks are stored in each shard.
```
dstansby, line 610: Could you add a sentence explaining what happens if the shards parameter isn't provided? Do we get one shard? Or something else?

dstansby, line 610:
```suggestion
This means that 10*10 chunks are stored in each shard, and there are 10*10 shards in total.
```

docs/user-guide/performance.rst
```diff
@@ -62,6 +62,43 @@ will be one single chunk for the array::
...
    (10000, 10000)
 
+
+Sharding
+~~~~~~~~
+
+If you have large arrays but need small chunks to efficiently access the data, you can
+use sharding. Sharding provides a mechanism to store multiple chunks in a single
+storage object or file. This can be useful because traditional file systems and object
+storage systems may have issues with many small files.
```
dstansby, line 72:
```suggestion
storage systems may have issues storing and accessing many files.
```
I dropped "small" because I think the issue is number of files, and not there size? I may be wrong there though...

docs/user-guide/performance.rst
```diff
@@ -62,6 +62,43 @@ will be one single chunk for the array::
...
+
+Picking a good combination of chunk shape and shard shape is important for performance.
+The chunk shape determines what unit of your data can be read independently, while the
+shard shape determines what unit of your data can be written efficiently.
+
+For an example, consider you have a 100 GB array and need to read small chunks of 1 MB.
+Without sharding, each chunk would be one file resulting in 10000 files. That can
```
dstansby, line 79:
```suggestion
Without sharding, each chunk would be one file resulting in 100,000 files. That can
```

dstansby (approved):
A couple of minor suggestions. I wasn't too confident reviewing the code, so would be worth a second check of those bits by someone else.

docs/user-guide/performance.rst
normanrz: Correct. Having small files can be inefficient, e.g. when the fs block size is larger than the file. But that is orthogonal to the performance.

jhamman (approved):
Thanks @normanrz - much better than I would have done :)

PR #2650, opened 2025-01-05
Speed up hypothesis tests

dstansby:
- `tests/test_store/test_stateful.py::test_zarr_hierarchy[memory]`
- `tests/test_store/test_stateful.py::test_zarr_store[memory]`

currently take up around 70% of our total test time (e.g., [see this recent test run](https://github.com/zarr-developers/zarr-python/actions/runs/12618843971/job/35162691283#step:6:158)).

Since these tests are not comparing two different store implementations, but the same one (comparing `MemoryStore` to `MemoryStore`), I think it makes sense to cut the number of examples right down here to save a **lot** of testing time.


dstansby:
What's the minimum number of examples to be useful?

My thinking was checking MemoryStore against MemoryStore was useful to make sure the test itself works, but it's just comparing instances of identical classes, so there's no benefit to running lots of tests, especially as it's taking up 70% of the total test time. Maybe I'm misunderstanding the tests though?

I'd be 👍 to doing a nightly action if there is good reason to run lots of examples for these tests.

dstansby:
How about cuting down to 50 examples then? Even that would give us a big and quick win in terms of test times. Then in the medium term we could look at 1) a nighly run of the stateful tests, and 2) testing LocalStore against MemoryStore.

dcherian:
OK sounds good.

tests/test_store/test_stateful.py
```diff
@@ -28,6 +28,11 @@ def mk_test_instance_sync() -> None:
 
     if isinstance(sync_store, ZipStore):
         pytest.skip(reason="ZipStore does not support delete")
-    if isinstance(sync_store, LocalStore):
+    elif isinstance(sync_store, LocalStore):
         pytest.skip(reason="This test has errors")
-    run_state_machine_as_test(mk_test_instance_sync, settings=Settings(report_multiple_bugs=True))
+    elif isinstance(sync_store, MemoryStore):
+        run_state_machine_as_test(mk_test_instance_sync, settings=Settings(max_examples=5))
```
dcherian, line 34:
```suggestion
        run_state_machine_as_test(mk_test_instance_sync, settings=Settings(max_examples=50))
```

dstansby: Woops, thanks for catching

dcherian (approved):

dcherian:
Sorry I had this all backwards. I have already done the nightly action: https://github.com/zarr-developers/zarr-python/actions/workflows/hypothesis.yaml

I think the error here is in running the hypothesis tests in all test commands instead of just that action. Are you able to take a look at that.

For example on this PR we run the tests [separately](https://github.com/zarr-developers/zarr-python/actions/runs/12638571454/job/35215120708?pr=2650) too.

dstansby:
I think it's worth merging this as a "ready to go" improvement on testing times - if the hypothesis tests aren't meant to be run on every normal test run, could you open an issue to track fixing that?

PR #2651, opened 2025-01-05
Don't draw invalid shapes in `test_vindex`

dstansby:
Quite a lot of test cases were not passing the `assume()` here. This makes sure generated data doesn't have a size-0 dimension to start with, which should speed up the test because it's not wasting time generating data that's immediately thrown away.


dcherian (approved):
nice catch!

dstansby:
pre-commit.ci autofix

Issue #2649, opened 2025-01-04
Array.info does not show the name of codecs


jhamman:
### Zarr version

3.0.0-rc.1

### Numcodecs version

0.14.1

### Python Version

3.12

### Operating System

Linux

### Installation

pip

### Description

Looking at the output from `Array.info`, I'm noticing that we are not showing the name of the codecs. This seems like an important bit of information that should be part of the output of `info`.

### Steps to reproduce

```
import zarr
import numpy as np

# Create a 2D Zarr array
z = zarr.create_array(
   store="data/example-1.zarr",
   shape=(100, 100),
   chunks=(10, 10),
   dtype="f4"
)

# Assign data to the array
z[:, :] = np.random.random((100, 100))
z.info
```
```
Type               : Array
Zarr format        : 3
Data type          : DataType.float32
Shape              : (100, 100)
Chunk shape        : (10, 10)
Order              : C
Read-only          : False
Store type         : LocalStore
Codecs             : [{'endian': <Endian.little: 'little'>}, {'level': 0, 'checksum': False}]
No. bytes          : 40000 (39.1K)
```

d-v-b:
will users creating arrays with `create_array` know what `codecs` means? see https://github.com/zarr-developers/zarr-python/issues/2633

jhamman:
Perhaps not, but we can teach them. But regardless of how we present it, we should include the name of the thing (e.g. zstd, blosc, etc.).

normanrz:
Fixed in #2652: https://github.com/zarr-developers/zarr-python/pull/2652/files#diff-4f1ac2418ce24c1097c753a26ca68452082002115a6d992405aa94aeb1ca60c5R112-R113

PR #2652, opened 2025-01-05
Adds filters, compressors and serializer props to Array

### LINKED ISSUE #2633: add `compressors`, `filters`, `serializer` to `Array` as attributes
#2463 added a new function for creating arrays that abstracts over the painful fact that v2 arrays and v3 arrays have overlapping-but-distinct specifications of how chunks are encoded. This function expresses the v2 and v3 chunk encoding specs via 3 keyword arguments: 
- `filters` (`filters` in v2 , the `array_array_codecs` elements of the `codecs` list attribute in v3)
- `serializer` (implicit in v2, the `array_bytes_codec` element of the `codecs` list attribute in v3)
- `compressors` (`compressor` in v2 (note the singular), `bytes_bytes_codec` elements of the `codecs` list attribute in v3)

logically, these terms describe "transformations made to chunks when they are n-dimensional arrays", "the transformation that turns an n-dimensional array into a flat bytestream", and "transformations made to chunks after they are flattened to bytestreams"

but we don't expose `filters`, `serializer`, `compressors` as attributes on the `Array` class, which is probably where users will expect to find them. So I'm proposing that we add `filters`, `serializer`, `compressors` to the `Array` class.

normanrz:
Adds `filters`, `compressors`, `serializer` properties to the `Array` and `AsyncArray` classes. These properties mirror the kwargs that users provide in `create_array`, i.e. inner codecs in sharding are surfaced. `Array.info` is adapted accordingly. Fixes #2633 #2649.

I also fixed a bug where strings/bytes dtypes couldn't be used with sharding due to validation errors.

TODO:


docs/user-guide/arrays.rst
```diff
@@ -236,23 +238,24 @@ built-in delta filter::
    >>> compressors = LZMA(filters=lzma_filters)
    >>> data = np.arange(100000000, dtype='int32').reshape(10000, 10000)
    >>> z = zarr.create_array(store='data/example-7.zarr', shape=data.shape, dtype=data.dtype, chunks=(1000, 1000), compressors=compressors)
-   >>> z.metadata.codecs
-   [BytesCodec(endian=<Endian.little: 'little'>), _make_bytes_bytes_codec.<locals>._Codec(codec_name='numcodecs.lzma', codec_config={'id': 'lzma', 'filters': [{'id': 3, 'dist': 4}, {'id': 33, 'preset': 1}]})]
+   >>> z.compressors
+   (_make_bytes_bytes_codec.<locals>._Codec(codec_name='numcodecs.lzma', codec_config={'id': 'lzma', 'filters': [{'id': 3, 'dist': 4}, {'id': 33, 'preset': 1}]}),)
```
d-v-b, line 242: out of scope for this PR, but this string serialization looks weird

normanrz: Will be solved with the next numcodecs release: https://github.com/zarr-developers/numcodecs/pull/675

src/zarr/core/_info.py
```diff
@@ -109,18 +109,19 @@ def __repr__(self) -> str:
...
+        kwargs = {field.name: getattr(self, field.name) for field in dataclasses.fields(self)}
+
         if self._chunk_shape is None:
             # for non-regular chunk grids
             kwargs["chunk_shape"] = "<variable>"
-        if self._compressor is not None:
-            template += "\nCompressor         : {_compressor}"
 
-        if self._filters is not None:
+        if len(self._filters) > 0:
```
d-v-b, line 119: Will this prevent arrays with no filters at all from displaying the fact that they have no filters via the `.info` method?

d-v-b: likely out of scope, since my question is really about the pre-existing behavior of this function

d-v-b (approved):
happy to see some simplified tests via these changes!

src/zarr/core/_info.py
normanrz: The "Filters     : ... " line wasn't shown if `filters` was `None`, but it was shown if `filters` was `()`. Now, it is not shown if `None` or `()`.

d-v-b: Don't we want to explicitly show the value of `filters`, even if it's `()`?

normanrz: We could. I don't have a strong preference here.

d-v-b: IMO it's better to show a static set of properties with this function, but if that's annoying to implement here we can defer for a separate PR

normanrz: What about "Serializer", which doesn't exist for v2 arrays?

d-v-b: maybe `None` there? in principle there's nothing stopping someone from implementing a v2-compatible `BytesCodec` in numcodecs, in which case we might end up with v2 arrays with a `serializer`

normanrz: Well, there is no place in the v2 metadata for a serializer.
I opted for not showing the serializer field in the Array.info.

PR #2653, opened 2025-01-05
Separate defaults for filters, serializers and compressors in v3

normanrz:
Instead of a single `v3_default_codecs` key in the config, this PR adds 3 separate keys for filters, serializers and compressors.

Ideally, we could get this in the 3.0 release, because it would require some backwards-compat code if we change it later.

Builds upon #2652, but could be made independent. 

TODO:


docs/user-guide/config.rst
```diff
@@ -54,19 +54,20 @@ This is the current default configuration::
...
-                                    'numeric': [{'configuration': {'endian': 'little'},
-                                                 'name': 'bytes'},
-                                                {'configuration': {'checksum': False,
-                                                                   'level': 0},
-                                                 'name': 'zstd'}],
-                                    'string': [{'name': 'vlen-utf8'},
-                                               {'configuration': {'checksum': False,
-                                                                  'level': 0},
-                                                'name': 'zstd'}]},
+              'v3_default_compressors': {'bytes': [{'configuration': {'checksum': False,
```
d-v-b, line 57: is there a simple way to ensure that this representation and the actual config don't go out of sync? not a big deal if no.

src/zarr/core/array.py
```diff
@@ -4006,31 +3998,34 @@ def _get_default_chunk_encoding_v3(
...
-            array_array.append(codec)
-        elif isinstance(codec, BytesBytesCodec):
-            bytes_bytes.append(codec)
-        else:
-            raise TypeError(f"Unexpected codec type: {type(codec)}")
+    filters_list: list[ArrayArrayCodec] = []
+    compressors_list: list[BytesBytesCodec] = []
 
-    if array_bytes_maybe is None:
+    serializer = _resolve_codec(default_serializer)
```
d-v-b, line 4008: can we use [`_parse_bytes_bytes_codec`](https://github.com/zarr-developers/zarr-python/blob/617e2cd84b28c4ca28b8b221abd77718561556bb/src/zarr/registry.py#L169-L186) here? (and similarly for the other codec varieties)

docs/user-guide/config.rst
normanrz: Yes, via doctests https://github.com/zarr-developers/zarr-python/blob/feat/separate-default-codecs/docs/user-guide/config.rst?plain=1#L43

src/zarr/core/array.py
normanrz: Good idea. That removes some lines of code.

src/zarr/core/array.py
```diff
@@ -4148,6 +4109,22 @@ def _parse_chunk_encoding_v3(
...
+    if compressors is None:
+        out_bytes_bytes = ()
+
+    elif compressors == "auto":
+        out_bytes_bytes = default_bytes_bytes
+
+    else:
+        if isinstance(compressors, dict | Codec):
+            maybe_bytes_bytes = (compressors,)
+        elif compressors is None:
```
d-v-b, line 4121: Is it possible for this to be true? don't we already check if `compressors` is `None` on L4112?

normanrz: No, that is not possible. Fixed.

src/zarr/api/synchronous.py
```diff
@@ -804,24 +803,24 @@ def create_array(
         List of compressors to apply to the array. Compressors are applied in order, and after any
         filters are applied (if any are specified) and the data is serialized into bytes.
 
-        For Zarr format 3, a "compressor" is a codec that takes a bytestream, and
+        For Zarr format 3, a "compressor" is a codec that takes a bytestrea, and
```
d-v-b, line 806:
```suggestion
        For Zarr format 3, a "compressor" is a codec that takes a bytestream, and
```

src/zarr/core/group.py
```diff
@@ -1067,26 +1066,26 @@ async def create_array(
             List of compressors to apply to the array. Compressors are applied in order, and after any
             filters are applied (if any are specified) and the data is serialized into bytes.
 
-            For Zarr format 3, a "compressor" is a codec that takes a bytestream, and
+            For Zarr format 3, a "compressor" is a codec that takes a bytestrea, and
```
d-v-b, line 1069:
```suggestion
            For Zarr format 3, a "compressor" is a codec that takes a bytestream, and
```

src/zarr/core/group.py
```diff
@@ -2323,26 +2321,26 @@ def create_array(
...
```
d-v-b, line 2324:
```suggestion
            For Zarr format 3, a "compressor" is a codec that takes a bytestream, and
```

src/zarr/core/group.py
```diff
@@ -2712,26 +2709,26 @@ def array(
...
```
d-v-b, line 2712:
```suggestion
            For Zarr format 3, a "compressor" is a codec that takes a bytestream, and
```

src/zarr/core/group.py
normanrz: This resurfaced due to merge conflicts that I resolved.

PR #2664, opened 2025-01-07
api: make buffer public api

jhamman:
This PR moves `zarr.core.buffer` to `zarr.abc.buffer` and `zarr.buffer`, paving the way to make the rest of `zarr.core` private in a follow up PR.

Towards #2621 

TODO:


src/zarr/core/buffer/__init__.py
```diff
@@ -1,19 +1,10 @@
...
-    "NDBuffer",
-    "default_buffer_prototype",
-    "numpy_buffer_prototype",
-]
+warnings.warn(
+    "The zarr.core.buffer module is deprecated and will be removed in a future version. "
+    "Please use zarr.buffer and zarr.abc.buffer modules instead.",
+    FutureWarning,
+    stacklevel=2,
+)
```
jhamman, line 10: put this in to prompt a discussion -- numcodecs depends on this module already... which sucks.

@normanrz - do you have thoughts on how to proceed? Options are:

1. keep `zarr.core` but document it as private. We've managed to do this in the Xarray project with few issues.
2. change the [import in `numcodecs`](https://github.com/zarr-developers/numcodecs/blob/6c0ea0f43788f22e4ad5d6655862c6c212a0dece/numcodecs/zarr3.py#L51-L54) and make a new release

dstansby: We can easily skip documenting some 'public' (without a `_`) modules, which would probably help: https://github.com/zarr-developers/zarr-python/blob/f9c20243d207835d15d7091e05e8ec82a265b7d1/docs/conf.py#L64-L76

normanrz: I would also lean towards 1. Changing numcodecs would almost certainly also require some backwards-compat shim in zarr-python.

PR #2665, opened 2025-01-07
feat/batch creation

d-v-b:
This PR adds a few routines for creating a collection of arrays and groups (i.e., a dict with path-like keys and `ArrayMetadata` / `GroupMetadata` values) in storage concurrently. 

-  `create_hierarchy` takes a dict representation of a hierarchy, parses that dict to ensure that there are no implicit groups (creating group metadata documents as needed), then invokes `create_nodes` and yields the results
- `create_nodes` concurrently writes metadata documents to storage, and yields the created `AsyncArray` / `AsyncGroup` instances.

I still need to wire up concurrency limits, and test them.

TODO:


d-v-b:
this is now working, so I would appreciate some feedback on the design. 

The basic design is the same as what I outlined earlier in this PR: there are two new functions that take a `dict[path, GroupMetadata | ArrayMetadata]` like `{'a': GroupMetadata(zarr_format=3), 'a/b': ArrayMetadata(...)}` and concurrently persist those metadata documents to storage, resulting in a hierarchy on disk that looks like the dict.

### approach
basically the same as concurrent group members listing, except we don't need any recursion. I'm scheduling writes and using `as_completed` to yield Arrays / Groups when they are available.

### new functions
- `create_nodes` is low-level and doesn't do any checking of its input, so it will happily create invalid hierarchies, e.g. nesting groups inside arrays, or mixing v2 and v3 metadata, and it won't create intermediate groups, either.

- `create_hierarchy` is higher level, it parses the input, checking it for invalid hierarchies, and inserting implicit groups as needed.

-`Group.create_hierarchy` is a new method on the `Group` / `AsyncGroup` classes that takes a hierarchy dict and creates the nodes specified in that dict at locations relative to the path of the group instance. the return value is `dict[str, AsyncGroup | AsyncArray]`, but I guess it also doesn't have tor return anything, or it could be an async iterator, so that you can interact with the nodes as they are formed. This is flexible right now, but I think the iterator idea is nice. 

-`_from_flat` (names welcome) is a new function that creates a group entirely from a hierarchy dict + a store. that dict must specify a root group, otherwise an exception is raised. We could revise this to create a root group if one is not specified. Open to suggestions here.

### Implicit groups
Partial hierarchies like `{'a': GroupMetadata(), 'a/b/c': ArrayMetadata(...)}` implicitly denote a group at `a/b`. When creating such a hierarchy, if we find an existing group at `a/b`, then we don't need to create a new one. So in the context of modeling a hierarchy, implicit groups are a little special -- by not specifying the properties of the group, the user / application is tolerant of _any_ group being there. So I introduced a subclass of `GroupMetadata` called `_ImplicitGroupMetadata`, which can be inserted into a hierarchy dict to explicitly denote groups that don't need to be written if one already exists. `_ImplicitGroupMetadata` is just like `GroupMetadata` except it errors if you try to set any parameter except `zarr_format`. 

### streaming v2 vs v3 node creation
creating v3 arrays / groups requires writing 1 metadata document, but v2 requires 2. To get the most concurrency I await the write of each metadata document separately, which means that `foo/.zattrs` might resolve before `foo/.zarray`. So in the v2 case I only yield up an array / group when _both_ documents were written. 

### Overlap with metadata consolidation logic
there's a lot of similarity between the stuff in this PR and routines used for consolidated metadata. it would be great to find ways to factor out some of the overlap areas

### still to do:
- write some more tests (checking that implicit groups don't get written if a group already exists)
- handle overwriting. I think the plan here is, if overwrite is False, then we do a check before any writing to ensure that there are no conflicts between the proposed hierarchy and the stuff that actually exists in storage. this check will involve more IO.

dcherian:
your proposal LGTM!

> The return value is dict[str, AsyncGroup | AsyncArray], but I guess it also doesn't have tor return anything,

Yes this return value is needed  for Xarray.

I think we would create the hierarchy dictionary here:
https://github.com/pydata/xarray/blob/5279bd15708e21ea582e56e4f27fc3e9663a8509/xarray/backends/zarr.py#L1194

then after the loop, call `create_nodes` about here:
https://github.com/pydata/xarray/blob/5279bd15708e21ea582e56e4f27fc3e9663a8509/xarray/backends/zarr.py#L1202

and then pass the returned `zarr.Array` and matching `v.data` to `ArrayWriter` which will have to learn to queue up those writes rather than execute them eagerly. 

As discussed earlier this week, zarr will need to add some new `zarr.set_many(list[zarr.Array], list[np.ndarray])` type API for bulk sets of chunks afterward, but that's out-of-scope for the current PR.

d-v-b:
I realized that there are a variety of complications to consider if want these methods to support creating arrays / groups in stores that already contain some arrays / groups. For example, if we use `set_if_not_exists` to write a zarr group, but a zarr array exists there already, then the end result will be a zarr array instead of a zarr group, which will be wrong. We will need to find the right abstraction level in the codebase for detecting / fixing these things. I worry that if we bake defensiveness too deep into the codebase, then we lose a lot of performance. So I don't want to handle all of these edge cases in this PR. 

Unless anyone disagrees, I'm going to design the low-level APIs in this PR under the assumption that we are creating arrays / groups in a completely empty store, and I will ensure that the hierarchy being stored is internally consistent. We can sort out the external consistency issues in a separate PR.

d-v-b:
this is ready for review. I'm going to work out the coverage issues.

d-v-b:
@tomnicholas you should have a look at some of these new functions / methods. I'd be happy to change things if you have some datatree conventions you'd like to suggest

src/zarr/core/group.py
```diff
@@ -1407,6 +1395,42 @@ async def _members(
...
+        nodes: dict[str, ArrayV2Metadata | ArrayV3Metadata | GroupMetadata],
+        overwrite: bool,
+    ) -> AsyncIterator[AsyncGroup | AsyncArray[ArrayV2Metadata] | AsyncArray[ArrayV3Metadata]]:
+        """
+        Create a hierarchy of arrays or groups rooted at this group.
+
+        This method takes a dictionary where the keys are the names of the arrays or groups
+        to create and the values are the metadata or objects representing the arrays or groups.
+
+        The method returns an asynchronous iterator over the created nodes.
```
dcherian, line 1411: A note about when to use this would be great

src/zarr/core/group.py
```diff
@@ -1407,6 +1395,42 @@ async def _members(
         ):
             yield member
 
+    # TODO: find a better name for this. create_tree could work.
```
dcherian, line 1398: hierarchy is fine to me.

src/zarr/core/group.py
d-v-b: good point, would something like "use this method to create an entire tree of sub-groups and / or sub-arrays efficiently." suffice?

src/zarr/core/group.py
```diff
@@ -2765,6 +2836,327 @@ def array(
...
+    nodes: dict[str, GroupMetadata | ArrayV2Metadata | ArrayV3Metadata],
+    semaphore: asyncio.Semaphore | None = None,
+) -> AsyncIterator[AsyncGroup | AsyncArray[ArrayV2Metadata] | AsyncArray[ArrayV3Metadata]]:
+    """
+    Create a collection of zarr arrays and groups concurrently and atomically. To ensure atomicity,
+    no attempt is made to ensure that intermediate groups are created.
+    """
+    ctx: asyncio.Semaphore | contextlib.nullcontext[None]
+
+    if semaphore is None:
```
dcherian, line 2972: why do we need `None` as an option here?

src/zarr/core/group.py
```diff
@@ -2765,6 +2836,327 @@ def array(
...
+    This function ensures that the input is transformed into a specification of a complete and valid
+    Zarr hierarchy.
+    """
+
+    observed_zarr_formats: dict[ZarrFormat, list[str | None]] = {2: [], 3: []}
+
+    # We will iterate over the dict again, but a full pass here ensures that the error message
+    # is comprehensive, and I think the performance cost will be negligible.
+    for k, v in data.items():
+        observed_zarr_formats[v.zarr_format].append(k)
```
dcherian, line 3092: can we short-circuit inside the loop and raise on the first instance of `zarr_format` that is not equal to `next(data.items())[1].zarr_format`?

src/zarr/core/group.py
```diff
@@ -2883,42 +3275,80 @@ async def _iter_members_deep(
...
+        return _build_metadata_v3(zarr_json)
+
+
+async def _read_metadata_v2(store_path: StorePath) -> ArrayV2Metadata | GroupMetadata:
+    """
+    Given a store_path, return ArrayV2Metadata or GroupMetadata defined by the metadata
+    document stored at store_path.path / (.zgroup | .zarray). If no such document is found,
+    raise a KeyError.
+    """
+    # TODO: consider first fetching array metadata, and only fetching group metadata when we don't
```
dcherian, line 3297: i think it's better to just grab all 3 at once honestly

src/zarr/core/group.py
d-v-b: `create_nodes` is low-level, kind of dangerous, and the expectation is that the user of this function knows what they are doing. So it doesn't default with any concurrency limit. It also doesn't check of nodes already exist, of the user wants to nest arrays inside arrays. A higher level function (`create_hierarchy`) does all that checking, and that's where the concurrency limit defaults to the value in the config.

src/zarr/core/group.py
d-v-b: we definitely could, but we get a much nicer error message if we can traverse the entire proposed hierarchy and identify all the problematic nodes.

src/zarr/core/group.py
dcherian: so why not require the semaphore if it's an advanced user?

tests/conftest.py
```diff
@@ -159,3 +173,210 @@ def zarr_format(request: pytest.FixtureRequest) -> ZarrFormat:
...
+    order: None,
+    zarr_format: Literal[3],
+    attributes: dict[str, JSON] | None,
+    chunk_key_encoding: ChunkKeyEncoding | ChunkKeyEncodingLike | None,
+    dimension_names: Iterable[str] | None,
+) -> ArrayV3Metadata: ...
+"""
+
+
+def create_array_metadata(
```
dcherian, line 218: seems like this would duplicate logic form elsewhere? Is there nowhere else this function could be used?

src/zarr/core/group.py
d-v-b: the semaphore is only necessary if you want some concurrency limiting mechanism, but I think that's a special case. A default of `None` means that users of the function don't need to run `from asyncio import Semaphore` before creating some nodes. E.g., there are a lot of places in the tests where I would want to use `create_nodes`, and in basically none of those places would a semaphore be necessary.

tests/conftest.py
d-v-b: this _should_ be duplicated, but sadly these functions don't exist in the codebase yet! I add some functions like this in another PR: https://github.com/zarr-developers/zarr-python/pull/2761

d-v-b:
question: should we export a sync version of `create_hierarchy` from the top-level `zarr` namespace?

dcherian:
> question: should we export a sync version of create_hierarchy from the top-level zarr namespace?

Yes, this would be used in Xarray.

d-v-b:
this PR adds a few functions that have async implementations and sync wrappers, like `create_hierarchy_a` (async) and `create_hierarchy`. Instead of putting (async, sync) pairs in the same module with slightly different names, I wonder if we should split the `group` module into an async-only namespace and a sync namespace (that imports stuff from the async namespace)? Then we don't need to mangle function names.

src/zarr/api/synchronous.py
```diff
@@ -1132,3 +1139,140 @@ def zeros_like(a: ArrayLike, **kwargs: Any) -> Array:
...
+    nodes: dict[str, GroupMetadata | ArrayV2Metadata | ArrayV3Metadata],
+    overwrite: bool = False,
+    allow_root: bool = True,
+) -> Iterator[Group | Array]:
+    """
+    Create a complete zarr hierarchy concurrently. Groups that are implicitly defined by the input
+    will be created as needed.
+
+    This function takes a parsed hierarchy dictionary and creates all the nodes in the hierarchy
+    concurrently. Arrays and Groups are yielded in the order they are created.
```
jhamman, line 1156:
```suggestion
    Create a complete zarr hierarchy from a collection of metadata objects.
    
    Groups that are implicitly defined by the input will be created as needed.

    This function takes a parsed hierarchy dictionary and creates all the nodes in the hierarchy
    concurrently. Arrays and Groups are yielded in the order they are created.
```

src/zarr/api/synchronous.py
```diff
@@ -1132,3 +1139,140 @@ def zeros_like(a: ArrayLike, **kwargs: Any) -> Array:
...
+        Whether to allow a root node to be created. If ``False``, attempting to create a root node
+        will result in an error. Use this option when calling this function as part of a method
+        defined on ``AsyncGroup`` instances, because in this case the root node has already been
+        created.
+
+    Yields
+    ------
+    Group | Array
+        The created nodes in the order they are created.
+    """
```
jhamman, line 1180: I think it would be worth adding a usage example here.

src/zarr/api/synchronous.py
```diff
@@ -1132,3 +1139,140 @@ def zeros_like(a: ArrayLike, **kwargs: Any) -> Array:
...
+    Group | Array
+        The created nodes.
+    """
+    coro = async_api.create_nodes(store=store, path=path, nodes=nodes)
+
+    for result in sync(_collect_aiterator(coro)):
+        yield _parse_async_node(result)
+
+
+def create_rooted_hierarchy(
```
jhamman, line 1222: I don't understand the use case for this function.

src/zarr/api/synchronous.py
```diff
@@ -1132,3 +1139,140 @@ def zeros_like(a: ArrayLike, **kwargs: Any) -> Array:
...
+    -------
+    Group | Array
+    """
+    async_node = sync(
+        async_api.create_rooted_hierarchy(store=store, path=path, nodes=nodes, overwrite=overwrite)
+    )
+    return _parse_async_node(async_node)
+
+
+def get_node(store: Store, path: str, zarr_format: ZarrFormat) -> Array | Group:
```
jhamman, line 1258: same here. this seems like a helper function but one that may not want to include as part of the public api

src/zarr/core/group.py
```diff
@@ -1407,6 +1394,62 @@ async def _members(
...
+
+        try:
+            async for node in create_hierarchy(
+                store=self.store,
+                path=self.path,
+                nodes=nodes,
+                overwrite=overwrite,
+                allow_root=False,
+            ):
+                yield node
```
jhamman, line 1442: this line is missing a test and feels important enough to call out.

src/zarr/core/group.py
```diff
@@ -2774,11 +2860,360 @@ def array(
...
+async def create_hierarchy(
+    *,
+    store: Store,
+    path: str,
+    nodes: dict[str, GroupMetadata | ArrayV3Metadata | ArrayV2Metadata],
+    overwrite: bool = False,
+    allow_root: bool = True,
+) -> AsyncIterator[AsyncGroup | AsyncArray[ArrayV2Metadata] | AsyncArray[ArrayV3Metadata]]:
+    """
+    Create a complete zarr hierarchy concurrently. Groups that are implicitly defined by the input
```
jhamman, line 2872: see suggested edits above to the api function version of this docstring

src/zarr/core/group.py
```diff
@@ -2774,11 +2860,360 @@ def array(
...
+
+            nodes_parsed = {
+                k: v for k, v in nodes_parsed.items() if k not in redundant_implicit_groups
+            }
+
+    async for node in create_nodes(store=store, path=path, nodes=nodes_parsed):
+        yield node
+
+
+async def create_nodes(
```
jhamman, line 2976: advocating to make this a private function and not export it to `zarr.api`.

tests/test_group.py
```diff
@@ -1443,7 +1459,454 @@ def test_delitem_removes_children(store: Store, zarr_format: ZarrFormat) -> None
 
 
 @pytest.mark.parametrize("store", ["memory"], indirect=True)
-def test_group_members_performance(store: MemoryStore) -> None:
+@pytest.mark.parametrize("impl", ["async", "sync"])
```
jhamman, line 1462: I'm interested to see if this pattern holds up. Running our `sync` function inside pytests IO loop may eventually lead to problems. But let's see.

jhamman (commented):
Gave this a review. My main suggestion is to reduce the scope of new public facing api to just `create_hierarchy`. We can come back and expose additional utilities here but for a first cut, I think its useful to limit the surface area of the API.

Really excited to get this into Xarray!

src/zarr/api/synchronous.py
d-v-b: this function returns a single zarr array or group (the root of the hierarchy); `create_hierarchy` returns an iterator over all the created nodes.

d-v-b: the use case is for when someone wants to create an entire hierarchy, and get as a return a value a handle to the root of that hierarchy. I suspect this is actually more typical than users wanting an iterator over everything in the hierarchy.

src/zarr/errors.py
```diff
@@ -57,3 +57,10 @@ class NodeTypeValidationError(MetadataValidationError):
     This can be raised when the value is invalid or unexpected given the context,
     for example an 'array' node when we expected a 'group'.
     """
+
+
+class RootedHierarchyError(BaseZarrError):
```
TomAugspurger, line 62: I find this name a bit confusing, but I might not understand the context.

If we're saying "you've tried to create a root, but it already exists" then perhaps a `RootExistsError`? But I think it's actually more like, you've tried to insert a root at a level below the root? So maybe something like `RootAsChildError`, `NestedRootError`, `ChildRootError`?

src/zarr/api/synchronous.py
TomNicholas: I think I agree with @jhamman - `create_rooted_hierarchy` doesn't really need to exist if it's pretty easy to get the root from `create_hierarchy`.

src/zarr/api/synchronous.py
```diff
@@ -1132,3 +1139,140 @@ def zeros_like(a: ArrayLike, **kwargs: Any) -> Array:
...
+    concurrently. Arrays and Groups are yielded in the order they are created.
+
+    Parameters
+    ----------
+    store : Store
+        The storage backend to use.
+    path : str
+        The name of the root of the created hierarchy. Every key in ``nodes`` will be prefixed with
+        ``path`` prior to creating nodes.
+    nodes : dict[str, GroupMetadata | ArrayV3Metadata | ArrayV2Metadata]
```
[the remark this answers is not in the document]
TomNicholas: Also the exact syntax of whether or not leading or trailing slashes are expected would be helpful too.

src/zarr/api/synchronous.py
```diff
@@ -1132,3 +1139,141 @@ def zeros_like(a: ArrayLike, **kwargs: Any) -> Array:
...
+    for result in sync(_collect_aiterator(coro)):
+        yield _parse_async_node(result)
+
+
+def create_nodes(
+    *, store: Store, path: str, nodes: dict[str, GroupMetadata | ArrayV2Metadata | ArrayV3Metadata]
+) -> Iterator[Group | Array]:
+    """Create a collection of arrays and / or groups concurrently.
+
+    Note: no attempt is made to validate that these arrays and / or groups collectively form a
```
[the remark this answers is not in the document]
TomNicholas: If so, we could just use `create_nodes` alone. 

IIUC the advantage of `create_hierarchy` is "safety" in the public zarr API. But to support writing multiple arbitrary new arrays/groups to an existing store concurrently requires the generality of `create_nodes`, so we need that one, and we hence have some "unsafe" public zarr API regardless.

src/zarr/core/group.py
TomNicholas: What's the position on using private zarr functions from within xarray?

src/zarr/errors.py
d-v-b: how about 
```python
class NestedRootError(BaseZarrError):
    """
    Exception raised when attempting to create a root node relative to a pre-existing root node.
    """
```

src/zarr/api/synchronous.py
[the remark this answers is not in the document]
d-v-b: > Is this the main / only difference between `create_nodes` and `create_hierarchy`?

yes. `create_hierarchy` attempts to model the rules of the zarr spec, and so it will not take an input like `{'a': ArrayMetadata, 'a/b': ArrayMetadata}`, which would nest an array inside another array. Whereas `create_nodes` doesn't do any input checking at all. it just creates nodes.

src/zarr/api/synchronous.py
d-v-b: Getting the root is easy, in a computer science sense, but it's also tedious. Look at the source code for `create_rooted_hierarchy` and tell me if you want every zarr user to write this themselves (or propose a simplification that renders it untedious :) ): 

https://github.com/d-v-b/zarr-python/blob/545cacb543a8e2c1e634530a1fdb530d6faa23f7/src/zarr/core/group.py#L3555-L3583

src/zarr/api/synchronous.py
d-v-b: I can remove it, but I think a function for getting an array or group is pretty useful to end users

src/zarr/errors.py
TomAugspurger: Yeah, that sounds good.

d-v-b:
thanks for the feedback all, I'm working on integrating the requested changes. I have a question about how we would expect `Group.create_hierarchy` to work if `overwrite` was set to `True`. Suppose we have a Group with the following layout:
```
{
  'a': Group
  'b': Group
}
```
i.e., our group contains two subgroups, named `'a'` and `'b'`. If I invoke `g.create_hierarchy({'b': GroupMetadata}, overwrite=True)`, then clearly this should overwrite the group at `'b'`, but would you also expect it to remove `'a'`? For me, I would expect `'a'` to survive. So I'm thinking the scope of "overwrite" should be limited to only affect the subnodes of the provided nodes, and not have any off-target effects. Does anyone disagree?

Slightly related to this, I don't think we have any methods for safely _removing_ arrays or groups from an existing group. If you run `g.store.delete_dir(g.path)`, for example, you will wipe all of the zarr metadata documents for `g`. This bug affects this PR (the solution is to do a `list_dir` and delete all the non-metadata documents), and I'm going to fix it, but I can imagine more places where this kind of thing will turn up.

d-v-b:
> Slightly related to this, I don't think we have any methods for safely _removing_ arrays or groups from an existing group. If you run `g.store.delete_dir(g.path)`, for example, you will wipe all of the zarr metadata documents for `g`. This bug affects this PR (the solution is to do a `list_dir` and delete all the non-metadata documents), and I'm going to fix it, but I can imagine more places where this kind of thing will turn up.

aha we do have methods for doing this defined on `AsyncGroup`. Very good.

TomNicholas:
I vote 3rd option. Let me just make whatever keys I want to populate, relative to the root of the store.

d-v-b:
> I vote 3rd option. Let me just make whatever keys I want to populate, relative to the root of the store.

that works for me. And what should the returned keys be for `d = dict(group.create_hierarchy(...))`? Relative to the path of the group, or the root of the store? IMO they should be relative to the path of the group (i.e., the same keys come out as went in)

TomNicholas:
> IMO they should be relative to the path of the group

That sounds fine as it's clear that the `group.create_hierarchy` API is already looking at one specific group.

src/zarr/errors.py
d-v-b: after a refactor, this exception isn't needed any more

src/zarr/api/synchronous.py
```diff
@@ -1132,3 +1139,141 @@ def zeros_like(a: ArrayLike, **kwargs: Any) -> Array:
         The new array.
     """
     return Array(sync(async_api.zeros_like(a, **kwargs)))
+
+
+def create_hierarchy(
+    store: Store,
+    path: str,
+    nodes: dict[str, GroupMetadata | ArrayV2Metadata | ArrayV3Metadata],
+    overwrite: bool = False,
```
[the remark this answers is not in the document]
d-v-b: this should now be documented in the docstrings, let me know if I missed any

src/zarr/api/synchronous.py
```diff
@@ -1132,3 +1139,141 @@ def zeros_like(a: ArrayLike, **kwargs: Any) -> Array:
...
+    allow_root: bool = True,
+) -> Iterator[Group | Array]:
+    """
+    Create a complete zarr hierarchy from a collection of metadata objects.
+    
+    Groups that are implicitly defined by the input will be created as needed.
+
+    This function takes a parsed hierarchy dictionary and creates all the nodes in the hierarchy
+    concurrently. Arrays and Groups are yielded in the order they are created.
```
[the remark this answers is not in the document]
d-v-b: the docstrings should now state that the order is not deterministic

src/zarr/api/synchronous.py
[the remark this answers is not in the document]
d-v-b: i think the new docstrings should clarify this.

src/zarr/core/group.py
d-v-b: it's now private

src/zarr/api/synchronous.py
d-v-b: removed from public api

src/zarr/api/synchronous.py
d-v-b: this is removed from public api

TomNicholas (approved):

dcherian (approved):
Nice. The public API `create_hierarchy` looks nice to me.

d-v-b:
test failure is unrelated to this PR (looks like an fsspec thing)

PR #2669, opened 2025-01-07
api: hide zarr.core from api docs

### LINKED ISSUE #2621: Make zarr.core private

jhamman:
Closes #2621

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -71,7 +71,7 @@ def skip_submodules(
       ) -> bool:
     # Skip documenting zarr.codecs submodules
     # codecs are documented in the main zarr.codecs namespace
-    if what == "module" and name.startswith("zarr.codecs."):
+    if what == "module" and name.startswith("zarr.codecs.") or name.startswith("zarr.core"):
         skip = True
     return skip
 
--- a/src/zarr/core/__init__.py
+++ b/src/zarr/core/__init__.py
@@ -1,3 +1,8 @@
+"""
+The ``zarr.core`` module is considered private API and should not be imported
+directly by 3rd-party code.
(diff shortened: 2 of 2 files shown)


normanrz (approved):

dstansby (changes requested):
👍 overall. There's a couple of places in the docs where we reference stuff in `zarr.core` in public docs though, which need fixing:

This breaks a link in the config guide: https://github.com/zarr-developers/zarr-python/blob/8bb0b3457bc31925e2ad0e737f1b29de9da74cbf/docs/user-guide/config.rst?plain=1#L6

Looking lower down the config [user guide](https://zarr--2669.org.readthedocs.build/en/2669/user-guide/config.html), several of the objects referenced in the default config are from the `zarr.core` namespace, so they should either be added back to the API documentation, or moved eleswhere in the API so we're not giving users references to API that we're considering private.

`zarr.core.group.ConsolidatedMetadata` is also publicly facing according to the consolidated metdata guide, so that needs making public in the docs too.

jhamman:
I don't think we have time today to do a proper fix to the consolidated metadata class but I'll fix the config link now.

jhamman:
I think this is as far as we should take it right now. Follow up actions could include:

- Move `zarr.core.buffer.cpu` to a public place (maybe even numcodecs)
- Move `zarr.core.codec_pipeline.BatchedCodecPipeline` to a public place

Since both of these are quite advanced parts of Zarr, I'm not too concerned about moving these after 3.0.0.

PR #2672, opened 2025-01-08
Improve development doc structure

dstansby:
- Remove the roadmap, since this seems to be a historical document that is out of date. Alternatively we could keep this here and put a disclaimer that it's a historical document at the top?
- Split the developer's guide into two pages, one for contributors, and one for maintainers. I think this makes the content more discoverable (for both sets of people), and also keeps the contributor's guide a bit simpler.


dstansby:
👍 - I didn't spot the note initially (my bad), I've just moved it to the top of the page in the latest revision.

Issue #829, opened 2021-09-09
Generate changelogs from fragements


joshmoore:
e.g. https://github.com/ansible-community/antsibull-changelog/blob/main/docs/other-projects.rst

At the moment, PRs regularly conflict due to conflicts to release.rst. By moving to a fragment based system, this _should_ go away. 

Any other recommendations for frameworks?

grlee77:
[Towncrier](https://towncrier.readthedocs.io/en/actual-freaking-docs/) is used by NumPy and we are considering it for scikit-image

joshmoore:
Had looked at that one previously as well. :+1: All the better if your experiences have been good.

grlee77:
We have a draft PR and were planning to start using it in the near future, so I don't have much experience with it yet. I was not aware of the ansible one you linked above, but it sounds like something similar.

Another approach that SciPy has used in the past is having contributors update a wiki page that the release manager can then pull from when making the release notes. I'm not sure how consistently it is applied, though.

Towncrier or antsibull will solve the conflicts problem, but there is still the separate problem of prompting contributors to create the fragments needed!

joshmoore:
I'd hope it'd be easy enough to add an action that says, "must have fragment". Worst case: we can push it to the PRs manually. :wink:

joshmoore:
Another possible option: https://sphinx-github-changelog.readthedocs.io/en/latest/ (...as I struggle with the anchors for pre-releases ...)

cc: @MSanKeys963

joshmoore:
- @jakirkham suggests https://github.com/regro/rever
- @alimanfoo suggests looking at how dask does it.

jakirkham:
We've used [`rever`]( https://github.com/regro/rever ) in `conda-smithy`

jakirkham:
Josh mentioned [this changelog proposal]( https://keepachangelog.com/en/1.0.0/ ).

jakirkham:
Another option somewhat similar to towncrier is [scriv]( https://scriv.readthedocs.io/en/latest/ )

joshmoore:
@MSanKeys963: this is something to review before we (potentially) launch the outreachy project.

MSanKeys963:
Sure. Adding to the list.
We can let the Outreachy intern experiment with multiple tools before we finalise a single one for Zarr.

jakirkham:
[...]

PR #2680, opened 2025-01-09
Bootstrap release notes post v3

dstansby:
This adds a new section for new release notes, and also updates the release guide to explain how to handle the release notes. The sections I chose for the release notes are very much up for change/suggestion.

### DIFF
--- a/docs/developers/contributing.rst
+++ b/docs/developers/contributing.rst
@@ -329,10 +329,16 @@ Release procedure
    Most of the release process is now handled by GitHub workflow which should
    automatically push a release to PyPI if a tag is pushed.
 
-Before releasing, make sure that all pull requests which will be
-included in the release have been properly documented in
-`docs/release.rst`.
+Pre-release
+"""""""""""
+1. Make sure that all pull requests which will be
+   included in the release have been properly documented in
+   :file:`docs/release-notes.rst`.
+2. Rename the "Unreleased" section heading in :file:`docs/release-notes.rst`
+   to the version you are about to release.
 
+Releasing
+"""""""""
 To make a new release, go to
 https://github.com/zarr-developers/zarr-python/releases and
 click "Draft a new release". Choose a version number prefixed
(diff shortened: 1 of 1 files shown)


docs/developers/contributing.rst
```diff
@@ -329,10 +329,16 @@ Release procedure
...
-Before releasing, make sure that all pull requests which will be
-included in the release have been properly documented in
-`docs/release.rst`.
-
+Pre-release
+"""""""""""
+1. Make sure that all pull requests which will be
+   included in the release have been properly documented in
+   :file:`docs/release-notes.rst`.
+2. Rename the "Unreleased" section heading in :file:`docs/release-notes.rst`
```
normanrz, line 337: That seems a bit annoying because it requires another commit before the release or am I understanding it wrong?

docs/developers/contributing.rst
```diff
@@ -355,5 +361,8 @@ https://readthedocs.io. Full releases will be available under
 pre-releases will be available under
 `/latest <https://zarr.readthedocs.io/en/latest>`_.
 
-Also review and merge the https://github.com/conda-forge/zarr-feedstock
-pull request that will be automatically generated.
+Post-release
+""""""""""""
+
+- Review and merge the pull request on the `conda-forge feedstock <https://github.com/conda-forge/zarr-feedstock>`_ that will be automatically generated.
+- Create a new "Unreleased" section in the release notes
```
normanrz, line 368: And another commit here?

docs/developers/contributing.rst
dstansby: Yeah, that's right. I guess with the current system, unless we decide on the version numbering ahead of time then this is unavoidable?

dstansby: On other projects I've used https://towncrier.readthedocs.io/en/stable/ with success, and it was suggested a while ago here too: https://github.com/zarr-developers/zarr-python/issues/829.

You still have to generate the changelog before release, but it makes having multiple branches easier (e.g., if we have a 3.0.x branch for bugfixes and `main` for 3.x.0 releases), and would mean the extra commit after release (your comment below) wouldn't be needed.

I'd be up for trying to set towncrier up, either instead of or after this PR?

normanrz:
Fwiw, I started release notes in #2679

normanrz (approved):

Issue #2134, opened 2024-08-28
[v3] Structured dtype support


jhamman:
Zarr-Python 2 [supported structured dtypes](https://zarr.readthedocs.io/en/stable/tutorial.html#indexing-fields-in-structured-arrays). I don't think this has been discussed much, if at all, in the context of version 3.

This once worked:

```python
a = np.array(
    [(b"aaa", 1, 4.2), (b"bbb", 2, 8.4), (b"ccc", 3, 12.6)],
    dtype=[("foo", "S3"), ("bar", "i4"), ("baz", "f8")],
)
za = zarr.array(a, chunks=2, fill_value=None)
za[:]
```

d-v-b:
The v3 spec lists a [finite set of dtypes](https://github.com/zarr-developers/zarr-specs/blob/main/docs/v3/core/v3.0.rst#data-types) and structured dtypes are not explicitly in the list. To be spec compliant and support structured dtypes, I see two options: 
- express a structured dtype via the raw bits `r*` dtype. This seems unsatisfactory, because two structured dtypes might have the same raw bits representation but very different semantics (e.g., a pair of float64s vs a pair of int64s will look the same as raw bits)
- define a spec extension for structured dtypes. I don't think this would be a technical challenge, but it would be important to get consensus across implementations about the JSON serialization for the dtype. We would not want to overfit to numpy conventions. This seems like a better option, but it requires some social logistics.

ilan-gold:
@d-v-b @jhamman Has there been any movement on this? I see there is a comment in the code base about this: https://github.com/zarr-developers/zarr-python/blob/2fe12a7f424ac78c18a688d0a75049a3df67c98f/tests/test_indexing.py#L161

This would be something of a regression for `anndata` as a project if we couldn't support this (or we'd have to go with option 1 and investigate the current status with readers in other languages to see if they could be made to support this functionality).

jhamman:
@ilan-gold - no progress and I agree its a regression. I think we should support it for v2 arrays but until there is an extension dtype for this in v3, we'll likely need to leave it out.

ilan-gold:
That would be very fine with me.  We'll need to "break" things for the v3 format anyway, so I don't have strong opinions there.

d-v-b:
> but until there is an extension dtype for this in v3

[...]

PR #2681, opened 2025-01-09
(fix): structured arrays for v2

ilan-gold:
This is a best guess based on https://numpy.org/doc/2.1/reference/generated/numpy.dtype.kind.html and the fact that `VLenBytes` appears to be explicitly for strings.

This addresses the v2 case of #2134

TODO:

### DIFF
--- a/src/zarr/core/buffer/core.py
+++ b/src/zarr/core/buffer/core.py
@@ -470,7 +470,9 @@ class NDBuffer:
         # every single time we have to write data?
         _data, other = np.broadcast_arrays(self._data, other)
         return np.array_equal(
-            self._data, other, equal_nan=equal_nan if self._data.dtype.kind not in "USTO" else False
+            self._data,
+            other,
+            equal_nan=equal_nan if self._data.dtype.kind not in "USTOV" else False,
         )
 
     def fill(self, value: Any) -> None:
--- a/src/zarr/core/metadata/v2.py
+++ b/src/zarr/core/metadata/v2.py
@@ -193,7 +193,12 @@ class ArrayV2Metadata(Metadata):
             zarray_dict["fill_value"] = fill_value
 
         _ = zarray_dict.pop("dtype")
-        zarray_dict["dtype"] = self.dtype.str
+        dtype_json: JSON
+        if self.dtype.kind == "V":
+            dtype_json = tuple(self.dtype.descr)
+        else:
+            dtype_json = self.dtype.str
+        zarray_dict["dtype"] = dtype_json
 
         return zarray_dict
 
@@ -220,6 +225,8 @@ class ArrayV2Metadata(Metadata):
 
 
 def parse_dtype(data: npt.DTypeLike) -> np.dtype[Any]:
+    if isinstance(data, list):  # this is a valid _VoidDTypeLike check
+        data = [tuple(d) for d in data]
     return np.dtype(data)
 
 
--- a/src/zarr/core/metadata/v2.py
+++ b/src/zarr/core/metadata/v2.py
@@ -194,7 +194,9 @@ class ArrayV2Metadata(Metadata):
 
         _ = zarray_dict.pop("dtype")
         dtype_json: JSON
-        if self.dtype.kind == "V":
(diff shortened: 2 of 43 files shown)


ilan-gold:
Hmm, this will need to handle the case where the array is not given the `dtype`.

ilan-gold:
(P.S I used `zarr.create(path=...)` which was allowed to pass but actually does nothing and uses a `MemoryStore`, which seems like a common error)

src/zarr/core/buffer/core.py
```diff
@@ -470,7 +470,9 @@ def all_equal(self, other: Any, equal_nan: bool = True) -> bool:
         # every single time we have to write data?
         _data, other = np.broadcast_arrays(self._data, other)
         return np.array_equal(
-            self._data, other, equal_nan=equal_nan if self._data.dtype.kind not in "USTO" else False
+            self._data,
+            other,
+            equal_nan=equal_nan if self._data.dtype.kind not in "USTOV" else False,
```
d-v-b, line 475: can we factor this check out into a stand-alone boolean function?

ilan-gold: Where? Is there somewhere this would be re-used in the core?

src/zarr/core/metadata/v2.py
```diff
@@ -193,7 +193,14 @@ def to_dict(self) -> dict[str, JSON]:
...
 
         _ = zarray_dict.pop("dtype")
-        zarray_dict["dtype"] = self.dtype.str
+        dtype_json: JSON
+        # In the case of zarr v2, the simplest i.e., '|VXX' dtype is represented as a string
+        dtype_descr = self.dtype.descr
+        if self.dtype.kind == "V" and dtype_descr[0][0] != "" and len(dtype_descr) != 0:
+            dtype_json = tuple(self.dtype.descr)
+        else:
+            dtype_json = self.dtype.str
```
ilan-gold, line 202: This is my attempt to match the old behavior.  I didn't look back at the old code yet, but if someone knows this to be wrong, would be great to know.

martindurant: Looks right to me

src/zarr/core/buffer/core.py
martindurant: The change here is just to add "V", so I wouldn't change it now

src/zarr/core/metadata/v2.py
```diff
@@ -220,6 +227,8 @@ def update_attributes(self, attributes: dict[str, JSON]) -> Self:
 
 
 def parse_dtype(data: npt.DTypeLike) -> np.dtype[Any]:
+    if isinstance(data, list):  # this is a valid _VoidDTypeLike check
```
martindurant, line 230: Any iterable?

martindurant (commented):
I am keen to see this go in

src/zarr/core/metadata/v2.py
ilan-gold: This is to handle the `[(field_name, field_dtype, field_shape), ...]` case on https://numpy.org/doc/2.1/reference/arrays.dtypes.html#specifying-and-constructing-data-types but at the same time to obey
Screenshot 2025-01-14 at 19 22 14

This might require more stringent checking or tests...Not sure.  The reason this `tuple` conversion happens is that lists incoming from on-disk reads are lists, not tuples.  So maybe we should check `list` and `data[0]` is also list? And throw an error if it isn't?  I'm not sure what else it could be though

martindurant: I guess the dtype constructor would make an exception (or our own comprehension fails) in the case the JSON on disk was edited - so I'm not too worried.

martindurant (approved):

martindurant:
ping @d-v-b - I'd be happy to merge if you have no objections.

d-v-b:
I think this looks good, but I'm admittedly not a structured dtype user, so I can't give it a very close examination. I think the only thing remaining is to use the new changelog system added by #2736

PR #2682, opened 2025-01-10
test: enable codecov in main test action

### LINKED ISSUE #2613: [TST] Bring back Codecov?
While we were developing Zarr-Python 3, test coverage checks stopped getting published to [Codecov](https://app.codecov.io/gh/zarr-developers/zarr-python).

Do we want to bring Codecov back? If so, what needs to be done to enable this in hatch / pytest / github actions?

jhamman:
Closes #2613 

TODO:


jhamman:
This is working. We may need to merge with codecov check failing because it is comparing against eaf5d7a which is way back in v2 land. Going forward, subsequent checks should compare against main and the behavior should be that the check will only fail if the coverage goes down.

d-v-b (approved):

Issue #2614, opened 2025-01-01
[TST] Bring test coverage to 100%


jhamman:
As of today, Zarr-Python's main branch has a test coverage of 87%. We should strive to bring this back to 100%. Here is the current report:

```
❯ hatch env run --env test.py3.12-2.1-optional run-coverage
---------- coverage: platform darwin, python 3.12.8-final-0 ----------
Name                                   Stmts   Miss  Cover
----------------------------------------------------------
src/zarr/__init__.py                       7      0   100%
src/zarr/_compat.py                        6      0   100%
src/zarr/_version.py                      10      1    90%
src/zarr/abc/__init__.py                   0      0   100%
src/zarr/abc/codec.py                     71     11    85%
src/zarr/abc/metadata.py                  26      3    88%
src/zarr/abc/store.py                    129     21    84%
src/zarr/api/__init__.py                   0      0   100%
src/zarr/api/asynchronous.py             311     50    84%
src/zarr/api/synchronous.py               77     20    74%
src/zarr/codecs/__init__.py               10      0   100%
src/zarr/codecs/_v2.py                    51      6    88%
src/zarr/codecs/blosc.py                  99     15    85%
src/zarr/codecs/bytes.py                  68      8    88%
src/zarr/codecs/crc32c_.py                38      4    89%
src/zarr/codecs/gzip.py                   39      6    85%
src/zarr/codecs/sharding.py              325     40    88%
src/zarr/codecs/transpose.py              47      5    89%
src/zarr/codecs/vlen_utf8.py              71      4    94%
src/zarr/codecs/zstd.py                   56      8    86%
src/zarr/convenience.py                    4      4     0%
src/zarr/core/__init__.py                  3      0   100%
src/zarr/core/_info.py                    78      4    95%
src/zarr/core/_tree.py                    35     35     0%
src/zarr/core/array.py                   572     37    94%
src/zarr/core/array_spec.py               59      3    95%
src/zarr/core/attributes.py               29      7    76%
src/zarr/core/buffer/__init__.py           3      0   100%
src/zarr/core/buffer/core.py             126     27    79%
src/zarr/core/buffer/cpu.py               56      4    93%
src/zarr/core/buffer/gpu.py               75     43    43%
src/zarr/core/chunk_grids.py              69      8    88%
src/zarr/core/chunk_key_encodings.py      58     11    81%
src/zarr/core/codec_pipeline.py          111     10    91%
src/zarr/core/common.py                   84      9    89%
[...]

PR #2688, opened 2025-01-11
Update set_partial_values to accept Buffer rather than BytesLike

maxrjones:
When working on https://github.com/zarr-developers/zarr-python/issues/2614 in https://github.com/maxrjones/zarr-python/tree/testing-storage, I found that the current implementation of `set_partial_values` for local store cannot work as written because it passes `BytesLike` to `_put` which expects a buffer. None of the other stores implement `set_partial_values`, so it's not possible to look to them for the correct way of implementing this method.

I believe that one of the following needs to happen in order for stores to correctly implement `set_partial_values`:
- Add a `prototype` argument to `set_partial_values` that can be used to go from BytesLike -> Buffer
- Update `set_partial_values` to accept `Buffer` types rather than `BytesLike` types

I chose the second option for this PR because all the other `set` methods seem to accept `Buffer` types. Of course, I welcome corrections if I am not understanding something here and the function signature should work as-is.

### DIFF
--- a/src/zarr/storage/_memory.py
+++ b/src/zarr/storage/_memory.py
@@ -38,7 +38,7 @@ class MemoryStore(Store):
 
     supports_writes: bool = True
     supports_deletes: bool = True
-    supports_partial_writes: bool = True
+    supports_partial_writes: bool = False
     supports_listing: bool = True
 
     _store_dict: MutableMapping[str, Buffer]
@@ -134,7 +134,7 @@ class MemoryStore(Store):
         except KeyError:
             logger.debug("Key %s does not exist.", key)
 
-    async def set_partial_values(self, key_start_values: Iterable[tuple[str, int, bytes]]) -> None:
+    async def set_partial_values(self, key_start_values: Iterable[tuple[str, int, Buffer]]) -> None:
         # docstring inherited
         raise NotImplementedError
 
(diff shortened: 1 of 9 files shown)


src/zarr/testing/store.py
```diff
@@ -181,6 +181,23 @@ async def test_set_many(self, store: S) -> None:
...
             assert (await self.get(store, k)).to_bytes() == v.to_bytes()
 
+    @pytest.mark.parametrize("key", ["zarr.json", "c/0", "foo/c/0.0", "foo/0/0"])
+    async def test_set_partial_values(self, store: S, key: str) -> None:
+        """
+        Ensure that data can be written to the store using the store.set_partial_values method.
+        """
+        assert not store.read_only
+        # Create empty key
+        await store.set(key, self.buffer_cls.from_bytes(b""))
```
maxrjones, line 191: `set_partial_values` for a `LocalStore` requires that a key exist. This test assumes this is the intended behavior and creates a key containing no data prior to testing set_partial_values.

src/zarr/storage/_memory.py
```diff
@@ -38,7 +38,7 @@ class MemoryStore(Store):
 
     supports_writes: bool = True
     supports_deletes: bool = True
-    supports_partial_writes: bool = True
+    supports_partial_writes: bool = False
```
maxrjones, line 41: I'm realizing this may be wrong. I interpreted `supports_partial_writes` to refer to whether `set_partial_values` is implemented whereas it may actually refer to whether `set` includes a `byte_range` parameter. I am not sure which, if either, is correct because `MemoryStore` had `supports_partial_writes=True` with `byte_range` implemented for `set` but no implemented `supports_partial_writes` whereas LocalStore had `supports_partial_writes=True` with no `byte_range` implemented for `set` but included a `supports_partial_writes` implementation.

Issue #2630, opened 2025-01-03
Test for group creation across all `mode` options


d-v-b:
Based on #2629 it seems that our test suite does not check group creation with `mode=w`. This probably means a test is not parametrized completely. This should be fixed.

dikwickley:
@d-v-b added a pr for this. thanks!

asimchoudhary:
Hii @d-v-b , i have opened the pr  #2840  for the above issue.

PR #2690, opened 2025-01-12
Add test for other modes

dikwickley:
Fixes: #2630 

TODO:

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -1086,10 +1086,14 @@ async def test_open_falls_back_to_open_group_async() -> None:
     assert group.attrs == {"key": "value"}
 
 
-def test_open_mode_write_creates_group(tmp_path: pathlib.Path) -> None:
+@pytest.mark.parametrize("mode", ["w", "a"])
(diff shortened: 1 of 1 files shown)


tests/test_api.py
```diff
@@ -1086,10 +1086,14 @@ async def test_open_falls_back_to_open_group_async() -> None:
...
 
 
-def test_open_mode_write_creates_group(tmp_path: pathlib.Path) -> None:
+@pytest.mark.parametrize("mode", ["w", "a"])
+def test_open_write_modes_creates_group(tmp_path: pathlib.Path, mode: str) -> None:
     # https://github.com/zarr-developers/zarr-python/issues/2490
-    zarr_dir = tmp_path / "test.zarr"
-    group = zarr.open(zarr_dir, mode="w")
+    zarr_dir = tmp_path / f"{mode}-test.zarr"
+    if mode in ["r", "r+"]:
```
dstansby, line 1093: This block won't every be executed, because the mode is only 'w' or 'a' in this test. Do you think we should get rid of the `if` block, or extend the parameterisation to include 'r' and 'r+'?

PR #2691, opened 2025-01-12
add enter and exit methods to groups.

dikwickley:
[Description of PR]

TODO:

### DIFF
--- a/src/zarr/core/group.py
+++ b/src/zarr/core/group.py
@@ -7,7 +7,7 @@ import logging
 import warnings
 from collections import defaultdict
 from dataclasses import asdict, dataclass, field, fields, replace
-from typing import TYPE_CHECKING, Literal, TypeVar, assert_never, cast, overload
+from typing import TYPE_CHECKING, Literal, TypeVar, assert_never, cast, overload, Self
 
 import numpy as np
 import numpy.typing as npt
@@ -1751,6 +1751,12 @@ class Group(SyncMixin):
         """
         obj = sync(AsyncGroup.open(store, zarr_format=zarr_format))
         return cls(obj)
+
+    def __enter__(self) -> Self:
+        return self
+
+    def __exit__(self) -> None:  # noqa: PYI036
+        self.store.close()
 
     def __getitem__(self, path: str) -> Array | Group:
         """Obtain a group member.


normanrz (commented):
Will you also add this for arrays?

src/zarr/core/group.py
```diff
@@ -1752,6 +1752,12 @@ def open(
         obj = sync(AsyncGroup.open(store, zarr_format=zarr_format))
         return cls(obj)
 
+    def __enter__(self) -> Self:
+        return self
+
+    def __exit__(self) -> None:  # noqa: PYI036
```
jhamman, line 1758:
```suggestion
    def __exit__(
        self,
        typ: type[BaseException] | None,
        exc: BaseException | None,
        tb: TracebackType | None
    ) -> None:
```

See https://docs.astral.sh/ruff/rules/bad-exit-annotation/

jhamman (changes requested):
Let's also get some unit tests for this. It will be important that we exercise this across all stores and establish expectations about behavior after the store has been closed.

src/zarr/core/group.py
dikwickley: updated in [d8f7d68](https://github.com/zarr-developers/zarr-python/pull/2691/commits/d8f7d686a15e3a62e87fb29f8d2288a08f4118ad). Also added tests.

tests/test_group.py
```diff
@@ -428,6 +428,28 @@ def test_group_len(store: Store, zarr_format: ZarrFormat) -> None:
     assert len(group) == 0
 
 
+def test_group_with_context_manager(store: Store, zarr_format: ZarrFormat, overwrite: bool) -> None:
+    spath = StorePath(store)
+
+    # attempt to open a group that does not exist.
+    with pytest.raises(FileNotFoundError):
+        with Group.open(store) as store:
```
jhamman, line 436:
```suggestion
        with Group.open(store) as g:
```

tests/test_group.py
```diff
@@ -428,6 +428,28 @@ def test_group_len(store: Store, zarr_format: ZarrFormat) -> None:
...
+    with Group.from_store(
+        store, attributes=attrs, zarr_format=zarr_format, overwrite=overwrite
+    ) as group:
+        assert store._is_open
+        assert group.attrs == attrs
+        assert group.metadata.zarr_format == zarr_format
+        assert group.store_path == spath
+
+    # Check if store was closed after exit.
+    assert not store._is_open
```
jhamman, line 450: rather than using this private store api, let's try to just to use the group again and make sure it fails.


```suggestion
    with pytest.raises(ValueError):
        group.attrs['path'] = 'bar'
```

tests/test_group.py
jhamman, line 436: also, instead of using `Group.open` and `Group.from_store`, can we use `zarr.open_group` and `zarr.create_group`? These are the public facing API methods we want most folks to be using.

jhamman (commented):
Super close. A few suggestions on how to make the tests just a bit better.

tests/test_group.py
dikwickley: updated in [3aee7b0](https://github.com/zarr-developers/zarr-python/pull/2691/commits/3aee7b00112896df8f871cc73bfa4120c4612b79)

dikwickley: updated in [3aee7b0](https://github.com/zarr-developers/zarr-python/pull/2691/commits/3aee7b00112896df8f871cc73bfa4120c4612b79)

tests/test_group.py
dikwickley: @jhamman apologise for such late response on this.
```
with pytest.raises(ValueError):
        group.attrs['path'] = 'bar'
```
does not raise any value error. the store is closed but the group instance still works.

jhamman: If you can edit the Group metadata after the Group has been closed, what is the point of the context manager?

PR #2693, opened 2025-01-13
Improve test coverage for storage classes

maxrjones:
This PR improves the test coverage for the various storage classes. While testing the storage classes, I fixed a few issues:
- Implement `open()` for `LoggingStore`
- Add `_is_open` property and setter for `WrapperStore`
- Use `stdout` rather than `stderr` as the default stream for `LoggingStore`
- Ensure that `ZipStore` is open before getting or setting any values
- Update equality for `LoggingStore` and `WrapperStore` such that the types much be equal. This is an opinionated change. For example, previously a LocalStore and LoggingStore instance could be evaluated as equal, whereas now they are distinct.

Here's the change in coverage:
```
src/zarr/abc/store.py                    84% -> 93%
src/zarr/storage/__init__.py             94% -> 94%
src/zarr/storage/_utils.py               94% -> 97%
src/zarr/storage/common.py               80% -> 91%
src/zarr/storage/fsspec.py               25% -> 90%
src/zarr/storage/local.py                86% -> 92%
src/zarr/storage/logging.py              62% -> 96%
src/zarr/storage/memory.py               82% -> 85%
src/zarr/storage/wrapper.py              56% -> 94%
src/zarr/storage/zip.py                  96% -> 97%
src/zarr/testing/store.py                92% -> 99%
```

`src/zarr/storage/memory.py` coverage is low because it includes the GPUStore and I don't have a test environment with cuda. I'm opening this PR now even though it's not at 100% coverage because I don't expect to have much time to work on it during the week and would rather the PR not get stale if the team has time for a review.

The set partial values methods are addressed separately because they require discussion (xref #2688).

### DIFF
--- a/changes/2693.bugfix.rst
+++ b/changes/2693.bugfix.rst
@@ -1,0 +1,14 @@
+Implement open() for LoggingStore
+LoggingStore is now a generic class.
+Add _is_open property and setter for WrapperStore
+Use stdout rather than stderr as the default stream for LoggingStore
+Ensure that ZipStore is open before getting or setting any values
+Update equality for LoggingStore and WrapperStore such that the types must be equal.
+Indicate StoreTest's `test_store_repr`, `test_store_supports_writes`, `test_store_supports_partial_writes`, and `test_store_supports_listing` need to be implemented using `@abstractmethod` rather than `NotImplementedError`.
+Separate instantiating and opening a store in StoreTests
+Test using Store as a context manager in StoreTests
+Match the errors raised by read only stores in StoreTests
+Test that a ValueError is raise for invalid byte range syntax in StoreTests
+Test getsize() and getsize_prefix() in StoreTests
+Test the error raise for invalid buffer arguments in StoreTests
+Test that data can be written to a store that's not yet open using the store.set method in StoreTests
--- a/changes/2693.bugfix.rst
+++ b/changes/2693.bugfix.rst
@@ -1,6 +1,5 @@
 Implement open() for LoggingStore
 LoggingStore is now a generic class.
-Add _is_open property and setter for WrapperStore
 Use stdout rather than stderr as the default stream for LoggingStore
 Ensure that ZipStore is open before getting or setting any values
 Update equality for LoggingStore and WrapperStore such that the types must be equal.
@@ -10,5 +9,5 @@ Test using Store as a context manager in StoreTests
 Match the errors raised by read only stores in StoreTests
 Test that a ValueError is raise for invalid byte range syntax in StoreTests
 Test getsize() and getsize_prefix() in StoreTests
-Test the error raise for invalid buffer arguments in StoreTests
+Test the error raised for invalid buffer arguments in StoreTests
 Test that data can be written to a store that's not yet open using the store.set method in StoreTests
--- a/changes/2693.bugfix.rst
+++ b/changes/2693.bugfix.rst
@@ -2,7 +2,7 @@ Implement open() for LoggingStore
 LoggingStore is now a generic class.
 Use stdout rather than stderr as the default stream for LoggingStore
 Ensure that ZipStore is open before getting or setting any values
-Update equality for LoggingStore and WrapperStore such that the types must be equal.
+Update equality for LoggingStore and WrapperStore such that 'other' must also be a LoggingStore or WrapperStore respectively, rather than only checking the types of the stores they wrap.
 Indicate StoreTest's `test_store_repr`, `test_store_supports_writes`, `test_store_supports_partial_writes`, and `test_store_supports_listing` need to be implemented using `@abstractmethod` rather than `NotImplementedError`.
 Separate instantiating and opening a store in StoreTests
 Test using Store as a context manager in StoreTests
--- a/src/zarr/storage/_fsspec.py
+++ b/src/zarr/storage/_fsspec.py
@@ -256,7 +256,7 @@ class FsspecStore(Store):
         self._check_writable()
         if not isinstance(value, Buffer):
             raise TypeError(
-                f"FsspecStore.set(): `value` must a Buffer instance. Got an instance of {type(value)} instead."
+                f"FsspecStore.set(): `value` must be a Buffer instance. Got an instance of {type(value)} instead."
             )
         path = _dereference_path(self.path, key)
         # write data
--- a/src/zarr/storage/_fsspec.py
+++ b/src/zarr/storage/_fsspec.py
@@ -10,6 +10,7 @@ from zarr.abc.store import (
     Store,
     SuffixByteRequest,
 )
+from zarr.core.buffer import Buffer
 from zarr.storage._common import _dereference_path
 
 if TYPE_CHECKING:
@@ -17,7 +18,7 @@ if TYPE_CHECKING:
 
     from fsspec.asyn import AsyncFileSystem
 
-    from zarr.core.buffer import Buffer, BufferPrototype
+    from zarr.core.buffer import BufferPrototype
     from zarr.core.common import BytesLike
 
 
@@ -253,6 +254,8 @@ class FsspecStore(Store):
         if not self._is_open:
             await self._open()
         self._check_writable()
+        if not isinstance(value, Buffer):
+            raise TypeError("FsspecStore.set(): `value` must a Buffer instance")
         path = _dereference_path(self.path, key)
         # write data
         if byte_range:
--- a/src/zarr/storage/_fsspec.py
+++ b/src/zarr/storage/_fsspec.py
@@ -173,6 +173,17 @@ class FsspecStore(Store):
         opts = {"asynchronous": True, **opts}
 
         fs, path = url_to_fs(url, **opts)
+        if not fs.async_impl:
+            try:
+                from fsspec.implementations.asyn_wrapper import AsyncFileSystemWrapper
+
+                fs = AsyncFileSystemWrapper(fs)
+            except ImportError as e:
+                raise ImportError(
+                    f"The filesystem for URL '{url}' is synchronous, and the required "
+                    "AsyncFileSystemWrapper is not available. Upgrade fsspec to version "
+                    "2024.12.0 or later to enable this functionality."
+                ) from e
 
         # fsspec is not consistent about removing the scheme from the path, so check and strip it here
         # https://github.com/fsspec/filesystem_spec/issues/1722
--- a/src/zarr/storage/_fsspec.py
+++ b/src/zarr/storage/_fsspec.py
@@ -255,7 +255,9 @@ class FsspecStore(Store):
             await self._open()
         self._check_writable()
         if not isinstance(value, Buffer):
-            raise TypeError(f"FsspecStore.set(): `value` must a Buffer instance. Got an instance of {type(value)} instead.")
+            raise TypeError(
+                f"FsspecStore.set(): `value` must a Buffer instance. Got an instance of {type(value)} instead."
+            )
         path = _dereference_path(self.path, key)
         # write data
         if byte_range:
--- a/src/zarr/storage/_fsspec.py
+++ b/src/zarr/storage/_fsspec.py
@@ -255,7 +255,7 @@ class FsspecStore(Store):
             await self._open()
         self._check_writable()
         if not isinstance(value, Buffer):
-            raise TypeError("FsspecStore.set(): `value` must a Buffer instance")
+            raise TypeError(f"FsspecStore.set(): `value` must a Buffer instance. Got an instance of {type(value)} instead.")
         path = _dereference_path(self.path, key)
         # write data
         if byte_range:
--- a/src/zarr/storage/_logging.py
+++ b/src/zarr/storage/_logging.py
@@ -165,7 +165,7 @@ class LoggingStore(WrapperStore[Store]):
 
     def __eq__(self, other: object) -> bool:
         with self.log(other):
-            return self._store == other
+            return type(self) is type(other) and self._store.__eq__(other._store)  # type: ignore[attr-defined]
 
     async def get(
         self,
--- a/src/zarr/storage/_logging.py
+++ b/src/zarr/storage/_logging.py
@@ -22,7 +22,7 @@ if TYPE_CHECKING:
 T_Store = TypeVar("T_Store", bound=Store)
 
 
-class LoggingStore(WrapperStore[Store]):
+class LoggingStore(WrapperStore[T_Store]):
     """
     Store wrapper that logs all calls to the wrapped store.
 
@@ -45,7 +45,7 @@ class LoggingStore(WrapperStore[Store]):
 
     def __init__(
         self,
-        store: Store,
+        store: T_Store,
         log_level: str = "DEBUG",
         log_handler: logging.Handler | None = None,
     ) -> None:
--- a/src/zarr/storage/_logging.py
+++ b/src/zarr/storage/_logging.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import inspect
 import logging
+import sys
 import time
 from collections import defaultdict
 from contextlib import contextmanager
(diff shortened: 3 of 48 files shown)


src/zarr/storage/_fsspec.py
```diff
@@ -253,6 +254,8 @@ async def set(
         if not self._is_open:
             await self._open()
         self._check_writable()
+        if not isinstance(value, Buffer):
+            raise TypeError("FsspecStore.set(): `value` must a Buffer instance")
```
d-v-b, line 258:
```suggestion
            raise TypeError("FsspecStore.set(): `value` must a Buffer instance. Got an instance of {type(value)} instead.")
```

src/zarr/storage/_logging.py
```diff
@@ -18,6 +19,8 @@
 
     counter: defaultdict[str, int]
 
+T_Store = TypeVar("T_Store", bound=Store)
+
 
 class LoggingStore(WrapperStore[Store]):
```
d-v-b, line 25: Should `WrapperStore` be generic w.r.t `T_Store` here?

src/zarr/testing/store.py
```diff
@@ -86,16 +110,23 @@ def test_store_read_only(self, store: S) -> None:
...
         assert store.read_only == read_only
 
-    async def test_read_only_store_raises(self, store_kwargs: dict[str, Any]) -> None:
-        kwargs = {**store_kwargs, "read_only": True}
+    async def test_store_context_manager(self, open_kwargs: dict[str, Any]) -> None:
+        # Test that the context manager closes the store
+        with await self.store_cls.open(**open_kwargs) as store:
+            assert store._is_open
+            # Test trying to open an already open store
+            with pytest.raises(ValueError):
```
d-v-b, line 124: can we check that the error message in the `ValueError` has the expected content? We don't want this test to succeed because of a `ValueError` unrelated to the store being already open.

src/zarr/testing/store.py
```diff
@@ -86,16 +110,23 @@ def test_store_read_only(self, store: S) -> None:
...
+                await store._open()
+        assert not store._is_open
+
+    async def test_read_only_store_raises(self, open_kwargs: dict[str, Any]) -> None:
```
d-v-b, line 128: contrary to the name, this test doesn't seem to check that an exception is raised

src/zarr/testing/store.py
```diff
@@ -135,6 +154,26 @@ async def test_get(self, store: S, key: str, data: bytes, byte_range: ByteReques
         expected = data_buf[start:stop]
         assert_bytes_equal(observed, expected)
 
+    async def test_get_not_open(self, store_not_open: S) -> None:
```
d-v-b, line 157: this is rather surprising -- I would expect that a non-open store would not support IO of any kind. what exactly does `open` mean? cc @jhamman

src/zarr/testing/store.py
```diff
@@ -157,6 +196,37 @@ async def test_get_many(self, store: S) -> None:
...
+        keys_values = [(k, data_buf) for k in keys]
+        await store._set_many(keys_values)
+        expected = len(data_buf) * len(keys)
+        observed = await store.getsize_prefix("c")
+        assert observed == expected
+
+    async def test_getsize_raises(self, store: S) -> None:
+        """
+        Test the result of store.getsize().
+        """
```
d-v-b, line 226: I think the method name and the docstring don't quite match the behavior of the test

src/zarr/testing/store.py
```diff
@@ -169,6 +239,17 @@ async def test_set(self, store: S, key: str, data: bytes) -> None:
         observed = await self.get(store, key)
         assert_bytes_equal(observed, data_buf)
 
+    async def test_set_not_open(self, store_not_open: S) -> None:
+        """
+        Ensure that data can be written to the store that's not yet open using the store.set method.
+        """
```
d-v-b, line 245: same as https://github.com/zarr-developers/zarr-python/pull/2693/files#r1913045841

src/zarr/testing/store.py
```diff
@@ -181,6 +262,13 @@ async def test_set_many(self, store: S) -> None:
         for k, v in store_dict.items():
             assert (await self.get(store, k)).to_bytes() == v.to_bytes()
 
+    async def test_set_raises(self, store: S) -> None:
```
d-v-b, line 265:
```suggestion
    async def test_set_invalid_bufferself, store: S) -> None:
```

tests/test_store/test_core.py
```diff
@@ -4,12 +4,50 @@
...
+        root.create_group("foo")
+    store_path = StorePath(local_store, path="foo")
+    assert await contains_group(store_path, zarr_format=zarr_format) == write_group
+
+
+@pytest.mark.parametrize("write_array", [True, False])
+@pytest.mark.parametrize("zarr_format", [2, 3])
+async def test_contains_array(local_store, write_array: bool, zarr_format: ZarrFormat) -> None:
+    """
+    Test contains group method
```
d-v-b, line 31:
```suggestion
    Test that the contains array method correctly reports the existence of an array.
```

tests/test_store/test_core.py
```diff
@@ -4,12 +4,50 @@
...
-from zarr.storage._common import make_store_path
+from zarr.storage._common import contains_array, contains_group, make_store_path
 from zarr.storage._utils import normalize_path
 
 
+@pytest.mark.parametrize("write_group", [True, False])
+@pytest.mark.parametrize("zarr_format", [2, 3])
+async def test_contains_group(local_store, write_group: bool, zarr_format: ZarrFormat) -> None:
+    """
+    Test contains group method
```
d-v-b, line 18:
```suggestion
     Test that the contains_group method correctly reports the existence of a group.
```

d-v-b, line 18: can we parametrize this over `path`, ensuring that we check a level of nesting? e.g. `@pytest.mark.parametrize('path', ['foo', 'foo/bar'])`

d-v-b: and similarly for the `contains_array` tests

tests/test_store/test_core.py
```diff
@@ -4,12 +4,50 @@
...
+
+async def test_contains_invalid_format_raises(local_store) -> None:
+    """
+    Test contains_group and contains_array raise errors for invalid zarr_formats
+    """
+    store_path = StorePath(local_store)
+    with pytest.raises(ValueError):
+        assert await contains_group(store_path, zarr_format="3.0")
+    with pytest.raises(ValueError):
+        assert await contains_array(store_path, zarr_format="3.0")
```
d-v-b, line 48:
```suggestion
```
deduplicate

tests/test_store/test_core.py
```diff
@@ -56,10 +94,19 @@ async def test_make_store_path_store_path(
...
 
 
+async def test_store_path_invalid_mode_raises(tmpdir: LEGACY_PATH) -> None:
+    """
+    Test that ValueErrors are raise for invalid mode.
+    """
+    with pytest.raises(ValueError):
+        await StorePath.open(LocalStore(str(tmpdir), read_only=True), path=None, mode="w")
+    with pytest.raises(ValueError):
+        await StorePath.open(LocalStore(str(tmpdir), read_only=False), path=None, mode="x")
```
d-v-b, line 107: lets parametrize the test over `mode` instead of repeating nearly identical checks

tests/test_store/test_local.py
```diff
@@ -53,3 +54,17 @@ def test_creates_new_directory(self, tmp_path: pathlib.Path):
 
         store = self.store_cls(root=target)
         zarr.group(store=store)
+
+    def test_invalid_root_raises(self):
```
d-v-b, line 58: add a docstring explaining what this test is checking

tests/test_store/test_local.py
```diff
@@ -53,3 +54,17 @@ def test_creates_new_directory(self, tmp_path: pathlib.Path):
...
+        with pytest.raises(TypeError):
+            LocalStore(root=0)
+
+    async def test_get_with_prototype_default(self, store: LocalStore):
+        """
+        Ensure that data can be read using the default prototype method.
```
d-v-b, line 64:
```suggestion
        Ensure that data can be read via ``store.get`` if the prototype keyword argument is unspecified, i.e. set to ``None``.
```

d-v-b:
this looks great, I had some minor suggestions.

tests/test_store/test_core.py
maxrjones: this is a distinct check for contains_array rather than contains_group. I can parameterize these functions to make it more concise and clear.

tests/test_store/test_core.py
d-v-b: looks like we would need to parametrize over `(read_only, mode)` tuples

dstansby:
Thanks for this, and thanks for fixing some issues along the way! For the user-facing issues you fixed, can you add a short entry in `docs/release-notes.rst`, so users know what's been fixed?

maxrjones:
> Thanks for this, and thanks for fixing some issues along the way! For the user-facing issues you fixed, can you add a short entry in `docs/release-notes.rst`, so users know what's been fixed?

For sure, I'm hoping to find time to address the remaining comments within the next day or two. I'll add release notes at that time.

src/zarr/testing/store.py
maxrjones: It tests for a ValueError - https://github.com/maxrjones/zarr-python/blob/1e96600acc744a1ff4bef43fb8e4f313b02f6873/src/zarr/testing/store.py#L134-L143. Could you please clarify your suggestion?

d-v-b: that's my mistake -- I think I stopped reading after the first `assert` block.

d-v-b:
I don't think my questions about `open` / `closed` behavior should block this PR. I can spin that out into a separate issue.

maxrjones:
> I don't think my questions about `open` / `closed` behavior should block this PR. I can spin that out into a separate issue.

Sounds good, thanks. There's still a couple new untested lines still in this PR - I should have time to fix that tomorrow morning.

src/zarr/storage/_logging.py
```diff
@@ -18,8 +19,10 @@
 
     counter: defaultdict[str, int]
 
+T_Store = TypeVar("T_Store", bound=Store)
 
-class LoggingStore(WrapperStore[Store]):
+
+class LoggingStore(WrapperStore[T_Store]):
```
dstansby, line 25: Could this get a changelog entry mentioning `LoggingStore` is now a generic class?

src/zarr/testing/store.py
```diff
@@ -37,30 +38,53 @@ class StoreTests(Generic[S, B]):
...
-        raise NotImplementedError
-
+    @abstractmethod
     @pytest.fixture
     def store_kwargs(self) -> dict[str, Any]:
-        return {"read_only": False}
+        """Kwargs for instantiating a store"""
+        ...
+
+    @abstractmethod
```
dstansby, line 65: Could these new abstract methods get a changelog entry? I also presume this is a breaking change, because anyone implementing a child of `StoreTests` will now be forced to implement these methods?

src/zarr/testing/store.py
```diff
@@ -86,39 +110,38 @@ def test_store_read_only(self, store: S) -> None:
...
-        store = await self.store_cls.open(**store_kwargs)
+    async def test_store_open_read_only(self, open_kwargs: dict[str, Any], read_only: bool) -> None:
+        open_kwargs["read_only"] = read_only
+        store = await self.store_cls.open(**open_kwargs)
         assert store._is_open
         assert store.read_only == read_only
 
-    async def test_read_only_store_raises(self, store_kwargs: dict[str, Any]) -> None:
-        kwargs = {**store_kwargs, "read_only": True}
+    async def test_store_context_manager(self, open_kwargs: dict[str, Any]) -> None:
```
dstansby, line 119: Could you add a changelog entry listing the new tests that have been added?

src/zarr/testing/store.py
dstansby: Yes, this is also surprising to me!

tests/test_store/test_local.py
```diff
@@ -53,3 +54,20 @@ def test_creates_new_directory(self, tmp_path: pathlib.Path):
 
         store = self.store_cls(root=target)
         zarr.group(store=store)
+
+    def test_invalid_root_raises(self):
+        """
+        Test that a TypeError is raised when a non-str/Path type is used for the `root` argument
+        """
+        with pytest.raises(TypeError):
```
dstansby, line 62: Could this get a `match=` argument to match the error message?

dstansby (changes requested):
Thanks for this, big 👍 for improving our testing.

Two major points:
- Since `StoreTests` is public API, we should carefully document everything that's been changed/added in the release notes. I added some comments on this PR inline to changes I spotted that need a release note.
- Like @d-v-b, I am confused by https://github.com/zarr-developers/zarr-python/pull/2693#discussion_r1913045841 - would be good to discuss/resolve that comment

src/zarr/testing/store.py
maxrjones: Yes, I can mention it in the changelog entry. I do not consider this a breaking change because the previous 'NotImplementedError' caused these tests to fail unless they were implemented in the inheriting class. I think this is just a more Pythonic way to indicate that they need to be implemented.

maxrjones:
> Like @d-v-b, I am confused by https://github.com/zarr-developers/zarr-python/pull/2693#discussion_r1913045841 - would be good to discuss/resolve that comment

@d-v-b @jhamman could I join the developers meeting tomorrow to discuss the Store open and set behavior (xref https://github.com/zarr-developers/zarr-python/pull/2688)?

src/zarr/testing/store.py
maxrjones: 👍 I listed all new tests in StoreTests in https://github.com/zarr-developers/zarr-python/pull/2693/commits/0d6eccfba8c13a894ccb78b054b6c40644b154c2

martindurant:
Labeler is fixed now on main

changes/2693.bugfix.rst
```diff
@@ -0,0 +1,14 @@
+Implement open() for LoggingStore
+LoggingStore is now a generic class.
+Add _is_open property and setter for WrapperStore
```
dstansby, line 3: Since this is private API, no need to put it in the changelog (which is user facing).
```suggestion
```

changes/2693.bugfix.rst
```diff
@@ -0,0 +1,14 @@
...
+Use stdout rather than stderr as the default stream for LoggingStore
+Ensure that ZipStore is open before getting or setting any values
+Update equality for LoggingStore and WrapperStore such that the types must be equal.
```
dstansby, line 6:
```suggestion
Update equality for LoggingStore and WrapperStore such that the type of the store they wrap must be equal for themselves to be equal.
```

dstansby (commented):
Thanks for all the changelog updates! I added a couple of minor suggestions. I think https://github.com/zarr-developers/zarr-python/pull/2693/files#r1913045841 is an outstanding question that still needs resolving too?

Other than that, is this ready to go from your point of view @maxrjones?

changes/2693.bugfix.rst
maxrjones: Your suggestion isn't comprehensive. Previously, this would not raise an error:
`assert LoggingStore(local_store: LocalStore) == local_store: LocalStore`
Now it raises an error because the types are different.

Would `Update equality for LoggingStore and WrapperStore such 'other' must also be a LoggingStore or WrapperStore respectively, rather than only checking the types of the stores they wrap.` be more clear?

maxrjones:
> Thanks for all the changelog updates! I added a couple of minor suggestions. I think [#2693 (files)](https://github.com/zarr-developers/zarr-python/pull/2693/files#r1913045841) is an outstanding question that still needs resolving too?
> 
> Other than that, is this ready to go from your point of view @maxrjones?

Thanks! Yes, I believe this is ready to go aside from the open questions about `._is_open`. I will raise a dedicated issue for that discussion shortly.

changes/2693.bugfix.rst
dstansby: Ah my mistake - you suggestion sounds good

dstansby (approved):

jhamman (approved):
Thanks so much for doing this @maxrjones. This is a great improvement AND exposes some rough edges in our store API. The good news there is that I think there are some relatively easy improvements we can make to clean up those rough edges.

PR #2714, opened 2025-01-15
Clean up warning filters in tests

dstansby:
I'm hoping to clean up the warning filters in this PR, starting with making them all error, and then either fix the warnings that come up or add them back to the filters. Opening because I noticed some deprecation warnings in our tests that should probably either be explicitly caught or the code fixed.

Fixes https://github.com/zarr-developers/zarr-python/issues/2968


d-v-b (approved):
thanks for this @dstansby!

PR #2718, opened 2025-01-15
Return scalar when accessing zero dimensional array

### LINKED ISSUE #2713: Inconsistent `()` indexing/size behavior from v2->v3
### Zarr version

v3.0.0

### Numcodecs version

0.14.1

### Python Version

3.12

### Operating System

Mac

### Installation

`uv`

### Description

Previously something like `z[()]` where `z` is a `()` shape array would return a scalar and now returns a one-elem array e.g., `np.array(1)` instead of `np.float64(1.0)`

### Steps to reproduce

v3:
```python
import numpy as np
import zarr

z = zarr.open_group('foo.zarr')
z.create_array("bar", dtype=float, shape=())
z["bar"][...] = np.array(1) # why doesn't [:] work?
z["bar"][()] # array
```
v2:
```python
import numpy as np
import zarr

z = zarr.open_group('foo.zarr')
z.create_dataset("bar", dtype=float, shape=(), data=np.array(1))
z["bar"][()] # np.float
```

### Additional output

_No response_

brokkoli71:
- fixes #2713

0-dimensional arrays are now returning a scalar. Therefore, the return type of ``__getitem__`` changed
to NDArrayOrScalarLike. This change is to make the behavior of 0-dimensional arrays consistent with
``numpy`` scalars.

TODO:


brokkoli71:
do we want to return scalars when accessing zero dimensional arrays?

dstansby:
To copy what numpy does (see https://github.com/zarr-developers/zarr-python/issues/2713#issuecomment-2596252524), I think we do want to return a scalar.

d-v-b:
> This change is to make the behavior of 0-dimensional arrays consistent with numpy scalars while not breaking the output format of the NDArrayLike interface.

Unfortunately the most important thing here for users is that we match numpy's behavior for scalar indexing, which I think argues against returning a special object. So in the interest of numpy compatibility, we it might be better to widen the return type of `__getitem__` to include scalars. 

FWIW I don't like being shackled to numpy's array indexing API, broadly speaking it's actually bad for users, but people are used to it so we are stuck with it for now.

src/zarr/core/buffer/core.py
```diff
@@ -106,6 +107,9 @@ def __eq__(self, other: object) -> Self:  # type: ignore[explicit-override, over
         """
 
 
+NDArrayLike = Union["np.ScalarType", NDArrayLike]
```
d-v-b, line 110:
```suggestion
NDArrayLike = "np.ScalarType" | NDArrayLike
```

brokkoli71: Oh, I wanted to create `NDArrayOrScalarLike` to see what mypy thinks about it, to see what problems arise when `__getitem__` returns no scalars.

brokkoli71:
@d-v-b what do you think about ``NDArrayLikeOrScalar`` return type for ``__getitem__``? (better ideas for the name of ``NDArrayLikeOrScalar`` are also welcome)

normanrz (approved):

PR #2721, opened 2025-01-16
Add hatch command for html coverage report

maxrjones:
I found the html coverage report simpler to use for monitoring progress in https://github.com/zarr-developers/zarr-python/pull/2693 and am wondering if you would consider a hatch command for this, such as the one suggested in the PR.

I also propose adding a `run-pytest` alias for `run` because I heard that using `hatch run --env <env> run` to run the tests was a bit confusing to new contributors at the Pangeo hackday.


d-v-b:
+1 for an easy path to making an html coverage report. no strong feelings about the new alias for `run`, but if people found it confusing, we should definitely fix that.

dstansby (approved):
Thanks, this is a nice addition!

PR #2736, opened 2025-01-20
Use towncrier for changelog generation

dstansby:
This changes the system of adding changelog entries from updating a single `.rst` file, to adding individual `.rst` files to a `changes` directory. At release time, these are then all automatically collated using [`towncrier`](https://towncrier.readthedocs.io/en/stable/). This has the advantages of:

- Removing potential merge conflicts in `releaset-notes.rst` during development
- Preventing release notes accidentally ending up in a alread-released section (e.g., see https://github.com/zarr-developers/zarr-python/pull/2681 where the changelog is in the wrong place)

For the release manager this is not much extra effort. Before, the changelog version already needed manually updating and commiting, with this PR the changelog just needs generating using towncrier and commiting.

Fixes https://github.com/zarr-developers/zarr-python/issues/829


docs/developers/contributing.rst
```diff
@@ -335,11 +335,9 @@ Release procedure
 
 Pre-release
 """""""""""
-1. Make sure that all pull requests which will be
-   included in the release have been properly documented in
-   :file:`docs/release-notes.rst`.
-2. Rename the "Unreleased" section heading in :file:`docs/release-notes.rst`
-   to the version you are about to release.
+1. Make sure that all pull requests which will be included in the release
+   have been properly documented as changelog files in :file:`changes`.
```
d-v-b, line 339: can you elaborate on / provide examples for what structure a changelog file should have?

dstansby: Does this line in `changes/README.md` answer your question?

> Inside the file, please write a short description of what you have changed, and how it impacts users of `zarr-python`.

joshmoore:
Thanks, @dstansby! I love this. A few questions from my side since I've not used it in anger:

- do PRs without a snippet get some form of notification? could they?
- do the snippets get cleaned up? if so, when and by whom?
- is there a sense that we want longer messages now? can they include their own markup?

docs/developers/contributing.rst
d-v-b: yep, that's nice and clear

jhamman:
One thing I like about our current system is that I can go to the [latest changelog](https://zarr.readthedocs.io/en/latest/release-notes.html#unreleased) to see what is going into the next release. How would we do that using this system?

changes/1234.doc.rst
```diff
@@ -0,0 +1,2 @@
+Changed the machinery for creating changelog entries.
```
joshmoore, line 1: Shouldn't this file be named be #2736 or #829?

dstansby: Woops, thanks for catching

joshmoore:
> I guess they could do, but they get rendered the same as before (RST lists), so this PR doesn't change how we want to write the messages one way or another.

Sounds good, guess we just need to be careful of too complex files that inject chaos into the README. As @d-v-b, I imagine some form of style examples might be worth long-term.

Anyhoo, this rocks.

joshmoore (approved):

martindurant:
Note that https://github.com/zarr-developers/zarr-python/pull/2681 put its note in the right place before merge, but of course that changelog line will need dealing with when this gets merged.

dstansby:
Since I think I've dealt with all the feedback/suggestions, and there's no objections, should we merge this to avoid stuff getting further out of sync with new PRs?

PR #2738, opened 2025-01-21
Use unsigned bytes to back Buffer

### LINKED ISSUE #2735: Buffer uses signed bytes with v2 compressors
### Zarr version

3.0.1

### Numcodecs version

0.15.0

### Python Version

3.13.1

### Operating System

Fedora Rawhide

### Installation

Fedora package

### Description

I'm looking at https://github.com/cgohlke/imagecodecs/issues/123 and after fixing some imports and setting `zarr_format=2`, I can run many more tests, but several are failing with mismatched types, namely that `imagecodecs` compressors are expecting `uint8_t`, but are getting `signed char`.

I have traced this to [`Buffer` requiring `dtype='b'`](https://github.com/zarr-developers/zarr-python/blob/main/src/zarr/core/buffer/core.py#L142), along with [casts in `cpu.Buffer.from_bytes`](https://github.com/zarr-developers/zarr-python/blob/main/src/zarr/core/buffer/cpu.py#L95).

If I modify those checks/casts to use the _unsigned_ `dtype='B'`, then I can get `imagecodecs` tests to pass.

I see this came in with GPU support in #1967. Was this actually intentional, or was it more that no-one noticed that the NumPy `b` dtype is a _signed_ byte? It would seem odd to me that `bytes` would be treated as signed as in regular Python they are treated as unsigned (e.g., `b'\xff'[0] == 255`, not -1).

If this is intentional, then it seems like something that should be documented in the migration guide that would break compressors for `zarr_format=2`.

### Steps to reproduce

Install `imagecodecs`, modify tests to use `zarr.storage.MemoryStorage` instead of `zarr.MemoryStorage`, and set `zarr_format=2`, then run its tests.

### Additional output

_No response_

QuLogic:
This makes compressors consistent with v2, and seems more correct than signed bytes.

Fixes #2735

Note that this does not implement accepting both signed and unsigned, as there wasn't a conclusion in #2735 about that.

TODO:


QuLogic:
PS, going to ignore tests for now.

dstansby:
I just fixed uploading code coverage for the GPU tests at https://github.com/zarr-developers/zarr-python/pull/2767, so hopefully when I merge in `main` here it should improve the patch coverage 🤞

maxrjones:
Hi @QuLogic and @dstansby, thanks for working on this! Could I help at all with resolving the codecov issue? I'd be super excited to see this completed so to enable further trying out imagecodecs with Zarr V3 (e.g., https://github.com/maxrjones/virtual-tiff/pull/4).

jakirkham:
(clicking update branch in case that fixes it)

jakirkham:
Closing and reopening to hopefully fix RTD (it has a `git checkout` error)

jakirkham:
Ok better RTD checked out the code 😌

jakirkham:
Looks like Codecov got stuck somewhere. Let's try another close/reopen

maxrjones:
I don't think that the codecov/patch target will be hit unless the overall coverage actually increases, rather than remains constant. The issue is that two of the lines changed (L87 and L110) are not tested in main either - https://app.codecov.io/gh/zarr-developers/zarr-python/blob/main/src%2Fzarr%2Fcore%2Fbuffer%2Fgpu.py, so the PR isn't decreasing code coverage but still will not hit the 80% target for code coverage for patches (i.e., all new/modified lines of code). Do you prefer to ignore the codecov/patch target or to include unit tests for adding and creating zero length buffers?

jakirkham:
If someone can add some tests, that would be welcome

What was concerning to me was one Codecov job says "failing after 1s". Though this may be a non-issue and just the report percentage (as you suggested)

d-v-b:
thanks for pushing this along @maxrjones. Based on the changes in this PR, I'm happy merging this regardless of what codecov says at the moment.

d-v-b (approved):

d-v-b (approved):
to repeat what I said in the [linked issue](https://github.com/zarr-developers/zarr-python/issues/2735#issuecomment-2606669991), the particular numpy dtype we use in our buffer implementation should be considered an implementation detail by consumers. But for the same reason we can alter that dtype as needed, e.g. to make life easier for imagecodecs users, which is what this PR does. I'm going to approve and merge this despite the low test coverage because adding those tests depends on getting the gpu buffer to work with sharding, which is IMO out of scope for this PR.

PR #2744, opened 2025-01-22
Fix label action

dstansby:
This seems to have gone wrong in https://github.com/zarr-developers/zarr-python/pull/2736, so trying to fix here.

Because the job is defined using `pull_request_target`, it's run from the GH actions config on the `main` branch, so this will need merging to fix the job.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,4 +1,4 @@
-- needs release notes:
+needs release notes:
   - all:
     - changed-files:
       - any-glob-to-any-file: 'changes/*.rst'


moradology:
This proved to require an odd and detailed intervention. Assuming the efforts in #2747 are accepted, I think this PR can be closed

martindurant:
> Because the job is defined using pull_request_target, it's run from the GH actions config on the main branch, so this will need merging to fix the job.

I like the simplicity of this fix, but how do we know if it works? Should we merge and see what happens?

dstansby:
Yes, it will need merging to verify if it works .

moradology:
Can technically swap over to target `pull_request` but I'd say just pull the trigger and clean up later

jhamman:
Can we get this merged now? The failing labeler is causing CI checks to fail on all PRs.

dstansby:
Sure, it just needs a maintainer to review/approve/merge.

d-v-b:
> @d-v-b , will auto-merge happen if the labeler run fails? (as it will, since it won't take action until its in main)

I was wondering the same thing. we shall see!

dstansby:
Looks like this worked: https://github.com/zarr-developers/zarr-python/actions/runs/12951743813/job/36127460117?pr=2714 🎉

PR #2746, opened 2025-01-22
Test labeler workflow fix

moradology:
Testing the workflow behavior, will delete PR after verification


moradology:
This was annoying. Have to temporarily change to 'pull_request' and THEN to 'pull_request_target'. Fixing in separate (clean) PR

PR #2747, opened 2025-01-22
Update labeler config for major version increment

moradology:
This PR simply prepares the repo to upgrade the major version of the labeler workflow (and fix all the workflow errors people are encountering). See docs here: https://github.com/actions/labeler?tab=readme-ov-file#updating-major-version-of-the-labeler

1. Once verified (PRIOR TO MERGING) `pull_request` on line 4 of `needs_release_notes.yml` will need to be reverted to `pull_request_target`. This *must* happen to avoid permissions issues! Cf. https://github.com/actions/labeler?tab=readme-ov-file#recommended-permissions
2. At this point things will fail again (the changes below won't exist so far as the github action is concerned). Ignore the failure. Merge.

This addresses the issue mentioned in #2744 and the unfortunate bug hanging up #2533 (among others)

### DIFF
--- a/.github/workflows/needs_release_notes.yml
+++ b/.github/workflows/needs_release_notes.yml
@@ -1,7 +1,7 @@
 name: "Pull Request Labeler"
 
 on:
-  - pull_request_target
+  pull_request
 
 jobs:
   labeler:
@@ -11,7 +11,12 @@ jobs:
       pull-requests: write
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/labeler@8558fd74291d67161a8a78ce36a881fa63b766a9  # v5.0.0
+    - uses: actions/checkout@v4
+      with:
+        ref: ${{ github.event.pull_request.head.sha }}
+
+    - name: Run Pull Request Labeler
+      uses: actions/labeler@v5.0.0
       with:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
         sync-labels: true
--- a/.github/workflows/needs_release_notes.yml
+++ b/.github/workflows/needs_release_notes.yml
@@ -1,7 +1,7 @@
 name: "Pull Request Labeler"
 
 on:
-  pull_request
+  pull_request_target
 
 jobs:
   labeler:
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,4 +1,3 @@
-- needs release notes:
-  - all:
-    - changed-files:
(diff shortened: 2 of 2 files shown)


moradology:
OK, here's verification:
image
image

moradology:
This was kind of confusing, so I'm going to avoid switching back to `pull_request_target` until I get some verification that my explanation/the validation of the fix here is acceptable

martindurant:
I also tried something similar in my PR, following from discussion I found online. I have no idea what's really going on! It would be OK by me to merge something to see if it gets fixed, without 100% certainty, since it's definitely broken right now.

moradology:
OK, switched over to `pull_request_target`. This can be merged, at which point it *should* work in other PRs

martindurant:
@dstansby , perhaps?

.github/workflows/needs_release_notes.yml
```diff
@@ -11,7 +11,12 @@ jobs:
       pull-requests: write
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/labeler@8558fd74291d67161a8a78ce36a881fa63b766a9  # v5.0.0
+    - uses: actions/checkout@v4
+      with:
+        ref: ${{ github.event.pull_request.head.sha }}
```
dstansby, line 16: Why do we need to checkout the repo?

.github/workflows/needs_release_notes.yml
```diff
@@ -11,7 +11,12 @@ jobs:
...
+
+    - name: Run Pull Request Labeler
+      uses: actions/labeler@v5.0.0
```
dstansby, line 19: Could we switch back to using the SHA instead? It will be updated find by dependabot, but it's generally recommended to use the SHA instead of a tag for security reasons.

dstansby:
I think a slightly simpler fix is at https://github.com/zarr-developers/zarr-python/pull/2744 - if another maintainer wants to review and merge that instead I think it would work? Otherwise I left a couple of comments above on this PR, and I'd be happy to merge this PR once they're resolved.

.github/workflows/needs_release_notes.yml
moradology: Failure to find `.github/labeler.yml`, which was the first error being encountered on #2533 and led me to looking into this

martindurant:
Thanks for working on this @moradology - it looks like it's fixed now on main

PR #2750, opened 2025-01-22
Prototype of new DType interface

nenb:
This is a PoC for a new standalone interface for managing dtypes in the `zarr-python` codebase. 

**Among other things, it tries to provide support for adding dtype extensions, as outlined by the Zarr V3 spec.**

I have provided several examples below of how it is currently intended to be used. I will also add a number of comments to the files in this PR to help further clarify intent.

### Core dtype registration

The following example demonstrates how to register a built-in `dtype` in the core codebase:

```python
from zarr.core.dtype import ZarrDType
from zarr.registry import register_v3dtype

class Float16(ZarrDType):
    zarr_spec_format = "3"
    experimental = False
    endianness = "little"
    byte_count = 2
    to_numpy = np.dtype('float16')

register_v3dtype(Float16)
```

### Entrypoint extension

The following example demonstrates how users can register a new `bfloat16` dtype for Zarr. This approach adheres to the existing Zarr entrypoint pattern as much as possible, ensuring consistency with other extensions. The code below would typically be part of a Python package that specifies the entrypoints for the extension:

```python
import ml_dtypes
import numpy as np
from zarr.core.dtype import ZarrDType  # User inherits from ZarrDType when creating their dtype

class Bfloat16(ZarrDType):
    zarr_spec_format = "3"
    experimental = True
    endianness = "little"
    byte_count = 2
    to_numpy = np.dtype('bfloat16')  # Enabled by importing ml_dtypes
    configuration_v3 = {
        "version": "example_value",
        "author": "example_value",
        "ml_dtypes_version": "example_value"
    }
```

### dtype lookup

The following examples demonstrate how to perform a lookup for the relevant `ZarrDType` inside the `zarr-python` codebase, given a string that matches the Zarr specification ID for the dtype, or a `numpy` dtype object:

```python
from zarr.registry import get_v3dtype_class, get_v3dtype_class_from_numpy

get_v3dtype_class('complex64')  # returns little-endian Complex64 ZarrDType
get_v3dtype_class('not_registered_dtype')  # ValueError

get_v3dtype_class_from_numpy('>i2')  # returns big-endian Int16 ZarrDType
get_v3dtype_class_from_numpy(np.dtype('float32'))  # returns little-endian Float32 ZarrDType
get_v3dtype_class_from_numpy('i10')  # ValueError
```

### String dtypes

The following indicates one possibility for supporting variable-length strings. It is via the entrypoint mechanism as in a previous example. The Apache Arrow specification does not currently include a dtype for fixed-length strings (only for fixed-length bytes) and so I am using string here to implicitly refer to a variable-length string data (there may be some subtleties with codecs that means this needs to be refined further):

```python
import numpy as np
from zarr.core.dtype import ZarrDType  # User inherits from ZarrDType when creating their dtype

try:
    to_numpy = np.dtypes.StringDType()
except AttributeError:
    to_numpy = np.dtypes.ObjectDType()

class String(ZarrDType):
    zarr_spec_format = "3"
    experimental = True
    endianness = 'little'
    byte_count = None  # None is defined to mean variable
    to_numpy = to_numpy
```

### int4 dtype

There is currently considerable interest in the AI community in 'quantising' models - storing models at reduced precision, while minimising loss of information content. There are a number of sub-byte dtypes that the community are using e.g. int4. Unfortunately `numpy` does not currently have much support for sub-byte dtypes. However, raw bits/bytes can still be held in a `numpy` array and then passed (in a zero-copy way) to something like `pytorch` which can handle more conveniently:

```python
import numpy as np
from zarr.core.dtype import ZarrDType  # User inherits from ZarrDType when creating their dtype

class Int4(ZarrDType):
    zarr_spec_format = "3"
    experimental = True
    endianness = 'little'
    byte_count = 1  # this is ugly, but I could change this from byte_count to bit_count if there was consensus 
    to_numpy = np.dtype('B')  # could also be np.dtype('V1'), but this would prevent bit-twiddling
    configuration_v3 = {
        "version": "example_value",
        "author": "example_value",
    }
```

### DIFF
--- a/src/zarr/core/dtype/core.py
+++ b/src/zarr/core/dtype/core.py
@@ -1,0 +1,204 @@
+"""
+# Overview
+
+This module provides a proof-of-concept standalone interface for managing dtypes in the zarr-python codebase. 
+
+The `ZarrDType` class introduced in this module effectively acts as a replacement for `np.dtype` throughout the
+zarr-python codebase. It attempts to encapsulate all relevant runtime information necessary for working with
+dtypes in the context of the Zarr V3 specification (e.g. is this a core dtype or not, how many bytes and what
+endianness is the dtype etc). By providing this abstraction, the module aims to:
+
+- Simplify dtype management within zarr-python
+- Support runtime flexibility and custom extensions
+- Remove unnecessary dependencies on the numpy API
+
+## Extensibility
+
+The module attempts to support user-driven extensions, allowing developers to introduce custom dtypes
+without requiring immediate changes to zarr-python. Extensions can leverage the current entrypoint mechanism, 
+enabling integration of experimental features. Over time, widely adopted extensions may be formalized through
+inclusion in zarr-python or standardized via a Zarr Enhancement Proposal (ZEP), but this is not essential.
+
+## Examples
+
+### Core `dtype` Registration
+
+The following example demonstrates how to register a built-in `dtype` in the core codebase:
+
+```python
+from zarr.core.dtype import ZarrDType
+from zarr.registry import register_v3dtype
+
+class Float16(ZarrDType):
+    zarr_spec_format = "3"
+    experimental = False
+    endianness = "little"
+    byte_count = 2
+    to_numpy = np.dtype('float16')
+
+register_v3dtype(Float16)
+```
+
+### Entrypoint Extension
+
+The following example demonstrates how users can register a new `bfloat16` dtype for Zarr.
+This approach adheres to the existing Zarr entrypoint pattern as much as possible, ensuring
+consistency with other extensions. The code below would typically be part of a Python package
+that specifies the entrypoints for the extension:
+
+```python
+import ml_dtypes
+from zarr.core.dtype import ZarrDType  # User inherits from ZarrDType when creating their dtype
+
+class Bfloat16(ZarrDType):
+    zarr_spec_format = "3"
+    experimental = True
+    endianness = "little"
+    byte_count = 2
+    to_numpy = np.dtype('bfloat16')  # Enabled by importing ml_dtypes
+    configuration_v3 = {
+        "version": "example_value",
+        "author": "example_value",
+        "ml_dtypes_version": "example_value"
+    }
+```
+
+### dtype lookup
+
+The following examples demonstrate how to perform a lookup for the relevant ZarrDType, given
+a string that matches the dtype Zarr specification ID, or a numpy dtype object:
+
+```
+from zarr.registry import get_v3dtype_class, get_v3dtype_class_from_numpy
+
+get_v3dtype_class('complex64')  # returns little-endian Complex64 ZarrDType
+get_v3dtype_class('not_registered_dtype')  # ValueError
+
+get_v3dtype_class_from_numpy('>i2')  # returns big-endian Int16 ZarrDType
+get_v3dtype_class_from_numpy(np.dtype('float32'))  # returns little-endian Float32 ZarrDType
+get_v3dtype_class_from_numpy('i10')  # ValueError
+```
+
+### String dtypes
+
+The following indicates one possibility for supporting variable-length strings. It is via the
+entrypoint mechanism as in a previous example. The Apache Arrow specification does not currently
+include a dtype for fixed-length strings (only for fixed-length bytes) and so I am using string
+here to implicitly refer to a variable-length string data (there may be some subtleties with codecs
+that means this needs to be refined further):
+
+```python
+import numpy as np
+from zarr.core.dtype import ZarrDType  # User inherits from ZarrDType when creating their dtype
+
+try:
+    to_numpy = np.dtypes.StringDType()
+except AttributeError:
+    to_numpy = np.dtypes.ObjectDType()
+
+class String(ZarrDType):
+    zarr_spec_format = "3"
+    experimental = True
+    endianness = 'little'
+    byte_count = None  # None is defined to mean variable
+    to_numpy = to_numpy
+```
+
+### int4 dtype
+
+There is currently considerable interest in the AI community in 'quantising' models - storing
+models at reduced precision, while minimising loss of information content. There are a number
+of sub-byte dtypes that the community are using e.g. int4. Unfortunately numpy does not
+currently have support for handling such sub-byte dtypes in an easy way. However, they can
+still be held in a numpy array and then passed (in a zero-copy way) to something like pytorch
+which can handle appropriately:
+
+```python
+import numpy as np
+from zarr.core.dtype import ZarrDType  # User inherits from ZarrDType when creating their dtype
+
+class Int4(ZarrDType):
+    zarr_spec_format = "3"
+    experimental = True
+    endianness = 'little'
+    byte_count = 1  # this is ugly, but I could change this from byte_count to bit_count if there was consensus 
+    to_numpy = np.dtype('B')  # could also be np.dtype('V1'), but this would prevent bit-twiddling
+    configuration_v3 = {
+        "version": "example_value",
+        "author": "example_value",
+    }
+```
+"""
+
+from __future__ import annotations
+
+from typing import Any, Literal
+
+import numpy as np
+
+
+# perhaps over-complicating, but I don't want to allow the attributes to be patched
+class FrozenClassVariables(type):
+    def __setattr__(cls, attr, value):
+        if hasattr(cls, attr):
+            raise ValueError(
+                f"Attribute {attr} on ZarrDType class can not be changed once set."
+            )
+
+
+class ZarrDType(metaclass=FrozenClassVariables):
+
+    zarr_spec_format: Literal["2", "3"]  # the version of the zarr spec used
+    experimental: bool  # is this in the core spec or not
+    endianness: Literal[
+        "big", "little", None
+    ]  # None indicates not defined i.e. single byte or byte strings
+    byte_count: int | None  # None indicates variable count
+    to_numpy: np.dtype[
+        Any
+    ]  # may involve installing a a numpy extension e.g. ml_dtypes;
+
+    configuration_v3: (
+        dict | None
+    )  # TODO: understand better how this is recommended by the spec
+
+    _zarr_spec_identifier: str  # implementation detail used to map to core spec
+
+    def __init_subclass__(  # enforces all required fields are set and basic sanity checks
+        cls,
+        **kwargs,
+    ) -> None:
+
+        required_attrs = [
+            "zarr_spec_format",
+            "experimental",
+            "endianness",
+            "byte_count",
+            "to_numpy",
+        ]
+        for attr in required_attrs:
+            if not hasattr(cls, attr):
+                raise ValueError(f"{attr} is a required attribute for a Zarr dtype.")
+
+        if not hasattr(cls, "configuration_v3"):
+            cls.configuration_v3 = None
+
+        cls._zarr_spec_identifier = (
+            "big_" + cls.__qualname__.lower()
+            if cls.endianness == "big"
+            else cls.__qualname__.lower()
+        )  # how this dtype is identified in core spec; convention is prefix with big_ for big-endian
+
+        cls._validate()  # sanity check on basic requirements
+
+        super().__init_subclass__(**kwargs)
+
+    # TODO: add further checks
+    @classmethod
+    def _validate(cls):
+
+        if cls.byte_count is not None and cls.byte_count <= 0:
+            raise ValueError("byte_count must be a positive integer.")
+
+        if cls.byte_count == 1 and cls.endianness is not None:
+            raise ValueError("Endianness must be None for single-byte types.")
--- a/src/zarr/registry.py
+++ b/src/zarr/registry.py
@@ -5,10 +5,14 @@ from collections import defaultdict
 from importlib.metadata import entry_points as get_entry_points
 from typing import TYPE_CHECKING, Any, Generic, TypeVar
 
+import numpy as np
+
 from zarr.core.config import BadConfigError, config
 
 if TYPE_CHECKING:
     from importlib.metadata import EntryPoint
+
+    import numpy.typing as npt
 
     from zarr.abc.codec import (
         ArrayArrayCodec,
@@ -19,6 +23,7 @@ if TYPE_CHECKING:
     )
     from zarr.core.buffer import Buffer, NDBuffer
     from zarr.core.common import JSON
+    from zarr.core.dtype import ZarrDType
 
 __all__ = [
     "Registry",
@@ -26,10 +31,14 @@ __all__ = [
     "get_codec_class",
     "get_ndbuffer_class",
     "get_pipeline_class",
+    "get_v3dtype_class",
+    "get_v2dtype_class",
     "register_buffer",
     "register_codec",
     "register_ndbuffer",
     "register_pipeline",
+    "register_v3dtype",
+    "register_v2dtype",
 ]
 
 T = TypeVar("T")
@@ -42,28 +51,52 @@ class Registry(dict[str, type[T]], Generic[T]):
 
     def lazy_load(self) -> None:
         for e in self.lazy_load_list:
-            self.register(e.load())
+            cls = e.load()
+
+            if (hasattr(cls, "_zarr_spec_identifier") and hasattr(cls, "to_numpy")):
+                for val in self.values():
+                    # allow the V and B dtypes to be overloaded - something of an escape hatch.
+                    if val.to_numpy == cls.to_numpy and cls.to_numpy.kind not in ["V", "B"]:
+                        raise ValueError(
+                            f"numpy dtype {cls.to_numpy} already exists for Zarr dtype {val}."
+                        )
+
+                self.register(cls, cls._zarr_spec_identifier)
+            else:
+                self.register(cls)
+
         self.lazy_load_list.clear()
 
-    def register(self, cls: type[T]) -> None:
-        self[fully_qualified_name(cls)] = cls
+    def register(self, cls: type[T], class_registry_key: str | None = None) -> None:
+        if class_registry_key is None:
+            self[fully_qualified_name(cls)] = cls
+        else:
+            if class_registry_key in self:
+                raise ValueError(
+                    f"{class_registry_key} already exists in the registry. Please try a different name."
+                )  # not great, but don't want to have the possibility of clobbering existing core dtypes
+            self[class_registry_key] = cls
 
 
 __codec_registries: dict[str, Registry[Codec]] = defaultdict(Registry)
 __pipeline_registry: Registry[CodecPipeline] = Registry()
 __buffer_registry: Registry[Buffer] = Registry()
 __ndbuffer_registry: Registry[NDBuffer] = Registry()
+__v3_dtype_registry: Registry[ZarrDType] = Registry()
+__v2_dtype_registry: Registry[ZarrDType] = Registry()
 
 """
 The registry module is responsible for managing implementations of codecs,
 pipelines, buffers and ndbuffers and collecting them from entrypoints.
 The implementation used is determined by the config.
+
+The registry module is also responsible for managing dtypes.
 """
 
 
 def _collect_entrypoints() -> list[Registry[Any]]:
     """
-    Collects codecs, pipelines, buffers and ndbuffers from entrypoints.
+    Collects codecs, pipelines, dtypes, buffers and ndbuffers from entrypoints.
     Entry points can either be single items or groups of items.
     Allowed syntax for entry_points.txt is e.g.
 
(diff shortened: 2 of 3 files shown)


src/zarr/core/dtype/core.py
```diff
@@ -0,0 +1,204 @@
+"""
```
nenb, line 1: Mostly the same information as what is included in the PR conversation.

src/zarr/core/dtype/core.py
```diff
@@ -0,0 +1,204 @@
...
+
+from __future__ import annotations
+
+from typing import Any, Literal
+
+import numpy as np
+
+
+# perhaps over-complicating, but I don't want to allow the attributes to be patched
```
nenb, line 140: Implementation detail: I decided that to try and freeze the class attributes after a dtype has been created. I used metaclasses for this. It's not essential.

src/zarr/core/dtype/core.py
```diff
@@ -0,0 +1,204 @@
...
+            raise ValueError(
+                f"Attribute {attr} on ZarrDType class can not be changed once set."
+            )
+
+
+class ZarrDType(metaclass=FrozenClassVariables):
+
+    zarr_spec_format: Literal["2", "3"]  # the version of the zarr spec used
+    experimental: bool  # is this in the core spec or not
+    endianness: Literal[
```
nenb, line 153: Zarr V3 has made the decision to use a codec for endianness. Endianness is not to be attached to the dtype.
This creates some problems for the Zarr API, which is still linked to `numpy`'s API in a number of ways, including
the ability to create in-memory arrays of arbitrary endianness. 

Currently, I think that the practical solution is for `zarr-python` to have dtypes that distinguish between big and little endianess in memory, but that when serialised to disk, always serialise the little endian dtype. 

I can elaborate on this with examples if helpful, but basically, endianness would just be an implementation detail for `zarr-python` that would allow it to track the endianness of an object in memory, and it wouldn't actually be used when serialising to disk.

src/zarr/core/dtype/core.py
```diff
@@ -0,0 +1,204 @@
...
+        "big", "little", None
+    ]  # None indicates not defined i.e. single byte or byte strings
+    byte_count: int | None  # None indicates variable count
+    to_numpy: np.dtype[
```
nenb, line 157: See the `bfloat16` example about how this might require new packages to be installed.

src/zarr/core/dtype/core.py
```diff
@@ -0,0 +1,204 @@
...
+        Any
+    ]  # may involve installing a a numpy extension e.g. ml_dtypes;
+
+    configuration_v3: (
```
nenb, line 161: Wasn't clear to me how this is intended to be used in the spec...

src/zarr/core/dtype/core.py
```diff
@@ -0,0 +1,204 @@
...
+        dict | None
+    )  # TODO: understand better how this is recommended by the spec
+
+    _zarr_spec_identifier: str  # implementation detail used to map to core spec
+
+    def __init_subclass__(  # enforces all required fields are set and basic sanity checks
```
nenb, line 167: Implementation detail: I thought it would be helpful to prevent class creation unless all attributes were defined.

src/zarr/core/dtype/core.py
```diff
@@ -0,0 +1,204 @@
...
+            "to_numpy",
+        ]
+        for attr in required_attrs:
+            if not hasattr(cls, attr):
+                raise ValueError(f"{attr} is a required attribute for a Zarr dtype.")
+
+        if not hasattr(cls, "configuration_v3"):
+            cls.configuration_v3 = None
+
+        cls._zarr_spec_identifier = (
```
nenb, line 186: Again, the way I am proposing endianness is just as an implementation detail for `zarr-python` to track the endianness of in-memory objects. When serialised to disk, this `big_` prefix would always be removed.

src/zarr/core/dtype/core.py
```diff
@@ -0,0 +1,204 @@
...
+            else cls.__qualname__.lower()
+        )  # how this dtype is identified in core spec; convention is prefix with big_ for big-endian
+
+        cls._validate()  # sanity check on basic requirements
+
+        super().__init_subclass__(**kwargs)
+
+    # TODO: add further checks
+    @classmethod
+    def _validate(cls):
```
nenb, line 198: Just an example of the sort of validation that could happen.

src/zarr/registry.py
```diff
@@ -5,11 +5,15 @@
 from importlib.metadata import entry_points as get_entry_points
 from typing import TYPE_CHECKING, Any, Generic, TypeVar
 
+import numpy as np
```
nenb, line 8: The changes in this module are a bit hacky. They are mainly just to show how something might work - the actual details can certainly be changed quite a bit.

src/zarr/registry.py
```diff
@@ -42,28 +51,52 @@ def __init__(self) -> None:
 
     def lazy_load(self) -> None:
         for e in self.lazy_load_list:
-            self.register(e.load())
+            cls = e.load()
+
+            if (hasattr(cls, "_zarr_spec_identifier") and hasattr(cls, "to_numpy")):
```
nenb, line 56: The motivation here is that I don't want a user to be able to add a dtype whose `numpy` representation overrides an already registered dtype e.g. MyFakeBool should not be able to override the core Bool dtype.

The exceptions are for the V and B dtypes, because these are raw byte strings, and multiple dtypes might be mapped to this e.g. `int2` and `int4` might both be mapped to the 'B' dtype in `numpy` because `numpy` has no support for sub-byte dtypes.

src/zarr/registry.py
```diff
@@ -131,6 +182,16 @@ def register_buffer(cls: type[Buffer]) -> None:
     __buffer_registry.register(cls)
 
 
+def register_v3dtype(cls: type[ZarrDType]) -> None:
+    assert cls.zarr_spec_format == "3"
+    __v3_dtype_registry.register(cls, class_registry_key=cls._zarr_spec_identifier)
+
+
+def register_v2dtype(cls: type[ZarrDType]) -> None:
```
nenb, line 190: Ideally, we could also backport this to Zarr V2? I'll admit that I haven't really researched this much though. Putting it here just to start a discussion.

src/zarr/registry.py
```diff
@@ -266,4 +328,50 @@ def get_ndbuffer_class(reload_config: bool = False) -> type[NDBuffer]:
     )
 
 
+# TODO: merge the get_vXdtype_class_ functions
+# these can be used instead of the various parse_X functions (hopefully)
+def get_v3dtype_class(dtype: str) -> type[ZarrDType]:
```
nenb, line 333: See the examples for how these can be used. As in the comment, the idea is for these to replace the various `parse_X` functions in the codebase.

d-v-b:
@nenb thank you so much for working on this. I really like the ideas in here -- it looks like a great addition to the library, and I will try to give this close examination, and I recommend the other @zarr-developers/python-core-devs  do the same. This is a big, forward-looking change so I imagine we will have a lot of opinions on multiple scales here. 

Our weekly dev meeting is tomorrow (here's the [calendar info](https://zarr.dev/community-calls/), the particular meeting I'm referring to is the [zarr-python developers meeting](https://calendar.google.com/calendar/u/0/event?eid=ZnBwaDg1MHFmZ3ZnOGFoaWludXVqcWFyY2xfMjAyNTAxMTBUMTUwMDAwWiBjX2JhMms3OWkzdTBsa2Y0OXZvMGpyZTI3ajE0QGc)), if the timing works for you, would it be possible for you to attend and give an overview of this PR?

src/zarr/core/dtype/core.py
normanrz: Basically, dtypes can be represented in the json metadata as a short-hand (`str`) or dict (`{ "name": str, "configuration": None | { ... } }`. The `configuration` key is optional and could be used for dtypes that need additional configuration. If there is no `configuration` key, the short-hand version is equivalent to the dict with just a `name` key.

normanrz: For example, `bfloat16` is equivalent to `{"name":"bfloat16"}`

src/zarr/core/dtype/core.py
```diff
@@ -0,0 +1,204 @@
...
+    def __setattr__(cls, attr, value):
+        if hasattr(cls, attr):
+            raise ValueError(
+                f"Attribute {attr} on ZarrDType class can not be changed once set."
+            )
+
+
+class ZarrDType(metaclass=FrozenClassVariables):
+
+    zarr_spec_format: Literal["2", "3"]  # the version of the zarr spec used
```
normanrz, line 151:
```suggestion
    zarr_format: Literal["2", "3"]  # the version of the zarr spec used
```

src/zarr/core/dtype/core.py
```diff
@@ -0,0 +1,204 @@
...
+    byte_count: int | None  # None indicates variable count
+    to_numpy: np.dtype[
+        Any
+    ]  # may involve installing a a numpy extension e.g. ml_dtypes;
+
+    configuration_v3: (
+        dict | None
+    )  # TODO: understand better how this is recommended by the spec
+
+    _zarr_spec_identifier: str  # implementation detail used to map to core spec
```
normanrz, line 165: I guess this would be mandatory? Or, how would the identifier for the metadata be specified otherwise?

nenb: **TL;DR** 
This isn't an essential detail, and the implementation can probably be considerably improved if there is consensus around adding a new dtype interface. I've included some details below about how I ended up with this attribute if you are interested though.

**Details**
At the moment, this is generated from the class name (see the logic in `__init_subclass__`). The user doesn't specify it, rather the user specifies the class name and the identifier is generated from the class name (potentially prefixed with `big_`).

**Example**

```python
class Float16(ZarrDType):
    zarr_spec_format = "3"
    experimental = False
    endianness = "big"
    byte_count = 2
    to_numpy = np.dtype('float16')
```

This would generate `big_float16` for `_zarr_spec_identifier`.

-------------------------

It's probably a bit clumsy in it's current implementation. I ended up with this pattern i) as a way of tracking the in-memory endianness of the dtype and ii) making sure the class name stays consistent with the identifier - the class name pops up a few types in the entrypoint registration logic, and I didn't want to have a situation where a class name could have a different value to the spec identifier.

Obviously, the convention is that a class needs to be named according to how the dtype is identified in the spec.

src/zarr/core/dtype/core.py
nenb: Ah, interesting, thanks!

My current thinking is that every dtype that is not in the core spec _should include_ a `configuration` key. I would like to introduce a convention where extension dtypes also provide metadata like 'author', 'extension_version', etc., to give the best chance of reproducibility/re-use in the future. At least, until an extension dtype becomes a core dtype.

Is the `configuration` key an appropriate location for such metadata?

nenb:
> @nenb thank you so much for working on this. I really like the ideas in here -- it looks like a great addition to the library, and I will try to give this close examination, and I recommend the other @zarr-developers/python-core-devs do the same. This is a big, forward-looking change so I imagine we will have a lot of opinions on multiple scales here.
> 
> Our weekly dev meeting is tomorrow (here's the [calendar info](https://zarr.dev/community-calls/), the particular meeting I'm referring to is the [zarr-python developers meeting](https://calendar.google.com/calendar/u/0/event?eid=ZnBwaDg1MHFmZ3ZnOGFoaWludXVqcWFyY2xfMjAyNTAxMTBUMTUwMDAwWiBjX2JhMms3OWkzdTBsa2Y0OXZvMGpyZTI3ajE0QGc)), if the timing works for you, would it be possible for you to attend and give an overview of this PR?

@d-v-b Thanks for being so receptive! I'm happy to attend and give an overview.

What's likely the best format to present at the meeting (I haven't attended before). If I give a 5 minute overview (including some motivation from the AI-side of things), and then leave 10 minutes for questions, would that work? Is there anything specific that you think it's worth focusing on e.g. extension mechanism?

d-v-b:
@nenb that plan sounds good. I'm looking forward to it!

src/zarr/registry.py
normanrz: I don't think there is a strong desire in the community to further evolve the v2 spec.

src/zarr/core/dtype/core.py
normanrz: Assigning names for extensions, such as dtypes, is something that the Zarr core spec should define to coordinate between the different Zarr implementations. However, there are some gaps in the spec that provide clear guidance for doing that. In the Zarr steering council, we are currently evaluating different options that we will propose to the community, shortly. Our goal is to achieve a naming mechanism that avoids naming conflicts. Our current favorite is to have 2 types of names:
- URI-based names, e.g. `https://nenb.github.io/bfloat16`, which can be freely used by anybody who reasonably controls the URI. The URI doesn't need to resolve to anything; it is just a name. However, it makes sense to have some useful information under the URI, e.g. a spec document.
- Raw names, e.g. `bfloat16`, which would be assigned through a centralized registry (e.g. a git repo) through a to-be-defined process. This will entail a bit more process than the URI-based names and will come with some expectations w.r.t. backwards compatibility.

> Is the configuration key an appropriate location for such metadata?

Not necessarily. I think this information would be better placed in specification documents of the dtypes.

src/zarr/core/dtype/core.py
normanrz: As I wrote in my other comment, we generally want to follow some guidelines around how and what metadata goes into the `zarr.json` files to achieve interoperability with other libraries/languages that implement Zarr. Unfortunately, some of these guidelines still need to be specced out.

In any case, maybe I missed it, why do we need to persist the endianness as part of the dtype? Shouldn't that be handled by the appropriate array-to-bytes codec (e.g. `bytes`)? The runtime endianness might need to be handled through runtime configuration, see the `ArrayConfig` class.

nenb:
@d-v-b @normanrz Thanks so much for the support during today's meeting, I'm very excited about the potential for `zarr` to help accelerate the AI field!

My current thinking is to close this PR (as it was just intended as an illustration of what I was hoping to gain support for), and then sketch out how to proceed elsewhere. Would you agree?

If so, could you give me a few pointers about how best to support coordination of this work please e.g. where to document a proposed list of tasks, who needs to be tagged etc.? 

As mentioned, I'm very flexible as to how the work proceeds! My main interests are i) allowing users to be able to register their own dtypes, ii) decoupling the `zarr-python` internal dtype API from the `numpy` dtype API (because I think `numpy` dtype API is not appropriate for AI work at the moment) and iii) trying to achieve these things in the next couple of weeks.

I know this last point is a little artificial, but as mentioned during the meeting earlier, a lot of this field moves fast at the moment!

joshmoore:
A heads up of a discussion around [ZEP9 & extension naming](https://github.com/zarr-developers/zeps/pull/65):

https://ossci.zulipchat.com/#narrow/channel/423692-Zarr/topic/ZEP9.3A.20extension.20points

*/me reminds himself yet again to look for a bot to do these cross-linkings...*

d-v-b:
@nenb could you explain the motivation for centering the design around classes that have no instance attributes? E.g., you have
```
class Float16(ZarrDType):
    zarr_spec_format = "3"
    experimental = False
    endianness = "little"
    byte_count = 2
    to_numpy = np.dtype('float16')
```
With this design, `__init__` is a no-op, so from what I can tell there's not really a difference between `Float16` and `Float16()`

An alternative implementation would be: 
```
float16 = ZarrDType(zarr_spec_format=3, experimental=False, ...)
```

I'm guessing there was a reason why you didn't opt for the "specify the dtype in `__init__`" approach?

nenb:
@d-v-b 
**TL;DR**
I don't remember there being anything important behind this. From memory, I was just trying to hack something together quickly, and I think this is what I ended up with. I had a couple of 'requirements' in mind, which are described in details below, and satisfying these were more important to me than any of the (throwaway) code that I wrote here. 

**Details**
Requirements that I had:
- once a dtype has been created, don't allow patching the attributes later
- don't allow creation of a dtype instance with invalid attributes or with missing required attributes
- make sure that the dtype can be discovered via the current entrypoint mechanism

For the first point, I thought it would lead to weird and confusing situations if users were able to (for example) change the `byte_count` to a different value after creation. I thought it best to make it as difficult as possible to modify the dtypes after creation, because they are fundamental to a lot of behaviour. (Personal choice)

For the second point, this was just my attempt at re-attempting Pydantic with metaclasses and convoluted code. 🤡 I think it's useful though to prevent creation/importing bad dtype classes, for the same reason I mentioned in the previous paragraph.

I didn't want to make many changes to the entrypoint code, as I felt it would distract from what I was trying to show here about the dtypes. This is probably why I ended up with the weird design for classes that you mentioned i.e. I just kept hacking until I got everything to work with the current entrypoint mechanism. Obviously, this is not a requirement when doing this work properly.

I would personally be happy with any code that can satisfy the above three requirements. And I'm happy to discuss any of these requirements if they considered unhelpful for whatever reason!

d-v-b:
@nenb what are you thoughts on the following things:
-  handling parametric dtypes, like [`datetime64[ms]`, `datetime64[s]`, ...], [`S1`, `S2`, ...]? For the datettimes the set of units is small enough that these can be created in advance, but the fixed length types like `<S | V><length>` are harder to pre-generate. If a user shows up with a concrete numpy dtype like `S2`, it's possible to create the corresponding class at runtime with something like `type(f"FixedLengthString{length}", (ZarrDType,), {...kwargs})`, but this feels a bit clunky compared to creating an instance of a variable-length string class, e.g. `VarLenString(length=2)`. 
- I'm not sure we should flag experimental vs non experimental dtypes. Any experimentalness should be represented by the context in which the dtype is used, not the dtype itself, because e.g. extension dtypes will mature, and then cease to be experimental, and that should not require changing the API of the dtype object itself.
- I'm not sure the zarr format should be part of the dtype metadata. We should instead have a zarr-format-independent representation of dtypes, and treat the zarr format as a serialization issue.
- Endianness is tricky because it isn't part of the dtype model of zarr v3 -- in v3, [endianness is specified by a codec](https://github.com/zarr-developers/zarr-specs/blob/main/docs/v3/codecs/bytes/v1.0.rst), not the dtype itself. So from a v3 POV, little-endian uint16 and big-endian uint16 specify two arrays with the same dtype but different codecs. This might mean that any dtype object has to take `endianness` as a parameter when serializing or generating numpy dtypes.

nenb:
@d-v-b (I'll answer the rest of this message tomorrow)

> handling parametric dtypes ...

Looking at the Arrow [data types](https://arrow.apache.org/docs/format/Columnar.html#data-types) (Arrow is reasonably mature at this point, I think that it's probably reasonable to adopt their design decisions, unless there is a clear reason not to?), it looks like having a type parameter is sensible. I don't think I gave this enough thought previously. I think an explicit reference to type parameters in the Zarr v3 spec is also lacking. What are your thoughts about this type parameters?

A potential challenge that I see would be in naming these parametric dtypes. Do you see issues with assigning the parameters in the `configuration` field e.g.

```
{
  "data_type": {
    "name": "DateTime",
    "configuration: {
        "unit": "ms",
        "timezone": "UTC"    
    }
  }
}
```

Something similar could happen for fixed-width strings. Is it critical that a dtype be identified by its name, or is it allowed to stuff identifying information in `configuration` like this?

The alternative seems a rather crude naming scheme where we attach the parameters to the type name.

> it's possible to create the corresponding class at runtime ...

For (`numpy`) data types such as fixed-width strings and bytestrings (codes `U` and `S` respectively), I was leaning towards creating them at runtime. I hadn't considered how to handle more parametric dtypes more generally (see previous paragraph), which was an oversight on my part.

I personally don't see an issue with creating parametric dtypes at runtime, but open to change my mind on this.

nenb:
> I'm leaning toward the second formulation, but I'd be open to feedback for what the first is better / preferred

I vote for this as well. If there is more than one type parameter, the first formulation could be confusing.

> I'm not sure we should flag experimental vs non experimental dtypes. Any experimentalness should be represented by the context in which the dtype is used, not the dtype itself, because e.g. extension dtypes will mature, and then cease to be experimental, and that should not require changing the API of the dtype object itself.

I had in mind that this would be a private part of the dtype API, so that someone could inspect a dtype at runtime and know whether or not it was part of the spec. Let's say I wanted to write code that only applied to experimental dtypes (e.g. emit warnings). How would code be able to infer if a dtype was experimental or not at runtime? Remember that the definition of what is a core type can change from one day to the next, but the (`zarr-python`) code itself will still be the same. The only way I can solve this is for the creator to signal the intent via an attribute like `experimental`. Perhaps what I'm saying is that experimentalness is a function of how the dtype is _created_ rather than how it is _used_? This might need more discussion though, I'm not terribly confident in what I've written.

> I'm not sure the zarr format should be part of the dtype metadata. We should instead have a zarr-format-independent representation of dtypes, and treat the zarr format as a serialization issue.

I think this is another variation of the last paragraph. How do I know what format a dtype belongs to at runtime? I could imagine this being useful at some point in the future (but maybe in the very distant future 😆). I don't have strong opinions on this though, and if we were able to infer the format of a dtype at runtime another way, I would also be happy with that.

> Endianness is tricky because it isn't part of the dtype model of zarr v3 -- in v3, [endianness is specified by a codec](https://github.com/zarr-developers/zarr-specs/blob/main/docs/v3/codecs/bytes/v1.0.rst), not the dtype itself. So from a v3 POV, little-endian uint16 and big-endian uint16 specify two arrays with the same dtype but different codecs. This might mean that any dtype object has to take endianness as a parameter when serializing or generating numpy dtypes.

Yeah, I remember this causing me quite a few headaches. I think given all the `zarr-python` code currently out there, folks are going to be trying to do things like `<i2` for quite a while into the future. I added the `endianess` attribute as a (private) part of the API, to try and help `zarr-python` developers support this use-case. Basically, `zarr-python` could use this attribute to track the endianess of the dtype _in memory_, but it would never be used when serialising to disk. So, when I create my array in `zarr-python` using `>i2`, `zarr-python` is still able to support this use-case by creating a big-endian dtype _in memory_. But, when serialising to disk, it would follow the spec and instead store this information in the codec. 

I think I remember you pointing out in earlier issues, that using codecs in this way diverges from the `numpy` dtype model, and this makes it difficult for `zarr-python`, give its close historical relationship with `numpy`. Adding the attribute on the dtype was my way of trying to support this. Again, it would ultiamtely be serialised to disk as part of the codec, but it would be part of the (Python) dtype private API as well, to continue support use cases like creating arrays with `>i2` and`<i2`.

Let me know if this isn't clear. Given the amount of issues related to endianness for v3 in `zarr-python`, I would love to try and solve as many of these as possible with a new dtype API.

d-v-b:
with the [spiritual successor](https://github.com/zarr-developers/zarr-python/pull/2874 ) to this pr merged, can we close this?

nenb:
Closing after merge of https://github.com/zarr-developers/zarr-python/pull/2874

Issue #2574, opened 2024-12-18
DOC: Using Zarr with GPU arrays


jhamman:
### Describe the issue linked to the documentation

Zarr Python 3 has new support for working with CuPy arrays. The only problem is no one knows how to use it!

@akshaysubr / @madsbk / @jakirkham - would you be up for adding a section to the docs? See #2568 for a in-progress docs refactor that should make it easy to add a page on this.

jhamman:
Sounds great @akshaysubr! The docs refactor landed this week so you could pick up and add something to [the advanced topics user-guide section](https://zarr.readthedocs.io/en/latest/user-guide/index.html#advanced-topics).

dcherian:
So does the current implementation only include `GpuMemoryStore`? For some reason, I thought this was superceding Kvikio.

TomAugspurger:
Here's what currently works:

```python
import cupy as cp

import zarr.api.asynchronous
import zarr.core.buffer.cpu
import zarr.core.buffer.gpu
import zarr.storage


async def main() -> None:
    src = cp.random.uniform(size=(100, 100))  # allocate on the device
    store = zarr.storage.LocalStore("data.zarr")
    z = await zarr.api.asynchronous.create_array(
        store=store, name="a", shape=src.shape, chunks=(10, 10), dtype=src.dtype, overwrite=True
    )

    await z.setitem(
        (slice(10), slice(10)), src[:10, :10], prototype=zarr.core.buffer.gpu.buffer_prototype
    )

    result = await z.getitem(
        (slice(10), slice(10)), prototype=zarr.core.buffer.gpu.buffer_prototype
    )
    print(type(result))  # <class 'cupy.ndarray'>

if __name__ == "__main__":
    import asyncio

    asyncio.run(main())
```

There's also `GpuMemoryStore`, but right now I *think* that's mostly just a convenience that automatically moves incoming data to on-device memory.

Notably, this uses the `prototype` argument to getitem and setitem, and so is only available through the async API. I think we'll want a single configuration option to control this, so that it's usable through the rest of Zarr's API (I'll write up a separate issue for designing this).

> For some reason, I thought this was superceding Kvikio.

I think kvikio will still be useful for users who want to want to read data directly into device memory (avoiding any intermediate allocations on the host before copying to the device). kvikio isn't currently compatible with zarr v3, but I'm hoping to look into that sometime.

akshaysubr:
[...]

PR #2751, opened 2025-01-22
Update and document GPU buffer handling

TomAugspurger:
This updates how we handle GPU buffers. See the new docs page for a simple example.

The basic idea, as discussed in https://github.com/zarr-developers/zarr-python/issues/2574 is to use host buffers for all metadata objects and device buffers for data.

Zarr has two types of buffers: plain buffers (used for a stream of bytes) and ndbuffers (used for bytes that represent ndarrays). To make it easier for users, I've added a new config option `zarr.config.enable_gpu()` that can be used to update those both. If we need additional customizations in the future (like gpu-accelerated codecs), we can add them here.

I've opened this as a draft for now. I want to look a bit more at exactly when data is copied between the host and device. Right now, it looks like the default Zarr v3 codec pipeline will automatically transfer bytes to the host in `ZstdCodec._encode_single`: https://github.com/zarr-developers/zarr-python/blob/0c154c327e92a40648b3dd95e9e9d9c71d3d53a3/src/zarr/codecs/zstd.py#L88

This PR ensures that you end up with GPU bytes when reading data, but all data I/O and encoding / decoding still happens on the CPU. I think in the future we'll be able to do more work on the GPU while avoiding copies back to the host.

TODO:

Clsoes #2574


TomAugspurger:
> I've opened this as a draft for now. I want to look a bit more at exactly when data is copied between the host and device. Right now, it looks like the default Zarr v3 codec pipeline will automatically transfer bytes to the host in ZstdCodec._encode_single:

Still looking at this, but there is currently a copy to the host for running Zstd (the `zarr.codecs.zstd.ZstdCodec` codec explicitly asks for a NumPy array, which (silently) copies from device to host if necessary). I'll probably leave codecs for a separate issue / PR, but I want to at least get an example working where everything stays on the GPU before moving this out of draft.

cc @madsbk.

TomAugspurger:
This should be ready for review now.

This fixes the issue with storing metadata in host memory, rather than device memory.

It adds a convenience function for setting up everything to use GPUs. Currently, that just sets the buffer prototype. In the future, I think it could also update the default codecs to use GPU codecs. https://gist.github.com/TomAugspurger/ba13bc29b27f587ae4709ac7f30d89c8 has a snippet hacking together an example of what that future might look like, but IIUC @akshaysubr is working on new APIs for compression / codecs, so IMO it makes sense to address that separately. For now, I've left a note that codecs will run on the CPU by default.

tests/test_api.py
```diff
@@ -1121,3 +1122,38 @@ def test_open_array_with_mode_r_plus(store: Store) -> None:
...
+            chunks=(10, 10),
+            dtype=src.dtype,
+            overwrite=True,
+            zarr_format=zarr_format,
+            compressors=compressors,
+        )
+        z[:10, :10] = src[:10, :10]
+
+        result = z[:10, :10]
+        cp.testing.assert_array_equal(result, src[:10, :10])
```
madsbk, line 1159:
```suggestion
        assert isinstance(got, result)
        cp.testing.assert_array_equal(result, src[:10, :10])
```

madsbk: I don't think `assert_array_equal` tests object type.

madsbk (approved):
Looks good, I only have a minor suggestion

akshaysubr (approved):
This looks good to me, approving!

TomAugspurger:
@dstansby would you be able to confirm that I did the changelog correctly? Happy to add a note to the contributing guide like

> Pull requests should usually include a note for the changelog as a [towncrier](https://towncrier.readthedocs.io/en/stable/tutorial.html) news fragment. Use the GitHub issue number along with the type of your change (`feature`, `bugfix`, `doc`, `removal`, `misc`) 
>
> ```
> towncrier create
> ```

> You'll be prompted to input the issue number, fix type, and text of the changelog. See the towncrier docs for more.

dstansby:
Release notes look good to me 👍 Adding that text in our contributor guide would be awesome (and TIL that you can use `towncrier create`, I've been doing them by hand until now!)

TomAugspurger:
Should be good to go.

src/zarr/core/config.py
```diff
@@ -56,6 +59,14 @@ def reset(self) -> None:
         self.clear()
         self.refresh()
 
+    def enable_gpu(self) -> ConfigSet:
+        """
+        Configure Zarr to use GPUs where possible.
+        """
+        return self.set(
```
TomAugspurger, line 66: I believe this shows it's covered: https://app.codecov.io/github/zarr-developers/zarr-python/blob/TomAugspurger%2Fzarr-python%3Atom%2Ffix%2Fgpu/src%2Fzarr%2Fcore%2Fconfig.py#L66

Maybe the warning comes from the CPU-only coverage run?

TomAugspurger (commented):
@jhamman or would @d-v-b would you be able to take a look at this sometime? Thanks!

dcherian (approved):

PR #2755, opened 2025-01-23
Deterministic chunk padding

### LINKED ISSUE #2696: Chunk padding is non-deterministic with `zarr_format=2`
### Zarr version

3.0.0

### Numcodecs version

0.14.0

### Python Version

3.11

### Operating System

Mac

### Installation

pip

### Description

When using `zarr_format=2` with Zarr Python 3.0.0 it's possible to write chunks that have a different size each time they are written (even with the same data). This doesn't occur if `zarr_format=3`, or with Zarr Python 2.

The problem seems to be that the chunk is initialized with no fill value (using `np.empty`).

This was found in https://github.com/sgkit-dev/bio2zarr/pull/288#issuecomment-2575426493.

### Steps to reproduce

```python
import zarr

def run():
    a = zarr.empty(
        store=store,
        shape=(3,),
        chunks=(1000,),
        dtype=int,
        zarr_format=2,
    )
    a[:] = 1
    print(len(store._store_dict["0"]))

run()
run()
```

Example output (changes from run to run):
```
1890
2954
```

Changing to `zarr_format=3`, and the key to `"c/0"` gives the following consistently:

```
27
27
```

### Additional output

_No response_

brokkoli71:
- fixes #2696

TODO:


d-v-b (approved):

src/zarr/core/buffer/cpu.py
```diff
@@ -154,7 +154,7 @@ def create(
         order: Literal["C", "F"] = "C",
         fill_value: Any | None = None,
     ) -> Self:
-        ret = cls(np.empty(shape=tuple(shape), dtype=dtype, order=order))
+        ret = cls(np.zeros(shape=tuple(shape), dtype=dtype, order=order))
         if fill_value is not None:
             ret.fill(fill_value)
```
jhamman, line 159:
```suggestion
        ret = cls(np.full(shape=tuple(shape), fill_value=fill_value or 0, dtype=dtype, order=order))
```

Untested but I'm fairly sure this pans out to be the same thing and is probably a bit faster.

changes/2755.bugfix.rst
```diff
@@ -0,0 +1 @@
+Upon creation, an empty ``zarr.core.buffer.cpu.NDBuffer`` is filled with zeros to ensure deterministic behavior.
```
dstansby, line 1:
```suggestion
Upon creation, an empty ``zarr.core.buffer.cpu.NDBuffer`` is filled with zeros to ensure deterministic behavior.
This fixes a bug where Zarr format 2 data with no fill value was written with un-predictable chunk sizes.
```

tests/test_store/test_memory.py
```diff
@@ -46,6 +52,28 @@ def test_store_supports_partial_writes(self, store: MemoryStore) -> None:
...
+                zarr_format=zarr_format,
+                overwrite=True,
+            )
+            a[...] = 1
+            key = "0" if zarr_format == 2 else "c/0"
+            return len(store._store_dict[key])
+
+        l1 = padding_size()
+        l2 = padding_size()
+        assert l1 == l2
```
dstansby, line 75: It would be good to add another check here to compare l1 and l2 to the exact value expected (1000?)

dstansby (commented):
A couple of suggestions (in particular, explaining in a bit more detail the user facing bug this fixes), otherwise looks good!

dstansby (changes requested):
Marking this as request changes, since it looks like it broke some tests - I had a look at the tests, and failure seems real (requesting a fill value of `""`, and getting `0` instead)

normanrz:
> Marking this as request changes, since it looks like it broke some tests - I had a look at the tests, and failure seems real (requesting a fill value of `""`, and getting `0` instead)

Fixed it.

dstansby (changes requested):
If this now means `zarr.emtpy` deafults to filling data with zeros, instead of undefined data, this is a behaviour change from zarr-python 2: https://zarr.readthedocs.io/en/v2.18.4/api/creation.html#zarr.creation.empty - I think this is fine, but we should document that change in the changelog.

The [docstring of zarr.creation.empty](https://zarr.readthedocs.io/en/v3.0.1/api/zarr/creation/index.html#zarr.creation.empty) is now also incorrect, and needs updating:

> The contents of an empty Zarr array are not defined. On attempting to retrieve data from an empty Zarr array, any values may be returned, and these are not guaranteed to be stable from one access to the next.

I also had a suggested improvement to the changelog above.

dstansby (approved):
Thanks for updating the docstrings!

I had another read and think about this, and this is quite a major change happening - the data being returned/created by a function is changing. I *think* in this case it's okay, because it's only filled data that the user hasn't specified, and it fixes a bad bug.

It also seems like this change makes `empty()` do the same thing as `zeros()`? (create an array filled with zeros) Is that correct, and do these two functions do the same thing now? If that's the case, should we deprecate and eventually remove `zarr.empty()`? That can happen in a follow up PR to keep this tightly scoped to the bug fix, but it's a consquence of this PR we should make sure we're happy with.

Because of the above issues I'm not 100% on just me reviewing this - I'd like someone else in @zarr-developers/python-core-devs to review and approve this (and explicitly say that they're okay with the data changing/potentially making `emtpy()` redundant).

src/zarr/api/asynchronous.py
```diff
@@ -1065,20 +1065,15 @@ async def create(
...
     Parameters
     ----------
     shape : int or tuple of int
         Shape of the empty array.
     **kwargs
         Keyword arguments passed to :func:`zarr.api.asynchronous.create`.
-
-    Notes
-    -----
-    The contents of an empty Zarr array are not defined. On attempting to
```
joshmoore, line 1079: A bit of a random thought, but this sentence could likely stay in the sense that not defined can also mean "with `fill_value OR 0`" if we are concerned that this isn't a constraint we want to impose in the future.

normanrz:
Both `zarr.empty` and `zarr.zeros` should be deprecated in favor of `zarr.create_array`.

The old docs might have said that `zarr.empty` returns unitialized data, but I don't think that was the case in the code. Here, entire are filled with `fill_value` even if outside the to-be-set selection: https://github.com/zarr-developers/zarr-python/pull/2755/files#diff-44efa3ae220ba9737fa8c5443b3c3af09a7c0f9549d5c267404f7fa3ce467318

d-v-b:
i had a quick look through the v2 code and it looks like v2 was padding with 0s to ensure consistency and compressibility: https://github.com/zarr-developers/zarr-python/blob/66e2982d4d16ace4e32bf0deba627357c8982844/zarr/core.py#L2309-L2313

d-v-b (approved):

d-v-b:
@brokkoli71 is this ready to go in?

brokkoli71:
> @brokkoli71 is this ready to go in?

yes 👍🏼 @d-v-b

PR #2763, opened 2025-01-24
Use ChunkKeyEncodingLike type alias

d-v-b:
previously, `ChunkKeyEncodingLike` was the name of a `TypedDict` instance that models the JSON form of the `chunk_key_encoding` attribute in zarr v3 array metadata. But `ChunkKeyEncodingLike` should be the name of a type alias that denotes the union of types that can be converted to an instance of `ChunkKeyEncoding`. So I renamed the typeddict to `ChunkKeyEncodingParams` and added a type alias (it is `ChunkKeyEncoding | ChunkKeyEncodingParams`)


d-v-b:
@dstansby can you check that my release notes are correct? and is it possible for the gha that adds the `needs release notes` label also _remove_ that label when release notes are added?

dstansby:
Looks good to me! Perhaps the note could say how the change impacts users? I'm guessing here the typing is just being made more permissive?

> and is it possible for the gha that adds the needs release notes label also remove that label when release notes are added?

Yes, it looks like there's a `sync-labels` option for the labeler action: https://github.com/actions/labeler/tree/main/?tab=readme-ov-file#inputs

d-v-b:
It should have no effect on users.

dstansby:
In that case I'd probably skip adding a release note? At least in my mind, release notes are targeted other folks using the package, not internally for developers.

d-v-b:
developers are also users :)

d-v-b:
in fact, I think developers will be more interested in reading the release notes than ordinary users. Thus I don't see a particular reason to omit infrastructure changes from the release notes.

dstansby:
I guess so - in that case LGTM, and perhaps the 'chore' release notes category is a good place to put developer facing notes 👍

dcherian (approved):

Issue #2731, opened 2025-01-19
botocore should be listed in test requirements


QuLogic:
### Zarr version

3.0.1

### Numcodecs version

0.14.0

### Python Version

3.13.1

### Operating System

Fedora rawhide

### Installation

source

### Description

It is always imported in tests, so should be part of the [`test` extra in `pyproject.toml`](https://github.com/zarr-developers/zarr-python/blob/main/pyproject.toml#L72-L85).
https://github.com/zarr-developers/zarr-python/blob/31d377b954eb639e3f4d51b02945fdd111f1b52e/tests/test_store/test_fsspec.py#L8

If instead you expect it to be required transitively through some other dependency, then it should be imported below the `importorskip` lines.
https://github.com/zarr-developers/zarr-python/blob/31d377b954eb639e3f4d51b02945fdd111f1b52e/tests/test_store/test_fsspec.py#L23-L27

### Steps to reproduce

pip install zarr[test]
pytest

QuLogic:
Perhaps also `aiohttp` or `fsspec[http]` for `test_make_store_path_fsspec` .

d-v-b:
thanks for the report. I think we need to refactor the how we test the fsspec-dependent stuff entirely, and that effort should include the changes you propose here.

PR #2768, opened 2025-01-26
Make botocore an optional test dependency

dstansby:
Fixes https://github.com/zarr-developers/zarr-python/issues/2731


pyproject.toml
```diff
@@ -73,10 +73,12 @@ test = [
     "coverage",
     "pytest",
     "pytest-cov",
+    'zarr[remote]',
```
d-v-b, line 76: Why do we want zarr itself listed as a dependency here? I feel like we never want to run our test suite against a version of zarr from pypi, instead tests should simply fail if zarr cannot be resolved as a local import.

dstansby: This is to pick up the `fsspec` dependency defined in the `remote` extra, to avoid specifying the `fsspec` pin in multiple locations in this file. `pip` is clever and recognises that `zarr` here means the 'local' installation, and doesn't try and get it from PyPI.

d-v-b: btw an alternative approach would be to define a `test-remote` extra, that includes botocore etc, and then our test environment would inherit from `remote`, `test`, and `test-remote`. Even if we drop fsspec, we would still need moto for testing some other s3-based store, so it does make sense to separate these things.

d-v-b: (maybe `test-s3` is actually a better name for the test dependencies that only exist to test s3)

dstansby: https://github.com/pypa/pip/issues/11296 is the best documentation I could find. FWIW we've been using something like this in the mindeps test environment for a while and it doesn't seem to have any problems:

https://github.com/zarr-developers/zarr-python/blob/fb37ff0942c7417a7767cc4d537de7173a9b3329/pyproject.toml#L237

d-v-b: cool, that took me to this discussion: https://discuss.python.org/t/where-is-nested-recursive-optional-dependencies-documented/35648, where the responses seemed to be "this behavior is completely obvious, why would you need it documented". Never change, python packaging!

d-v-b: if it's already working then lets keep doing it :)

d-v-b (approved):

PR #2778, opened 2025-01-30
Use removeprefix rather than replace to avoid separator deletion

### LINKED ISSUE #2777: Potentially unexpected behavior on fsspecstore list method
### Zarr version

3.0.1

### Numcodecs version

0.15.0

### Python Version

3.12

### Operating System

Mac

### Installation

pip

### Description

Looks as though this line will have some unexpected results (wiping out all separators) if the empty string is passed as the path during store initialization. The empty string will otherwise work as a way of specifying the current working directory, so this may be undesirable. A couple of resolutions make sense here: 
1. guard against empty string path arguments in __init__
2. modify this line to avoid wholesale replace of `self.path + "/"` with `""`: https://github.com/zarr-developers/zarr-python/blob/e602aa1d19f26bb06669994231e524c55bcecbeb/src/zarr/storage/_fsspec.py#L344

### Steps to reproduce

Simplest path to reproduction is just to simulate values. These come from a grib in the kerchunk tests:

```python
allfiles = ['.zattrs', '.zgroup', 'refc/.zattrs', 'refc/.zgroup', 'refc/instant.zattrs', 'refc/instant.zgroup', 'refc/instant/atmosphere/.zattrs', 'refc/instant/atmosphere/.zgroup', 'refc/instant/atmosphere/atmosphere/.zarray']
path = ""

# following the logic here https://github.com/zarr-developers/zarr-python/blob/e602aa1d19f26bb06669994231e524c55bcecbeb/src/zarr/storage/_fsspec.py#L344:
for onefile in (a.replace(path + "/", "") for a in allfiles):
    print(onefile)
```

### Additional output

_No response_

moradology:
The point of replace here is just to remove the prefix instance but it has the unintended consequence of deleting separators when path is the empty string. This uses `removeprefix` explicitly and avoids that potential pitfall

Fixes #2777 

TODO:

### DIFF
--- a/changes/2778.bugfix.rst
+++ b/changes/2778.bugfix.rst
@@ -1,0 +1,1 @@
+Use removeprefix rather than replace when removing filename prefixes in `FsspecStore.list`
--- a/src/zarr/storage/_fsspec.py
+++ b/src/zarr/storage/_fsspec.py
@@ -341,7 +341,7 @@ class FsspecStore(Store):
     async def list(self) -> AsyncIterator[str]:
         # docstring inherited
         allfiles = await self.fs._find(self.path, detail=False, withdirs=False)
-        for onefile in (a.replace(self.path + "/", "") for a in allfiles):
+        for onefile in (a.removeprefix(self.path + "/") for a in allfiles):
             yield onefile
 
     async def list_dir(self, prefix: str) -> AsyncIterator[str]:
--- a/src/zarr/testing/store.py
+++ b/src/zarr/testing/store.py
@@ -400,6 +400,37 @@ class StoreTests(Generic[S, B]):
             expected = tuple(sorted(expected))
             assert observed == expected
 
+    async def test_list_empty_path(self, store: S) -> None:
+        """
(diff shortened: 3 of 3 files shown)


dstansby (changes requested):
Thansk a lot, code change looks 👍 

Could you add a test using an example that would have failed before this change to make sure this doesn't fail in the future?

d-v-b:
> Thansk a lot, code change looks 👍
> 
> Could you add a test using an example that would have failed before this change to make sure this doesn't fail in the future?

seconding this request -- the real bug is the hole in our test suite

moradology:
Test added. Seems like a fairly general behavior but it wasn't obvious that there was a more abstract way to add it, so I opted for covering just the `fsspec` store path

d-v-b:
those tests look good. if it's not too much work, you could consider adding them to the store test base class -- maybe we catch some more bugs this way. But that's extra credit. If you don't change the base class, I can just open an issue indicating that we need to include these fsspec tests in the base class.

d-v-b (approved):

PR #2784, opened 2025-01-31
Always skip reads when completely overwriting chunks

### LINKED ISSUE #757: behavior of `is_total_slice` for boundary chunks

dcherian:
Teach the *Indexer objects to detect when a selection corresponds to a whole chunk. Propagate that information down to the codec pipeline where reads can be skipped if it makes sense locally.

Closes #757

~I'm opening this for comments about the approach (@d-v-b , @normanrz ) it could definitely use more tests (edit: and some fixes)~

TODO:

### DIFF
--- a/src/zarr/abc/codec.py
+++ b/src/zarr/abc/codec.py
@@ -357,7 +357,7 @@ class CodecPipeline:
     @abstractmethod
     async def read(
         self,
-        batch_info: Iterable[tuple[ByteGetter, ArraySpec, SelectorTuple, SelectorTuple]],
+        batch_info: Iterable[tuple[ByteGetter, ArraySpec, SelectorTuple, SelectorTuple, bool]],
         out: NDBuffer,
         drop_axes: tuple[int, ...] = (),
     ) -> None:
@@ -379,7 +379,7 @@ class CodecPipeline:
     @abstractmethod
     async def write(
         self,
-        batch_info: Iterable[tuple[ByteSetter, ArraySpec, SelectorTuple, SelectorTuple]],
+        batch_info: Iterable[tuple[ByteSetter, ArraySpec, SelectorTuple, SelectorTuple, bool]],
         value: NDBuffer,
         drop_axes: tuple[int, ...] = (),
     ) -> None:
--- a/src/zarr/core/indexing.py
+++ b/src/zarr/core/indexing.py
@@ -403,24 +403,16 @@ class SliceDimIndexer:
                 dim_chunk_sel_start = self.start - dim_offset
                 dim_out_offset = 0
 
-            # Use None to indicate this selection ends at the chunk edge
-            # This is useful for is_total_slice at boundary chunks,
             if self.stop > dim_limit:
                 # selection ends after current chunk
-                dim_chunk_sel_stop = None  # dim_chunk_len
+                dim_chunk_sel_stop = dim_chunk_len
 
             else:
                 # selection ends within current chunk
-                if dim_chunk_ix == (self.nchunks - 1):
-                    # all of the last chunk is included
-                    dim_chunk_sel_stop = None
-                else:
-                    dim_chunk_sel_stop = self.stop - dim_offset
+                dim_chunk_sel_stop = self.stop - dim_offset
 
             dim_chunk_sel = slice(dim_chunk_sel_start, dim_chunk_sel_stop, self.step)
-            dim_chunk_nitems = ceildiv(
-                ((dim_chunk_sel_stop or dim_chunk_len) - dim_chunk_sel_start), self.step
-            )
+            dim_chunk_nitems = ceildiv((dim_chunk_sel_stop - dim_chunk_sel_start), self.step)
 
             # If there are no elements on the selection within this chunk, then skip
             if dim_chunk_nitems == 0:
@@ -1386,17 +1378,7 @@ def is_total_slice(item: Selection, shape: ChunkCoords) -> bool:
                 isinstance(dim_sel, slice)
                 and (
                     (dim_sel == slice(None))
-                    or (
-                        dim_sel.stop is not None
-                        and (dim_sel.stop - dim_sel.start == dim_len)
-                        and (dim_sel.step in [1, None])
-                    )
-                    # starts exactly at a chunk
-                    or (
-                        (dim_sel.start % dim_len == 0)
-                        and dim_sel.stop is None
-                        and (dim_sel.step in [1, None])
-                    )
+                    or ((dim_sel.stop - dim_sel.start == dim_len) and (dim_sel.step in [1, None]))
                 )
             )
             for dim_sel, dim_len in zip(item, shape, strict=False)
--- a/src/zarr/core/indexing.py
+++ b/src/zarr/core/indexing.py
@@ -421,7 +421,9 @@ class SliceDimIndexer:
 
             dim_out_sel = slice(dim_out_offset, dim_out_offset + dim_chunk_nitems)
 
-            is_complete_chunk = dim_chunk_sel_start == 0 and (self.stop >= dim_limit)
+            is_complete_chunk = (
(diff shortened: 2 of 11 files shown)


src/zarr/core/indexing.py
```diff
@@ -403,16 +403,24 @@ def __iter__(self) -> Iterator[ChunkDimProjection]:
...
+                if dim_chunk_ix == (self.nchunks - 1):
+                    # all of the last chunk is included
+                    dim_chunk_sel_stop = None
+                else:
+                    dim_chunk_sel_stop = self.stop - dim_offset
 
             dim_chunk_sel = slice(dim_chunk_sel_start, dim_chunk_sel_stop, self.step)
-            dim_chunk_nitems = ceildiv((dim_chunk_sel_stop - dim_chunk_sel_start), self.step)
+            dim_chunk_nitems = ceildiv(
+                ((dim_chunk_sel_stop or dim_chunk_len) - dim_chunk_sel_start), self.step
```
d-v-b, line 422: This is checking if `dim_chunk_sel_stop` is None, but can either `dim_chunk_sel_stop` or `dim_chunk_len` be 0? Because that would mess up this conditional.

src/zarr/storage/_logging.py
```diff
@@ -88,7 +88,7 @@ def log(self, hint: Any = "") -> Generator[None, None, None]:
         op = f"{type(self._store).__name__}.{method}"
         if hint:
             op = f"{op}({hint})"
-        self.logger.info("Calling %s", op)
+        self.logger.info(" Calling %s", op)
```
d-v-b, line 91: is this intended?

dcherian: yes it makes them line up!

```
2025-01-31 08:52:11,360 - LoggingStore(memory://4701418816) - INFO -  Calling MemoryStore.set(foo/c/0/0/1/0)
2025-01-31 08:52:11,360 - LoggingStore(memory://4701418816) - INFO - Finished MemoryStore.set(foo/c/0/0/1/0) [0.00 s]
2025-01-31 08:52:11,361 - LoggingStore(memory://4701418816) - INFO -  Calling MemoryStore.set(foo/c/0/0/0/1)
2025-01-31 08:52:11,361 - LoggingStore(memory://4701418816) - INFO - Finished MemoryStore.set(foo/c/0/0/0/1) [0.00 s]
```

src/zarr/core/indexing.py
```diff
@@ -1378,7 +1386,17 @@ def is_total_slice(item: Selection, shape: ChunkCoords) -> bool:
                 isinstance(dim_sel, slice)
                 and (
                     (dim_sel == slice(None))
-                    or ((dim_sel.stop - dim_sel.start == dim_len) and (dim_sel.step in [1, None]))
+                    or (
+                        dim_sel.stop is not None
+                        and (dim_sel.stop - dim_sel.start == dim_len)
+                        and (dim_sel.step in [1, None])
```
d-v-b, line 1392: it's kinda of weird that the step matters at all, if we know the stop. I understand this is how slice objects work, but maybe zarr indexing should not rely on python slice object semantics.

tests/test_array.py
```diff
@@ -1335,5 +1335,14 @@ async def test_orthogonal_set_total_slice() -> None:
...
         array[0, slice(4, 10)] = np.arange(6)
+
+    array = zarr.create_array(
+        store, shape=(20, 21), chunks=(1, 2), dtype=int, fill_value=-1, overwrite=True
+    )
+    with mock.patch("zarr.storage.MemoryStore.get", side_effect=RuntimeError):
+        array[0, :] = np.arange(21)
+
+    with mock.patch("zarr.storage.MemoryStore.get", side_effect=RuntimeError):
+        array[:] = 1
```
d-v-b, line 1348: i like this test

src/zarr/core/indexing.py
dcherian: but you can do `a[slice(None, None, 2] = 1` for example...

d-v-b: `__getitem__` should totally accept slices as input, but we should _immediately_ convert slices into objects that make sense for zarr array indices (finite sets of regularly spaced non-negative integers). Then we don't have to handle `Nones` for checks like this here, and we could frame the problem in terms of sets of integers. Like, "is_total_slice" is equivalent to "is the set of indices for the chunk completely contained by this other set of indices"

d-v-b: (out of scope for this PR of course)

dcherian: > "is_total_slice" is equivalent to "is the set of indices for the chunk completely contained by this other set of indices"

right but then handling the potentially-smaller-last-chunk is painful. Today's logic is basically that, except it can't handle a smaller last chunk because it has no context.

src/zarr/abc/codec.py
```diff
@@ -379,7 +379,7 @@ async def read(
     @abstractmethod
     async def write(
         self,
-        batch_info: Iterable[tuple[ByteSetter, ArraySpec, SelectorTuple, SelectorTuple]],
+        batch_info: Iterable[tuple[ByteSetter, ArraySpec, SelectorTuple, SelectorTuple, bool]],
```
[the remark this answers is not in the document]
dcherian: this will cause cascading changes throughout this file, so I refrained.

IMO the real refactor is switching from NamedTuple to a dataclass and then passing that around. The code is already ugly, just got a little more uglier.

src/zarr/core/indexing.py
```diff
@@ -1198,7 +1206,8 @@ def __iter__(self) -> Iterator[ChunkProjection]:
                 for (dim_sel, dim_chunk_offset) in zip(self.selection, chunk_offsets, strict=True)
             )
 
-            yield ChunkProjection(chunk_coords, chunk_selection, out_selection)
+            is_complete_chunk = False  # TODO
```
dcherian, line 1209: I'm not sure where this is used. Is it always True?

dcherian:
https://github.com/zarr-developers/zarr-python/pull/2784 fixes the `is_total_slice` problem.

d-v-b:
I find the indexing code pretty hard to follow, but this looks good to me

d-v-b (approved):

jhamman:
@normanrz - do you have the bandwidth to give this a review?

dcherian:
@normanrz if you have time, please prioritize reviewing https://github.com/zarr-developers/numcodecs/pull/700 over this one.

dcherian:
I'll merge tomorrow if there are no more comments.

PR #2785, opened 2025-01-31
Pluralize glob and file in labeler.yml

moradology:
Well, this is embarrassing. Need to pluralize `glob` and `file` when using `all` as opposed to `any`. Makes sense. Missed it. Going to test things first by changing target to `pull_request` rather than `pull_request_target`

TODO:

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,4 +1,4 @@
 needs release notes:
   - all:
     - changed-files:
-      - all-glob-to-all-file: '!changes/*.rst'
+      - all-globs-to-all-files: '!changes/*.rst'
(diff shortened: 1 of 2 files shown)


moradology:
Naturally, perms are an issue. :facepalm:
image

moradology:
OK, the good news is that with the .rst added, this thing does, in fact, work. I will now revert to pull_request_target (which will fail for the above permissions issues). 
image

jhamman:
Thanks @moradology for the quick fix here.

moradology:
OK, the fix should be ready to go in and it should Just Work™

jhamman (approved):

PR #2795, opened 2025-02-04
create_array creates explicit groups

### LINKED ISSUE #2794: zarr v3 still generates implicit groups
### Zarr version

v3.0.2

### Numcodecs version

v0.15

### Python Version

3.12

### Operating System

Linux

### Installation

pip install zarr

### Description

Following https://zarr.readthedocs.io/en/stable/user-guide/groups.html#working-with-groups and doing
```
import zarr
root = zarr.open_group('group.zarr', mode='w')
z = root.create_array(name='foo/bar/baz', shape=(10000, 10000), chunks=(1000, 1000), dtype='int32')
```

This generates the following file hierarchy:
```
group.zarr
group.zarr/zarr.json
group.zarr/foo
group.zarr/foo/bar
group.zarr/foo/bar/baz
group.zarr/foo/bar/baz/zarr.json
```

Note the absence of a zarr.json file in the ``foo`` and ``foo/bar`` subdirectory, making them implicit groups, which have been removed in the Zarr V3 spec (https://github.com/zarr-developers/zarr-specs/pull/292)

So either zarr-python shouldn't accept a name like 'foo/bar/baz' for an array, or explicitly create groups

### Steps to reproduce

-

### Additional output

_No response_

d-v-b:
Fixes #2794. 
- `init_array` used to return a metadata object, and it would only create a single metadata document (the source of the bug reported in #2794). `init_array` is modified to create parent groups as needed, and it returns an `AsyncArray` instance, and accordingly it now takes a `config` parameter. 

A few off-target changes of note, that were necessary to make the above change smoother:
- The dict form of the `ArrayConfig` was renamed from `ArrayConfigLike` to `ArrayConfigParams`, and the name `ArrayConfigLike` is now used for the union of `ArrayConfig | ArrayConfigParams`. 
- Config parsing now happens in `AsyncArray.__init__`, instead of higher up, since we were already accepting `None` in that function this is a simplifying change.
- `create_array` tests are grouped in a test class. I combined some redundant tests.


d-v-b:
i can build the docs locally, so I'm confused why the build in ci is failing. Any ideas @dstansby ?

dstansby:
> i can build the docs locally, so I'm confused why the build in ci is failing. Any ideas @dstansby ?

Have you looked at the difference between the build here, and the last readthedocs build that passed to see if there's any obvious differences that could be causing the issue?

d-v-b:
it looks like in the last 20 hours all docs builds for PRs against `main` have failed with warnings like 
```
[AutoAPI] Reading files... [  2%] /home/docs/checkouts/readthedocs.org/user_builds/zarr/checkouts/2795/src/zarr/_version.py
WARNING: Unable to read file: /home/docs/checkouts/readthedocs.org/user_builds/zarr/checkouts/2795/src/zarr/_version.py [autoapi.not_readable]
[AutoAPI] Reading files... [  3%] /home/docs/checkouts/readthedocs.org/user_builds/zarr/checkouts/2795/src/zarr/creation.py
WARNING: Unable to read file: /home/docs/checkouts/readthedocs.org/user_builds/zarr/checkouts/2795/src/zarr/creation.py [autoapi.not_readable]
``` 

the only difference I can see between this and the successful builds is the lack of warnings, and thus the lack of failure. That's not super informative to me unfortunately.

dstansby:
Are all the dependencies being installed the same between a passing/failing build too?

d-v-b:
> Are all the dependencies being installed the same between a passing/failing build too?

as far as I can tell there are no differences in the docs build process, but it's possible that I missed something.

d-v-b:
since #2796 is failing for the same reasons, I feel confident that nothing in this PR is actually causing the docs build problems.

dcherian (approved):
I added a property test. Please take a look at the last commit.

tests/test_array.py
```diff
@@ -992,339 +932,397 @@ def test_auto_partition_auto_shards(
...
+        explicitly
+        """
+        with zarr.config.set({"array.write_empty_chunks": write_empty_chunks}):
+            arr = await create_array(store, shape=(2, 2), dtype="i4")
+            assert arr._config.write_empty_chunks == write_empty_chunks
+
+    @staticmethod
+    @pytest.mark.parametrize("zarr_format", [2, 3])
+    @pytest.mark.parametrize("path", [None, "", "/", "/foo", "foo", "foo/bar"])
+    async def test_name(store: Store, zarr_format: ZarrFormat, path: str | None) -> None:
```
dcherian, line 1304: this is the only test I reviewed. It looks good.

tests/test_properties.py
```diff
@@ -30,6 +31,24 @@ def test_roundtrip(data: st.DataObject, zarr_format: int) -> None:
...
+        if array.metadata.zarr_format == 2:
+            assert (
+                sync(array.store.get(f"{parent}/.zgroup", prototype=default_buffer_prototype()))
+                is not None
+            )
+        elif array.metadata.zarr_format == 3:
+            assert (
+                sync(array.store.get(f"{parent}/zarr.json", prototype=default_buffer_prototype()))
+                is not None
+            )
```
d-v-b, line 49: looks good, thanks for adding this. #2665 will include a function that makes retrieving an array or group from a path in storage that we can use instead of calling `store.get` here.

tests/test_array.py
```diff
@@ -992,339 +932,397 @@ def test_auto_partition_auto_shards(
...
+        """
+        Test that the value of write_empty_chunks is sensitive to the global config when not set
+        explicitly
+        """
+        with zarr.config.set({"array.write_empty_chunks": write_empty_chunks}):
+            arr = await create_array(store, shape=(2, 2), dtype="i4")
+            assert arr._config.write_empty_chunks == write_empty_chunks
+
+    @staticmethod
+    @pytest.mark.parametrize("zarr_format", [2, 3])
```
d-v-b, line 1302:
```suggestion
```
we don't actually need this parametrization here

PR #2798, opened 2025-02-04
Parse 0 fill value as "" for str dtype

### LINKED ISSUE #2792: Error in string data type handling with 0 fill value
### Zarr version

3.0.1

### Numcodecs version

-

### Python Version

3.12

### Operating System

Linux

### Installation

see reproducer

### Description

There are Zarr V2 string arrays in the wild with a `0` fill value. `zarr-python` interprets this as an empty string (e.g. when partially writing a chunk), but a completely missing chunk returns `0`'s rather than `""`s. See reproducer.

### Steps to reproduce

```python
#!/usr/bin/env -S uv run
# /// script
# requires-python = ">=3.12"
# dependencies = [
#     "zarr==3.0.1",
# ]
# ///

import zarr

array = zarr.create_array(
    dtype=str,
    shape=(5,),
    chunks=(2,),
    filters=zarr.codecs.vlen_utf8.VLenUTF8(),
    compressors=[None],
    fill_value=0,
    zarr_format=2,
    overwrite=True,
)
array[:3] = ["a", "bb", ""]
print(array.info)
# assert (array[:] == ["a", "bb", "", "", ""]).all() # EXPECTED
# array[:] is ["a", "bb", "", "", 0]
```

### Additional output

_No response_

moradology:
This PR implements `fill_value` parsing for v2 metadata such that if the dtype specified is `str`, and 0 is the provided `fill_value`, this value is replaced with `""`

Closes #2792

TODO:

As a side note, I was a bit surprised at how many places there were that independently seemed to be concerned with metadata parsing. This... was slightly confusing, as values undergo multiple `parse_*` steps after having been parsed already (so that it is unclear without printing stack and dropping into a debugger when e.g. dtype was the provided dtype vs when it was the already parsed dtype. In this case, as the dtype of `object` is produced by `parse_dtype` in `init_array`, the only place where it was possible to check on the `dtype` specified by the user was at that point (way up the stack from calls into v2.py). Not sure I quite understand the organization (though I'm sure there are good reasons for it; perhaps historical and compatibility related) and I'd be very interested in a rough sketch of the motivation that I'm surely failing to see


src/zarr/core/array_spec.py
```diff
@@ -102,7 +102,9 @@ def __init__(
     ) -> None:
         shape_parsed = parse_shapelike(shape)
         dtype_parsed = np.dtype(dtype)
-        fill_value_parsed = parse_fill_value(fill_value)
+        # No zarr_format available here...
+        # fill_value_parsed = parse_fill_value(fill_value, dtype, zarr_format=2)
+        fill_value_parsed = fill_value
```
moradology, line 107: Not sure what to make of this. The fill value is parsed way up the stack from here or else the dtype we have to inspect is lost. Should I just toss this out?

d-v-b:
> As a side note, I was a bit surprised at how many places there were that independently seemed to be concerned with metadata parsing. This... was slightly confusing, as values undergo multiple `parse_*` steps after having been parsed already (so that it is unclear without printing stack and dropping into a debugger when e.g. dtype was the provided dtype vs when it was the already parsed dtype. In this case, as the dtype of `object` is produced by `parse_dtype` in `init_array`, the only place where it was possible to check on the `dtype` specified by the user was at that point (way up the stack from calls into v2.py). Not sure I quite understand the organization (though I'm sure there are good reasons for it; perhaps historical and compatibility related) and I'd be very interested in a rough sketch of the motivation that I'm surely failing to see

Sorry that all the parsing was confusing, we should definitely avoid parsing the same things twice! Could you open an issue for the values that are getting parsed multiple times?

src/zarr/core/common.py
```diff
@@ -150,9 +150,11 @@ def parse_shapelike(data: int | Iterable[int]) -> tuple[int, ...]:
     return data_tuple
 
 
-def parse_fill_value(data: Any) -> Any:
+def parse_fill_value(fill_value: Any, dtype: Any, zarr_format: ZarrFormat) -> Any:
+    if zarr_format == 2 and (dtype is str or dtype == "str") and fill_value == 0:
```
d-v-b, line 154: I think this needs at least a comment explaining why we are doing this. what was zarr-python 2's behavior in this situation?

moradology:
Looks like this has been superseded by 2799

LDeakin:
@moradology #2799 is resolving a separate issue (`None` fill values) and does not supersede this PR. But there is uncertainty over how #2792 should be handled, as the behaviour has changed between `zarr-python` 2/3, see https://github.com/zarr-developers/zarr-python/issues/2792#issuecomment-2637954231.

PR #2799, opened 2025-02-05
fix: implicit fill value initialisation

LDeakin:
Fixes omissions in https://github.com/zarr-developers/zarr-python/pull/2274

- initialise empty chunks to the default fill value during writing
- add default fill value for datetime, timedelta, structured, and other (void* fixed size) data types
- add test for "other" data type

Fixes the below:
```python
array = zarr.create_array(
    dtype=np.uint8,
    shape=(4,),
    chunks=(2,),
    fill_value=None,
    zarr_format=2,
)
array[:3] = [1, 2, 3]
print(array[:]) # [1, 2, 3, undefined]
```

TODO:

### DIFF
--- a/src/zarr/core/metadata/v2.py
+++ b/src/zarr/core/metadata/v2.py
@@ -350,7 +350,7 @@ def _default_fill_value(dtype: np.dtype[Any]) -> Any:
     elif dtype.kind in "UO":
         return ""
     elif dtype.kind in "Mm":
-        return dtype.type('nat')
+        return dtype.type("nat")
     elif dtype.kind == "V":
         if dtype.fields is not None:
             default = tuple([_default_fill_value(field[0]) for field in dtype.fields.values()])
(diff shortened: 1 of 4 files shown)


src/zarr/core/metadata/v2.py
```diff
@@ -349,6 +349,14 @@ def _default_fill_value(dtype: np.dtype[Any]) -> Any:
         return b""
     elif dtype.kind in "UO":
         return ""
+    elif dtype.kind in "Mm":
```
LDeakin, line 352: This could use some keen eyes, I've never used these dtypes.

d-v-b: I'm also unfamiliar with those dtypes. perhaps you could find examples from the tests on the v2 branch?

d-v-b (approved):

Issue #2741, opened 2025-01-21
Broken NaN encoding when writing v2 storage format from v3 library


d70-t:
### Zarr version

3.0.1

### Numcodecs version

0.15.0

### Python Version

3.12.8

### Operating System

Mac

### Installation

micromamba create -n zarr3 python 'zarr>=3'

### Description

When writing an float array with `fill_value=np.nan` to a version 2 store, [according to the spec](https://zarr-specs.readthedocs.io/en/latest/v2/v2.0.html#fill-value-encoding), the fill value has to be encoded as a string (i.e.: `"NaN"`). This has to be the case, as JSON doesn't support a NaN literal. However, when doing so using `zarr>=3`, the fill value is **not** encoded as string. The resulting `.zarray` is thus not a valid JSON and can't be read using other JSON parsers.

This behaviour could be a regession of #412.
The bug was originally found by @lkluft.

### Steps to reproduce

Using this Python script as `check_nan_encoding.py`:

```python
import zarr
import numpy as np

import numcodecs
import sys

print("zarr version", zarr.__version__)
print("numcodecs version", numcodecs.__version__)
print("python version", sys.version)

def make_array(**kwargs):
    if zarr.__version__ >= "3":
        return zarr.create_array(zarr_format=2, **kwargs)
    else:
        return zarr.create(**kwargs)

def to_str(buffer):
    if zarr.__version__ >= "3":
        return buffer.to_bytes().decode("utf-8")
    else:
        return buffer.decode("utf-8")

def get_fill_value_line(store):
    return [line.strip(" ,")
            for line in to_str(store[".zarray"]).split("\n")
            if "fill_value" in line][0]

store = {}
z = make_array(
    store=store,
    shape=(1,),
    chunks=(1,),
    dtype="f4",
    fill_value=np.nan,
)
print(get_fill_value_line(store))
```

When running in a `zarr>=3` environment:
```bash
micromamba create -n zarr3 python 'zarr>=3'
micromamba run -n zarr3 python3 check_nan_encoding.py
```
the following is printed:
```
zarr version 3.0.1
numcodecs version 0.15.0
python version 3.12.8 | packaged by conda-forge | (main, Dec  5 2024, 14:19:53) [Clang 18.1.8 ]
"fill_value": NaN
```
Which is **not** a valid JSON encoding.

---

When running the same code in a `zarr>=2,<3` environment:
```bash
[...]

PR #2802, opened 2025-02-05
Fix fill_value serialization issues

moradology:
The current serialization of `fill_value` in `ArrayV2Metadata` does not fully conform to the spec, particularly for:
- NaN and Infinity values, which must be serialized as strings ("NaN", "Infinity", "-Infinity").
- Complex numbers (np.complex64, np.complex128), which must be stored as two-element arrays [real, imag] according to the above Nan/Inf rules.

#### Changes
- Updated _sanitize_fill_value() to enforce correct JSON serialization.
- Fixed test_v2meta_fill_value_serialization() to compare expected and actual JSON using a normalized representation.
- Introduced property-based testing with Hypothesis to generate valid input cases and verify compliance.
- Enforce compliance (to some degree) by setting `allow_nan` to `False` on `json.dumps`

Resolves: #2741

TODO:


moradology:
The bad news is, this issue is going to be slightly more than I've said here. The good news is that the property-based tests caught some edge cases.

dcherian:
Nice, I've been meaning to add this to Zarr:
```python
@st.composite
def v3_array_metadata(draw: st.DrawFn) -> bytes:
    from zarr.codecs.bytes import BytesCodec
    from zarr.core.chunk_grids import RegularChunkGrid
    from zarr.core.chunk_key_encodings import DefaultChunkKeyEncoding
    from zarr.core.metadata.v3 import ArrayV3Metadata

    # separator = draw(st.sampled_from(['/', '\\']))
    shape = draw(array_shapes)
    ndim = len(shape)
    chunk_shape = draw(npst.array_shapes(min_dims=ndim, max_dims=ndim))
    dtype = draw(zrst.v3_dtypes())
    fill_value = draw(npst.from_dtype(dtype))
    dimension_names = draw(
        st.none() | st.lists(st.none() | simple_text, min_size=ndim, max_size=ndim)
    )

    metadata = ArrayV3Metadata(
        shape=shape,
        data_type=dtype,
        chunk_grid=RegularChunkGrid(chunk_shape=chunk_shape),
        fill_value=fill_value,
        attributes=draw(simple_attrs),
        dimension_names=dimension_names,
        chunk_key_encoding=DefaultChunkKeyEncoding(separator="/"),  # FIXME
        codecs=[BytesCodec()],
        storage_transformers=(),
    )

    return metadata.to_buffer_dict(prototype=default_buffer_prototype())["zarr.json"]
```

What do you think of a `array_metadata_json(zarr_formats...)` strategy that just returns the JSON and we can test whether that satisfies the spec for V2 and V3?

moradology:
I love the idea. I was thinking the other day that the obvious path out of the bugs that are currently popping up would be property-based testing, so I was pretty pleased to see that there's already some work in that direction

Out of curiosity, do we have something like json schema that we could apply against the outputs to at least verify structure? We'd still need to define all the rules that exist in terms of value/type dependencies etc. but if it's an easy win if that exists somewhere

d-v-b:
I'm not aware of a JSON schema definition for the array metadata. If one existed, it would necessarily only support partial validation, because JSON schema can't express certain invariants in the metadata document, like the requirement that dimensional attributes (shape, chunk_shape, etc) be consistent.

dcherian:
A fairly easy alternative way to handle this would be to simply write  a test that takes the `arrays` strategy, extracts the metadata, converts to JSON, and then asserts that the JSON meets spec (as best as we can). 

I still think a generic metadata strategy is probably useful.

moradology:
Did some refactoring and organization of property-based testing code and added round trip testing for `ArrayV2Metadata` (which hopefully captures some important expectations about these things. Still need to fully satisfy the linter and make sure similar guarantees are tested for `ArrayV3Metadata`

dcherian:
Hye @moradology I added an `array_metadata` strategy in https://github.com/zarr-developers/zarr-python/blob/e8bfb6484aeb543c21b9b0f6de2e0f0a034fc5d7/src/zarr/testing/strategies.py#L110 would you mind merging your changes in with that strategy please?

moradology:
Definitely - do you mind pulling the generators up into their own submodule as I've done here, given their similar and unique functionality?

dcherian:
I think testing/strategies.py is right place. Everything in there already handles the v2 vs v3 complexity

dcherian:
It feels a bit premature to me. In any case, I find it nice to keep such refactoring PRs separate so that we can review (much) smaller diffs. In my experience, keeping the diffs small is the best way to keep open-source contributions easy to merge.

moradology:
Sounds like a plan. As a relatively new contributor to this library, I appreciate the firm opinions!

dcherian:
Thank *you* for considering the opinions in a constructive manner!

moradology:
OK, so to get the property-based tests working I had to make some decisions about what serialization strategies looked like. I tried to follow the instructions available [here](https://zarr-specs.readthedocs.io/en/latest/v2/v2.0.html), but other issues that have come to the surface in the history of `zarr-python` suggest there are likely blindspots in this document. [This part](https://github.com/zarr-developers/zarr-python/pull/2802/files#diff-535b8bc19e7bf0cd5b07eb0d481ba8676f21b41cd352ddcbfc92c4667f124804R197-R213), in particular, deserves special attention

tests/test_properties.py
```diff
@@ -135,3 +185,47 @@ async def test_roundtrip_array_metadata(
 #     nparray = data.draw(np_arrays)
 #     zarray = data.draw(arrays(arrays=st.just(nparray)))
 #     assert_array_equal(nparray, zarray[:])
+
+
+@given(data=st.data(), zarr_format=zarr_formats)
+def test_meta_roundtrip(data: st.DataObject, zarr_format: int) -> None:
```
dcherian, line 191: can we update the `test_roundtrip_array_metadata` test further up in the file instead please?

tests/test_properties.py
```diff
@@ -135,3 +185,47 @@ async def test_roundtrip_array_metadata(
...
+        metadata_roundtripped = ArrayV3Metadata.from_dict(zarray_dict)
+
+    orig = dataclasses.asdict(metadata)
+    rt = dataclasses.asdict(metadata_roundtripped)
+
+    assert deep_equal(orig, rt), f"Roundtrip mismatch:\nOriginal: {orig}\nRoundtripped: {rt}"
+
+
+@given(npst.from_dtype(dtype=np.dtype("float64"), allow_nan=True, allow_infinity=True))
+def test_v2meta_nan_and_infinity(fill_value):
```
dcherian, line 212: This could look more like https://github.com/zarr-developers/zarr-python/pull/2847/files#diff-d318cba7c9e4a6983338cf21df1db66aab796137a2fb4a76ce48c0afa17de2f9

We could abstract out a `assert_valid_v2_json_dict` and similarly for v3.

tests/test_properties.py
moradology: Honestly, missed it. Will do.

moradology: OK, updated test names and added docstrings to clarify what's going on

tests/test_properties.py
dcherian: Let's just copy the test over and you can expand them here. I can handle the merge conflicts. It may very well be that this one goes in first :)

moradology: OK, so I'm curious about something. See this line: https://github.com/zarr-developers/zarr-python/blob/main/src/zarr/core/metadata/v3.py#L413C1-L414C1
Is that type correct? If so, it seems that numpy types should be serialized after `to_dict`. As of now, they're not. So what's the desired behavior?

dcherian: don't know. ping @d-v-b

dcherian:
as promised, I fixed the conflict :)

tests/test_properties.py
d-v-b: no, that type is wrong and it bugs me! for v3 metadata `to_dict` returns an instance of `DataType` for the `data_type` key, and I think we registered that type with a custom JSON encoder.

d-v-b: I don't remember why this decision was made but we should definitely fix it. fwiw, `dict[str, JSON]` is also sub-optimal, given that the keys of the metadata document are (almost) entirely static, and so we could do typeddicts here

moradology: OK, so the immediately obvious thing is that the serialization logic I've added to `to_dict` needs to be pushed down into `to_buffer_dict` and `to_dict` should retain its (potentially) not-directly-serializable python types.

From there, a typed-dict can be implemented for `to_dict` output.

moradology: Alright, after a fair bit of back and forth I've gotten to a better (but imperfect) place in terms of serialization/deserialization. A couple of high points:
1. fill value serialization is happening in exactly one function and inside of the `to_buffer_dict` method rather than `to_dict`. V2's to_dict simply 
2. deserialization/parsing happens entirely inside of one function rather than in a few different places. Hopefully this makes things a bit easier to follow: https://github.com/zarr-developers/zarr-python/pull/2802/files#diff-535b8bc19e7bf0cd5b07eb0d481ba8676f21b41cd352ddcbfc92c4667f124804R318-R392

I also played around with removing the `dict[str, JSON` type signature from the to_dict/from_dict methods but upon further inspection, those seem to be expected/baked in for other `Metadata` subclasses. In particular, `ChunkKeyEncoding`, `GroupMetadata`, and `ChunkGrid`. It seems to me that really disentangling this type signature might be out of scope here and probably deserves a separate PR.

moradology:
Codecov is being overly sensitive at this point. The added lines are *extensively* tested via the round tripping now done via property-based tests

moradology:
@d-v-b or @dcherian - thoughts on this set of changes? No pressure, I just don't want this to get too stale if possible. The PR is now non-trivial, but I think it does a decent job of organizing and defining serialization/deserialization behaviors that were previously hard to see at a glance or else not well defined. It also verifies round tripping with property-based tests, which is how certain undefined behaviors were uncovered (e.g. 'NaT' for `datetime` instances)

src/zarr/core/metadata/v2.py
```diff
@@ -332,6 +367,18 @@ def parse_fill_value(fill_value: object, dtype: np.dtype[Any]) -> Any:
             raise ValueError(
                 f"fill_value {fill_value!r} is not valid for dtype {dtype}; must be a unicode string"
             )
+    elif dtype.kind in "SV" and isinstance(fill_value, str):
+        fill_value = base64.standard_b64decode(fill_value)
```
d-v-b, line 371: out of scope but does anyone know why we base64 encode scalars with dtype `S` ? `S` is ascii-encoded strings, which shouldn't need any encoding / decoding.

d-v-b (approved):

d-v-b:
this looks good to me, I'm going to merge soon unless there are objections. good to keep things moving.

src/zarr/core/metadata/v2.py
```diff
@@ -108,6 +109,29 @@ def shards(self) -> ChunkCoords | None:
...
 
     def to_buffer_dict(self, prototype: BufferPrototype) -> dict[str, Buffer]:
+        def _serialize_fill_value(fv: Any) -> JSON:
+            if self.fill_value is None:
+                pass
+            elif self.dtype.kind in "SV":
+                # There's a relationship between self.dtype and self.fill_value
+                # that mypy isn't aware of. The fact that we have S or V dtype here
+                # means we should have a bytes-type fill_value.
+                fv = base64.standard_b64encode(cast(bytes, self.fill_value)).decode("ascii")
```
dstansby, line 119: Instead of overwriting the `fv` variable, can you use an new variable (`serialised`? `fvs`?) to avoid the type casting right at the end of this function?

src/zarr/core/metadata/v2.py
```diff
@@ -108,6 +109,29 @@ def shards(self) -> ChunkCoords | None:
         return None
 
     def to_buffer_dict(self, prototype: BufferPrototype) -> dict[str, Buffer]:
+        def _serialize_fill_value(fv: Any) -> JSON:
```
dstansby, line 112: There's several branches of this functions that codecov is reporting as being untested - do tests need adding for these?

src/zarr/core/metadata/v2.py
```diff
@@ -108,6 +109,29 @@ def shards(self) -> ChunkCoords | None:
...
+            if self.fill_value is None:
+                pass
+            elif self.dtype.kind in "SV":
+                # There's a relationship between self.dtype and self.fill_value
+                # that mypy isn't aware of. The fact that we have S or V dtype here
+                # means we should have a bytes-type fill_value.
+                fv = base64.standard_b64encode(cast(bytes, self.fill_value)).decode("ascii")
+            elif isinstance(fv, np.datetime64):
+                if np.isnat(fv):
+                    fv = "NaT"
```
dstansby, line 122: As far as I can see this branch isn't needed, because `np.datetime_as_string()` converts `NaT` correctly:

```python
>>> t
np.datetime64('NaT')
>>> np.datetime_as_string(t)
np.str_('NaT')
```

src/zarr/core/metadata/v2.py
dstansby, line 112: This new funciton is inside an already really long function definition (with another nested function underneath). To make it more readable, could you extract this function by itself? It would need to gain a `array_dtype` argument to replace to `self.dtype`, but I think that's much better than having these nested functions.

src/zarr/core/metadata/v2.py
```diff
@@ -191,18 +213,11 @@ def from_dict(cls, data: dict[str, Any]) -> ArrayV2Metadata:
 
         _data = {k: v for k, v in _data.items() if k in expected}
 
-        return cls(**_data)
+        return cls(**cast(Mapping[str, Any], _data))
```
dstansby, line 216: Why does this need the new cast?

src/zarr/core/metadata/v2.py
```diff
@@ -300,7 +315,27 @@ def parse_metadata(data: ArrayV2Metadata) -> ArrayV2Metadata:
     return data
 
 
-def parse_fill_value(fill_value: object, dtype: np.dtype[Any]) -> Any:
+def parse_structured_fill_value(fill_value: Any, dtype: np.dtype[Any]) -> Any:
+    """Handle structured dtype/fill value pairs"""
+    try:
+        if isinstance(fill_value, list):
+            fill_value = tuple(fill_value)
```
dstansby, line 322: I think it would be clearer in this function to just return all the values in the branches, intead of assigning to `fill_value` and then immediately returning it after the try/except statement

src/zarr/core/metadata/v2.py
```diff
@@ -317,13 +352,13 @@ def parse_fill_value(fill_value: object, dtype: np.dtype[Any]) -> Any:
     """
 
     if fill_value is None or dtype.hasobject:
-        # no fill value
         pass
+    elif dtype.fields is not None:
```
dstansby, line 356: What case is the `dtype.fields is not None:` guarding against? I think worth adding a comment explaining why this is here, it's not clear to me.

src/zarr/core/metadata/v2.py
```diff
@@ -332,6 +367,18 @@ def parse_fill_value(fill_value: object, dtype: np.dtype[Any]) -> Any:
             raise ValueError(
                 f"fill_value {fill_value!r} is not valid for dtype {dtype}; must be a unicode string"
             )
+    elif dtype.kind in "SV" and isinstance(fill_value, str):
```
dstansby, line 370: All of this logic is the same as new lines added above. Would it be possible to deduplicate it somehow?

tests/test_properties.py
```diff
@@ -25,8 +30,53 @@
 )
 
 
+def deep_equal(a, b):
```
dstansby, line 33: Would be good to type this funciton.

tests/test_properties.py
```diff
@@ -155,3 +234,68 @@ def test_array_metadata_meets_spec(store: Store, meta: ArrayV2Metadata | ArrayV3
 #     nparray = data.draw(np_arrays)
 #     zarray = data.draw(arrays(arrays=st.just(nparray)))
 #     assert_array_equal(nparray, zarray[:])
+
+
+def serialized_float_is_valid(serialized):
```
dstansby, line 239: Would be good to type this function

tests/test_properties.py
```diff
@@ -155,3 +234,68 @@ def test_array_metadata_meets_spec(store: Store, meta: ArrayV2Metadata | ArrayV3
...
+    """
+    if isinstance(serialized, numbers.Real):
+        return True
+    return serialized in ("NaN", "Infinity", "-Infinity")
+
+
+@given(meta=array_metadata())  # type: ignore[misc]
+def test_array_metadata_meets_spec(meta: ArrayV2Metadata | ArrayV3Metadata) -> None:
+    """
+    Validate that the array metadata produced by the library conforms to the relevant spec (V2 vs V3).
```
dstansby, line 261: Nice docs on the tests 👍

src/zarr/core/metadata/v2.py
moradology: I'm not really sure what to make of that - the round tripping being done *should* very adequately cover the relevant behaviors. Do you think it would be appropriate to test this function separately as well?

src/zarr/core/metadata/v2.py
moradology: Not entirely sure what the type inference was choking on there - been a bit since I was tangling with `mypy` here but I'll double check

src/zarr/core/metadata/v2.py
dstansby: It might be that codecov is just incorrect - if you're happy that all these branches are being tested, then I definitely trust you more than codecove 😄

src/zarr/core/metadata/v2.py
moradology: Yeah, this was actually super confusing. Basically, once I unified all fill_value deserialization structured datatype tests started failing(!). It appears that they were being handled correctly before this as a result of code path differences between `from_dict` and `__init__`, so what is going on here is *explicit* handling of that case inside the singular deserialization function for v2 fill values.

That's kind of hard to convey succinctly in a comment form and it seems to be of mostly historical relevance but I'll add a comment that explains what this branch is doing. Thinking this as of now:
```python
    # The dtype is structured (has multiple fields), so the fill_value might be a
    # compound value (e.g., a tuple or dict) that needs field-wise processing.
    # We use parse_structured_fill_value to correctly convert each component.
```

d-v-b: fwiw, over in #2874 I'm associating fill value definition and ser/deserialization with separate dtype classes, which should remove a lot of these confusing `if .. elif ...` blocks

src/zarr/core/metadata/v2.py
moradology: I assume you're referring to the handling in `parse_structured_fill_value`. There are slight differences between these cases that make deduplication seem pretty difficult to my mind. In the structured case, what's being handled is a potentially serialized (to string) structured fill value (hence the next step which takes that and constructs the compound value): `fill_value = np.frombuffer(decoded, dtype=dtype)[0]`.

src/zarr/core/metadata/v2.py
moradology: Was able to remove this. It was the result of me trying to be consistent with the superclass `Metadata` in the type signature of `from_dict` and the fact that dicts are invariant (while we need covariance for JSON being recognized as Any). I've reverted that change (though at some point it might be good to button that down)

moradology:
Glad I got the pushback to go through things again because the property-based tests came across an error related to floating point precision at really big numbers that (in extremely rare cases of integers near `sys.maxsize`) was causing problems. Also trimmed some redundancy

moradology:
Well, this timeout error is pretty annoying! Can't reliably reproduce it locally but don't want to bump the deadline up *too* much... 
image

tests/test_v2.py
```diff
@@ -315,6 +316,90 @@ def test_structured_dtype_roundtrip(fill_value, tmp_path) -> None:
...
+        "string_input",
+        "dictionary_input",
+        "dictionary_input_extra_fields",
+    ],
+)
+def test_parse_structured_fill_value_valid(
+    fill_value: Any, dtype: np.dtype[Any], expected_result: Any
+) -> None:
+    result = _parse_structured_fill_value(fill_value, dtype)
+    print(result)
```
dstansby, line 370:
```suggestion
```

dstansby (approved):
I spotted a wild `print()` statement that should probably be removed 😄 - otherwise looks good to me now, thanks for making the changes!

PR #2804, opened 2025-02-06
Enable stateful tests for LocalStore

dcherian:
Also restrict stateful tests to slow hypothesis CI workflow to speed up the CI a bit (cc @dstansby)

Seems to work:
-  `SKIPPED [6] tests/test_store/test_stateful.py: need --run-slow-hypothesis option to run`

Also fixes the LocalStore stateful test

TODO:

### DIFF
--- a/changes/2804.feature.rst
+++ b/changes/2804.feature.rst
@@ -1,0 +1,1 @@
+:py:class:`LocalStore` learned to ``delete_dir``.
--- a/changes/2804.feature.rst
+++ b/changes/2804.feature.rst
@@ -1,1 +1,1 @@
-:py:class:`LocalStore` learned to ``delete_dir``.
+:py:class:`LocalStore` learned to ``delete_dir``. This makes array and group deletes more efficient.
--- a/src/zarr/storage/_local.py
+++ b/src/zarr/storage/_local.py
@@ -214,6 +214,12 @@ class LocalStore(Store):
         path = self.root / prefix
         if path.is_dir():
             shutil.rmtree(path)
+        elif path.is_file():
+            raise ValueError(f"delete_dir was passed a {prefix=!r} that is a file")
+        else:
+            # Non-existent directory
+            # This path is tested by test_group:test_create_creates_parents for one
+            pass
 
     async def exists(self, key: str) -> bool:
         # docstring inherited
--- a/src/zarr/storage/_local.py
+++ b/src/zarr/storage/_local.py
@@ -208,6 +208,13 @@ class LocalStore(Store):
         else:
             await asyncio.to_thread(path.unlink, True)  # Q: we may want to raise if path is missing
 
+    async def delete_dir(self, prefix: str) -> None:
+        # docstring inherited
+        self._check_writable()
+        path = self.root / prefix
+        if path.is_dir():
+            shutil.rmtree(path)
+
     async def exists(self, key: str) -> bool:
         # docstring inherited
         path = self.root / key
--- a/src/zarr/testing/stateful.py
+++ b/src/zarr/testing/stateful.py
@@ -25,9 +25,6 @@ from zarr.testing.strategies import keys as zarr_keys
 
 MAX_BINARY_SIZE = 100
 
-# Handle possible case-insensitive file systems (e.g. MacOS)
-node_names = node_names.map(lambda x: x.lower())
-
 
 def split_prefix_name(path: str) -> tuple[str, str]:
     split = path.rsplit("/", maxsplit=1)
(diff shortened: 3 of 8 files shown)


changes/2804.feature.rst
```diff
@@ -0,0 +1 @@
+:py:class:`LocalStore` learned to ``delete_dir``.
```
dstansby, line 1:
```suggestion
Implemented ``delete_dir`` for :py:class:`LocalStore`. This means it is now possible to delete an entire directory at once, instead of having to delete every individual key under a given directory in the store.
```

src/zarr/storage/_local.py
```diff
@@ -208,6 +208,13 @@ async def delete(self, key: str) -> None:
         else:
             await asyncio.to_thread(path.unlink, True)  # Q: we may want to raise if path is missing
 
+    async def delete_dir(self, prefix: str) -> None:
+        # docstring inherited
+        self._check_writable()
+        path = self.root / prefix
+        if path.is_dir():
+            shutil.rmtree(path)
```
dstansby, line 216: Am I right in thinking this will just silently do nothing if a file is provided to this function? In that case we should either:

1. Document this behaviour, or 
2. Do something else like raise a warning or error

Is there a precedent from other stores which option we should take?

src/zarr/testing/stateful.py
```diff
@@ -68,6 +68,9 @@ def can_add(self, path: str) -> bool:
     # -------------------- store operations -----------------------
     @rule(name=node_names, data=st.data())
     def add_group(self, name: str, data: DataObject) -> None:
+        # Handle possible case-insensitive file systems (e.g. MacOS)
+        if isinstance(self.store, LocalStore):
+            name = name.lower()
```
dstansby, line 73: This line doesn't seem to be covered by the tests, but looks like it's intended to be run? Do you know whether this is a coverage upload issue, or if the line actually isn't being run in tests?

src/zarr/testing/strategies.py
```diff
@@ -218,10 +219,14 @@ def key_ranges(
         [(key, (range_start, range_end)),
          (key, (range_start, range_end)),...]
     """
+
+    def make_request(start: int, length: int) -> RangeByteRequest:
+        return RangeByteRequest(start, end=min(start + length, max_size))
```
dstansby, line 224: This also isn't showing up in code coverage.

src/zarr/testing/strategies.py
```diff
@@ -209,7 +210,7 @@ def basic_indices(draw: st.DrawFn, *, shape: tuple[int], **kwargs: Any) -> Any:
 
 
 def key_ranges(
-    keys: SearchStrategy = node_names, max_size: int | None = None
+    keys: SearchStrategy = node_names, max_size: int = sys.maxsize
```
dstansby, line 213: This is [part of our public API](https://zarr--2804.org.readthedocs.build/en/2804/api/zarr/testing/strategies/index.html#zarr.testing.strategies.key_ranges), so changing the default here is an API change. What was the motivation for changing it - does hypothesis not work well with `max_value=None`?

tests/test_store/test_stateful.py
```diff
@@ -28,11 +27,7 @@ def mk_test_instance_sync() -> None:
...
-        pytest.skip(reason="This test has errors")
-    elif isinstance(sync_store, MemoryStore):
-        run_state_machine_as_test(mk_test_instance_sync, settings=Settings(max_examples=50))
-    else:
-        run_state_machine_as_test(
-            mk_test_instance_sync, settings=Settings(report_multiple_bugs=True)
-        )
+
+    if isinstance(sync_store, LocalStore):
+        pytest.skip(reason="Test isn't suitable for LocalStore.")
```
dstansby, line 32: Could you expand the reason here a bit to explain why it's not suitable? Just thinking of down the line when someone tries to understand this a bit more info would be helpful.

dstansby (changes requested):
👍 great that the slow tests aren't running on all the test runs now! I left some comments/requests for updates/questions inline.

src/zarr/testing/strategies.py
dcherian: that's a bad report, or perhaps out-of-date. this PR also enables coverage recording for the hypothesis tests.

src/zarr/testing/stateful.py
dcherian: that's a bad report, or perhaps out-of-date. this PR also enables coverage recording for the hypothesis tests.

src/zarr/testing/strategies.py
dcherian: Cleans up the use of `min` below, that's all.

changes/2804.feature.rst
dcherian: I'll add another line.

src/zarr/storage/_local.py
dcherian: Nice catch. Now it raises an error if a file is passed and does nothing if passed a non-existent path.

src/zarr/testing/strategies.py
dstansby: 👍

dstansby (approved):

PR #2807, opened 2025-02-06
Strengthen pickling test and fix ZipStore __getstate__()

maxrjones:
It's important for stores that manually implement `__getstate__` and `__setstate__` that the store behavior is tested beyond a simply equality check. This is motivated by https://github.com/zarr-developers/zarr-python/pull/1661, where I first added incorrect serialization functionality even though it passed this test. This PR adds a simple set/get check on the serialization test and exposes an issue with ZipStore.

TODO:

### DIFF
--- a/src/zarr/storage/_zip.py
+++ b/src/zarr/storage/_zip.py
@@ -108,7 +108,8 @@ class ZipStore(Store):
         self._sync_open()
 
     def __getstate__(self) -> dict[str, Any]:
-        state = self.__dict__
+        # We need a copy to not modify the state of the original store
+        state = self.__dict__.copy()
         for attr in ["_zf", "_lock"]:
             state.pop(attr, None)
         return state
--- a/changes/2807.bugfix.rst
+++ b/changes/2807.bugfix.rst
@@ -1,0 +1,1 @@
+Fix pickling for ZipStore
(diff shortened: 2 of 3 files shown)


src/zarr/storage/_zip.py
```diff
@@ -108,7 +108,8 @@ async def _open(self) -> None:
         self._sync_open()
 
     def __getstate__(self) -> dict[str, Any]:
-        state = self.__dict__
+        # We need a copy to not modify the state of the original store
+        state = self.__dict__.copy()
```
maxrjones, line 112: I'm not an expert on these methods, so I'd appreciate if someone with more experience could verify that we need a copy. Maybe @martindurant could help?

maxrjones: Meaning, is there a better way to approach this than the current implementation?

martindurant: Since the `__init__` doesn't take many arguments, you could pluck the ones you need, and make this a `__reduce__` - slightly simpler. It means deciding which arguments you yes need, rather than the `pop` statements here deciding which ones you don't.

But yes, you need a copy() in order not to mutate the state of the original object.

maxrjones: thank you!

maxrjones: I looked into using `__reduce__` but find the current implementation to be simpler given that `ZipStore` requires kwargs, so I would need to use most of the optional tuple values returned by `__reduce__`. So, I'll leave this as is. thanks again for the advice

jhamman (approved):

PR #2815, opened 2025-02-11
Multiprocessing support

### LINKED ISSUE #2812: multiprocessing support
We have had multiple reports of problems using zarr-python 3.x in a multiprocessing context: #2729, https://ossci.zulipchat.com/#narrow/channel/423692-Zarr/topic/Zarr.20v3.20force.20single.20threaded.20for.20forked.20processes. 

We should:
- ensure that our array / group classes play nicely in a multiprocessing context
- ensure that multiprocessing support is properly tested in the test suite

d-v-b:
I wanted to use this PR to add tests to ensure that basic array operations work in a multiprocessing context. But right now I just have 1 test, and it's failing with an infinite stall. Since this breaks the entire test suite, we should also fix the underlying bug in this PR. Stay tuned!

closes #2812

### DIFF
--- a/src/zarr/core/sync.py
+++ b/src/zarr/core/sync.py
@@ -101,7 +101,9 @@ def reset_resources_after_fork() -> None:
     _executor = None
 
 
-os.register_at_fork(after_in_child=reset_resources_after_fork)
+# this is only available on certain operating systems
+if hasattr(os, "register_at_fork"):
+    os.register_at_fork(after_in_child=reset_resources_after_fork)
 
 
 async def _runner(coro: Coroutine[Any, Any, T]) -> T | BaseException:
--- a/src/zarr/core/sync.py
+++ b/src/zarr/core/sync.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import asyncio
 import atexit
 import logging
+import os
 import threading
 from concurrent.futures import ThreadPoolExecutor, wait
 from typing import TYPE_CHECKING, TypeVar
@@ -87,6 +88,20 @@ def cleanup_resources() -> None:
 
 
 atexit.register(cleanup_resources)
+
+
+def reset_resources_after_fork() -> None:
+    """
+    Ensure that global resources are reset after a fork. Without this function,
+    forked processes will retain invalid references to the parent process's resources.
+    """
+    global loop, iothread, _executor
+    loop[0] = None
+    iothread[0] = None
+    _executor = None
+
+
+os.register_at_fork(after_in_child=reset_resources_after_fork)
 
 
 async def _runner(coro: Coroutine[Any, Any, T]) -> T | BaseException:
--- a/src/zarr/core/sync.py
+++ b/src/zarr/core/sync.py
@@ -96,9 +96,13 @@ def reset_resources_after_fork() -> None:
     forked processes will retain invalid references to the parent process's resources.
     """
     global loop, iothread, _executor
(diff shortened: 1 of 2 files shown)


d-v-b:
the underlying issue here is that store operations don't work in child processes.

d-v-b:
@martindurant this seems like your wheelhouse -- would you mind checking this over?

martindurant:
> sets the persistent loops to None in child processes

This is absolutely what you must do.
asyncio event loops, threads and file handles are the main category of things that "don't work" in a child process, with the specific problem depending on which thing it is and how the process is made.

d-v-b:
I think this PR is ready for review. The test I added checks that array indexing works in a multiprocessing context, which is a fine test, but the actual substance is whether our asynchronous store methods work with multiprocessing, and that can be tested without arrays. But I'd rather treat that as a test refactor for the future.

src/zarr/core/sync.py
```diff
@@ -89,6 +90,20 @@ def cleanup_resources() -> None:
 atexit.register(cleanup_resources)
 
 
+def reset_resources_after_fork() -> None:
+    """
+    Ensure that global resources are reset after a fork. Without this function,
+    forked processes will retain invalid references to the parent process's resources.
+    """
+    global loop, iothread, _executor
```
martindurant, line 98: strictly speaking, loop and iothread don't need to be global, since they are mutated in-place.

src/zarr/core/sync.py
```diff
@@ -89,6 +90,20 @@ def cleanup_resources() -> None:
...
+    loop[0] = None
+    iothread[0] = None
+    _executor = None
```
martindurant, line 101: Fix coverage: the main process never sees this function getting called. I suppose a test could explicitly call it.

tests/test_array.py
```diff
@@ -1382,3 +1384,39 @@ def test_roundtrip_numcodecs() -> None:
...
+    return arr[index]
+
+
+@pytest.mark.parametrize(
+    "method",
+    [
+        pytest.param(
+            "fork",
+            marks=pytest.mark.skipif(
+                sys.platform in ("win32", "darwin"), reason="fork not supported on Windows or OSX"
```
martindurant, line 1399: Because windows should only perform spawn? This decorator is a bit verbose, it would be OK to put `if ... : pytest.skip()` in the body of the function.

martindurant (approved):
I'm OK keeping these, effectively, as integration tests rather than store-specific. The top-level sync() we call internally is important here.

src/zarr/core/sync.py
d-v-b: I will just exempt these lines from coverage, since they are being tested indirectly

d-v-b:
maybe someone can explain what codecov is going on about? I find this hard to interpret:

image

I'm going to merge this later today unless there are objections.

tests/test_array.py
d-v-b: I think I prefer (verbose) parametrization over checking the OS in the test itself. The former gives better feedback in the test summary.

martindurant:
The "indirect changes" tab is picking up many lines in testing/stateful.py . I don't know why they show up here - maybe this file wasn't being tracked at all before?

martindurant:
OK, it looks like not all of the coverage reports were in yet :) 
Good now.

d-v-b:
thanks for your help martin!

PR #2819, opened 2025-02-12
Make `create_array` signatures consistent

### LINKED ISSUE #2810: `Group.create_array` does not take a `data` keyword argument
This is inconsistent with the `create_array` function and should be fixed. 

I'd be interested in ideas for how we can catch these omissions in tests. One idea would be to define functions / methods that are wrapping some other function, and checking that the signature of the wrapping functions / methods satisfies some check.

d-v-b:
This PR ensures that the various invocations of `create_array` are consistent. For reference, we have 4 ways to call `create_array`:
- `zarr.core.array.create_array` (the actual function that does stuff)
- `zarr.api.synchronous.create_array` (synchronous wrapper around the async function)
- `zarr.core.group.AsyncGroup.create_array` (method on `AsyncGroup` class that invokes `create_array`)
- `zarr.core.group.Group.create_array` (synchronous wrapper around the AsyncGroup method)

All of these functions should have _consistent_ signatures, but in main they don't, and a big part of this is missing tests. So this PR adds some tests that check that certain pairs of functions have identical parameters (we can't force the return types to match, because the async functions will return coroutines). To make the tests pass, this PR also ensures that all of the invocations of `create_array` have parameters that are consistent.

I say "consistent" because, when invoking `Group.create_array`, that method does not take a `store` argument, because we have one already from the group instance. Also, `Group.create_array` was recently given an extra keyword argument (`compressor`) that we need to deprecate. So the test that compares `zarr.core.array.create_array` with `zarr.core.group.AsyncGroup.create_array` only checks that all of `create_array` parameters are present in the group method. 

closes #2810

### DIFF
--- a/src/zarr/core/group.py
+++ b/src/zarr/core/group.py
@@ -20,6 +20,7 @@ from zarr.abc.metadata import Metadata
 from zarr.abc.store import Store, set_or_delete
 from zarr.core._info import GroupInfo
 from zarr.core.array import (
+    DEFAULT_FILL_VALUE,
     Array,
     AsyncArray,
     CompressorLike,
@@ -1009,7 +1010,7 @@ class AsyncGroup:
         compressors: CompressorsLike = "auto",
         compressor: CompressorLike = "auto",
         serializer: SerializerLike = "auto",
-        fill_value: Any | None = None,
+        fill_value: Any | None = DEFAULT_FILL_VALUE,
         order: MemoryOrder | None = None,
         attributes: dict[str, JSON] | None = None,
         chunk_key_encoding: ChunkKeyEncodingLike | None = None,
@@ -2430,7 +2431,7 @@ class Group(SyncMixin):
         compressors: CompressorsLike = "auto",
         compressor: CompressorLike = "auto",
         serializer: SerializerLike = "auto",
-        fill_value: Any | None = None,
+        fill_value: Any | None = DEFAULT_FILL_VALUE,
         order: MemoryOrder | None = None,
         attributes: dict[str, JSON] | None = None,
         chunk_key_encoding: ChunkKeyEncodingLike | None = None,
@@ -2835,7 +2836,7 @@ class Group(SyncMixin):
         compressors: CompressorsLike = "auto",
         compressor: CompressorLike = None,
         serializer: SerializerLike = "auto",
-        fill_value: Any | None = 0,
+        fill_value: Any | None = DEFAULT_FILL_VALUE,
         order: MemoryOrder | None = "C",
         attributes: dict[str, JSON] | None = None,
         chunk_key_encoding: ChunkKeyEncodingLike | None = None,
--- a/src/zarr/api/asynchronous.py
+++ b/src/zarr/api/asynchronous.py
@@ -11,6 +11,7 @@ from typing_extensions import deprecated
 
 from zarr.abc.store import Store
 from zarr.core.array import (
+    DEFAULT_FILL_VALUE,
     Array,
     AsyncArray,
     CompressorLike,
@@ -863,7 +864,7 @@ async def create(
     chunks: ChunkCoords | int | bool | None = None,
     dtype: ZDTypeLike | None = None,
     compressor: CompressorLike = "auto",
-    fill_value: Any | None = None,  # TODO: need type
+    fill_value: Any | None = DEFAULT_FILL_VALUE,
     order: MemoryOrder | None = None,
     store: str | StoreLike | None = None,
     synchronizer: Any | None = None,


LDeakin:
> worth noting that this changes some defaults (we were using a default fill_value of `None` in some places, and 0 in other places, for example). When in doubt, I picked the default that seemed more default.

`fill_value = None` seems a more sensible default to me, with some docs somewhere explaining how this maps to a default fill value for each data type. `0` needs conversion for any non-integer data type, and this is exactly what caused https://github.com/zarr-developers/zarr-python/issues/2792#issuecomment-2644362122 in `zarr-python` 2.

d-v-b:
> > worth noting that this changes some defaults (we were using a default fill_value of `None` in some places, and 0 in other places, for example). When in doubt, I picked the default that seemed more default.
> 
> `fill_value = None` seems a more sensible default to me, with some docs somewhere explaining how this maps to a default fill value for each data type. `0` needs conversion for any non-integer data type, and this is exactly what caused [#2792 (comment)](https://github.com/zarr-developers/zarr-python/issues/2792#issuecomment-2644362122) in `zarr-python` 2.

That works for me. In `main` I think we have a mix of 0 and `None` floating around, I think I chose 0 for this PR because it seemed like the majority, but I'm happy to switch to `None` here

d-v-b:
with https://github.com/zarr-developers/zarr-python/pull/2819/commits/642272dcaf0ae313034046a494225ae4a87d6d31 the default fill value is `None` for the array creation funcs.

src/zarr/core/group.py
```diff
@@ -987,22 +987,24 @@ async def create_array(
...
+        dtype: npt.DTypeLike | None = None,
+        data: np.ndarray[Any, np.dtype[Any]] | None = None,
         chunks: ChunkCoords | Literal["auto"] = "auto",
         shards: ShardsLike | None = None,
         filters: FiltersLike = "auto",
         compressors: CompressorsLike = "auto",
         compressor: CompressorLike = "auto",
         serializer: SerializerLike = "auto",
-        fill_value: Any | None = 0,
+        fill_value: Any | None = None,
```
dstansby, line 999: Does this mean the default fill value is no longer `0`?

changes/2819.chore.rst
```diff
@@ -0,0 +1 @@
+Ensure that invocations of ``create_array`` use consistent keyword arguments.
```
dstansby, line 1: I think it would be helpful for users to add more concretely what has changed, ie which functions have gained which arguments.

dstansby (commented):
👍 for this in general - I left one request for a more verbose changelog entry, and one question.

src/zarr/core/group.py
d-v-b: in `main` the default fill value is not consistent. in some places it's 0 (e.g., here), and in [other places](https://github.com/zarr-developers/zarr-python/blob/64b9a370a94c8d8afcfd06f93ca25d77e668cdd6/src/zarr/core/array.py#L4019) it's `None`. This PR makes it consistently `None`.

changes/2819.chore.rst
d-v-b: https://github.com/zarr-developers/zarr-python/pull/2819/commits/3fa6526511200c25256c0c9abc5afac95803ff47

src/zarr/core/group.py
dstansby: I think worth shouting loudly that this has changed (for certain functions) in a changelog entry then, if it changes the default behaviour.

d-v-b: let me know what you think of the language I added to the changelog in my recent commit

changes/2819.chore.rst
```diff
@@ -0,0 +1,4 @@
+Ensure that invocations of ``create_array`` use consistent keyword arguments, with consistent defaults.
+Specifically, ``zarr.api.synchronous.create_array`` now takes a ``write_data`` keyword argument; The
+``create_array`` method on ``zarr.Group`` takes ``data`` and ``write_data`` keyword arguments. The ``fill_value``
+keyword argument of the various invocations of ``create_array`` has been consistently set to ``None``, where previously it was either ``None`` or ``0``.
```
dstansby, line 4:
```suggestion
Ensure that invocations of ``create_array`` use consistent keyword arguments, with consistent defaults.
Specifically, ``zarr.api.synchronous.create_array`` now takes a ``write_data`` keyword argument; The
``create_array`` method on ``zarr.Group`` takes ``data`` and ``write_data`` keyword arguments.
The ``fill_value`` keyword argument of the various invocations of ``create_array`` has been consistently set to ``None``, where previously it was either ``None`` or ``0``.
In practice this does not change the default fill value assigned to arrays when the fill value is not manually set.
```

dstansby (approved):
Thanks for the changelog update, looks great to me. I left one more suggestion - I did a bit of testing and this doesn't seem to change the fill value that's set on the arrays (which is good, but I think worth reassuring users)

d-v-b:
> Thanks for the changelog update, looks great to me. I left one more suggestion - I did a bit of testing and this doesn't seem to change the fill value that's set on the arrays (which is good, but I think worth reassuring users)

how did you check this? because for zarr v2 data, it _should_ make a difference -- `None` is a valid v2 fill value, and it's not the same as 0

dstansby:
I didn't check for v2 🙈 . If it does make a difference, that should be made very clear in the changelog (and we should think about putting this PR in a non-bugfix release since it's a breaking change?)

d-v-b:
@LDeakin that was my impression as well. I think for v2 data the contents of the array metadata will be sensitive to the `None` / 0 distinction, but at runtime the actual array data should be the same in zarr-python. but this might vary in other implementations.

d-v-b:
The nullable fill value is also clunky because zarr v2 supports the python `Object` dtype, and `None` is a valid instance of the `Object` dtype, which makes it impossible to distinguish "no fill value" from "the fill value is the literal `None`" for arrays with that dtype (or any other dtype that admits `None` as a value). 

We would have to get a measure of the impact on users, but if it doesn't inconvenience a lot of people then I would definitely support dropping the creation of arrays with a `null` fill value.

tests/test_metadata/test_consolidated.py
```diff
@@ -521,7 +521,7 @@ async def test_consolidated_metadata_v2(self):
                         dtype=dtype,
                         attributes={"key": "a"},
                         chunks=(1,),
-                        fill_value=0,
+                        fill_value=None,
```
dstansby, line 524: So to clarify, this change is a breaking one for writing v2 data? Is this something we want to avoid breaking?

d-v-b: it's change to how v2 metadata is written. the actual array data will be written as before. Because the signatures of the affected functions are inconsistent, there is _no way_ to achieve the goal of this PR without changing the default behavior of some of those functions, which will result in different metadata being written to disk as an effect.

d-v-b: so to your questions: it's breaking if someone relied on the default metadata document produced by some, but not all, of our array creation routines. And no, we do not want to avoid breaking, because the cost of a confusing, inconsistent API is worth paying in this case

dstansby (changes requested):
Sorry to be a bit of a pain, but I'm going to put a request changes on this until we've resolved how we're going to bump the version number (and/or update our [versioning policy](https://zarr.readthedocs.io/en/stable/developers/contributing.html#api-compatibility)) for this change. See https://github.com/zarr-developers/zarr-python/pull/2819#discussion_r1980210543 for more context.

tests/test_metadata/test_consolidated.py
d-v-b: I think we should adjust the versioning policy, and also release the changes in this PR in a minor (non-patch) release. We were already planning on making breaking changes without incrementing the major version number -- deprecated functions like `create` were slated to be removed long before a 4.x release. So I think the versioning policy needs updating.

d-v-b:
that works for me! would you like to open an issue about the versioning policy, or should I?

dstansby:
I can 👍

d-v-b:
it turns out that the default fill value of `0` is problematic for #2874, so I'd like to elevate the priority of this PR. Given our new versioning policy, do you think this is good to go after I sort out the merge conflicts @dstansby?

dstansby:
Is changing the default value of the `fill_value` argument a breaking change? Either way that should be made clearer in the changelog entry - and if it is breaking, we need a release plan for 3.1 before merging this, so we don't accidentally end up in a place where stuff targeted for 3.1 ends up blocking bug fix releases for a long time while all the work for 3.1 is finished.

d-v-b:
ultimately I think whether these changes are breaking depends on whether you believe it's a bug that different definitions of the same function have different default values.

dstansby:
Does changing the values affect any behaviour, or any metadata that gets written to zarr arrays or groups? If not, then I'd say this isn't breaking. If it does, then I'd say it is breaking.

d-v-b:
I think there are two questions:
- Is the current situation (the same function with different default values) bugged? I believe the answer to this is "yes".
- Is it possible to fix that bug without changing behavior of zarr-python? I believe the answer to this is "no".

Because this library produces data, we will inevitably have bugs that affect how data is produced. Therefore, fixing those bugs will change how data is produced. If the bug fix takes us from "writes invalid data" to "writes valid data", that's a clear win. But if the bug fix takes us from "writes valid data" to "writes slightly different valid data", then things are less clear cut. I think we are in this second scenario. 

For simplicity lets say that these changes are breaking, and thus a good fit for a 3.1 release. it might be time to think about a 3.1 branch!

d-v-b:
> how that changes the data values returned from emtpy portions of an array with `fill_value=None` vs. `fill_value=0`.

there is no change here, zarr-python uses an array of zeros (or the zeros-equivalent, per data type) in either case.

dstansby:
I still need to wrap my head around the fill value changes a bit - in the meantime, perhaps adding the new arguments could be done in a separate PR, since that's definitely backwards compatible and probably something we could put in a 3.0.x release?

changes/2819.chore.rst
dstansby, line 1: Since "chore type" changelong entries don't get rendered (e.g., see https://zarr.readthedocs.io/en/stable/release-notes.html#misc), I'd recommend splitting this up into a "feature" entry for the updated signatures, and a separate "feature" for the change in default fill_value arguments.

dstansby (approved):
👍 had another think about this, and I think it's good as long as the changelog is fixed so it's visible. I would like someone else to review this though, just to check the default changes to the fill value are okay.

dstansby:
Pinging @zarr-developers/python-core-devs - would be good to get a second review on this

d-v-b:
the test failure here is windows-specific and fixed in https://github.com/zarr-developers/zarr-python/pull/3151

d-v-b:
these function signature tests are gold. they caught a _bunch_ of recently-added inconsistencies

changes/2819.chore.rst
d-v-b: the change in default fill value has already been done by https://github.com/zarr-developers/zarr-python/pull/3198

PR #2824, opened 2025-02-13
Add compressor, codec pipeline strategy

dcherian:
I could use input on whether we can generate `filters` and `compressors` independently of each other or whether they must be consistent with each other. Currently, I am assuming the latter and the `codecs` strategy creates both `filters` and `compressors`

TODO:


src/zarr/testing/strategies.py
```diff
@@ -86,11 +89,51 @@ def safe_unicode_for_dtype(dtype: np.dtype[np.str_]) -> st.SearchStrategy[str]:
...
+        st.builds(numcodecs.LZMA),
+        st.builds(numcodecs.Zstd),
+        st.builds(numcodecs.Zlib),
+    )
+    num_codecs_v3 = st.one_of(
+        st.builds(ncodecs.Blosc),
+        st.builds(ncodecs.LZMA),
+        # st.builds(ncodecs.PCodec),
+        # st.builds(ncodecs.ZFPY),
+    )
```
dcherian, line 127: This search space is quite large. How can we be strategic here? 

I think we should test the codec pipeline separately with many different array types. For the `arrays` strategy, we could restrict ourselves to a much more limited choice (e.g. choose one among [a codec from numcodecs, a codec from zarr codecs, or None])

PR #2827, opened 2025-02-13
WIP: Consolidate workflows for GitHub actions

maxrjones:
It might be helpful to run the upstream and min-deps jobs as part of the main 'test' workflow in order to capture caught ImportErrors in the code coverage. Motivated by #2774 where the only missed lines are ImportErrors.

TODO:

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -43,7 +43,7 @@ jobs:
             os: 'windows-latest'
           - python-version: '3.11'
             dependency-set: 'min_deps'
-          - python-version: '3.33'
+          - python-version: '3.13'
             dependency-set: 'upstream'
     runs-on: ${{ matrix.os }}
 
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -41,6 +41,10 @@ jobs:
             numpy-version: '2.1'
             dependency-set: 'optional'
             os: 'windows-latest'
+          - python-version: '3.11'
+            dependency-set: 'min_deps'
+          - python-version: '3.33'
```
dcherian, line 46:
```suggestion
          - python-version: '3.13'
```

maxrjones: yikes 🤦‍♂️

maxrjones:
I'm not sure why exactly the test workflow isn't running - it could require maintainer approval.

An alternative to this approach would be to just upload a report at the end of the `test-upstream-and-min-deps` run, as was done for the GPU tests in https://github.com/zarr-developers/zarr-python/pull/2767.

maxrjones:
Sorry for the noise, I'm going to close this because I think uploading a coverage report for `test-upstream-and-min-deps` would be a better strategy because the yaml file would be simpler even if there a more components.

PR #2839, opened 2025-02-15
Array creation with strings for filters, serializer, compressors

### LINKED ISSUE #2705: Array creation should take strings for filters, serializer, compressors
For a user who is happy to use crc32c and zstd with whatever the defaults happen to be, `create_array(..., compressors=('crc32c', 'zstd'))` is a lot more ergonomic than `create_array(..., compressors=({'name': 'crc32c'}, {'name' : 'zstd'}))`. So we should accept the iterable of strings.

brokkoli71:
- fixes #2705 

TODO:

### DIFF
--- a/src/zarr/core/array.py
+++ b/src/zarr/core/array.py
@@ -4266,9 +4266,6 @@ def _parse_chunk_encoding_v2(
     elif isinstance(compressor, tuple | list) and len(compressor) == 1:
         _compressor = parse_compressor(compressor[0])
     else:
-        if isinstance(compressor, Iterable) and not isinstance(compressor, dict):
-            msg = f"For Zarr format 2 arrays, the `compressor` must be a single codec. Got an iterable with type {type(compressor)} instead."
-            raise TypeError(msg)
         _compressor = parse_compressor(compressor)
 
     if filters is None:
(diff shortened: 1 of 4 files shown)


src/zarr/core/array.py
```diff
@@ -4264,24 +4264,13 @@ def _parse_chunk_encoding_v2(
...
-            msg = f"For Zarr format 2 arrays, the `compressor` must be a single codec. Got an iterable with type {type(compressor)} instead."
-            raise TypeError(msg)
         _compressor = parse_compressor(compressor)
 
     if filters is None:
         _filters = None
     elif filters == "auto":
         _filters = default_filters
     else:
-        if isinstance(filters, Iterable):
```
brokkoli71, line 4277: This is also checked in ``parse_filters'', so it could be removed

src/zarr/core/array.py
```diff
@@ -4264,24 +4264,13 @@ def _parse_chunk_encoding_v2(
     elif isinstance(compressor, tuple | list) and len(compressor) == 1:
         _compressor = parse_compressor(compressor[0])
     else:
-        if isinstance(compressor, Iterable) and not isinstance(compressor, dict):
```
brokkoli71, line 4267: This is also checked in ``parse_compressor``, so it could be removed

tests/test_array.py
```diff
@@ -1146,6 +1171,50 @@ async def test_v2_chunk_encoding(
...
+        msg = "Delta.__init__() missing 1 required positional argument: 'dtype'"
+        with pytest.raises(TypeError, match=re.escape(msg)):
+            await create_array(
+                store=store,
+                dtype="uint8",
+                shape=(10,),
+                zarr_format=2,
+                filters="delta",
+            )
+        msg = "TransposeCodec.__init__() missing 1 required keyword-only argument: 'order'"
```
d-v-b, line 1208: i don't think this kind of error message is the best outcome here. it would be better to have an error like "you provided the string foo, which specifies codec fooCodec, but that codec cannot be specified by a string because it takes a required configuration. use either the dict form of fooCodec, or pass in a concrete fooCodec instance instead"

d-v-b:
i think this is a good idea, and for codecs that have simple defaults (blosc and gzip and most conventional compressors), it's very user-friendly. but IMO we need to give users a very clear error when they use a string name for a codec that takes required parameters.

brokkoli71:
@d-v-b what do you think of the updated error messages?

normanrz (approved):

Issue #2842, opened 2025-02-16
Only use `None` for "no filters"


d-v-b:
In v2, we allow `filters` to be an empty array. But the v2 spec states that `filters` should be an array with at least one element, or `None`. So we should fix our implementation, and no longer generate metadata with an empty array of filters. For backwards compatibility, we should read the empty array of filters, and probably emit a warning.

leuchthelp:
https://github.com/zarr-developers/zarr-python/discussions/2841#discussion-7972692

Linking the discussion for context. Can also show up when creating a v3 zarr file, but I do not know if it might be intended for zarr v3.

d-v-b:
> [#2841 (comment)](https://github.com/zarr-developers/zarr-python/discussions/2841#discussion-7972692)
> 
> Linking the discussion for context. Can also show up when creating a `format=3` zarr file, but I do not know if it might be intended for zarr `format=3`.

in the case of zarr v3, "filters" is a shorthand name for a type of codec that transforms arrays into arrays, and we do want to represent 0 of them as an empty tuple, so I don't think there's a problem there.

PR #2847, opened 2025-02-18
don't serialize empty tuples for v2 filters, and warn when reading such metadata

d-v-b:
We were violating the v2 spec by serializing "no filters" as the empty array. We created arrays in the wild with this metadata, so any fix should be backwards compatible.

To fix this bug, this PR changes the filters parsing function so that it no longer returns empty tuples, and instead return `None` when no filters are specified. For backwards compatibility with datasets created with earlier versions of zarr-python 3.x, the `from_dict` method will replace empty lists / tuples with the value `None` and emit a warning. 

After a reasonable amount of time (a few minor releases?) we should remove this special-casing and treat `filters = []` as an error. Happy to chat about the timeline.

closes https://github.com/zarr-developers/zarr-python/issues/2842

TODO:

### DIFF
--- a/tests/test_properties.py
+++ b/tests/test_properties.py
@@ -113,6 +113,13 @@ async def test_roundtrip_array_metadata(
         assert actual == expected
 
 
+@given(store=stores, meta=array_metadata())  # type: ignore[misc]
+def test_array_metadata_meets_spec(store: Store, meta: ArrayV2Metadata | ArrayV3Metadata) -> None:
+    asdict = meta.to_dict()
+    if isinstance(meta, ArrayV2Metadata):
+        assert asdict["filters"] != ()
+
+
 # @st.composite
 # def advanced_indices(draw, *, shape):
 #     basic_idxr = draw(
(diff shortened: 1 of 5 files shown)


d-v-b:
i don't think the coverage report is accurate

dcherian (approved):
_seems_ OK to me.

tests/test_properties.py
```diff
@@ -113,6 +113,13 @@ async def test_roundtrip_array_metadata(
         assert actual == expected
 
 
+@given(store=stores, meta=array_metadata())  # type: ignore[misc]
+def test_array_metadata_meets_spec(store: Store, meta: ArrayV2Metadata | ArrayV3Metadata) -> None:
```
d-v-b, line 117: since this test does nothing for v3 metadata, would it be possible to only run it on `ArrayV2Metadata`?

d-v-b:
i will look over this again tomorrow. I think we should definitely try to fix the underlying bug this week, e.g. via this PR

tests/test_properties.py
dcherian: here I made it do something ;) . really we should fill it out.

d-v-b: i was more thinking that it would make sense to test the v2 and v3 specs in separate tests, since they are logically independent, but we can do that refactor later

PR #2856, opened 2025-02-21
Open zip store from path with zip suffix

### LINKED ISSUE #2831: Can't conveniently open zip store from path with zarr v3
### Zarr version

v3

### Numcodecs version

v0.15.1

### Python Version

3.12

### Operating System

Linux

### Installation

Mamba

### Description

In zarr v2, passing a path (str) ending with `*.zip` to `zarr.open` or `zarr.open_consolidated` would work transparently, the same way as is a `*.zarr` path was given.

In zarr v3, this fails with `FileNotFoundError : Unable to find group`. The way around is to open a `ZipStore` explicitely and then pass that to `zarr.open`.

The `zarr.open` docstring says:
https://github.com/zarr-developers/zarr-python/blob/99621ecf0b81400e323828111363fe21cf0c7592/src/zarr/api/synchronous.py#L166-L169
which seems to imply that passing a string path ending in `zip` should work.

Moreover, I found it more convenient when one didn't need to test for file extensions and explicitly handle storage objects. I use zarr dataset through xarray and it seems to me that `xr.open_dataset('example.zarr.zip', engine='zarr')` should usually be how a normal user should open a such a file ?

I understand that the ZipStore is still « experimental » in v3, and I really hope you keep it in the officiel scheme because it is very useful, to us at least.  Zipped zarrs have many of the benefits of zarr (over netCDF for example), but without the inode-explosion that pure zarr folders create on unix filesystems (slowing down the disk operations).

I think I see that the store guessing happens in `zarr.storage._common.make_store_path` ? 
Like it could happen here: 
https://github.com/zarr-developers/zarr-python/blob/99621ecf0b81400e323828111363fe21cf0c7592/src/zarr/storage/_common.py#L298-L309

Would this convenience be welcomed back in zarr-python ? I could do a PR if the team here agrees with adding this case handling.

Otherwise, I guess this could be done by `xarray` itself ? Many of my scripts go through `intake-esm` also, I guess we could fix it there too if the proposal gets refused here.

### Steps to reproduce

Example adapted from the doc.
```python3
import numpy as np
import zarr

store = zarr.storage.ZipStore("example-3.zip", mode='w')
z = zarr.create_array(
    store=store,
    shape=(100, 100),
    chunks=(10, 10),
    dtype="f4"
)
z[:, :] = np.random.random((100, 100))
store.close()

zarr.open('example-3.zip', mode='r')
```

### Additional output

_No response_

aulemahal:
- Opening a string or Path path is redirected to `ZipStore` if the path has a `.zip` suffix.

Fixes #2831 

The added test is not as elegant as the other from the suite. Is there a way to get an _existing_ zipped zarr dataset within the test suite ? Should I create another fixture instead of writing my own zarr in the test ?


aulemahal:
I did not add anything in the `user-guide/*.rst` files as I didn't feel it was necessary. The docstring of `open(` already makes mention of this feature.

src/zarr/storage/_common.py
```diff
@@ -295,6 +297,8 @@ async def make_store_path(
             store = store_like
         elif store_like is None:
             store = await MemoryStore.open(read_only=_read_only)
+        elif isinstance(store_like, Path) and store_like.suffix == ".zip":
+            store = await ZipStore.open(path=store_like, mode=mode or "r")
```
dstansby, line 301: Instead of putting the `mode or "r"` check here, and having the same check below, I think it's better to do

```python
if mode is None:
    mode = "r"
```
above the `_read_only = mode == "r"` line. This way the default mode is always "r" if `None` is passed.

dstansby (changes requested):
Looks good! Just one minor comment, but I think worth fixing 👍

aulemahal:
Hi @dstansby , I'm not sure, but it seems your suggested change that worked in march now breaks the tests. Did some assumptions change about `mode=None` ?

aulemahal:
Hi @dstansby , I updated the PR with the current main. 

#### To justify coming back to this PR after almost one year:
In between the moment I opened this PR and now, I saw there was the objective of ZEP8 which would have been another way of implementing this convenience. However, this project seems to have been stalled for the last 6 months.

My organisation has a lot of zipped zarrs in its database. We are running a lot of code on a traditionnal HPC which has a relatively low limit on inodes. Zipped zarr offer much better performance than netCDFs and a single node. Local zarrs are more performant, but yield insane amounts of inodes. Xarray not being able to read zipped zarr directly is an issue that blocks us from upgrading to Zarr 3. 

We have a few solutions:
- Having zarr do it natively : the easiest one, this PR
- A custom xarray backend + changes in `intake-esm` : more effort, less elegant, needs development in a low-activity repo.
- Monkeypatch zarr : extra ugly, much less nice than contributing to open source software
- Flip to netCDF / unzipped zarr : lots of work to rewrite everything, great loss in performance on the HPC

I would understand if my patch here is not the most elegant way forward in the grand scheme of Zarr. But, I still think this could be useful for more than us. It could be seen as a "temporary solution" before ZEP8 is implemented fully.

d-v-b:
> Zipped zarr offer much better performance than netCDFs and a single node. Local zarrs are more performant, but yield insane amounts of inodes. 

I will look at this PR but in the meantime have you considered sharding? that gives you most of the benefits of zipping your data

aulemahal:
Thanks!

Yes, we did, but not that much yet, we were a bit slow to begin to look into new Zarr v3 features. For a typical gridded climate datasets with a single data variable, you still have at least 3 inodes per variable (folder, data and json), which can easily yield 20x more inodes considering you have a few auxiliary coords, even if writing the data in a single shard (not sure yet if this is good or not for multi-GB dataseets). So sharding could be a compromise indeed, but this is still a lot of inodes.

ospinajulian:
Hi, is there any update on this PR?

aulemahal:
Hi @d-v-b , I updated the PR with the main branch.

I read the release blog post and I see big plans are coming with the stackification of zarr, that's great! I still think this PR could act as small temporary feature before one of the new stack element can tackle the "auto-open" issue with a more complete approach (like ZEP8, if still in the plans).

However, if you don't think this is the case and that this PR is not useful, please tell me and I'll redirect my efforts for fixing our issue to another solution. Thanks!

Issue #2861, opened 2025-02-25
Missmatch of Array.store_path.store.list_prefix() and Array._iter_chunk_keys() in chunks_initialized() for an array in a group


relativityhd:
### Zarr version

v3.0.2

### Numcodecs version

v0.15.1

### Python Version

3.11

### Operating System

Linux

### Installation

uv add zarr

### Description

Currently, it is not possible to view how many chunks are initialized in an array inside a group.

I was able to track it down to a mismatch of the outputs of the functions `z.store_path.store.list_prefix(prefix=z.store_path.path)` and `z._iter_chunk_keys()`.

### Steps to reproduce

```py
import zarr

# Create a group with two arrays
root = zarr.group("reproduction.zarr")
root.create("x", shape=(1000, 1000), chunks=(100, 100), dtype="f4")
root.create("y", shape=(1000, 1000), chunks=(100, 100), dtype="uint8")
root.tree()
```

```py
from zarr.core.array import chunks_initialized

# Now check which chunks are initialized (should be an empty tuple)
await chunks_initialized(root["x"])
> ()
```

```py
# Now fill two chunks
root["x"][0:100, 0:100] = 1.0
root["x"][100:200, 0:100] = 2.0

# Now check which chunks are initialized (should be a tuple with two chunks)
await chunks_initialized(root["x"])
> ()  # <-- This shouldn't be empty
```

```py
# Further investigate what happens in "chunks_initialized()"
store_contents = [x async for x in root["x"].store_path.store.list_prefix(prefix=root["x"].store_path.path)]
store_contents
> ['x/zarr.json', 'x/c/1/0', 'x/c/0/0']
```

```py
tuple(chunk_key for chunk_key in root["x"]._iter_chunk_keys())

> ('c/0/0',
 'c/0/1',
 'c/0/2',
 'c/0/3',
 'c/0/4',
 'c/0/5',
 'c/0/6',
 'c/0/7',
 'c/0/8',
 'c/0/9',
 'c/1/0',
 'c/1/1',
 'c/1/2',
 'c/1/3',
 'c/1/4',
...
 'c/9/5',
 'c/9/6',
 'c/9/7',
 'c/9/8',
 'c/9/9')
```

d-v-b:
good report, and sorry for the bug. seems like we need to relativize the output of `list_prefix`,  or absolutifiy the values in `iter_chunk_keys`. Since this is so obviously wrong that I'm wondering how our tests did not catch it.

[...]

PR #2862, opened 2025-02-25
fix/unbreak chunks initialized

d-v-b:
fixes https://github.com/zarr-developers/zarr-python/issues/2861 by making extant prefixes relative before comparing them to the list of chunk names.

some odd noise in the commit log here, but these are all commits that landed in main so I don't think it's a problem.


src/zarr/storage/_utils.py
```diff
@@ -74,11 +74,60 @@ def _join_paths(paths: Iterable[str]) -> str:
...
+    'a/b'
+    >>> _join_paths(["a", "b", "c"])
+    'a/b/c'
     """
     return "/".join(filter(lambda v: v != "", paths))
 
 
+def _relativize_path(path: str, prefix: str) -> str:
+    """
+    Make a "\"-delimited path relative to some prefix. If the prefix is '', then the path is
```
TomAugspurger, line 97: Is `"\"-delimited` right here, or should this be `"/"-delimited`?

src/zarr/storage/_utils.py
```diff
@@ -74,11 +74,60 @@ def _join_paths(paths: Iterable[str]) -> str:
...
+    returned as-is. Otherwise, the prefix is removed from the path as well as the separator
+    string "\".
+
+    If ``prefix`` is not the empty string and``path`` does not start with ``prefix``
```
TomAugspurger, line 101:
```suggestion
    If ``prefix`` is not the empty string and ``path`` does not start with ``prefix``
```

TomAugspurger: And I guess since every string technically starts with the empty string, you can rephrase this as just "If path does not start with prefix ..."

src/zarr/storage/_utils.py
```diff
@@ -74,11 +74,60 @@ def _join_paths(paths: Iterable[str]) -> str:
...
+    --------
+    >>> _relativize_paths("", "a/b")
+    'a/b'
+    >>> _relativize_paths("a/b", "a/b/c")
+    'c'
+    """
+    if prefix == "":
+        return path
+    else:
+        _prefix = prefix + "/"
```
TomAugspurger, line 125: What behavior do we want for `prefix="<some-prefix>/"`, i.e. a prefix ending with a `/`?

IIUC, this will currently always raise a `ValueError`, unless `path` for some reason starts with `<some-prefix>//`

Maybe we should have `_prefix = prefix.rstrip("/") + "/"`? Or if you're confident the callers are careful about not providing a prefix with a trailing slash this is fine.

TomAugspurger (approved):
Thanks for handling this. A couple small questions, but I think this looks good.

src/zarr/storage/_utils.py
d-v-b: I think I was assuming here that paths have already been normalized by [`normalize_path`](https://github.com/zarr-developers/zarr-python/blob/5ff3fbe5fe1488310301e9d2ae56a9880d1ddfb2/src/zarr/storage/_utils.py#L16), in which case prefixes cannot end with `"/"`. I should make this explicit in the docstring for the function here.

TomAugspurger (approved):

TomAugspurger:
Sounds good, thanks. It looks like the codecov failure might be legitimate, so good to go once that's fixed.

Issue #2658, opened 2025-01-06
Device support in `zarr-python` (especially for GPU)


nenb:
**Problem**
I would like to load `zarr` data directly onto non-CPU devices (especially GPU). The current approach appears to rely on using `cupy` to load onto `cupy`-supported devices e.g. https://github.com/rapidsai/kvikio/blob/branch-25.02/notebooks/zarr.ipynb.

Unfortunately, there are a number of devices that are not supported by `cupy` e.g. I don't believe that my Apple Metal GPU is supported. This means that I must load from `zarr` via CPU if I would like to use these devices e.g. `zarr` on disk -> `numpy` -> `torch` (which has Metal support).

This is slow(er) and I don't believe is necessary from the `zarr` specification alone (?).

**Background**
Multi-device support is a very important requirement in the AI/ML community. I would like to use `zarr` (and specifically the Python implementation) to run models such as LLMs on multiple devices. The quicker it is to load the model onto device (and with reduced memory usage etc), the better the UX and developer experience is.

**Questions**
1. Is `cupy` the correct/only way to load direct to GPU with `zarr-python`?
2. Is there/will there be any way of loading direct to devices such as Metal with `zarr-python`?
3. (Related) What is the best way to load a PyTorch neural network on GPU with `zarr-python`? Is it `cupy` and then using something like [dlpack](url) for zero-copy exchange? Are there alternatives?

**Related issues**
https://github.com/zarr-developers/zarr-python/pull/1967
https://github.com/zarr-developers/zarr-python/issues/2574

cc @jhamman (as suggested by @TomNicholas)

ziw-liu:
CuPy/kvikio relies on nvidia's GPUDirect storage (GDS) driver and goes through PCIe. Metal GPUs are using unified memory, so CPU-to-GPU transfer can _in theory_ be almost zero-cost (passing an address). If there is a way to pass the ownership of an array from CPU to GPU, nothing needs to be done in zarr unless there is need for GPU-accelerated decompression.

In practice though, at least torch implements the `to("mps")` method by cloning the tensor (memcpy-ish cost), and each ML framework may do different things. Another reference point is jax, which implements [(experimental) serialization](https://github.com/jax-ml/jax/blob/55119490ebe9810dd6faace0a27ef0fbdf0e25fd/jax/experimental/array_serialization/serialization.py) to zarr using tensorstore.

nenb:
[...]

PR #2863, opened 2025-02-25
Zstd Codec on the GPU

akshaysubr:
This PR adds a Zstd codec that runs on the GPU using the nvCOMP 4.2 python APIs.

TODO:
* [ ] Make fully async
* [ ] Performance benchmarking
* [ ] CPU-GPU roundtrip testing


src/zarr/codecs/gpu.py
```diff
@@ -0,0 +1,195 @@
...
+except ImportError:
+    cp = None
+
+try:
+    from nvidia import nvcomp
+except ImportError:
+    nvcomp = None
+
+
+class AsyncCUDAEvent(Awaitable[None]):
```
TomAugspurger, line 33: A couple comments on this, based on some brief reading and conversation with @akshaysubr.

Currently, the call to `self._zstd_codec.encode` in `NvcompZstdCodec.encode` blocks. Since this is an aysnc function we'll want to avoid that blocking call, but it's fine for now for experimentation / testing.

Our easiest way of making that non-blocking is with using `asyncio.to_thread` somewhere. Exactly what's done in `to_thread` or before it depends on how much we want run on the main event loop vs. the worker thread.

I'm not sure what the overhead of using `to_thread` is (a bit of overhead from creating a `Future`, a bit of overhead from the OS having to switch between the threads). It's probably minor, assuming that `nvcom.Codec.encode` doesn't hold onto the GIL. My understanding is that that (host) thread itself is just going to be waiting around for the decompression to finish on the device.

I briefly looked into whether using [selectors](https://docs.python.org/3/library/selectors.html) might be an option to bridge the CUDA Event completion to Python aysncio, without the need for a separate host worker thread. But `select` is all based around file descriptors. AFAIK, there isn't a way for the OS to monitor the completion of some CUDA event. We could perhaps hack something together, but I'm pessimistic that it would actually be faster than `to_thread`.

dstansby:
Thanks for opening this PR! At the moment we do not have any codecs implemented in the `zarr-python` package, but instead store them in [`numcodecs`](https://github.com/zarr-developers/numcodecs). So although it looks like there are some zarr-python specific changes that are needed to support the new GPU codec, the actual codec should be implemented in `numcodecs`, and then imported in `zarr-python`.

akshaysubr:
@dstansby My understanding was that `numcodecs` is the place that python bindings to native codec implementations live and that with v3, the `Codec` class itself lives in `zarr-python`. The GPU codecs and python bindings are implemented in nvCOMP and imported through the `nvidia-nvcomp-cu12` python package so I'm not sure which part of this would need to go in `numcodecs`. What did you have in mind?

src/zarr/codecs/gpu.py
```diff
@@ -0,0 +1,195 @@
...
+        Returns
+        -------
+        Iterable[Buffer | None]
+        """
+        chunks_and_specs = list(chunks_and_specs)
+
+        # Convert to nvcomp arrays
+        filtered_inputs, none_indices = await self._convert_to_nvcomp_arrays(chunks_and_specs)
+
+        outputs = self._zstd_codec.decode(filtered_inputs) if len(filtered_inputs) > 0 else []
```
TomAugspurger, line 157: Question related to https://github.com/zarr-developers/zarr-python/issues/2904 (which is looking into memory usage). Would it be possible for nvcomp-python to provide an `out` argument to `decode`? If I'm reading the [C++ docs](https://docs.nvidia.com/cuda/nvcomp/cpp_api.html#_CPPv4N6nvcomp17nvcompManagerBase10decompressEP7uint8_tPK7uint8_tRK19DecompressionConfigP6size_t) correctly, that does seem to decompress into an output buffer. Eventually it would be nice to do that all the way into Zarr's `out` buffer.

TomAugspurger: It's not currently supported, but I've seen similar requests from elsewhere. Hopefully someday.

src/zarr/codecs/gpu.py
```diff
@@ -0,0 +1,195 @@
...
+            Ordered set of encoded chunks with their accompanying chunk spec.
+
+        Returns
+        -------
+        Iterable[Buffer | None]
+        """
+        chunks_and_specs = list(chunks_and_specs)
+
+        # Convert to nvcomp arrays
+        filtered_inputs, none_indices = await self._convert_to_nvcomp_arrays(chunks_and_specs)
```
TomAugspurger: This should just be reinterpreting some bytes / pointers, right? There's no possibility of doing any kind of (blocking) I/O? If so, then I'd recommend making this a regular `sync` method.

TomAugspurger: Same comment on L161 with the `return await self._convert_from_nvcomp_arrays`.

src/zarr/codecs/gpu.py
TomAugspurger, line 133: This is the line where we should be careful about what happens where.

Do you know whether `nvcomp.Codec.decode` is blocking, or does it just asynchronously schedule the decode on the GPU? If it's non-blocking, then for simplicity, I'd recommend using `asyncio.to_thread` around a (regular sync Python) function that schedules the decode and then uses the `event.wait()` to block to do the stream synchronization.

```python
async def decode_wrapper(codec, filtered_inputs):
    result = codec.decode(filtered_inputs)
    # wait for the decode to complete. cupy.cuda.Event?
    return result
```

I have some longer-term thoughts around how zarr-python handles concurrency for different types of workloads, but for now I think it's probably best to follow the other codecs, which use `to_thread`.

src/zarr/codecs/gpu.py
TomAugspurger: Also in `encode`.

src/zarr/codecs/gpu.py
TomAugspurger: More specifically, I thinking something like this on the class:

```python
    async def _synchronize_stream(self) -> None:
        # this is the blocking operation. Offload it to a worker thread to not block the main thread
        await asyncio.to_thread(self.stream.synchronize)
```

and then in decode:

```python
        ...
        decoded = self._convert_from_nvcomp_arrays(outputs, chunks_and_specs)
        # Uphold zarr-python's guarantee that the decode is finished before returning
        await self._synchronize_stream()
        return decoded
```

That way, the main thread will schedule everything to happen on the GPU via `self._zstd_codec.decode()`, but the actual stream synchronization will happen on another thread, so that it doesn't block the event loop.

That requires putting a `stream` (and maybe `Device`?) object on the `ZstdCodec` class so that we can make sure we synchronize the right stream (the same one passed to `nvcomp.Codec`)

akshaysubr: The way that synchronization is handled in nvcomp is through a first touch policy of any attribute of an nvcomp output array. The work is scheduled in a non-blocking fashion but attempting to query an attribute of an output will trigger a sync. So in the example above, `_convert_from_nvcomp_arrays` will implicitly issue the synchronization. I think here, we don't want to rely on nvcomp's sync mechanism but add our own sync mechanism that renders nvcomp's eventual sync a no-op right?

TomAugspurger: > I think here, we don't want to rely on nvcomp's sync mechanism but add our own sync mechanism that renders nvcomp's eventual sync a no-op right?

That sounds about right... The main thing we want to avoid is anything on the main thread blocking the event loop. It sounds like calling `nvidia.nvcomp.Codec.encode` and `decode` are themselves non-blocking and so probably fine to do from the main thread. But everything after that should probably be done off the main thread, or at least after another thread has done a synchronization (I think doing things like `cp.from_dlpack` / viewing reinterpreting bytes on the main thread is considered fine).

akshaysubr: Instead of using `asyncio.to_thread`, I ended up implementing a custom awaitable object that checks on a CUDA event and uses `asyncio.sleep()` to relinquish cycles back to the event loop. Just pushed those changes.

TomAugspurger: Here's my understanding of the context and what we want, and then maybe we can discuss pros / cons of different implementations:

We're inside an `async def decode` with a batch of `Buffer | None`s to decode. Because we're inside an `async def` function, we *cannot* do any kind of blocking operation like `event.wait()` or stream  / device synchronize, since that would block the event loop. The main Python event loop thread used by Zarr needs to progress other reads, decodes, and respond to additional input from the user.

AFAIK, the three ways we can try to address this type of issue are

1. Offloading the work to a separate thread in the asyncio event loops Executor pool.
2. Writing some kind polling wrapper that makes a non-blocking check for completion and uses `asyncio.sleep` to not block while giving other coroutines a chance to run
3. Set up some kind of callback that fires when (this would essentially mimic what `asyncio.to_thread` does via https://github.com/python/cpython/blob/da79ac9d26860db62500762c95b7ae534638f9a7/Lib/asyncio/futures.py#L368). This relies on the blocking API providing some kind of callback, which I think we don't have here.

Option 1 is inefficient (because we'll have a bunch of threads just doing `event.wait()`; everything is happening on the PGU), but simple and consistent with the rest of zarr.

Option 2 feels a bit complicated to me, but maybe just because it's new. And I think it'll be relatively slow (there's always *some* delay between when the event finishes and when the poller notices that it's finished, thanks to the `sleep`) and *might* put some stress on the event loop (if we have *many* futures being polled; I suspect we won't hit this since we have one per *batch*, not chunk, but still...)

Option 3 is intriguing. It seems to be a thing that CUDA supports through [`cuda.bindings.driver.cuLaunchHostFunc`](https://nvidia.github.io/cuda-python/cuda-bindings/latest/module/driver.html#cuda.bindings.driver.cuLaunchHostFunc) but I've never used that API before. If we can schedule that to fire a *python* callback, then it could take care of setting the result of an `asyncio.Future` and things might just work out. I'll spend a bit of time looking into this.

src/zarr/codecs/gpu.py
```diff
@@ -0,0 +1,178 @@
...
+try:
+    from nvidia import nvcomp
+except ImportError:
+    nvcomp = None
+
+
+def parse_zstd_level(data: JSON) -> int:
+    if isinstance(data, int):
+        if data >= 23:
+            raise ValueError(f"Value must be less than or equal to 22. Got {data} instead.")
```
TomAugspurger, line 35: Ideally nvcomp would do this, but if we're taking it on then we'll need unit tests for the `ValueError` here, the `TypeError` [here]https://github.com/zarr-developers/zarr-python/pull/2863/files#diff-49b79e9bc6f6a9c0f6714c1c1be4b6f4c1539b0afabbb72568b8383ead14a5f0R37) and the `TypeError` [here](https://github.com/zarr-developers/zarr-python/pull/2863/files#diff-49b79e9bc6f6a9c0f6714c1c1be4b6f4c1539b0afabbb72568b8383ead14a5f0R43).

src/zarr/codecs/gpu.py
```diff
@@ -0,0 +1,178 @@
...
+    def to_dict(self) -> dict[str, JSON]:
+        return {
+            "name": "zstd",
+            "configuration": {"level": self.level, "checksum": self.checksum},
+        }
+
+    @cached_property
+    def _zstd_codec(self) -> nvcomp.Codec:
+        # config_dict = {algorithm = "Zstd", "level": self.level, "checksum": self.checksum}
+        # return Zstd.from_config(config_dict)
```
TomAugspurger:
```suggestion
```

src/zarr/codecs/gpu.py
```diff
@@ -0,0 +1,178 @@
...
+
+@dataclass(frozen=True)
+class NvcompZstdCodec(BytesBytesCodec):
+    is_fixed_size = True
+
+    level: int = 0
+    checksum: bool = False
+
+    def __init__(self, *, level: int = 0, checksum: bool = False) -> None:
+        # TODO: Set CUDA device appropriately here and also set CUDA stream
```
TomAugspurger, line 54: Agreed with leaving devices / streams as a TODO for now.

I want to enable users to overlap host-to-device memcpys with compute operations (like decode, but their own compute operations as well), but I'm not sure yet what that API will look like.

If you have any thoughts on how best to do this I'd love to hear them, and write them up as an issue.

src/zarr/codecs/gpu.py
```diff
@@ -0,0 +1,178 @@
...
+        # TODO: add CUDA stream here
+        return nvcomp.as_arrays(filtered_inputs), none_indices
+
+    async def _convert_from_nvcomp_arrays(
+        self,
+        arrays: Iterable[nvcomp.Array],
+        chunks_and_specs: Iterable[tuple[Buffer | None, ArraySpec]],
+    ) -> Iterable[Buffer | None]:
+        return [
+            spec.prototype.buffer.from_array_like(cp.array(a, dtype=np.dtype("b"), copy=False))
```
TomAugspurger: This is one spot where @weiji14's idea to use dlpack in https://github.com/zarr-developers/zarr-python/issues/2658#issuecomment-2907038338 would help. If `NDBuffer` new how to consume objects implementing the dlpack protocol, we could (maybe) get rid of the `cp.array`

src/zarr/codecs/gpu.py
```diff
@@ -0,0 +1,178 @@
...
+        device = cp.cuda.Device()  # Select the current default device
+        stream = cp.cuda.get_current_stream()  # Use the current default stream
+        return nvcomp.Codec(
+            algorithm="Zstd",
+            bitstream_kind=nvcomp.BitstreamKind.RAW,
+            device_id=device.id,
+            cuda_stream=stream.ptr,
+        )
+
+    async def _convert_to_nvcomp_arrays(
```
TomAugspurger:
```suggestion
    def _convert_to_nvcomp_arrays(
```

src/zarr/codecs/gpu.py
```diff
@@ -0,0 +1,178 @@
...
+        self,
+        chunks_and_specs: Iterable[tuple[Buffer | None, ArraySpec]],
+    ) -> tuple[list[nvcomp.Array], list[int]]:
+        none_indices = [i for i, (b, _) in enumerate(chunks_and_specs) if b is None]
+        filtered_inputs = [b.as_array_like() for b, _ in chunks_and_specs if b is not None]
+        # TODO: add CUDA stream here
+        return nvcomp.as_arrays(filtered_inputs), none_indices
+
+    async def _convert_from_nvcomp_arrays(
```
TomAugspurger:
```suggestion
    def _convert_from_nvcomp_arrays(
```

src/zarr/codecs/gpu.py
```diff
@@ -0,0 +1,178 @@
...
+            Ordered set of encoded chunks with their accompanying chunk spec.
+
+        Returns
+        -------
+        Iterable[Buffer | None]
+        """
+        chunks_and_specs = list(chunks_and_specs)
+
+        # Convert to nvcomp arrays
+        filtered_inputs, none_indices = await self._convert_to_nvcomp_arrays(chunks_and_specs)
```
TomAugspurger:
```suggestion
        filtered_inputs, none_indices = self._convert_to_nvcomp_arrays(chunks_and_specs)
```

src/zarr/codecs/gpu.py
```diff
@@ -0,0 +1,178 @@
...
+
+        Returns
+        -------
+        Iterable[Buffer | None]
+        """
+        # TODO: Make this actually async
+        chunks_and_specs = list(chunks_and_specs)
+
+        # Convert to nvcomp arrays
+        filtered_inputs, none_indices = await self._convert_to_nvcomp_arrays(chunks_and_specs)
```
TomAugspurger:
```suggestion
        filtered_inputs, none_indices = self._convert_to_nvcomp_arrays(chunks_and_specs)
```

src/zarr/codecs/gpu.py
```diff
@@ -0,0 +1,178 @@
...
+
+        # Record event for synchronization
+        event = cp.cuda.Event()
+        # Wait for decode to complete in a separate async thread
+        await asyncio.to_thread(event.synchronize)
+
+        for index in none_indices:
+            outputs.insert(index, None)
+
+        return await self._convert_from_nvcomp_arrays(outputs, chunks_and_specs)
```
TomAugspurger:
```suggestion
        return self._convert_from_nvcomp_arrays(outputs, chunks_and_specs)
```

TomAugspurger (commented):
Looks nice overall. I think the async side of things ended up in a pretty good spot. The code itself is pretty easy to follow (a normal stream synchronize). Having to do that on another host thread is a bit unfortunate, but there's only one synchronize per batch so this should be fine.

I left comments on a few things to clean up that I can help out with if you want @akshaysubr.

src/zarr/codecs/gpu.py
TomAugspurger: Based on some conversations with teammates, trying to use `cuLaunchHostFunc` from Python might be too fragile for us. The callback needs to be some C wrapper, but we need to interact with Python objects (in order to set the result of the `Future`). This requires acquiring the GIL, which risks deadlocks with other things happening in the CUDA runtime.

It's perhaps worth exploring, but we'd need to be very careful. Using `asyncio.to_thread` for each batch is a fine workaround for now.

TomAugspurger:
Fixed some merge conflicts and changed the `_convert_{to,from}_nvcomp_arrays` methods to be sync.

src/zarr/core/buffer/gpu.py
```diff
@@ -59,7 +59,7 @@ def __init__(self, array_like: ArrayLike) -> None:
 
         if array_like.ndim != 1:
             raise ValueError("array_like: only 1-dim allowed")
-        if array_like.dtype != np.dtype("B"):
+        if array_like.dtype.itemsize != 1:
```
TomAugspurger, line 61: The new tests in `test_nvcomp.py` were failing without this change.

I'd like to get us to a point where we don't care as much about the details of the buffer passed in here. This is an OK start I think.

d-v-b: what exactly does this check for? It's not clear to me why any numpy array that can be viewed as bytes wouldn't be allowed in here

d-v-b: and same for the dimensionality check, since any N-dimensional numpy array can be viewed as a 1D array.

TomAugspurger: Yeah, I'm not really sure...

I agree that the actual data we store internally here needs to be a byte dtype. Just doing `cp.asarray(input).view("b")` seems pretty reasonable to me.

d-v-b: i'm not even convinced that we need `Buffer` / `NDBuffer`, when `Buffer` is just a special case of `NDBuffer` where there's 1 dimension and the data type is bytes

d-v-b: we could even express this formally by:
- making `NDBuffer` generic with two type parameters (number of dimensions and dtype)
- having APIs that insist on consuming a `Buffer` instead insist on consuming `NDBuffer[Literal[1], np.uint8]`

d-v-b: (super out of scope for this PR ofc)

src/zarr/codecs/gpu.py
```diff
@@ -0,0 +1,176 @@
...
+        Returns
+        -------
+        Iterable[Buffer | None]
+        """
+        chunks_and_specs = list(chunks_and_specs)
+
+        # Convert to nvcomp arrays
+        filtered_inputs, none_indices = self._convert_to_nvcomp_arrays(chunks_and_specs)
+
+        outputs = self._zstd_codec.decode(filtered_inputs) if len(filtered_inputs) > 0 else []
```
TomAugspurger, line 133: @akshaysubr do you know whether `codec.decode` is blocking or not?

https://github.com/TomAugspurger/cuda-streams-sample/issues/4 is looking into this a bit, but it seems like it might be blocking (an `nvtx.annotate` around the `codec.decode(filtered_inputs)` spans the duration of the decode).

TomAugspurger: This might have been an artifact of me immediately discarding the result of `codec.decode` and the destructor of `nvcomp.Array` (potentially?) blocking while running `cudaFreeHost`. See https://github.com/TomAugspurger/cuda-streams-sample/issues/4#issuecomment-3063822279

akshaysubr: @TomAugspurger `codec.decode` itself should be non-blocking. However, the synchronization point is implicit when either:

1. Any attribute of an output `nvcomp.Array` object is accessed or
2. The codec object itself is destroyed (since the codec internally could be managing some temporary allocations, etc.)

1 is mainly why we need to have our own async synchronization point before any other part of the pipeline touches an unavailable output and let the event loop assign other tasks during the decode. But 2 might be what you're running into?

src/zarr/core/buffer/gpu.py
akshaysubr: > It's not clear to me why any numpy array that can be viewed as bytes wouldn't be allowed in here

I think this is mainly because `NDBuffer` objects don't need to be contiguous, but `Buffer` objects must be contiguous in memory which might be important when we send those out to codecs that expect a contiguous memory slice.

But I agree that we can probably merge those two and make `Buffer` a specialization of `NDBuffer`.

TomAugspurger:
@d-v-b f16d730b9b5b13bcfec934cd26ccee0cd8a6e466 fixes an issue after https://github.com/zarr-developers/zarr-python/pull/3228 I think. Where providing an alternative zstd implementation here, so when we figure out which codec to use when creating an array, we need to look up the codec class registered for `zstd` instead of just using `zarr.codecs.ZstdCodec()`. LMK if that sounds right to you.

d-v-b:
> @d-v-b [f16d730](https://github.com/zarr-developers/zarr-python/commit/f16d730b9b5b13bcfec934cd26ccee0cd8a6e466) fixes an issue after #3228 I think. Where providing an alternative zstd implementation here, so when we figure out which codec to use when creating an array, we need to look up the codec class registered for `zstd` instead of just using `zarr.codecs.ZstdCodec()`. LMK if that sounds right to you.

This definitely seems like a good approach given the current state of our codec registry. More broadly, I'm not convinced that we have an ergonomic way of supporting multiple implementations of the same codec...

src/zarr/codecs/gpu.py
```diff
@@ -0,0 +1,176 @@
...
+
+        outputs = self._zstd_codec.encode(filtered_inputs) if len(filtered_inputs) > 0 else []
+
+        # Record event for synchronization
+        event = cp.cuda.Event()
+        # Wait for decode to complete in a separate async thread
+        await asyncio.to_thread(event.synchronize)
+
+        for index in none_indices:
+            outputs.insert(index, None)
```
TomAugspurger, line 175: This bit isn't covered by tests. I'll work on that today.

TomAugspurger (approved):
Thanks @akshaysubr, I think this is in a good state.

I'll open up a follow up issue to discuss how to handle devices and streams. I'll leave this open for a few days if anyone is interested in taking a look, and we can hopefully merge it next week.

src/zarr/codecs/gpu.py
TomAugspurger: https://github.com/zarr-developers/zarr-python/issues/3271 for planning on devices and streams.

src/zarr/codecs/gpu.py
```diff
@@ -0,0 +1,176 @@
...
+except ImportError:
+    cp = None
+
+try:
+    from nvidia import nvcomp
+except ImportError:
+    nvcomp = None
+
+
+def parse_zstd_level(data: JSON) -> int:
```
dstansby:
```suggestion
def _parse_zstd_level(data: JSON) -> int:
```
This should be private API

src/zarr/codecs/gpu.py
```diff
@@ -0,0 +1,176 @@
...
+    if isinstance(data, int):
+        if data >= 23:
+            raise ValueError(f"Value must be less than or equal to 22. Got {data} instead.")
+        return data
+    raise TypeError(f"Got value with type {type(data)}, but expected an int.")
+
+
+def parse_checksum(data: JSON) -> bool:
```
dstansby:
```suggestion
def _parse_checksum(data: JSON) -> bool:
```
Similar to above, this should be private

src/zarr/core/array.py
```diff
@@ -4680,9 +4680,9 @@ def default_compressors_v3(dtype: ZDType[Any, Any]) -> tuple[BytesBytesCodec, ..
     """
     Given a data type, return the default compressors for that data type.
 
-    This is just a tuple containing ``ZstdCodec``
+    This is just a tuple containing an instance of the default "zstd" codec class.
     """
-    return (ZstdCodec(),)
+    return (cast(BytesBytesCodec, get_codec_class("zstd")()),)
```
dstansby, line 5056: Why is the extra `cast` needed now?

src/zarr/core/buffer/gpu.py
```diff
@@ -59,7 +58,7 @@ def __init__(self, array_like: ArrayLike) -> None:
...
             raise ValueError("array_like: only byte dtype allowed")
```
dstansby:
```suggestion
            raise ValueError("array_like: only an itemsize of 1 allowed")
```

dstansby (changes requested):
Thanks for the really nice work here! I'm afraid I'm going to mark this as request changes, because we still need to work out what our policy on including new codecs is - should they be included in `numcodecs`, or should they be included directly in `zarr-python`. There's a few reason we need to work this out before adding new codecs:

1. For developers, make it clear where to contribute new codecs (I know someone interested in implementing `blosc2` at the moment for example)
2. For users, we need a clear story/documentation about where codecs live. If the answer is across *both* numcodecs and zarr-python, then we need clear cross linking documentation to make sure this is clear.

I've opened an issue at https://github.com/zarr-developers/zarr-python/issues/3272 to discuss this; once that's resolved we can either continue here or move over to numcodecs.

I left some comments inline - might be worth waiting for the above to be resolved before looking at them?

Finally, this should also get some documentation in the user guide about how to swap between using the CPU and GPU when both are an option.

thanks again for the work - it's super appreciated, and I'm sorry that merging is going to be a bit delayed until we work out what our long term plan with implementing new codecs is.

TomAugspurger:
Akshay responded on this point [in February](https://github.com/zarr-developers/zarr-python/pull/2863#issuecomment-2691011494)(!) @dstansby could you at least respond to that directly: that this is a *wrapper* around a codec engine, in exactly the same way that https://github.com/zarr-developers/zarr-python/blob/abbdbf2be70a24e7d662b5ed449c68f6718977f9/src/zarr/codecs/zstd.py is?

> Finally, this should also get some documentation in the user guide about how to swap between using the CPU and GPU when both are an option.

That's already included in `gpu.rst`.

d-v-b:
where is the source code for the nvcomp zstd implementation, and the python bindings?

TomAugspurger:
> where is the source code for the nvcomp zstd implementation, and the python bindings?

I don't believe that the source code for those are published publicly.

d-v-b:
> > where is the source code for the nvcomp zstd implementation, and the python bindings?
> 
> I don't believe that the source code for those are published publicly.

That's potentially quite problematic. We recently had [problems](https://github.com/zarr-developers/numcodecs/issues/424) relating to skew across zstd implementations. If we cannot inspect the source code for this codec, and we cannot submit patches, then I would definitely not be interested in experiencing bugs from it.

TomAugspurger:
> If we cannot inspect the source code for this codec, and we cannot submit patches, then I would definitely not be interested in experiencing bugs from it.

Is that you speaking as a user of zarr or a maintainer?

From the user perspective, this will only be used if you activate it.

From the maintainer's perspective, I'd hope that through documentation and clear error messages we can pinpoint the issue for users.

dstansby:
> Akshay responded on this point https://github.com/zarr-developers/zarr-python/pull/2863#issuecomment-2691011494(!) @dstansby could you at least respond to that directly: that this is a wrapper around a codec engine, in exactly the same way that [abbdbf2/src/zarr/codecs/zstd.py](https://github.com/zarr-developers/zarr-python/blob/abbdbf2be70a24e7d662b5ed449c68f6718977f9/src/zarr/codecs/zstd.py) is?

Sorry, I missed that. And sorry for the slow reply - I am not paid to work on zarr-python, and review stuff on a best efforts basis.

re. where stuff lives, I did not realise that we had codec classes implemented directly in `zarr-python`. Indeed, our [migration guide says](https://zarr.readthedocs.io/en/latest/user-guide/v3_migration.html#getting-ready-for-3-0) (!!)

> zarr.codecs has gone, use numcodecs instead

Given that codecs classes are in `zarr-python` already, and given the direction of conversation in https://github.com/zarr-developers/zarr-python/issues/3272, I'm happy to put this in `zarr-python`.

> That's already included in gpu.rst.

Am I right in thinking that GPU arrays and now decompression are a global option, not a per-array or a per-operation configuration? If so it would be good to clarify in gpu.rst that a) this is only a global option b) if it can be changed during a python session, and c) how decompression of GPU buffers is handled for codecs without a GPU implementation (I presume it just falls back to the CPU impelementation)

re. @d-v-b's point about about skew across zstd implementations, I think this is a very important point. I'm personally on the fence about depending on a closed source library, but as a minimum I think there should be tests that compare the compression/decompression with the (open source I presume?) CPU implementation and the new GPU implementation, and make sure the results are the same.

d-v-b:
> Is that you speaking as a user of zarr or a maintainer?

Both? Zarr users and maintainers routinely expose bugs in libraries we depend on, and we routinely forward issues to those libraries, and occasionally contribute patches. If users report a bug or problem in nvcomp, what do we do? Suppose you or akshay leave nvidia, should I email support@nvidia.com when we have problems?

TomAugspurger:
> Sorry, I missed that. And sorry for the slow reply - I am not paid to work on zarr-python, and review stuff on a best efforts basis.

Likewise :/

> Am I right in thinking that GPU arrays and now decompression are a global option, not a per-array or a per-operation configuration?

I'm not sure... In `AsyncArray.getitem` there is a `prototype` argument that seems to be used for the buffers. I'm not sure whether the intent is for that to control all intermediate buffers or just the user-facing output buffer. The whole codec pipeline is still a bit of a black box to me.

I think that https://zarr.readthedocs.io/en/stable/user-guide/config.html is the relevant documentation here:

> For selecting custom implementations of codecs, pipelines, buffers and ndbuffers, first register the implementations in the registry and then select them in the config.

This isn't adding any new concepts: just a new implementation that uses the existing configuration system (which I think is fair to summarize as a global managed through our configuration system).

>  if it can be changed during a python session,

Sure... It's the same as any other configuration option. We don't explicitly show using any option in a context manager, but we do link to https://github.com/pytroll/donfig, which documents that behavior. I can duplicate that in our documentation if you want.

> and c) how decompression of GPU buffers is handled for codecs without a GPU implementation (I presume it just falls back to the CPU implementation)

Yeah, it's following the [documented behavior](https://zarr.readthedocs.io/en/stable/user-guide/config.html): the codec class associated with a given codec ID is used. We can repeat that in the GPU documentation and cross-reference the configuration docs.

So from a generic codec side of things, it'll just use whatever codec implementation is associated with that codec name. From a GPU-specific side of things, we'll want to document which codecs are currently implemented (I imagine the remaining codecs will be a relatively straightforward refactor based on top of what this branch implements).

> If users report a bug or problem in nvcomp, what do we do?

An issue filed at https://github.com/NVIDIA/CUDALibrarySamples would be best (I personally would ask the reporter to submit that rather than filing on their behalf, but you're probably more patient than me 😄).

>  I think there should be tests that compare the compression/decompression

I'll add those to this PR.

src/zarr/core/array.py
TomAugspurger: `get_codec_class` returns `type[Codec]` but this function specifically returns a `tuple[BytesBytesCodec]`.

TomAugspurger:
>  I think there should be tests that compare the compression/decompression 

https://github.com/zarr-developers/zarr-python/pull/2863/files#diff-b4b3e1bcdd519d839ba90f79f062ef0147362dbac854af26f871b53796791f21R69 has that.

I believe that all the comments have been addressed now but let me know if I missed anything.

I'm unsure why [codecov](https://app.codecov.io/gh/zarr-developers/zarr-python/pull/2863) is reporting that the coverage dropped here. Locally I see 100% coverage for `codec/gpu.py` and the lines changed in `core/buffer/gpu.py` changed in this PR are covered.

d-v-b:
I confess that I'm worried about retracing the steps that led to a large number of poorly-maintained storage backends in zarr-python 2. I'm not saying that this functionality will be poorly maintained, but it _is_ a risk, and I think we need to make a decision as a project about how we weigh that risk. For this particular feature, what would it look like if we spun it off into a separate repo under the zarr-developers org?

TomAugspurger:
> For this particular feature, what would it look like if we spun it off into a separate repo under the zarr-developers org?

Let's see:

For users, I think it could in principle be mostly seamless. We have entry points for loading a codec class without requiring users to import it. If this were in a separate package, I'd push for the `gpu` extra to depend on it, so that getting zarr-python with GPU support is still just `pip install zarr[gpu]` (this does induce a circular dependency, but it's in an extra so maybe that's not so bad).

We'd need some system to ensure that `zarr.config.enable_gpu()` has the ability to ask 3rd-party packages what changes to make. I don't think that system currently exists and so would need to be designed and implemented.

For maintainers, it'd be additional overhead from a separate project (CI, packaging, docs), which is minor but non-negligible.

My main question: is zarr-python really willing to commit to this as a stable interface, to the degree that third parties can build on it without worrying about things breaking? I worry that we don't know enough about the codec interface and its interaction with buffers yet to really commit to that (e.g. all your suggestions about buffers earlier in the thread).

d-v-b:
> My main question: is zarr-python really willing to commit to this as a stable interface, to the degree that third parties can build on it without worrying about things breaking? I worry that we don't know enough about the codec interface and its interaction with buffers yet to really commit to that (e.g. all your suggestions about buffers earlier in the thread).

I am personally not convinced that our codec API is as good as it could be, and I think the process of developing new codecs is a really important part of finding out how we can improve it. In particular, I'm worried that we are missing out on really simple performance optimizations like making the basic encoding and decoding routines take an optional buffer. But this is a separate question from whether we have the _will_ to change this API any time soon.

If we optimistically assume that the codec API might evolve in the future, then it does makes things easiest if the functionality in this PR is defined inside zarr python. So that's a solid point in favor of keeping it in this repo.

As a counter-point, the other day I was doing development on my mac laptop and I made changes that caused failures in the GPU tests. My mac doesn't have an nvidia gpu, so I couldn't run any of the cuda-specific tests, and so I had to rely on github actions to debug. That's a pretty bad development experience, so I do worry that this will only get worse if we add more cuda-specific code that needs to be updated as part of codebase-wide refactoring efforts, but can only be developed on a system with an nvidia gpu. The development friction this adds might actually cancel out the friction reduced by keeping the gpu-zstd code close to the codec definitions. So in the end I'm really not sure where the cuda zstd codec should live.

I don't want this effort to be stalled, and I think the functionality here is really important. It would be great to get input on this from other @zarr-developers/python-core-devs

dstansby:
Most OSS devs I know don't have access to a NVIDIA GPU on their machine, so having NVIDIA-specific code in here does greatly increase the barrier to development (basically, I'm echoing the above comment).

GPU development also requires provision of a non-standard GitHub actions runner, that I might be right in thinking we're paying for somehow? There's a risk that this goes away in the future.

So I'd be +1 to moving code designed for non-CPU hardware outside of zarr-python, perhaps into zarr-python-gpu or similar? The other advantage is you could move faster and break stuff (and not rely on zarr-python core dev reviews), compared to developing in zarr-python that has to stay very stable for it's wide user base.

dstansby:
I'm not that deep in the existing GPU code, so practically I'm not sure how feasible moving it out would be. But at least for codecs, it seems like it should definitely be possible to implement custom codecs in other packages?

TomAugspurger:
We talked about this a bit at last week's dev call. I think there's general agreement that native support for GPUs is important. And that at a technical level developing GPU-specific bits here or in some other repo is feasible.

For better or worse, we've defined public interfaces for many parts of Zarr-python. Having `gpu.{Buffer,NDBuffer}` in this repo and running on CI makes it clear when a change breaks something that the `gpu` implementation was relying on. But actually debugging and fixing those issues is extremely challenging without access to the right hardware.

We can split that out to another repo that contains the (NVIDIA) GPU specific bits (and likewise for MLX and other accelerators that want their own memory types, codecs, store, etc). But I do think it's important that the zarr-python project *commits* to supporting an interface. That involves a higher commitment to stability, and might mean some kind of (informal) policy on reverting commits that happen to break downstream libraries, assuming the issue is noticed and reported quickly. If we pair a strong commitment to stability from zarr-python core, along with "champions" from any of the downstream projects that commit to reporting issues and helping debug problems at the boundary, then I think splitting this out to a third-party repo and building on the extension points should work.

d-v-b:
we talked about this PR (and non-cpu-device-dependent extensions  to zarr python in general) at the recent zarr-python dev meeting. 

While asymptotically I think we would be happy seeing a stand-alone `zarr-developers/zarr-zstd` repo containing CUDA (and CPU!) implementations of the Zstd codec, it's not clear that we can ensure stability for the codec, buffer, store APIs etc needed for that to be a smooth developer experience. That points to putting this functionality in Zarr Python. The prospect of exposing this under the potential `zarr.experimental` module seemed like a good balance between getting it in user's hands while also giving us flexibility down the road. That path relies on getting #3490 merged.

TomAugspurger:
Thanks for the update. I can get the merge conflicts and any remaining issues ironed out this week.

src/zarr/codecs/gpu.py
```diff
@@ -0,0 +1,183 @@
...
+        result: list[Buffer | None] = []
+
+        for a, (_, spec) in zip(arrays, chunks_and_specs, strict=True):
+            if a is None:
+                result.append(None)
+            else:
+                a2 = cp.array(a, dtype=a.dtype, copy=False)
+                if a2.dtype != np.dtype("B"):
+                    a2 = a2.view(dtype=np.dtype("B"))
+                result.append(spec.prototype.buffer.from_array_like(a2))
```
TomAugspurger, line 108: Previously we just had `cp.array(a, dtype=np.dtype("B"), copy=False)`. cupy 14.x seems to be a bit stricter about this reinterpret / cast.

I think this mismatch comes in the first place from nvcomp using `int8` for some 1-D buffers, while zarr-python uses `uint8`.

src/zarr/codecs/gpu.py
```diff
@@ -0,0 +1,183 @@
...
+
+        # Convert to nvcomp arrays
+        filtered_inputs, none_indices = self._convert_to_nvcomp_arrays(chunks_and_specs)
+
+        outputs = self._zstd_codec.decode(filtered_inputs) if len(filtered_inputs) > 0 else []
+
+        # Record event for synchronization
+        event = cp.cuda.Event()
+        # Wait for decode to complete in a separate async thread
+        await asyncio.to_thread(event.synchronize)
```
TomAugspurger, line 138: Most of what I learned as part of https://tomaugspurger.net/posts/gpu-accelerated-zarr/ was the importance of avoiding unnecessary synchronizations being important for performance.

At the moment, it doesn't really matter since `_convert_from_nvcomp_arrays` will hit the stream sync inside of `nvcomp.Array`, but in the future I think we might want to offer an API that's actually asynchronous in the GPU sense (i.e. we'd return before the result is valid, and let the downstream operation synchronize when it needs something to be syncrhonized).

TomAugspurger:
@d-v-b remind me, was there anything else you wanted to see here?

d-v-b:
I think we talked about putting this in the experimental module but I no longer think that's really necessary, since this is not really an experimental feature, and users opt-in directly.

longer term I think codecs like this should live outside zarr-python, and be declared as optional zarr-python dependencies (e.g., `pip install zarr-python[cuda-codecs]`, but see #3867 for an explanation for why we can't do that today without a circular dependency mess.

docs/user-guide/gpu.md
```diff
@@ -29,3 +20,9 @@ type(z[:10, :10])
 ```
 
 Note that the output type is a `cupy.ndarray` rather than a NumPy array.
+
+For supported codecs, data will be decoded using the GPU via the [nvcomp] library.
+See [runtime-configuration][config-runtime-configuration] for more. Isseus and feature requestsfor NVIDIA nvCOMP can be reported in the nvcomp [issue tracker].
```
ljstrnadiii: "Issues" and ”requests for" typos.

TomAugspurger:
```suggestion
See [runtime-configuration][config-runtime-configuration] for more. Issues and feature requests for NVIDIA nvCOMP can be reported on the nvcomp [issue tracker].
```

JeppeKlitgaard:
I just happened to stumble over this and was wondering if it might be more appropriate to use `cuda` as the key for the extra (and perhaps also changing over in a few other places, where relevant)?

I think some parts of the Python eco-system have been guilty of conflating GPU support with CUDA, which can be problematic if AMD or other accelerators are added later — if ROCm support was added later, we would not want `zarr[gpu]` to have to install both CUDA and ROCm libraries.

I imagine `zarr.config.enable_gpu()` would keep its name, but the extra and perhaps some filenames, etc, would be changed to `cuda`. It may also be worthwhile to have separate extras for CUDA 13 and 12, for example, with `cuda` selecting the latest version.

An example of how this is handled elsewhere is the JAX project:
- https://docs.jax.dev/en/latest/installation.html (`jax[cuda12/cuda13/rocm7-local/tpu]`)

Hopefully this makes sense, otherwise I apologise for the intrusion!

d-v-b:
> I think some parts of the Python eco-system have been guilty of conflating GPU support with CUDA, which can be problematic if AMD or other accelerators are added later — if ROCm support was added later, we would not want `zarr[gpu]` to have to install both CUDA and ROCm libraries.

100% agree, I think the jax example is much better than calling CUDA devices "GPU"

TomAugspurger:
Agreed.

> It may also be worthwhile to have separate extras for CUDA 13 and 12, for example, with cuda selecting the latest version.

I think extras are fine (it's unfortunate we can't easily deprecate extras so anyone using the `"gpu"` extra will need to read the release notes, but that's OK). I'd really like to avoid using separate packages with suffixes like `cupy-cuda12x` and `cupy-cuda13x`, since that would force things like `zarr-cuda12x` or `zarr-cuda13x` (TBD how we express a dependency on cupy, and I'm now noticing our [current](https://github.com/zarr-developers/zarr-python/blob/main/pyproject.toml#L82) cupy dependency does just use `cupy-cuda12x`. Darn).

And just a note that this discussion around extras and making "gpu" generic is mostly independent of the earlier discussion around where to develop and package these features (https://github.com/zarr-developers/zarr-python/pull/2863#issuecomment-3127622619).

PR #2864, opened 2025-02-25
DOC: Fixed incorrect link in blosc.py

AdityaChaubeyIITM:
The "see more" under BloscCname class contains [old link](https://zarr.readthedocs.io/en/stable/tutorial.html#configuring-blosc) to refer more information about configuring Blosc but it is currently shifted to [new link](https://zarr.readthedocs.io/en/stable/user-guide/performance.html#configuring-blosc)

### DIFF
--- a/src/zarr/codecs/blosc.py
+++ b/src/zarr/codecs/blosc.py
@@ -55,7 +55,7 @@ class BloscCname(Enum):
     zlib = "zlib"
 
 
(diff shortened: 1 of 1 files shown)


joshmoore:
Thanks for the link correction :+1:, but of course, the target of the link says "Coming soon" so ... 🤷🏽

:smile:

dstansby (approved):
Thanks - I think better to link to a page that exists, even if the guide doesn't exist yet.

Issue #2867, opened 2025-02-26
update_attributes deletes existing attributes


relativityhd:
### Zarr version

v3.0.2

### Numcodecs version

v.0.15.1

### Python Version

3.13

### Operating System

Linux

### Installation

uv add zarr

### Description

When updating attributes via the `z.update_attributes` function, existing attributes get deleted instead of merged.
This is wrongly stated by the docs.

See code in zarr.core.array.py -> AsyncArray.update_attributes:

```py
        ...
        Notes
        -----
        - This method is asynchronous and should be awaited.
        - The updated attributes will be merged with existing attributes, and any conflicts will be
          overwritten by the new values.
        """
        # metadata.attributes is "frozen" so we simply clear and update the dict
        self.metadata.attributes.clear()
        self.metadata.attributes.update(new_attributes)

        # Write new metadata
        await self._save_metadata(self.metadata)

        return self
```

I think `self.metadata.attributes.clear()` just delete everything from the dictionary

### Steps to reproduce

```py
import zarr

z = zarr.create(10, store="data.zarr", overwrite=True)
z.attrs["a"] = []
z.attrs["b"] = 3
print(dict(z.attrs))
z.update_attributes({"a": [3, 4], "c": 4})
print(dict(z.attrs))
```

Outputs:

```sh
$ uv run repro.py
{'a': [], 'b': 3}
{'a': [3, 4], 'c': 4}
```

relativityhd:
Similar behaviour for groups

d-v-b:
thanks for the report, this is definitely not intended

moradology:
Can't assign myself but I'll pick this up if no one has, as yet

moradology:
Looks like the tests fixed in and some of the comments removed in #2870 suggest this behavior *was* intended. Kind of seems like an oversight though, because that isn't consistent across `update_attributes` implementations from what I saw

PR #2870, opened 2025-02-26
Prevent update_attributes from erasing all prior attributes

moradology:
This PR fixes attribute updating logic by clearing in calls to `put` and removing `clear` calls from all `update_attributes` methods. It also fixes an apparent oversight in the JSON type (advise if I'm wrong) which didn't allow lists. Sequence should handle both `tuple` and `list`

closes: #2867 

TODO:

### DIFF
--- a/src/zarr/core/array.py
+++ b/src/zarr/core/array.py
@@ -1609,8 +1609,6 @@ class AsyncArray(Generic[T_ArrayMetadata]):
         - The updated attributes will be merged with existing attributes, and any conflicts will be
           overwritten by the new values.
         """
-        # metadata.attributes is "frozen" so we simply clear and update the dict
-        self.metadata.attributes.clear()
         self.metadata.attributes.update(new_attributes)
 
         # Write new metadata
--- a/src/zarr/core/attributes.py
+++ b/src/zarr/core/attributes.py
@@ -50,6 +50,7 @@ class Attributes(MutableMapping[str, JSON]):
            >>> attrs
            {'a': 3, 'c': 4}
         """
+        self._obj.metadata.attributes.clear()
         self._obj = self._obj.update_attributes(d)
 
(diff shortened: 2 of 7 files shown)


moradology:
OK, maybe I'm being dense here but I'm pretty sure I didn't decrease code coverage by deleting a function call and adding three tests that cover the function in which that deletion occurred

src/zarr/core/array.py
```diff
@@ -1609,8 +1609,6 @@ async def update_attributes(self, new_attributes: dict[str, JSON]) -> Self:
         - The updated attributes will be merged with existing attributes, and any conflicts will be
           overwritten by the new values.
         """
-        # metadata.attributes is "frozen" so we simply clear and update the dict
-        self.metadata.attributes.clear()
         self.metadata.attributes.update(new_attributes)
```
d-v-b, line 1612: in principle we _do_ want the metadata documents to be immutable, but as long as we are using plain python dicts for the attributes, immutability is not realistic. so updating in-place is fine.

dcherian (approved):
LGTM. thanks

src/zarr/core/attributes.py
```diff
@@ -50,6 +50,7 @@ def put(self, d: dict[str, JSON]) -> None:
            >>> attrs
            {'a': 3, 'c': 4}
         """
+        self._obj.metadata.attributes.clear()
```
dstansby, line 53: Is this a separate bug fix to the change in `zarr/core/array.py`? If so, it should get it's own changelog entry, and test.

moradology: This is to keep `put` (which formerly was using the undesired behavior in `update_attributes`) working as expected rather than a new feature or bugfix

dstansby (approved):

PR #2872, opened 2025-02-27
Fix hatch matrix setup for minimal and optional dependencies

maxrjones:
I noticed after Martin's comment in https://github.com/zarr-developers/zarr-python/pull/2774#discussion_r1973748712 that even the 'minimal' tests include optional dependencies including fsspec and universal_pathlib. I think it's worth only including the required dependencies and those necessary for running tests / type checking in the "minimal" test environment.


d-v-b:
sounds good to me

maxrjones:
Tests are failing because the environment name changed, I'll fix this in a bit

tests/test_store/test_core.py
```diff
@@ -119,6 +120,8 @@ async def test_make_store_path_invalid() -> None:
         await make_store_path(1)  # type: ignore[arg-type]
 
 
+@pytest.mark.skipif(not importlib.util.find_spec("requests"), reason="requires requests")
+@pytest.mark.skipif(not importlib.util.find_spec("aiohttp"), reason="requires aiohttp")
```
dstansby, line 124: Might be a bit nicer to use [`@pytest.importorskip`](https://docs.pytest.org/en/stable/reference/reference.html#pytest.importorskip) here?

dstansby (commented):
Very 👍 this

tests/test_store/test_core.py
maxrjones: Ah, good call, I missed that I could use that on the test (rather than module) level

maxrjones: Done in [`b2442ba` (#2872)](https://github.com/zarr-developers/zarr-python/pull/2872/commits/b2442bab8277247ef530805bb70b0bedd9b52503)

pyproject.toml
```diff
@@ -143,28 +144,21 @@ hooks.vcs.version-file = "src/zarr/_version.py"
...
+deps = ["minimal", "optional"]
 
-[[tool.hatch.envs.test.matrix]]
-python = ["3.11", "3.12", "3.13"]
-numpy = ["1.25", "2.1"]
-features = ["gpu"]
+[tool.hatch.envs.test.overrides]
+matrix.deps.dependencies = [
+  {value = "zarr[remote, remote_tests, test, optional]", if = ["optional"]}
+]
```
maxrjones, line 158: I determined by looking at the old logs that there were no actual differences between the minimal and optional groups of the matrix. This override setup properly includes optional dependencies for the optional group and excludes them from the minimal group.

pyproject.toml
```diff
@@ -143,28 +144,21 @@ hooks.vcs.version-file = "src/zarr/_version.py"
...
-[[tool.hatch.envs.test.matrix]]
-python = ["3.11", "3.12", "3.13"]
-numpy = ["1.25", "2.1"]
-features = ["optional"]
+deps = ["minimal", "optional"]
 
-[[tool.hatch.envs.test.matrix]]
-python = ["3.11", "3.12", "3.13"]
-numpy = ["1.25", "2.1"]
-features = ["gpu"]
```
maxrjones, line 163: We separately have the `testgpu` environment, so there's no need to include another gpu feature set in the regular `test` matrix. I removed it because it only makes the setup more confusing.

.github/workflows/test.yml
```diff
@@ -64,6 +64,7 @@ jobs:
       run: |
         hatch env run --env test.py${{ matrix.python-version }}-${{ matrix.numpy-version }}-${{ matrix.dependency-set }} run-coverage
     - name: Upload coverage
+      if: ${{ matrix.dependency-set == 'optional' && matrix.os == 'ubuntu-latest' }}
```
maxrjones, line 67: The minimal environment will have less coverage than optional, so there's no value in uploading it to codecov

d-v-b (approved):

.github/workflows/test.yml
dstansby: Is there a downside to uploading it? I think it might be worth just keeping all the uploads to keep the complexity of this file down.

dstansby (approved):
For simplicity, I think might be worth keeping the codecov upload? Let me know what you think either way - otherwise this is looking 👍

.github/workflows/test.yml
maxrjones: It seemed to me a tradeoff between complexity in the workflow and complexity in understanding the codecov logs - https://app.codecov.io/gh/zarr-developers/zarr-python/commit/ebadec01e74acdd5bc3273be771c20a0f6a67cce. But not many people will interact with those logs so I can remove this line

dstansby: That's fair - I tend not to look at the logs, but agree that it makes the logs a bit easier. Lets leave the line in and I'll merge 🚢

PR #2873, opened 2025-02-28
Deprecate v3 code on v2 branch

dstansby:
For zarr-python 2.19.0 I thought it would be good to tear out support for partial reads/writes (deprecated in https://github.com/zarr-developers/zarr-python/pull/2844), and any zarr 3 code (this PR). That should make it easier to maintain for at least the next few months.

This PR deprecates v3 code on the v2 branch. It also removes all the v3 tests - since this API is unsupported, I thought the easiest thing to do here was just tear out the tests, instead of trying to keep them but catch deprecation warnings.

I also skipped the sqlite store tests because they are failing (e.g., [see this GH actions run](https://github.com/zarr-developers/zarr-python/actions/runs/13491208294/job/37689643311?pr=2858) from another recent PR). I think debugging/supporting this is pretty low priority, so I'm happy with just plain skipping these to get the v2 tests passing again.


jhamman (changes requested):
I would like to suggest we keep the experimental v3 code as-is in Zarr-Python until we hit the end of 6-month support window following the 3.0. I don't think there is much of a maintenance burden with keeping this in the support branch, is there?

dstansby:
I managed to fix the sqlite failures at https://github.com/zarr-developers/zarr-python/pull/2880

jhamman:
> What's the motivation for wanting to keep it around?

We are still using it and until there is a complete migration story for datatypes/codecs in v3, we can't fully migrate.

_Noting that we've also discussed this on [Zulip](https://ossci.zulipchat.com/#narrow/channel/423692-Zarr/topic/zarr-python.202.2E18.2E6.20and.202.2E19.2E0)_

-----

I'll also note that we already have a `FutureWarning` in place warning users that this version will be removed in 3.0. 

```ipython
In [1]: import zarr

In [2]: g = zarr.group(zarr_version=3)
/Users/jhamman/miniforge3/envs/earthmover-dev-311/lib/python3.11/site-packages/zarr/_storage/store.py:39: FutureWarning: The experimental Zarr V3 implementation in this version of Zarr-Python is not in alignment with the final V3 specification. This version will be removed in Zarr-Python 3 in favor of a spec compliant version.
  warnings.warn(
```

Based on the wording here, we've already communicated that the experimental v3 support would be removed in 3.0 -- not 2.18.

Issue #2871, opened 2025-02-27
Public API for `zarr.core.buffer`


TomAugspurger:
### Zarr version

v3

### Numcodecs version

na

### Python Version

na

### Operating System

na

### Installation

na

### Description

Over in https://github.com/pydata/xarray/pull/10079, I'd like to use `zarr.core.buffer.cpu`. But after https://github.com/zarr-developers/zarr-python/issues/2621 `zarr.core` is offically private.

I'd suggest `zarr.buffer.{cpu,gpu}` as a public API for these modules. Similar-ish to `zarr.storage`.

### Steps to reproduce

na

### Additional output

xref https://github.com/zarr-developers/zarr-python/pull/2641, but that covered the ABC. In this case I'll need the concrete implementation (CPU specifically).

jhamman:
Yes! Let's do this.

PR #2876, opened 2025-02-28
Public API for buffer objects

TomAugspurger:
This moves the public imports from buffer things out of `zarr.core.buffer`.

- Abstract stuff is availble under `zarr.abc.buffer`.
- Concrete implementations are available under `zarr.buffer.{cpu,gpu}`.

I haven't added any *new* tests, but I updated the tests in `tests/test_buffer.py` to use the public API.

Closes https://github.com/zarr-developers/zarr-python/issues/2871


dstansby:
Do you think it's worth just moving the actual code to the new public location, so it doesn't live in `zarr.core.buffer` any more? That would make for a simpler code base and be my preference, but I might be missing some reason not to move the code.

TomAugspurger:
Mmm, I think I'd still want the implementation to be in a private module so that implementation details like numpy imports don't leak into the public API's namespace. So it'd end up looking pretty similar. And if we're moving stuff out of `zarr.core` I'd want to do it properly, since I wouldn't be surprised if people were already using it. IMO not worth it right now.

TomAugspurger:
> I wouldn't be surprised if people were already using it.

Yep: https://github.com/earth-mover/icechunk/blob/584d4f2b0aad1160f8e4ba9a48ebbf770c058c00/icechunk-python/python/icechunk/store.py#L13

I think this PR takes care of those `zarr.core.buffer` imports. IMO, we should ensure that all the types in our public API (like `BytesLike`) are exported somewhere, but that should be done separately.

src/zarr/buffer/__init__.py
```diff
@@ -0,0 +1,12 @@
+"""
+Public API for implementations of the Zarr Buffer interface.
+
+See Also
+========
+arr.abc.buffer: Abstract base class for the Zarr Buffer interface.
```
dstansby, line 6:
```suggestion
:mod:`arr.abc.buffer` : Abstract base class for the Zarr Buffer interface.
```
Would be good to give this a link

src/zarr/buffer/__init__.py
```diff
@@ -0,0 +1,12 @@
+"""
+Public API for implementations of the Zarr Buffer interface.
```
dstansby, line 2:
```suggestion
Implementations of the Zarr Buffer interface.
```
If it's in the docs, it's implicitly public API already

src/zarr/buffer/__init__.py
```diff
@@ -0,0 +1,12 @@
...
+
+See Also
+========
+arr.abc.buffer: Abstract base class for the Zarr Buffer interface.
+"""
+
+from ..core.buffer import default_buffer_prototype
+from . import cpu, gpu
```
dstansby, line 10: I don't think we import submodules like this anywhere else? If we don't, to be consistent I would remove these imports, and we can always discuss doing this throughout the package in an issue.

src/zarr/buffer/__init__.py
```diff
@@ -0,0 +1,12 @@
+"""
+Public API for implementations of the Zarr Buffer interface.
+
+See Also
+========
+arr.abc.buffer: Abstract base class for the Zarr Buffer interface.
+"""
+
+from ..core.buffer import default_buffer_prototype
```
dstansby, line 9:
```suggestion
from zarr.core.buffer import default_buffer_prototype
```
AFAIK it's a bit safer and clearer to do an absolute import.

src/zarr/buffer/__init__.py
TomAugspurger: I confirmed locally that numpydoc / sphinx do this automatically.

TomAugspurger:
I don't plan to move the implementation at this time. We already know that some libraries, like icechunk, are depending on it. I'm don't want to break those implementations because Zarr lacked a public API for this previously. We need to offer a public API first and then give them some time to migrate over.

I'll post a proposed plan over in https://github.com/zarr-developers/zarr-python/issues/2621.

docs/user-guide/extending.rst
```diff
@@ -83,7 +83,10 @@ Coming soon.
 Custom array buffers
 --------------------
 
-Coming soon.
+zarr-python provides control where and how arrays stored in memory through
```
dstansby, line 86:
```suggestion
Zarr-Python provides control for where and how arrays stored in memory through
```

docs/user-guide/extending.rst
```diff
@@ -83,7 +83,10 @@ Coming soon.
...
+:mod:`zarr.buffer`. Currently both CPU (the default) and GPU implementations are
+provided (see :ref:`user-guide-gpu` for more). You can implement your own buffer
```
dstansby, line 88:
```suggestion
provided (see :ref:`user-guide-gpu` for more info). You can implement your own buffer
```

src/zarr/buffer/__init__.py
dstansby: It doesn't look like `cpu` and `gpu` have been removed as of the latest commit? Because we don't do this kind of import elsewhere (e.g., see `zarr/api/__init__.py`, that doesn't import `synchronous` or `asyncrhonous`), I think we should be consistent and not import `cpu` or `gpu` here.

dstansby (changes requested):
As well as my inline comments/suggestions, I noticed that there are parts of the API that are typed with `zarr.core.buffer...` - it would be good to change those to use the new public interface. As an example, [`zarr.abc.codec.ArrayArrayCodec`](https://zarr--2876.org.readthedocs.build/en/2876/api/zarr/abc/codec/index.html#module-contents)

dstansby:
> I don't plan to move the implementation at this time. We already know that some libraries, like icechunk, are depending on it. I don't want to break those implementations because Zarr lacked a public API for this previously. We need to offer a public API first and then give them some time to migrate over.

👍 - my suggestion would be to introduce those deprecations in this PR for the buffer stuff that's been made public, but happy to punt that to a later PR if that's easier.

TomAugspurger:
The latest pair of commits updates our `zarr.config`, which references `zarr.core.buffer`. We needed to slightly update our registration code to let the class being registered have a different qualname / import path than where they're defined.

src/zarr/buffer/__init__.py
TomAugspurger: Sorry I misunderstood your earlier comment.

This is to ease migration for users: Previously `import zarr` would make, e.g. `zarr.core.buffer.cpu.Buffer` available, as a side-effect of us importing modules internally.

The closest analog here is `zarr.storage`, which imports all store subclasses, rather than making users use the full import path. We can't do that exact same thing here hough, since both `.gpu` and `.cpu` use the same names (Buffer, NDBuffer, etc), so the imports need to be namespaced.
 
We could just not import `cpu` and `gpu` and make the user do that import, but in this case I think consistency with `zarr.store` is most important.

docs/user-guide/extending.rst
TomAugspurger: I think the current version sounds OK.

TomAugspurger:
Thanks for the review.

> I noticed that there are parts of the API that are typed 

This is proving to be a bit more work than I can do right now. Feel free to push a commit if you're able to, or we can merge this as is since and handle that later.

TomAugspurger:
> 👍 - my suggestion would be to introduce those deprecations in this PR for the buffer stuff that's been made public, but happy to punt that to a later PR if that's easier.

I'd recommend waiting. A release with this public API will give us a chance to migrate packages like icechunk without their users ever seeing a warning.

d-v-b:
It would be nice to get this merged. In the interest of moving forward here, @dstansby are you OK with sorting out some of the requested changes in follow-up PRs?

dstansby:
I had a think about this again, and I actually think quite strongly that if we are moving where we want users to import this code from, then we should deprecate importing it from the old locations at the same time, in the same PR. This would make sure:

1. It's logistically possible to deprecate the old namespace (otherwise, if it's impossible there's no point complicating things with a new namespace)
2. Doing the deprecation happens and isn't lost.

I know it's extra work, but I think in order to make sure the current nice bit of work is worth it, pushing users towards using the new work (ie deprecating the old import path) should happen at the same time.

dstansby:
that's my two cents - if there are a couple of other developers that feel strongly in the other direction (splitting creation of new API/deprecating old so there's a period with duplicate API), then happy to do that. I have a gut feeling that it has the potential to go wrong and create a mess if not done in one go though.

TomAugspurger:
If you want to do that then go for it.I don’t like introducing warnings to users of downstream libraries before those libraries have had a chance to adapt.On Apr 22, 2025, at 5:00 PM, David Stansby ***@***.***> wrote:﻿
  that's my two cents - if there are a couple of other developers that feel strongly in the other direction (splitting creation of new API/deprecating old so there's a period with duplicate API), then happy to do that. I have a gut feeling that it has the potential to go wrong and create a mess if not done in one go though.—Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you authored the thread.Message ID: ***@***.***>



dstansby left a comment (zarr-developers/zarr-python#2876)
that's my two cents - if there are a couple of other developers that feel strongly in the other direction (splitting creation of new API/deprecating old so there's a period with duplicate API), then happy to do that. I have a gut feeling that it has the potential to go wrong and create a mess if not done in one go though.

d-v-b:
should we fold this into the 3.1 release?

TomAugspurger:
If @dstansby is OK with things more or less as is I'll fix the merge conflict. Otherwise I don't plan to work on it.

dstansby:
This *definitely* shouldn't be merged without a clear migration plan explaining what exsiting users of `zarr.core.buffer` should do (presumably just switch their imports to `zarr.buffer`?).

I'm okay with adding the new API, then deprecating later 👍  - could you open an issue to make sure deprecating isn't forgotten though?

TomAugspurger:
> clear migration plan explaining 

Updated the release note.

> then deprecating later

https://github.com/zarr-developers/zarr-python/issues/2621

dstansby:
When conflicts are fixed, this is good to be merged

Issue #2616, opened 2025-01-02
[v3] support for datetime and timedelta dtypes


jhamman:
Zarr-Python 2 [supported datetime and timedelta dtypes](https://zarr.readthedocs.io/en/v2.18.4/tutorial.html#datetimes-and-timedeltas). This functionality has not made it into Zarr-Python 3 yet (in part because the v3 spec does not have established extension dtypes for these yet).

Example of the behavior for Zarr 2:

```python
z = zarr.array(['2007-07-13', '2006-01-13', '2010-08-13'], dtype='M8[D]')
z
<zarr.core.Array (3,) datetime64[D]>
z[:]
array(['2007-07-13', '2006-01-13', '2010-08-13'], dtype='datetime64[D]')
z[0]
numpy.datetime64('2007-07-13')
z[0] = '1999-12-31'
z[:]
array(['1999-12-31', '2006-01-13', '2010-08-13'], dtype='datetime64[D]')
```

This issue tracks the development of datetime and timedelta support in Zarr-Python 3.

christine-e-smit:
Hey Joe! All our zarr stores use the datetime type for our time dimension variable. It sounds like we'll have to transform all our zarr stores or help you make the change to support this data. What kind of effort do you think would be required for relative newbies to add datetime data type support to zarr v3?

jhamman:
https://github.com/zarr-developers/zarr-python/pull/2874 should be merged in the next week or so and will include support for datetime dtypes.

christine-e-smit:
Oh I'm sorry. My eyes went right over the links in this ticket for some reason. Thank you.

christine-e-smit:
We are now actively in the process of updating our code. It looks as though #2874 has not been merged. Any idea on the ETA? Do we need to delay our migration?

d-v-b:
@christine-e-smit #2874 has been merged, and we plan to issue release with this functionality once the documentation is finished and any remaining bugs are ironed out (see #3090)

d-v-b:
closing as completed via #2874

christine-e-smit:
> [@christine-e-smit](https://github.com/christine-e-smit) [#2874](https://github.com/zarr-developers/zarr-python/pull/2874) has been merged, and we plan to issue release with this functionality once the documentation is finished and any remaining bugs are ironed out (see [#3090](https://github.com/zarr-developers/zarr-python/issues/3090))

[...]

PR #2874, opened 2025-02-28
refactor v3 data types

d-v-b:
As per #2750, we need a new model of data types if we want to support more data types. Accordingly, this PR will refactor data types for the zarr v3 side of the codebase and make them extensible. I would also like to handle v2 as well with the same data structures, and confine the v2 / v3 differences to the places where they vary.

In `main`,all the v3 data types are encoded as variants of an enum (i.e., strings). Enumerating each dtype as a string is cumbersome for datetimes, that are parametrized by a time unit, and plain unworkable for parametric dtypes like fixed-length strings, which are parametrized by their length. This means we need a model of data types that can be parametrized, and I think separate classes is probably the way to go here. Separating the different data types into different objects also gives us a natural way to capture some of the per-data type variability baked into the spec: each data type class can define its own default value, and also define methods for how its scalars should be converted to / from JSON.

This is a very rough draft right now -- I'm mostly posting this for visibility as I iterate on it.

PR #2884, opened 2025-03-03
support for datetime and timedelta dtypes (#2616)

RJ-Vestrum:
My first contribution addressing https://github.com/zarr-developers/zarr-python/issues/2616! 

Summary of changes:
    * Add support for the datetime dtypes
    * Add support for the timedelta dtypes
    * Update tests to validate the fill_values for for datetime and timedelta
    * Towncrier file for changes

TODO:


d-v-b:
@RJ-Vestrum thanks so much for this work! We definitely want to have these datatypes in zarr, but there are few complications:
- The Zarr v2 spec made support for datatimes and timedeltas very easy, but the zarr v3 spec makes this a bit harder, because it doesn't provide any guidance at all for adding new data types. https://github.com/zarr-developers/zarr-specs/pull/330 is an effort to fix this, but that discussion is ongoing. One certainty is that we will allow data types to be specified in metadata as a dict with the structure `{"name": <dtype name>, "configuration": {...params}}`. For parametric dtypes like datetime and timedelta, I think we should consider storing them in JSON like so:
```
"data_type": {"name": "numpy.timedelta64", "configuration": {"unit": "ns"}}
``` 
IMO this is a lot clearer than using a stringly-typed identifier like `"datetime64[ns]"`. But it would also complicate your work in this PR

A second complication for your work is my current effort to completely refactor how dtypes work in zarr-python (see #2874). In that PR, I'm unifying v2 and v3 data type handling via a completely new set of python objects to model dtypes. If you're interested, I'd love to get your feedback / input on that work, because I haven't added datetimes or timedeltas yet. I'd be happy to explain how you could add those under the proposed framework in that PR if you're interested. Otherwise, there's a risk that the work in _this_ PR would need to be re-done in the new framework, provided #2874 is successful.

Since there's so many moving parts right now (spec changes, multiple PRs), we should probably invest some effort in coordinating our efforts. let me know if you're interested!

RJ-Vestrum:
@d-v-b thanks for the detailed response and for explaining all the moving parts here! I really appreciate the context on both the spec changes and the dtype refactor in #2874.

This PR was mostly a workaround I had been using locally, and I wanted to share it in case it could contribute to the development process. I don’t have much time to coordinate right now, but I’ll take a look at #2874 and keep it in mind. I’m planning to revisit this in April, so if #2874 is finalized by mid-April, I can rework this PR to align with the new framework then.

I also fully agree that specifying data types in metadata rather than using stringly-typed identifiers will make things much clearer.

Thanks again for taking the time to lay this all out!

rabernat:
I'm curious...how specific is the on-disk storage of `numpy.timedelta64` etc. to numpy? Is it actually just `int64` data, plus a few parameters like units?

d-v-b:
> I'm curious...how specific is the on-disk storage of `numpy.timedelta64` etc. to numpy? Is it actually just `int64` data, plus a few parameters like units?

great question, I have no idea, but if the answer is "yes", then it's even more reason to think about an arrow-style dtype hierarchy where primitive types form the basis of logical types.

d-v-b:
@RJ-Vestrum if you have a chance, could you try out https://github.com/zarr-developers/zarr-python/pull/2874 and see if it works for your purposes?

d-v-b:
now that the functionality in this PR has been added via #2874, should we close this PR?

dstansby:
I'll close this - if this is still relevant after #2874, please feel free to re-open or open a new PR.

PR #2890, opened 2025-03-04
DOC: Add data= examples and async note (fix #2809)

### LINKED ISSUE #2809: DOC: create_array(..., data=,...)

I am very confused about the argument `data` in `create_array`. A common use case is to simply serialize an in memory array, in which case I tend to pass it as the `data=in_memory_array` argument. However, I cannot find the `data` argument in the documentation.

Using IPyhon, on the other hand, `zarr.create_array` clearly has the argument, while `zarr.Group.create_array` doesn't seem to expose the interface. I am quite confused about the discrepancy. If this is intentional, please document it.
LLM also suggest that
```Python
zarr.create_array("store.zarr", data=in_memory_data)
```
is more efficient than
```Python
arr = zar.create_arra("store.zarr", shape=in_memory_data.shape, dtype=in_memory_data.dtype)
arr[...] = in_memory_data
```
I have no idea whether this is true or not. `zarr.create_array(..., data=in_memory_data)` might be indeed more efficient as it seems to be written asynchronously. But the documentation seems to by quite lacking, what the best practice is.

------

This might be a bit out of scope for this issue, this issue, so please tell me if it's out of scope. But from the documentation, I don't really see how to leverage the asynchronous nature of the `zarr` implementation. A common pattern I encounter is, that data is generated in parallel using multiprocessing (as it is CPU bound) and persisted using `zarr` (probably disc bound). Is there a preferred pattern, to use `zarr` as an asynchronous sink for the generated data? If so, it would be great to include it in the docs.

_No response_

OPMTerra:
This is a documentation improvements for #2809! This PR adds clear examples and guidance for using the `data` parameter in array creation.

### What's Included
- Practical example of `create_array(data=...)` usage
- Performance comparison with assignment method
- Async storage note for cloud implementations

### Checklist (For Maintainers)
✅ Docs updated in `user-guide/arrays.rst`  
✅ Pure documentation changes (no code modifications)  
✅ Built docs locally to verify formatting  

These changes help users:
1. Discover the efficient `data` parameter  
2. Understand performance best practices  
3. Leverage async writes in cloud environments  

I'm happy to make any adjustments needed! 😊  
@OPMTerra

### DIFF
--- a/docs/user-guide/arrays.rst
+++ b/docs/user-guide/arrays.rst
@@ -30,6 +30,29 @@ module documentation.
 
 .. _user-guide-array:
 
+Initializing with Data
+---------------------
+Pass existing data during array creation for better performance:
+
+.. code-block:: python
+
+   import zarr
+   import numpy as np
+
+   data = np.random.rand(1000, 1000)
+   arr = zarr.create_array("data.zarr", shape=data.shape, data=data)
+
+This avoids a separate write step and is more efficient than:
+
+.. code-block:: python
+
+   # Less efficient approach
+   arr = zarr.create_array("data.zarr", shape=(1000, 1000))
+   arr[:] = data
+
+.. note::
+   For cloud stores like S3, use ``data=`` with async writes for best performance.
+
 Reading and writing data
 ------------------------
 


docs/user-guide/arrays.rst
```diff
@@ -30,6 +30,29 @@ module documentation.
...
+
+.. code-block:: python
+
+   # Less efficient approach
+   arr = zarr.create_array("data.zarr", shape=(1000, 1000))
+   arr[:] = data
+
+.. note::
+   For cloud stores like S3, use ``data=`` with async writes for best performance.
+
```
jhamman, line 55: this section isn't quite right. 

1. There isn't a performance gain by passing `data` directly. The operations are entirely sequential and equivalent to the example above.
2. RE async -- both examples write data concurrently and take advantage of the async backend.

OPMTerra: hello @jhamman 
Thank you for pointing this! I'll update the docs.

OPMTerra: @jhamman 
I've updated the docs to:  

1. Remove performance claims  
2. Clarify async behavior applies to both methods  
3. Focus on code brevity over efficiency  

would this be better?

jhamman: Much better. Thanks @OPMTerra!

docs/user-guide/arrays.rst
```diff
@@ -30,6 +30,29 @@ module documentation.
 
 .. _user-guide-array:
 
+Initializing with Data
+----------------------
+Pass existing data during array creation for better performance:
```
d-v-b, line 35: I don't think we should emphasize the performance benefit here, as we do not have measurements to support this claim. I would instead direct people to use the `data` parameter because it is convenient.

docs/user-guide/arrays.rst
```diff
@@ -30,6 +30,29 @@ module documentation.
...
+This provides a more concise syntax compared to separate assignment:
+
+.. code-block:: python
+
+   # Equivalent but verbose
+   arr = zarr.create_array("data.zarr", shape=(1000, 1000))
+   arr[:] = data
+
+.. note::
+   Both methods leverage async writes when using async-compatible stores like S3.
```
d-v-b, line 54: All of our storage classes have an asynchronous API, so all I/O operations can be described as leveraging asynchronous writes. So I would remove this note.

OPMTerra: @d-v-b Thank you for both the feedbacks I"ll update the docs

OPMTerra: Hi @d-v-b 👋,  

I've updated the docs per your feedback:  
1. Removed performance claims  
2. Deleted async note  
3. Focused on code brevity  

Please re-review it and let me know if further changes are required! 🙏

docs/user-guide/arrays.rst
```diff
@@ -30,6 +30,25 @@ module documentation.
 
 .. _user-guide-array:
 
+Initializing with Data
```
jhamman, line 33:
```suggestion
Initializing with data
```

docs/user-guide/arrays.rst
```diff
@@ -30,6 +30,25 @@ module documentation.
...
+----------------------
+Pass existing data during array creation for convenience:
+
+.. code-block:: python
+
+   import zarr
+   import numpy as np
+
+   data = np.random.rand(1000, 1000)
+   arr = zarr.create_array("data.zarr", shape=data.shape, data=data)
```
jhamman, line 43:
```suggestion
   arr = zarr.create_array("data.zarr", data=data)
```

Q: is shape required here?

OPMTerra:
I've updated the screenshot to reflect the latest changes:  
- Removed redundant `shape` parameter  
- Simplified code example

Issue #2881, opened 2025-03-03
remove partial codec initialization


d-v-b:
Some codec classes (like the blosc codec) can be partially initialized, where codec creation succeeds but the resulting object is not actually serializable because certain fields that are required for serialization are nullable for codec creation. I believe the logic for this pattern was to spare users the burden of specifying codec properties that depend on the data type of their array, like bloscshuffle. The cost of this pattern is code complexity, indirection, and irregularity of the codec API surface.

Now that we have easier methods for automatically specifying codecs (e.g, the "auto" value in `create_array`), I don't think we need to support partial codec initialization any more. So we should remove this functionality from the codecs that use it, and instead require codecs are fully valid for serialization when they are constructed.

AdityaChaubeyIITM:
Hello @d-v-b , I see 2 ways to implement it, either make details like typesize and shuffle compulsory while defining, or ask data type instead and then set typesize and shuffle. Which of the 2 would you suggest? Or is there any other way that I missed? Please let me know

AdityaChaubeyIITM:
Sure, I shall start working on this then.

Issue #2889, opened 2025-03-04
Change versioning policy to allow API breaking in minor versions?


dstansby:
Our [current versioning policy](https://zarr.readthedocs.io/en/stable/developers/contributing.html#api-compatibility) says:

> If a release contains any backwards-incompatible public API changes, the major version number should be incremented (e.g., 2.3.0 -> 3.0.0).

 An example of where we could currently have to increment to v4 is at https://github.com/zarr-developers/zarr-python/pull/2819, and @d-v-b also says:

> We were already planning on making breaking changes without incrementing the major version number -- deprecated functions like create were slated to be removed long before a 4.x release

So I think the proposal on the table is to update the versioning policy so:

- The major version of `zarr-python` is tied to the Zarr format specification version it supports
- The minor version is for breaking changes or major new features
- The micro version is for bugfixes

Ping @zarr-developers/python-core-devs for comment, since this is a major change. Comments/suggestions/disagreements very welcome - if you just generally agree with the above proposal feel free to react with a 👍 to this post.

d-v-b:
Generally speaking I don't have strong feelings here, but I do think we should ensure that we have a versioning policy that allows small breaking changes as needed. Although we released v3, I don't think all the code / APIs are fully settled, and it would be a shame to preserve loads of warts because we are afraid of breaking changes. The alternative to changing the versioning policy would be to just release 4.x, 5.x, etc as needed. I don't mind that, as long as I can push out my breaking changes :)

d-v-b:
> I don't think we would even need to increment the major version if we supported a new Zarr version, if it is unnoticable by users. 

I agree with this in principle but the [current versioning guidelines say otherwise](https://zarr.readthedocs.io/en/stable/developers/contributing.html#data-format-compatibility): 

> If a release of the Zarr library implements a different version of the storage specification, then the major version number of the Zarr library should be incremented. 

It might be worth removing this statement from the guidelines. I doubt that will trouble anyone.

TomAugspurger:
> It might be worth removing this statement from the guidelines. I doubt that will trouble anyone.

+1. IMO, the versioning of zarr-python the library should be tied to the source code, rather than the spec version.

And for folks who haven't read it, https://hynek.me/articles/semver-will-not-save-you/ is a nice thing to keep in mind when discussing semver. I don't have a super strong preference for whether we try to stick strictly to semver vs. something else, as long as

1. we're clear about it in our versioning policy
[...]

PR #2896, opened 2025-03-06
Resolves issue #2881 by removing partial codec initialization

AdityaChaubeyIITM:
[Description of PR]
Changed the way Blosc and Bytes codecs are initialized by making all the fields are mandatory 
Removed checks for is None as they would never be none.


d-v-b:
thanks for working on this! As these are breaking changes to the blosc codec, I don't think we can merge this until we decide on when the next breaking release should be. See #2889 for context.

PR #2910, opened 2025-03-14
update version policy to use effver

d-v-b:
This changes the versioning policy from something semver-ish to [effver](https://jacobtomlinson.dev/effver/).

based on conversations in [this issue](https://github.com/zarr-developers/zarr-python/issues/2889#issuecomment-2708165027)

TODO:

closes https://github.com/zarr-developers/zarr-python/issues/2889


docs/developers/contributing.rst
```diff
@@ -261,90 +261,64 @@ Merging pull requests
...
+The Zarr library is an implementation of a file format standard defined externally -- see the `Zarr specifications website <https://zarr-specs.readthedocs.io>`_ for the list of Zarr file format specifications.
 
-Ideally, any bug fixes that don't change the public API should be released as soon as
-possible. It is fine for a micro release to contain only a single bug fix.
 
-When to make a minor release is at the discretion of the core developers. There are no
-hard-and-fast rules, e.g., it is fine to make a minor release to make a single new
-feature available; equally, it is fine to make a minor release that includes a number of
-changes.
+If an existing Zarr file format version changes, or a new version of the Zarr file format is released, then the Zarr library will generally require changes. It is very likely that
```
jhamman, line 317:
```suggestion
If an existing Zarr format version changes, or a new version of the Zarr format is released, then the Zarr library will generally require changes. It is very likely that
```

jhamman: I don't like calling Zarr things "files"

docs/developers/contributing.rst
```diff
@@ -261,90 +261,64 @@ Merging pull requests
...
-Ideally, any bug fixes that don't change the public API should be released as soon as
-possible. It is fine for a micro release to contain only a single bug fix.
 
-When to make a minor release is at the discretion of the core developers. There are no
-hard-and-fast rules, e.g., it is fine to make a minor release to make a single new
-feature available; equally, it is fine to make a minor release that includes a number of
-changes.
+If an existing Zarr format version changes, or a new version of the Zarr format is released, then the Zarr library will generally require changes. It is very likely that
+a new Zarr file format will require extensive breaking changes to the Zarr library, and so support for a new Zarr file format in the Zarr library will almost certainly come in new ``major`` release.
+For some period of time after the Zarr library adds support for a new Zarr file format, there may be a period of accelerated changes as developers refine APIs that were added to support the new format.
```
jhamman, line 319:
```suggestion
a new Zarr format will require extensive breaking changes to the Zarr library, and so support for a new Zarr format in the Zarr library will almost certainly come in new ``major`` release.
For some period of time after the Zarr library adds support for a new Zarr format, there may be a period of accelerated changes as developers refine APIs that were added to support the new format.
```

jhamman (approved):
Thanks @d-v-b -- this looks great!

docs/developers/contributing.rst
```diff
@@ -261,90 +261,64 @@ Merging pull requests
...
-
-Backwards-incompatible changes to the experimental API can be included in a minor release,
-although this should be minimised if possible. I.e., it would be preferable to save up
-backwards-incompatible changes to the experimental API to be included in a major release, and to
-stabilise those features at the same time (i.e., move from experimental to public API), rather than
-frequently tinkering with the experimental API in minor releases.
+Versioning
+""""""""""
+The Zarr library uses a version identifier with the form ``<major>.<minor>.<patch>``, where each of value is
+an integer. For example, when this document was written the version of Zarr was ``3.0.4``. A release of the Zarr library is associated with a new
```
d-v-b, line 276:
```suggestion
Versions of the Zarr are identified by a triplet of integers with the form  ``<major>.<minor>.<patch>``. For example, when this document was written the version of Zarr was ``3.0.4``. A release of the Zarr library is associated with a new
```

d-v-b:
@zarr-developers/python-core-devs it would be good to get yalls eyes on this

normanrz (approved):

docs/developers/contributing.rst
```diff
@@ -261,90 +261,63 @@ Merging pull requests
...
-the major version number should be incremented (e.g., 2.3.0 -> 3.0.0).
-
-Backwards-incompatible changes to the experimental API can be included in a minor release,
-although this should be minimised if possible. I.e., it would be preferable to save up
-backwards-incompatible changes to the experimental API to be included in a major release, and to
-stabilise those features at the same time (i.e., move from experimental to public API), rather than
-frequently tinkering with the experimental API in minor releases.
+Versioning
+""""""""""
+Versions of this library are identified by a triplet of integers with the form  ``<major>.<minor>.<patch>``. For example, when this document was written the version of Zarr was ``3.0.4``. A release of the Zarr library is associated with a new
```
dstansby, line 275:
```suggestion
Versions of this library are identified by a triplet of integers with the form  ``<major>.<minor>.<patch>``, for example 3.0.4. A release of ``zarr-python`` is associated with a new
```

dstansby (commented):
Thanks for tackling this! I left some comments inline. I think a major gap in the policy as it stands is that it doesn't embody @TomAugspurger's [excellent points](https://github.com/zarr-developers/zarr-python/issues/2889#issuecomment-2701566052) of being considerate to downstream users by:

> - preserving backwards compatibility where possible
> - giving notice of deprecations through FutureWarnings
> - giving users and libraries time to adapt to those deprecations

Matplotlib has some good [guidelines on making breaking API changes using deprecations](https://matplotlib.org/devdocs/devel/api_changes.html#rules) which could be taken as inspiration here.

TomAugspurger (approved):

d-v-b:
> Thanks for tackling this! I left some comments inline. I think a major gap in the policy as it stands is that it doesn't embody @TomAugspurger's [excellent points](https://github.com/zarr-developers/zarr-python/issues/2889#issuecomment-2701566052) of being considerate to downstream users by:
> 
> > * preserving backwards compatibility where possible
> > * giving notice of deprecations through FutureWarnings
> > * giving users and libraries time to adapt to those deprecations
> 
> Matplotlib has some good [guidelines on making breaking API changes using deprecations](https://matplotlib.org/devdocs/devel/api_changes.html#rules) which could be taken as inspiration here.

@dstansby have a look at the content I added in https://github.com/zarr-developers/zarr-python/pull/2910/commits/b90f0ef358d3dded5b7462e8594ef4dccfd41dc1

dcherian (approved):

d-v-b:
does anyone else think a community announcement (e.g. a blog post) would make sense here? If so, I can start working on one

docs/developers/contributing.rst
```diff
@@ -261,90 +261,64 @@ Merging pull requests
...
-possible. It is fine for a micro release to contain only a single bug fix.
 
-When to make a minor release is at the discretion of the core developers. There are no
-hard-and-fast rules, e.g., it is fine to make a minor release to make a single new
-feature available; equally, it is fine to make a minor release that includes a number of
-changes.
+If an existing Zarr format version changes, or a new version of the Zarr format is released, then the Zarr library will generally require changes. It is very likely that
+a new Zarr format will require extensive breaking changes to the Zarr library, and so support for a new Zarr format in the Zarr library will almost certainly come in new ``major`` release.
+For some period of time after the Zarr library adds support for a new Zarr format, there may be a period of accelerated changes as developers refine APIs that were added to support the new format.
+Breaking changes may more frequent until a new API is fully stabilized, and these breaking changes may occur in ``minor`` releases.
```
d-v-b, line 320:
```suggestion
When the Zarr library adds support for a new Zarr format, there may be a period of accelerated changes as developers refine newly added APIs and deprecate old APIs. In such a transitional phase breaking changes may be more frequent than usual.
```

sanketverma1704:
Thanks for working on this, @d-v-b! Important change.

> does anyone else think a community announcement (e.g. a blog post) would make sense here? If so, I can start working on one

We should at least make a social media announcement on BlueSky, LinkedIn, and Mastodon. A blog post would be great! I can take care of the announcements and help with blog writing/review.

A minor suggestion: Can we wrap the text to 80 (or similar) columns?

docs/developers/contributing.rst
```diff
@@ -261,90 +261,63 @@ Merging pull requests
...
+
+  Users should always feel safe upgrading to a the latest patch release.
+
+Note that this versioning scheme is not consistent with `Semantic Versioning <https://semver.org/>`_. The Zarr library
+may release breaking changes in ``minor`` releases, or even ``patch`` releases under exceptional circumstances.
+But we should strive to avoid doing so. A better model for our versioning scheme is `Intended Effort Versioning <https://jacobtomlinson.dev/effver/>`_, or "EffVer".
+The guiding principle off EffVer is to categorize releases based on the *expected effort required to upgrade to that release*.
+
+Zarr developers should make changes as smooth as possible for users. This means making backwards-compatible changes wherever possible.
+When a backwards-incompatible change is necessary, users should be notified well in advance, e.g. via informative deprecation warnings.
```
dstansby, line 309:
```suggestion
Zarr developers should make changes as smooth as possible for users. This means making backwards-compatible changes wherever possible.
When a backwards-incompatible change is necessary, users should be notified well in advance, e.g. via informative deprecation warnings
Breaking changes should be made no earlier than 6 months after the deprecation is first made available in a release.
```

I think it would be helpful for both developers and users to set expectations a bit more concretely here. From my experience 6 months strikes a good balance between giving users time to adapt, and giving developers room to make changes in the long run, but that's a suggestion that I think should be open to discussion.

d-v-b: I would rather not introduce language about specific time ranges.

d-v-b: The right metric is probably N minor releases, not N months. And I think fleshing out our deprecation policy should probably happen in another PR.

dstansby: What's the reasoning for using number of releases, and not a timeframe ?

My thinking on time was there isn't a regular release schedule (at the moment it's just whenever one or a few core devs feel like it's worth doing a release), so giving a minimum time of deprecation gives users a constant time frame that they have to adapt to deprecations.

d-v-b: - A deprecation warning can confidently state `"zarr-python 3.<current minor version + 2>.0 will drop support for feature foo"`. You can't do the same thing with a fixed time range. 
- If the version that drops the deprecated feature is well-defined, users can reliably pin to a version that precedes it. This requires a fixed relationship between version numbers and breaking changes, which you can't do if there's a temporal variable.
- The time range is reasonable but also completely arbitrary. This means it can be undermined by equally reasonable-but-arbitrary reasoning, which sets the stage for people to deviate from this recommendation on a case-by-case basis. ("this fix really needs to happen but it's been 5 months since release, who else thinks we can drop this feature 1 month early", etc). Normalizing deviance is bad, so if we want to introduce rules for feature deprecation (and we should), those rules should be really easy to follow. If we really want to drop a deprecated feature and we link deprecation to releases, then we can just release more (and users can always pin). But if deprecation is linked to the calendar, we can't speed up the calendar.

dstansby: Just to compare with other similar (stable) projects, some prior art here is:

- [SciPy](https://docs.scipy.org/doc/scipy/dev/core-dev/index.html#deprecations): "Wait till at least 6 months after the release date of the release that introduced the DeprecationWarning before removing the functionality."
- [Matplotlib](https://matplotlib.org/devdocs/devel/api_changes.html#deprecate-api): "Deprecated API is generally removed (expired) two point-releases after introduction of the deprecation."
- [numpy](https://numpy.org/doc/stable/dev/depending_on_numpy.html#understanding-numpy-s-versioning-and-api-abi-stability): "Minor versions are released regularly, typically every 6 months. Minor versions contain new features, deprecations, and removals of previously deprecated code. "

So there does seem to be a range of options currently in use.

I see the advantages of having a clear "feature X will be dropped in version x.y.z" message. How about having a release schedule in that case, with minor releases every (at least) 6 months, and deprecations have to be in one minor release before being actioned? That way users know they only have to check every 6 months.

d-v-b: I'd love to have a release schedule, but I think it's out of scope for this PR

d-v-b: I'm happy to hear other opinions here -- I can definitely see the advantage to putting all our "here's how we version" in a single PR, but I don't feel confident in saying "this is our release schedule" without getting a lot of feedback.

dstansby: As it stands this PR gets rid of API stability guarentees provided by semantic versioning, and doesn't replace them with any other guarentees (ie, after this PR it would be possible to deprecate something tomorrow in a v3.1, then remove it next week in a v3.2). So I think if we're going to get rid of a guarentee that gives users time to adapt, it's reasonable to replace it with something similar at the same time instead of waiting for a follow up PR.

dstansby:
```suggestion
Zarr developers should make changes as smooth as possible for users. This means making changes backwards-compatible wherever possible.
When a backwards-incompatible change is necessary, where possible users will be notified in advance with a deprecation warning raised when user code depends on the old behaviour.
Deprecations can be introduced in minor releases (not patch releases).
The deprecated behaviour will be removed one minor release after the release that introduced the deprecation.
```
How about this as a compromise? It doesn't have any time based guarentees (perhaps we can discuss that in a follow up?), but it does at least give users one minor version to adapt their code if there's any deprecations.

d-v-b: > after this PR it would be possible to deprecate something tomorrow in a v3.1, then remove it next week in a v3.2

My read of effver is that it explicitly discourages this behavior, because what you describe would require a lot of effort for users to adjust to.

d-v-b: > How about this as a compromise? It doesn't have any time based guarentees (perhaps we can discuss that in a follow up?), but it does at least give users one minor version to adapt their code if there's any deprecations.

I think this proposal is reasonable but if we are widening the scope of this PR to include specifying our deprecation policy then I would like to input from all of @zarr-developers/python-core-devs

jni: > The deprecated behaviour will be removed one minor release after the release that introduced the deprecation.

I would say *at least* one minor release after. I don't think that we should be overly prescriptive here. Some deprecations are easy for downstream to adopt, others are hard, so the time/number of releases should not be set in stone.

I agree @d-v-b that having to make decisions for every tiny change is exhausting for everyone and should be avoided, but until we are doing regular releases, we *can't* avoid it. And this @d-v-b 

> If we really want to drop a deprecated feature and we link deprecation to releases, then we can just release more (and users can always pin).

is unhinged 😂 and goes completely against the spirit of effver — as you yourself noted a few comments later:

> My read of effver is that it explicitly discourages this behavior, because what you describe would require a lot of effort for users to adjust to.

tbh I would postpone this whole discussion to a later issue and leave the paragraph unchanged:

> When a backwards-incompatible change is necessary, users should be notified well in advance, e.g. via informative deprecation warnings.

leaves us enough room to adopt any of the strategies discussed here. So let's leave it in, and then open a new issue to discuss which way we want to go, if any, or if we want to stay flexible for some time longer.

jni: > Also I don't think calling ideas "unhinged" adds much to this discussion.

Apologies for my clumsy attempt at humorous hyperbole. 🙏

jhamman (approved):
Thanks @d-v-b for leading us through this.

PR #2911, opened 2025-03-17
fix(storage): accept pytest tmp_dir as valid store_like paths

jhamman:
This PR adjusts the logic in `make_store_path` to accept pytest's `tmp_dir` fixture as a valid store_like argument. This worked in Zarr 2 and the fix here brings back the prior behavior.

An alternative approach would have been to cast any unknown argument type to a string but this seemed less safe.

TODO:

### DIFF
--- a/src/zarr/storage/_common.py
+++ b/src/zarr/storage/_common.py
@@ -346,7 +346,7 @@ def _is_pytest_legacy_path(path: Any) -> bool:
     # https://docs.pytest.org/en/stable/how-to/tmp_path.html#tmp-path
     try:
         from _pytest.compat import LEGACY_PATH
-    except ImportError:
+    except ImportError:  # pragma: no cover
         return False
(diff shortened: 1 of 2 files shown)


dcherian:
Why is anyone sending in a fixture instead of the requested parameter value?

jhamman:
> Why is anyone sending in a fixture instead of the requested parameter value?

Zarr-Python 2 quite eagerly called str() on any input that wasn't a MutableMapping. The Xarray test suite took advantage of this and used `tmp_dir` as an input to many tests. 

xref: https://github.com/pydata/xarray/pull/10020

d-v-b (approved):
It would be better for this to be fixed in xarray but I understand that it might not be so easy.

dcherian:
Hmmm... fine to merge, but looks like we've fixed the xarray tests in any case.

PR #2918, opened 2025-03-19
Fix bug in fsspec store delete_dir.

### LINKED ISSUE #2917: Varibles are unintenionally overwritten on a google cloud store
### Zarr version

3.0.5

### Numcodecs version

0.15.1

### Python Version

3.11

### Operating System

Mac (any)

### Installation

from source

### Description

When writing to gcsfs there is a potential for variables to overwrite eachother when the start of the second variable to be written is contained in an already written variable. i.e. `a` will overwrite `a*`

### Steps to reproduce

```python
import zarr
import time

url = zarr.storage.FsspecStore.from_url(
    "gs://some-bucket/overwrite-test.zarr",
    # "s3://some-bucket/overwrite-test.zarr", # this works fine
)
root = zarr.create_group(store=url, overwrite=True)
shape = (10, 10)
dtype = int
encoding = { "overwrite": True}

# aa gets overwritten by 'a'
root.create_array(name="aa", shape=shape, dtype=dtype, **encoding)
root.create_array(name="a", shape=shape, dtype=dtype, **encoding)

# b does not get overwritted by bb
root.create_array(name="b", shape=shape, dtype=dtype, **encoding)
root.create_array(name="bb", shape=shape, dtype=dtype, **encoding)
z = zarr.open(url)
print(sorted(list(root.keys())))
print(sorted(list(z.keys())))
```

### Additional output

gcsfs version: '2025.2.0+1.gd0fd744.dirty

dcherian:
Closes #2917

I don't know how to test it but have confirmed the MVCE works now.

cc @martindurant 

TODO:

### DIFF
--- a/changes/2918.bugfix.rst
+++ b/changes/2918.bugfix.rst
@@ -1,0 +1,1 @@
+Fix bug with `FsspecStore.delete_dir` (:issue:`2917`)
--- a/src/zarr/storage/_fsspec.py
+++ b/src/zarr/storage/_fsspec.py
@@ -298,7 +298,7 @@ class FsspecStore(Store):
         path_to_delete = _dereference_path(self.path, prefix)
 
         with suppress(*self.allowed_exceptions):
-            await self.fs._rm(path_to_delete, recursive=True)
+            await self.fs._rm(path_to_delete + "/", recursive=True)
 
     async def exists(self, key: str) -> bool:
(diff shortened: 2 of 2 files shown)


martindurant:
I suppose it would be a good idea to include the specific test with "aa" and "a" arrays for various storage backends.

Issue #2929, opened 2025-03-24
Metadata comparison fails for NaN fill_values


TomNicholas:
### Zarr version

main

### Numcodecs version

n/a

### Python Version

3.12

### Operating System

linux

### Installation

pip editable

### Description

Two `Metadata` objects with identical attributes will compare not equal if they both have `NaN` for a `fill_value`. This is because the `__eq__` check introspects deeper until it finds e.g. the `np.float32(nan)` type, but 

```python
In [4]: bool(np.float32('nan') == np.float32('nan'))
Out[4]: False
```

(See https://stackoverflow.com/a/10059796 for why numpy NaNs behave like this.)

The solution needs to be to actually check two `Metadata` classes are `__eq__` with dedicated code, not just trusting the python dataclasses' automatically-generated `__eq__` method to do it correctly.

xref https://github.com/zarr-developers/VirtualiZarr/issues/501

### Steps to reproduce

```python
In [12]: metadata1 = ArrayV3Metadata(
    ...:     shape=(2,),
    ...:     data_type=np.float32,
    ...:     chunk_grid={
    ...:         "name": "regular",
    ...:         "configuration": {"chunk_shape": (2,)},
    ...:     },
    ...:     chunk_key_encoding={"name": "default"},
    ...:     fill_value=np.float32('nan'),
    ...:     codecs=({'name': 'bytes', 'configuration': {'endian': 'little'}},),
    ...:     attributes={},
    ...:     dimension_names=None,
    ...:     storage_transformers=None,
    ...: )

In [13]: metadata2 = ArrayV3Metadata(
    ...:     shape=(2,),
    ...:     data_type=np.float32,
    ...:     chunk_grid={
    ...:         "name": "regular",
    ...:         "configuration": {"chunk_shape": (2,)},
    ...:     },
    ...:     chunk_key_encoding={"name": "default"},
    ...:     fill_value=np.float32('nan'),
    ...:     codecs=({'name': 'bytes', 'configuration': {'endian': 'little'}},),
    ...:     attributes={},
    ...:     dimension_names=None,
    ...:     storage_transformers=None,
    ...: )

In [14]: bool(metadata1 == metadata2)
Out[14]: False
```

d-v-b:
closed as completed via #2874

PR #2930, opened 2025-03-24
Fix nan fill value metadata comparison

TomNicholas:
Fix for #2929

TODO:
* [ ] ~~Add docstrings and API docs for any new/modified user-facing classes and functions~~

### DIFF
--- a/src/zarr/abc/metadata.py
+++ b/src/zarr/abc/metadata.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from collections.abc import Sequence
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any
 
 if TYPE_CHECKING:
     from typing import Self
@@ -9,6 +9,8 @@ if TYPE_CHECKING:
     from zarr.core.common import JSON
 
 from dataclasses import dataclass, fields
+
+import numpy as np
 
 __all__ = ["Metadata"]
 
@@ -44,3 +46,23 @@ class Metadata:
         """
 
         return cls(**data)
+    
+    def __eq__(self, other: Any) -> bool:
+        """Checks metadata are identical, including special treatment for NaN fill_values."""
+        if not isinstance(other, type(self)):
+            return False
+        
+        metadata_dict1 = self.to_dict()
+        metadata_dict2 = other.to_dict()
+
+        # fill_value is a special case because numpy NaNs cannot be compared using __eq__, see https://stackoverflow.com/a/10059796
+        fill_value1 = metadata_dict1.pop("fill_value")
+        fill_value2 = metadata_dict2.pop("fill_value")
+        if np.isnan(fill_value1) and np.isnan(fill_value2):
+            fill_values_equal = fill_value1.dtype == fill_value2.dtype
+        else:
+            fill_values_equal = fill_value1 == fill_value2
+
+        # everything else in ArrayV3Metadata is a string, Enum, or Dataclass
+        return fill_values_equal and metadata_dict1 == metadata_dict2
+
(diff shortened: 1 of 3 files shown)


d-v-b:
Thanks for this! When we add datetimes, we will need to change this code again with a separate block for [`np.nat`](https://stackoverflow.com/questions/38509538/numpy-checking-if-a-value-is-nat).

src/zarr/abc/metadata.py
```diff
@@ -44,3 +46,23 @@ def from_dict(cls, data: dict[str, JSON]) -> Self:
         """
 
         return cls(**data)
+    
+    def __eq__(self, other: Any) -> bool:
```
TomNicholas, line 50: Overriding this means we lose the nice auto-generated dataclass `__eq__` method, which prints out like this:

```python
Matching attributes:
['shape',
 'data_type',
 'chunk_grid',
 'chunk_key_encoding',
 'codecs',
 'attributes',
 'dimension_names',
 'zarr_format',
 'node_type',
 'storage_transformers']
Differing attributes:
['fill_value']

Drill down into differing attribute fill_value:
  fill_value: np.float32(nan) != np.float32(nan)
```

But I don't see an easy way to keep that formatting and make it handle `fill_value` properly too...

tests/test_metadata/test_v3.py
```diff
@@ -411,3 +411,20 @@ def test_dtypes(dtype_str: str) -> None:
     else:
         # return type for vlen types may vary depending on numpy version
         assert dt.byte_count is None
+
+
+def test_metadata_comparison_with_nan_fill_value():
```
TomNicholas, line 416: This test could presumably be parametrized over both v2 and v3 metadata, but you don't appear to have any other tests which do that. Is there a reason for that?

d-v-b:
more broadly, using `np.isnan` is picky about the types it accepts: `np.isnan('im not nan') will raise a type error, for example. So I think some try... excepts might be necessary here.

tests/test_metadata/test_v3.py
d-v-b: the two metadata documents are separate entities, so they get tested separately. if there's shared functionality that they both depend on, but isn't part of their shared base class, then that shared functionality should have its own test. 

higher up in the stack we test the array classes against the two metadata varieties, because part of the job of the array classes is to abstract over the v2 v3 differences.

TomNicholas:
> just a thought but what if the fill_value was already JSON-encoded
>
> This would totally prevent the need to handle edge cases in __eq__

I like this idea - it seems what you actually care about is whether the representation of the metadata _on-disk_ is equal, which is what the JSON-encoded version is. Would this involve changing `Metadata`'s internal representation of the `fill_value`?

d-v-b:
> > just a thought but what if the fill_value was already JSON-encoded
> > This would totally prevent the need to handle edge cases in **eq**
> 
> I like this idea - it seems what you actually care about is whether the representation of the metadata _on-disk_ is equal, which is what the JSON-encoded version is. Would this involve changing `Metadata`'s internal representation of the `fill_value`?

yes, it would require pushing the "treat this JSON value as a numpy scalar" job higher up in the stack, or we could add a method on the metadata classes.

FWIW, over in #2874 I'm giving each data type object a method that converts the JSON representation of a scalar to an in-memory representation. That's exactly what we would use here.

d-v-b:
as of #2874, the `to_dict` method of the metadata classes returns JSON-serializable values that shouldn't be problematic for equality checking. So maybe this can be closed?

TomNicholas:
Yes that works nicely now! See https://github.com/zarr-developers/VirtualiZarr/pull/677/files#r2212963076 - I was able to delete the internal metadata-comparing function I had. 🥳 

So I think this and #2929 can be closed.

Issue #2931, opened 2025-03-26
Appending to re-opened zarr array does not respect `write_empty_chunks`


ianhi:
### Zarr version

3.0.7.dev3+g9e8b50ae

### Numcodecs version

0.15.1

### Python Version

3.13.2

### Operating System

mac

### Installation

from source

### Description

When you open an existing array in order to resize and write new values to it the `write_empty_chunks` config is no longer respected. 

### Steps to reproduce

```python
import os
import zarr

z = zarr.create_array(
    store="appending.zarr",
    name="test",
    shape=(2,),
    dtype=int,
    chunks=(1,),
    config={"write_empty_chunks": True},
    overwrite=True,
    fill_value=0,
    zarr_format=3,
)

z[:] = 0

# if you comment out this re-opening then everything works as intended
z = zarr.open(
    "appending.zarr", path="test", config={"write_empty_chunks": True}, fill_value=0
)
print(os.listdir("appending.zarr/test/c/"))

z.resize((4,))

print(os.listdir("appending.zarr/test/c/"))
l = os.listdir("appending.zarr/test/c/")
print(l)
assert l == ['0', '1',]

# now put a value into the newly appended chunks
z[2:] = 0
l = os.listdir("appending.zarr/test/c/")
print(l)
# assert l == ['0', '1', '2', '3']


# putting in a value of zero makes all empty chunks be removed from disk
z[:] = 0

l = os.listdir("appending.zarr/test/c/")
print(l)
# assert l == ['0', '1', '2', '3']
```

gives
```
['0', '1']
['0', '1']
['0', '1']
['0', '1']
[]
```

not only are chunks 2 and 3 not written, but unexpectedly chunks 0 and 1 are removed from disk.

if you comment out the `zarr.open...` line then the output looks like:

```
['0', '1']
['0', '1']
['0', '1']
['0', '1', '3', '2']
['0', '1', '3', '2']
```

ianhi:
This was fixed by https://github.com/zarr-developers/zarr-python/pull/3378

PR #2933, opened 2025-03-26
Fix: respect `config` when opening an existing array

ianhi:
Fixes https://github.com/zarr-developers/zarr-python/issues/2931 

Previously there were edge cases with open which would result in the `write_empty_chunks` config was being ignored. This PR has two separate fixes.
1st: for `zarr.open` line 327
2nd: `zarr.open_array` required some config parsing as well as a change to the api of `AsyncArray.open`

* [NA] New/modified features documented in `docs/user-guide/*.rst`

### DIFF
--- a/tests/test_array.py
+++ b/tests/test_array.py
@@ -746,6 +746,44 @@ def test_resize_2d(store: MemoryStore, zarr_format: ZarrFormat) -> None:
     assert new_shape == z[:].shape
 
 
+@pytest.mark.parametrize("store", ["local"], indirect=True)
+@pytest.mark.parametrize("open", ["open", "open_array"])
+def test_append_config_passed(store: LocalStore, open: str, zarr_format: ZarrFormat) -> None:
+    z = zarr.create_array(
+        store=store,
+        name="test",
+        shape=(2,),
+        dtype=int,
+        fill_value=0,
+        chunks=(1,),
+        config={"write_empty_chunks": True},
+        overwrite=True,
+        zarr_format=zarr_format,
+    )
+    z[:] = 0
+    print(store)
+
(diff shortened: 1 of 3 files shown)


ianhi:
Discussed with @jhamman at the dev meeting and we came to the following conclusion:

1. Un-deprecate `order` and `write_empty_chunks` in favor of keeping these explicit
2. deprecate the `config` argument to `create` and to `AsyncArray`
3. default to the values in the global config when these values are `None` which should be the default

This keeps better consistency with zarr-python 2 and is more explicit in what the user is passing rather a layer of obfuscation by wrapping in `ArrayConfigLike` as a dict.

ianhi:
Motivated by https://github.com/pydata/xarray/pull/10177

cc @d-v-b

tests/test_array.py
```diff
@@ -746,6 +746,43 @@ def test_resize_2d(store: MemoryStore, zarr_format: ZarrFormat) -> None:
     assert new_shape == z[:].shape
 
 
+@pytest.mark.parametrize("store", ["local"], indirect=True)
+@pytest.mark.parametrize("open", ["open", "open_array"])
+def test_append_config_passed(store: LocalStore, open: str, zarr_format: ZarrFormat) -> None:
```
d-v-b, line 751: Why do our existing `write_empty_chunks` not work here?

d-v-b: for context, it would probably be preferable to expand the scope of an existing test instead of adding a new one

ianhi:
Part of the fixes here ended up in https://github.com/zarr-developers/zarr-python/pull/3378

but the remaining confusing with the possibliity of passing mulltiple ways of defining the same config remains

Issue #2859, opened 2025-02-24
StoreTests don't test `Store.set_partial_values`


TomAugspurger:
### Zarr version

v3

### Numcodecs version

na

### Python Version

na

### Operating System

Linux

### Installation

source

### Description

[`zarr.testing.store.StoreTests`](https://github.com/zarr-developers/zarr-python/blob/8d2fb47252a7ed5eb163ccfb0401e4e0486f72f6/src/zarr/testing/store.py#L37) provides some unit tests that any class implementing the Store interface should pass. Currently, there aren't any tests that exercise the abstract `Store.set_partial_values` API.

### Steps to reproduce

Spotted this while working on zarr-v3 support in kvikio: https://github.com/TomAugspurger/kvikio/tree/tom/zarr-v3

d-v-b:
I suspect we have `set_partial_values` only because it's [defined in the spec](https://github.com/zarr-developers/zarr-specs/blob/d38b0db4440560171a1145e8fb5b19c96cd63a79/docs/v3/core/v3.0.rst?plain=1#L1321) alongside `set`. In `zarr-python` we defined `set` to take a byte_range keyword argument, rendering `set_partial_values` is unnecessary. We should probably remove that method entirely. And maybe the spec should not have tried to define a store API.

TomAugspurger:
Gotcha, thanks. Although it looks like only some store classes support it, and it's not part of the `Store` abc yet. Happy to repurpose this issue to discuss that.

Would it be OK to add that to the `zarr.abc.Store.set` method, even with a default of `None`? Just unconditionally passing through a `byte_range=...` argument is liable to break stores that haven't been updated to accept that argument (even if they ignore it). I guess we have a couple options:

1. Maybe that's the purpose of the `supports_partial_writes` abstract method? Should we look at that, and pass it through when that's True?
2. Maybe use `inspect.signature` to see `set` accepts `support_partial_writes` keyword. This is a bit fragile, but might work.

> And maybe the spec should not have tried to define a store API.

Yeah, IMO that seems a bit beyond the scope of the spec. I can see something like that being useful in the spec for examples, but not necessarily something that implementations should follow.

d-v-b:
[...]

PR #2944, opened 2025-04-02
Avoid memory copy in local store write

tomwhite:
This removes an unnecessary memory copy when writing to a local file, bringing the memory usage in line with Zarr Python v2 for this case (see https://github.com/tomwhite/memray-array).

TODO:

### DIFF
--- a/changes/2944.misc.rst
+++ b/changes/2944.misc.rst
@@ -1,0 +1,1 @@
+Avoid an unnecessary memory copy when writing Zarr to a local file
--- a/src/zarr/storage/_local.py
+++ b/src/zarr/storage/_local.py
@@ -51,7 +51,7 @@ def _put(
     if start is not None:
         with path.open("r+b") as f:
             f.seek(start)
-            f.write(value.as_numpy_array())
+            f.write(value.as_numpy_array())  # type: ignore[arg-type]
         return None
     else:
         view = memoryview(value.as_numpy_array())
--- a/src/zarr/storage/_local.py
+++ b/src/zarr/storage/_local.py
@@ -51,6 +51,7 @@ def _put(
     if start is not None:
         with path.open("r+b") as f:
             f.seek(start)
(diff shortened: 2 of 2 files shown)


d-v-b (approved):

d-v-b:
thanks! this looks great.

dstansby (changes requested):
As the `pre-commit` typing failure indicates, doesn't `f.write()` need bytes given instead of a numpy array here? Maybe it doesn't since the tests all seem to pass - if that's the case, could you add an appropriate `# ignore` comment to fix the mypy error, and add a comment in the code explaining why it's okay to hand a numpy array directly to `write()`?

TomAugspurger:
`write` supports anything implementing the buffer protocol, which ndarrays do. Prior to 3.12 that wasn't available in Python (https://peps.python.org/pep-0688/).

Pumping the python version in our mypy config at https://github.com/zarr-developers/zarr-python/blob/9e8b50ae19cc63ad573f58569c3ef5826a5c60fc/pyproject.toml#L352 might let this work.

dstansby:
[SPEC 0](https://scientific-python.org/specs/spec-0000/#support-window) has support Python 3.11 for a little while yet, so I don't think we should bump the Python version in the config. If it's all fine in practice, then just an ignore comment to make `mypy` happy should be fine.

TomAugspurger (approved):
Also needs a release note: https://zarr.readthedocs.io/en/stable/developers/contributing.html#changelog

Looks good once those are addressed.

dstansby (approved):

TomAugspurger:
The codecov failure seems to come from [increasing the number of lines in an uncovered branch](https://app.codecov.io/gh/zarr-developers/zarr-python/pull/2944?dropdown=coverage&src=pr&el=h1). https://github.com/zarr-developers/zarr-python/issues/2859 is already tracking that, so I'll go ahead and merge this.

Issue #2934, opened 2025-03-26
Add (or document migration from) `contains_array` and `contains_group`


dstansby:
In zarr-python 2 there were `zarr.storage.contains_{array, group}` functions. These don't seem to have been replaced, or mentioned in the migration guide. It would be good to either:

- Add replacements for them in zarr-python 3
- Explain in the migration guide what code that used these should do now

dstansby:
@d-v-b pointed out these still exist in https://github.com/zarr-developers/zarr-python/pull/2947#discussion_r2049750948 - they just need putting back in the public API somewhere, and if their import path has changed that needs documenting in the migration guide.

PR #2947, opened 2025-04-03
Expose get_node() in the public API

dstansby:
Fixes https://github.com/zarr-developers/zarr-python/issues/2934.

TODO:


docs/user-guide/v3_migration.rst
```diff
@@ -154,8 +153,30 @@ Common replacements:
    change ensures that all store methods are non-blocking and are as performant as
    possible.
 
-Beyond the changes store interface, a number of deprecated stores were also removed in
-Zarr-Python 3. See :issue:`1274` for more details on the removal of these stores.
+Removed store API
+^^^^^^^^^^^^^^^^^
+
+``zarr.storage.contains_array()`` and ``zarr.storage.contains_group()`` have been removed.
```
d-v-b, line 159: [`contains_array`](https://github.com/zarr-developers/zarr-python/blob/f9b5a3bf47e981c58a66222a15bd0df9413b8662/src/zarr/storage/_common.py#L437) and [`contains_group`](https://github.com/zarr-developers/zarr-python/blob/f9b5a3bf47e981c58a66222a15bd0df9413b8662/src/zarr/storage/_common.py#L472) have not been removed

dstansby: From a user perspective they have, because they're in a private sub-module. I can make them public in this PR if they were always intended to be public?

d-v-b: I think we can say that `contains_array` and `contains_group` are not yet part of the public API. I don't have any problem adding them to the public API but maybe that's better for another PR.

Issue #2950, opened 2025-04-03
Saving and re-opening a v2 array does not report (or save) correct memory order


dstansby:
```python
import zarr
import zarr.storage

store = zarr.storage.MemoryStore()
arr = zarr.create((100,), store=store, path="a", zarr_format=2, order="F")
print(arr.order)
# "F"

arr = zarr.open(store, path="a")
print(arr.order)
# "C"
```

I would expect the memory order to be persisted and the same when I re-open the array.

d-v-b:
iirc this is because we made order part of the [runtime configuration](https://github.com/zarr-developers/zarr-python/blob/9e8b50ae19cc63ad573f58569c3ef5826a5c60fc/src/zarr/core/config.py#L78), due to `order` not being part of the zarr v3 metadata model. I'm pretty sure a _global_ config is the wrong location for this parameter -- I think we need a per-array namespace for runtime configuration like this.

d-v-b:
if you inspect the `array.metadata.order`, I think you should see that it has the expected `F` value. The issue is that when zarr opens an array, it defers to the global config when setting the order of the in-memory array associated with that zarr array, whether the array is v2 or v3. The problem is our use of this global config.

dstansby:
> if you inspect the array.metadata.order, I think you should see that it has the expected F value. 

Nope, `arr.metadata.order` still gives the wrong order.

d-v-b:
:( that's bad! we definitely need to fix this!

d-v-b:
i can get to this in the next few days unless someone does it first

d-v-b:
> if you inspect the `array.metadata.order`, I think you should see that it has the expected `F` value. The issue is that when zarr opens an array, it defers to the global config when setting the order of the in-memory array associated with that zarr array, whether the array is v2 or v3. The problem is our use of this global config.

this explanation is actually very wrong -- we are already using the array-specific config. the underlying issue, addressed in #2951, is a simple bug in our implementation.

PR #2951, opened 2025-04-03
Fix specifying memory order in v2 arrays

dstansby:
Fixes https://github.com/zarr-developers/zarr-python/issues/2950

TODO:


d-v-b:
this looks good! and my explanation from https://github.com/zarr-developers/zarr-python/issues/2950 was wrong -- this has nothing to do with the global config, it's just buggy code (that i think I wrote).

dstansby:
So I updated the test, and it now passes a bit more. But the check that the actual array data is F-order now fails, so it looks like the data still isn't (and never was) actually being written in F-order. More digging needed...

tests/test_v2.py
```diff
@@ -194,12 +194,8 @@ def test_create_array_defaults(store: Store):
         )
 
 
-@pytest.mark.parametrize("array_order", ["C", "F"])
-@pytest.mark.parametrize("data_order", ["C", "F"])
-@pytest.mark.parametrize("memory_order", ["C", "F"])
-def test_v2_non_contiguous(
-    array_order: Literal["C", "F"], data_order: Literal["C", "F"], memory_order: Literal["C", "F"]
-) -> None:
```
dstansby, line 202: @normanrz you originally wrote this test in https://github.com/zarr-developers/zarr-python/pull/2679, but I don't understand why there are three different types of order, and there's no documentation, or even comments on this test explaining.

I've therefore chosen to be a little bit bold, and for at least v2 data this PR now assumes there is only one order. If we need to deal with different types of orders, I'm certainly open to that, but that increased complexity probably needs to come with some very good user docs, given in zarr-python v2 there was only one type of memory order.

Since you originally wrote this test, I wanted to see what you thought about this PR?

normanrz: array_order = order in the metadata
data_order = order of the numpy array that is used to write
memory_order = order in the runtime config which is used for reading data, ie the order of the output numpy array

At least the first two are necessary for this test. If you want to change the behavior to always read in array_order, then you can drop the memory_order.

dstansby: Maybe I'm missing something, but don't we just need to parametrize with one order? We pass that one order parameter to the array creation routine, and then check it's correct in both the metadata and the order the data is stored in memory, right (which is how the test works currently in this PR)?

normanrz: Consider this: You create a Zarr array with F order. Next you create a numpy array with C order. You write the numpy array in the Zarr array. This test makes sure that the data on-disk is in F order. That's why you need at least 2 orders.

dstansby: Aha, that makes sense - thanks!

d-v-b (approved):

PR #2954, opened 2025-04-03
Remove `mode` arg usage in `create_array`

kushaangupta:
Quickstart tutorial code gave an error upon incorrect usage of mode argument in `zarr.create_array()`. So I tried removing it in this fix. Let me know if same is needed for `zarr.open_array()`.
It was interesting to see that `zarr.storage.ZipStore()` still takes mode argument.


d-v-b:
thanks for this contribution @kushaangupta, this looks good, we just need release notes. Something like "Remove the ``mode`` keyword argument from usage of ``create_array`` in the quickstart docs." would be good.

PR #2962, opened 2025-04-07
(fix): use `typesize` on `Blosc` codec

### LINKED ISSUE #2171: Poor blosc compression ratios compared to v2
### Zarr version

v3

### Numcodecs version

n/a

### Python Version

n/a

### Operating System

n/a

### Installation

n/a

### Description

While playing around with v3 a bit, I noticed that the blosc codec wasn't compressing my sample data as well as I expected to, and I have confirmed after multiple comparisons that I am getting different results between v3 and v2, in some cases v3 blosc compressed chunks end up being 10-20x larger than in v2 (see below).

### Steps to reproduce

```python
import numpy as np
import zarr
import zarr.v2 as zarr_v2
import numpy as np
from numcodecs import Blosc

a = np.random.rand(1000000)

# v3
codecs = [zarr.codecs.BytesCodec(), zarr.codecs.BloscCodec()]
z3 = zarr.array(a, chunks=(10000), codecs=codecs)
v3_size = len(z3.store_path.store._store_dict["c/0"]) # 75136

# v2
compressor = Blosc("zstd") # Should be equivalent to default settings for v3 Blosc codec
z2 = zarr_v2.array(a, chunks=z3.chunks, compressor=compressor)
v2_size = len(z2.store["0"]) 

print(f"v2 compressed chunk size: {v2_size}") 
print(f"v3 compressed chunk size: {v3_size}") 
# v2 compressed chunk size: 70113
# v3 compressed chunk size: 75136
```

The difference isn't huge in this case, but it can be much more noticeable in others. For example:
```python
b = np.arange(1000000)
z3[:] = b
v3_size = len(z3.store_path.store._store_dict["c/0"]) 

z2[:] = b
v2_size = len(z2.store["0"]) 

print(f"v2 compressed chunk size: {v2_size}") 
print(f"v3 compressed chunk size: {v3_size}") 

# v2 compressed chunk size: 1383
# v3 compressed chunk size: 11348
```

### Cause and Possible Solution
In numcodecs, the blosc compressor is able to improve compression ratios by inferring the item size through the input buffer's numpy array dtype. But in v3, the blosc codec is implemented as a `BytesBytesCodec` and requires each chunk to be fed as bytes on encode (hence, `BytesCodec()` is required in the list of codecs in my example) and thus numcodecs infers an item size of 1.

A simple fix for this is to make the following change in `blosc.py`:
```python
    async def _encode_single(
        self,
        chunk_bytes: Buffer,
        chunk_spec: ArraySpec,
    ) -> Buffer | None:
        # Since blosc only support host memory, we convert the input and output of the encoding
        # between numpy array and buffer
        return await to_thread(
            lambda chunk: chunk_spec.prototype.buffer.from_bytes(
                self._blosc_codec.encode(chunk.as_numpy_array().view(chunk_spec.dtype))
            ),
            chunk_bytes,
```
### Thoughts
I am still just getting started with v3, but this has made me curious about one thing. Why is the blosc codec implemented as a `BytesBytesCodec` rather than as an `ArrayBytesCodec`, considering that it can accept (and is optimized for) numpy array input? Although the above solution does work, because I need to include the `BytesCodec` first when specifying my codecs in v3, it essentially first encodes each chunk into bytes, then decodes it back into an array in its original dtype, making the bytes codec effectively a pointless noop in this case.

### LINKED ISSUE #2766: Difference in behavior between 2.x and 3.x using identical compressor settings
### Zarr version

3.0.1

### Numcodecs version

0.15.0

### Python Version

3.13

### Operating System

Linux

### Installation

Using pip into virtual environment

### Description

Writing the same data using identical compressor settings in Zarr-Python 2.x and 3.x yields differences in compression results.

Using `zarr==2.18.4, numcodecs==0.13.1`, this produces two chunks each of size 1.7 KiB on disk. The metadata is:

```json
{
    "attributes": {},
    "chunk_grid": {
        "chunk_shape": [
            64,
            64,
            2
        ],
        "separator": "/",
        "type": "regular"
    },
    "chunk_memory_layout": "C",
    "compressor": {
        "codec": "https://purl.org/zarr/spec/codec/blosc/1.0",
        "configuration": {
            "blocksize": 0,
            "clevel": 1,
            "cname": "lz4",
            "shuffle": 1
        }
    },
    "data_type": "<f8",
    "extensions": [],
    "fill_value": 0.0,
    "shape": [
        64,
        64,
        4
    ]
}
```

Using `zarr==3.0.1, numcodecs==0.15.0`, this produces two chunks of size 30 KiB.

```json
{
  "shape": [
    64,
    64,
    4
  ],
  "data_type": "float64",
  "chunk_grid": {
    "name": "regular",
    "configuration": {
      "chunk_shape": [
        64,
        64,
        2
      ]
    }
  },
  "chunk_key_encoding": {
    "name": "default",
    "configuration": {
      "separator": "/"
    }
  },
  "fill_value": 0.0,
  "codecs": [
    {
      "name": "bytes",
      "configuration": {
        "endian": "little"
      }
    },
    {
      "name": "blosc",
      "configuration": {
        "typesize": 8,
        "cname": "lz4",
        "clevel": 1,
        "shuffle": "shuffle",
        "blocksize": 0
      }
    }
  ],
  "attributes": {},
  "zarr_format": 3,
  "node_type": "array",
  "storage_transformers": []
}
```

### Steps to reproduce

**MVCE**

Run the following code in a 2.x and 3.x environment and inspect contents of `/tmp/foo.zarr`.

```python
import numpy as np

import zarr

store = "/tmp/foo.zarr"
shape = (64, 64, 4)
chunks = (64, 64, 2)
dtype = np.float64

cname = "lz4"
clevel = 1

if zarr.__version__[0] == "2":
    import numcodecs.blosc

    compressor = numcodecs.Blosc(
        cname=cname,
        clevel=clevel,
        shuffle=numcodecs.blosc.SHUFFLE,
    )

    za = zarr.open(
        store,
        mode="w",
        zarr_version=3,
        shape=shape,
        chunks=chunks,
        dtype=dtype,
        compressor=compressor,
    )
else:
    import zarr.codecs

    compressors = zarr.codecs.BloscCodec(
        cname=cname,
        clevel=clevel,
        shuffle=zarr.codecs.BloscShuffle.shuffle,
    )

    za = zarr.create_array(
        store,
        shape=shape,
        chunks=chunks,
        dtype=dtype,
        zarr_format=3,
        compressors=compressors,
    )

arr = np.arange(np.prod(shape)).reshape(shape)
za[:] = arr
```

### Additional output

_No response_

ilan-gold:
Fixes #2766 and fixes #2171
TODO:


normanrz (approved):

ilan-gold:
Apologies did not mean for you guys to do an immediate review, will keep that in mind next time, this was mostly to remind myself to finish up :)

d-v-b:
no worries, I was trigger-happy here

tests/test_codecs/test_blosc.py
```diff
@@ -54,3 +57,29 @@ async def test_blosc_evolve(store: Store, dtype: str) -> None:
         assert blosc_configuration_json["shuffle"] == "bitshuffle"
     else:
         assert blosc_configuration_json["shuffle"] == "shuffle"
+
+
+async def test_typesize() -> None:
+    a = np.arange(1000000)
```
dstansby, line 63:
```suggestion
    a = np.arange(2**16, dtype=np.uint16)
```
As a thought, worth explicitly specifying the data type (and making the data smaller)? Don't know if it will fix the windows issue, but I think worth doing anyway os there's a concrete bytesize, and perhaps using integer data type will help with linux/windows because perhaps they have different floating point implementations (although that's wild speculation on my part...)

ilan-gold: `arange` is default `uint64` so I'll push something with that added.

dstansby:
> This would indicate to me the typesize is being incorrectly encoded but I (a) don't know why and (b) don't know what the @ means.

Weird... I'm guessing that would be an upstream numcodecs issue/fix, so we could probably cut our losses here and just xfail the test on windows for now.

ilan-gold:
Ok @dstansby great call - it looks like it was just being explicit, there must be different behavior on windows for that version.  There is a warning in the documentation https://numpy.org/doc/stable/reference/generated/numpy.arange.html but I figured we hadn't actually hit any of those conditions.

dstansby (approved):

Issue #2925, opened 2025-03-21
Add `Buffer.as_buffer_like` method.


kylebarron:
The core Buffer class represents a block of memory in either CPU or GPU memory.

There's no current method on `Buffer` to expose the memory via the buffer protocol without _always_ forcing a copy. The closest thing is the `as_array_like` method, which returns an `ArrayLike`, but there's no way to get a `BytesLike` from it.

So what I want is 
```py
    @abstractmethod
    def as_bytes_like(self) -> collections.abc.Buffer:
```
(https://github.com/zarr-developers/zarr-python/issues/2814 tracks the unification of `BytesLike` and `collections.abc.Buffer` types)

In practice, this means that the obstore integration (#1661) won't need to force a memory copy when uploading data.

ref https://github.com/zarr-developers/zarr-python/pull/1661#discussion_r2007656426, cc @TomAugspurger

tomwhite:
BTW there's a typo in the title; it should be 'Add `Buffer.as_bytes_like` method'.

Also, this will be fixed in #2972.

kylebarron:
I called it "buffer" like to refer to the buffer protocol, hence the name in `collections.abc.Buffer`.

carloshorn:
I see that the `as_buffer_like` method now exists. However, I was wondering why the Zarr `Buffer` itself does not implement the `__buffer__` method to become a `collections.abc.Buffer`.

Here a simple demonstration:
```python
class MyBuffer(zarr.core.buffer.cpu.Buffer):
    def __buffer__(self, flags: int) -> memoryview:
        return self.as_buffer_like()

bfr_old = zarr.core.buffer.cpu.Buffer.from_bytes(b"abc")
bfr_new = MyBuffer.from_bytes(b"abc")

#memoryview(bfr_old)  # raises TypeError
memoryview(bfr_new)  # works fine
```

I know that this is Python >= 3.12, so the `as_buffer_like` is likely still relevant. In general, I would feel comfortable if the Zarr buffer were also a Python buffer.

d-v-b:
> if the Zarr buffer were also a Python buffer.

[...]

PR #2972, opened 2025-04-08
Avoid memory copy in obstore write

tomwhite:
This removes an unnecessary memory copy when writing using obstore, similar to #2944

TODO:

### DIFF
--- a/src/zarr/storage/_obstore.py
+++ b/src/zarr/storage/_obstore.py
@@ -7,6 +7,8 @@ from collections import defaultdict
 from collections.abc import Iterable
 from typing import TYPE_CHECKING, Any, TypedDict
 
+import numpy as np
+
 from zarr.abc.store import (
     ByteRequest,
     OffsetByteRequest,
@@ -145,7 +147,7 @@ class ObjectStore(Store):
 
         self._check_writable()
 
-        buf = value.to_bytes()
+        buf = value.as_numpy_array().view(np.uint8)
         await obs.put_async(self.store, key, buf)
 
     async def set_if_not_exists(self, key: str, value: Buffer) -> None:
@@ -153,7 +155,7 @@ class ObjectStore(Store):
         import obstore as obs
 
         self._check_writable()
-        buf = value.to_bytes()
+        buf = value.as_numpy_array().view(np.uint8)
         with contextlib.suppress(obs.exceptions.AlreadyExistsError):
             await obs.put_async(self.store, key, buf, mode="create")
 


src/zarr/storage/_obstore.py
```diff
@@ -145,15 +147,15 @@ async def set(self, key: str, value: Buffer) -> None:
 
         self._check_writable()
 
-        buf = value.to_bytes()
+        buf = value.as_numpy_array().view(np.uint8)
```
jakirkham, line 150: Is it worthwhile to make this a standard method? This would make it easier to use throughout the codebase

Some possible names:

* `to_bytelike`
* `to_binary`
* `to_uint8s`
* ?

jakirkham (commented):
Thanks Tom! 🙏

Had a question below

src/zarr/storage/_obstore.py
d-v-b: i guess i'm also curious to know why we have a `to_bytes()` method that makes a copy, when `.as_numpy_array().view(np.uint8)` achieves the same effective result (an iterable of bytes) without a copy.

tomwhite: This was suggested in #2925, so I've added an `as_bytes_like` method.

d-v-b (approved):

src/zarr/storage/_obstore.py
jakirkham: `bytes` objects always own their memory. So creating a new `bytes` object means making a copy

If we prefer to move away from this approach, we could add a deprecation cycle to move from `to_bytes` to `as_bytes_like`

tomwhite: I don't think `FsspecStore` works with `BytesLike` since it didn't accept a `memoryview` object when I tried it. So we probably need to keep `to_bytes()`.

jakirkham: Think this would be [a good issue to raise with `fsspec`]( https://github.com/fsspec/filesystem_spec/issues/new ) if we can reproduce with it directly

If not, a Zarr issue would be welcome

Ideally these copies should be avoidable in the `FsspecStore` case as well

tomwhite: Opened https://github.com/fsspec/s3fs/issues/959

kylebarron: > Some possible names:

I'd also suggest `as_buffer_like`, because in my head "bytes" in Python is a type that always owns its own memory, while "buffer" doesn't necessarily own its memory, as in the buffer protocol or `collections.abc.Buffer`.

d-v-b:
@tomwhite anything new here or is it ready to go in?

tomwhite:
@d-v-b This is ready to go in I think.

PR #2965, opened 2025-04-07
Pin numcodecs to < 0.16 to fix v2

dstansby:
Fixes https://github.com/zarr-developers/zarr-python/issues/2963 - I think worth getting this in and doing a release straight away, since it unbreaks an installation with just `pip install zarr<3`

PR #2973, opened 2025-04-08
Add a test with up to date dependencies & fix zarr-python v2

dstansby:
Currently on the v2 branch the dependencies in the test are pinned in `requirements_*.txt` files at the root of the repository, and these pins are bumped by dependabot periodically (I think weekly?). This means when a new version of a dependency is released, there is a period where we don't test against it. This prevents us from catching issues such as in https://github.com/zarr-developers/zarr-python/pull/2965.

This PR adds a new workflow that creates a fresh conda environment, then pip installs `zarr-python`. This emulates how a user installs, and will automatically pick up the latest versions of dependencies.

Hopefully this will initially fail, because it will pull in `numcodecs` 0.16. Then I plan to introduce the (correct...) pin, and hopefully that will fix the new test.

EDIT: everything seems to have worked, so I think this is good to review

### DIFF
--- a/.github/workflows/latest-deps.yml
+++ b/.github/workflows/latest-deps.yml
@@ -25,6 +25,6 @@ jobs:
         ZARR_V3_EXPERIMENTAL_API: 1
         ZARR_V3_SHARDING: 1
       run: |
-        python -m pip install pytest pytest-cov pytest-timeout
+        python -m pip install pytest pytest-cov pytest-timeout msgpack-python
         python -m pip install .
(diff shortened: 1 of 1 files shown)


dstansby:
Good news, the new CI job fails as expected.

dstansby:
🎉 and with the correct numcodecs pin, the latest deps job passes now.

dstansby:
Going to self merge and do a release to unbreak v2

PR #2980, opened 2025-04-12
Ignore stale children when reconsolidating metadata

### LINKED ISSUE #2921: zarr.consolidated_metadata should not use previously consolidated (stale) metadata
### Zarr version

3.0.5

### Numcodecs version

v0.15.1

### Python Version

3.11

### Operating System

Mac

### Installation

conda/pip

### Description

We're noticing some ugly behavior when re-consolidating a store with pre-existing consolidated metadata. If a child node in the store already has consolidated metadata in it, any subsequent consolidation attempts will fail to find children that were not present before the consolidation. Because there is no way to invalidate the consolidated metadata, it is not possible to consolidate the entire store.

Note: In https://github.com/zarr-developers/zarr-python/issues/2920, we also noticed a distinct but related issue where the behavior of `consolidate_metadata` has changed to honor the `path` argument. 

### Steps to reproduce

```python

root = zarr.group(store=store, zarr_format=3)
root.create_group("foo")
zarr.consolidate_metadata(store, path="foo")
root.create_group("foo/bar/spam")
zarr.consolidate_metadata(store)

root = zarr.open_consolidated(store, zarr_format=3)
root.members(max_depth=100)
# (('foo', <Group memory://140600666693888/foo>),)

# expected
# (('foo', <Group memory://140600662927488/foo>),
#  ('foo/bar', <Group memory://140600662927488/foo/bar>),
#  ('foo/bar/spam', <Group memory://140600662927488/foo/bar/spam>))
```

TomAugspurger:
This fixes a bug when recreating consolidated metadata, when one of the child groups picked up had (stale) consolidated metadata.

In `consolidate_metadata`, we already ignore the `group.metadata.consolidated_metadata` for the root group. The fix is to *also* ignore the consolidated metadata for any child nodes loaded by `.members()`.

For now, I've made this a new public parameter in `.members()`. It might be worth keeping it private (by making a `_members_private` method that takes this parameter). I don't immediately see any other use case for this parameter.

Closes #2921 

TODO:

### DIFF
--- a/changes/2921.bugfix.rst
+++ b/changes/2921.bugfix.rst
@@ -1,0 +1,1 @@
+Ignore stale child metadata when reconsolidating metadata.
--- a/src/zarr/api/asynchronous.py
+++ b/src/zarr/api/asynchronous.py
@@ -193,7 +193,10 @@ async def consolidate_metadata(
     group = await AsyncGroup.open(store_path, zarr_format=zarr_format, use_consolidated=False)
     group.store_path.store._check_writable()
 
-    members_metadata = {k: v.metadata async for k, v in group.members(max_depth=None)}
+    members_metadata = {
+        k: v.metadata
+        async for k, v in group.members(max_depth=None, use_consolidated_for_children=False)
+    }
     # While consolidating, we want to be explicit about when child groups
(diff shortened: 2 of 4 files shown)


d-v-b (approved):
thanks tom!

Issue #2989, opened 2025-04-16
[v3] Function `zeros_like` ignores overrides in keyword arguments (e.g. `chunks`, `dtype`).


tasansal:
### Zarr version

3.0.6

### Numcodecs version

0.15.1

### Python Version

3.13

### Operating System

Mac

### Installation

Using uv into virtual env

### Description

The `zeros_like` function doesn't honor the overrides. e.g. `chunks` and `dtype`.

### Steps to reproduce

```python
import zarr

root = zarr.open("test.zarr", mode="w")
arr = root.create_array(name="array", shape=(256, 512, 512), chunks=(256, 512, 512), dtype="float32")

arr2 = root.zeros_like(data=arr, name="clone", chunks=(16, 16, 16), dtype="uint32")
```

`arr2.chunks` should show (16, 16, 16) but it keeps original (256, 512, 512)

### Additional output

Here is metadata for that `clone` array.

```json
{
  "shape": [
    256,
    512,
    512
  ],
  "data_type": "float32",
  "chunk_grid": {
    "name": "regular",
    "configuration": {
      "chunk_shape": [
        256,
        512,
        512
      ]
    }
  },
  "chunk_key_encoding": {
    "name": "default",
    "configuration": {
      "separator": "/"
    }
  },
  "fill_value": 0.0,
  "codecs": [
    {
      "name": "bytes",
      "configuration": {
        "endian": "little"
      }
    },
    {
      "name": "zstd",
      "configuration": {
        "level": 0,
        "checksum": false
      }
    }
  ],
  "attributes": {},
  "zarr_format": 3,
  "node_type": "array",
  "storage_transformers": []
}
```

d-v-b:
@tasansal thanks for this bug report. I think I fixed this in https://github.com/zarr-developers/zarr-python/pull/2992, please let me know if that PR doesn't resolve this

BrianMichell:
@d-v-b Thanks for the quick response! I did some quick testing and it appears that the fix behaves as expected, but unfortunately our full test suite currently keels over due to #2998.

PR #2992, opened 2025-04-17
fix:*-like creation routines take kwargs

d-v-b:
In main, array creation routines that take an existing array, like `full_like`, `zeros_like`, etc,  don't allow users to override certain array properties (like shape and dtype) via keyword arguments. E.g., `full_like(arr, shape=(10,))` will ignore the `shape` keyword argument and produce an array with the same shape as `arr`.

This PR makes these array creation routines sensitive to keyword arguments like `shape`, `dtype`, `chunks`, etc. 

closes https://github.com/zarr-developers/zarr-python/issues/2989

### DIFF
--- a/src/zarr/api/asynchronous.py
+++ b/src/zarr/api/asynchronous.py
@@ -108,10 +108,10 @@ def _get_shape_chunks(a: ArrayLike | Any) -> tuple[ChunkCoords | None, ChunkCoor
     return shape, chunks
 
 
-def _like_args(a: ArrayLike, kwargs: dict[str, Any]) -> dict[str, Any]:
+def _like_args(a: ArrayLike) -> dict[str, object]:
     """Set default values for shape and chunks if they are not present in the array-like object"""
 
-    new = kwargs.copy()
+    new: dict[str, object] = {}
 
     shape, chunks = _get_shape_chunks(a)
     if shape is not None:
@@ -1077,7 +1077,7 @@ async def empty(
     shape: ChunkCoords, **kwargs: Any
 ) -> AsyncArray[ArrayV2Metadata] | AsyncArray[ArrayV3Metadata]:
     """Create an empty array with the specified shape. The contents will be filled with the
-    array's fill value or zeros if no fill value is provided.
+    specified fill value or zeros if no fill value is provided.
 
     Parameters
     ----------
@@ -1092,8 +1092,9 @@ async def empty(
     retrieve data from an empty Zarr array, any values may be returned,
     and these are not guaranteed to be stable from one access to the next.
     """
-
-    return await create(shape=shape, fill_value=None, **kwargs)
+    if "fill_value" not in kwargs:
+        kwargs["fill_value"] = None
+    return await create(shape=shape, **kwargs)
 
 
 async def empty_like(
@@ -1120,8 +1121,10 @@ async def empty_like(
     retrieve data from an empty Zarr array, any values may be returned,
     and these are not guaranteed to be stable from one access to the next.
     """
-    like_kwargs = _like_args(a, kwargs)
-    return await empty(**like_kwargs)
+    like_kwargs = _like_args(a) | kwargs
+    if isinstance(a, (AsyncArray | Array)):
+        kwargs.setdefault("fill_value", a.metadata.fill_value)
+    return await empty(**like_kwargs)  # type: ignore[arg-type]
 
 
 # TODO: add type annotations for fill_value and kwargs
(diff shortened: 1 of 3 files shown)


d-v-b:
requesting review from @jhamman since I think you touched this code most recently

src/zarr/api/asynchronous.py
```diff
@@ -109,10 +109,10 @@ def _get_shape_chunks(a: ArrayLike | Any) -> tuple[ChunkCoords | None, ChunkCoor
     return shape, chunks
 
 
-def _like_args(a: ArrayLike, kwargs: dict[str, Any]) -> dict[str, Any]:
+def _like_args(a: ArrayLike) -> dict[str, object]:
```
TomAugspurger, line 112: Why the change to `object`? And if you're feeling eager, this could maybe be a typed dict?

```
class LikeArgs(TypedDict):
    shape: tuple[int, ...]
    chunks: tuple[int, ...]
    dtype: np.typing.DTypeLike
    order: ...
```


but that's probably out of scope for this PR.

src/zarr/api/asynchronous.py
```diff
@@ -1120,8 +1119,10 @@ async def empty_like(
     retrieve data from an empty Zarr array, any values may be returned,
     and these are not guaranteed to be stable from one access to the next.
     """
-    like_kwargs = _like_args(a, kwargs)
-    return await empty(**like_kwargs)
+    like_kwargs = _like_args(a) | kwargs
+    if isinstance(a, (AsyncArray | Array)):
+        like_kwargs.setdefault("fill_value", a.metadata.fill_value)
```
TomAugspurger, line 1124: To make sure I understand the precedence, it's

1. User keyword arguments (via `| kwargs`)
2. `_like_kwargs(a)`
3. `fill_value` via `a.metadata.fill_value`

? It'd be worth confirming we have each of those cases covered with tests.

src/zarr/api/asynchronous.py
```diff
@@ -1120,8 +1119,10 @@ async def empty_like(
...
+    return await empty(**like_kwargs)  # type: ignore[arg-type]
```
TomAugspurger, line 1125: Why the type ignore here? I guess the `LikeArgs` typed dict might solve that.

tests/test_api.py
```diff
@@ -68,6 +68,81 @@ def test_create(memory_store: Store) -> None:
         z = create(shape=(400, 100), chunks=(16, 16.5), store=store, overwrite=True)  # type: ignore [arg-type]
 
 
+LikeFuncName = Literal["zeros_like", "ones_like", "empty_like", "full_like", "open_like"]
+
+
+@pytest.mark.parametrize("func_name", get_args(LikeFuncName))
```
TomAugspurger, line 74: Maybe there's a reason to structure it like this, but my usual preference is to use functions instead of names. So the actual params would be `[zarr.api.asynchronous.zeros_like, zarr.api.asynchronous.one_like, ...`].

src/zarr/api/asynchronous.py
d-v-b: and re: the typeddict thing, yeah that would be nice. typeddictification across the codebase is on my to-do list.

src/zarr/api/asynchronous.py
d-v-b: Like many of the v2-era creation routines, `empty` has a nearly [untyped signature](https://github.com/zarr-developers/zarr-python/blob/f6742368229b8817a50335c28bb19a10e7a778e9/src/zarr/api/synchronous.py#L1129). The payoff for adding static typing to this particular invocation of `empty` is low until we add a fully typed signature to `empty` itself, which I'm not interested in doing in this PR

tests/test_api.py
d-v-b: good catch, I don't remember why I used strings, but in https://github.com/zarr-developers/zarr-python/pull/2992/commits/19b1ce2650b69d3d879ef79e00cbc9a0ff3fe95f i switch to iterating over functions directly

src/zarr/api/asynchronous.py
d-v-b: the first two cases were handled already, and I added fill_value testing in https://github.com/zarr-developers/zarr-python/pull/2992/commits/d29f68275505596b7ac14c1445b61a68c748d142

d-v-b:
our upstream tests are failing because fsspec and s3fs tips are incompatible. this failure has nothing to do with this PR.

jhamman (approved):

Issue #2991, opened 2025-04-16
is dot syntax gone now?


christine-e-smit:
### Zarr version

v3.0.6

### Numcodecs version

0.15.1

### Python Version

3.13.3

### Operating System

M<ac

### Installation

using conda

### Description

I don't actually know what you call this, but I'm calling it dot syntax for accessing arrays. With zarr version 2.18.4, I can do the following:

```python
import zarr
import numpy as np
z = zarr.open("two.zarr",mode="w")
d = z.create_dataset("d",shape=(2),dtype=np.int32)
# this works!
z.d
```

But with zarr version 3.0.6, the following throws an exception:
```python
import zarr
import numpy as np
z = zarr.open("three.zarr",mode="w")
d = z.create_array("d",shape=(2),dtype=np.int32)
# this now throws the exception "AttributeError: 'Group' object has no attribute 'd'"
z.d
```

Is this expected? Unexpected? I didn't see anything in the 3.0 Migration Guide, but it wouldn't surprise me if I missed something.

### Steps to reproduce

```python
import zarr
import numpy as np
z = zarr.open("three.zarr",mode="w")
d = z.create_array("d",shape=(2),dtype=np.int32)
# this now throws the exception "AttributeError: 'Group' object has no attribute 'd'"
z.d
```

christine-e-smit:
Thank you for the clarification. I've submitted a pull request with some suggested language for the documentation.

rabernat:
@christine-e-smit - did you find this syntax valuable?

For context, IIRC it was removed because many IDEs (e.g. JupyterLab, VCCode) will introspect python objects to provide auto-completion. With this dot syntax, that would actually trigger listing of groups, reading metadata, etc. operations with significant side effects and performance implications.

christine-e-smit:
Interesting! I never personally ran into the IDE issue in Visual Studio Code, but I can see how it could be painful. I think I've used both `g.x` and `g["x"]` syntax without thinking terribly deeply about it, so I don't think I'll miss the "." syntax. I'm more concerned that finding all the instances where we used "." syntax is going to be tricky because it's not something you can just grep for. But if it's causing that many issues with IDEs, then I understand why you've dropped it.

PR #2997, opened 2025-04-17
updated migration documention to say that '.' syntax is no longer all…

christine-e-smit:
…owed for variables in groups

In #2991, @d-v-b said that "." syntax is gone for group members in v3.0. This is my suggested update to the 3.0 migration documentation.


d-v-b (approved):
thanks for fixing this!

Issue #1758, opened 2024-04-05
[v3] Request coalescing


jhamman:
Various threads have recently highlighted that supporting request coalescing would be a nice feature to add to the v3 effort. This may be particularly impactful with sharding coming online. We need to decide if we should handle this and where (at the store level or elsewhere).

aldenks:
I'm interested in taking this on for coalescing requests to chunks within the same shard.  (side question: are there other types of coalescing in play?)

I did some profiling of slower than I'd anticipated reading from a [zarr with many small chunks per shard](https://dynamical.org/catalog/noaa-gefs-forecast-35-day/) and found that [this loop in `ShardingCodec._decode_partial_single`](https://github.com/zarr-developers/zarr-python/blob/main/src/zarr/codecs/sharding.py#L510-L518) was the culprit. 
```python
            for chunk_coords in all_chunk_coords:
                chunk_byte_slice = shard_index.get_chunk_slice(chunk_coords)
                if chunk_byte_slice:
                    chunk_bytes = await byte_getter.get(
                        prototype=chunk_spec.prototype,
                        byte_range=RangeByteRequest(chunk_byte_slice[0], chunk_byte_slice[1]),
                    )
                    if chunk_bytes:
                        shard_dict[chunk_coords] = chunk_bytes
```

I had Claude wip up coalescing for those requests and while the code is a bit of an AI hot mess it's working to get way higher read throughput (went from roughly 4MB/s to 70MB/s aka saturating my current network).

-----
In terms of solutions, two obvious ones to me are:
- replace the serial for loop with a `concurrent_map`. 
- do actual coalescing to reduce request count.

Coalescing seems to me like the ideal longer term solution (less request overhead, lower backend cost overhead if paying for requests) and a decent implementation shouldn't be too hard. The key questions I see if we go this route are:
- is there a max size we want to stop coalescing at or do we let it be naturally bounded by the shard size?
- what's the maximum gap of unnecessary bytes to coalesce over? Should that be a config option? (one point of reference is rust's object_store which [defaults to 1MiB](http://docs.rs/object_store/latest/src/object_store/util.rs.html#92))
- do we make concurrent or serial requests to each of the coalesced groups? My default would be concurrent, but this is happening within an outer concurrent_map in `ArrayBytesCodecPartialDecodeMixin.decode_partial` so you're potentially not respecting the `async.concurrency` config value if you make a nested concurrent_map call.


----
[...]

PR #3004, opened 2025-04-22
Optimize partial shard reads

aldenks:
This PR optimizes reading more than one, but not all, chunks from a shard.  Towards #1758 and #3953.

Optimizes partial reads from sharded arrays by replacing per-chunk serial `Store.get()` calls with a single `Store.get_ranges()` call per shard. This lets stores coalesce nearby byte ranges and fetch chunk payloads concurrently while preserving existing behavior for full-shard reads (no need to fetch the chunk index). `Store.get_ranges()`  was recently introduced in #3925.

To be explicit, in my mind the optimizations in this have two goals:
1. Speed up the time it takes to read many chunks from a single shard. Most of the benefit on this point comes from concurrently rather than serially fetching bytes. With especially small chunks coalescing can also help reduce request overhead.
2. Reduce the total count of calls to the underlying store. If the store (e.g. object storage) has a per operation charge this results in reduced costs.

Mostly these goals aren't in conflict. The maximum coalesced size option provides a knob to ensure we aren't coalescing requests into a single very large request which would be faster to make as multiple concurrent calls.
   

TODO:


aldenks:
I'm curious what folks think of this and looking for input! One particular design question to note: this adds a nested concurrent_map (there's already one in the codec pipeline's read/read_batch) so it's technically possible to have  concurrency of `async.concurrency**2`. That said, I think we want _some_ concurrency here.

d-v-b:
@aldenks thanks for working on this, it's a great optimization direction to pursue. 

That being said, since this is a performance optimization, it might be helpful to get some metrics for how much perf we get with these changes, in terms of speed and also latency / number of requests. If reducing requests is the goal, you could consider implementing a `WrapperStore` as a demonstration or test fixture that tracks the number of times its methods are invoked, and use that to show that your approach reduces requests. But any other demonstration would work too.

As for the specific code changes (and the nested concurrent map), nothing jumps out to me as problematic, but that might just be my ignorance of async python and the sharding code 😁

aldenks:
I'm working on a couple performance tests as Davis suggested above. Trying [airspeed velocity](https://asv.readthedocs.io/en/latest/index.html) which is new to me but looks useful.

dcherian:
Something like this might be more useful than asv here. A regression test of this form would also be great, if it can be made to work

https://github.com/pydata/xarray/blob/e37bfbf1bff63476de9e9d95eedbbd66d864f588/xarray/tests/test_backends.py#L3415

aldenks:
Got pulled away after pushing those changes but before describing them -- ill update this tomorrow with a description of what im seeing but here's the plot

zarr-metrics-by-statement

aldenks:
alright!
- Tests added that check for coalesced (reduced count) of calls to store
- Profiled execution time and store request count

Let me know what yall think of these changes. Once i get the thumbs up ill work on the remaining todos in the description.

## Tests
Added `test_sharding_multiple_chunks_partial_shard_read` which is parameterized to test both with and with out coalescing and I'm seeing the behavior I'd expect. e.g.:
```
    if coalesce_reads:
        # 2 shard index requests + 2 coalesced chunk data byte ranges (one for each shard)
        assert store_mock.get.call_count == 4
    else:
        # 2 shard index requests + 6 chunks
        assert store_mock.get.call_count == 8
```

## Profiling
I added a slow test which reads a few subsets of a large-ish shard containing small-ish chunks under a range of settings. See `test_partial_shard_read_performance` for details.

The test is currently skipped and I don't necessarily think it should be a test, probably belongs in `bench` or I'm happy to remove now it's done its job.

To recreate the figures below:
- Check out the commit before the optimization changes https://github.com/zarr-developers/zarr-python/pull/3004/commits/91fbdf9299bef95624459e7722259cdaf04f5afe and run `hatch env run --env test.py3.12-2.1-optional -- pytest -s tests/test_codecs/test_sharding.py::test_partial_shard_read_performance --run-slow-hypothesis`. Then checkout the latest commit https://github.com/zarr-developers/zarr-python/pull/3004/commits/9f6727da283d60cd6187c2876271fa59e76db23b and run the same command again. (yes, this is what airspeed velocity is for but i opted for simpler for now.)
- Run the plotting code below

<details><summary>Plotting code</summary>

```python
import pandas as pd
import matplotlib.pyplot as plt
import seaborn as sns

df_coalesce = pd.read_json(
    "zarr-python-partial-shard-read-performance-with-coalesce.json"
)
df_no_coalesce = pd.read_json(
    "zarr-python-partial-shard-read-performance-no-coalesce.json"
)
df_no_coalesce["coalesce_max_gap"] = "Before optimization"
df = pd.concat([df_no_coalesce, df_coalesce])

# Define statements and metrics
statements = df["statement"].unique()
metrics = ["time", "store_get_calls"]
metric_titles = ["Execution Time", "Number of Store Get Calls"]
metric_ylabels = ["Time (s)", "Number of Calls"]

# Create labels for coalesce values
coalesce_values = df["coalesce_max_gap"].unique()
coalesce_labels = [
    x if isinstance(x, str) else "Disabled" if x == -1 else f"{x/1024/1024:.0f}MiB"
    for x in coalesce_values
]

# Get unique get_latency values
latency_values = df["get_latency"].unique()

# For each unique get_latency value, create a separate figure
for latency in latency_values:
    latency_df = df[df["get_latency"] == latency]

    max_time = latency_df["time"].max()
    max_store_get_calls = latency_df["store_get_calls"].max()

    plt.figure(figsize=(16, 12))
    plt.suptitle(
        "Performance when reading a subset of a shard\n"
        "Execution time and count of store `get` calls for 3 array access patterns\n"
        "Array and shard shape: (512, 512, 512), chunk shape: (64, 64, 64)\n"
        "Store get latency: local ssd"
        + ("" if latency == 0 else f" + {latency * 1000:.0f}ms"),
        fontsize=14,
    )

    # One row per statement, two columns: time and store_get_calls
    for i, statement in enumerate(statements):
        statement_df = latency_df[latency_df["statement"] == statement]

        # Plot for time
        plt.subplot(3, 2, i * 2 + 1)
        sns.barplot(
            data=statement_df,
            x="concurrency",
            y="time",
            hue="coalesce_max_gap",
            hue_order=coalesce_values,
            palette="muted",
            errorbar=None,
        )
        plt.title(f"{statement} - {metric_titles[0]}")
        plt.xlabel("Concurrency")
        plt.ylabel(metric_ylabels[0])
        plt.ylim(0, max_time * 1.1)
        plt.legend(title="Coalesce Max Gap", labels=coalesce_labels)

        # Plot for store_get_calls
        plt.subplot(3, 2, i * 2 + 2)
        sns.barplot(
            data=statement_df,
            x="concurrency",
            y="store_get_calls",
            hue="coalesce_max_gap",
            hue_order=coalesce_values,
            palette="muted",
            errorbar=None,
        )
        plt.title(f"{statement} - {metric_titles[1]}")
        plt.xlabel("Concurrency")
        plt.ylabel(metric_ylabels[1])
        plt.ylim(0, max_store_get_calls * 1.1)
        plt.legend(title="Coalesce Max Gap", labels=coalesce_labels)

    plt.tight_layout()
    plt.savefig(f"zarr-metrics-by-statement-with-latency-{latency}s.png", dpi=300)
    plt.show()
    plt.close()
```

</details> 


#### Local SSD
The results shown in this figure read from my local SSD. Before this optimization and when run on my ssd, the time to read chunks serially in the loop in `ShardingCodec._decode_partial_single` was about the same as the time to decode chunks (which was parallelized already within the `self.codec_pipeline.read` call).

zarr-metrics-by-statement-latency-0 0


#### Simulated Object Storage
This optimization is most important reading from higher latency storage. To simulate this I added 10ms of latency to each get call. (I read that with specific networking setups you can get down to 1ms latency from S3, I also regularly see >100ms in practice, so 10ms feels fair)

zarr-metrics-by-statement-latency-0 01


### Takeaways
None of these are particularly surprising, but my takeaways are:

- Most of the execution time benefit comes from adding concurrency
- We can reduce store request count by enabling coalescing and get equal or better performance to adding concurrency alone
- If we coalesce over gaps that are too large execution time increases a bit as we're requesting more data we don't need. 1MiB seems like a decent middle ground as a maximum gap between chunks to coalesce.

JackKelly:
This is very cool!

On the topic of benchmarking: When reading compressed & sharded datasets, [LDeakin's `zarr_benchmarks`](https://github.com/LDeakin/zarr_benchmarks) suggests that `zarr-python` (without this PR) is quite slow. Note that the extent of the slowness isn't immediately obvious from a quick glance at the `zarr_benchmark` graphs, until you read the numbers above each bar in the bar charts! For example, in the "[round trip benchmark](https://github.com/LDeakin/zarr_benchmarks?tab=readme-ov-file#round-trip-benchmark)", `zarrs` takes 3 seconds to read a compressed and sharded dataset from a local SSD, whilst `zarr-python` takes 181 seconds!

Hopefully this PR will help `zarr-python` to catch up. Thank you, Alden!

If it's helpful, and if this PR is ready, then I could try running LDeakin's `zarr_benchmarks` with this PR.

aldenks:
Thanks @JackKelly! I'd be happy for more benchmarking of this. I took a look at LDeakin's zarr_benchmarks and I would  expect the changes in the PR to _not_ have any impact. All the read tests in there appear to read complete shards and this PR doesn't touch that path. If you're adding benchmarks I might advocate for one that reads subsets of multiple shards as I find that's a pretty common read pattern in practice.

JackKelly:
Noted! I've started making some notes for my plans for benchmarking [here](https://github.com/JackKelly/zarr_cloud_benchmarks). Feel free to comment!

JackKelly:
(Just a quick note to say that I'm afraid I'm now planning to experiment with storing NWP data in Parquet (using Polars)... so I might not get round to benchmarking Zarr readers soon, I'm sorry...)

aldenks:
Any maintainers, I don't necessarily need a detailed review at this point -- if yall let me know any open questions you have or give a thumbs up ill work on the remaining todos in the description.

aldenks:
I'm approaching having time to wrap up the TODOs in the description and still eager to get this merged

pyproject.toml
```diff
@@ -347,6 +347,7 @@ ignore = [
 [tool.mypy]
 python_version = "3.11"
 ignore_missing_imports = true
+mypy_path = "src"
```
aldenks: I needed to do this for the mypy run in pre-commit  to succeed when it was running on tests/test_config.py. Not sure if we want this at all, or if it should go in its own PR.

src/zarr/codecs/sharding.py
```diff
@@ -196,7 +198,9 @@ async def from_bytes(
 
     @classmethod
     def create_empty(
-        cls, chunks_per_shard: ChunkCoords, buffer_prototype: BufferPrototype | None = None
+        cls,
+        chunks_per_shard: ChunkCoords,
+        buffer_prototype: BufferPrototype | None = None,
```
aldenks: I'm leaving these formatting changes in since they were produced by the pre-commit run

src/zarr/codecs/sharding.py
```diff
@@ -483,39 +501,31 @@ async def _decode_partial_single(
...
-                    )
-                    if chunk_bytes:
-                        shard_dict[chunk_coords] = chunk_bytes
+            shard_dict_maybe = await self._load_partial_shard_maybe(
+                byte_getter, chunk_spec.prototype, chunks_per_shard, all_chunk_coords
+            )
+
+        if shard_dict_maybe is None:
+            return None
+        shard_dict = shard_dict_maybe
```
aldenks, line 528: Here's where the non-formatting changes start in this file

tests/test_codecs/test_sharding.py
```diff
@@ -197,6 +204,254 @@ def test_sharding_partial_read(
     assert np.all(read_data == 1)
 
 
+@pytest.mark.skip("This is profiling rather than a test")
+@pytest.mark.slow_hypothesis
+@pytest.mark.parametrize("store", ["local"], indirect=["store"])
+def test_partial_shard_read_performance(store: Store) -> None:
```
aldenks: Should I remove this? It's what was used to create the data for the performance plots

aldenks:
**Ready for review!**

at long last

The PR optimizes reading part of a shard in two ways:
1. Requests for nearby chunks in the same shard are combined into a single request to the store. This combining process respects two config options which define 1) the maximum gap in bytes between two chunks to combine over and 2) the maximum size of a combined group.
2. The resulting requests to the store are performed concurrently if there are > 1 groups within a shard after coalescing.

Motivation:
- speed, see profiling in [this comment above](https://github.com/zarr-developers/zarr-python/pull/3004#issuecomment-2845059934)
- reduced request count to stores, and therefore operation costs when using most object storage providers


A couple questions to draw attention to:
- The newly added concurrent_map over groups within a shard is nested within the existing concurrent_map in BatchedCodecPipeline.read_batch. This means one could have concurrency of up to `async.concurrency ^ 2`. I don't see a clean way to limit concurrency overall and the concurrency is the main source of the speedup. The default concurrency (10) is also pretty conservative imo so the extreme case of squaring it is perhaps ok?
- The defaults for the config options. Here's my thinking and experience with them.
  - max gap: see the bar plots in the performance table, its a tradeoff. The lower the store latency, the less gap you want and latencies are trending down so 1MiB feels like a fairly conservative default. It also is the [value](https://github.com/apache/arrow-rs-object-store/blob/main/src/util.rs#L92) used by rust's object_store in their [get_ranges](https://github.com/apache/arrow-rs-object-store/blob/cd931979bdb753fbbfca90db118dbbd93a4f1fdc/src/lib.rs#L659) function (and that value isn't configurable) and that's a very well tested in production library.
  - max group size. I added this to avoid the case where coalescing causes you to loose concurrency when reading from a large (~200MB+) shard and that lower concurrency reduces throughput. I didn't tune this value much, but from my experience and what I read online 100MiB is likely large enough to amorize out most of the per request overhead and get close to the maximum _per individual store request_ throughput.
  
  
Minor notes/questions
- Patch code cov is not 100%, but when i look at the report it looks like its 100% of the functional lines of code
- Is passing `fill_value=0` rather than `shard_spec.fill_value` to the [two](https://github.com/zarr-developers/zarr-python/blob/main/src/zarr/codecs/sharding.py#L459) [cases](https://github.com/zarr-developers/zarr-python/blob/main/src/zarr/codecs/sharding.py#L506) of `nd_buffer.create` in the sharding codec intentional? The overall behavior is correct currently because it gets filled later, but feels a bit like a bug in waiting.

aldenks:
@d-v-b who should I request a review from on this?

d-v-b:
@zarr-developers/python-core-devs who wants to check out this sweet PR

src/zarr/codecs/sharding.py
```diff
@@ -329,9 +335,18 @@ async def finalize(
         return await shard_builder.finalize(index_location, index_encoder)
 
 
+class _ChunkCoordsByteSlice(NamedTuple):
+    """Holds a chunk's coordinates and its byte range in a serialized shard."""
+
+    coords: ChunkCoords
```
d-v-b: Is `coords` the location of the chunk in the indexing space of the full array?

src/zarr/codecs/sharding.py
```diff
@@ -760,6 +771,115 @@ async def _load_full_shard_maybe(
             else None
         )
 
+    async def _load_partial_shard_maybe(
```
d-v-b, line 777: What would happen if we got rid of `_load_full_shard_maybe` and instead treated loading the full shard as a special case of loading some of the shard?

src/zarr/codecs/sharding.py
```diff
@@ -760,6 +771,115 @@ async def _load_full_shard_maybe(
...
+        byte_getter: ByteGetter,
+        prototype: BufferPrototype,
+        chunks_per_shard: ChunkCoords,
+        all_chunk_coords: set[ChunkCoords],
+    ) -> ShardMapping | None:
+        """
+        Read chunks from `byte_getter` for the case where the read is less than a full shard.
+        Returns a mapping of chunk coordinates to bytes or None.
+        """
+        shard_index = await self._load_shard_index_maybe(byte_getter, chunks_per_shard)
```
d-v-b, line 788: this will fetch the shard index every time. Should we push reading the shard index higher up in the stack, and have this function take the content of the index as a parameter? This might be out of scope for this PR.

d-v-b: (we could also build caching into the `_load_shard_index` function)

src/zarr/codecs/sharding.py
```diff
@@ -760,6 +771,115 @@ async def _load_full_shard_maybe(
...
+        if shard_index is None:
+            return None
```
d-v-b, line 790: what does this mean? when would the shard index be `None`?

d-v-b: i don't think your PR is adding these semantics btw, I'm just curious what this code path means

src/zarr/codecs/sharding.py
```diff
@@ -760,6 +771,115 @@ async def _load_full_shard_maybe(
...
+        shard_dicts = await concurrent_map(
+            [(group, byte_getter, prototype) for group in groups],
+            self._get_group_bytes,
+            config.get("async.concurrency"),
+        )
+
+        shard_dict: ShardMutableMapping = {}
+        for d in shard_dicts:
+            if d is None:
+                return None
```
d-v-b: why do we return `None` from this function if any one of the shard dicts is None? maybe a comment would help

d-v-b:
a high level comment -- the tests are all at the array level, which is effectively an integration test, testing how this functionality fits into the entire stack. It would be a lot more direct if we could do lower-level tests, e.g. create a model of a shard, compare the proposed reading plan against the reading plan we expect. This kind of test might not be easy to write in our codebase today, but if so we should identify things we need to change to bring us there.

(not a blocker for this PR if these tests are hard to write)

src/zarr/codecs/sharding.py
```diff
@@ -760,6 +771,115 @@ async def _load_full_shard_maybe(
...
+        Combine chunks from a single shard into groups that should be read together
+        in a single request to the store.
+
+        Respects the following configuration options:
+        - `sharding.read.coalesce_max_gap_bytes`: The maximum gap between
+          chunks to coalesce into a single group.
+        - `sharding.read.coalesce_max_bytes`: The maximum number of bytes in a group.
+        """
+        max_gap_bytes = config.get("sharding.read.coalesce_max_gap_bytes")
+        coalesce_max_bytes = config.get("sharding.read.coalesce_max_bytes")
```
d-v-b: lets not fetch parameters from a global mutable variable (the config) inside this function. better to define these as parameters of the function, and pass them in. we should get values out of the config as early as possible in this process, and after that it's just regular function parameters

d-v-b: we could make these keyword-only so that there's flexibility to change the configuration in the future

aldenks:
Thank you for the first review @d-v-b! I didn't get to this before going out of office but ill address your comments when im back in ~1 week.

tasansal:
this is great! With #3561 now creating sharded arrays is almost as fast as regular chunked arrays; but reads are still painfully slow and does too many requests even for sequential reads in the slow direction. Any idea if this PR will be pushed forward?

tasansal:
> this is great! With #3561 now creating sharded arrays is almost as fast as regular chunked arrays; but reads are still painfully slow and does too many requests even for sequential reads in the slow direction. Any idea if this PR will be pushed forward?

For example:

I am using this to create data on GCS:

```python
import zarr
from distributed import Client
import dask.array as da

client = Client(n_workers=8, threads_per_worker=2)

path = "gs://bucket/prefix/bench.zarr"
root = zarr.create_group(path, overwrite=True)

# Example: Create a sharded array
sharded = root.create_array(
    name="sharded",
    shape=(1355, 2456, 1501),
    shards=(256, 256, 256), 
    chunks=(32, 32, 32),
    dtype='float32',
)

chunked = root.create_array(
    name="chunked",
    shape=(1355, 2456, 1501),
    chunks=(128, 128, 128),
    dtype='float32',
)

data = da.random.normal(size=sharded.shape, chunks=(256, 256, 256))
```

```python
%%time
data.to_zarr(chunked)
```
gives 40.4s

```python
%%time
data.to_zarr(sharded)
```
gives 64s.

Which is somewhat reasonable.

However, the reads:
```python
%%time
root["chunked"][256]
```
9.22s

```python
%%time
root["sharded"][256]
```
17.7s

The same read with tensorstore on the sharded version is significantly faster.

tasansal:
Here is the tensorstore version:

```python
import tensorstore
import tensorstore as ts

spec = {
    "driver": "zarr3",
    "kvstore": {
        "driver": "gcs",
        "bucket": "bucket",
        "path": "prefix/bench.zarr/sharded"
    }
}

# Open the Zarr array (async, use .result() for blocking)
array = ts.open(spec).result()
```

then 

```python
%%time
array[256].read().result()
```
results in 2.78s

aldenks:
Thanks for the encouragement! this has been close to getting back to the top of my list and I'll get working on it this week

src/zarr/codecs/sharding.py
aldenks: good point, i should call it `chunk_coords`. updated this to
```
@dataclass(frozen=True)
class _ChunkCoordsByteSlice:
    """Holds a core.indexing.ChunkProjection.chunk_coords and its byte range in a serialized shard."""

    chunk_coords: tuple[int, ...]
    byte_slice: slice
```

src/zarr/codecs/sharding.py
aldenks: Agreed it seems like it should be. I started going down this road but
1. we do want to avoid making an extra request for the shard index if its a full shard
1. without having the shard index, I don't see a way to go down the partial shard path that doesn't make that code messy

aldenks:
got spun back up on this and started making progress through Davis' review. will keep at it

d-v-b:
btw @aldenks we now have a benchmark suite -- if we don't have a workload in there that exercises the code path your are improving, let me know

src/zarr/codecs/sharding.py
aldenks: Agreed we should do this, but it's not entirely trivial because at this point we only have a ByteGetter that doesn't expose the _path_ to the shard that bytes are being retrieved from which we need as part of the cache key. I'd propose leaving for a follow up (especially since it took me so long to get back this)

aldenks:
@d-v-b awesome! right now here's the list im going to work through. It looks like code speed did not see any performance change which i assume means this code path is not tested in the benchmark suite (edit: confirmed, e2e test reads whole shards from a sharded array, indexing benchmark uses non-sharded array)

- [x] confirm `return None` in a couple places is maintaining existing behavior and document if so
- [x] fetch config values early. planning to in `_decode_partial_single` the highest level method in sharding.py that calls into the optimized path)
- [x] create sharding.py unit test module and add tests for coalescing behavior added here

---
not for this PR: caching shard indexes ([rationale](https://github.com/zarr-developers/zarr-python/pull/3004#discussion_r2705359720))

src/zarr/codecs/sharding.py
aldenks: ByteGetter.get can return None and if it does there's no useful fallback i can think of (I'm adding commenting)

aldenks: ByteGetter.get can return None and if it does there's no useful fallback i can think of (I'm adding commenting)

src/zarr/codecs/sharding.py
aldenks: Needing to handle ByteGetter returning none, this time for chunk data. I added commenting and changed the logic to exactly match what it was before. 

Previously if there was a mix of successful and None responses for ByteGetter.get it would have just dropped the keys from the dict for the failed ones and passed the valid ones along. I suspect that would have failed anyways later in the decoding pipeline due to a mismatch between expected indexes and available chunk bytes, but I've updated to maintain previous behavior.

Bigger picture it might be nicer to not allow None responses from ByteGetter.get and instead require useful exceptions but that's quite an api change.

aldenks (commented):
Ok! @d-v-b  I believe I've addressed all your feedback. I did in fact review each of the new unit tests even though opus did the drafting based on the integration tests. (Edit, fixing pre-commit mypy errors, was having issue w/ my local precommit)

src/zarr/codecs/sharding.py
d-v-b: > Bigger picture it might be nicer to not allow None responses from ByteGetter.get and instead require useful exceptions but that's quite an api change.

Totally agree! But it would be a pretty sweeping change.

aldenks:
I added a benchmark in #3663 and will get it on this branch if/once we get that on main.

The benchmark in there which shows the improvement is `tests/benchmarks/test_indexing.py::test_slice_indexing[(50, 50, 50)-(slice(10, -10, 4), slice(10, -10, 4), slice(10, -10, 4))-local]`. That's sharded, indexing partial shards, and using a local store that has some latency. Mean timing from that on my computer: main 137ms, this branch 58ms.

src/zarr/core/config.py
```diff
@@ -100,6 +100,12 @@ def enable_gpu(self) -> ConfigSet:
             },
             "async": {"concurrency": 10, "timeout": None},
             "threading": {"max_workers": None},
+            "sharding": {
+                "read": {
+                    "coalesce_max_bytes": 100 * 2**20,  # 100MiB
+                    "coalesce_max_gap_bytes": 2**20,  # 1MiB
```
aldenks: I think 1MiB default for coalesce_max_gap_bytes is roughly a good choice right now (if you're using object storage via fsspec implementations for example, a very common case). But with the combination of io latencies (networks, ssds, object stores) going down and high performance stores (obstore, icechunk, etc) becoming more popular, im thinking we may want a lower default to set us up for the "lots of small requests barely cost more than a few big ones" world. As the per-request latency goes down, the time to download the extra bytes in a gap between required chunks increases as a fraction of the total time. 256KiB sounds to me like a very conservative default (although I can also see arguments for going all the way to zero). Anyone have Strong Opinions?

aldenks:
Looking at the codspeed regression, `test_slice_indexing[(50, 50, 50)-(0, 0, 0)-memory] 1.9 ms -> 2.1 ms (-10.04%)` (access a single element in a sharded array with a memory store), my guess is the extra .2ms is the overhead of starting a thread introduced by the concurrent_map when there's no benefit to concurrency (only one call to store.get).  I'll try skipping the concurrent_map in that case, that should be very minimal additional complexity.

aldenks:
Ready for review. Key points for review:
* **Benchmarking**
  * I wouldn't expect the current benchmarks to quantify this improvement because the maximum store latency in the benchmarks is 0.1ms. We'd see a pretty significant improvement with store latencies of 1-10ms which is still quite optimistic for object storage
  * We should not be seeing any benchmark regressions after my [commit](https://github.com/zarr-developers/zarr-python/pull/3004/changes/4bec665221c14c06f24aca6d755f18d9c0e4a15f) earlier today to avoid starting threads if there's only one call to the store needed. I'm not sure why codspeed has skipped in my last few pushes to prove this on CI. 
 * **This PR adds a nested `concurrent_map`.** There's already one in the codec pipeline's read/read_batch, so it's technically possible to have concurrency of `async.concurrency**2`. However, that concurrency is critical to have anywhere near decent performance when interacting with modest latency stores so that feels like a worthwhile trade off to me.
 * **Default `coalesce_max_gap_bytes` value.** I think 1MiB as the default for `coalesce_max_gap_bytes` is roughly a good choice right now if you're using object storage via fsspec implementations. But with the combination of io latencies (networks, ssds, object stores) going down and high performance stores (obstore, icechunk, etc) becoming more popular, I'm thinking we may want a lower default to set us up for the "lots of small requests barely cost more than a few big ones" world (already the case for local disk based stores). As the per-request latency goes down, the time to retrieve the extra bytes in a gap between required chunks increases as a fraction of the total time. _256KiB sounds to me like a very conservative default, although I can also see arguments for going all the way to zero._

d-v-b:
>  I'm not sure why codspeed has skipped in my last few pushes to prove this on CI.

we blew our codspeed budget by running benchmarks on every PR. now it requires a `benchmark` label, which I'm applying now!

a high-level question before I get into the review: is there anything about this the logic in this PR that's specific to sharding? Because if not, we might want to explore applying this optimization at the store level.

d-v-b:
> Because if not, we might want to explore applying this optimization at the store level.

To elaborate on what this would look like:
- we implement a "new" `get_many` class on the store ABC. `get_many` takes a `tuple[tuple[str, byte range], ...]` and returns `Iterable[tuple[(str, byte range), Buffer]]`  I'm calling it "new" because we used to have this method, but it got removed because we weren't using it and that bugged someone.
- Inside `get_many`, we apply the optimization in your PR. The sharding codec never knows about it.

aldenks:
@d-v-b good question and that might be cleaner and offer benefit to new extensions in the future. Theres two optimizations in here.

1. Making concurrent requests - this is not specific to sharding, but obviously is achieved/controlled in other situations like requests to different underlying objects/files in a preexisting way with concurrent_map/async.concurrency 

2. coalescing nearby requests - technically not sharding specific, but also only offers benefit when you have multiple nearby requests to the same object/file. Does that happen in any other situation?

d-v-b:
> Making concurrent requests - this is not specific to sharding, but obviously is achieved/controlled in other situations like requests to different underlying objects/files in a preexisting way with concurrent_map/async.concurrency

Yes, and insofar as concurrency limits are part of runtime configuration, it might be an easier knob to turn if it's defined in one place: on the store.

> Does that happen in any other situation?

not currently outside of sharding, but if we ever improve our zip support to make it backend-agnostic, then we would absolutely be making dense byte-range requests against a single bytestream

This might come as a curveball, so I would only pursue this direction if it makes this PR easier. We can always lift this implementation into the store API later if needed.

d-v-b:
btw obstore supports range request coalescing at the store level

cc @kylebarron

kylebarron:
based on the title "coalesce _and_ parallelize", just FYI obstore will coalesce (currently up to 1MB gaps, hardcoded pending https://github.com/developmentseed/obstore/issues/574) 

But do note that obstore will not split one large request into multiple concurrent smaller requests.

maxrjones:
> But do note that obstore will not split one large request into multiple concurrent smaller requests.

You could use https://obspec-utils.readthedocs.io/en/latest/api/splitting/ for this, though the API has yet to be finalized

aldenks:
I'm convinced this would be valuable to move into the Store. I started looking into that and ObjectStore's impl and I'm wondering how `get_partial_values` differs from the proposed `get_many`?  Or.. do we just need to call get_partial_values and most of this can go away.

Btw, this PR does not _split_ requests. I does have a feature to stop coalescing at some max size, but that's pretty incidental performance wise.

kylebarron:
> wondering how `get_partial_values` differs from the proposed `get_many`?

What's the difference between these two? One is getting multiple ranges from a single file and the other is across different files?

aldenks:
I dug more and `get_partial_values` coalescing and concurrency behavior is up to the implementation and many give you concurrency and of commonly available stores `ObjectStore` is the only that I've found that also coalesces.

| store | concurrent | coalesce | source |
| :--- | :---: | :---: | :--- |
| `ObjectStore` | ✅ | ✅ | [src](https://github.com/zarr-developers/zarr-python/blob/main/src/zarr/storage/_obstore.py#L439) |
| `FsspecStore` | ✅ | ❌ | [src](https://github.com/fsspec/filesystem_spec/blob/master/fsspec/asyn.py#L483) |
| `ZipStore` | ❌ | ❌ | [src](https://github.com/zarr-developers/zarr-python/blob/879e1cec04e5008fc215b8753c37eea0d54aaa4e/src/zarr/storage/_zip.py#L185) |
| `LocalStore` | ✅ | ❌ | [src](https://github.com/zarr-developers/zarr-python/blob/879e1cec04e5008fc215b8753c37eea0d54aaa4e/src/zarr/storage/_local.py#L209) |
| `IcechunkStore` | ✅ | ❌ | [src](https://github.com/earth-mover/icechunk/blob/main/icechunk/src/store.rs#L223) |

---
Thoughts
- There's concurrency on all the stores where it matters
- It's probably uphill to go get all those stores to coalesce, but possible. **option a**
- I _think_ a correct and general implementation at the abc Store level would be totally possible: just coalesce and then async gather a bunch of calls to `Store.get`, respecting the concurrency limit.
   - Replacing all existing get_partial_values impls with a generic one sounds like a significant testing burden. **option b**
   - We could make another method (e.g. `get_many`) that is generic and use that. **option c**
- Since `get_partial_values` is concurrent in most stores used w/ high latency storage, we can update the sharding codec to call that method. If users also want coalescing, they use `ObjectStore`. **option d**

Options c and d both sound good to me, with d seeming slightly better because it doesn't expand any api surface.

d-v-b:
> > wondering how `get_partial_values` differs from the proposed `get_many`?
> 
> What's the difference between these two? One is getting multiple ranges from a single file and the other is across different files?

There's not really a difference. `get_partial_values` is IMO just a confusing name for `get_many`. Funnily enough we have `_get_many` implemented on the ABC as a private method, which slightly differs from `get_partial_values` insofar as `_get_many` supports streaming, while `get_partial_values` eagerly returns a list of results.

@aldenks what looks like the best way forward for you? Should we overload the `get_partial_values` implementations for stores where request coalescing makes sense?

aldenks:
@d-v-b perfect, thank you. given that I'm thinking we
- update this PR so the sharding codec calls `get_partial_values`. In practice that will get us concurrency for all major stores. Concurrency is the main lever that will improve wall clock time on high latency stores so we'll be in a much better performance situation with that. The diff will be much smaller.
- Then as stores are interested we/their maintainers can implement coalescing if it is worthwhile for that store. That is primarily a _cost_ optimization for situations where users pay per request to their store. ObjectStore already supports this and that's a go-to store for people who are doing large scale work where these costs can add up.

I'll put up an implementation of that this week unless i hear other suggestions!

aldenks:
@d-v-b May I have a review when you're able?

Serial calls to `get` in the sharding codec have been replaced with a single call to `get_partial_values`. All explicit coalescing and concurrency are no longer in this PR -- store implementations may do that in their `get_partial_values` implementations (see concurrency + coalescing [availability by store table](https://github.com/zarr-developers/zarr-python/pull/3004/#issuecomment-3963684637)).

Implementation notes:
- At the site where we want to call `get_partial_values` in the sharding codec we have a `ByteGetter` (the protocol, not a `Store` instance) and `ByteGetter` has only a `get` method. These changes add a `get_partial_values` method to the protocol (and to `ByteSetter`, which is used as a `ByteGetter` in `_encode_partial_single`.)
- Simple/pass through implementations of `get_partial_values` are added to the two objects that assume `ByteGetter` at the moment: `StorePath` and `_ShardingByteGetter`.

Because the main `Store` implementations used with higher latency stores all implement concurrency this change will effectively make what were serial requests concurrent where it counts. It keeps the choice about how to make a batch of requests in the `Store` which arguably has the most information about how to do that efficiently for the specific store type.

src/zarr/abc/store.py
```diff
@@ -695,13 +695,25 @@ async def get(
...
 class ByteSetter(Protocol):
     async def get(
         self, prototype: BufferPrototype, byte_range: ByteRequest | None = None
     ) -> Buffer | None: ...
 
+    async def get_partial_values(
+        self,
+        prototype: BufferPrototype,
+        byte_ranges: Iterable[ByteRequest | None],
+    ) -> list[Buffer | None]: ...
```
aldenks: This is on the _setter_ because there are places in the sharding codec that pass a setter into a ByteGetter. Setter is effectively treated as a subclass/superset of ByteGetter. That's a preexisting thing I think we don't want to pull apart here.

aldenks:
@d-v-b despite the build being red, this is ready for review. The failing step is codspeed runner usage exceeded. Probably you were already aware and if so sorry for bugging!

d-v-b:
hi @aldenks sorry for taking so long with this! over in https://github.com/zarr-developers/zarr-python/pull/3925 I propose some changes to the store API that would support the direction you are going here. I hope that if we can get those changes in, this PR gets easier. Let me know if that's true!

aldenks:
@d-v-b #3925 looks very reasonable. Changing this PR to use that would basically be a `get_partial_ranges` -> `get_ranges` rename and we could remove some of the adding get_partial_ranges methods to ByteGetter/Setter and StorePath parts of this PR.

I'll watch that PR and update here accordingly!

aldenks:
I'll give my latest changes to use Store.get_ranges a self-review tomorrow, update the PR description, and highlight two places for input (`isinstance(byte_getter, StorePath)` branch and if we should add config options for coalescing options).

But tl;dr the diff is looking nice and clean now that it builds on #3925

src/zarr/codecs/sharding.py
```diff
@@ -807,6 +802,66 @@ async def _load_full_shard_maybe(
...
+            if chunk_byte_slice is not None:
+                chunk_coord_byte_ranges.append(
+                    (chunk_coord, RangeByteRequest(chunk_byte_slice[0], chunk_byte_slice[1]))
+                )
+
+        if not chunk_coord_byte_ranges:
+            return {}
+
+        shard_dict: ShardMutableMapping = {}
+        if isinstance(byte_getter, StorePath):
```
aldenks, line 805: Highlighting for review: This `isinstance` check lets us avoid adding `get_ranges` to the `ByteGetter` protocol and everything that implements it (`StorePath`, `_ShardingByteGetter`, `_ShardingByteSetter`).  I can see good arguments for and against doing this.

src/zarr/codecs/sharding.py
```diff
@@ -807,6 +802,66 @@ async def _load_full_shard_maybe(
...
+            # External store: use Store.get_ranges for coalescing + concurrency.
+            byte_ranges = [byte_range for _, byte_range in chunk_coord_byte_ranges]
+            try:
+                async for group in byte_getter.store.get_ranges(
+                    byte_getter.path, byte_ranges, prototype=prototype
+                ):
```
aldenks, line 811: This uses `get_ranges` default argument values for `max_concurrency`, `max_gap_bytes`, and `max_coalesced_bytes` values. **Should we expose any of these as config options?** If so, are we namespacing them as part of sharding or get_ranges (`sharding.max_coalesced_bytes` or `get_ranges.max_coalesced_bytes` or ..)?

aldenks (commented):
@d-v-b (or others), this is ready for review! Two places I'd particularly like input on in the comments.

Benchmark regression: The partial sharded reads from memory store benchmarks have each gone from 1.8ms to 2ms. I'm pretty sure this is the tiny overhead of spinning up concurrent threads. We could opt memory store out of the default get_ranges implementation, but that also seems like a pretty small performance change.

src/zarr/codecs/sharding.py
d-v-b: IMO they should be config options, but I think we can defer that for a follow-up PR. what do you think?

src/zarr/codecs/sharding.py
d-v-b: looks fine to me

chase-dwelle:
Thanks for this work, just wanted to put in a comment with a reproducible test case on larger data which highlights improvements. I see a ~4x speedup against the same call on `main` reading from a sharded dataset with many chunks per shard.

Setup:

- GEM3 weather reforecast, hosted as zarr v3 at `https://gemv3-reforecast.salient-open-data.com/forecast`
- Backend: `obstore.store.HTTPStore` wrapped in `zarr.storage.ObjectStore`
- Wrapped in `LoggingStore` to count store API calls
- Array layout: `shape=(3443, 168, 50, 721, 1440)`, `chunks=(1, 7, 50, 36, 36)`, `shards=(1, 7, 50, 288, 288)`, `dims=(forecast_date, lead, sample, lat, lon)`
- Inner chunks are ~1.5 MB compressed; ~64 inner chunks per shard

```python
import xarray as xr
import obstore as obs
from zarr.storage import ObjectStore, LoggingStore

http_store = obs.store.HTTPStore.from_url(
    "https://gemv3-reforecast.salient-open-data.com/forecast"
)
store = LoggingStore(ObjectStore(http_store, read_only=True), log_level="WARNING")
ds = xr.open_zarr(store, chunks=None, consolidated=True)
```

Computed some stats on number of reads and time to make those reads across 11 sequential calls for each use case/branch.

### Case 1: regional spatial subset: ~3.9× faster

```python
ds.sel(forecast_date="2025-01-01",
       lat=slice(20, 55), lon=slice(-130, -60))["2m_temperature"].values
# shape: (168, 50, 141, 281) — spans 96 shards, ~11 inner chunks per shard
```

| branch | median   | mean     | min      | max      | `get` (HTTP) | `get_ranges` | input ranges |
| ------ | -------- | -------- | -------- | -------- | ------------ | ------------ | ------------ |
| `main` | 10.806s  | 10.731s  | 9.887s   | 11.110s  | 1,176        | 0            | 0            |
| PR     |  2.752s  |  2.787s  | 2.626s   |  3.176s  |   256        | 96           | 1,080        |

96 shards × ~11 chunks each → 1,080 input chunk ranges across 96 `get_ranges` calls, coalesced down to 256 actual HTTP fetches (≈ 1 index suffix-read + 1-2 merged-group fetches per shard). vs main's 1,176 individual range `get`s.

### Case 2: global tile, single lead slice — ~3.9× faster

```python
ds["2m_temperature"].sel(forecast_date="2025-01-01").isel(lead=0).values
# shape: (50, 721, 1440) — 15 shards: 10 fully-covered, 5 lat-edge coalesces
# since the last lat shard is only partially full
```

| branch | median   | mean     | min      | max      | `get` | `get_ranges` | input ranges |
| ------ | -------- | -------- | -------- | -------- | ----- | ------------ | ------------ |
| `main` |  3.659s  |  3.697s  | 3.554s   |  3.871s  | 215   | 0            | 0            |
| PR     |  0.937s  |  0.933s  | 0.868s   |  0.980s  |  20   | 5            | 200          |

src/zarr/codecs/sharding.py
aldenks: Follow up PR sounds great to me.

aldenks:
@d-v-b given our conversations above, from my perspective this is ready to merge!

d-v-b (approved):
awesome work!

PR #3021, opened 2025-04-25
Implement Store.move

### LINKED ISSUE #2108: [v3] Implement store methods for erase and rename operations

brokkoli71:
- fixes #2108

This PR implements a move method for stores. This only affects stores that contain a path. afaics thats:
- LocalStore
- ZipStore
- ~~(maybe FsspecStore)~~

Am I missing one?
<del> 
Maybe, this PR could also implement moving the content from one store to a new one of a different type e.g.

```python
store = LocalStore(root="path")
store.move(dest=MemoryStore())
```
</del> 
Feedback/thoughts welcome

TODO:


paraseba:
I don't think this can be meaningfully implemented by any other Stores. Maybe it shouldn't be a method in the abstract class, but purely a concrete implementation for the Stores that can do it efficiently.

In a real world scenario, moving a cloud Store could take days, cost thousands of dollars and fail. People wouldn't want to run this kind of operation in this way.

brokkoli71:
good point! @paraseba 
what do you think of `store.move(dest=another_store)` ? 
for example from memory to local stores?

paraseba:
@brokkoli71 
> what do you think of store.move(dest=another_store) ?

That would be an extremely useful thing, I have wanted that to exist many times, but only if I can use it confidently across different stores. That would mean it needs checkpointing, retrying, restarting, progress feedback, etc. So it goes beyond what we should implement in a store. In particular, being part of the `Store` instance doesn't benefit it, because it won't have internal knowledge of the `dest` store anyway.

Regarding the name, I think `move` probably should be reserved for renaming members of the store, for example, moving an array from group A to group B, or changing the name of a group. That is an operation that some stores can run very efficiently, and that is also very useful to users. Anybody with a few real-world datasets wanted at least once to rename something.

I don't think we should implement `move(dest: Store)` at this point. I think that's a higher level tool. As @d-v-b , I'm also not sure what to say about the "internal" move you originally proposed. I'm pretty sure we shouldn't implement it for the cloud stores, because it would be misleading to users. And for memory stores it doesn't make sense. So, maybe it's better implemented initially as a concrete method on file system stores and then we see if people ask for it in other types of stores?

src/zarr/storage/_local.py
```diff
@@ -253,5 +253,21 @@ async def list_dir(self, prefix: str) -> AsyncIterator[str]:
...
             pass
 
+    async def move(self, dest_root: Path | str) -> None:
+        """
+        Move the store to another path. The old root directory is deleted.
+        """
+        if isinstance(dest_root, str):
+            dest_root = Path(dest_root)
+        os.makedirs(dest_root, exist_ok=True)
+        for src_file in self.root.rglob("*"):
```
normanrz, line 263: Why not use `shutil.move` to move the entire folder?
If implementing this ourselves, the newly created folders might need to copy the permissions and settings of the source folders. And, we could put the move operations on a thread pool.

brokkoli71: @normanrz Using `shutil.move('some/path/origin', 'other/path/destination')`, the folder `origin` would be moved into the `destination` folder, resulting in `root=other/path/destination/origin`. I think we would want the new root to be `other/path/destination` instead.
Perhaps the following would be the way to go:
```python
shutil.move(self.root, dest_root.parent)
os.rename(dest_root.parent/self.root.name, dest_root).
```
However, a thread pool seems like a good idea as well for large stores.

normanrz: Is that also true, if dest_root doesn't exist? That would be equivalent to `mv`.

brokkoli71: You're right, I wasn't aware of that.

normanrz (approved):

PR #3031, opened 2025-05-01
Add flexible deserialization for datetime fill values

### LINKED ISSUE #3028: serializing not a time (NaT) to metadata
Numpy datetime and timedelta dtypes have a nan-like value called `NaT`. The `NaT` value internally represented by the 64 bit signed integer -9223372036854775808:

```python
np.int64(-9223372036854775808).view('datetime64')
# np.datetime64('NaT')
```

The zarr v2 spec does not define a JSON-serializable fill value encoding for datetimes or timedeltas. In lieu of a spec, we can go by the implementation of zarr v2, which used the integer form of the datetime / timedelta for all fill value, including `NaT`:

```python
# /// script
# requires-python = ">=3.11"
# dependencies = [
#     "zarr==2.18",
#     "numcodecs<0.16"
# ]
# ///


import zarr
import json
z = zarr.create((10,), dtype='>M8[s]', fill_value='NaT')
print(json.loads(z.store['.zarray'])['fill_value'])
# -9223372036854775808
```

But we have [a property test ](https://github.com/zarr-developers/zarr-python/blob/36a1bac850c50fbaa1708f8986edd5766f5fd67b/tests/test_properties.py#L262)that checks if the numpy [`NaT` value is serialized to metadata as the string `"NaT"`](https://github.com/zarr-developers/zarr-python/blob/36a1bac850c50fbaa1708f8986edd5766f5fd67b/tests/test_properties.py#L303-L304), but this goes against the behavior of v2. It should be serialized as an int.

moradology:
This PR addresses a backwards incompatibility issue created in recent standardization of the fill_value serialization/deserialization work. It does so by adopting a flexible decoding approach that takes either the minimum signed 64bit value or else 'NaT' as meaning `Not a Time` while sticking to writing fill values as 'NaT'.

Fixes #3028 

TODO:


src/zarr/core/metadata/v2.py
```diff
@@ -327,7 +327,15 @@ def parse_fill_value(fill_value: Any, dtype: np.dtype[Any]) -> Any:
     """
 
     if fill_value is None or dtype.hasobject:
+        # Pass through None or if dtype is object
         pass
+    elif dtype.kind in "M":
```
d-v-b, line 332:
```suggestion
    elif dtype.kind in ("M", "m"):
```
handle timedelta64 as well

d-v-b:
I think this PR has been superseded by the work in #2874 @moradology could you confirm? If so, maybe this can be closed.

dstansby:
I'll close this - if this is still relevant after #2874, please feel free to re-open or open a new PR.

Issue #3038, opened 2025-05-06
Can't use `zarr.create` to make v2 array with no compressor


dstansby:
In zarr-python 2:

```python
import zarr

arr = zarr.create(store={}, shape=(1), dtype='uint8', compressor=None)
print(arr.compressor)
# None
```

In zarr-python 3:

```python
import zarr

arr = zarr.create(store={}, shape=(1), dtype='uint8', zarr_format=2, compressor=None)
print(arr.compressors)
# (Zstd(level=0),)
```

I was expecting the list of compressors to be empty, becuase by passing `compressor=None` I'm specifying explicitly to have no compressors.

PR #3039, opened 2025-05-06
Allow no compressor for v2 arrays

dstansby:
Fixes https://github.com/zarr-developers/zarr-python/issues/3038.


src/zarr/api/asynchronous.py
```diff
@@ -990,7 +992,7 @@ async def create(
         dtype = parse_dtype(dtype, zarr_format)
         if not filters:
             filters = _default_filters(dtype)
-        if not compressor:
```
d-v-b, line 993: i haaaaaaate `if not x`! I'm guessing this was getting tripped because the empty tuple is falsey?

dstansby: No, this was correct before, the change is deliberate because the default value of `compressor` has changed.

src/zarr/core/array.py
```diff
@@ -3785,7 +3797,11 @@ def _get_default_codecs(
     | Literal["auto"]
     | None
 )
-CompressorLike: TypeAlias = dict[str, JSON] | BytesBytesCodec | numcodecs.abc.Codec | None
+# Union of acceptable types for v2 and v3 compressors
+CompressorLike: TypeAlias = (
```
d-v-b, line 3801: Do we want `None` in here? I feel like `CompressorLike` should be all the different ways of specifying a particular compressor, but `None` is not that.

dstansby: `None` is an acceptable compressor for v2

d-v-b: `None` is an acceptable value for the `"compressor"` field in metadata, but it's not a compressor. that's the distinction i'm getting at

d-v-b: and if we were to write a type annotation for the `"compressor"` field in metadata, it would be `Compressor | None`

src/zarr/core/array.py
```diff
@@ -3785,7 +3797,11 @@ def _get_default_codecs(
     | Literal["auto"]
     | None
 )
-CompressorLike: TypeAlias = dict[str, JSON] | BytesBytesCodec | numcodecs.abc.Codec | None
+# Union of acceptable types for v2 and v3 compressors
```
dstansby, line 3800:
```suggestion
# Union of acceptable types for users to pass in for both v2 and v3 compressors
```

src/zarr/core/array.py
dstansby: Ah, I see what you mean. This is intended as a type annotation for the public API, which means `None` and 'auto' are acceptable values. I updated the code comment to clarify this a bit.

d-v-b: Yes and my suggestion is about the public API. We should use `CompressorLike` for something that is like a compressor. That's compatible with a function that takes `CompressorLike | Literal["auto"] | None`.

dstansby: I had a quick look at what other existing `*Like` types do, and the ones already in use with `Group.create_array` (e.g., `ShardsLike`, `FiltersLike`) also contain `'auto'` and `None`. So while I don't disagree with changing that, I think that's something that should be done consistently and is outside the scope of this PR?

d-v-b: I don't agree that we need specific typevar for this type, as long as we can use a short union, but this is probably a discussion for another PR entirely

d-v-b: > I had a quick look at what other existing `*Like` types do, and the ones already in use with `Group.create_array` (e.g., `ShardsLike`, `FiltersLike`) also contain `'auto'` and `None`. So while I don't disagree with changing that, I think that's something that should be done consistently wiht all the types and is outside the scope of this PR?

agreed! sorry for the noise

dstansby:
I don't think it's changing the default behaviour (if you don't specify `compressor`, you still get a default compressor as before), but it is changing the behaviour of explicitly passing `compressor=None`. I suspect not many people are explicitly doing that (and expecting a compressor to be automatically chosen), so I'm erring on the side of it's more of a bugfix and fine for a micro release. But happy for a minor release too.

tests/test_api.py
```diff
@@ -1190,3 +1192,20 @@ def test_gpu_basic(store: Store, zarr_format: ZarrFormat | None) -> None:
         # assert_array_equal doesn't check the type
         assert isinstance(result, type(src))
         cp.testing.assert_array_equal(result, src[:10, :10])
+
+
+def test_v2_without_copmpressor() -> None:
```
d-v-b, line 1197:
```suggestion
def test_v2_without_compressor() -> None:
```
typo

tests/test_api.py
```diff
@@ -1190,3 +1192,20 @@ def test_gpu_basic(store: Store, zarr_format: ZarrFormat | None) -> None:
...
+    # Make sure it's possible to set no compressor for v2 arrays
+    arr = zarr.create(store={}, shape=(1), dtype="uint8", zarr_format=2, compressor=None)
+    assert arr.compressors == ()
+
+
+def test_v2_with_v3_copmpressor() -> None:
```
d-v-b, line 1203:
```suggestion
def test_v2_with_v3_compressor() -> None:
```

d-v-b:
gotcha, if the default has the same behavior after this change then I think it can go in immediately

d-v-b (approved):

PR #3045, opened 2025-05-07
Fix typing errors in testing.stateful

dstansby:
This fixes typing errors in the stateful testing sub-module.

Along the way I also realised that the dimension names parameter was typed wrong - it should allow None to be in the list too. So I made a type variable for that, and fixed the typing throughout.


src/zarr/testing/strategies.py
```diff
@@ -232,14 +238,14 @@ def np_array_and_chunks(
...
-@st.composite  # type: ignore[misc]
+@st.composite
 def arrays(
     draw: st.DrawFn,
     *,
     shapes: st.SearchStrategy[tuple[int, ...]] = array_shapes,
     compressors: st.SearchStrategy = compressors,
     stores: st.SearchStrategy[StoreLike] = stores,
-    paths: st.SearchStrategy[str | None] = paths(),  # noqa: B008
+    paths: st.SearchStrategy[str] = paths(),  # noqa: B008
```
dcherian, line 248: None should be allowed here, no?

src/zarr/testing/strategies.py
```diff
@@ -30,17 +32,17 @@
...
+@st.composite
+def keys(draw: st.DrawFn, *, max_num_nodes: int | None = None) -> str:
     return draw(st.lists(node_names, min_size=1, max_size=max_num_nodes).map("/".join))
 
 
-@st.composite  # type: ignore[misc]
-def paths(draw: st.DrawFn, *, max_num_nodes: int | None = None) -> Any:
+@st.composite
+def paths(draw: st.DrawFn, *, max_num_nodes: int | None = None) -> str:
     return draw(st.just("/") | keys(max_num_nodes=max_num_nodes))
```
dcherian, line 42:
```suggestion
    return draw(st.none() | st.just("/") | keys(max_num_nodes=max_num_nodes))
```

If `path=None` is allowed.

src/zarr/testing/strategies.py
dstansby: Shouldn't paths just be strings? At least `Group.create_array()` (where this variable is passed in the tests) doesn't accept `path=None`.

dcherian (approved):

PR #3047, opened 2025-05-08
Fix typing errors in test_consolidated

dstansby:
I didn't get the chance to finish this, but I see no harm in merging the fixes I have now, and fixing the rest later.


dstansby:
Meh, I can't be bothered sorting out the errors at the moment, so will close.

PR #3048, opened 2025-05-08
Fix readtehdocs build

dstansby:
Playing whack-a-mole with broken dependencies again; see https://github.com/sphinx-doc/sphinx/issues/13533.


dstansby:
Fixed upstream, so this is no longer needed

PR #3052, opened 2025-05-10
Fix typing in a bunch of store tests

dstansby:


tests/test_store/test_fsspec.py
```diff
@@ -109,10 +114,12 @@ async def test_basic() -> None:
     data = b"hello"
     await store.set("foo", cpu.Buffer.from_bytes(data))
     assert await store.exists("foo")
-    assert (await store.get("foo", prototype=default_buffer_prototype())).to_bytes() == data
+    assert (buf := (await store.get("foo", prototype=default_buffer_prototype()))) is not None
+    assert buf.to_bytes() == data
```
d-v-b, line 118:
```suggestion
    buf = await store.get("foo", prototype=default_buffer_prototype())
    assert buf is not None
    assert buf.to_bytes() == data
```

I don't think we gain much from mixing assignment with the is-none-check in a single line

d-v-b (changes requested):
one tiny nit otherwise this look good

tests/test_store/test_fsspec.py
dstansby: Haha fair I might have got too carried away with the walrus here 😄

d-v-b (approved):

Issue #3061, opened 2025-05-15
zarr.load deletes data


alfieroddan:
### Zarr version

3.0.7

### Numcodecs version

0.16.0

### Python Version

3.11.11

### Operating System

MacOS

### Installation

using pip into venv

### Description

If I load a folder with zarr but there is no .zarray, it deletes my data! This is really not good and not obvious.

Is this intended or a bug?

I now know that the load function is not read only!

Instead use open_array with read only mode:

```python
z = zarr.open_array(path, mode='r')
```

### Steps to reproduce

```bash
(env) /Users/Documents/test_folder/video_2025-04-22_16-47-43-917 $ ls
zarr_folders rgb
(env) /Users/Documents/test_folder $ python 
Python 3.11.11 (main, Dec  3 2024, 17:20:40) [Clang 16.0.0 (clang-1600.0.26.4)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import zarr
>>> zarr.load("/Users/Documents/test_folder/video_2025-04-22_16-47-43-917")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/Documents/EnPlayback/env/lib/python3.11/site-packages/zarr/api/synchronous.py", line 151, in load
    return sync(
           ^^^^^
  File "/Users/Documents/EnPlayback/env/lib/python3.11/site-packages/zarr/core/sync.py", line 163, in sync
    raise return_result
  File "/Users/Documents/EnPlayback/env/lib/python3.11/site-packages/zarr/core/sync.py", line 119, in _runner
    return await coro
           ^^^^^^^^^^
  File "/Users/Documents/EnPlayback/env/lib/python3.11/site-packages/zarr/api/asynchronous.py", line 275, in load
    raise NotImplementedError("loading groups not yet supported")
NotImplementedError: loading groups not yet supported
(env) /Users/Documents/test_folder/video_2025-04-22_16-47-43-917 $ ls
zarr.json
```

joshmoore:
can confirm

### bar.py
```
# /// script
# requires-python = ">=3.12"
# dependencies = [
#     "zarr",
# ]
# ///

import zarr
zarr.load("/tmp/bar")
```

### setup and run
```
cd /tmp
mkdir bar
touch bar/IMPORTANT
uv run bar.py
Installed 5 packages in 10ms
Traceback (most recent call last):
  File "/private/tmp/bar.py", line 9, in <module>
[...]

PR #3062, opened 2025-05-15
Fix overwrite modes

dstansby:
Fixes https://github.com/zarr-developers/zarr-python/issues/3061. @jhamman, you last modified these lines in https://github.com/zarr-developers/zarr-python/pull/2442, is there a reason why "a" and "r+" was in ovewrite modes? It seems to me like they definitely shouldn't be there.


dstansby:
What would be the logic in leaving "a" and "r+" in overwrite modes, given those modes should not overwrite any data?

d-v-b (approved):

PR #3064, opened 2025-05-15
Test using latest numpy

dstansby:
~~This is a bit of an experiment to see if using `'2'` will give us the latest verison of numpy to test on~~

This bumps to the latest version of numpy. We still need to fix typing to work with numpy 2.2, but worth running tests for now at least.

### DIFF
--- a/.github/workflows/gpu_test.yml
+++ b/.github/workflows/gpu_test.yml
@@ -25,7 +25,7 @@ jobs:
     strategy:
       matrix:
         python-version: ['3.11']
-        numpy-version: ['2.1']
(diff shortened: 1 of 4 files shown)


jhamman:
+1 on this change if we can make it work. I seem to remember trying and failing but it seems like it should work.

dstansby:
So the problem with using ~ is that

> This operator MUST NOT be used with a single segment version number such as ~=1.

(see [PEP 440](https://peps.python.org/pep-0440/#version-specifiers))

I'm trying out using `.*` instead with ==

d-v-b (approved):

Issue #2298, opened 2024-10-02
Raise specific errors when failing to open a group.


TomAugspurger:
### Zarr version

v3

### Numcodecs version

na

### Python Version

na

### Operating System

na

### Installation

na

### Description

Zarr 2.x raised a `zarr.errors.GroupNotFoundError` when failing to open a group.

3.x currently raises a `ValueError` while trying to fall back to creating a Group on a read-only store.

### Steps to reproduce

```pytb
In [6]: zarr.open_group(zarr.storage.MemoryStore(), mode="r")  # 2.x
---------------------------------------------------------------------------
GroupNotFoundError                        Traceback (most recent call last)
Cell In[6], line 1
----> 1 zarr.open_group(zarr.storage.MemoryStore(), mode="r")

File ~/gh/zarr-developers/zarr-v2/.direnv/python-3.10/lib/python3.10/site-packages/zarr/hierarchy.py:1578, in open_group(store, mode, cache_attrs, synchronizer, path, chunk_store, storage_options, zarr_version, meta_array)
   1576         if contains_array(store, path=path):
   1577             raise ContainsArrayError(path)
-> 1578         raise GroupNotFoundError(path)
   1580 elif mode == "w":
   1581     init_group(store, overwrite=True, path=path, chunk_store=chunk_store)

GroupNotFoundError: group not found at path ''
```

versus 3.x:

```pytb
In [2]: zarr.open_group(zarr.storage.MemoryStore(), mode="r")
---------------------------------------------------------------------------
FileNotFoundError                         Traceback (most recent call last)
File ~/gh/zarr-developers/zarr-python/src/zarr/api/asynchronous.py:609, in open_group(store, mode, cache_attrs, synchronizer, path, chunk_store, storage_options, zarr_version, zarr_format, meta_array, attributes)
    608 try:
--> 609     return await AsyncGroup.open(store_path, zarr_format=zarr_format)
    610 except (KeyError, FileNotFoundError):

File ~/gh/zarr-developers/zarr-python/src/zarr/core/group.py:179, in AsyncGroup.open(cls, store, zarr_format)
    178 if zarr_json_bytes is None and zgroup_bytes is None:
--> 179     raise FileNotFoundError(
    180         f"could not find zarr.json or .zgroup objects in {store_path}"
    181     )
    182 # set zarr_format based on which keys were found

FileNotFoundError: could not find zarr.json or .zgroup objects in memory://4519879936

During handling of the above exception, another exception occurred:

[...]

PR #3066, opened 2025-05-16
Add a GroupNotFoundError

dstansby:
Fixes https://github.com/zarr-developers/zarr-python/issues/2298.

Note I also updated the base zarr exception to no longer inherit from ValueError, because in general errors we raise are not always ValueErrors. I'll add a separate release note for this change.

### DIFF
--- a/tests/test_metadata/test_v3.py
+++ b/tests/test_metadata/test_v3.py
@@ -20,7 +20,7 @@ from zarr.core.metadata.v3 import (
     parse_fill_value,
     parse_zarr_format,
 )
-from zarr.errors import MetadataValidationError
+from zarr.errors import MetadataValidationError, NodeTypeValidationError
 
 if TYPE_CHECKING:
     from collections.abc import Sequence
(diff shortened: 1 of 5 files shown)


dstansby:
I didn't spot #3012, thanks! I'll put this back in draft and have a think when I've got a bit more time.

dstansby:
I reverted this to just adding a `GroupNotFoundError`, which isn't breaking and a self-contained change. This could be merged as is, but I'm +/- 0 on whether to try and put in more errors in the same PR. Pros: less confusing to users if we put all the new exceptions in in one go. Cons: harder to do/review. Anyone have any thoughts either way?

d-v-b:
> I reverted this to just adding a `GroupNotFoundError`, which isn't breaking and a self-contained change. This could be merged as is, but I'm +/- 0 on whether to try and put in more errors in the same PR. Pros: less confusing to users if we put all the new exceptions in in one go. Cons: harder to do/review. Anyone have any thoughts either way?

incremental progress is progress IMO, so I'm fine with just the one exception added here.

tests/test_metadata/test_v3.py
```diff
@@ -88,7 +89,8 @@ def test_parse_node_type_invalid(node_type: Any) -> None:
 @pytest.mark.parametrize("data", [None, "group"])
 def test_parse_node_type_array_invalid(data: Any) -> None:
     with pytest.raises(
-        ValueError, match=f"Invalid value for 'node_type'. Expected 'array'. Got '{data}'."
+        NodeTypeValidationError,
```
d-v-b, line 92: I am wondering now why we have a special exception for this case...

Issue #2949, opened 2025-04-03
Array.read_only incorrect


dstansby:
```python
import zarr
import zarr.storage

store = zarr.storage.MemoryStore()
zarr.create((100,), store=store, zarr_format=2, path="a")


arr = zarr.open_array(store=store, path="a", zarr_format=2, mode="r")
print(arr.read_only)
# False
```

Given the mode is "r", I think `.read_only` should return `True`?

d-v-b:
In zarr-python 3, arrays and groups don't have their own read / write permission model, instead all of that is handled by the storage layer.

Thus the `read_only` property was added for backwards compatibility with v2, and it doesn't do what you would expect -- [it reports whether the array is backed by a read-only store.](https://github.com/zarr-developers/zarr-python/blob/9e8b50ae19cc63ad573f58569c3ef5826a5c60fc/src/zarr/core/array.py#L1059-L1068). So in your example, because you are are-using a store instance, it keeps its original mode (which is probably `a`). I agree that this is confusing.

dstansby:
Hmm, in that case should passing a mode to `open_array` raise an exception, when the passed mode is different to the mode the store is open with?

d-v-b:
i believe `open_array` has a `mode` parameter for backwards compatibility reasons. Since we are controlling access modes with the storage layer, IMO it doesn't make sense to pretend that it's also controlled by the array / group layer, and we should deprecate the `mode` keyword argument.

dstansby:
> we should deprecate the mode keyword argument.

👍  - I think it's pretty bad that it silently does nothing at the moment!

PR #3068, opened 2025-05-18
Error on invalid store mode

dstansby:
This avoids instances where mode='r' could be passed, by the resulting array is not read-only. Fixes https://github.com/zarr-developers/zarr-python/issues/2949

### DIFF
--- a/changes/3068.bugfix.rst
+++ b/changes/3068.bugfix.rst
@@ -1,0 +1,1 @@
+Trying to open an array with ``mode='r'`` when the store is not read-only now raises an error.
--- a/src/zarr/storage/_common.py
+++ b/src/zarr/storage/_common.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import json
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, Literal
+from typing import TYPE_CHECKING, Any, Literal, Self
 
 from zarr.abc.store import ByteRequest, Store
 from zarr.core.buffer import Buffer, default_buffer_prototype
@@ -48,9 +48,7 @@ class StorePath:
         return self.store.read_only
 
     @classmethod
-    async def open(
-        cls, store: Store, path: str, mode: AccessModeLiteral | None = None
-    ) -> StorePath:
+    async def open(cls, store: Store, path: str, mode: AccessModeLiteral | None = None) -> Self:
         """
         Open StorePath based on the provided mode.
 
@@ -67,6 +65,9 @@ class StorePath:
         ------
         FileExistsError
             If the mode is 'w-' and the store path already exists.
+        ValueError
+            If the mode is not "r" and the store is read-only, or
+            if the mode is "r" and the store is not read-only.
         """
 
         await store._ensure_open()
@@ -78,6 +79,8 @@ class StorePath:
 
         if store.read_only and mode != "r":
             raise ValueError(f"Store is read-only but mode is '{mode}'")
+        if not store.read_only and mode == "r":
+            raise ValueError(f"Store is not read-only but mode is '{mode}'")
 
         match mode:
             case "w-":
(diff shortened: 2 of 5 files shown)


d-v-b:
this looks good to me, curious to hear your thoughts @jhamman  since you are the mode architect

dstansby:
My thinking here with an error was to avoid situations where one opened an array with `mode='r'`, but writing to it still worked. So although it has the potential to be disruptive, it's avoiding confusion with the API so I think worth it.

At a slightly higher level, I don't really understand why arrays can't have their own read/write permissions, but I missed that design decision.

dstansby:
We should decide one way or another before doing a 3.0.9 release. I'm still minded to go ahead with this, but expand the error message to say to open the store in read only mode if you want to open an array in read only mode. I'd also like to look at the xarray test failures, to see if there's an easy fix, or if there is a sensible use case that we're breaking with this change.

So before we release 3.0.9 with this, I think the todo is:
- Improve the error message that was added in this PR
- Review https://github.com/zarr-developers/zarr-python/pull/3138, and hopefully merge
- Check if there's a use case that this change prevents (by looking at failing xarray tests)

@maxrjones does that sound good? Anything I missed? Thanks for the downstream testing, it helps a lot!

dstansby:
Okay, I had a poke around at https://github.com/pydata/xarray/pull/10430, and convinced myself that we should revert this change for now. Long term we really need to sort out the complete mess of permissions and work out what our permission model actually is...

d-v-b:
for posterity can you elaborate a bit more on how the linked xarray PR convinced you to revert?

dstansby:
In the xarray tests, there are examples where they create a memory store that is not read only, write an array to it, and then try and open that array in read only mode. I think it's reasonable to ask for a read only array from a writeable store, and not be able to write to the array.

d-v-b:
an alternative solution would be to go with #https://github.com/zarr-developers/zarr-python/pull/3138 and modify xarray to use that method as needed. I'd be curious to hear the pros / cons for the revert solution vs the "add functionality" solution.

dstansby:
My sense is #3138 is a bit of a hack anyway, when it *should* be possible to open an array in read-only mode from a writeable store. Because no-one did a write up (?) of the permission model changes for zarr-python v3 I don't really understand whether this is deliberately not possible, or if it was just an oversight.

d-v-b:
>  it should be possible to open an array in read-only mode from a writeable store.

This is the key question. In Zarr-python 2 arrays and groups had a permission model. But in Zarr-python 3, we expressly moved the permissions model to the storage level, and so arrays / groups have defer entirely to the store for that. If we stick with that decision, then it _should not_ be possible to open an array in read-only mode from a writeable store, without creating a new read-only version of that store (which #3138 implements).

maxrjones:
> > it should be possible to open an array in read-only mode from a writeable store.
> 
> This is the key question. In Zarr-python 2 arrays and groups had a permission model. But in Zarr-python 3, we expressly moved the permissions model to the storage level, and so arrays / groups have defer entirely to the store for that. If we stick with that decision, then it _should not_ be possible to open an array in read-only mode from a writeable store, without creating a new read-only version of that store (which #3138 implements).

Rather than erroring on invalid store mode, we could emit a warning and create a copy of the store in the specified mode using #3138 (if implemented in the store class). This would balance matching the store mode with the API call and avoiding breaking changes. What would be the downsides of that approach?

dstansby:
👍 that sounds like a good approach to me - I'd even consider not warning and just silently returning a copy.

How shall we move forward then? Should we roll that change into https://github.com/zarr-developers/zarr-python/pull/3138?

dcherian:
> I think it's reasonable to ask for a read only array from a writeable store, and not be able to write to the array.

Yes. for example I'd like to read from array `a` and write to another array `b` in the same store, and have some protections around not writing to `a`.

> Rather than erroring on invalid store mode, we could emit a warning and create a copy of the store in the specified mode using https://github.com/zarr-developers/zarr-python/pull/3138 (if implemented in the store class).

:+1: great idea!

maxrjones:
> How shall we move forward then? Should we roll that change into https://github.com/zarr-developers/zarr-python/pull/3138?

I can update https://github.com/zarr-developers/zarr-python/pull/3138 to add `with_read_only()` to the other store classes as the first step. To keep PRs well-constrained, I think it would work well for you to either modify https://github.com/zarr-developers/zarr-python/pull/3145 to use with_read_only() as the second step or someone does that as a different PR after #3138 is finished/reviewed/merged.

PR #3074, opened 2025-05-20
(chore): release notes cleanup

ilan-gold:
[Description of PR]
oops on us @dstansby!
TODO:

### DIFF
--- a/changes/2950.bufgix.rst
+++ b/changes/2950.bufgix.rst
@@ -1,1 +1,0 @@
-Specifying the memory order of Zarr format 2 arrays using the ``order`` keyword argument has been fixed.
--- a/changes/2962.fix.rst
+++ b/changes/2962.fix.rst
@@ -1,1 +1,0 @@
-Internally use `typesize` constructor parameter for :class:`numcodecs.blosc.Blosc` to improve compression ratios back to the v2-package levels.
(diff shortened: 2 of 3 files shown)


dstansby:
Huh, odd - I wonder why `towncrier` didn't pick these up when I ran it?

ilan-gold:
@dstansby towncrier matches exactly on the type of issue, I believe, if memory serves, probably.  Mine was titled `fix` and the other was misspelled, I think, so it didn't recognize them as fragments.

dstansby:
Ah that would explain it. I wonder if there's a way we can catch these and error on CI before release... anyway, thanks for the fix here!

dstansby (approved):

d-v-b:
> Ah that would explain it. I wonder if there's a way we can catch these and error on CI before release... anyway, thanks for the fix here!

Could the "needs release notes" label check for valid (releasable) release notes?

dstansby:
Looks like https://github.com/twisted/towncrier/issues/137 is a relevant issue

ilan-gold:
@dstansby We also use towncrier in `anndata` and I think we essentially have the same issue.  The "best" we do is calling `towncrier` before actually building the docs so we can _in theory_ (big emphasis there) check:

https://github.com/scverse/anndata/blob/bb2d44c0ad5757c5fc59844e2cf686009bb6073d/.readthedocs.yml#L10-L12

ilan-gold:
You could actually probably edit that regex so it fails upon finding a non-supported `md` file (in our case).

PR #3075, opened 2025-05-21
Add full copy of zarr-python v2.18.7 under src/zarr/v2

jni:
This proposes to include a complete copy of zarr v2.18.7 under the zarr.v2
namespace. The goal is for libraries using the v2 API to be able to migrate
gradually, and not have to be environment-incompatible with libraries that have
fully migrated to v3. Such libraries could then do:

```python
import zarr

if not zarr.__version__.startswith('2.'):
    import zarr.v2 as zarr
```

It would also buy *us* more time to improve the migration guide.

For more discussion, see [this Zulip thread][zulip].

Currently, I have not imported it in `zarr.__init__.py`, so one would have to
explicitly import it to discover it. I think that's probably the right thing.

[zulip]: https://ossci.zulipchat.com/#narrow/channel/423692-Zarr/topic/Funding.20opportunity.3A.20NF.20small.20devel.20grant/near/519038080

TODO:
* [x] ~~Add unit tests and/or doctests in docstrings~~
* [x] ~~Add docstrings and API docs for any new/modified user-facing classes and functions~~


jeromekelleher:
Thanks for doing this @jni - I think would personally very much appreciate this feature.

Issue #2948, opened 2025-04-03
Incorrect warning when creating zarr_format=2 array with order keyword


dstansby:
```python
import zarr
import zarr.storage

zarr.create(
    (100,),
    store=zarr.storage.MemoryStore(),
    order="F",
    zarr_format=2,
)
```

warns
```
/Users/dstansby/software/zarr/pydantic-zarr/.venv/lib/python3.12/site-packages/zarr/api/asynchronous.py:1015: RuntimeWarning: The `order` keyword argument has no effect for Zarr format 3 arrays. To control the memory layout of the array, either use the `config` keyword argument, as in `config={'order: 'C'}`,or change the global 'array.order' configuration variable.
  _warn_order_kwarg()
```

But I'm creating a zarr format 2 array, so I don't think this should warn?

PR #3078, opened 2025-05-21
Don't warn when order passed for v2 arrays

dstansby:
The correct warning is raised further down the call stack in `AsyncArray._create`, so no need for it in `create()`.

Edit: this is turning into a bit of a "fix all the order arguments" rabbit hole...

Fixes https://github.com/zarr-developers/zarr-python/issues/2948


dstansby:
I'm going to work on this in a fresh PR

PR #3082, opened 2025-05-22
(feat): use `np.zeros` for buffer creation with `fill_value=0`

ilan-gold:
I was profiling code and noticed a non-trivial amount of time spent on `np.full` with a `fill_value=0` so I did a little digging and at least on my machine:

```python
In [1]: import numpy as np

In [2]: %timeit np.full((10_000, 10_000), dtype=np.float32, fill_value=0, order="C")
33.2 ms ± 784 μs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [3]: %timeit np.zeros((10_000, 10_000), dtype=np.float32, order="C")
1.04 μs ± 1.48 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

For example: 

https://github.com/zarr-developers/zarr-python/blob/481550a2b98d1bf51d731653208c26b5bcfce454/src/zarr/core/array.py#L1285-L1290

and

https://github.com/zarr-developers/zarr-python/blob/481550a2b98d1bf51d731653208c26b5bcfce454/src/zarr/codecs/sharding.py#L441-L443

are both points where this happens

TODO:

### DIFF
--- a/src/zarr/core/buffer/cpu.py
+++ b/src/zarr/core/buffer/cpu.py
@@ -154,7 +154,8 @@ class NDBuffer(core.NDBuffer):
         order: Literal["C", "F"] = "C",
         fill_value: Any | None = None,
     ) -> Self:
-        if fill_value is None or fill_value == 0:
+        # np.zeros is much faster than np.full, and therefore using it when possible is better.
+        if fill_value is None or (isinstance(fill_value, int) and fill_value == 0):
             return cls(np.zeros(shape=tuple(shape), dtype=dtype, order=order))
         else:
             return cls(np.full(shape=tuple(shape), fill_value=fill_value, dtype=dtype, order=order))
--- a/src/zarr/core/buffer/cpu.py
+++ b/src/zarr/core/buffer/cpu.py
@@ -154,7 +154,7 @@ class NDBuffer(core.NDBuffer):
         order: Literal["C", "F"] = "C",
         fill_value: Any | None = None,
     ) -> Self:
-        if fill_value is None:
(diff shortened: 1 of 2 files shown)


d-v-b:
out of curiosity, is there any performance difference between `np.zeros(...)[:] = fill_value` and `np.full(..., fill_value)`? I would expect not but I have no idea.

src/zarr/core/buffer/cpu.py
```diff
@@ -154,7 +154,7 @@ def create(
         order: Literal["C", "F"] = "C",
         fill_value: Any | None = None,
     ) -> Self:
-        if fill_value is None:
+        if fill_value is None or fill_value == 0:
```
d-v-b, line 157: could we get a comment here that explains why we are branching on the fill value?

ilan-gold:
Not much :/
```python
In [1]: import numpy as np
In [2]: %timeit d = np.zeros((10_000, 10_000), dtype=np.float32, order="C"); d[:] = 1
35.1 ms ± 934 μs per loop (mean ± std. dev. of 7 runs, 10 loops each)
In [3]: %timeit np.full((10_000, 10_000), dtype=np.float32, fill_value=1, order="C")
37.7 ms ± 2.15 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

src/zarr/core/buffer/cpu.py
ilan-gold: yes!

dstansby (approved):

d-v-b:

PR #3083, opened 2025-05-23
Add async oindex method to AsyncArray

TomNicholas:
`Array` has an `.oindex` method (and `.vindex`), but `AsyncArray` has no equivalent. This PR adds one.

I want it for https://github.com/pydata/xarray/pull/10327#discussion_r2103674520

TODO:

### DIFF
--- a/src/zarr/core/array.py
+++ b/src/zarr/core/array.py
@@ -1370,7 +1370,7 @@ class AsyncArray(Generic[T_ArrayMetadata]):
         if prototype is None:
             prototype = default_buffer_prototype()
         indexer = OrthogonalIndexer(selection, self.shape, self.metadata.chunk_grid)
-        return await self._async_array._get_selection(
+        return await self._get_selection(
             indexer=indexer, out=out, fields=fields, prototype=prototype
         )
 
--- a/src/zarr/core/array.py
+++ b/src/zarr/core/array.py
@@ -79,6 +79,7 @@ from zarr.core.indexing import (
     MaskIndexer,
     MaskSelection,
     OIndex,
+    AsyncOIndex,
     OrthogonalIndexer,
     OrthogonalSelection,
     Selection,
@@ -1358,6 +1359,21 @@ class AsyncArray(Generic[T_ArrayMetadata]):
         )
         return await self._get_selection(indexer, prototype=prototype)
 
+    async def get_orthogonal_selection(
+        self,
+        selection: OrthogonalSelection,
+        *,
+        out: NDBuffer | None = None,
+        fields: Fields | None = None,
+        prototype: BufferPrototype | None = None,
+    ) -> NDArrayLike:
+        if prototype is None:
+            prototype = default_buffer_prototype()
+        indexer = OrthogonalIndexer(selection, self.shape, self.metadata.chunk_grid)
+        return await self._async_array._get_selection(
+            indexer=indexer, out=out, fields=fields, prototype=prototype
+        )
+
     async def _save_metadata(self, metadata: ArrayMetadata, ensure_parents: bool = False) -> None:
         """
         Asynchronously save the array metadata.
@@ -1487,6 +1503,12 @@ class AsyncArray(Generic[T_ArrayMetadata]):
             chunk_grid=self.metadata.chunk_grid,
         )
         return await self._set_selection(indexer, value, prototype=prototype)
+
+    @property
+    def oindex(self) -> AsyncOIndex:
+        """Shortcut for orthogonal (outer) indexing, see :func:`get_orthogonal_selection` and
+        :func:`set_orthogonal_selection` for documentation and examples."""
+        return AsyncOIndex(self)
 
     async def resize(self, new_shape: ShapeLike, delete_outside_chunks: bool = True) -> None:
         """
--- a/src/zarr/core/indexing.py
+++ b/src/zarr/core/indexing.py
@@ -28,7 +28,7 @@ import numpy.typing as npt
 from zarr.core.common import product
 
 if TYPE_CHECKING:
-    from zarr.core.array import Array
+    from zarr.core.array import Array, AsyncArray
     from zarr.core.buffer import NDArrayLikeOrScalar
     from zarr.core.chunk_grids import ChunkGrid
     from zarr.core.common import ChunkCoords
@@ -950,13 +950,32 @@ class OIndex:
         return self.array.get_orthogonal_selection(
             cast(OrthogonalSelection, new_selection), fields=fields
         )
-
+    
     def __setitem__(self, selection: OrthogonalSelection, value: npt.ArrayLike) -> None:
         fields, new_selection = pop_fields(selection)
         new_selection = ensure_tuple(new_selection)
         new_selection = replace_lists(new_selection)
         return self.array.set_orthogonal_selection(
             cast(OrthogonalSelection, new_selection), value, fields=fields
+        )
+
+
+@dataclass(frozen=True)
+class AsyncOIndex:
+    array: AsyncArray
+
+    async def getitem(self, selection: OrthogonalSelection | Array) -> NDArrayLike:
+        from zarr.core.array import Array
+
+        # if input is a Zarr array, we materialize it now.
+        if isinstance(selection, Array):
+            selection = _zarr_array_to_int_or_bool_array(selection)
+
+        fields, new_selection = pop_fields(selection)
+        new_selection = ensure_tuple(new_selection)
+        new_selection = replace_lists(new_selection)
+        return await self.array.get_orthogonal_selection(
+            cast(OrthogonalSelection, new_selection), fields=fields
         )
 
 


TomNicholas:
@dcherian suggested making the sync oindex and vindex getitem methods call the new async versions.

EDIT: I think this is already the case?

src/zarr/core/indexing.py
```diff
@@ -1268,6 +1290,30 @@ def __setitem__(
...
+        # if input is a Zarr array, we materialize it now.
+        if isinstance(selection, Array):
+            selection = _zarr_array_to_int_or_bool_array(selection)
+        fields, new_selection = pop_fields(selection)
+        new_selection = ensure_tuple(new_selection)
+        new_selection = replace_lists(new_selection)
+        if is_coordinate_selection(new_selection, self.array.shape):
+            return await self.array.get_coordinate_selection(new_selection, fields=fields)
+        elif is_mask_selection(new_selection, self.array.shape):
+            return await self.array.get_mask_selection(new_selection, fields=fields)
```
TomNicholas, line 1308: I had to add `.get_mask_selection` to `AsyncArray` to cover this codepath. But I only realised I needed to thanks to mypy. This means that this codepath is
- not needed for me right now (I think)
- definitely not covered by the property tests

src/zarr/core/array.py
```diff
@@ -1556,6 +1608,19 @@ async def setitem(
         )
         return await self._set_selection(indexer, value, prototype=prototype)
 
+    @property
+    def oindex(self) -> AsyncOIndex[T_ArrayMetadata]:
```
TomNicholas, line 1612: I chose this API to try to follow this pattern:

- `Array.__getitem__` (exists)
- `Array.oindex.__getitem__` (exists)
- `Array.vindex.__getitem__` (exists)
- `AsyncArray.getitem` (exists)
- `AsyncArray.oindex.getitem` (new)
- `AsyncArray.vindex.getitem` (new)

because python doesn't let you make an async version of the `__getitem__` magic method.

tests/test_properties.py
TomNicholas, line 1: Should I also be adding tests to `test_indexing.py`?

src/zarr/core/array.py
```diff
@@ -1425,6 +1427,56 @@ async def getitem(
         )
         return await self._get_selection(indexer, prototype=prototype)
 
+    async def get_orthogonal_selection(
```
TomNicholas, line 1430: `get_basic_selection` also doesn't exist on `AsyncArray` - should I add that too?

tests/test_properties.py
```diff
@@ -143,6 +164,22 @@ def test_oindex(data: st.DataObject) -> None:
     assert_array_equal(nparray, zarray[:])
 
 
+@pytest.mark.asyncio
+@given(data=st.data())
+@pytest.mark.filterwarnings("ignore::zarr.core.dtype.common.UnstableSpecificationWarning")
+async def test_oindex_async(data: st.DataObject) -> None:
```
dcherian: can we just merge these with the sync versions of these tests. we can test both async and sync within one async test function

src/zarr/core/indexing.py
dcherian: Yes I haven't added mask indexing to the property test suite

src/zarr/core/array.py
dcherian: Can we update the sync versions to use `sync(self._async_array.oindex.getitem)` instead of going to `_get_selection` directly? That will get us some test coverage

tests/test_properties.py
TomNicholas: good idea, merged in https://github.com/zarr-developers/zarr-python/pull/3083/commits/b9197e5a5017f65270f85e796ffab0ff895c8e5b

src/zarr/core/array.py
TomNicholas: That sounds very sensible, but when I try to refactor to do that I find that it causes a problem where `sync()` gets called from within an async coroutine. I wouldn't have thought this should happen, but here's claude's explanation of why it does:

<details>

```python
> Explain how exactly sync is being called from within an async coroutine

⏺ Looking at the error and the code, here's exactly how sync() is being called from within an async coroutine:

  The Call Chain

  1. Test calls zarr array indexing: za[zix] in test_indexing_with_zarr_array
  2. Array.getitem (src/zarr/core/array.py:2566):
  return self.vindex[cast("CoordinateSelection | MaskSelection", selection)]
  3. VIndex.getitem (src/zarr/core/indexing.py:1248):
  return sync(self.array._async_array.vindex.getitem(selection))
  4. sync() tries to run the coroutine but detects it's already in a running loop

  The Problem

  The issue is in the __array__ method call chain:

  1. VIndex.__getitem__ calls sync() with self.array._async_array.vindex.getitem(selection)
  2. This eventually leads to AsyncArray._get_selection() being called
  3. But somewhere in the process, Array.__array__() gets called (src/zarr/core/array.py:2413)
  4. __array__() calls self[...] which goes back to Array.__getitem__
  5. This creates a nested call where sync() is called while already inside an async context

  The Root Cause

  The function _zarr_array_to_int_or_bool_array() at line 85 calls np.asarray(arr), which triggers the __array__ protocol on the zarr array. This causes:

  def _zarr_array_to_int_or_bool_array(arr: Array) -> npt.NDArray[np.intp] | npt.NDArray[np.bool_]:
      if arr.dtype.kind in ("i", "b"):
          return np.asarray(arr)  # <-- This calls arr.__array__()

  When np.asarray() is called on a zarr Array, it calls Array.__array__(), which calls self[...], which eventually calls sync() again - but we're already
  inside a sync() call from the VIndex, creating the nested async context error.

  The original code before the changes avoided this by handling the zarr array conversion within the sync methods directly, rather than delegating to async
  methods that would create this nested sync situation.
```
</details>

tests/test_properties.py
d-v-b: yes i think so

d-v-b:
The new code needs clear tests. Our current indexing tests set an extremely bad example here, because they don't test the indexing classes very specifically. So ignore the rest of the indexing tests, and instead  for each method defined on each new class write a test that checks if that method does what you expect it to do.

src/zarr/core/array.py
dcherian: I guess for indexing with a Zarr array, we should convert to numpy array _before_ the sync call

tests/test_properties.py
TomNicholas: Now added

tests/test_indexing.py
```diff
@@ -1994,3 +1994,65 @@ def test_iter_chunk_regions():
         assert_array_equal(a[region], np.ones_like(a[region]))
         a[region] = 0
         assert_array_equal(a[region], np.zeros_like(a[region]))
+
+
+class TestAsync:
+    @pytest.mark.parametrize(
+        ("indexer", "expected"),
+        [
```
dcherian, line 2002: Can we share this list between the sync and async tests please?

dcherian:
Shall we also have the sync array getitme methods use these async methods?
Example: https://github.com/zarr-developers/zarr-python/blob/6fa9f376a38748decff1f87fa6ed7b576b772a89/src/zarr/core/array.py#L2785-L2792

TomNicholas:
Yea I would like to, but don't fully understand how to get that to work. So I thought I could leave that for a follow-up.

https://github.com/zarr-developers/zarr-python/pull/3083#discussion_r2230737448

dcherian:
OK but presumably that errors means async indexing with Zarr arrays also doesn't work (https://github.com/zarr-developers/zarr-python/pull/3083/files#r2231114456). Can you open an issue to track please?

TomNicholas:
> indexing with Zarr arrays

I didn't even know it was possible to index a zarr array with another zarr array!

> presumably that errors means async indexing with Zarr arrays also doesn't work

Actually I just added a test that seems to show that indexing a zarr array with a (sync) zarr array does work. I also tried indexing a zarr array with the `AsyncArray` (same test but using `._async_array`) which raises with

<details>

```python
        # try indexing with async zarr array
>       result = await async_zarr.oindex.getitem(z2._async_array)

tests/test_indexing.py:2061: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
src/zarr/core/indexing.py:974: in getitem
    return await self.array.get_orthogonal_selection(
src/zarr/core/array.py:1440: in get_orthogonal_selection
    indexer = OrthogonalIndexer(selection, self.shape, self.metadata.chunk_grid)
src/zarr/core/indexing.py:878: in __init__
    dim_indexer = BoolArrayDimIndexer(dim_sel, dim_len, dim_chunk_len)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <[AttributeError("'BoolArrayDimIndexer' object has no attribute 'dim_sel'") raised in repr()] BoolArrayDimIndexer object at 0x107e8e990>
dim_sel = <AsyncArray memory://4427184448/z2 shape=(2,) dtype=bool>, dim_len = 2, dim_chunk_len = 1

    def __init__(self, dim_sel: npt.NDArray[np.bool_], dim_len: int, dim_chunk_len: int) -> None:
        # check number of dimensions
        if not is_bool_array(dim_sel, 1):
            raise IndexError("Boolean arrays in an orthogonal selection must be 1-dimensional only")
    
        # check shape
        if dim_sel.shape[0] != dim_len:
            raise IndexError(
                f"Boolean array has the wrong length for dimension; expected {dim_len}, got {dim_sel.shape[0]}"
            )
    
        # precompute number of selected items for each chunk
        nchunks = ceildiv(dim_len, dim_chunk_len)
        chunk_nitems = np.zeros(nchunks, dtype="i8")
        for dim_chunk_ix in range(nchunks):
            dim_offset = dim_chunk_ix * dim_chunk_len
            chunk_nitems[dim_chunk_ix] = np.count_nonzero(
>               dim_sel[dim_offset : dim_offset + dim_chunk_len]
            )
E           TypeError: 'AsyncArray' object is not subscriptable

src/zarr/core/indexing.py:613: TypeError
```

</details>

If that is supposed to work I can raise an issue for it, but it doesn't seem to be the same `sync` problem that we were discussing before.

dcherian:
The Claude diagnosis points to a sync `np.array` being called in `_zarr_array_to_ints_or_bool`. (https://github.com/zarr-developers/zarr-python/pull/3083#discussion_r2230737448

tests/test_indexing.py
TomNicholas: I'm actually not totally sure what you mean. I could:

1. Add sync versions of the async tests I just added, using the same list of indexers I added in this PR, and not touch existing tests.
2. Edit the existing sync tests to also run the test cases I added for the new async tests.
3. Edit the existing sync tests to merge all these cases into one bigger list, and run both sync and async tests using that.

Which are you asking for? I'm happy to do whatever, but I was trying to avoid stalling this PR by having the scope creep to include refactoring existing tests.

dcherian: aha, my mistake. I thought you had copied over some from the existing sync tests, but you have not.

dcherian:
Not sure what to do about codecov, except add more tests

dcherian:
Thanks for the extra tests!

PR #3087, opened 2025-05-23
issues: add pep-723 to issue template

ianhi:
Adds `pep-723` compliant metadata to the placeholder for the reproducer in the bug report

You can see how it will render by making a new issue here: https://github.com/ianhi/git-expt/issues

image

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -57,7 +57,19 @@ body:
     id: reproduce
     attributes:
       label: Steps to reproduce
-      description: Minimal, reproducible code sample, a copy-pastable example if possible.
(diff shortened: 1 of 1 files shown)


d-v-b:
thoughts on whether the zarr version should point to main branch on github?

d-v-b:
do we want to make it uv-specific? I think you can use regular pypi dependency syntax in the pep-723 block

ianhi:
Good catch. I used the generic form, I also linked to active docs as suggested by the PEP documetn

d-v-b (approved):

d-v-b:
the upstream tests are failing right now because of incompatibility between s3fs and fsspec. I will merge this anyway

d-v-b:
> the upstream tests are failing right now because of incompatibility between s3fs and fsspec. I will merge this anyway

the github UI will not let me merge until the checks pass. I'm trying to sort out the upstream test failures [here](https://github.com/fsspec/s3fs/pull/963)

Issue #3092, opened 2025-05-24
make deprecated positional arguments keyword-only


d-v-b:
we have a lot of functions decorated with [this decorator](https://github.com/zarr-developers/zarr-python/blob/f6742368229b8817a50335c28bb19a10e7a778e9/src/zarr/_compat.py#L12), which by default warns that the decorated function will ~only take keyword-only arguments from version 3.1.0 and onwards.~ not accept certain arguments as positional from 3.1.0 and onwards. 

As we are preparing a 3.1.0 release, we should ~transform these functions to all keyword-only arguments, and~ remove the decoration.

d-v-b:
worth noting that [`create_array`](https://github.com/zarr-developers/zarr-python/blob/f6742368229b8817a50335c28bb19a10e7a778e9/src/zarr/core/array.py#L4298) doesn't use keyword-only arguments (the store is positional). Is that something we want to eventually change?

d-v-b:
similarly, [`create_group`](https://github.com/zarr-developers/zarr-python/blob/f6742368229b8817a50335c28bb19a10e7a778e9/src/zarr/core/group.py#L916-L921) is not all keyword-only, and I don't think we had deprecations for that

d-v-b:
closed via #3225

PR #3093, opened 2025-05-24
remove use of _deprecated_positional_args decorator

d-v-b:
For every function decorated with [`_deprecate_positional_args`](https://github.com/zarr-developers/zarr-python/blob/f6742368229b8817a50335c28bb19a10e7a778e9/src/zarr/_compat.py#L12), I removed the decorator ~and made the function arguments keyword-only.~

This should only go in 3.1.0. 

addresess #3092

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -57,7 +57,22 @@ body:
     id: reproduce
     attributes:
       label: Steps to reproduce
-      description: Minimal, reproducible code sample, a copy-pastable example if possible.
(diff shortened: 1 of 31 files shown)


dstansby (changes requested):
The intent of `_deprecate_positional_args` is not to make every argument positional, but only the ones that are already marked as keyword-only. So no actual arguments should be made keyword only here - the decorator should merely be removed.

As a working example, note how the first call does not raise a warning, but the second call does here:

```python
# /// script
# dependencies = [
#   "zarr==3.0.8",
# ]
# ///
import zarr

root = zarr.group(store='data/example.zarr')
foo = root.create_array('foo', shape=(1000, 100), chunks=(10, 10), dtype='f4')
foo = root.create_array('foo', (1000, 100), chunks=(10, 10), dtype='f4')
```

d-v-b:
thanks for the clarification @dstansby! I'll adjust the PR accordingly.

dstansby:
Woops, deleting the 3.1.0 branch accidentally closed this 🙈 - I think it needs to be reopened targeting the `main` branch

PR #3097, opened 2025-05-25
Fix typing in store tests

dstansby:


dstansby:
I'm confused as to why `mypy` is still failing, as I think I've marked the lines that are erroring with `type: ignore` statements...

PR #3098, opened 2025-05-25
refactor warnings

### LINKED ISSUE #3096: improve control over warnings
As we deprecate old v2 functionality and continue to support Zarr features that are not defined in a specification document (like many codecs), I think zarr-python will continue to emit a lot of warnings in the forseeable future. But these warnings are annoying for users who have already been warned, so we should provide a simple way to turn them off.

Two concrete proposals:
- Make all the warnings we emit zarr-flavored, e.g.:
  
```python
class ZarrFutureWarning(FutureWarning): 
    ...

...
warning.warns('this will change', ZarrFutureWarning)
```

- Provide some top-level functions for silencing warnings. these would be light wrappers around the built in system for suppressing warnings.

d-v-b:
closes #3096 

this PR adds zarr-specific subclasses of `FutureWarning` and `DeprecationWarning`, ~so that we can expose routines for silencing future / deprecation warnings specifically emitted by zarr-python.~

~This PR is a draft until I add the silencing routines later in this PR.~

### DIFF
--- a/changes/3098.misc.rst
+++ b/changes/3098.misc.rst
@@ -1,2 +1,3 @@
-Define Zarr-specific warning classes `ZarrDeprecationWarning` and `ZarrFutureWarning`, that
-subclass `DeprecationWarning` and `FutureWarning`, respectively.
+Define Zarr-specific warning classes that subclass the Python built-in warnings.
+These classes makes it easier to control the visibility of warnings emitted by Zarr Python.
+See `zarr.errors` for these warning classes.


dstansby:
👍 to having our own zarr warning sub-classes, I think I'm 👎 on having some of our own warning silencing helpers, but depends on what the proposal is I guess. So would be good to have separate PRs for those.

d-v-b:
I amended the PR description to remove the goal of providing top-level functions for users to silence zarr warnings. For now, users can use the built-in warnings library to control warnings.

d-v-b:
this will allow users who are tired of seeing zarr-specific warnings to suppress them.

dstansby:
I milestoned as 3.2.0, but I guess this isn't breaking (because the new warnings subclass the old warnings), so could go in 3.1.2 instead?

pyproject.toml
```diff
@@ -409,7 +409,7 @@ filterwarnings = [
     "ignore:Automatic shard shape inference is experimental and may change without notice.*:UserWarning",
     "ignore:The codec .* is currently not part in the Zarr format 3 specification.*:UserWarning",
     "ignore:The dtype .* is currently not part in the Zarr format 3 specification.*:UserWarning",
```
dstansby: Should other warnings in this list also be adapted so they raise the new zarr-specific warnings?

src/zarr/errors.py
```diff
@@ -66,3 +66,9 @@ class NodeTypeValidationError(MetadataValidationError):
     This can be raised when the value is invalid or unexpected given the context,
     for example an 'array' node when we expected a 'group'.
     """
+
+
+class ZarrFutureWarning(FutureWarning): ...
+
+
+class ZarrDeprecationWarning(DeprecationWarning): ...
```
dstansby: These aren't showing up in the API reference docs. Perhaps they need adding to `__all__` at the top of this file to fix this? Would be good if they could get short docstrings too.

tests/test_api.py
```diff
@@ -8,6 +8,8 @@
 import zarr.storage
 from zarr.core.array import init_array
 from zarr.storage._common import StorePath
+from zarr.storage._local import LocalStore
+from zarr.storage._zip import ZipStore
```
dstansby: Why move these imports? If it works, it would be better to import them from `zarr.storage` instaed, since this is the public API path.

src/zarr/errors.py
d-v-b: they're added now

tests/test_api.py
d-v-b: probably due to a mistaken merge conflict resolution. I'll revert

d-v-b: done in https://github.com/zarr-developers/zarr-python/pull/3098/commits/917fd3d70bcdd2edc90f690aa8eb9e55f692c64a

d-v-b:
I removed almost all the ignored warnings from pyproject.toml, at the cost of inserting a bunch of explicit `with pytest.warns()` statements throughout our tests. This was useful because it revealed how many warnings zarr emits during normal usage. For consolidated metadata in particular I think we need to rethink emitting a warning every time its used. I don't think this feature is typically used by individual users but rather libraries like xarray, in which case the best way to convey the off-spec-ness of consolidated metadata is via documentation.

d-v-b:
we should also consider changing our `ZipStore` implementation to hide the `UserWarning` emitted whenever a key is modified.

TomAugspurger:
> removed almost all the ignored warnings from pyproject.toml, at the cost of inserting a bunch of explicit with pytest.warns() statements throughout our tests

Thanks for doing this. I was going to suggest that, but didn't want to assign work to you that's only tangentially related to the PR. I think that being explicit in the tests about where warnings are raised is a good practice.

d-v-b:
consistent with my "test code models user code" theory, excluding every warning individually from the tests has given me a newfound appreciation for all the warnings zarr-python emits on a regular basis

pyproject.toml
d-v-b: this is done

d-v-b:
I don't think the coverage report is very useful here. If we discount it, I think this is ready to go

docs/user-guide/consolidated_metadata.rst
```diff
@@ -27,6 +27,8 @@ In Python, the consolidated metadata is available on the ``.consolidated_metadat
 attribute of the ``GroupMetadata`` object.
 
    >>> import zarr
+   >>> import warnings
+   >>> warnings.filterwarnings("ignore", category=UserWarning)
```
dstansby, line 31: Is it possible to remove these new lines, and instead include the warnings in the docstest output? Instead of hiding the warnings, it would be nice to be honest about them 😄

src/zarr/errors.py
```diff
@@ -61,8 +65,38 @@ class MetadataValidationError(BaseZarrError):
...
+    Specialized exception when the node_type of the metadata document is incorrect.
 
     This can be raised when the value is invalid or unexpected given the context,
     for example an 'array' node when we expected a 'group'.
     """
+
+
+class ZarrFutureWarning(FutureWarning):
+    """
+    A warning raised to indicate when a construct will change semantically in the future.
```
dstansby: What does "construct will change semantically" mean (certainly I don't know!)? Can this be rewritten in less jargon-y language?

src/zarr/errors.py
```diff
@@ -61,8 +65,38 @@ class MetadataValidationError(BaseZarrError):
...
+
+class UnstableSpecificationWarning(ZarrFutureWarning):
+    """
+    A warning raised to indicate that a feature is outside the Zarr specification.
+    """
+
+
+class ZarrDeprecationWarning(DeprecationWarning):
+    """
+    A warning raised to indicate that a construct will be removed in a future release.
```
dstansby:
```suggestion
    A warning raised to indicate that a feature will be removed in a future release.
```

src/zarr/errors.py
```diff
@@ -61,8 +65,38 @@ class MetadataValidationError(BaseZarrError):
...
+
+class ZarrUserWarning(UserWarning):
+    """
+    A warning raised to report problems with user code.
+    """
+
+
+class ZarrRuntimeWarning(RuntimeWarning):
+    """
+    A warning for dubious runtime behavior.
```
dstansby, line 101: What's the difference between a runtime warning and a user warning? It's not clear to me from these descriptions how to choose between them when introducing a new warning.

tests/test_codecs/test_codecs.py
```diff
@@ -290,97 +293,37 @@ async def test_dimension_names(store: Store) -> None:
...
-            shape=(16, 16),
-            chunk_shape=(16, 16),
-            dtype=np.dtype("uint8"),
+@pytest.mark.parametrize(
+    "codecs",
+    [
+        (BytesCodec(), TransposeCodec(order=order_from_dim("F", 2))),
+        (TransposeCodec(order=order_from_dim("F", 2)),),
+    ],
+)
```
dstansby, line 302: Why the change in paramterization here? Could the store parameterization go back as well as the new codec paraetrization?

changes/3098.misc.rst
```diff
@@ -0,0 +1,2 @@
+Define Zarr-specific warning classes ``ZarrDeprecationWarning`` and ``ZarrFutureWarning``, that
+subclass ``DeprecationWarning`` and ``FutureWarning``, respectively.
```
dstansby:
```suggestion
Define Zarr-specific warning classes `ZarrDeprecationWarning` and `ZarrFutureWarning`, that
subclass `DeprecationWarning` and `FutureWarning`, respectively.
```
Because these are in the API docs, use single quotes to link to them.

src/zarr/errors.py
d-v-b: this is nearly identical to the docstring for `FutureWarning`. I think someone interested in writing a warning can handle a little jargon.

docs/user-guide/consolidated_metadata.rst
d-v-b: no, I don't think we want to do that in our docs.

tests/test_codecs/test_codecs.py
d-v-b: the test doesn't depend on storage. it's about metadata validation.

src/zarr/errors.py
d-v-b: these categories exist in the standard library, and I took their docstrings from there.

src/zarr/errors.py
dstansby: Python has this definition, which I much prefer:

> Base category for warnings about deprecated features when those warnings are intended for end users of applications that are written in Python.

So how about this
```suggestion
    A warning intended for end users raised to indicate deprecated features.
```

src/zarr/errors.py
dstansby: Can you give an example of what is "dubious runtime behaviour", or an example of what would be a `RuntimeWarning` vs. what would be a `UserWarning`?

src/zarr/errors.py
d-v-b: sure

src/zarr/errors.py
d-v-b: no i can't, that's an exercise for the reader. I did not invent these categories, these are warnings we are already using in zarr python. I am simply introducing zarr-specific subclasses for each one.

d-v-b: for clarity, this PR contains the result of the following procedure:

- find warnings we are currently emitting
- for each warning, replace the standard library `FooWarning` with `ZarrFooWarning(FooWarning)`. 
 
I am not creating new warning semantics. I am not introducing brand new warnings. These are not breaking changes. I am simply ensuring that every warning we raise is "branded" with its zarrness, so someone can silence all the zarr warnings, without silencing non-zarr warnings.

dstansby: for clarity all I'm trying to do is make new docstrings as clear as possible for develoeprs and users 🤷

d-v-b: I appreciate that but this goal is out of scope for this PR. if you have a concrete suggestion for a better docstring for the warning, I'm happy to take it, but the goal of this PR is not to get the best docstrings. We can do that in a separate PR.

changes/3098.misc.rst
```diff
@@ -0,0 +1,2 @@
+Define Zarr-specific warning classes `ZarrDeprecationWarning` and `ZarrFutureWarning`, that
+subclass `DeprecationWarning` and `FutureWarning`, respectively.
```
dstansby: There's a few more new Warnings that should be listed here, or alternatively just point to `zarr.errors` and say there's a list in the API docs there.

dstansby (approved):
One final comment on the release note, otherwise 👍 Once that's fixed, feel free to self-merge

changes/3098.misc.rst
d-v-b: done in https://github.com/zarr-developers/zarr-python/pull/3098/commits/61ff062135ca9c6dd7a5aa0643b4ebcbb1025af7

PR #3099, opened 2025-05-26
Regression testing

### LINKED ISSUE #3065: time travel testing
In order to track changes in the stored representation of data, we should test the current version of zarr-python against the output of old versions of zarr-python. As discussed in this [zulip topic](https://ossci.zulipchat.com/#narrow/channel/423692-Zarr/topic/ideas.20for.20multi-version.20compatibility.20testing/with/515673134)[#Zarr > ideas for multi-version compatibility testing](https://ossci.zulipchat.com/#narrow/channel/423692-Zarr/topic/ideas.20for.20multi-version.20compatibility.20testing/with/515673134).

d-v-b:
this PR adds regression testing to our support for zarr v2 arrays. The design of the test is to create an array from `main`, then attempt to read + resave that array from `zarr-python==2.18`, then read it again from `main` and check that nothing changed in the round-trip. I implemented the `zarr-python==2.18` dependency via a script with PEP-723-compliant metadata that exposes a CLI for this read -> write operation.

I haven't wired this up to CI yet, because it's not clear when we would want to run it in the life cycle of a random PR, and these tests do take several tens of seconds, so they are not free.

Locally, I got failures round-tripping `datetime`,  `timedelta`, and vlen-string dtypes on the zarr-python 3 side, meaning we could not create those arrays correctly. On the zarr-python 2 side, I got failures attemping to open fixed-length utf32 and ascii arrays saved by zarr-python 3. I will explore how these can be fixed in `main`, and whether these failures persist if I use the new dtype infrastructure in #2874 

closes #3065


d-v-b:
funny, the regression tests are [actually running](https://github.com/zarr-developers/zarr-python/actions/runs/15252756424/job/42893238812) as part of the main test suite, so I guess CI has `uv` installed?

d-v-b:
this code block makes `U` / `S` dtype arrays unreadable from zarr-python 2 by inserting the vlen-utf8 codec where it is not needed: https://github.com/zarr-developers/zarr-python/blob/0dd797f8a7eeab05982a1e4115990200d97d9d98/src/zarr/core/array.py#L771-L777

I think this needs to be removed ASAP

d-v-b:
regression tests were added as part of #2874, so this is not needed

PR #3100, opened 2025-05-27
remove insertion of vlen-string codec for v2 metadata creation

d-v-b:
This PR removes the following code block:

https://github.com/zarr-developers/zarr-python/blob/0dd797f8a7eeab05982a1e4115990200d97d9d98/src/zarr/core/array.py#L771-L777

Neither of the two numpy dtypes that can model variable-length strings are subdtypes of `np.str_`:

```python
>>> np.issubdtype(np.dtypes.ObjectDType(), np.str_)         
False
>>> np.issubdtype(np.dtypes.StringDType(), np.str_)         
False
```

so there is no reason to use vlen-string codec here.

This was revealed by [regression testing against 2.18](https://github.com/zarr-developers/zarr-python/pull/3099). Zarr-python 2.18 cannot read certain arrays generated by zarr-python main because we inserting this codec.

### DIFF
--- a/changes/3100.bugfix.rst
+++ b/changes/3100.bugfix.rst
@@ -1,3 +1,3 @@
-Zarr V2: allow fixed-length string arrays to be created without automatically inserting a
+For Zarr format 2, allow fixed-length string arrays to be created without automatically inserting a
 ``Vlen-UT8`` codec in the array of filters. Fixed-length string arrays do not need this codec. This
 change fixes regressions against Zarr Python 2.18.
--- a/changes/3100.bugfix.rst
+++ b/changes/3100.bugfix.rst
@@ -1,3 +1,3 @@
 For Zarr format 2, allow fixed-length string arrays to be created without automatically inserting a
 ``Vlen-UT8`` codec in the array of filters. Fixed-length string arrays do not need this codec. This
-change fixes regressions against Zarr Python 2.18.
+change fixes a regression where fixed-length string arrays created with Zarr Python 3 could not be read with Zarr Python 2.18.
--- a/changes/3100.bugfix.rst
+++ b/changes/3100.bugfix.rst
@@ -1,0 +1,3 @@
+Zarr V2: allow fixed-length string arrays to be created without automatically inserting a
+``Vlen-UT8`` codec in the array of filters. Fixed-length string arrays do not need this codec. This
+change fixes regressions against Zarr Python 2.18.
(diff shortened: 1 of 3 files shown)


changes/3100.bugfix.rst
```diff
@@ -0,0 +1,3 @@
+Zarr V2: allow fixed-length string arrays to be created without automatically inserting a
```
dstansby, line 1:
```suggestion
For Zarr format 2, allow fixed-length string arrays to be created without automatically inserting a
```

changes/3100.bugfix.rst
```diff
@@ -0,0 +1,3 @@
...
+``Vlen-UT8`` codec in the array of filters. Fixed-length string arrays do not need this codec. This
+change fixes regressions against Zarr Python 2.18.
```
dstansby, line 3:
```suggestion
change fixes a regression where fixed-length string arrays created with Zarr Python 3 could not be read with Zarr Python 2.18.
```

dstansby (commented):
👍 , but probably worth adding a test to make sure this change doesn't accidentally get reverted in the future? I guess just creating a fixed-length string array and checking the list of filters would work.

dstansby:
Independently of cross testing with zarr-python 2, I don't see why we shouldn't put in a test here (or modify an existing one) that checks the filters for a fixed length string array are as expected.

How does one actually hit this bug? I tried the code below, but the list of filters is empty.

```python
# /// script
# dependencies = [
#   "zarr==3.0.8",
#   "numpy"
# ]
# ///
import numpy as np
import zarr

arr = zarr.create_array(store={}, data=np.array(["a", "b"]))
print(arr.filters)
```

d-v-b:
for example, here is a test that checks that the default filters are empty: https://github.com/zarr-developers/zarr-python/blob/0dd797f8a7eeab05982a1e4115990200d97d9d98/tests/test_config.py#L67

I think this is sufficient for this PR. If you feel like another layer of testing is needed, maybe open an issue about that and we can fix it in another PR.

dstansby:
Isn't that just checking the default filters in the config though, not the filters that end up in an array after it's created?

I think this should get a test, so I won't approve/merge, but another dev can merge if they're happy with not adding a test here.

d-v-b:
it doesn't make any sense to me to add a test that specifically checks that the code I deleted has been deleted. Can you find an existing test that we could alter to cover this case?

d-v-b:
fwiw, I do think we have some real gaps in our tests here, so it would be helpful to identify which test needs to be expanded

PR #3103, opened 2025-05-29
Use automatic chunking in array creation routines

dstansby:
The original goal here was to reduce duplication, by pushing the logic for deciding chunk shape down the stack from `zarr.api.asynchronous.create` to `AsyncArray._create`. In the process I discovered this fixes automatically setting chunks with most array creation routines (see the changelog entry), which I think is intended behaviour instad of just setting chunk shape to the input data array shape?


tests/test_api.py
```diff
@@ -1301,3 +1301,9 @@ def test_no_overwrite_load(tmp_path: Path) -> None:
     with contextlib.suppress(NotImplementedError):
         zarr.load(store)
     assert existing_fpath.exists()
+
+
+def test_auto_chunks() -> None:
+    # Check chunks are automatically set to a sensible default
+    a = zarr.ones(shape=(1000, 1000), dtype=np.uint8)
+    assert a.chunks == (500, 500)
```
d-v-b, line 1309: this test will break if we make small changes the auto-chunking behavior, so maybe it's more future-proof to check that the resulting chunks are the ones you would get by feeding the array parameters into the auto chunk routine.

d-v-b: that's reasonable, and the change I'm suggesting can always be made later

PR #3110, opened 2025-06-03
Use context manager for temporary directories in store tests

maxrjones:
This PR creates a pytest fixture so that the temporary directory used in the store tests get cleaned up after use. The motivation is that the current usage of temporary directories was causing the failures seen in https://github.com/zarr-developers/zarr-python/actions/runs/15405315759/job/43346759238?pr=2774.

### DIFF
--- a/tests/test_store/test_core.py
+++ b/tests/test_store/test_core.py
@@ -18,6 +18,26 @@ from zarr.storage._utils import (
 )
 
 
+@pytest.fixture(
+    params=["none", "temp_dir_str", "temp_dir_path", "store_path", "memory_store", "dict"]
+)
+def store_like(request):
+    if request.param == "none":
+        yield None
+    elif request.param == "temp_dir_str":
+        with tempfile.TemporaryDirectory() as temp_dir:
+            yield temp_dir
+    elif request.param == "temp_dir_path":
+        with tempfile.TemporaryDirectory() as temp_dir:
+            yield Path(temp_dir)
+    elif request.param == "store_path":
+        yield StorePath(store=MemoryStore(store_dict={}), path="/")
+    elif request.param == "memory_store":
+        yield MemoryStore(store_dict={})
+    elif request.param == "dict":
+        yield {}
+
+
 @pytest.mark.parametrize("path", ["foo", "foo/bar"])
 @pytest.mark.parametrize("write_group", [True, False])
 @pytest.mark.parametrize("zarr_format", [2, 3])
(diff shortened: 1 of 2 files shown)


maxrjones:
Darn, this helped locally but apparently there are other problems

d-v-b:
I'm surprised pytest's [built-in temporary directory management](https://docs.pytest.org/en/stable/how-to/tmp_path.html#temporary-directory-location-and-retention) is insufficient for our needs here. can we not make the `store_like` fixture take a `tmp_dir` fixture as input (even if some of the stores don't use it)?

dstansby:
Perhaps we're running into https://github.com/pytest-dev/pytest-cov/issues/693? Pinning `pytest<8.4` would be the quickest way of checking.

dstansby:
I'm giving the pytest pin a go over at https://github.com/zarr-developers/zarr-python/pull/3113 - this PR is definitely a good idea regardless 👍

tests/test_store/test_core.py
```diff
@@ -18,6 +18,26 @@
...
+        with tempfile.TemporaryDirectory() as temp_dir:
+            yield Path(temp_dir)
+    elif request.param == "store_path":
+        yield StorePath(store=MemoryStore(store_dict={}), path="/")
+    elif request.param == "memory_store":
+        yield MemoryStore(store_dict={})
+    elif request.param == "dict":
+        yield {}
+
+
```
chuckwondo, line 40: I think you can simplify like so:

```suggestion
@pytest.fixture(
    params=[
        None,
        str,
        Path,
        StorePath(store=MemoryStore(store_dict={}), path="/"),
        MemoryStore(store_dict={}),
        {},
    ],
)
def store_like(request, tmp_path: Path):
    return param if not callable(param := request.param) else param(tmp_path)


```

maxrjones: thanks, this is more concise at the expense of being less obviously interpretable. I'm fine with either version.

chuckwondo: If you want something closer to your original, but slightly simpler:

```
@pytest.fixture(
    params=[None, "temp_dir_str", "temp_dir_path", "store_path", "memory_store", "dict"]
)
def store_like(request, tmp_path: Path):
    if request.param is None:
        return None
    if request.param == "temp_dir_str":
        return str(tmp_path)
    if request.param == "temp_dir_path":
        return tmp_path
    if request.param == "store_path":
        return StorePath(store=MemoryStore(store_dict={}), path="/")
    if request.param == "memory_store":
        return MemoryStore(store_dict={})
    if request.param == "dict":
        return {}
```

dstansby (approved):

dstansby:
Thanks - when we do switch back to pytest 8.4, hopefully this will be a few less errors to fix!

PR #3111, opened 2025-06-03
Cleanups to implementation of `create()`

dstansby:
This cleans up the implementation of `create()` a bit by:

- Fixing the typing of `filters`
- Removing duplicate parsing of data type and compressor (the parsing is already done further down the call stack)
- Moving the parsing of filters to be in the same place (further down the call stack) as parsing of `compressor`

### DIFF
--- a/src/zarr/api/asynchronous.py
+++ b/src/zarr/api/asynchronous.py
@@ -995,8 +995,6 @@ async def create(
         dtype = parse_dtype(dtype, zarr_format)
         if not filters:
             filters = _default_filters(dtype)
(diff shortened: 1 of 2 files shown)


d-v-b:
thank you for this cleanup work :pray:

dstansby:
Since this seems uncontroversial, and has plenty of time for review, I'm going to self merge if CI passes.

dstansby:
Bah, I can't self merge 😄 - can someone review please?

d-v-b (approved):

Issue #3072, opened 2025-05-19
Column-major arrays are returned as row-major


kapadia:
### Zarr version

v3.0.8

### Numcodecs version

0.16.0

### Python Version

3.12.5

### Operating System

macOS

### Installation

uv pip into virtual environment

### Description

I recall using fortran-style arrays with zarr v2 with no issue. Since zarr v3, I've observed that fortran-style arrays are read back as C-style arrays. If this is in fact a bug, and not an error on my part, I'd be happy to contribute a fix. 

### Steps to reproduce

```
import zarr

# Write fortran style array
group = zarr.group("column-major-example.zarr")
array = group.create_array(
    name="example",
    shape=(128, 128),
    dtype="float32",
    order="F",
    dimension_names=("row", "col"),
)
array[:] = 1


# Read back and ensure the order is preserved
group = zarr.open_group("column-major-example.zarr/", mode="r")
array = group["example"]
assert array[:].flags["F_CONTIGUOUS"]
```

ianhi:
I agree that this is a bug.

**zarr format**
This is only an issue if you use `zarr_format=3`, which is the default for zarr-python version 3. If you use `zarr_format=2` when creating the group, the behavior is as expected.


**ordering**

This is pretty clearly not the correct behavior to silently drop this user input. I think the issue is because that in zarr format 3 the `order` is now delegated to the codec as noted in the [spec](https://zarr-specs.readthedocs.io/en/latest/v3/core/index.html)

> order has been replaced by the [transpose](https://zarr-specs.readthedocs.io/en/latest/v3/codecs/transpose/index.html#transpose-codec-v1) codec,


However, it's pretty opaque to me how to actually implement that with zarr-python 3. My best guess would is to add:

`    filters=zarr.codecs.TransposeCodec(order=[1,0]),` to the `create_array` argument (which would ideally happen automatically)  but doing so does not seem to have the desired effect.

ianhi:
These tests have some clues about what the intended behavior is: https://github.com/zarr-developers/zarr-python/blob/57107260291fc9c6f32c95346c321b8a28a6b6e8/tests/test_codecs/test_transpose.py#L14-L19


though so seemingly the read byte order is set via the global config rather than as it was saved on disk? 

https://github.com/zarr-developers/zarr-python/blob/57107260291fc9c6f32c95346c321b8a28a6b6e8/tests/test_codecs/test_transpose.py#L44-L46


Indeed doing so passes your assert:

```python
[...]

PR #3112, opened 2025-06-03
Fix order handling

dstansby:
This is many fixes to how 'order' is handled. See the changelong entries for details of bugs fixed.

Fixes https://github.com/zarr-developers/zarr-python/issues/2948, and helps with https://github.com/zarr-developers/zarr-python/issues/3072.


changes/xxx1.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fixed the error message when passing both ``config`` and ``write_emtpy_chunks`` arguments to reflect the current behaviour (``write_empty_chunks`` takes precedence).
```
d-v-b:
```suggestion
Fixed the error message when passing both ``config`` and ``write_empty_chunks`` arguments to reflect the current behaviour (``write_empty_chunks`` takes precedence).
```

tests/test_array.py
```diff
@@ -1400,52 +1397,6 @@ async def test_data_ignored_params(store: Store) -> None:
         ):
             await create_array(store, data=data, shape=None, dtype=data.dtype, overwrite=True)
 
-    @staticmethod
-    @pytest.mark.parametrize("order", ["C", "F", None])
-    @pytest.mark.parametrize("with_config", [True, False])
-    def test_order(
```
dstansby, line 1453: I've merged this with another existing test to reduce duplication - they were doing a lot of the same things.

src/zarr/core/array.py
```diff
@@ -4521,8 +4518,18 @@ def _parse_keep_array_attr(
                 serializer = "auto"
         if fill_value is None:
             fill_value = data.fill_value
-        if order is None:
+
+        if data.metadata.zarr_format == 2 and zarr_format == 3 and data.order == "F":
+            # Can't set order="F" for v3 arrays
+            warnings.warn(
+                "Zarr format 3 arrays are always stored with order='C'. "
```
d-v-b: I don't think it's accurate to say that Zarr V3 array are stored in C order. In fact, even in Zarr V2 the order parameter can be overridden by the chunk encoding! An array-array codec (filter, in zarr v2 parlance) can take an array and return an arbitrarily re-ordered version of that array to the next codec. The same goes for endianness. 

I'm not really sure how to change this warning in a helpful way, though. We might just say that the `"order"` attribute of a Zarr V2 array does not have a direct analogue in Zarr V3, and so it will be ignored.

dstansby:
I will definitely add more tests to get test coverage up to 100%, so this isn't quite ready yet

src/zarr/core/array.py
```diff
@@ -4267,10 +4266,8 @@ async def init_array(
             chunks_out = chunk_shape_parsed
             codecs_out = sub_codecs
 
-        if config is None:
-            config = {}
-        if order is not None and isinstance(config, dict):
-            config["order"] = config.get("order", order)
+        if order is not None:
+            _warn_order_kwarg()
```
d-v-b, line 4329: codecov is reporting this as untested, could get get a test that checks for the warning?

test.py
```diff
@@ -0,0 +1,11 @@
+import zarr
+
+# Write fortran style array
```
joshmoore: Is this file intentional?

d-v-b: i'm guessing a side-effect of `git add test` instead of `git add tests`

dstansby: Woops, thanks - it was not intentional!

src/zarr/core/array.py
d-v-b: this warning needs to be changed IMO

d-v-b:
this looks good other than the text of the warning that claims that zarr v3 arrays are always stored in C order. That text needs to be made more precise.

d-v-b (changes requested):

src/zarr/core/array.py
dstansby: Good point - I changed the error message to say there's no direct analogue in format 3.

src/zarr/core/array.py
dstansby: Odd, I'm pretty sure I added a test specifically for this line. The test are running again, will see what happens...

dstansby: Yep, this is covered now - looks like your comment is from before when I added the new test.

PR #3113, opened 2025-06-03
Add pytest pin

dstansby:
Maybe this fixes tests?


maxrjones:
thanks, this seems to have resolved the failures in https://github.com/zarr-developers/zarr-python/pull/2774 (I merged in your branch to test)

d-v-b (approved):
pinning seems like the right call. we will need to remove the pin when the underlying problem is fixed.

PR #3119, opened 2025-06-06
Allow Stores to opt out of consolidated metadata.

paraseba:
Some Stores don't benefit from Zarr's consolidated metadata mechanism. These Stores usually implement their own consolidation mechanism, or provide good performance for metadata retrieval out of the box.

These Stores can now implement the `supports_consolidated_metadata` property returning `False`. In this situation, Zarr will silently ignore any requests to consolidate the metadata.

### DIFF
--- a/docs/user-guide/consolidated_metadata.rst
+++ b/docs/user-guide/consolidated_metadata.rst
@@ -118,7 +118,7 @@ removed, or modified, consolidated metadata may not be desirable.
 Stores Without Support for Consolidated Metadata
 ------------------------------------------------
 
-Some stores may want to opt out of the conolidated metadata mechanism. This
+Some stores may want to opt out of the consolidated metadata mechanism. This
 may be for several reasons like:
 
 * They want to maintain read-write consistency, which is challenging with
@@ -127,7 +127,7 @@ may be for several reasons like:
 * They offer good enough performance without need for consolidation.
 
 This type of store can declare it doesn't want consolidation by implementing
-`Store.supports_consolidated_metadata`. For stores that don't support
+`Store.supports_consolidated_metadata` and returning `False`. For stores that don't support
 consolidation, Zarr will:
 
 * Raise an error on `consolidate_metadata` calls, maintaining the store in


src/zarr/api/asynchronous.py
```diff
@@ -194,11 +195,16 @@ async def consolidate_metadata(
...
         The group, with the ``consolidated_metadata`` field set to include
-        the metadata of each child node.
+        the metadata of each child node. If the Store doesn't prefer
+        consolidated metadata, this is function does nothing and returns
+        the group without modifications. See ``Store.supports_consolidated_metadata``.
     """
     store_path = await make_store_path(store, path=path)
 
     group = await AsyncGroup.open(store_path, zarr_format=zarr_format, use_consolidated=False)
+    if not group.store_path.store.supports_consolidated_metadata:
```
dcherian:
```suggestion
    if not store.supports_consolidated_metadata:
```

src/zarr/api/synchronous.py
```diff
@@ -101,7 +102,10 @@ def consolidate_metadata(
     -------
     group: Group
         The group, with the ``consolidated_metadata`` field set to include
-        the metadata of each child node.
+        the metadata of each child node. If the Store doesn't prefer
+        consolidated metadata, this is function does nothing and returns
```
dcherian:
```suggestion
        consolidated metadata, this function does nothing and returns
```

src/zarr/core/group.py
```diff
@@ -503,6 +506,8 @@ async def open(
             to load consolidated metadata from a non-default key.
         """
         store_path = await make_store_path(store)
+        if not store_path.store.supports_consolidated_metadata:
```
dcherian, line 507:
```suggestion
        if not store.supports_consolidated_metadata:
```

src/zarr/api/asynchronous.py
paraseba: I cannot use `store` directly because it's a StoreLike, but I can certainly use `store_path.store`. Will change

src/zarr/core/group.py
paraseba: I don't think this works, store could be a string for example.

src/zarr/api/synchronous.py
paraseba: done

dstansby (changes requested):
Can you give some context in to why this is useful? What's an example of a store that would want to put a blanket ban on consolidating metadata?

Can you also explain what the benefit of this approach is over stores overriding `consolidate_metadata` to raise an error or warning + noop?

If this is implemented this way, I don't think trying to consolidate in this case should silently do nothing, because that provides confusion for a user who is trying to consolidate but nothing is happening. It should raise an error, or at least a warning, so the user knows what is going on, and can correct their code to remove attempts at consolidation.

paraseba:
@dstansby:

> Can you give some context in to why this is useful? What's an example of a store that would want to put a blanket ban on consolidating metadata?

[Icechunk](https://icechunk.io/) is an example of such a Store and my main motivation for this PR. Currently we are having issues with consolidated metadata. It's the default in XArray but it produces inconsistencies. Icechunk users have very concurrent workloads, and the whole point of Icechunk is to maintain consistency. Consolidated metadata breaks consistency before information gets to the Store, so it's too late to fix it.

> Can you also explain what the benefit of this approach is over stores overriding `consolidate_metadata` to raise an error or warning + noop?

Our first approach In Icechunk was to error, see [this PR](https://github.com/earth-mover/icechunk/pull/965) for example. But we checked that it breaks a lot of code. From XArray backend tests (that Icechunk uses) to, more importantly, most code out there that is not passing `consolidated=False` to XArray. We also analyzed the possibility of silently (or with a waning) ignore the metadata key and not store it. But that would break probably the most important invariant for a store: `set(key, x); get(key) == x`. A lot of test and user code depend on that invariant being maintained.

> If this is implemented this way, I don't think trying to consolidate in this case should silently do nothing, because that provides confusion for a user who is trying to consolidate but nothing is happening. It should raise an error, or at least a warning, so the user knows what is going on, and can correct their code to remove attempts at consolidation.

I like the idea of warning in `consolidate_metadata`. That would definitely provide useful information and help the user improve their code. I'll add it. Thank you!

TomAugspurger:
What's the proposed behavior of `open_consolidated` or `open_group(..., consolidated=True)`? IMO that should error, since the user has explicitly asked that consolidated metadata be used.

> From XArray backend tests (that Icechunk uses) to, more importantly, most code out there that is not passing consolidated=False to XArray

Would it be fair to say that this is a case where the `Store` has a better idea of whether consolidated metadata is appropriate? If so, it sounds like we should update xarray to defer to the store for what the default should be.

Note that the top-level Zarr APIs all have a `use_consolidated=None` parameter (except for `open_consolidated`), which means "use consolidated if it happens to be there. Does this sound like what icechunk wants out of xarray?

> These Stores usually implement their own consolidation mechanism, or provide good performance for metadata retrieval out of the box

I see two benefits from consolidated metadata:

1. Reducing the number of reads / HTTP requests to load a bunch of related nodes under a group.
2. Providing an API to navigate a hierarchy that you know is I/O free; Without consolidated metadata, getting the metadata off a child node of a `Group` requires I/O with the store. With consolidated metadata, there's no additional I/O (at the cost of your snapshot potentially not reflecting changes to the store behind your back).

Does that sound right to you @paraseba. Does one or the other provide particular challenges for icechunk?

If icechunk has its own consolidated metadata mechanism, is there any reason not to use it to populate the in-memory `Group.metadata.consolidated_metadata` field (doing this would require new APIs I think)? i.e. is the issue with storing `consolidated_metadata` in the root `zarr.json`, or is the issue with the behavior in zarr-python after the metadata is loaded?

paraseba:
@TomAugspurger:

> What's the proposed behavior of `open_consolidated` or `open_group(..., consolidated=True)`? IMO that should error, since the user has explicitly asked that consolidated metadata be used. 

I'm basically reinterpreting `open_consolidated` to be "open the group in the most performant and correct way the Store supports". If the store does its own consolidation that works better than Zarr's default one, it can implement `supports_consolidated_metadata` to return `False`.

> Would it be fair to say that this is a case where the `Store` has a better idea of whether consolidated metadata is appropriate?

Exactly. Zarr is currently assuming the Store is "dumb" in terms of metadata, which is perfectly true for most stores. But smarter Stores such as Icechunk, need a way to indicate their behavior to Zarr. There is a lot more that could (and probably should) be done in this front. Zarr should be able to "reflect" on the Store capabilities to optimize its algorithms.

> If so, it sounds like we should update xarray to defer to the store for what the default should be.

Yes. In the name of reducing impact I'm reinterpreting `consolidated=True` to "let the Store + Zarr decide what's best". The XArray user has no interest in the details of how the Store makes that happen.

> Note that the top-level Zarr APIs all have a `use_consolidated=None` parameter (except for `open_consolidated`), which means "use consolidated if it happens to be there. Does this sound like what icechunk wants out of xarray?.

Kind of, but is this three-state value useful? Why XArray users would distinguish between `None` and `True`. Don't they want to just say "do what is best and possible"? As an XArray user I want to do what is faster. If tomorrow a store I'm using learns how to do better consolidation I don't want to have to change my code.

> Providing an API to navigate a hierarchy that you know is I/O free; Without consolidated metadata, getting the metadata off a child node of a `Group` requires I/O with the store. With consolidated metadata, there's no additional I/O (at the cost of your snapshot potentially not reflecting changes to the store behind your back).

Yes. There is some risk of people depending on this behavior, but I'd argue they are depending on an implementation detail. The right way to discover hierarchy should be using the `tree` or `members` APIs. Those should provide the ability to do I/O free navigation.
 
> If icechunk has its own consolidated metadata mechanism, is there any reason not to use it to populate the in-memory `Group.metadata.consolidated_metadata` field (doing this would require new APIs I think)?

We analyzed that possibility. A couple of thoughts:

* This generates a (minor) performance impact on calls that are not interested in discovering structure. Now every `get` operation needs to check if it's group metadata and potentially do orders of magnitude more compute to generate the full hierarchy.
* This would be quite surprising behavior from a store, again breaking the invariant of `get after set is identity`.
* Previous point is made worse by `consolidated_metadata` key not being part of the spec.

> i.e. is the issue with storing `consolidated_metadata` in the root `zarr.json`,

To make matters more complex, I understand the key can be stored at any level of the hierarchy, not only the root.

 > or is the issue with the behavior in zarr-python after the metadata is loaded

One of the main issues is there is no way to implement the current API consistently in concurrent scenarios.

TomAugspurger:
> Kind of, but is this three-state value useful? Why XArray users would distinguish between None and True. Don't they want to just say "do what is best and possible"? 

In my experience, yes it's useful for "dumb" stores like blob storage. I want to know whether my `open_dataset` call is going to make one HTTP or one per variable. I wouldn't want it to silently fall back to the slow path, which I gather isn't a problem for Icechunk since all paths are fast :)

I think we're saying essentially the same thing. xarray / Zarr should query the store for the "recommended" way to write stuff by default. I'm probably more bothered more than you by `use_consolidated=True` (silently or with a warning) meaning something other than "use consolidated metadata".  And there's conceptual issue of whether consolidated metadata means "fast mode" or whether it means "write this extra bit of metadata to `zarr.json`.

> Zarr should be able to "reflect" on the Store capabilities to optimize its algorithms.

+100. I have a WIP branch for Stores to read (uncompressed) data into a reallocated buffer without any intermediate copies. I think we'll be seeing this pattern in a few places.

> The right way to discover hierarchy should be using the tree or members APIs. Those should provide the ability to do I/O free navigation.

LMK if you want to move this to a separate issue, but focusing just on zarr-python's in-memory model, if I do something like

```python
group = await open_group(store, path)
async for member in group.members():
    ...
```

do you expect `group.members()` to always make an HTTP request? This is something I struggled with in consolidated metadata. But at the moment, zarr-python *doesn't* make an HTTP request as long as `group.metadata.consolidated_metadata` is set (from the original `open_group`).

> This generates a (minor) performance impact on calls that are not interested in discovering structure. Now every get operation needs to check if it's group metadata and potentially do orders of magnitude more compute to generate the full hierarchy.
> This would be quite surprising behavior from a store, again breaking the invariant of get after set is identity.

I didn't follow these two points, but I feel like we're getting a bit far afield from the PR and it's not clear whether this is worth spending time on. I might ping you on a separate issue if I have time to write up some thoughts.

> Previous point is made worse by consolidated_metadata key not being part of the spec.

Not for lack of trying https://github.com/zarr-developers/zarr-specs/pull/309

---

tl/dr my only qualm right now is with `use_consolidated=True` meaning something other than use consolidated metadata, especially when I *think* we can use `use_consolidated=None` to mean "query the store for what's appropriate".

TomAugspurger:
I just skimmed through https://github.com/earth-mover/icechunk/issues/962 and https://github.com/pydata/xarray/issues/10122 and I think I agree with https://github.com/pydata/xarray/issues/10122#issuecomment-2761798642:

1. Implement `Store.supports_consolidated_metadata` here in zarr-python
2. Update xarray to query that before trying to consolidate metadata

paraseba:
@TomAugspurger:

> I'm probably more bothered more than you by `use_consolidated=True` (silently or with a warning) meaning something other than "use consolidated metadata". 

This is a very valid point. I think things still work for Icechunk, if I change `open` to fail on `use_consolidated=True` on a non-supporting Store. As long as I can preserve the `None` behavior of asking the Store what it wants. Would that work for you? 

> do you expect `group.members()` to always make an HTTP request? This is something I struggled with in consolidated metadata. But at the moment, zarr-python _doesn't_ make an HTTP request as long as `group.metadata.consolidated_metadata` is set (from the original `open_group`).

Consistency issues aside, I think this is the right behavior. And this is why I think depending on the `consolidated_metadata` metadata key as a user is not necessary.

TomAugspurger:
> Icechunk, if I change open to fail on use_consolidated=True on a non-supporting Store. 

Yep, I think that'll be perfect.

paraseba:
@TomAugspurger letting the Store select behavior when `use_consolidated=None` has been implemented. I like this version more.

src/zarr/api/asynchronous.py
```diff
@@ -194,11 +195,21 @@ async def consolidate_metadata(
...
-        the metadata of each child node.
+        the metadata of each child node. If the Store doesn't prefer
+        consolidated metadata, this is function does nothing and returns
+        the group without modifications. See ``Store.supports_consolidated_metadata``.
     """
     store_path = await make_store_path(store, path=path)
 
     group = await AsyncGroup.open(store_path, zarr_format=zarr_format, use_consolidated=False)
+    if not store_path.store.supports_consolidated_metadata:
+        store_name = type(store_path.store).__name__
```
TomAugspurger, line 205: I would expect this to raise an exception, probably a `RuntimeError` or perhaps a [`TypeError`](https://docs.python.org/3/library/exceptions.html#TypeError):

> This exception may be raised by user code to indicate that an attempted operation on an object is not supported, and is not meant to be.

paraseba:
>     1. Raise an exception on `consolidate_metadata` on a store when `store.supports_consolidated_metadata` is False

This is what I mentioned before. If I do this, the change no longer works for Icechunk. XArray will try to consolidate by default, and Icechunk has no way to stop it. User code breaks until they explicitly ask for `consolidated=False`.

> 
>     2. Raise an exception on `open_group(store, ..., use_consolidated=True)` when `store.supports_consolidated_metadata` is False

This is implemented, I believe, because it just calls `AsyncGroup.open`. Do you see a problem with the current behavior?
 
> 
>     3. Update [`AsyncGroup.open`](https://github.com/zarr-developers/zarr-python/blob/6193fd9b3bd23f0aa9676b489e62f224e4325b3e/src/zarr/core/group.py#L476) to
> 
> 
>     * check `if use_consolidated is True and not store.use_consolidated:` to raise an error

This is implemented. 


>     * _not_ make a request for the consolidated metadata key when `use_consolidated=None and not store.use_consolidated` (effectively turning `use_consolidated=None` into `use_consolidated=False`

Yes, I believe this is done https://github.com/zarr-developers/zarr-python/pull/3119/files#diff-2ad4cb8e97a61d216098a06921255a4be58352e51ae4a857b731b8e12080fea4R516

TomAugspurger:
> This is what I mentioned before. If I do this, the change no longer works for Icechunk. XArray will try to consolidate by default, and Icechunk has no way to stop it. User code breaks until they explicitly ask for consolidated=False.

We should update xarray to check `store.supports_consolidated_metadata` and only consolidate if that's True. I can take care of that if you want.

paraseba:
> We should update xarray to check `store.supports_consolidated_metadata` and only consolidate if that's True. I can take care of that if you want.

I see. That makes sense to me. I can make the change in here. Do you think the XArray change is reasonably uncontroversial?

TomAugspurger:
@TomNicholas suggested this in https://github.com/pydata/xarray/issues/10122#issuecomment-2761798642

> there should be an actual property on zarr.Store which expresses this preference for each implementation. Then xarray can just quietly look at that instead of making users think about it.

and there didn't seem to be any pushback. I can chime in over there with a +1 and request that that'd be acceptable.

jhamman:
@dstansby - this has advanced quite a bit from your last review. Can you re-review or clear your requested changes?

src/zarr/abc/store.py
```diff
@@ -264,6 +264,18 @@ async def _set_many(self, values: Iterable[tuple[str, Buffer]]) -> None:
         """
         await gather(*starmap(self.set, values))
 
+    @property
+    def supports_consolidated_metadata(self) -> bool:
+        """
+        Does the store support and benefit from consolidated metadata?.
```
dstansby:
```suggestion
        Does the store support consolidated metadata?.
```

src/zarr/abc/store.py
```diff
@@ -264,6 +264,18 @@ async def _set_many(self, values: Iterable[tuple[str, Buffer]]) -> None:
...
+
+        If it doesn't Zarr will ignore requests to consolidate the metadata.
```
dstansby:
```suggestion
        If it doesn't an error will be raised on requests to consolidate the metadata.
```

src/zarr/abc/store.py
```diff
@@ -264,6 +264,18 @@ async def _set_many(self, values: Iterable[tuple[str, Buffer]]) -> None:
...
+        Stores that would return `True` are the ones that implement their own
+        consolidation mechanism, that allows fast querying of metadata keys.
```
dstansby:
```suggestion
        Returning `False` can be useful for stores which implement their own
        consolidation mechanism outside of the zarr-python implementation.
```

src/zarr/api/asynchronous.py
```diff
@@ -174,7 +174,8 @@ async def consolidate_metadata(
     Consolidate the metadata of all nodes in a hierarchy.
 
     Upon completion, the metadata of the root node in the Zarr hierarchy will be
-    updated to include all the metadata of child nodes.
+    updated to include all the metadata of child nodes. For Stores that prefer
+    not to use consolidated metadata, this operation raises a ``TypeError``.
```
dstansby:
```suggestion
    updated to include all the metadata of child nodes. For Stores that do
    not support consolidated metadata, this operation raises a ``TypeError``.
```

Alternatively, you could make use of the ["Raises" docstring section](https://numpydoc.readthedocs.io/en/latest/format.html#raises) for this information.

src/zarr/api/asynchronous.py
```diff
@@ -194,10 +195,18 @@ async def consolidate_metadata(
     -------
     group: AsyncGroup
         The group, with the ``consolidated_metadata`` field set to include
-        the metadata of each child node.
+        the metadata of each child node. If the Store doesn't prefer
+        consolidated metadata, this is function does nothing and returns
+        the group without modifications. See ``Store.supports_consolidated_metadata``.
```
dstansby:
```suggestion
        the metadata of each child node. If the Store doesn't support
        consolidated metadata, this is function does raises a `TypeError`.
```

src/zarr/api/synchronous.py
```diff
@@ -81,7 +81,8 @@ def consolidate_metadata(
     Consolidate the metadata of all nodes in a hierarchy.
 
     Upon completion, the metadata of the root node in the Zarr hierarchy will be
-    updated to include all the metadata of child nodes.
+    updated to include all the metadata of child nodes. For Stores that prefer
+    not to use consolidated metadata, this operation does nothing.
```
dstansby:
```suggestion
    updated to include all the metadata of child nodes. For Stores that do
    not use consolidated metadata, this operation raises a `TypeError`.
```

tests/test_metadata/test_consolidated.py
```diff
@@ -651,3 +651,38 @@ async def test_consolidated_metadata_encodes_special_chars(
...
+class NonConsolidatedStore(zarr.storage.MemoryStore):
+    """A store that doesn't support consolidated metadata"""
+
+    @property
+    def supports_consolidated_metadata(self) -> bool:
+        return False
+
+
+async def test_consolidate_metadata_raises_for_self_consolidating_stores():
+    """Verify calling consolidate_metadata on a non supporting stores does nothing"""
```
dstansby:
```suggestion
    """Verify calling consolidate_metadata on a non supporting stores raises an error."""
```

TomAugspurger (approved):
Thanks @paraseba, looks good once the docs are updated.

paraseba:
@dstansby I updated your suggested docstrings and a few more places. Thank you!

docs/user-guide/consolidated_metadata.rst
```diff
@@ -114,3 +114,23 @@ removed, or modified, consolidated metadata may not be desirable.
    metadata.
 
 .. _Consolidated Metadata: https://github.com/zarr-developers/zarr-specs/pull/309
+
+Stores Without Support for Consolidated Metadata
+------------------------------------------------
+
+Some stores may want to opt out of the conolidated metadata mechanism. This
```
dstansby:
```suggestion
Some stores may want to opt out of the consolidated metadata mechanism. This
```

docs/user-guide/consolidated_metadata.rst
```diff
@@ -114,3 +114,23 @@ removed, or modified, consolidated metadata may not be desirable.
...
+may be for several reasons like:
+
+* They want to maintain read-write consistency, which is challenging with
+  consolidated metadata.
+* They have their own consolidated metadata mechanism.
+* They offer good enough performance without need for consolidation.
+
+This type of store can declare it doesn't want consolidation by implementing
+`Store.supports_consolidated_metadata`. For stores that don't support
```
dstansby:
```suggestion
`Store.supports_consolidated_metadata` and returning `False`. For stores that don't support
```

dstansby (approved):

PR #3127, opened 2025-06-10
Fix `zarr.save` for given `path` and multiple `args` arrays

brokkoli71:
When `save` has an argument `path=some/path/` and multiple arrays in `args`, the path resulted in `some/path/some/path` due to using the `path` argument twice while building the array path. This gets fixed in this PR

TODO:


src/zarr/api/asynchronous.py
```diff
@@ -505,13 +505,12 @@ async def save_group(
...
     aws = []
     for i, arr in enumerate(args):
-        _path = f"{path}/arr_{i}" if path is not None else f"arr_{i}"
         aws.append(
             save_array(
                 store_path,
                 arr,
                 zarr_format=zarr_format,
-                path=_path,
+                path=f"arr_{i}",
```
dstansby, line 513: Doesn't this now ignore the `path` that's passed in? Shouldn't it be  `f"{path}/arr_{i}"`?

brokkoli71: the `path` will be applied in `make_store_path`

brokkoli71: if I'm not mistaken, the problem was that the `path` was added in `make_store_path` to the store_path AND with `path=f"{path}/arr_{i}"`

dstansby: I thought the issue was the path being applied multiple times, but it still needs to be applied once, no? Either way, the right way to find out would be to add a test that re-loads the saved group and checks that the array paths are as expected.

brokkoli71: I think that is what I did by adding `path` to `test_save` to check different `path` values. That test failed before my changes in asynchronous.py

brokkoli71: The issue was that it was applied two times. One time in line 485:
```python
store_path = await make_store_path(store, path=path, mode="w", storage_options=storage_options)
```
and line 508:
```python
_path = f"{path}/arr_{i}" if path is not None else f"arr_{i}"
```
Therefore the `store_path`, as well as the `path` argument for `save_array` was extended by the `path` argument of `save_group`

brokkoli71: Therefore I removed it from the `path` while keeping it in `store_path`. I would be open to do it the other way round and not extend the `store_path`, but the `path` instead

dstansby: That makes sense - thanks for the explanation!

dstansby (approved):

PR #3128, opened 2025-06-10
Fix `zarr.open` default for argument `mode` when `store` is `read_only`

### LINKED ISSUE #3042: `zarr.load` fails with `ZipStore` unless `read_only=False`
### Zarr version

3.0.7

### Numcodecs version

0.16.0

### Python Version

3.13.1

### Operating System

WSL2

### Installation

uv pip install zarr

### Description

`zarr.load` fails when reading data from `ZipStore`.

`ValueError: Store is read-only but mode is 'a'`

### Steps to reproduce

```python
import sys

import numcodecs
import zarr

file = "test_data.zip"


def main():
    with ZipStore(file, mode="r", read_only=False) as zs:
        data = zarr.load(store=zs, path="some/data")  # This works
        print(data.shape)

    with ZipStore(file, mode="r") as zs:
        data = zarr.load(store=zs, path="some/data")  # This does not work
        print(data.shape)


if __name__ == "__main__":
    print(zarr.__version__)
    print(numcodecs.__version__)
    print(sys.version)
    main()


```

### Additional output

_No response_

brokkoli71:
- fixes #3042

If `zarr.open` receives a `Store` Object, its `read_only` value should be considered for the default of the `mode` parameter to avoid errors like in #3042 

TODO:


src/zarr/api/asynchronous.py
```diff
@@ -324,7 +325,11 @@ async def open(
         Return type depends on what exists in the given store.
     """
     zarr_format = _handle_zarr_version_or_format(zarr_version=zarr_version, zarr_format=zarr_format)
-
+    if mode is None:
+        if isinstance(store, Store) and store.read_only:
```
dstansby: This should also check for `StorePath`, and if it is a `StorePath` extract the `Store` and follow the same logic.

brokkoli71:
Thanks for the feedback @dstansby !
What do you think of the changes?

dstansby (approved):

d-v-b:

PR #3129, opened 2025-06-10
Fix infinite reload loop on latest release notes page

dstansby:
Currently https://zarr.readthedocs.io/en/latest/release-notes.html just infinitely reload loops, due to changes in https://github.com/zarr-developers/zarr-python/pull/3106 🙈 - this should fix it.


joshmoore (approved):
Thanks for the quick fix, @dstansby

PR #3133, opened 2025-06-13
Update pre-commit hooks + mypy fixes

dstansby:
Replaces https://github.com/zarr-developers/zarr-python/pull/3109.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -22,7 +22,7 @@ repos:
       - id: check-yaml
       - id: trailing-whitespace
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v1.15.0
+    rev: v1.16.1
     hooks:
       - id: mypy
         files: src|tests
--- a/src/zarr/core/buffer/core.py
+++ b/src/zarr/core/buffer/core.py
@@ -93,7 +93,7 @@ class NDArrayLike(Protocol):
 
     def all(self) -> bool: ...
 
-    def __eq__(self, other: object) -> Self:  # type: ignore[explicit-override, override]
+    def __eq__(self, other: object) -> Self:  # type: ignore[override]
         """Element-wise equal
 
         Notes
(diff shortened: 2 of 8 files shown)


dstansby:
@d-v-b any ideas what's going on with the typing error here? I can't wrap my head around why it's erroring. Perhaps we should just stick an # type: ignore comment on the failing line?

DimitriPapadopoulos:
I wonder whether this would fix the mypy error:
```diff
-    if dtype_category == "variable-length-string":
-        zdtype = VariableLengthUTF8()
-    else:
-        zdtype = Int8()
+    if dtype_category == "variable-length-string":
+        zdtype = VariableLengthUTF8
+    else:
+        zdtype = Int8
```

.pre-commit-config.yaml
```diff
@@ -22,7 +22,7 @@ repos:
       - id: check-yaml
       - id: trailing-whitespace
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v1.15.0
+    rev: v1.16.0
```
DimitriPapadopoulos: Bump to `1.16.1`?

dstansby:
The issue with the dtype mypy error was the variable length string type is not actually a valid `ZarrDtype`. Since there are existing `#type: ignore` comments to acocunt for this, I added another one here.

d-v-b:
> @d-v-b any ideas what's going on with the typing error here? I can't wrap my head around why it's erroring. Perhaps we should just stick an # type: ignore comment on the failing line?

which specific typing error? a basic challenge with `VariableLengthUTF8` is that it's conditionally defined: if numpy < 2 is installed, we define 1 version of the class (using the numpy object dtype) and if numpy > 2 is installed, we define a different version of the class (using np.dtypes.StringDType). That can be why mypy singles this type out as unknown

dstansby:
This line here: https://github.com/zarr-developers/zarr-python/pull/3133/files#diff-da24b355349b53e4c8047034767d008e4c2ce9ea6c8933a3acd50d0828f00528R334 - if you remove the typing ignores, it gives an error.

d-v-b:
> This line here: https://github.com/zarr-developers/zarr-python/pull/3133/files#diff-da24b355349b53e4c8047034767d008e4c2ce9ea6c8933a3acd50d0828f00528R334 - if you remove the typing ignores, it gives an error.

yeah, mypy is confused there, so the type ignore is correct. This will get simpler in a few months when we drop numpy < 2.

PR #3138, opened 2025-06-17
Add with_read_only() convenience method to store

maxrjones:
This PR adds a `Store.with_read_only()` convenience method to create a new store with a different read mode. By default, the method raises a `NotImplementedError`. The motivation is to make it simpler for downstream libraries to adapt to the changes in https://github.com/zarr-developers/zarr-python/pull/3068. For example, this would make it much simpler to address the [269 failures in Xarray's upstream test suite](https://github.com/pydata/xarray/actions/runs/15646254224/job/44084198866) and the [failures in VirtualiZarr's upstream test suite](https://github.com/zarr-developers/VirtualiZarr/actions/runs/15666293033/job/44130820539).

This is similar to the functionality added in https://github.com/zarr-developers/zarr-python/pull/2249 and removed in https://github.com/zarr-developers/zarr-python/pull/2442. @jhamman would you be able to offer any details on the motivation for previously removing the similar functionality?

TODO:


src/zarr/abc/store.py
```diff
@@ -83,6 +83,22 @@ async def open(cls, *args: Any, **kwargs: Any) -> Self:
         await store._open()
         return store
 
+    def with_read_only(self, read_only: bool = False) -> Store:
+        """
+        Return a new store of the same type pointing to the same location with the specified read_only state.
```
dstansby:
```suggestion
        Return a new store with a new read_only setting.
        
        The new store points to the same location with the specified new read_only state.
```
To keep the first sentence short?

src/zarr/abc/store.py
```diff
@@ -83,6 +83,22 @@ async def open(cls, *args: Any, **kwargs: Any) -> Self:
...
+        The returned Store is not automatically opened.
```
dstansby:
```suggestion
        The returned Store is not automatically opened, and this store is
        not automatically closed.
```

src/zarr/abc/store.py
```diff
@@ -83,6 +83,22 @@ async def open(cls, *args: Any, **kwargs: Any) -> Self:
...
+        Parameters
+        ----------
+        read_only
+            If True, the store will be created in read-only mode. Defaults to False.
+
+        Returns
+        -------
+            A new store of the same type with the new read only attribute.
+        """
+        raise NotImplementedError("with_read_only is not implemented for this store type.")
```
dstansby: Is there a way to get the store name and include it in the error message?

src/zarr/abc/store.py
```diff
@@ -83,6 +83,22 @@ async def open(cls, *args: Any, **kwargs: Any) -> Self:
         await store._open()
         return store
 
+    def with_read_only(self, read_only: bool = False) -> Store:
```
dstansby, line 86: Given the intent is clear from the method name, should we make `read_only` positional only? (ie., force `with_read_only(True)` and forbid `with_read_only(read_only=True)`?

src/zarr/abc/store.py
maxrjones: The codecov check warning is picking up an old version of the tests - https://github.com/zarr-developers/zarr-python/pull/3138/checks?check_run_id=44355956277. This is actually covered.

src/zarr/testing/store.py
```diff
@@ -149,6 +149,58 @@ async def test_read_only_store_raises(self, open_kwargs: dict[str, Any]) -> None
         ):
             await store.delete("foo")
 
+    async def test_with_read_only_store(self, open_kwargs: dict[str, Any]) -> None:
+        kwargs = {**open_kwargs, "read_only": True}
+        store = await self.store_cls.open(**kwargs)
```
dstansby, line 154:
```suggestion
        store = await self.store_cls.open(read_only=True, **open_kwargs)
```

src/zarr/testing/store.py
```diff
@@ -149,6 +149,58 @@ async def test_read_only_store_raises(self, open_kwargs: dict[str, Any]) -> None
...
+        # Test that you cannot write to a read-only store
+        with pytest.raises(
+            ValueError, match="store was opened in read-only mode and does not support writing"
+        ):
+            await store.set("foo", self.buffer_cls.from_bytes(b"bar"))
+
+        # Check if the store implements with_read_only
+        try:
+            writer = store.with_read_only(read_only=False)
+
```
dstansby: To avoid the big `try` block, I'd be tempted to put the except block here, then put the rest of the code for testing stores that implement `store.with_read_only` underneath in a dedented block.

dstansby (commented):
Nice - looks great overall, have left some minor suggestions.

src/zarr/abc/store.py
maxrjones: I find the keyword argument easier to document and understand even in this case with some redundancy, so my slight preference would be to allow both positional and keyword arguments.

src/zarr/testing/store.py
maxrjones: This leads to `TypeError: zarr.abc.store.Store.open() got multiple values for keyword argument 'read_only'`

src/zarr/abc/store.py
maxrjones: I realized a reviewer would suggest this in between your comment writing and review posting 😆 [`11abad1` (#3138)](https://github.com/zarr-developers/zarr-python/pull/3138/commits/11abad120bc5f5237f49c3824b5fe6ef0ecd75d5)

src/zarr/testing/store.py
maxrjones: thanks, done in [`dff7034` (#3138)](https://github.com/zarr-developers/zarr-python/pull/3138/commits/dff7034c0a30a30ad9050da0a90b1cf53a531cb6)

dstansby (approved):

d-v-b:

PR #3139, opened 2025-06-18
Ignore new ruff rules (ruff 0.12)

DimitriPapadopoulos:
TODO:


dstansby:
What do these rules do? Can we fix for them instead of ignoring them?

DimitriPapadopoulos:
We could, but some maintainers don't like them:
```
PT011 `pytest.raises(ValueError)` is too broad, set the `match` parameter or use a more specific exception
PT012 `pytest.raises()` block should contain a single simple statement
PT030 `pytest.warns(DeprecationWarning)` is too broad, set the `match` parameter or use a more specific warning
PT031 `pytest.warns()` block should contain a single simple statement
```

dstansby:
In my opinion:

> PT011 `pytest.raises(ValueError)` is too broad, set the `match` parameter or use a more specific exception
> PT030 `pytest.warns(DeprecationWarning)` is too broad, set the `match` parameter or use a more specific warning

We should definitely enforce these. 

> PT012 `pytest.raises()` block should contain a single simple statement
> PT031 `pytest.warns()` block should contain a single simple statement

I'd be interested to see what needs fixing to enable these, and if it's reasonable changes enforce them.

dstansby:
If you'd be up for it, a PR to fix the first two (adding `match` to pytest.raises) would be great, and a separate PR to see what fixing the second two would be good too.

DimitriPapadopoulos:
I'd rather avoid adding the `match` argument in this PR, changes are extensive as far as I can remember.

As for the second two, they imply transforming large `pytest.raises()`/`pytest.warns()` blocks into a small block that contains only the statement that raises or emits the warning. The rest being is moved outside the block.

DimitriPapadopoulos:
Superseded by #3214.

PR #3140, opened 2025-06-18
Suppress FileNotFoundError when deleting keys in the obstore adapter

### LINKED ISSUE #3136: FileNotFoundError when writing an array that contains empty chunks
### Zarr version

v3.0.8

### Numcodecs version

v0.16.1

### Python Version

3.13.1

### Operating System

Mac

### Installation

using uv

### Description

There seems to be an issue with the zarr + obstore integration, which results in a `FileNotFoundError` when trying to write data into a newly created, empty array in case some of the chunks trying to write are empty (all zeros).

Curiously though the issue only surfaces when using a `GCSStore` or a `LocalStore`, for `S3` it seems to work as expected.

```python
import numpy as np
import zarr
from obstore.store import LocalStore

zarr_store = ObjectStore(LocalStore("test_zarr_store"))  # issue also comes up with GCSStore
arr = zarr.create_array(zarr_store, name="arr", shape=(5, 128, 128), dtype=np.uint16, chunks=(1, 32, 32))

# this will fail with: FileNotFoundError: Object at location debug/zarr_issue/arr/c/0/1/2 not found
arr[0, :, :] = np.zeros((128, 128), dtype=np.uint16)
```

<details>

<summary>FileNotFoundError, expand for Traceback</summary>

```
  File "/Users/lukasbindreiter/Documents/tilebox/playground/zarr/gcs_issue.py", line 20, in <module>
    arr[0, :, :] = np.zeros((128, 128), dtype=np.uint16)
    ~~~^^^^^^^^^
  File "/Users/lukasbindreiter/Library/Caches/uv/environments-v2/repr-6df0040f128f3d90/lib/python3.13/site-packages/zarr/core/array.py", line 2553, in __setitem__
    self.set_orthogonal_selection(pure_selection, value, fields=fields)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/lukasbindreiter/Library/Caches/uv/environments-v2/repr-6df0040f128f3d90/lib/python3.13/site-packages/zarr/_compat.py", line 43, in inner_f
    return f(*args, **kwargs)
  File "/Users/lukasbindreiter/Library/Caches/uv/environments-v2/repr-6df0040f128f3d90/lib/python3.13/site-packages/zarr/core/array.py", line 3009, in set_orthogonal_selection
    return sync(
        self._async_array._set_selection(indexer, value, fields=fields, prototype=prototype)
    )
  File "/Users/lukasbindreiter/Library/Caches/uv/environments-v2/repr-6df0040f128f3d90/lib/python3.13/site-packages/zarr/core/sync.py", line 163, in sync
    raise return_result
  File "/Users/lukasbindreiter/Library/Caches/uv/environments-v2/repr-6df0040f128f3d90/lib/python3.13/site-packages/zarr/core/sync.py", line 119, in _runner
    return await coro
           ^^^^^^^^^^
  File "/Users/lukasbindreiter/Library/Caches/uv/environments-v2/repr-6df0040f128f3d90/lib/python3.13/site-packages/zarr/core/array.py", line 1446, in _set_selection
    await self.codec_pipeline.write(
    ...<12 lines>...
    )
  File "/Users/lukasbindreiter/Library/Caches/uv/environments-v2/repr-6df0040f128f3d90/lib/python3.13/site-packages/zarr/core/codec_pipeline.py", line 481, in write
    await concurrent_map(
    ...<6 lines>...
    )
  File "/Users/lukasbindreiter/Library/Caches/uv/environments-v2/repr-6df0040f128f3d90/lib/python3.13/site-packages/zarr/core/common.py", line 76, in concurrent_map
    return await asyncio.gather(*[asyncio.ensure_future(run(item)) for item in items])
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/lukasbindreiter/Library/Caches/uv/environments-v2/repr-6df0040f128f3d90/lib/python3.13/site-packages/zarr/core/common.py", line 74, in run
    return await func(*item)
           ^^^^^^^^^^^^^^^^^
  File "/Users/lukasbindreiter/Library/Caches/uv/environments-v2/repr-6df0040f128f3d90/lib/python3.13/site-packages/zarr/core/codec_pipeline.py", line 431, in write_batch
    await concurrent_map(
    ...<8 lines>...
    )
  File "/Users/lukasbindreiter/Library/Caches/uv/environments-v2/repr-6df0040f128f3d90/lib/python3.13/site-packages/zarr/core/common.py", line 76, in concurrent_map
    return await asyncio.gather(*[asyncio.ensure_future(run(item)) for item in items])
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/lukasbindreiter/Library/Caches/uv/environments-v2/repr-6df0040f128f3d90/lib/python3.13/site-packages/zarr/core/common.py", line 74, in run
    return await func(*item)
           ^^^^^^^^^^^^^^^^^
  File "/Users/lukasbindreiter/Library/Caches/uv/environments-v2/repr-6df0040f128f3d90/lib/python3.13/site-packages/zarr/core/codec_pipeline.py", line 427, in _write_key
    await byte_setter.delete()
  File "/Users/lukasbindreiter/Library/Caches/uv/environments-v2/repr-6df0040f128f3d90/lib/python3.13/site-packages/zarr/storage/_common.py", line 165, in delete
    await self.store.delete(self.path)
  File "/Users/lukasbindreiter/Library/Caches/uv/environments-v2/repr-6df0040f128f3d90/lib/python3.13/site-packages/zarr/storage/_obstore.py", line 184, in delete
    await obs.delete_async(self.store, key)
FileNotFoundError: Object at location debug/zarr_issue/arr/c/0/1/2 not found: Error performing DELETE https://storage.googleapis.com/workflow%2Dcache%2D15c9850/debug%2Fzarr%5Fissue%2Farr%2Fc%2F0%2F1%2F2 in 72.532333ms - Server returned non-2xx status code: 404 Not Found: <?xml version='1.0' encoding='UTF-8'?><Error><Code>NoSuchKey</Code><Message>The specified key does not exist.</Message><Details>No such object: workflow-cache-15c9850/debug/zarr_issue/arr/c/0/1/2</Details></Error>

Debug source:
NotFound {
    path: "debug/zarr_issue/arr/c/0/1/2",
    source: RetryError {
        method: DELETE,
        uri: Some(
            https://storage.googleapis.com/workflow%2Dcache%2D15c9850/debug%2Fzarr%5Fissue%2Farr%2Fc%2F0%2F1%2F2,
        ),
        retries: 0,
        max_retries: 10,
        elapsed: 72.532333ms,
        retry_timeout: 180s,
        inner: Status {
            status: 404,
            body: Some(
                "<?xml version='1.0' encoding='UTF-8'?><Error><Code>NoSuchKey</Code><Message>The specified key does not exist.</Message><Details>No such object: workflow-cache-15c9850/debug/zarr_issue/arr/c/0/1/2</Details></Error>",
            ),
        },
    },
}
```
</details>

The issue only surfaces when one (or all of the chunks) contains all zeros, if I fill up my array with random values it works:

```python
# fails with FileNotFoundError
arr[0, :, :] = np.zeros((128, 128), dtype=np.uint16)

# works
arr[0, :, :] = np.random.randint(0, 100, size=(128, 128), dtype=np.uint16)
```

It makes sense to me that `zarr` would delete chunks here, instead of writing ones with only zeros in them.
However, it seems that instead of a `delete` operation it should use a `delete_if_exists`, right?

### Steps to reproduce

```python
# /// script
# requires-python = ">=3.11"
# dependencies = [
#   "obstore==0.6.0",
# ]
# ///
#

import shutil
from pathlib import Path

import numpy as np
import zarr
from obstore.store import LocalStore

store_location = Path("test_zarr_store")
if store_location.exists():
    # allow to run the script multiple times by deleting the previous run
    shutil.rmtree(store_location)
store_location.mkdir()

zarr_store = ObjectStore(LocalStore(store_location))
arr = zarr.create_array(
    zarr_store, name="arr", shape=(5, 128, 128), dtype=np.uint16, chunks=(1, 32, 32)
)
# this will fail with: # FileNotFoundError: Object at location test_zarr_store/arr/c/0/1/0 not found
arr[0, :, :] = np.zeros((128, 128), dtype=np.uint16)

# however, if we don't use zeros but random values it works
# arr[0, :, :] = np.random.randint(0, 100, size=(128, 128), dtype=np.uint16)
```

### Additional output

_No response_

lukasbindreiter:
Suppress `FileNotFoundError` in `delete` operation of the obstore adapter.

Since some underlying stores allow this, while others raise an error, supressing the error results in consistent behaviour across all stores. This also mimics how the same thing is implemented in the `fsspec` adapter.

Fixes #3136 

TODO:

### DIFF
--- a/src/zarr/storage/_obstore.py
+++ b/src/zarr/storage/_obstore.py
@@ -181,7 +181,12 @@ class ObjectStore(Store):
         import obstore as obs
 
         self._check_writable()
-        await obs.delete_async(self.store, key)
+
+        # Some stores such as local filesystems, GCP and Azure raise an error
+        # when deleting a non-existent key, while others such as S3 and in-memory do
+        # not. We suppress the error to make the behavior consistent across all stores
+        with contextlib.suppress(FileNotFoundError):
+            await obs.delete_async(self.store, key)
 
     @property
     def supports_partial_writes(self) -> bool:
--- a/tests/test_store/test_object.py
+++ b/tests/test_store/test_object.py
@@ -75,6 +75,9 @@ class TestObjectStore(StoreTests[ObjectStore, cpu.Buffer]):
         with pytest.raises(TypeError):
             ObjectStore("path/to/store")
 
+    async def test_store_delete_nonexistent_key_does_not_raise(self, store: ObjectStore) -> None:
+        await store.delete("nonexistent_key")
+
 
 @pytest.mark.slow_hypothesis
 def test_zarr_hierarchy():


dstansby:
Thanks for the PR/fix! This is slightly surprising behaviour to me - I would expect trying to delete a non-existent file to raise an error. But I guess if this is existing behaviour/design across other stores, this is good.

To be sure that we're getting consistent behaviour across stores, could you move the test you added to the `StoreTests` class, so it's run across all stores that are tested?

d-v-b:
I think a better long-term solution would be for the stores to all raise the same exception when attempting to delete a missing key, and for the caller of the store method to handle that exception specifically in the context of attempting to remove an empty chunk. But this fix also works!

kylebarron:
> I think a better long-term solution would be for the stores to all raise the same exception when attempting to delete a missing key

From the obstore side I don't have total control over that. The upstream s3 implementation does not raise an error: https://docs.rs/object_store/latest/object_store/trait.ObjectStore.html#method.delete_stream

lukasbindreiter:
> Thanks for the PR/fix! This is slightly surprising behaviour to me - I would expect trying to delete a non-existent file to raise an error. But I guess if this is existing behaviour/design across other stores, this is good.

Yes this seems to be the behaviour across stores:

- `local store` uses `path.unlink(missing_ok=True)` 
    https://github.com/zarr-developers/zarr-python/blob/main/src/zarr/storage/_local.py#L218 
- `memory store` suppresses `KeyError`
    https://github.com/zarr-developers/zarr-python/blob/main/src/zarr/storage/_memory.py#L143
- `fsspec store` also suppresses `FileNotFoundError`:
    https://github.com/zarr-developers/zarr-python/blob/main/src/zarr/storage/_fsspec.py#L359

so I believe it makes sense to do the same for `obstore` as well.

And when thinking about the main purpose of the function (to clear chunks) this behaviour makes sense - it seems the function isn't really about deleting an object, instead it is to make sure that an object doesn't exist in the store.


> To be sure that we're getting consistent behaviour across stores, could you move the test you added to the StoreTests class, so it's run across all stores that are tested?

Good idea, I've moved the test now to `StoreTests` 👍 



Across stores is a bit ambiguous in this context though - there is the zarr adapter stores (`fsspec`, `local`, `memory`, `obstore`) - the test now runs across all of those.

However the issue / bugfix was only for the `obstore` adapter, since there is inconsistent behaviour across obstore stores (`S3`, `GCS`, `Azure`, `Local`, `InMemory`) - but as @kylebarron has mentioned that seems to be out of control of obstore itself.

src/zarr/storage/_obstore.py
```diff
@@ -186,9 +186,17 @@ def supports_deletes(self) -> bool:
     async def delete(self, key: str) -> None:
         # docstring inherited
         import obstore as obs
+        from obstore.exceptions import NotFoundError
```
kylebarron: IMO let's just handle `FileNotFoundError` for parity with existing code:
https://github.com/zarr-developers/zarr-python/blob/2911be8f20357c7f717eed410ccc99782d487808/src/zarr/storage/_obstore.py#L153-L154

I can't get multiple inheritance of exceptions to work in obstore, so it's possible I'll just remove the `NotFoundError` in a future release.

maxrjones:
```suggestion
```

src/zarr/storage/_obstore.py
```diff
@@ -186,9 +186,17 @@ def supports_deletes(self) -> bool:
...
 
         self._check_writable()
-        await obs.delete_async(self.store, key)
+
+        # Some stores such as local filesystems, GCP and Azure raise an error
+        # when deleting a non-existent key, while others such as S3 and in-memory do
+        # not. We suppress the error to make the behavior consistent across all stores
+        # currently obstore raises FileNotFoundError, but in the future might raise
+        # NotFoundError instead, so let's suppress that too
+        with contextlib.suppress(FileNotFoundError, NotFoundError):
```
maxrjones:
```suggestion
        with contextlib.suppress(FileNotFoundError):
```

maxrjones (approved):
My approval doesn't mean anything really, but I'm submitting it because I ran into this issue and tested your branch locally to confirm that it solved the problem. I also made two code suggestions to apply Kyle's recommendation for convenience.

src/zarr/storage/_obstore.py
kylebarron: We should also update the comment here

kylebarron:
In https://github.com/developmentseed/obstore/pull/487 I propose deprecating the `obstore.exceptions.NotFoundError` class

src/zarr/storage/_obstore.py
lukasbindreiter: Makes sense 👍

lukasbindreiter:
Thanks @maxrjones and @kylebarron for the review - I've now updated the comment and removed the `obstore.exceptions.NotFoundError` again

d-v-b (approved):

lukasbindreiter:
@d-v-b Thanks for enabling the merge.

I checked the failing test for the `windows-latest py3.11` run, which is this one. 

```python
def test_sync_timeout() -> None:
    duration = 0.02

    async def foo() -> None:
        await asyncio.sleep(duration)

    with pytest.raises(asyncio.TimeoutError):
        sync(foo(), timeout=duration / 10)
```

It seems completely unrelated to the changes in the PR, probably a flaky failure? Could you potentially retry it?

d-v-b:
@lukasbindreiter the flaky test theory seems valid! thanks for this contribution.

(I accidentally posted this in https://github.com/zarr-developers/zarr-python/issues/3136#issuecomment-3004339926)

d-v-b:

PR #3151, opened 2025-06-18
Fix/windows int32 issue

d-v-b:
There's a windows-specific issue where `np.dtype('i')` is an instance of `np.dtypes.IntDType` instead of `np.dtypes.Int32DType`, which breaks dtype inference. This PR adds logic to the dtype inference for `Int32` to handle this case, and adds a test that fails without that fix.

TODO:


d-v-b:
rtd build is failing but I can't replicate this locally...

jhamman:
docs should be fixed now. Thanks @d-v-b

Issue #3147, opened 2025-06-18
Able to delete group even when opened as readonly


TomNicholas:
### Zarr version

v3.0.8

### Numcodecs version

0.16.1

### Python Version

3.12.4

### Operating System

linux

### Installation

pip

### Description

I was not expecting to be able to delete a group when opening from the store using `mode='r'`.

### Steps to reproduce

```python
# /// script
# requires-python = ">=3.11"
# dependencies = [
#   "zarr@git+https://github.com/zarr-developers/zarr-python.git@main",
# ]
# ///

import zarr

store = zarr.storage.MemoryStore()
root = zarr.create_group(store=store)
foo = root.create_group('foo')
bar = root.create_group('bar')
root.members()
# (('foo', <Group memory://139839476802624/foo>),
# ('bar', <Group memory://139839476802624/bar>))

# re-open root group in a supposedly readonly state
root_readonly = zarr.open_group(store, path='/', mode='r')

# delete a child group
del root_readonly['foo']

root_readonly.members()
# (('bar', <Group memory://139839476802624/bar>),)
```

---

```
zarr.print_debug_info()
```
```
platform: Linux-5.10.220-209.869.amzn2.x86_64-x86_64-with-glibc2.35
python: 3.12.4
zarr: 3.0.8

**Required dependencies:**
packaging: 24.1
numpy: 1.26.4
numcodecs: 0.16.1
typing_extensions: 4.12.2
donfig: 0.8.1.post1

**Optional dependencies:**
botocore: 1.34.106
fsspec: 2024.6.0
numcodecs: 0.16.1
s3fs: 2024.6.0
gcsfs: 2024.6.0
rich: 13.7.1

**Not Installed:**
cupy-cuda12x
universal-pathlib
obstore
```

d-v-b:
at the moment `mode=r` for arrays and groups is a placebo. This is related to https://github.com/zarr-developers/zarr-python/pull/3068#issuecomment-2923535068

maxrjones:
Since https://github.com/zarr-developers/zarr-python/pull/3068 (currently unreleased), `zarr.open_group(store=writeable_store, read_only=True)` errors:
```
Store is not read-only but mode is 'r'
  File "/Users/max/Documents/Code/zarr-developers/zarr-python/src/zarr/storage/_common.py", line 90, in open
    raise ValueError(f"Store is not read-only but mode is '{mode}'")
[...]

PR #3155, opened 2025-06-19
Only warn if opening writeable store with mode='r'

dstansby:
See much discussion in https://github.com/zarr-developers/zarr-python/pull/3068 - instead of erroring, this just warns now. ping @maxrjones

PR #3156, opened 2025-06-19
Create read only copy if needed when opening a store path

maxrjones:
This PR makes a read-only copy of a store when using `StorePath.open(mode="r")`.

See discussion in #3068. This augments #3068 (it still errors when attempting to open a writeable store path using a read-only store. This is an alternative to #3155.

Closes https://github.com/zarr-developers/zarr-python/issues/3147

TODO:

### DIFF
--- a/src/zarr/storage/_common.py
+++ b/src/zarr/storage/_common.py
@@ -2,8 +2,9 @@ from __future__ import annotations
 
 import importlib.util
 import json
+import warnings
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, Literal, Self, TypeAlias
+from typing import TYPE_CHECKING, Any, Literal, Self, TypeAlias, get_args
 
 from zarr.abc.store import ByteRequest, Store
 from zarr.core.buffer import Buffer, default_buffer_prototype
@@ -55,58 +56,87 @@ class StorePath:
         return self.store.read_only
 
     @classmethod
+    async def _create_open_instance(cls, store: Store, path: str) -> Self:
+        """Helper to create and return a StorePath instance."""
+        await store._ensure_open()
+        return cls(store, path)
+
+    @classmethod
     async def open(cls, store: Store, path: str, mode: AccessModeLiteral | None = None) -> Self:
         """
         Open StorePath based on the provided mode.
 
-        * If the mode is 'w-' and the StorePath contains keys, raise a FileExistsError.
-        * If the mode is 'w', delete all keys nested within the StorePath
-        * If the mode is 'a', 'r', or 'r+', do nothing
+        * If the mode is None, return an opened version of the store with no changes.
+        * If the mode is 'r+', 'w-', 'w', or 'a' and the store is read-only, raise a ValueError.
+        * If the mode is 'r' and the store is not read-only, return a copy of the store with read_only set to True.
+        * If the mode is 'w-' and the store is not read-only and the StorePath contains keys, raise a FileExistsError.
+        * If the mode is 'w'  and the store is not read-only, delete all keys nested within the StorePath.
 
         Parameters
         ----------
         mode : AccessModeLiteral
             The mode to use when initializing the store path.
 
+            The accepted values are:
+
+            - ``'r'``: read only (must exist)
+            - ``'r+'``: read/write (must exist)
+            - ``'a'``: read/write (create if doesn't exist)
+            - ``'w'``: read/write (overwrite if exists)
+            - ``'w-'``: read/write (create if doesn't exist).
+
         Raises
         ------
         FileExistsError
             If the mode is 'w-' and the store path already exists.
-        ValueError
-            If the mode is not "r" and the store is read-only, or
-            if the mode is "r" and the store is not read-only.
         """
-
-        await store._ensure_open()
-        self = cls(store, path)
 
         # fastpath if mode is None
         if mode is None:
-            return self
+            return await cls._create_open_instance(store, path)
 
-        if store.read_only and mode != "r":
-            raise ValueError(f"Store is read-only but mode is '{mode}'")
-        if not store.read_only and mode == "r":
-            raise ValueError(f"Store is not read-only but mode is '{mode}'")
+        if mode not in get_args(AccessModeLiteral):
+            raise ValueError(f"Invalid mode: {mode}, expected one of {AccessModeLiteral}")
 
+        if store.read_only:
+            # Don't allow write operations on a read-only store
+            if mode != "r":
+                raise ValueError(
+                    f"Store is read-only but mode is '{mode}'. Create a writable store or use 'r' mode."
+                )
+            self = await cls._create_open_instance(store, path)
+        elif mode == "r":
+            # Create read-only copy for read mode on writable store
+            try:
+                warnings.warn(
+                    "Store is not read-only but mode is 'r'. Creating a read-only copy. "
+                    "This behavior may change in the future with a more granular permissions model.",
+                    UserWarning,
+                    stacklevel=2,
+                )
+                self = await cls._create_open_instance(store.with_read_only(True), path)
+            except NotImplementedError as e:
+                raise ValueError(
+                    "Store is not read-only but mode is 'r'. Unable to create a read-only copy of the store."
+                ) from e
+        else:
+            # writable store and writable mode
+            await store._ensure_open()
+            self = await cls._create_open_instance(store, path)
+
+        # Handle mode-specific operations
         match mode:
             case "w-":
                 if not await self.is_empty():
-                    msg = (
-                        f"{self} is not empty, but `mode` is set to 'w-'."
-                        "Either remove the existing objects in storage,"
-                        "or set `mode` to a value that handles pre-existing objects"
-                        "in storage, like `a` or `w`."
+                    raise FileExistsError(
+                        f"Cannot create '{path}' with mode 'w-' because it already contains data. "
+                        f"Use mode 'w' to overwrite or 'a' to append."
                     )
-                    raise FileExistsError(msg)
             case "w":
                 await self.delete_dir()
             case "a" | "r" | "r+":
                 # No init action
                 pass
-            case _:
-                raise ValueError(f"Invalid mode: {mode}")
-
         return self
 
     async def get(
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -1318,7 +1318,7 @@ def test_no_overwrite_open(tmp_path: Path, open_func: Callable, mode: str) -> No
     existing_fpath = add_empty_file(tmp_path)
 
     assert existing_fpath.exists()
-    with contextlib.suppress(FileExistsError, FileNotFoundError, ValueError):
+    with contextlib.suppress(FileExistsError, FileNotFoundError, UserWarning):
         open_func(store=store, mode=mode)
     if mode == "w":
         assert not existing_fpath.exists()
--- a/tests/test_store/test_core.py
+++ b/tests/test_store/test_core.py
@@ -263,8 +263,11 @@ def test_relativize_path_invalid() -> None:
         _relativize_path(path="a/b/c", prefix="b")
 
 
-def test_invalid_open_mode() -> None:
+def test_different_open_mode() -> None:
     store = MemoryStore()
     zarr.create((100,), store=store, zarr_format=2, path="a")
-    with pytest.raises(ValueError, match="Store is not read-only but mode is 'r'"):
+    with pytest.warns(
+        UserWarning,
+        match="Store is not read-only but mode is 'r'. Attempting to create a read-only copy. This behavior may change in the future with a more granular permissions model.",
+    ):
         zarr.open_array(store=store, path="a", zarr_format=2, mode="r")


src/zarr/storage/_common.py
```diff
@@ -54,59 +55,90 @@ def __init__(self, store: Store, path: str = "") -> None:
...
 
         Parameters
         ----------
         mode : AccessModeLiteral
             The mode to use when initializing the store path.
 
+            The accepted values are:
+
+            - ``'r'``: read only (must exist)
+            - ``'r+'``: read/write (must exist)
```
maxrjones, line 89: currently there are actually no checks that the store contains any keys, but I think this should be addressed as a separate PR.

src/zarr/storage/_common.py
```diff
@@ -54,59 +55,90 @@ def __init__(self, store: Store, path: str = "") -> None:
...
+                        f"Use mode 'w' to overwrite or 'a' to append."
                     )
-                    raise FileExistsError(msg)
             case "w":
                 await self.delete_dir()
             case "a" | "r" | "r+":
                 # No init action
                 pass
-            case _:
-                raise ValueError(f"Invalid mode: {mode}")
```
maxrjones, line 108: I moved the mode checking up to the start of the function.

src/zarr/storage/_common.py
maxrjones, line 89: After #3118 the `AccessModeLiteral` should be clickable in the docs for users to quickly get to the options, but until then I think we need to duplicate the meaning here. In particular, I think a lot of people may not know that "r+" means writeable because file modes are not as prevalent as they used to be.

src/zarr/storage/_common.py
```diff
@@ -54,59 +55,90 @@ def __init__(self, store: Store, path: str = "") -> None:
...
         # fastpath if mode is None
         if mode is None:
-            return self
+            return await cls._create_open_instance(store, path)
 
-        if store.read_only and mode != "r":
-            raise ValueError(f"Store is read-only but mode is '{mode}'")
-        if not store.read_only and mode == "r":
-            raise ValueError(f"Store is not read-only but mode is '{mode}'")
+        if mode not in get_args(AccessModeLiteral):
```
d-v-b: a static alternative to invoking `get_args` here is to define a tuple of strings (all the elements of the literal type), annotate it as `Final`, and refer to that variable here. That would be _marginally_ more performant. See an example [here](https://github.com/zarr-developers/zarr-python/blob/286bef86b040b76b458a2f4db2e3215fbd9457ac/src/zarr/core/dtype/__init__.py#L69-L70). Just floating this as an option.

src/zarr/storage/_common.py
```diff
@@ -54,59 +55,90 @@ def __init__(self, store: Store, path: str = "") -> None:
...
+            raise ValueError(f"Invalid mode: {mode}, expected one of {AccessModeLiteral}")
 
+        if store.read_only:
+            # Don't allow write operations on a read-only store
+            if mode != "r":
+                raise ValueError(
+                    f"Store is read-only but mode is '{mode}'. Create a writable store or use 'r' mode."
```
d-v-b:
```suggestion
                    f"Store is read-only but mode is {mode!r}. Create a writable store or use 'r' mode."
```
the !r will ensure that strings get displayed with quotes around them, but other python objects will not.

src/zarr/storage/_common.py
maxrjones: done in [`a42dc7c` (#3156)](https://github.com/zarr-developers/zarr-python/pull/3156/commits/a42dc7cd9e552fb82a2be8d90852057188c68b35)

src/zarr/core/common.py
```diff
@@ -39,6 +40,7 @@
 JSON = str | int | float | Mapping[str, "JSON"] | Sequence["JSON"] | None
 MemoryOrder = Literal["C", "F"]
 AccessModeLiteral = Literal["r", "r+", "a", "w", "w-"]
+ANY_ACCESS_MODE: Final = "r", "r+", "a", "w", "w-"
```
dstansby, line 43:
```suggestion
ANY_ACCESS_MODE: Final = get_args(AccessModeLiteral)
```
It might be possible to de-duplicate here (untested) using [`typing.get_args`](https://docs.python.org/3/library/typing.html#typing.get_args)?

src/zarr/storage/_common.py
```diff
@@ -54,59 +62,90 @@ def __init__(self, store: Store, path: str = "") -> None:
...
     @classmethod
     async def open(cls, store: Store, path: str, mode: AccessModeLiteral | None = None) -> Self:
         """
         Open StorePath based on the provided mode.
 
-        * If the mode is 'w-' and the StorePath contains keys, raise a FileExistsError.
-        * If the mode is 'w', delete all keys nested within the StorePath
-        * If the mode is 'a', 'r', or 'r+', do nothing
+        * If the mode is None, return an opened version of the store with no changes.
+        * If the mode is 'r+', 'w-', 'w', or 'a' and the store is read-only, raise a ValueError.
```
dstansby, line 76: Given we can go one way (make a store read-only), is there a reason not to go the other way (make a store writeable) in this case, instead of raising an error?

src/zarr/storage/_common.py
```diff
@@ -54,59 +62,90 @@ def __init__(self, store: Store, path: str = "") -> None:
...
+                )
+            self = await cls._create_open_instance(store, path)
+        elif mode == "r":
+            # Create read-only copy for read mode on writable store
+            try:
+                warnings.warn(
+                    "Store is not read-only but mode is 'r'. Creating a read-only copy. "
+                    "This behavior may change in the future with a more granular permissions model.",
+                    UserWarning,
+                    stacklevel=2,
```
dstansby:
```suggestion
                    stacklevel=1,
```
I think a stacklevel of 1 (raising from this function, instead of the one calling it) makes more sense here?

src/zarr/storage/_common.py
```diff
@@ -54,59 +62,90 @@ def __init__(self, store: Store, path: str = "") -> None:
...
+                    raise FileExistsError(
+                        f"Cannot create '{path}' with mode 'w-' because it already contains data. "
+                        f"Use mode 'w' to overwrite or 'a' to append."
                     )
-                    raise FileExistsError(msg)
             case "w":
                 await self.delete_dir()
             case "a" | "r" | "r+":
                 # No init action
                 pass
```
dstansby:
```suggestion
```
Given mode checking has moved up, these lines could go too.

dstansby (dismissed):
Looks good to me. Requesting changes because:
- This should get a release note
- My old release note from when I added the error should be deleted

src/zarr/core/common.py
d-v-b: at least on my machine, the type inference machinery thinks `get_args(...)` returns `tuple[Any,...]`. So this won't work.

dstansby: Does it matter what the type is of `ANY_ACCESS_MODE` though, since it's only used for runtime checking? Not a big deal given these two lines of code are next to each other though, so should be obvious if they end up out of sync.

d-v-b: it does matter, because when you annotate a tuple of strings as `Final`, type checkers will know that it's `Literal[*stuff_in_the_tuple]`, which means you can use that variable for "compile-time" and runtime type checks.

src/zarr/storage/_common.py
maxrjones: I viewed only going from writable -> readable as an optimization for safety because it seems more likely that someone could make a mistake. I haven't seen the pattern of creating a (theoretically) writeable array from a read-only store as much as the opposite case, so could we see if there's a request for that functionality rather than adding it to this PR?

maxrjones:
> This should get a release note
> My old release note from when I added the error should be deleted

Thanks, done in [`10e50e4` (#3156)](https://github.com/zarr-developers/zarr-python/pull/3156/commits/10e50e45524bf330cd4b3daad2b2bdb1f515c52e)

src/zarr/core/common.py
maxrjones: It worked on my machine and I was curious what the CI would say so I pushed [`7ad760f` (#3156)](https://github.com/zarr-developers/zarr-python/pull/3156/commits/7ad760fba89e77a3ec5ad9649ada3a99799feeb6) (would be easy to revert again)

maxrjones: The CI passes. @d-v-b are you getting a type error or runtime error?

src/zarr/storage/_common.py
dstansby: 👍 sounds good

src/zarr/core/common.py
d-v-b: CI will pass because the return type of `get_args()` is `tuple[Any,...]`, and so any type is assignable to its contents. 

```
lit_typ = Literal["a", "b"]
lit_val: Final = ("a", "b")

x: lit_typ = lit_val[0]
# the type of x is inferred to be Literal["a"]
y = get_args(lit_typ)[0]
# the type of y is inferred to be Any
```

d-v-b: another argument against `get_args` here:
```
y: lit_typ = get_args(lit_val)[15]
# mypy is OK with this
z: lit_typ = lit_val[15]
# mypy sees the indexing error
```

maxrjones: thanks for the explanation, I reverted the change in [`a1949f1` (#3156)](https://github.com/zarr-developers/zarr-python/pull/3156/commits/a1949f1044ae8867e7ec1af8631d459bdce4df81).

It seems like our options are redundancy or using `get_args` inside `StorePath.open()`. @d-v-b previously suggested that the redundancy option offers minor performance benefits.

d-v-b: yeah it's a handful strings, so writing it down twice in essentially the exact same location is not expensive. If we are _really_ worried that the type and the final variable might diverge, we can test for that (using get_args).

maxrjones: > yeah it's a handful strings, so writing it down twice in essentially the exact same location is not expensive. If we are really worried that the type and the final variable might diverge, we can test for that (using get_args).

this seems like a reasonable compromise. I added a test in [`202d606` (#3156)](https://github.com/zarr-developers/zarr-python/pull/3156/commits/202d60689fe900537f40fcced8365edf41793805).

src/zarr/storage/_common.py
```diff
@@ -54,59 +62,87 @@ def __init__(self, store: Store, path: str = "") -> None:
...
+            self = await cls._create_open_instance(store, path)
+        elif mode == "r":
+            # Create read-only copy for read mode on writable store
+            try:
+                warnings.warn(
+                    "Store is not read-only but mode is 'r'. Creating a read-only copy. "
+                    "This behavior may change in the future with a more granular permissions model.",
+                    UserWarning,
+                    stacklevel=1,
+                )
```
TomNicholas: I feel like this warning could be outside of the `try... except` block. The start of the message is redundant with that in the re-raised `ValueError`, so why not just move this warning to before the `try...except`, effectively factoring out that part of the message?

src/zarr/storage/_common.py
```diff
@@ -54,59 +62,87 @@ def __init__(self, store: Store, path: str = "") -> None:
...
+                self = await cls._create_open_instance(store.with_read_only(True), path)
+            except NotImplementedError as e:
+                raise ValueError(
+                    "Store is not read-only but mode is 'r'. Unable to create a read-only copy of the store."
```
TomNicholas: _User when this error is thrown:_ "But why was it unable to create a read-only copy?"

The fact it's because the store doesn't implement something specific is totally swallowed by this error message, with no further direction given. I think the re-raised error message would be more helpful if it ended with `because store class {type(store)} does not implement method ...`.

src/zarr/storage/_common.py
```diff
@@ -54,59 +62,87 @@ def __init__(self, store: Store, path: str = "") -> None:
...
+        # Handle mode-specific operations
         match mode:
             case "w-":
                 if not await self.is_empty():
-                    msg = (
-                        f"{self} is not empty, but `mode` is set to 'w-'."
-                        "Either remove the existing objects in storage,"
-                        "or set `mode` to a value that handles pre-existing objects"
-                        "in storage, like `a` or `w`."
+                    raise FileExistsError(
```
TomNicholas, line 134: I personally think all these `FileExistsErrors` should be `KeyErrors` but I guess that complaint is orthogonal to this PR.

src/zarr/storage/_common.py
TomNicholas: Motivated by the idea that it's good practice to have the `try...except` wrap the absolute minimum amount of code that could raise to minimize the chance of it raising on something unexpected.

TomNicholas: In fact going even further I guess the `try... except` only needs to wrap 

```python
read_only_store = store.with_read_only(True)
```
not the outer part
```python
self = await cls._create_open_instance(read_only_store, path)
```

TomNicholas:
FYI I tried running xarray's zarr tests with this PR and while it fixes the basic read_only failure that #3068 caused (see for example [this test](https://github.com/pydata/xarray/actions/runs/15768379641/job/44448894454#step:8:65650), which passes with this PR), I still see a lot of other errors. I'm just trying to work out if those errors are caused by this PR or by zarr's dtypes refactor.

I also see a lot of things like this

```
FAILED xarray/tests/test_backends.py::TestZarrDictStore::test_warning_on_bad_chunks[2] - AssertionError: Got 1 unexpected warning(s): ["UserWarning('Store is not read-only but mode is 'r'. Creating a read-only copy. This behavior may change in the future with a more granular permissions model.'))"]
```
suggesting that xarray might need to swallow the warning being raised in this PR.

d-v-b:
Why do we need to raise a warning here? Doesn't the new `with_read_only` method fully legitimize the creation of a new store in read-only mode?

TomNicholas:
I agree, this isn't something to warn for, it's just a change of behaviour to be documented.

rabernat:
> I agree, this isn't something to warn for, it's just a change of behaviour to be documented.

:+1: to this. FWIW I think we have been way to eager to issue warnings in Zarr. No warning is needed here.

TomNicholas:
> I'm just trying to work out if those errors are caused by this PR or by zarr's dtypes refactor.

After talking to @ianhi I think those other errors are related to the dtypes changes rather than to the changes in this PR.

maxrjones:
> > I'm just trying to work out if those errors are caused by this PR or by zarr's dtypes refactor.
> 
> After talking to @ianhi I think those other errors are related to the dtypes changes rather than to the changes in this PR.

Yes they are definitely unrelated to this and instead are relevant to the dtypes changes.

src/zarr/storage/_common.py
maxrjones: I'll add direction to first create a read-only store or use a store than implements `with_read_only`. I don't agree with the value of duplication in the tracebace and think it's reasonable to expect users to read tracebacks e.g.,:
```bash
Exception has occurred: ValueError       (note: full exception trace is shown but execution is paused at: _run_module_as_main)
Store is not read-only but mode is 'r'. Unable to create a read-only copy of the store.
  File "/Users/max/Documents/Code/zarr-developers/zarr-python/src/zarr/storage/_common.py", line 126, in open
    # writable store and writable mode
  File "/Users/max/Documents/Code/zarr-developers/zarr-python/src/zarr/abc/store.py", line 103, in with_read_only
        f"with_read_only is not implemented for the {type(self)} store type."
    
    )
NotImplementedError: with_read_only is not implemented for the <class 'zarr.storage._zip.ZipStore'> store type.

The above exception was the direct cause of the following exception:

  File "/Users/max/Documents/Code/zarr-developers/zarr-python/src/zarr/storage/_common.py", line 128, in open
    
    
    # Handle mode-specific operations
  File "/Users/max/Documents/Code/zarr-developers/zarr-python/src/zarr/storage/_common.py", line 373, in make_store_path
    return result
    
  File "/Users/max/Documents/Code/zarr-developers/zarr-python/src/zarr/api/asynchronous.py", line 1262, in open_array
    store_path = await make_store_path(store, path=path, mode=mode, storage_options=storage_options)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/max/Documents/Code/zarr-developers/zarr-python/src/zarr/core/sync.py", line 119, in _runner
    return await coro
           ^^^^^^^^^^
  File "/Users/max/Documents/Code/zarr-developers/zarr-python/src/zarr/core/sync.py", line 163, in sync
    raise return_result
  File "/Users/max/Documents/Code/zarr-developers/zarr-python/src/zarr/api/synchronous.py", line 1297, in open_array
            async_api.open_array(
            ^^^^^^^^^^^^^^^^^^^^^
    ...<7 lines>...
                storage_options=storage_options,
    
  File "/Users/max/Documents/Code/zarr-developers/zarr-python/.vscode/test-zip-mode.py", line 5, in <module>
    arr = zarr.open_array(store=store, path="a", zarr_format=2, mode="r")
  File "/opt/homebrew/Cellar/python@3.13/3.13.5/Frameworks/Python.framework/Versions/3.13/lib/python3.13/runpy.py", line 88, in _run_code
    exec(code, run_globals)
  File "/opt/homebrew/Cellar/python@3.13/3.13.5/Frameworks/Python.framework/Versions/3.13/lib/python3.13/runpy.py", line 198, in _run_module_as_main (Current frame)
    return _run_code(code, main_globals, None,
ValueError: Store is not read-only but mode is 'r'. Unable to create a read-only copy of the store.
```

src/zarr/storage/_common.py
maxrjones: thanks for the suggestions, I removed the warning in [`4e099e4` (#3156)](https://github.com/zarr-developers/zarr-python/pull/3156/commits/4e099e4f8c49090b182cf1c60e25006446f26253) and shortened the try...except in [`2616b2f` (#3156)](https://github.com/zarr-developers/zarr-python/pull/3156/commits/2616b2f8d030e6adfede6202c88d5331b3992446)

src/zarr/storage/_common.py
maxrjones: yeah there's definitely room for improvements with the selection and standardization of errors, but also as you said this PR isn't focused on that problem

maxrjones:
> Why do we need to raise a warning here? Doesn't the new `with_read_only` method fully legitimize the creation of a new store in read-only mode?

thanks for this suggestion, Davis, and the affirmations, Tom and Ryan. The PR no longer includes a warning. My decision to initially include one was influenced by the prevalence of warnings elsewhere so it's helpful to have that general correction.

src/zarr/storage/_common.py
TomNicholas: You're right - I missed the fact that raising from the `NotImplementedError` wouldn't actually completely swallow it, so the user would get to see the `NotImplementedError` message and hence why it actually failed.

d-v-b:

d-v-b:

PR #3157, opened 2025-06-19
document dtype extension

d-v-b:
This PR adds a working example of custom dtype creation and registration. because it's a lot of code, I put this in a new top-level directory called `examples`, which contains the executable python file `dtype_example.py`. This file uses PEP-723 metadata to declare a `ml_dtypes` dependency, and it uses a local zarr install, which means it can be tested properly against local changes.

I also expanded the current dtype docs in the user guide to include content about the data type resolution process.

TODO:

### DIFF
--- a/src/zarr/core/dtype/npy/bool.py
+++ b/src/zarr/core/dtype/npy/bool.py
@@ -246,7 +246,7 @@ class Bool(ZDType[np.dtypes.BoolDType, np.bool_], HasItemSize):
         """
         if self._check_scalar(data):
             return np.bool_(data)
-        msg = (
+        msg = (  # pragma: no cover
             f"Cannot convert object {data!r} with type {type(data)} to a scalar compatible with the "
             f"data type {self}."
         )
--- a/src/zarr/core/dtype/npy/string.py
+++ b/src/zarr/core/dtype/npy/string.py
@@ -385,7 +385,7 @@ class FixedLengthUTF32(
 
             return self.to_native_dtype().type(str(data)[: self.length])
 
-        msg = (
+        msg = (  # pragma: no cover
             f"Cannot convert object {data!r} with type {type(data)} to a scalar compatible with the "
             f"data type {self}."
         )
@@ -711,11 +711,11 @@ class UTF8Base(ZDType[TDType_co, str], HasObjectCodec):
         """
         if self._check_scalar(data):
             return self._cast_scalar_unchecked(data)
-        msg = (
+        msg = (  # pragma: no cover
             f"Cannot convert object {data!r} with type {type(data)} to a scalar compatible with the "
             f"data type {self}."
         )
-        raise TypeError(msg)
+        raise TypeError(msg)  # pragma: no cover
 
 
 if _NUMPY_SUPPORTS_VLEN_STRING:
--- a/tests/test_dtype/test_npy/test_bytes.py
+++ b/tests/test_dtype/test_npy/test_bytes.py
@@ -24,6 +24,7 @@ class TestNullTerminatedBytes(BaseTestZDType):
         "|S",
         "|U10",
         "|f8",
+        {"name": "|S4", "object_codec_id": "vlen-bytes"},
     )
     invalid_json_v3 = (
         {"name": "fixed_length_ascii", "configuration": {"length_bits": 0}},


d-v-b:
cc @nenb @ianhi, since yall were the most involved in this example over in the main dtypes PR.

examples/custom_dtype.py
```diff
@@ -0,0 +1,170 @@
+# /// script
+# requires-python = ">=3.11"
+# dependencies = [
+#   "zarr @ {root}",
```
ianhi: Is there any way around doing this? When this is being tested is it being called with `run` or just run via pytest? If the latter then would it be possible to change this to to just `zarr`? I think that would a small but nice improvement as this is first and foremore for documentation, not testing.

examples/custom_dtype.py
```diff
@@ -0,0 +1,170 @@
...
+    check_dtype_spec_v2,
+)
+
+int2_dtype_cls = type(np.dtype("int2"))
+int2_scalar_cls = ml_dtypes.int2
+
+
+class Int2(ZDType[int2_dtype_cls, int2_scalar_cls]):
+    """
+    This class provides a Zarr compatibility layer around the int2 data type and the int2
```
ianhi: Is there a nice link explaining the difference between these? I think I've inferred it but would be nice to make it explicit.

examples/custom_dtype.py
```diff
@@ -0,0 +1,170 @@
...
+        msg = f"Cannot convert object with type {type(data)} to a 2-bit integer."
+        raise TypeError(msg)
+
+    def default_scalar(self) -> ml_dtypes.int2:
+        """Get the default scalar value"""
+        return ml_dtypes.int2(0)
+
+    def to_json_scalar(self, data: object, *, zarr_format: ZarrFormat) -> int:
+        """Convert a python object to a scalar."""
+        return int(data)
```
ianhi: Can this be more specific to the example? e.g. explain something to the effect of "needs to be `int` to be compatible with json." and mention `int2` somewhere.

examples/custom_dtype.py
```diff
@@ -0,0 +1,170 @@
...
+        """Serialize this ZDType to v2- or v3-flavored JSON"""
+        if zarr_format == 2:
+            return {"name": "int2", "object_codec_id": None}
+        elif zarr_format == 3:
+            return self._zarr_v3_name
+        raise ValueError(f"zarr_format must be 2 or 3, got {zarr_format}")  # pragma: no cover
+
+    def _check_scalar(self, data: object) -> TypeGuard[int]:
+        """Check if a python object is a valid scalar"""
+        return isinstance(data, (int, int2_scalar_cls))
```
ianhi: why does this check against `int` instead of just `int2`?

docs/user-guide/data_types.rst
```diff
@@ -170,3 +170,100 @@ Deserialize a scalar value from JSON:
...
+`NumPy structured data type <https://numpy.org/doc/stable/user/basics.rec.html#structured-datatypes>`_.
+This data type is a container that can contain any NumPy data type, which makes it recursive. It is
+not possible to make a lookup table that relates all NumPy structured data types to their Zarr
+equivalents, as there is a nearly unbounded number of different structured data types. So instead of
+a static lookup table, Zarr Python relies on a dynamic approach to data type resolution.
+
+Zarr Python defines a collection of Zarr data types. This collection, called a "data type registry",
+is essentially a dict where the keys are strings (a canonical name for each data type), and the values are
+the data type classes themselves. Dynamic data type resolution entails iterating over these data
+type classes, invoking a special class constructor defined on each one, and returning a concrete
```
ianhi: Maybe mention the name of method, or link to on ZDType.

examples/custom_dtype.py
d-v-b: It's essential that these examples be tested as part of the regular test suite. Otherwise, the examples will break, and users will complain, and we will have to fix them, and to check that our fix worked, we will need to write tests...

But I agree that the "zarr @ {root}" declaration sucks. So here is an idea -- our test suite doesn't test this file. Instead, at test time we generate a copy of this file with a transformed `/// script` header, where the zarr version has been replaced with a reference to the local development version of zarr.

examples/custom_dtype.py
d-v-b: no I don't actually think there is a nice link that explains the data type / scalar type difference. The numpy docs should explain this, but they don't. I can add something to our docs.

examples/custom_dtype.py
d-v-b: because `_check_scalar(0)` should be OK

docs/user-guide/data_types.rst
```diff
@@ -1,33 +1,54 @@
...
-encoding and decoding both the data type itself, and scalars from that data type to and from Zarr array metadata. And these serialization procedures
-depend on the Zarr format.
 
 Data types in Zarr version 2
 -----------------------------
 
 Version 2 of the Zarr format defined its data types relative to
 `NumPy's data types <https://numpy.org/doc/2.1/reference/arrays.dtypes.html#data-type-objects-dtype>`_,
-and added a few non-NumPy data types as well. Thus the JSON identifier for a NumPy-compatible data
+and added a few non-NumPy data types as well. With one exception, the Zarr V2 JSON identifier for a data
```
ianhi:
```suggestion
and added a few non-NumPy data types as well. With one exception (`<the exception>`), the Zarr V2 JSON identifier for a data
```

otherwise confusing with the mention of two special cases below

examples/custom_dtype.py
d-v-b: this is fixed using the rewrite-the-script-at-test-time strategy

docs/user-guide/data_types.rst
d-v-b: done

docs/user-guide/data_types.rst
d-v-b: done

d-v-b:
I think the content is basically done -- I added docstrings to basically every new class and method (please let me know if i missed anything). I am however currently stymied on some [sphinx build issues](https://app.readthedocs.org/projects/zarr/builds/28625135/), so if anyone is proficient in that kind of debugging I would appreciate help.

d-v-b:
one question for sphinx experts: how can I get sphinx to not mangle the pipe symbol (`|`), e.g.:

image

ianhi:
Build issue is somehow related to this i think: 

https://github.com/sphinx-doc/sphinx/issues/12026

https://github.com/readthedocs/sphinx-autoapi/issues/422

ianhi:
Broken in a new way! Crazy that that fixed that

d-v-b:
all the sphinx issues are sorted, the only failing check is coverage, and I'm not sure I believe the claims that these changes reduced test coverage by 20%.

examples/custom_dtype.py
d-v-b: I added something to this effect

examples/custom_dtype.py
d-v-b: I put something about this in the data type guide

examples/custom_dtype.py
```diff
@@ -0,0 +1,245 @@
...
+# dependencies = [
+#   "zarr @ git+https://github.com/zarr-developers/zarr-python.git@main",
+#   "ml_dtypes==0.5.1",
+#   "pytest==8.4.1"
+# ]
+# ///
+#
+# Note: the zarr version must be changed in order to run this outside of the
+# zarr source tree. For example, to make this script truly stand-alone, specify the zarr
+# dependency as just "zarr"
```
d-v-b:
```suggestion
```

docs/user-guide/data_types.rst
```diff
@@ -1,34 +1,54 @@
-Data types
-==========
+Array data types
```
jhamman, line 3: These docs are great. Since the TOC order is defined elsewhere and not modified in this PR, I wanted to suggest we relocate this page to the `Advanced Topics` section in the User Guide. Most users will not need to know 98% of this in order to use Zarr (which is good).

jhamman (approved):

maxrjones:
@d-v-b FYI I think that some of these missing features in 3.0 in the array docs can now instead be linked to the new dtypes docs - https://zarr.readthedocs.io/en/stable/user-guide/arrays.html#missing-features-in-3-0.

docs/user-guide/data_types.rst
d-v-b: done

d-v-b:
> @d-v-b FYI I think that some of these missing features in 3.0 in the array docs can now instead be linked to the new dtypes docs - https://zarr.readthedocs.io/en/stable/user-guide/arrays.html#missing-features-in-3-0.

I removed the "this stuff is missing" sections, leaving only the part that is still missing (the copying + migrating data stuff)

Issue #3168, opened 2025-06-25
disallow 0-length fixed-size data types


d-v-b:
NumPy lets you define 0-length data types, but they are useless in the context of arrays:

```python
# /// script
# requires-python = ">=3.11"
# dependencies = [
#   "numpy",
# ]
# ///

import numpy as np

dt = np.dtype('str')
print(dt.itemsize)
# 0
arr = np.array([], dtype=dt)
print(arr.dtype)
# dtype('<U1')
print(arr.dtype == dt)
# False
```

NumPy's data type API happily treats the user's requested data type as a suggestion rather than an instruction. Oh well.

In Zarr-python 2, these 0-length data types exposed some weird behavior, where the data type of an array did not match the data type of the numpy arrays returned when indexing that array:

```python
# /// script
# requires-python = ">=3.11"
# dependencies = [
#   "zarr==2.18",
#   "numcodecs<0.16"
# ]
# ///

import zarr

array = zarr.create(store={}, shape=(1,), dtype='U0')
array[:] = ''
print(array.dtype)
# <U0
print(array[:].dtype)
# <U1
print(array.dtype == array[:].dtype)
# False
```

A few options for zarr python, ranked by goodness, descending:
- copy NumPy's bad behavior. If a user requests a 0-length data type, silently replace it with a 1-length version
- allow creating 0-length data types, but raise an exception when creating an array with one.
- disallow the creation of 0-length data types at the data type level.

I think option 3 makes the most sense, so I'm going to open a PR with that. But we can talk about alternatives here as well. Maybe there is a big market for these 0-length data types that I am not aware of.

addresses part of #3167

d-v-b:
this was closed via #3170

Issue #3167, opened 2025-06-24
Broken behavior (unable to load written data) with `str` dtype on main


ianhi:
### Zarr version

main

### Numcodecs version

0.16.1

### Python Version

3.11

### Operating System

Mac

### Installation

script metadata

### Description

On main you can write an array with the `str` dtype. However, you cannot read it back. This is different than the behavior for 3.0.8 and seemingly an uncaught consequence of the dtype changes attn: @d-v-b 

### Steps to reproduce

```python
# /// script
# requires-python = ">=3.11"
# dependencies = [
#   "zarr@git+https://github.com/zarr-developers/zarr-python.git@main",
# ]
# ///

import zarr

# each option here breaks in the same way
# string_data = ["abc", "ab", "a"]
# string_data = ["ab", "cd", "ef"]
string_data = ["ab"]

group = zarr.group("store.zarr", overwrite=True)
array = group.create("blah", shape=len(string_data), dtype=str)
array[:] = string_data

open_group = zarr.open("store.zarr")
open_group["blah"][:]
```

Gives:

<details>
<summary><b>traceback</b></summary>

```python
Traceback (most recent call last):
  File "/Users/ian/Documents/dev/xarray/tmp/string_test.py", line 19, in <module>
    open_group["blah"][:]
    ~~~~~~~~~~~~~~~~~~^^^
  File "/Users/ian/.cache/uv/environments-v2/string-test-a602d4454d39a8df/lib/python3.12/site-packages/zarr/core/array.py", line 2454, in __getitem__
    return self.get_orthogonal_selection(pure_selection, fields=fields)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ian/.cache/uv/environments-v2/string-test-a602d4454d39a8df/lib/python3.12/site-packages/zarr/_compat.py", line 43, in inner_f
    return f(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^
  File "/Users/ian/.cache/uv/environments-v2/string-test-a602d4454d39a8df/lib/python3.12/site-packages/zarr/core/array.py", line 2896, in get_orthogonal_selection
    return sync(
           ^^^^^
  File "/Users/ian/.cache/uv/environments-v2/string-test-a602d4454d39a8df/lib/python3.12/site-packages/zarr/core/sync.py", line 163, in sync
    raise return_result
[...]

PR #3170, opened 2025-06-25
string dtype fixes

d-v-b:
this PR fixes the issues reported in #3167.

There were two problems driving that issue:
- zarr python was not using the string `"str"` to denote a variable length string data type. Instead, in main we use NumPy's behavior and map `"str"` to a fixed-length UTF-32 string, which is not a useful data type for most people. 

  In this PR, when a user requests `dtype=str` or `dtype="str"` or `dtype="string"` they get a `VariableLengthUTF8` dtype instead of the nearly-useless `U` dtype.

- zarr python could create data types for length-0 scalars. Since these data types cannot be associated with arrays that contain any values, they are useless. See #3168. This is a quirk of NumPy that we don't need to support. This PR contains logic to ensure that we don't create data types for length-0 scalars.

note that while debugging this, I found that zarr-python 2.x could create arrays with where the dtype in metadata indicates length 0, but the actual array uses a dtype with size 1. We will need some support for this, but I don't think we need that in this PR.

### DIFF
--- a/tests/test_dtype/test_npy/test_string.py
+++ b/tests/test_dtype/test_npy/test_string.py
@@ -133,3 +133,12 @@ def test_unstable_dtype_warning(zdtype: FixedLengthUTF32 | VariableLengthUTF8) -
     """
     with pytest.raises(UnstableSpecificationWarning):
         zdtype.to_json(zarr_format=3)
+
+def test_invalid_size() -> None:
+    """
+    Test that it's impossible to create a data type that has no length
+    """
+    length = 0
+    msg = f"length must be >= 1, got {length}."
+    with pytest.raises(ValueError, match=msg):
+        FixedLengthUTF32(length=length)
--- a/tests/test_dtype/test_npy/test_string.py
+++ b/tests/test_dtype/test_npy/test_string.py
@@ -134,6 +134,7 @@ def test_unstable_dtype_warning(zdtype: FixedLengthUTF32 | VariableLengthUTF8) -
     with pytest.raises(UnstableSpecificationWarning):
         zdtype.to_json(zarr_format=3)
 
+
 def test_invalid_size() -> None:
     """
     Test that it's impossible to create a data type that has no length
--- a/tests/test_dtype/test_npy/test_string.py
+++ b/tests/test_dtype/test_npy/test_string.py
@@ -120,7 +120,12 @@ class TestFixedLengthUTF32(BaseTestZDType):
     )
 
 
-@pytest.mark.parametrize("zdtype", [FixedLengthUTF32(length=10), VariableLengthUTF8()])
+@pytest.mark.parametrize(
+    "zdtype",
+    [
+        FixedLengthUTF32(length=10),
+    ],
+)
 def test_unstable_dtype_warning(zdtype: FixedLengthUTF32 | VariableLengthUTF8) -> None:
     """
     Test that we get a warning when serializing a dtype without a zarr v3 spec to json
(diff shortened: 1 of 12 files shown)


d-v-b:
another important change: I altered the zarr v3 name of the `VariableLengthUTF8` dtype to be `"string"` instead of `"variable_length_utf8"`. This should resolve compatibility issues with zarr-python 3.0.8.

tests/test_regression/scripts/v3.0.8.py
```diff
@@ -0,0 +1,75 @@
+# /// script
+# requires-python = ">=3.11"
+# dependencies = [
+#  "zarr==3.0.8"
+# ]
+# ///
+
+# /// script
+# requires-python = ">=3.11"
```
ianhi: Should this pin a python version? Conceivably this is still useful to test in a few years and we would want zarr 2.18 to run with a python version that it was actually like to have been used with (i.e. if someone wrote data today but reads it in 5 years with whatever zarr version)

d-v-b: yeah I think it should. i'll pin to 3.11

d-v-b: ended up pinning to 3.12

tests/test_dtype/test_npy/test_string.py
```diff
@@ -101,30 +101,45 @@ class TestFixedLengthUTF32(BaseTestZDType):
         {"name": "numpy.fixed_length_utf32", "configuration": {"length_bits": "invalid"}},
     )
 
-    scalar_v2_params = ((FixedLengthUTF32(length=0), ""), (FixedLengthUTF32(length=2), "hi"))
+    scalar_v2_params = ((FixedLengthUTF32(length=1), ""), (FixedLengthUTF32(length=2), "hi"))
     scalar_v3_params = (
-        (FixedLengthUTF32(length=0), ""),
+        (FixedLengthUTF32(length=1), ""),
```
ianhi, line 106: This line leaves me a little confused on what the intended behavior for FIxedLengthString with an empty fill value is.  or in general does the fill vlaue of a fixed length string need to be the same length?

d-v-b: I'm going with numpy's semantics, which assigns the empty string to a sequence of null bytes:
```
>>> np.array([''], dtype="U1").tobytes()
b'\x00\x00\x00\x00'
```

d-v-b: this is a situation where the array data type and the scalar data type (the thing you get when you index that array) are not the same:

```python
>>> np.dtype('U0').type('a').dtype
dtype('<U1')
>>> np.dtype('U0').type('').dtype 
dtype('<U')
>>> np.array([''], dtype="U0").dtype   
dtype('<U1')
>>> np.array([''], dtype="U0")[0].dtype
dtype('<U')
```

d-v-b:
for posterity, numpy _does_ permit the creation of arrays with a `"V0"` data type, i.e. raw bytes where each element has length 0:

```python
>>> np.array([b''], dtype="V0").dtype
dtype('V')
>>> np.array([b''], dtype="V0").tobytes()
b''
>>> np.array([b'',b'',b'',b''], dtype="V0").tobytes()
b''
```
Arrays with this data type are all the empty byte string b"". I don't think this is meaningful for zarr.

The changes in this PR will make this data type unrepresentable in zarr python. I hope we are OK with this.

PR #3190, opened 2025-07-01
Add human readable size for No. bytes stored to `info_complete`

jsignell:
Example:

```python
>>> z.info_complete()
Type               : Array
Zarr format        : 3
Data type          : Int32(endianness='little')
Fill value         : 0
Shape              : (10000, 10000)
Chunk shape        : (1000, 1000)
Order              : C
Read-only          : False
Store type         : LocalStore
Filters            : ()
Serializer         : BytesCodec(endian=<Endian.little: 'little'>)
Compressors        : (BloscCodec(typesize=4, cname=<BloscCname.zstd: 'zstd'>, clevel=3, shuffle=<BloscShuffle.bitshuffle: 'bitshuffle'>, blocksize=0),)
No. bytes          : 400000000 (381.5M)
No. bytes stored   : 3558573 (3.4M)
Storage ratio      : 112.4
Chunks Initialized : 100
```

Note: I think this was the intention, looks like just a dict key misalignment.

TODO:

### DIFF
--- a/docs/user-guide/arrays.rst
+++ b/docs/user-guide/arrays.rst
@@ -211,7 +211,7 @@ prints additional diagnostics, e.g.::
    Serializer         : BytesCodec(endian=<Endian.little: 'little'>)
    Compressors        : (BloscCodec(typesize=4, cname=<BloscCname.zstd: 'zstd'>, clevel=3, shuffle=<BloscShuffle.bitshuffle: 'bitshuffle'>, blocksize=0),)
    No. bytes          : 400000000 (381.5M)
-   No. bytes stored   : 3558573
+   No. bytes stored   : 3558573 (3.4M)
    Storage ratio      : 112.4
    Chunks Initialized : 100
 
@@ -285,7 +285,7 @@ Here is an example using a delta filter with the Blosc compressor::
    >>> compressors = zarr.codecs.BloscCodec(cname='zstd', clevel=1, shuffle=zarr.codecs.BloscShuffle.shuffle)
    >>> data = np.arange(100000000, dtype='int32').reshape(10000, 10000)
    >>> z = zarr.create_array(store='data/example-9.zarr', shape=data.shape, dtype=data.dtype, chunks=(1000, 1000), filters=filters, compressors=compressors)
-   >>> z.info
+   >>> z.info_complete()
    Type               : Array
    Zarr format        : 3
    Data type          : Int32(endianness='little')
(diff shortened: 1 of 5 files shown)


src/zarr/core/_info.py
```diff
@@ -133,7 +133,7 @@ def __repr__(self) -> str:
 
         if self._count_bytes_stored is not None:
             template += "\nNo. bytes stored   : {_count_bytes_stored}"
-            kwargs["_count_stored"] = byte_info(self._count_bytes_stored)
+            kwargs["_count_bytes_stored"] = byte_info(self._count_bytes_stored)
```
d-v-b, line 136: we got bitten by `kwargs` here :(

docs/user-guide/arrays.rst
```diff
@@ -285,7 +285,7 @@ Here is an example using a delta filter with the Blosc compressor::
    >>> compressors = zarr.codecs.BloscCodec(cname='zstd', clevel=1, shuffle=zarr.codecs.BloscShuffle.shuffle)
    >>> data = np.arange(100000000, dtype='int32').reshape(10000, 10000)
    >>> z = zarr.create_array(store='data/example-9.zarr', shape=data.shape, dtype=data.dtype, chunks=(1000, 1000), filters=filters, compressors=compressors)
-   >>> z.info
+   >>> z.info_complete()
```
jsignell, line 289: This was the only real change to docs. Seems like if you are explaining about compression it's useful to show this extra information.

tests/test_info.py
```diff
@@ -79,7 +79,7 @@ def test_array_info(zarr_format: ZarrFormat) -> None:
 
 
 @pytest.mark.parametrize("zarr_format", ZARR_FORMATS)
-@pytest.mark.parametrize("bytes_things", [(1_000_000, "976.6K", 500_000, "500000", "2.0", 5)])
+@pytest.mark.parametrize("bytes_things", [(1_000_000, "976.6K", 500_000, "488.3K", "2.0", 5)])
```
d-v-b, line 82: i love "bytes_things"

d-v-b (approved):
thanks! this is a nice improvement

src/zarr/core/_info.py
d-v-b: for posterity, if we really want to use `kwargs` (and I think we should not), we should define some typeddicts to model what goes in `kwargs`. something for a later pr.

dstansby:
Thanks a lot, this is a nice improvement! Could you add a file to the `/changes` directory with a changelog entry? Otherwise I think this is good to go.

d-v-b:
thanks @jsignell!

Issue #2904, opened 2025-03-10
Memory usage


TomAugspurger:
We discussed memory usage on Friday's community call. https://github.com/TomAugspurger/zarr-python-memory-benchmark started to look at some stuff.

https://rawcdn.githack.com/TomAugspurger/zarr-python-memory-benchmark/refs/heads/main/reports/memray-flamegraph-read-compressed.html has the memray flamegraph for reading an uncompressed array (400 MB total, split into 10 chunks of 40 MB each). I think the optimal memory usage here is about 400 MB. Our peak memory is about 2x that.

https://rawcdn.githack.com/TomAugspurger/zarr-python-memory-benchmark/refs/heads/main/reports/memray-flamegraph-read-compressed.html has the zstd compressed version. Peak memory is about 1.1 GiB.

I haven't looked too closely at the code, but I wonder if we could be smarter about a few things in certain cases:

1. For the uncompressed case, we *might* be able to do a `readinto` directly into (an appropriate slice of)the `out` array.
2. For the compressed case, we might be able to improve things once we know the size of the output buffers. I see that numcodec's `zstd.decode` takes an output buffer [here](https://github.com/zarr-developers/numcodecs/blob/3c933cf19d4d84f2efc5f3a36926d8c569514a90/numcodecs/zstd.pyx#L251C11-L252C1) that we could maybe use. And past that point, maybe all the codecs could reuse one or two buffers, rather than allocating a new buffer for each stage of the codec (one buffer if doing stuff inplace, two buffers if something can't be done inplace)?

I'm not too familiar with the codec pipeline stuff, but will look into this as I have time. Others should feel free to take this if someone gets an itch though. There's some work to be done :)

TomAugspurger:
https://github.com/TomAugspurger/zarr-python-memory-benchmark/blob/4039ba687452d65eef081bce1d4714165546422a/sol.py#L41 has a POC for using `readinto` to read an uncompressed zarr dataset into a pre-allocated buffer. https://rawcdn.githack.com/TomAugspurger/zarr-python-memory-benchmark/3567246b852d7adacbc10f32a58b0b3f6ac3d50b/reports/memray-flamegraph-sol-read-uncompressed.html shows that that takes ~exactly the size of the output ndarray (so no overhead from Zarr).


[...]

PR #3191, opened 2025-07-01
Add NDBuffer.empty

TomAugspurger:
In https://github.com/zarr-developers/zarr-python/issues/2904, I've found that `np.empty(...)` can be quite a bit faster than the `np.full` we use in places
(https://github.com/zarr-developers/zarr-python/blob/baabf08d07e8518e3d37bd83c493a1d46ea7ac1d/src/zarr/core/buffer/cpu.py#L149. Though note that `np.empty` and `np.zeros` are about the same for the cases where we use that.)

For the common case of chunk-aligned reads, the memset used by `np.full` or `np.zeros` should be unnecessary, because the codec pipeline will overwrite the memory anyway (or overwritten with `fill_value` if the store is missing the key).

This adds a new method `NDBuffer.empty`, mirroring `np.empty`. To preserve backwards compatibility, it's *not* abstract. It delegates to the less efficient `NDBuffer.create`. But I've implemented it for our `gpu` and `cpu` buffers.

### DIFF
--- a/src/zarr/testing/buffer.py
+++ b/src/zarr/testing/buffer.py
@@ -13,6 +13,8 @@ if TYPE_CHECKING:
     from collections.abc import Iterable
     from typing import Self
 
+    from zarr.core.common import ChunkCoords
+
 
 __all__ = [
     "NDBufferUsingTestNDArrayLike",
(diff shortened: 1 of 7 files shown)


src/zarr/testing/buffer.py
```diff
@@ -51,6 +53,15 @@ def create(
...
         return ret
 
+    @classmethod
+    def empty(
+        cls,
+        shape: ChunkCoords,
+        dtype: npt.DTypeLike,
+        order: Literal["C", "F"] = "C",
+    ) -> Self:
+        return super(cpu.NDBuffer, cls).empty(shape=shape, dtype=dtype, order=order)
```
TomAugspurger, line 63: This is to test the base `core.NDBuffer` implementation. This class inherits from `cpu.NDBuffer` so we use `super(cpu.NDBuffer, cls)` to get its parent's impelentation (the base class).

d-v-b (approved):
This looks good to me! Thanks Tom.

TomAugspurger:
Just a note: this *might* cause issues for people running the `StoreTests`. We needed https://github.com/rapidsai/kvikio/commit/bdca513912dd6aa3053d6450c660ffab679250aa in kvikio to avoid an error from pytest about an unregistered marker.

PR #3195, opened 2025-07-02
Add missing import for AsyncFileSystemWrapper for `_make_async` in `_fsspec.py`

tasansal:
When using FSSpec's memory filesystem, the new `_make_async` function falls back to using below:

https://github.com/zarr-developers/zarr-python/blob/baabf08d07e8518e3d37bd83c493a1d46ea7ac1d/src/zarr/storage/_fsspec.py#L71

However, this is not imported and results in an `AttributeError`. This is expected since the submodule is not imported and we are trying to call it from `fsspec` module itself.

Added explicit import of the submodule.

Reproducer:

```python
import zarr
zarr.open_group("memory://group.zarr")
```

Python 3.13
`zarr-python` version `3.0.9`
`fsspec` version `2025.5.1`

I'm not sure if any unit tests are necessary here since its a simple fix of a minor oversight. May I suggest doing a `3.0.9` post-release and yanking the `3.0.9` ASAP. 

TODO:

### DIFF
--- a/src/zarr/storage/_fsspec.py
+++ b/src/zarr/storage/_fsspec.py
@@ -68,9 +68,9 @@ def _make_async(fs: AbstractFileSystem) -> AsyncFileSystem:
             "2024.12.0 or later to enable this functionality."
         )
 
-    import fsspec.implementations.asyn_wrapper
+    from fsspec.implementations.asyn_wrapper import AsyncFileSystemWrapper
 
-    return fsspec.implementations.asyn_wrapper.AsyncFileSystemWrapper(fs, asynchronous=True)
+    return AsyncFileSystemWrapper(fs, asynchronous=True)
 
 
 class FsspecStore(Store):


tasansal:
Anything to worry about in the failing tests or are those the usual flaky tests?

joshmoore:
```
FAILED tests/test_store/test_fsspec.py::test_wrap_sync_filesystem_raises - AssertionError: Regex pattern did not match.
 Regex: 'The filesystem .*'
 Input: "No module named 'fsspec.implementations.asyn_wrapper'"
```

tasansal:
> ```
> FAILED tests/test_store/test_fsspec.py::test_wrap_sync_filesystem_raises - AssertionError: Regex pattern did not match.
>  Regex: 'The filesystem .*'
>  Input: "No module named 'fsspec.implementations.asyn_wrapper'"
> ```

Ok I think needs to be imported past the `fsspec` version condition.

src/zarr/storage/_fsspec.py
```diff
@@ -68,6 +68,8 @@ def _make_async(fs: AbstractFileSystem) -> AsyncFileSystem:
             "2024.12.0 or later to enable this functionality."
         )
 
+    import fsspec.implementations.asyn_wrapper
```
maxrjones:
```suggestion
    from fsspec.implementations.asyn_wrapper import AsyncFileSystemWrapper
```

src/zarr/storage/_fsspec.py
```diff
@@ -68,6 +68,8 @@ def _make_async(fs: AbstractFileSystem) -> AsyncFileSystem:
...
+
     return fsspec.implementations.asyn_wrapper.AsyncFileSystemWrapper(fs, asynchronous=True)
```
maxrjones:
```suggestion
    return AsyncFileSystemWrapper(fs, asynchronous=True)
```

maxrjones (commented):
I suggested the slightly more targeted approach that I included in https://github.com/zarr-developers/zarr-python/pull/3193.

d-v-b:
i feel like this should have been caught in our tests. why wasn't it?

maxrjones:
> i feel like this should have been caught in our tests. why wasn't it?

I think because the lines below in the tests mean that the module gets loaded, whereas in a user's workflow the module may not already be loaded since there are the modules are not imported in `fsspec/__init__.py` and `fsspec/implementation/__init__.py`.

- https://github.com/zarr-developers/zarr-python/blob/baabf08d07e8518e3d37bd83c493a1d46ea7ac1d/tests/test_store/test_fsspec.py#L301
- https://github.com/zarr-developers/zarr-python/blob/baabf08d07e8518e3d37bd83c493a1d46ea7ac1d/tests/test_store/test_fsspec.py#L325
- https://github.com/zarr-developers/zarr-python/blob/baabf08d07e8518e3d37bd83c493a1d46ea7ac1d/tests/test_store/test_fsspec.py#L411

src/zarr/storage/_fsspec.py
tasansal: done

tasansal:
Yea having the lazy import inside function could cause some issues like this. One way to fix that is module level import conditional on fsspec version etc; but it also gets ugly :)

tasansal:
@joshmoore seems like this is ready to merge. What are your thoughts on 3.0.9 post-release and yanking 3.0.9?

d-v-b:
IMO we don't need to yank 3.0.9, but we should issue a 3.0.10 release very soon

maxrjones (approved):

d-v-b (approved):

d-v-b:
@meeseeksdev backport to 3.0.10

PR #3198, opened 2025-07-02
fix: restore fill_value=None for zarr_format=2

### LINKED ISSUE #3197: Breaking Change in Fill Value Behavior from 3.0.9 to main for zarr format 2
### Zarr version

3.0.9 vs main

### Numcodecs version

0.16.1

### Python Version

3.11

### Operating System

mac

### Installation

uv run

### Description

The fill value written for the same inputs has changed from zarr 3.0.9 to main. 

With a fill value of `None` for an `int` dtype (and time and I think a few others) and zarr format 2, 3.0.9 writes a fill value of `null` but main writes `0`.

I know there are breaking changes in API with dtypes, but this feels like it changes what data is written, without a warning or error. This is causing data corruption errors in the xarray tests for zarr format 2

cc @d-v-b 

### Steps to reproduce

```python
# /// script
# requires-python = ">=3.11"
# dependencies = [
# "zarr @ git+https://github.com/zarr-developers/zarr-python.git@main"
# ]
# ///

#     "zarr==3.0.9",
import json
import os
from pathlib import Path

import numpy as np
import zarr

# zarr.print_debug_info()
print(f"Zarr version: {zarr.__version__}")

store_path = Path(__file__).parent / "test.zarr"

# Create group exactly like xarray does
group = zarr.open_group(store_path, mode="w", zarr_format=2)

# Create array with None fill_value for int64 dtype
array = group.create(
    "test_array", shape=(10,), dtype=np.int64, fill_value=None, chunks="auto"
)

print(f"array.fill_value: {array.fill_value}")
print(f"array.fill_value type: {type(array.fill_value)}")

# Check JSON metadata
with open(os.path.join(store_path, "test_array", ".zarray")) as f:
    metadata = json.load(f)

print(f"JSON fill_value: {metadata['fill_value']}")
print(f"JSON fill_value type: {type(metadata['fill_value'])}")
```

**3.0.9 output**

```
Zarr version: 3.0.9
array.fill_value: None
array.fill_value type: <class 'NoneType'>
JSON fill_value: None
JSON fill_value type: <class 'NoneType'>
```

**main output**

```
Zarr version: 3.0.9.dev43+gbaabf08d
array.fill_value: 0
array.fill_value type: <class 'numpy.int64'>
JSON fill_value: 0
JSON fill_value type: <class 'int'>
```

### Additional output
You might need to run with uv run --no-cache to force uv to reinstall zarr between runs

ianhi:
Should fix #3197

Without this it is not possible to write "null" to the zarr format 2 fill value. 

@d-v-b 

TODO:

* [NA] New/modified features documented in `docs/user-guide/*.rst`
* [NA] Changes documented as a new file in `changes/`

### DIFF
--- a/src/zarr/core/array.py
+++ b/src/zarr/core/array.py
@@ -774,15 +774,13 @@ class AsyncArray(Generic[T_ArrayMetadata]):
         chunks: ChunkCoords,
         order: MemoryOrder,
         dimension_separator: Literal[".", "/"] | None = None,
-        fill_value: float | None = None,
+        fill_value: Any | None = None,
         filters: Iterable[dict[str, JSON] | numcodecs.abc.Codec] | None = None,
         compressor: CompressorLikev2 = None,
         attributes: dict[str, JSON] | None = None,
     ) -> ArrayV2Metadata:
         if dimension_separator is None:
             dimension_separator = "."
-        if fill_value is None:
-            fill_value = dtype.default_scalar()  # type: ignore[assignment]
         return ArrayV2Metadata(
             shape=shape,
             dtype=dtype,
@@ -806,7 +804,7 @@ class AsyncArray(Generic[T_ArrayMetadata]):
         order: MemoryOrder,
         config: ArrayConfig,
         dimension_separator: Literal[".", "/"] | None = None,
-        fill_value: float | None = None,
+        fill_value: Any | None = None,
         filters: Iterable[dict[str, JSON] | numcodecs.abc.Codec] | None = None,
         compressor: CompressorLike = "auto",
         attributes: dict[str, JSON] | None = None,


src/zarr/core/array.py
```diff
@@ -138,11 +138,43 @@
...
+
+    This is implemented as a singleton.
+    """
+
+    _instance = None
+
+    def __new__(cls) -> Self:
+        if cls._instance is None:
+            cls._instance = super().__new__(cls)
+        return cls._instance
```
d-v-b: is this perhaps too fancy? I feel like a vanilla class is OK here, without the need for singleton-ness. The code that will use instances of this class is already checking against a specific instance. That seems like all we need.

d-v-b: to put it differently, what condition are we guarding against with this implementation of `__new__`, and why would that condition be bad.

d-v-b: > The code that will use instances of this class is already checking against a specific instance.

Correction -- this PR uses an `isinstance` check, so we aren't even checking if `fill_value == DEFAULT_FILL_VALUE`, which IMO is just fine

ianhi: Good call, I'll simplify. I think this was reflex from working with microscope hardware, gotta have single source of truth for interfacing with the real world, but not here :)

tests/test_array.py
```diff
@@ -1012,6 +1012,28 @@ def test_default_fill_value(dtype: ZDType[Any, Any], store: Store) -> None:
...
 
+    @staticmethod
+    # @pytest.mark.parametrize("zarr_format", [2, 3])
+    @pytest.mark.parametrize("dtype", zdtype_examples)
+    @pytest.mark.filterwarnings("ignore::zarr.core.dtype.common.UnstableSpecificationWarning")
+    def test_default_fill_value_None(
+        dtype: ZDType[Any, Any], store: Store, zarr_format: ZarrFormat
+    ) -> None:
+        """
+        Test that the fill value of an array is set to the default value for an explicit None arguement for
```
d-v-b:
```suggestion
        Test that the fill value of an array is set to the default value for an explicit None argument for
```

src/zarr/core/array.py
```diff
@@ -138,11 +138,33 @@
...
+    """
+    Sentinel class to indicate that the default fill value should be used.
+
+    This class is used to distinguish between:
+    Zarr Format 3:
+        - fill_value = None: dtype default value
+    Zarr Format 2:
+        - fill_value = None: fill_value saved as null
+
+    This allows backwards compatibility with zarr format 2.
```
d-v-b:
```suggestion
    Sentinel class to indicate that the default fill value should be used.

   This class exists because the normal values used to convey "defaultness" 
   like ``None`` or ``"auto"` do not work for the fill value parameter of a Zarr array. 
   ``None`` is a valid fill value for Zarr V2 (resulting in ``"fill_value: null"`` in array metadata). 
   A string like ``"auto"`` is a valid fill value for an array with a string data type. 
   But an instance of this class lies outside the space of valid fill values, which makes it 
   suitable for expressing that the default fill value should be used.
```

src/zarr/core/array.py
```diff
@@ -138,11 +138,32 @@
...
+    """
+    Sentinel class to indicate that the default fill value should be used.
+
+   This class exists because the normal values used to convey "defaultness" 
+   like ``None`` or ``"auto"` do not work for the fill value parameter of a Zarr array. 
+   ``None`` is a valid fill value for Zarr V2 (resulting in ``"fill_value: null"`` in array metadata). 
+   A string like ``"auto"`` is a valid fill value for an array with a string data type. 
+   But an instance of this class lies outside the space of valid fill values, which makes it 
+   suitable for expressing that the default fill value should be used.
+    """
```
d-v-b:
```suggestion
    Sentinel class to indicate that the default fill value should be used.

    This class exists because the normal values used to convey "defaultness" 
    like ``None`` or ``"auto"` do not work for the fill value parameter of a Zarr array. 
    ``None`` is a valid fill value for Zarr V2 (resulting in ``"fill_value: null"`` in array metadata). 
    A string like ``"auto"`` is a valid fill value for an array with a string data type. 
    But an instance of this class lies outside the space of valid fill values, which makes it 
    suitable for expressing that the default fill value should be used.
    """
```

src/zarr/api/synchronous.py
```diff
@@ -606,7 +606,7 @@ def create(
     chunks: ChunkCoords | int | bool | None = None,
     dtype: ZDTypeLike | None = None,
     compressor: CompressorLike = "auto",
-    fill_value: Any | None = None,  # TODO: need type
+    fill_value: Any | None = DEFAULT_FILL_VALUE,  # TODO: need type
```
d-v-b, line 609: for posterity, we now have a type alias for the scalar types of all the dtypes which we can use instead of `Any` here

d-v-b: although maybe we should just use `object`, since we do want to allow inputs to be castable

d-v-b (approved):
thanks for this ian! In the interest of velocity I pushed some changes. I'm going to merge this as soon as all the checks pass.

Issue #3196, opened 2025-07-02
Issue Opening Files via SFTP with fsspec in Zarr v3


dgegen:
I'm currently using Zarr and fsspec to access a Zarr file stored on a server via SFTP. However, I've encountered an issue while transitioning from Zarr v2 to v3.

## Description
```python
import zarr
import fsspec

# Specify the SFTP filesystem
storage_options = {
    "key_filename": "/path/to/key",
    "username": "username",
    "host": "host",
}
sftp_fs = fsspec.filesystem('sftp', **storage_options)
zarr_path = "/path/to/file/on/server.zarr"

fs_map = fsspec.FSMap(zarr_path, sftp_fs)

print(f"File exists: {sftp_fs.exists(zarr_path)}")  # True
```

In Zarr v2.18.3 I am able to open this file using
```python
# Zarr v2.18.3
zarr_file = zarr.open(fs_map)
```

Now, attempting to achieve the same thing using Zarr v3.0.9 (including an additional import because of #3195), I use:
```python
# Zarr v3.0.9
import fsspec.implementations.asyn_wrapper  # missing import in zarr.storage.FsspecStore
from zarr.storage import FsspecStore

store = FsspecStore.from_mapper(fs_map, read_only=True)
zarr_file = zarr.open(store)
```
However, the call `zarr_file = zarr.open(store)`, never finishes, i.e. Python freezes.

Is this a bug? Or will SFPT no longer be supported in Zarr v3? Or am I using `FsspecStore` incorrectly?

Thanks a lot in advance!

d-v-b:
> However, the call `zarr_file = zarr.open(store)`, never finishes, i.e. Python freezes.

that's a surprising failure mode. As our 3.0.9 release has been a bit bumpy, can you try running the same thing in 3.0.8? You will need to skip the fsmap and just give the filesystem to the fsspecstore I think.

dgegen:
For me, Python also freezes using the following in v3.0.8
```python
# Zarr v3.0.8
from fsspec.implementations.asyn_wrapper import AsyncFileSystemWrapper
from zarr.storage import FsspecStore

store = FsspecStore(AsyncFileSystemWrapper(sftp_fs, asynchronous=True), path=remote_file_path)
zarr.open(store)
```

d-v-b:
ok,that's "good" in the sense that it's not related to the 3.0.9 release, but it's bad insofar as it's a bug. Can you confirm that this isn't an fsspec issue?

[...]

PR #3207, opened 2025-07-05
Preventing Deadlocks When Reading Metadata Concurrently via `asyncio.gather`

dgegen:
As described in #3196, I encountered issues opening Zarr v3 arrays stored over SFTP using `fsspec`. Specifically, python would freeze opening zarr arrays.

#### Root Cause

The issue stems from the use of `asyncio.gather` in `zarr.core.array.get_array_metadata`, which attempts to read multiple metadata files (e.g., `.zarray`, `.zattrs`, `zarr.json`) concurrently. This works well for truly asynchronous filesystems, but breaks when using systems like `SFTPFileSystem`, which does not seem to be concurrency-safe in async contexts (potentially relying on blocking I/O internally or managing connection states using global locks) leading to deadlocks or indefinite hangs when `asyncio.gather` is used to perform multiple reads simultaneously.

#### Solution

To address this, I’ve implemented a fallback to **sequential reads** for filesystems that are not concurrency-safe. The logic is as follows: For non asynchronous file systems, the user sets `store.fs.asynchronous=False`. The helper function `is_concurrency_safe(store_path: StorePath) -> bool`, checks this `getattr(fs, "asynchronous", True)`. If True `asyncio.gather` is used, else we fall back to sequential `await`. This Preserves the performance benefit of concurrent reads for safe filesystems (e.g., local disk, S3, GCS), while preventing deadlocks and improved robustness when using backends like SFTP.

These changes may not address all scenarios not asynchronous file systems could cause issues, as there are several other instances of `asyncio.gather` in `zarr.core.array` and `zarr.core.group`. However, I opted to focus on this specific problem first, as enabling the opening of arrays and groups is likely the highest priority, and I wanted to discuss this approach before making too many changes.

I look forward to hearing your thoughts and seeing this issue resolved!

TODO:

### DIFF
--- a/src/zarr/abc/store.py
+++ b/src/zarr/abc/store.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
-from asyncio import gather
+from asyncio import as_completed, gather
 from dataclasses import dataclass
 from itertools import starmap
 from typing import TYPE_CHECKING, Protocol, runtime_checkable
@@ -414,8 +414,27 @@ class Store(ABC):
         that objects will be retrieved in the order in which they were requested, so this method
         yields tuple[str, Buffer | None] instead of just Buffer | None
         """
-        for req in requests:
-            yield (req[0], await self.get(*req))
+
+        async def _get_with_name(
+            key: str, prototype: BufferPrototype, byte_range: ByteRequest | None
+        ) -> tuple[str, Buffer | None]:
+            value = await self.get(key, prototype, byte_range)
+            return key, value
+
+        tasks = [_get_with_name(*req) for req in requests]
+        for completed in as_completed(tasks):
+            task = await completed
+            yield task
+
+    async def _get_many_ordered(
+        self, requests: Iterable[tuple[str, BufferPrototype, ByteRequest | None]]
+    ) -> tuple[Buffer | None, ...]:
+        """
+        Retrieve a collection of objects from storage in the order they were requested.
+        """
+        tasks = [self.get(*req) for req in requests]
+
+        return tuple(await gather(*tasks))
 
     async def getsize(self, key: str) -> int:
         """
--- a/src/zarr/abc/store.py
+++ b/src/zarr/abc/store.py
@@ -426,16 +426,6 @@ class Store(ABC):
             task = await completed
             yield task
 
-    async def _get_many_ordered(
-        self, requests: Iterable[tuple[str, BufferPrototype, ByteRequest | None]]
-    ) -> tuple[Buffer | None, ...]:
-        """
-        Retrieve a collection of objects from storage in the order they were requested.
-        """
-        tasks = [self.get(*req) for req in requests]
-
-        return tuple(await gather(*tasks))
-
     async def getsize(self, key: str) -> int:
         """
         Return the size, in bytes, of a value in a Store.
--- a/src/zarr/core/group.py
+++ b/src/zarr/core/group.py
@@ -513,19 +513,23 @@ class AsyncGroup:
             consolidated_key = use_consolidated
 
         if zarr_format == 2:
-            paths = [store_path / ZGROUP_JSON, store_path / ZATTRS_JSON]
+            requests = [
+                (key, default_buffer_prototype(), None) for key in [ZGROUP_JSON, ZATTRS_JSON]
+            ]
             if use_consolidated or use_consolidated is None:
-                paths.append(store_path / consolidated_key)
+                requests.append((consolidated_key, default_buffer_prototype(), None))
 
-            zgroup_bytes, zattrs_bytes, *rest = await asyncio.gather(
-                *[path.get() for path in paths]
+            retrieved_buffers = {key: value async for key, value in store_path.get_many(requests)}
+            zgroup_bytes, zattrs_bytes = (
+                retrieved_buffers[ZGROUP_JSON],
+                retrieved_buffers[ZATTRS_JSON],
             )
+
             if zgroup_bytes is None:
                 raise FileNotFoundError(store_path)
 
             if use_consolidated or use_consolidated is None:
-                maybe_consolidated_metadata_bytes = rest[0]
-
+                maybe_consolidated_metadata_bytes = retrieved_buffers[consolidated_key]
             else:
                 maybe_consolidated_metadata_bytes = None
 
@@ -534,17 +538,18 @@ class AsyncGroup:
             if zarr_json_bytes is None:
                 raise FileNotFoundError(store_path)
         elif zarr_format is None:
+            requests = [
+                (key, default_buffer_prototype(), None)
+                for key in [ZARR_JSON, ZGROUP_JSON, ZATTRS_JSON, consolidated_key]
+            ]
+            retrieved_buffers = {key: value async for key, value in store_path.get_many(requests)}
             (
                 zarr_json_bytes,
                 zgroup_bytes,
                 zattrs_bytes,
                 maybe_consolidated_metadata_bytes,
-            ) = await store_path.get_many_ordered(
-                ZARR_JSON,
-                ZGROUP_JSON,
-                ZATTRS_JSON,
-                consolidated_key,
-            )
+            ) = tuple(retrieved_buffers.get(req[0]) for req in requests)
+
             if zarr_json_bytes is not None and zgroup_bytes is not None:
                 # warn and favor v3
                 msg = f"Both zarr.json (Zarr format 3) and .zgroup (Zarr format 2) metadata objects exist at {store_path}. Zarr format 3 will be used."
@@ -3476,12 +3481,14 @@ async def _read_metadata_v2(store: Store, path: str) -> ArrayV2Metadata | GroupM
     """
     # TODO: consider first fetching array metadata, and only fetching group metadata when we don't
     # find an array
-    zarray_bytes, zgroup_bytes, zattrs_bytes = await store._get_many_ordered(
-        [
-            (_join_paths([path, ZARRAY_JSON]), default_buffer_prototype(), None),
-            (_join_paths([path, ZGROUP_JSON]), default_buffer_prototype(), None),
-            (_join_paths([path, ZATTRS_JSON]), default_buffer_prototype(), None),
-        ]
+    requests = [
+        (_join_paths([path, ZARRAY_JSON]), default_buffer_prototype(), None),
+        (_join_paths([path, ZGROUP_JSON]), default_buffer_prototype(), None),
+        (_join_paths([path, ZATTRS_JSON]), default_buffer_prototype(), None),
+    ]
+    retrieved_buffers = {key: value async for key, value in store._get_many(requests)}
+    zarray_bytes, zgroup_bytes, zattrs_bytes = tuple(
+        retrieved_buffers.get(req[0]) for req in requests
     )
 
     if zattrs_bytes is None:
--- a/src/zarr/core/group.py
+++ b/src/zarr/core/group.py
@@ -539,11 +539,11 @@ class AsyncGroup:
                 zgroup_bytes,
                 zattrs_bytes,
                 maybe_consolidated_metadata_bytes,
-            ) = await asyncio.gather(
-                (store_path / ZARR_JSON).get(),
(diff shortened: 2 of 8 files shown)


d-v-b:
Good detective work here! I think the ideal solution would keep store implementation details confined to the store classes themselves. So instead of the solution here, what if we override the [get_many](https://github.com/zarr-developers/zarr-python/blob/6c40629cadea3bd22699bcee36502bf268778745/src/zarr/abc/store.py#L409) method on the fsspec store to include the logic you have added here, and then use that method instead of multiple gets

dgegen:
Very good point! Perhaps something along these lines?

```python
class StorePath:
    # ...
    async def _is_concurrency_save(self):
        fs = getattr(self.store, "fs", None)
        return getattr(fs, "asynchronous", True)

    async def get_many(
        self,
        *suffixes : str,
        prototype: BufferPrototype | None = None,
        byte_range: ByteRequest | None = None,
    ):
        tasks = [
            (self / suffix).get(prototype=prototype, byte_range=byte_range) for suffix in suffixes
        ]
        if await self._is_concurrency_save():
            return await gather(*tasks)
        else:
            results = []
            for task in tasks:
                result = await task
                results.append(result)
            return results
            
class FsspecStore:
    # ...
    async def _get_many(
        self, requests: Iterable[tuple[str, BufferPrototype, ByteRequest | None]]
    ) -> AsyncGenerator[tuple[str, Buffer | None], None]:
        if getattr(self.fs, "asynchronous", True):
            async for result in super()._get_many(requests=requests):
                yield result
        else:
            for key, prototype, byte_range in requests:
                value = await self.get(key, prototype, byte_range)
                yield (key, value)
                
```

src/zarr/storage/_common.py
```diff
@@ -263,6 +282,10 @@ def __eq__(self, other: object) -> bool:
             pass
         return False
 
+    async def _is_concurrency_save(self):
+        fs = getattr(self.store, "fs", None)
+        return getattr(fs, "asynchronous", True)
```
d-v-b: this should be a method specific to the fsspec store. the fsspec store should call it inside `_get_many` in order to chose which implementation to use.

src/zarr/core/group.py
```diff
@@ -3476,11 +3476,20 @@ async def _read_metadata_v2(store: Store, path: str) -> ArrayV2Metadata | GroupM
...
     # find an array
-    zarray_bytes, zgroup_bytes, zattrs_bytes = await asyncio.gather(
-        store.get(_join_paths([path, ZARRAY_JSON]), prototype=default_buffer_prototype()),
-        store.get(_join_paths([path, ZGROUP_JSON]), prototype=default_buffer_prototype()),
-        store.get(_join_paths([path, ZATTRS_JSON]), prototype=default_buffer_prototype()),
-    )
+    requests = [
+        (_join_paths([path, ZARRAY_JSON]), default_buffer_prototype(), None),
+        (_join_paths([path, ZGROUP_JSON]), default_buffer_prototype(), None),
+        (_join_paths([path, ZATTRS_JSON]), default_buffer_prototype(), None),
```
d-v-b, line 3487: lets bind `_join_paths([path, X])` to a variable so that we don't call the `_join_paths` function so many times. for example:

```
zarray_path = _join_paths([path, ZARRAY_JSON])
...
```

src/zarr/storage/_common.py
dgegen: Done!

dgegen:
Thanks to the feedback I received and some additional testing, I have substantially improved my initial solution. I look forward to hearing your thoughts on it! :)

PS: I'm also not sure why two of the tests failed because they don't seem to relate to any of the changes I've made.

src/zarr/abc/store.py
```diff
@@ -414,8 +414,27 @@ async def _get_many(
...
+            task = await completed
+            yield task
+
+    async def _get_many_ordered(
+        self, requests: Iterable[tuple[str, BufferPrototype, ByteRequest | None]]
+    ) -> tuple[Buffer | None, ...]:
+        """
+        Retrieve a collection of objects from storage in the order they were requested.
+        """
+        tasks = [self.get(*req) for req in requests]
```
d-v-b: I'm not sure I see the use for this method. If store users want fetches to happen in a specific order, then users can call `get` in a loop. If users only want the _results_ of their fetches to be ordered, they can re-order the results _after_ receiving them.

dgegen: I introduced the method `_get_many_ordered` because it is the only used case in the current implementation. Not using it leads to repetitive boilerplate code because we would always find ourselves awaiting all fetches and then sorting the results afterwards. Using `_get_many_ordered` encapsulates this logic, resulting in code that is less verbose, less repetitive, and easier-to-read.

## Example
Compare for example
```python
zarray_bytes, zgroup_bytes, zattrs_bytes = await store._get_many_ordered(
    [
        (_join_paths([path, ZARRAY_JSON]), default_buffer_prototype(), None),
        (_join_paths([path, ZGROUP_JSON]), default_buffer_prototype(), None),
        (_join_paths([path, ZATTRS_JSON]), default_buffer_prototype(), None),
    ]
)
```
with
```python
ordered_keys = [
    (_join_paths([path, ZARRAY_JSON]), default_buffer_prototype(), None),
    (_join_paths([path, ZGROUP_JSON]), default_buffer_prototype(), None),
    (_join_paths([path, ZATTRS_JSON]), default_buffer_prototype(), None),
]

retrieved_objects = {}
async for key, value in store._get_many(ordered_keys):
    retrieved_objects[key] = value

zarray_bytes, zgroup_bytes, zattrs_bytes = tuple(retrieved_objects.get(key[0]) for key in ordered_keys)
```
The first block can be read much faster.

At the same time, we cannot use the original,
```python
zarray_bytes, zgroup_bytes, zattrs_bytes = await asyncio.gather(
    store.get(_join_paths([path, ZARRAY_JSON]), prototype=default_buffer_prototype()),
    store.get(_join_paths([path, ZGROUP_JSON]), prototype=default_buffer_prototype()),
    store.get(_join_paths([path, ZATTRS_JSON]), prototype=default_buffer_prototype()),
```
as this would lead to deadlocks in synchronous file systems, and we of course also don't want to call `.get` sequentially because it reduces performance in asynchronous systems.

d-v-b: > .get sequentially because it reduces performance in asynchronous systems.

`get` is async. caling it returns a coroutine. these can be scheduled together with `asyncio.gather`, which will preserve order. So calling `get` sequentially is not a performance problem.

src/zarr/abc/store.py
```diff
@@ -414,8 +414,27 @@ async def _get_many(
...
+        async def _get_with_name(
+            key: str, prototype: BufferPrototype, byte_range: ByteRequest | None
+        ) -> tuple[str, Buffer | None]:
+            value = await self.get(key, prototype, byte_range)
+            return key, value
+
+        tasks = [_get_with_name(*req) for req in requests]
+        for completed in as_completed(tasks):
+            task = await completed
+            yield task
```
d-v-b, line 427: what is the advantage of this new implementation? the previous implementation was extremely simple, which I think is good for an abc.

src/zarr/abc/store.py
dgegen: Of course, but it is, if we were to await them sequentially, e.g. if we instead used
```python
zarray_bytes = await store.get(_join_paths([path, ZARRAY_JSON]), prototype=default_buffer_prototype()) 
zgroup_bytes = await store.get(_join_paths([path, ZGROUP_JSON]), prototype=default_buffer_prototype())
zattrs_bytes = await store.get(_join_paths([path, ZATTRS_JSON]), prototype=default_buffer_prototype())
```
This would assure that synchronous file systems don't run into deadlocks. But it would not be a good alternative for asynchronous systems.

d-v-b: i'm confused, aren't you doing all this exactly because the fsspec sftp backend is not async? so then sequential awaiting (inside the logic of `_get_many`) is exactly what's we expect to happen, no?

dgegen: Yes, but in the general case, the implementation should be asynchronous. It must be implemented, however, in such a way that we can make it synchronous by overwriting store methods in the FSSpec store, making what is generally asynchronous synchronous if `store.fs.asynchronous==False`.

Note also, that originally, we were not using `await.gather` wrapping multiple `get` statements, so this was not possible. And a general sequential solution is not desirable in an I/O-limited system.

src/zarr/abc/store.py
dgegen: The claim in the docstring is incorrect given the previous implementation.

This loop is sequential: it awaits each self.get(*req) and yields it before moving on to the next. Each request is handled one at a time, in the exact order provided. Therefore, results are always yielded in the same order as the input requests.

It is thus not fully concurrent which would  be desirable in an I/O-limited system and, at least as I understand, kind of defeats the purpose of having an asynchronous `_get_many` method yielding results in the first place. Because if we stick to the order, we might as well await all results and simply replace the implementation of `_get_many` with that of `_get_many_ordered`, making it faster and arguable more easy to used in the asynchronous case. If we want to give the extra flexibility of not awaiting all at once, but still requesting all at the same time, the new implementation would be the right one.

d-v-b: the point of the default implementation is to be the simplest possible implementation of `_get_many` that any child class can safely support, given an implementation of `get`. But child classes should also be able to override this with more efficient methods where applicable, and in these cases the order of results is not guaranteed. hence the type annotation in the original method.

dgegen: I find this somewhat confusing, as I would have expected the standard implementation to be fully asynchronous. However, if the goal is to maximize simplicity, then having an asynchronous implementation that runs synchronously might be the way to go.

That being said, if we revert this to the original, we would only have to also remove the `FsspecStore._get_many` from my current solution. Unless you think we should not have a `_get_many_ordered` method and use the `_get_many` method instead and then always sort the values locally, as they could be of a different order in other implementations.

dgegen:
Resolving this issue would be really useful for my current research project, so I am wondering what I should change specifically to get this pull request approved :)
Should I revert the `_get_many` method to its synchronous version, even though this would mean that the metadata would no longer be retrieved asynchronously? And should I remove the `_get_many_ordered` method and sort the output of the `_get_many` method locally whenever it is called, or can we keep it as it does make the code less verbose, less repetitive, and more readable?

d-v-b:
> Should I revert the `_get_many` method to its synchronous version, even though this would mean that the metadata would no longer be retrieved asynchronously? And should I remove the `_get_many_ordered` method and sort the output of the `_get_many` method locally whenever it is called, or can we keep it as it does make the code less verbose, less repetitive, and more readable?

What if we make the default implementation of `_get_many` async (it could either by a gather or something fancier using as_completed), and remove the `_get_many_sorted` routines for simplicity. I'm not convinced that the burden of sorting results post-hoc is worth the extra method at this time.

dgegen:
Ok. Just to clarify, as you were hinting at potentially using `asyncio.gather` in the `_get_many` method: We still want `_get_many` be a generator, right? In other words, we think that scenarios where an asynchronous generator is indispensable are common enough to warrant its inclusion in the base class.

Assuming this is the case, I will provide a solution with `_get_many` remaining an asynchronous generator.

jhamman:
@dgegen - thanks for working on this. Curious if we've ruled out a potential fix upstream in fsspec?

cc @martindurant

martindurant:
I thought zarr wrapped all non-asynchronous fsspec FSs in the async wrapper, so that you should not have to do anything like this. Then, zarr should only ever be calling asynchronous methods. Is that wrong?

dgegen:
> I thought zarr wrapped all non-asynchronous fsspec FSs in the async wrapper, so that you should not have to do anything like this. Then, zarr should only ever be calling asynchronous methods. Is that wrong?

Zarr does indeed wrap everything in an `AsyncFileSystemWrapper(fs, asynchronous=True)` if `fs.async_impl` is false. However, this does not solve the current problem, which occurs when the file system wrapped by `AsyncFileSystemWrapper` does not actually support concurrent requests. This is the case for `fsspec.filesystem('sftp', **storage_options)`, where we encountered deadlocks when trying to read different metadata files concurrently (as described in detail at the beginning of this PR).

Issue #3185, opened 2025-06-30
keys() returning an empty list on a FsspecStore


bevilacqc:
### Zarr version

3.0.8

### Numcodecs version

0.16.1

### Python Version

3.11.7

### Operating System

Mac

### Installation

using pip into a virtual environment

### Description

When calling the `keys()` function on the root group of a zarr file hosted in a S3 bucket I get an empty list, while accessing directly the key (of a group which I know exists in the file) works.

I am not sure whether it is a bug or there is something wrong in the way I stored the zarr file in the S3 bucket or opened it through the zarr library.

### Steps to reproduce

```python
# /// script
# requires-python = ">=3.11"
# dependencies = [
#   "zarr >= 3.0",
#   "zarr[remote]>=3"
#   "aiohttp"
# ]
# ///
#

import zarr
filename = r'https://s3.embl.de/bls-app-test/test.bls.zarr'

store = zarr.storage.FsspecStore.from_url(url=filename, read_only=True)
print(f"listing supported: {store.supports_listing}")

root = zarr.open_group(store=store, mode='r')

root["Brillouin_data"] # the key 'Brillouin_data' can e found
print(f"keys: {list(root.keys())}") # but when listing the keys in the root an empty list is returned
```

### Additional output

listing supported: True
keys: []

d-v-b:
It's possible that your bucket is not configured to allow listing. can you list the objects in the bucket using the s3 api (or fsspec)?

bevilacqc:
@d-v-b thanks for your input. I checked and listing was already enabled.

I managed to make it work by creating a FsspecStore directly from a fsspec filesystem explicitly set to S3:
```Python
import fsspec                    

fs = fsspec.filesystem('s3', anon=True, asynchronous=True,
                    client_kwargs={'endpoint_url': f"https://s3.embl.de"})

store = zarr.storage.FsspecStore(fs, path = 'bls-app-test/test.bls.zarr',
                                read_only=True)


print(f"listing supported: {store.supports_listing}")

root = zarr.open_group(store=store, mode='r')

print(f"keys: {list(root.keys())}") # Now the keys are listed correctly
```

d-v-b:
[...]

PR #3212, opened 2025-07-07
Update documentation of remote store

bevilacqc:
Add documentation both in the user guide and in the `FsspecStore.from_url` to address issue #3185


d-v-b (approved):
this looks great, thank you!

PR #3216, opened 2025-07-08
Write chunks with negative zero values and a zero fill value

### LINKED ISSUE #3144: Negative zero not preserved
### Zarr version

v3.0.8

### Numcodecs version

v0.15.1

### Python Version

3.11

### Operating System

Mac

### Installation

pip in a venv

### Description

Negative zero (-0.0) is not distinguished from zero when reading back values from a Zarr array. This differs from the expected behaviour (and Zarr Python v2).

### Steps to reproduce

```python
# /// script
# requires-python = ">=3.11"
# dependencies = [
# ]
# ///
#

import numpy as np
import zarr
print(zarr.array(np.array([0.0, -0.0]))[:])
#zarr.print_debug_info()
```

This prints

```
[0. 0.]
```

But with Zarr 2.18.7 it correctly prints

```
[ 0. -0.]
```

### Additional output

_No response_

bojidar-bg:
Fixes #3144.

Using `np.any(self._data)` was inspired by how [Zarr v2 checks for equality with a falsey fill value](https://github.com/zarr-developers/zarr-python/blob/v2.18.7/zarr/util.py#L721).

TODO:


tests/test_array.py
```diff
@@ -904,6 +904,34 @@ def test_write_empty_chunks_behavior(
...
+        zarr_format=zarr_format,
+        dtype="f4",
+        fill_value=fill_value,
+        chunks=(1,),
+        config={"write_empty_chunks": False},
+    )
+
+    # initialize the with the fill value
+    arr[:] = fill_value
+    print(arr[:], fill_value)
```
d-v-b:
```suggestion
```

tests/test_array.py
```diff
@@ -904,6 +904,34 @@ def test_write_empty_chunks_behavior(
...
+    assert arr.nchunks_initialized == 0
+
+    # initialize the with the negated fill value (-0.0 for +0.0, +0.0 for -0.0)
+    arr[:] = -fill_value
+    print(arr[:], fill_value)
```
d-v-b:
```suggestion
```

tests/test_array.py
```diff
@@ -904,6 +904,34 @@ def test_write_empty_chunks_behavior(
...
+        shape=(2,),
+        zarr_format=zarr_format,
+        dtype="f4",
+        fill_value=fill_value,
+        chunks=(1,),
+        config={"write_empty_chunks": False},
+    )
+
+    # initialize the with the fill value
+    arr[:] = fill_value
```
d-v-b: I don't think we need to initialize the array values here. by defining the fill value in the array definition, we already ensure that `array[:]` will return the fill value

bojidar-bg:
Oh, oops, thanks :sweat_smile:

changes/3144.bugfix.rst
```diff
@@ -0,0 +1 @@
+For arrays with ``config={"write_empty_chunks: False}`` and ``fill_value=0.0`` (default), chunks containing negative zeroes are written out.
```
d-v-b:
```suggestion
Ensure that -0.0 is not considered equal to 0.0 when checking if all the values in a chunk are equal to an array's fill value.```

File: src/zarr/core/buffer/core.py [hunk 5]
```diff
@@ -523,6 +523,12 @@ def all_equal(self, other: Any, equal_nan: bool = True) -> bool:
         if other is None:
             # Handle None fill_value for Zarr V2
             return False
+        if other == 0.0 and self._data.dtype.kind not in ("U", "S", "T", "O", "V"):
+            # Handle positive and negative zero
+            if np.any(self._data):  # Check for any truthy value
```
bojidar-bg: ~~Actually, thinking about it, this might break with NaN values...~~
Nevermind, `not not math.nan` is `True` :sweat_smile: JavaScript brain strikes again :joy:

d-v-b:
this test failure seems significant: https://github.com/zarr-developers/zarr-python/actions/runs/16172926021/job/45650861381?pr=3216#step:8:420

dcherian:
> this test failure seems significant

Yes looks like this approach doesn't work for complex number types

d-v-b:
what if we view the array as raw bytes (should be cheap) and compare the raw bytes?

```python
>>> import numpy as np
>>> np.array([0.0]) == np.array([-0.0])
array([ True])
>>> np.array([0.0]).view('V') == np.array([-0.0]).view('V')
array([False])
```

bojidar-bg:
I wonder if that would somehow break with floating point subnormal-s and the like. Will have to experiment :thinking:

bojidar-bg:
Took me a bit, but finally got around to it. Subnormals are fine, and behave as expected; the only difference between Python's float equality and bitwise float equality is that signed zeroes compare as un-equal when comparing their bits, and that nan numbers can sometimes compare as equal when comparing their bits; the former is exactly what we want, and the latter won't occur since the code path is triggered only for signed zero fill values.
```python
>>> import numpy as np
>>> np.array(1e-323).view('V') == np.array(0.0).view('V'), 1e-323 == 0.0
(array(False), False)
>>> np.array(1e-324).view('V') == np.array(0.0).view('V'), 1e-324 == 0.0
(array(True), True)
>>> np.array(-1e-323).view('V') == np.array(-0.0).view('V'), -1e-323 == -0.0
(array(False), False)
>>> np.array(-1e-324).view('V') == np.array(-0.0).view('V'), -1e-324 == -0.0
(array(True), True)
>>> np.array(-0.0).view('V') == np.array(0.0).view('V'), 0.0 == -0.0
(array(False), True)
>>> np.inf * 0.0
nan
>>> np.array(np.nan).view('V') == np.array(np.nan).view('V'), np.nan == np.nan
(array(True), False)
>>> np.array(np.inf * 0.0).view('V') == np.array(np.nan).view('V'), np.inf * 0.0 == np.nan
(array(False), False)
```

d-v-b:
> nan numbers can sometimes compare as equal when comparing their bits

This is actually potentially super useful, because the zarr v3 spec distinguishes between different types of nans, even though numpy does not. In order to ensure that arrays round-trip correctly through zarr python, we need to generate _exactly_ the specific nan defined in the metadata. I did a quick check and numpy will preserve the underlying byte representation of different nans, so this should be possible.

```python
np.array([b'\x00\x00\x00\x00\x00\x00\xFF\xFF'], dtype='|V8').view('float').view('V')
array([b'\x00\x00\x00\x00\x00\x00\xFF\xFF'], dtype='|V8')
```

bojidar-bg:
Oh, that's curious! Probably not something I can quite incorporate in the code here... unless we make all floating point arrays use bitwise comparison for empty chunks.. :thinking:

bojidar-bg:
That `.view("V")` trick fails on GPU with a `ZeroDivisionError`, presumably in [cupy:_core/core.pyx:81](https://github.com/cupy/cupy/blob/c7b45b96c3be6acfab47464abe268ccf89e993b4/cupy/_core/core.pyx#L819) when `v_is == dtype.itemsize == 0`, as is the case for the `"V"` dtype... 
Use structured `np.void` dtypes to achieve the same idea doesn't work because `np.void` isn't hashable, but it works when using a sized "V" dtype, like "V16". ~~Let's see if that's enough to get all tests green :joy:~~ YES IT IS! :tada: That GPU test was stubborn! :joy::joy:

tests/test_array.py
```diff
@@ -870,6 +870,30 @@ def test_write_empty_chunks_behavior(
...
+        dtype=dtype,
+        fill_value=fill_value,
+        chunks=(1,),
+        config={"write_empty_chunks": False},
+    )
+    assert arr.nchunks_initialized == 0
+
+    # initialize the with the negated fill value (-0.0 for +0.0, +0.0 for -0.0)
+    arr[:] = -fill_value
+    assert arr.nchunks_initialized == arr.nchunks
```
d-v-b, line 894: this test is fine but ideally we would be testing the altered function explicitly, instead of indirectly via array creation + chunk writing. this is not a blocker for this PR, just something to sort out down the road

bojidar-bg: That test is basically copied from `test_write_empty_chunks_behavior` right above it. But yeah, it might be worth to have both a unit and an integration test in this case (:

PR #3225, opened 2025-07-10
remove usage of the _deprecate_positional_args decorator, and tests for its behavior

d-v-b:
TODO:


d-v-b:
there's a reduction in coverage, but the previously covered lines of code were not actually being tested for correctness -- they were just being run to check the behavior of the positional args decorator. So I think we can disregard the coverage reduction here.

d-v-b:
[this failure](https://github.com/zarr-developers/zarr-python/actions/runs/16227331719/job/45822176450?pr=3225#step:8:339) is due to recent changes to our hypothesis tests, presumably revealing a real bug, but likely one unrelated to this PR (cc @dcherian)

dstansby:
This breaks the hypothesis tests and pre-commit...

d-v-b:
pre-commit is failing, and hypothesis is also reporting a failed test, but not because of this PR

dstansby:
There were pre-existing pre-commit errors before this PR, but this introduced more. And I'm pretty sure this broke the hypothesis tests? They were working before this PR was merged: https://github.com/zarr-developers/zarr-python/actions/runs/16271528357/job/45940137517

d-v-b:
compare the changes in this PR (removing a decorator that only raised a warning, and some tests for that decorator) with the pre-commit errors (type-checking failures in obstore and hypothesis tests). I don't see how they are related other than by timing. We should just fix the pre-commit errors (by correcting type hints as needed). 

As for the hypothesis test failure, I don't see how that could be related to the changes in this PR since it pertains to deleting groups from a local file system-backed store. we should of course also fix this. but it seems very unlikely to be caused by the changes in this PR.

dstansby:
> I don't see how they are related other than by timing.

If I run pre-commit on the `main` commit before this was merged (2f64d670098b6f06f6c752c40dee481858bbfb62), I get 7 errors. If I run it after this was merged (b267ff66f76f8344fcb29f92972604a947869f75) I get 14 errors. I think that's a pretty clear smoking gun that the new errors were caused by this PR, and this PR changed the acceptable call signatures of lots of functions/methods, I don't think it's that unlikely. For the same reason I don't think it's unlikely that the hypothesis failures were caused by this PR.

d-v-b:
Good detective work! Do you want to work on fixing the mypy problems? And @dcherian do you have any idea what's causing the hypothesis test failures?

Issue #3104, opened 2025-05-30
consider removing default compressor / filters / serializer from config


d-v-b:
Our config right now contains this logic for defining a default encoding scheme for a given data type:

https://github.com/zarr-developers/zarr-python/blob/af55fcfaefa42b5ef556b1b5be33dcdd06a7fd0b/src/zarr/core/config.py#L85-L107

This approach is problematic because it requires dividing our data types into separate categories which are not very well defined -- is a fixed-length utf32 data type a "string" or "numeric" type?

Given the changes coming in #2874, I propose the following alteration to our approach here:

- Pull this stuff out of the config entirely.

- Confine all this logic to a single function for automatically picking a chunk encoding based on a data type + a requested chunk encoding. This function should also check for incompatibility between a data type and a requested chunk encoding. For example, if someone requests a variable-length string data type but does not specify vlen-utf8 as a serializer, then they should get a clear, early error.

These would be breaking changes, but our current approach is, IMO, unworkable.

d-v-b:
note: any changes to the config API, e.g. those in #2874, should be preceded by deprecations wherever possible. This is supported by donfig: https://donfig.readthedocs.io/en/latest/configuration.html#deprecations. 

xref: https://github.com/zarr-developers/zarr-python/pull/2874#discussion_r2132945621

d-v-b:
closed via #3228

PR #3228, opened 2025-07-10
remove chunk encoding from config

d-v-b:
This PR removes the default chunk encoding parameters from the config.

Prior to the data types refactor, when we inferred the chunk encoding automatically, we categorized the array data type into one of a few categories like "numeric", "string", "bytes". These categories were created before we added datetimes and fixed-length strings, and they don't really make sense after those data types were added: 

- "string" and "bytes" are not real categories, because they do not distinguish between fixed length and variable length bytes / strings. The fixed-length types are basically "numeric", but the variable length types must use the vlen utf8 and vlen bytes codecs.  
- datetime data, fixed-length strings, and structured data types are not obviously "numeric"

It's not really possible to preserve the old categories in the config now that we have more data types that resist categorization. I don't think it's really possible break the data types down into a small number of general categories, and it's less possible / attractive to put this categorization logic in our config. 

I think a better solution is for zarr python to choose sane defaults; when users want something other than the default, they specify it in the function call. In practice these sane defaults are the same codecs for all data types except the variable-length data types, which get special treatment (data types with endianness in zarr v3 also get a `bytes` codec with an endianness parameter). That's what I added in this PR. The default chunk encoding parameters each come from functions that are inaccessible to configuration.

I did a quick github survey to find users who might be using the zarr config to change the default codecs, and I could only find 2: @TomAugspurger  uses the config [here](https://github.com/TomAugspurger/zarr-python-memory-benchmark/blob/6d6d74e6d5d3d877a974757cfa196ae0a3b2e695/main.py#L62) and https://github.com/dcangst in [here](https://github.com/ethz-tb/orcAI/blob/d47ead3010726837a2732611bfa5ceb2416f4aee/src/orcAI/io.py#L321).

### DIFF
--- a/src/zarr/core/array.py
+++ b/src/zarr/core/array.py
@@ -30,6 +30,8 @@ from zarr.abc.codec import ArrayArrayCodec, ArrayBytesCodec, BytesBytesCodec, Co
 from zarr.abc.store import Store, set_or_delete
 from zarr.codecs._v2 import V2Codec
 from zarr.codecs.bytes import BytesCodec
+from zarr.codecs.vlen_utf8 import VLenBytesCodec, VLenUTF8Codec
+from zarr.codecs.zstd import ZstdCodec
 from zarr.core._info import ArrayInfo
 from zarr.core.array_spec import ArrayConfig, ArrayConfigLike, parse_array_config
 from zarr.core.attributes import Attributes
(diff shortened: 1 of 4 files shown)


d-v-b:
closes https://github.com/zarr-developers/zarr-python/issues/3104

normanrz (approved):
I think that makes sense!

TomAugspurger:
FYI @d-v-b there's still some references to `v3_default_compressors` in the docs and docstrings: https://github.com/search?q=repo%3Azarr-developers%2Fzarr-python%20v3_default_compressors&type=code

d-v-b:
> FYI @d-v-b there's still some references to `v3_default_compressors` in the docs and docstrings: https://github.com/search?q=repo%3Azarr-developers%2Fzarr-python%20v3_default_compressors&type=code

Are these not fixed by #3390?

TomAugspurger:
Possibly! I'm behind on reviews :)

PR #3242, opened 2025-07-14
fix obstore type hints

d-v-b:
mypy was flagging problems with some of the obstore type annotations. I don't know what change precipitated these new errors, but this PR silences them by adding type: ignore statements. Obviously fixing the type annotations would be even better, but several obvious things that I attempted did not work, so I think we need an obstore expert for a real fix (@kylebarron)

until then, this PR keeps our CI clean

TODO:

### DIFF
--- a/src/zarr/storage/_obstore.py
+++ b/src/zarr/storage/_obstore.py
@@ -16,7 +16,7 @@ from zarr.abc.store import (
 from zarr.core.config import config
 
 if TYPE_CHECKING:
-    from collections.abc import AsyncGenerator, Coroutine, Iterable
+    from collections.abc import AsyncGenerator, Coroutine, Iterable, Sequence
     from typing import Any
 
     from obstore import ListResult, ListStream, ObjectMeta, OffsetRange, SuffixRange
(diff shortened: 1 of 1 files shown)


d-v-b:
I'm very confused now, because the type:ignore statements necessary to pass mypy locally fail our mypy CI job. I'm removing those type hints and re-running things.

d-v-b:
test failures are unrelated.

TomAugspurger:
> mypy locally fail our mypy CI job

You might need to update your version of obstore locally.

TomAugspurger (approved):

d-v-b:
> > mypy locally fail our mypy CI job
> 
> You might need to update your version of obstore locally.

I was running mypy against obstore upstream, so maybe I needed to _downgrade_

kylebarron:
Indeed in Obstore we switched from `list` to `Sequence` because `Sequence` is covariant while `list` is invariant. While the physical type returned is indeed a list, typing it as a `Sequence` more faithfully represents the semantics. It was also necessary for the work with obspec

PR #3243, opened 2025-07-14
don't test self-deletion in stateful tests

d-v-b:
This PR prevents the stateful group tests from deleting the root group

TODO:

### DIFF
--- a/src/zarr/testing/stateful.py
+++ b/src/zarr/testing/stateful.py
@@ -343,6 +343,8 @@ class ZarrHierarchyStateMachine(SyncMixin, RuleBasedStateMachine):
     @precondition(lambda self: len(self.all_groups) >= 2)  # fixme don't delete root
     @rule(data=st.data())
     def delete_group_using_del(self, data: DataObject) -> None:
+        # ensure that we don't include the root group in the list of member names that we try
+        # to delete
         member_names = tuple(filter(lambda v: "/" in v, sorted(self.all_groups)))
         group_path = data.draw(st.sampled_from(member_names), label="Group deletion target")
         prefix, group_name = split_prefix_name(group_path)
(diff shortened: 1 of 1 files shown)


src/zarr/testing/stateful.py
```diff
@@ -343,9 +343,8 @@ def delete_array_using_del(self, data: DataObject) -> None:
     @precondition(lambda self: len(self.all_groups) >= 2)  # fixme don't delete root
     @rule(data=st.data())
     def delete_group_using_del(self, data: DataObject) -> None:
-        group_path = data.draw(
-            st.sampled_from(sorted(self.all_groups)), label="Group deletion target"
-        )
+        members = tuple(filter(lambda v: "/" in v, sorted(self.all_groups)))
```
dcherian: 1. this will filter out `foo/bar` too no?
```suggestion
        members = tuple(filter(lambda v: v == "/", sorted(self.all_groups)))
```

2. Should this op be an error?

d-v-b: > this will filter out foo/bar too no?

It will reject any member name that _doesn't_ contain "/", so it will keep `foo/bar`

d-v-b:
closes #3244

src/zarr/testing/stateful.py
dcherian: I'm confused about how this is working. Can you add a comment please?

d-v-b: added in https://github.com/zarr-developers/zarr-python/pull/3243/commits/724e3c9d2ceafc8bca2130a53b97a6c3667804d8

dcherian (approved):

src/zarr/testing/stateful.py
d-v-b: let me know if this explanation needs more context, or if there's just a better way to get the same result -- I'm assuming it's a given that the name of the root group does not contain a "/" character.

Issue #3247, opened 2025-07-15
Confusing constructors for new data types


TomNicholas:
I'm trying to [update VirtualiZarr](https://github.com/zarr-developers/VirtualiZarr/issues/676) to work with the new data types, and just wanted to note some ways in which I found the new API confusing.

I briefly skimmed the [narrative docs page](https://zarr.readthedocs.io/en/stable/user-guide/data_types.html), read the phrase "native dtypes", and then found a the [`zarr.dtype.ZDType.from_native_dtype()`](https://zarr.readthedocs.io/en/stable/api/zarr/dtype/index.html#zarr.dtype.ZDType.from_native_dtype) constructor classmethod in the API docs. But that behaved very counterintuitively:

```python
In [1]: from zarr.dtype import ZDType, parse_data_type

In [2]: import numpy as np

In [3]: np.dtype('int32')
Out[3]: dtype('int32')

In [4]: ZDType.from_native_dtype(np.dtype('int32'))
```
This returns `None`?! 

Looking at the docs more closely, apparently I would have to do `Int32.from_native_dtype(np.dtype('int32'))`? But that's no help to me if I don't already have the zarr dtype.

I'm also confused as to why passing the base class succeeds at all. If `Int32` is a subclass of `ZDType`, and `.from_native_dtype` is intended to be used on the subclasses, then why isn't it abstract on the base class? 

It seems what I actually needed was `parse_data_type`:

```python
In [6]: parse_data_type(np.dtype('int32'), zarr_format=3)
Out[6]: Int32(endianness='little')
```

which isn't mentioned explicitly in the narrative docs, nor uses the word "native" anywhere, even though it seems to parse native dtypes.

After reading the full docs page, I still don't think it would be clear how to get from a native dtype to a zarr data type. Presumably that information should be part of the section on [Data Type Resolution](https://zarr.readthedocs.io/en/stable/user-guide/data_types.html#data-types-in-zarr-version-3).

---

Aside: It's also confusing how the docs page says

> Zarr V3 data types do not have endianness.

but that `Int32(endianness='little')` object certainly seems to have endianess...

/ end rant 😅

d-v-b:
> But that's no help to me if I don't already have the zarr dtype.

There isn't in general a 1:1 mapping between numpy data types and zarr data types. So you either need to know the zarr data type you want, or you use the dynamic data type resolution routines.

> `.from_native_dtype` is intended to be used on the subclasses, then why isn't it abstract on the base class?

[...]

PR #3251, opened 2025-07-15
make zdtype abstract methods raise notimplementederror

d-v-b:
[Description of PR]

TODO:


d-v-b:
partially addresses #3247

src/zarr/core/dtype/wrapper.py
```diff
@@ -135,7 +132,7 @@ def to_native_dtype(self: Self) -> TDType_co:
         TDType
             The native data type wrapped by this ZDType.
         """
-        ...
+        raise NotImplementedError  # pragma: no cover
```
TomNicholas, line 135: It's bizarre that this would even be necessary - the method is abstract!

d-v-b: i think abstract means that it's an error to create a subclass that doesn't implement this method, but that's technically compatible with this method doing something.

TomNicholas: Riiiight. So perhaps the better question is: why isn't this base class an `abc.ABC`? 

EDIT: I guess that only prevents you instantiating it, but classmethods don't require instantiation...? This whole thing is weird.

d-v-b:
> I think you missed a couple of methods, specifically `_from_json_v2` and `_from_json_v3`

good catch, I added these in https://github.com/zarr-developers/zarr-python/pull/3251/commits/f7978d97defaf3d80ccd1b4700b5f5d39713f7c5

dstansby (approved):

Issue #1798, opened 2024-04-17
[V3] v2 -> v3 data migration


d-v-b:
We should invest in tools to make the v2 -> v3 conversion simple for people who are motivated to convert their data. A few high-level ideas:
- A simple CLI that converts an array, or a group (recursive or not) from v2 to v3, in a new location.
- - ~~Someone should investigate how complicated in-place conversions would be. On a local filesystem where `mv` is cheap, this could be attractive.~~. V3 is designed to make array conversions easy, requiring only the creation of new metadata.
- - The CLI should use functions that are accessible from scripts that don't use the CLI. We can look at work @normanrz did in [Zarrita](https://github.com/scalableminds/zarrita/blob/b58c257221f012621f9084c9e1be8d526d8b34eb/zarrita/array_v2.py#L452).
- Documentation of the key differences between `zarr-python` v2 and v3, and a migration guide. This should have its own page in the docs.
- - We should consider options for people who don't want to re-save their data. I'm not presently a kerchunk user, but I presume that kerchunk could map v2 to v3, for people who don't want to convert their data? cc @martindurant.

jeromekelleher:
Big +1 on this. I'm working on a conversion tool for large-scale genomics data (100s TB scale) which is usually held in file systems (for the moment, it will probably migrate to object stores later on). A CLI tool that does an in-place migration from v2 to v3 would be a big help. I'm hoping to move the v3 early on, before too many datasets are converted into v2 format and so most users won't ever know about v2.

My assumptions was that the migration was largely a case of writing a new JSON metadata file per-array, and should be possible to do both cheaply and safely?

d-v-b:
> My assumptions was that the migration was largely a case of writing a new JSON metadata file per-array, and should be possible to do both cheaply and safely?

Yes, I think this is right. Besides the metadata, which will live in a completely new JSON document (`zarr.json`), V3 supports a [backwards-compatible layout for the chunks](https://github.com/zarr-developers/zarr-specs/blob/81317acfc1441ff4274bcfbf67386e9852892969/docs/v3/core/v3.0.rst?plain=1#L543)

jeromekelleher:
Thanks yes, I've been aiming for v3 forwards compatibility by using "/" as the default dimension separator.  Then, iterating over the chunks in the first dimension and renaming to have a "c" prefix should be relatively cheap (I forgot about this difference).

Is there some developer documentation with recommendations for forwards/backwards compatibility?

[...]

Issue #3256, opened 2025-07-16
Numcodecs Delta filter throws AttributeError when astype is specified


K-Meech:
### Zarr version

v3.1.0

### Numcodecs version

v0.16.1

### Python Version

3.13

### Operating System

Windows

### Installation

pip into a virtual environment

### Description

Running the code snippet in the 'steps to reproduce' section below, throws the following error:
```
  File "C:\Users\kimme\miniforge3\envs\zarr-stock-only-2\Lib\site-packages\zarr\core\common.py", line 89, in run
    return await func(*item)
           ^^^^^^^^^^^^^^^^^
  File "C:\Users\kimme\miniforge3\envs\zarr-stock-only-2\Lib\site-packages\zarr\abc\codec.py", line 447, in wrap
    return await func(chunk, chunk_spec)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\kimme\miniforge3\envs\zarr-stock-only-2\Lib\site-packages\zarr\codecs\bytes.py", line 80, in _decode_single
    dtype = chunk_spec.dtype.to_native_dtype()
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'numpy.dtypes.Int32DType' object has no attribute 'to_native_dtype'
```

In previous zarr-python versions (e.g. `3.0.10`), this passes without error. Removing the `astype` option to `Delta` also results in the code passing without error.

### Steps to reproduce

Code snippet based on [example from zarr-python docs](https://zarr.readthedocs.io/en/stable/user-guide/arrays.html#filters) with different `dtype` / `astype` options to the `Delta` filter. This throws an error:
```python
# /// script
# requires-python = ">=3.11"
# dependencies = [
#   "zarr@git+https://github.com/zarr-developers/zarr-python.git@main",
# ]
# ///
#
# This script automatically imports the development branch of zarr to check for issues

from numcodecs.zarr3 import Delta
import numpy as np
import zarr

filters = [Delta(dtype='<i4', astype='<i4')]
compressors = zarr.codecs.BloscCodec(cname='zstd', clevel=1, shuffle=zarr.codecs.BloscShuffle.shuffle)
data = np.arange(100000000, dtype='int32').reshape(10000, 10000)
z = zarr.create_array(
    store='data/delta-filter-v3.zarr', 
    shape=data.shape, 
    dtype=data.dtype, 
    chunks=(1000, 1000), 
[...]

Issue #3162, opened 2025-06-22
a codec simplification plan


d-v-b:
I propose that this code should work:

```
# /// script
# requires-python = ">=3.11"
# dependencies = [
#     "zarr",
#     "pytest"
# ]
# ///

import zarr
from numcodecs import GZip
import pytest

compressors = (GZip(), zarr.codecs.GzipCodec())
zarr_formats = (2,3)
@pytest.mark.parametrize('compressor', compressors)
@pytest.mark.parametrize('zarr_format', zarr_formats)
def test(compressor, zarr_format):
    x = zarr.create_array(
        {},
        shape=(10,), 
        dtype='uint8', 
        compressors=compressor, 
        zarr_format=zarr_format)

if __name__ == '__main__':
    pytest.main([__file__, f'-c {__file__}'])
```
```
../../.cache/uv/environments-v2/test-48d946355ee2ef42/lib/python3.11/site-packages/zarr/core/array.py:4723: TypeError
==================================================================================================== short test summary info ====================================================================================================
FAILED  /home/bennettd/dev/zarr-python::test[2-compressor1] - ValueError: Invalid compressor. Expected None, a numcodecs.abc.Codec, or a dict representation of a numcodecs.abc.Codec. Got <class 'zarr.codecs.gzip.GzipCodec'> instead.
FAILED  /home/bennettd/dev/zarr-python::test[3-compressor0] - TypeError: 'GZip' object is not iterable
```

Ignoring the details of the errors here, the reason those errors exist is because  our codec handling is weird and requires the use of separate objects (`zarr.codecs.GzipCodec` for v3, `numcodecs.GZip` for v2) to express the same thing (gzip compression).

Here is a proposal to fix this:

- we add methods to our codec base class which enables it to handling zarr v2 and zarr v3 metadata.  This means the exact same codec class can be used for zarr v2 or zarr v3. This is how the new dtypes work and I think it's a good design.
- We define a protocol, in this repo, that models the structure of the numcodecs codec abstract base class. That codec would look like this:
```python

from typing import Protocol, ClassVar, runtime_checkable

@runtime_checkable
class Numcodec(Protocol):
    codec_id: ClassVar[str]

    def encode(self, buf: ArrayLike) -> ArrayLike:
        ...

    def decode(self, buf: ArrayLike) -> ArrayLike:
        ...
    
    def get_config() -> Mapping[str, Object] # this return type enables typed dicts
        ...
    @classmethod
[...]

PR #3257, opened 2025-07-16
Add CLI for converting v2 metadata to v3

K-Meech:
For https://github.com/zarr-developers/zarr-python/issues/1798

Adds a CLI using [`typer`](https://typer.tiangolo.com/) to convert v2 metadata (`.zarray` / `.zattrs`...) to v3 metadata `zarr.json`. 

To test, you will need to install the new optional cli dependency e.g.
`pip install -e ".[remote,cli]"`

This should make the `zarr-converter` command available e.g. try:
```
zarr-converter --help
zarr-converter convert --help
zarr-converter clear --help
```
`convert` adds `zarr.json` files to every group / array, leaving the v2 metadata as-is. A zarr with both sets of metadata can still be opened with `zarr.open`, but will give a UserWarning: `Both zarr.json (Zarr format 3) and .zarray (Zarr format 2) metadata objects exist... Zarr v3 will be used.`. This can be avoided by passing `zarr_format=3` to `zarr.open`, or by using the `clear` command to remove the v2 metadata.

`clear` can also remove v3 metadata. This is useful if the conversion fails part way through e.g. if one of the arrays uses a codec with no v3 equivalent.

All code for the cli is in `src/zarr/core/metadata/converter/cli.py`, with the actual conversion functions in `src/zarr/core/metadata/converter/converter_v2_v3.py`. These functions can be called directly, for those who don't want to use the CLI (although currently they are part of `/core` which is considered private API, so it may be best to move them elsewhere in the package).

Some points to consider:
- I had to modify `set_path` from `test_dtype_registry.py` and `test_codec_entrypoints.py`, as they were causing the CLI tests to fail if they were run after. This seems to be due to the `lazy_load_list` of the numcodecs codecs registries being cleared, meaning they were no longer available in my code which finds the `numcodecs.zarr3` equivalent of a numcodecs codec.
- I tested this on local zarr images, so it would be great if someone with access to s3 / google cloud etc., could try it out on some small example images there.
- I'm happy to add docs about how to use the CLI, but wanted to get feedback on the general structure first

TODO:


d-v-b:
this is awesome! A few high-level suggestions:

- can we have the name of the cli be just `zarr`, where `convert` is a subcommand: 
  ```
  $ zarr --help
   # prints info including available commands
  $ zarr convert --help
  # info about the convert operation
  ```

- can the 2->3 migration just be a particular invocation of a more general command, like
  ```zarr convert source/data.zarr --in-place --zarr-format 3```

  I'm not sure how to express the "clean up a broken conversion" process through this API, but it should be doable

- is "convert" the best name here? "resave" also seems good. I'm not opposed to "convert", just interested in exploring the space of names a bit.

K-Meech:
Thanks for the comments @d-v-b !
- Happy to change the top-level name to `zarr` 

- Just to check - are you suggesting it would be best to combine the `convert` and ` clear` commands into one command? If so, then it should be possible with something like `zarr convert source/data.zarr --in-place --allow-overwrite` where:
  - `--in-place` would create v3 metadata + remove any v2 metadata (without this specified, both v2/v3 metadata would remain)
  - `--allow-overwrite` would remove any existing v3 metadata, then create new (otherwise it errors if an existing `zarr.json` is encountered)
  - I think the only situation this doesn't cover is if the conversion fails part way through + you just want to remove any `zarr.json` that were written. Perhaps it would be best to prevent this scenario from occurring in the first place? I could always run `convert_v2_to_v3` inside a `try / except`, then run `remove_metadata` if any exceptions are encountered. This could be optional behaviour with something like `--clear`

- As for the naming - I don't have a strong preference here. I'd probably prefer `convert` as I think it's a simpler  / easier to understand term, but happy to change this.

d-v-b:
> I think the only situation this doesn't cover is if the conversion fails part way through + you just want to remove any zarr.json that were written. Perhaps it would be best to prevent this scenario from occurring in the first place? I could always run convert_v2_to_v3 inside a try / except, then run remove_metadata if any exceptions are encountered. This could be optional behaviour with something like --clear

Attempting to clean up on failure is a good idea, but that can also fail. It would still be possible for the program to be interrupted while `remove_metadata` is being run, so I don't think there's anything we can do prevent conversion from failing. But as long as we are only creating json documents, and we aren't deleting the originals (the v2 metadata), then re-running should be cheap.  This seems like a detail that we can iron out later on.

> --in-place would create v3 metadata + remove any v2 metadata (without this specified, both v2/v3 metadata would remain)

There might actually be situations where people want the v2 and v3 metadata in the same hierarchy, and this is not against the rules of the spec. So I'd say that creating v3 metadata should be a separate (and precedent) operation from deleting the v2 metadata. `--in-place` should be considered shorthand for this:

```
$ zarr convert foo.zarr/bar foo.zar/bar
```
i.e., taking a v2 hierarchy from foo.zarr/bar, and creating a v3 hierarchy at foo.zarr/bar. Thus I think 
```
$ zarr convert foo.zarr baz.zarr
```
i.e., resaving the hierarchy somewhere else, should also be supported. For now users would be responsible for copying chunks in this case. 

> --allow-overwrite would remove any existing v3 metadata, then create new (otherwise it errors if an existing zarr.json is encountered)

that seems right, but instead of `--allow-overwrite` I would just do `--overwrite`.

And a broader request: I am sure that while working on this kind of low-level tool, you find tasks that should be direct, but are made difficult by design choices in zarr-python, or the lack of low-level features. If you have the patience for it, please keep track of the stuff we could change in the core library to make this kind of tool easier to write, and feel free to open issues to track these features.

K-Meech:
Ok - that all sounds good. How about the following for an updated CLI:

The base command is `zarr convert path/to/input.zarr` - this creates v3 metadata, writing it in-place back to `input.zarr`, leaving the v2 metadata as-is. There are then the following optional commands on top:
- `--output path/to/output.zarr` - if provided, v3 metadata will be written to this separate output location (leaving `input.zarr` completely unchanged). Only the metadata will be created (no copying of chunks).
- `--overwrite`: allows overwrite of v3 metadata at the output location (either `path/to/input.zarr` or `path/to/output.zarr` if provided)
- `--remove-v2`: removes any v2 metadata at the output location

I think it's probably worth also keeping the `clear` command to give maximum flexibility. I've certainly used it a lot during testing of the CLI e.g. when converting in-place, then doing some testing of the v3 output to check it looks ok, then later removing the v2 metadata when it's no longer needed.

d-v-b:
> * `--output path/to/output.zarr` - if provided, v3 metadata will be written to this separate output location (leaving `input.zarr` completely unchanged). Only the metadata will be created (no copying of chunks).

This design would make the in-place conversion the default, and writing the hierarchy to a new path would be an opt-in kind of thing. Speaking for myself, I think I would expect the convert-to-somewhere-else mode to be the default, and the in-place conversion to be opt-in. It might be helpful to sample opinions from other folks here (cc @zarr-developers/python-core-devs). In any case this is a minor thing.


> I think it's probably worth also keeping the `clear` command to give maximum flexibility. I've certainly used it a lot during testing of the CLI e.g. when converting in-place, then doing some testing of the v3 output to check it looks ok, then later removing the v2 metadata when it's no longer needed.

That makes perfect sense. Since "clear" is kind of ambiguous on its own, and since the `convert` command is pretty narrowly scoped to converting from v2 to v3 data, I wonder if we should rename this command to something like `convert-v3`, `migrate-v3`, etc, something that makes it really obvious that all the options will be scoped to the process of moving from v2 data to v3 data. This would free up the `convert` command for later use, e.g. things like changing chunking, compression, etc.

d-v-b:
also, in case anyone wants to try this tool out, here's a 1-liner with uvx:

```
uvx --with typer --from git+https://github.com/K-Meech/zarr-python/@km/v2-v3-conversion zarr-converter --help
```

src/zarr/core/metadata/converter/converter_v2_v3.py
```diff
@@ -0,0 +1,256 @@
...
+        the Store constructor for that implementation. Ignored otherwise.
+    dry_run : bool, optional
+        Enable a 'dry run' - files that would be created are logged, but no files are actually created.
+    """
+
+    zarr_v2 = zarr.open(store=store, mode="r+", path=path, storage_options=storage_options)
+    convert_array_or_group(zarr_v2, dry_run=dry_run)
+
+
+def convert_array_or_group(zarr_v2: Array | Group, dry_run: bool = False) -> None:
```
dstansby: To keep the API/function calling a bit simpler, I would:
- define a `convert_array` and `convert_group` function
- rename this to `convert_to_v3` (makes it a bit clearer what it's doing)

src/zarr/core/metadata/converter/converter_v2_v3.py
```diff
@@ -0,0 +1,256 @@
...
+    """Convert all v2 metadata in a zarr array/group to v3. Note - if a group is provided, then
```
dstansby:
```suggestion
    """Convert all v2 metadata in a zarr array/group to v3.
    
    Note - if a group is provided, then
```
Docstrings should start with a single sentence, then break into a longer paragraph if needed (see https://numpydoc.readthedocs.io/en/latest/format.html#short-summary)

src/zarr/core/metadata/converter/converter_v2_v3.py
```diff
@@ -0,0 +1,256 @@
...
+        )
+        sync(_save_v3_metadata(zarr_v2, group_metadata_v3, dry_run=dry_run))
+
+    else:
+        # write array's converted metadata
+        array_metadata_v3 = _convert_array_metadata(zarr_v2.metadata)
+        sync(_save_v3_metadata(zarr_v2, array_metadata_v3, dry_run=dry_run))
+
+
+async def remove_metadata(
```
dstansby: Am I right in thinking this isn't required by the conversion functionality? If not, I'd be tempted not to implement it at least in this PR to keep this PR tightly scoped and therefore easier to review. I have some comments/questions about performance implications here, but thought maybe that could be discussed in a follow up PR?

src/zarr/core/metadata/converter/converter_v2_v3.py
```diff
@@ -0,0 +1,256 @@
...
+
+    return ArrayV3Metadata(
+        shape=metadata_v2.shape,
+        data_type=metadata_v2.dtype,
+        chunk_grid=metadata_v2.chunk_grid,
+        chunk_key_encoding=chunk_key_encoding,
+        fill_value=metadata_v2.fill_value,
+        codecs=codecs,
+        attributes=metadata_v2.attributes,
+        dimension_names=None,
```
dstansby: A thought - it might be nice to give the option of adding dimension names during conversion? That could go in a follow up PR though.

src/zarr/core/metadata/converter/converter_v2_v3.py
```diff
@@ -0,0 +1,256 @@
...
+
+
+def _convert_filters(metadata_v2: ArrayV2Metadata) -> list[ArrayArrayCodec]:
+    if metadata_v2.filters is None:
+        return []
+
+    filters_codecs = [_find_numcodecs_zarr3(filter) for filter in metadata_v2.filters]
+    for codec in filters_codecs:
+        if not isinstance(codec, ArrayArrayCodec):
+            raise TypeError(f"Filter {type(codec)} is not an ArrayArrayCodec")
```
dstansby: If filters aren't ArrayArray codecs, instead of erroring here would it be possible to include them somehow in the new metadata?

dstansby (commented):
I have left some comments - not finished reviewing yet, but I have to bounce for a few hours now and thought having some to think about/work on might be helpful in the meantime. Will carry on reviewing later today!

src/zarr/core/metadata/converter/converter_v2_v3.py
dstansby: Sorry, I just read the comment thread, feel free to ignore this and leave clear here!

dstansby:
In addition to my 'hit-and-run' comments above, I've now read the conversation ☝️ and had a think about the design/CLI API. My thoughts:

- I like using `migrate` instead of `convert`, in order to leave `convert` free for potentially other stuff. Another option to throw into the pot is `upgrade`?
- Whether `convert`/`migrate`/`upgrade`, I prefer leaving the version out of the top level command, ie. doing `zarr migrate v3` and not `zarr migrate-v3`. This leaves open the oppurtunity to do other invocations like `zarr migrate v4`, `zarr migrate v2`. I'm not saying these are a good or bad idea or will happen, but I think best not to close the option off for now.
- For migrating in-place, I think we could drop an explicit `--in-place` option? Then the call would be `zarr migrate v3 /path/to/arr.zarr`. If we're not deleting any data and just adding metadata documents, I see no reason to make the user manually specify `--in-place`
- Perhaps `clear` could be called `remove-metadata`, with a call like `zarr remove-metadata v2 /path/to/arr.zarr`?
- And perhaps `clear` could also (at least by default) first check for valid v3 metadata before deleting v2 metadata? This would ensure that you're always left with a valid array/group and don't accidentally delete all the metadata. A `--force` option or similar could override this.

K-Meech:
Thanks for the comments @dstansby! I'll start making the changes you suggested in the review above.

As for the overall structure - how about:

`zarr migrate v3` command:
- Simplest call is either `zarr migrate v3 path/to/input.zarr path/to/output.zarr` (add metadata at separate location) or `zarr migrate v3 path/to/input.zarr` (in place)
- The following extra options are available:
  - `--overwrite`: allows overwrite of existing v3 metadata at the output location
  -  `--remove-v2`: removes any v2 metadata at the output location after conversion
  -  `--dry-run`: log changes that would be made, without making them
  - Previously I'd supported a `path` option which, similar to [`zarr.open`](https://zarr.readthedocs.io/en/stable/api/zarr/index.html#zarr.open), specified the path within the store to open. Now that there's possibly separate input / output locations - this could be expanded to `--input-path` and `--output-path`.
  - The underlying functions e.g. [convert_v2_v3](https://github.com/K-Meech/zarr-python/blob/km/v2-v3-conversion/src/zarr/core/metadata/converter/converter_v2_v3.py#L37) also support `storage_options` similar to [`zarr.open`](https://zarr.readthedocs.io/en/stable/api/zarr/index.html#zarr.open) - should this be supported in the CLI too?

`zarr remove-metadata` command:
- Simplest call is `zarr remove-metadata v3 /path/to/.zarr` or `zarr remove-metadata v2 /path/to/.zarr`
- Options are:
  - `--path`: path within the store to open
  - `--dry-run`: log files that would be deleted, without removing them
  - `--force`: allows deletion of metadata, when there is no valid alternative e.g. can delete v3 metadata, when no valid v2 metadata OR delete v2 metadata when there is no valid v3 metadata

dstansby:
That broadly sounds great to me 👍 A couple of questions/suggestions:

> `--overwrite`: allows overwrite of existing v3 metadata at the output location

What's the use case for this? 

> `--remove-v2`: removes any v2 metadata at the output location after conversion

Perhaps `--remove-v2-meta` or `--remove-v2-metadata` to make it clear no array data is being deleted? That does result in a longer option to type out though, so I'm not sure which I prefer.

re. stores and paths, in the Python API I think there's a distinction because one might want to open one store and access multiple paths in one session. On the command line you only ever pass one store/path combination at a time on the command line, so leaving the store/path as a single option/string makes most sense to me instead of splitting them up with a `--path` option.

re. storage options, I guess in some cases they're required (e.g., see the FTP example in the [fsspec docs](https://filesystem-spec.readthedocs.io/en/latest/usage.html#instantiate-a-file-system)). Finding a consistent way to map from a command line string to a Python dict containing keword arguments that are then passed to `fsspec` might be a bit tricky though, so maybe worth puting down the line as an extra feature to add in a follow up PR? If there's an obvious and easy answer though it could be done here.

K-Meech:
- `--overwrite` is mainly provided for the situation where a conversion fails part way through, and you want to run it again. For example, say you were converting a large remote zarr and your connection failed part way through. You may have `zarr.json` in some locations in your hierarchy, but not others, which would need to overwritten by a new conversion call. (You could also call `remove-metadata` first, then `migrate`, but this is a shortcut for that)

- Happy to change to `--remove-v2-metadata` to make this clearer.

- Ok - I can remove the `path` option, and people will need to provide the full path to the array/group they want to convert in the input/output store part.

- Happy to keep `storage_options` for a future PR, as it isn't super-simple to allow an arbitrary Python dict of options

K-Meech:
I've updated the structure of the CLI - hopefully this addresses both @dstansby and @d-v-b 's comments! You should be able to test with:
```
zarr --help
zarr migrate --help
zarr remove-metadata --help
```
I haven't addressed [this comment yet](https://github.com/zarr-developers/zarr-python/pull/3257#discussion_r2221835719), but will do so in the next round of changes (I know @dstansby had some additional comments to make on the converter implementation).

One known issue:
- If you run `zarr migrate --dry-run` with a non-existent location e.g. `zarr migrate v3 data/images/example-1.zarr --dry-run`, it will create an empty directory at that location. This is related to `r+` opens - reported on this issue: https://github.com/zarr-developers/zarr-python/issues/3295

dstansby:
Very nice! I had a play with it locally and it worked well. I'll do a fuller review of the code now.

src/zarr/core/metadata/converter/cli.py
```diff
@@ -0,0 +1,180 @@
...
+
+import zarr.core.metadata.converter.migrate_to_v3 as migrate_metadata
+from zarr.core.sync import sync
+
+app = typer.Typer()
+
+logger = logging.getLogger(__name__)
+
+
+def _set_logging_config(verbose: bool) -> None:
```
dstansby:
```suggestion
def _set_logging_config(*, verbose: bool) -> None:
```

src/zarr/core/metadata/converter/cli.py
```diff
@@ -0,0 +1,180 @@
...
+    if verbose:
+        lvl = logging.INFO
+    else:
+        lvl = logging.WARNING
+    fmt = "%(message)s"
+    logging.basicConfig(level=lvl, format=fmt)
```
dstansby:
```suggestion
    logger.basicConfig(level=lvl, format=fmt)
```
I *think* you want to configure the logger instance, not global settings?

src/zarr/core/metadata/converter/cli.py
```diff
@@ -0,0 +1,180 @@
...
+
+
+def _set_verbose_level() -> None:
+    logging.getLogger().setLevel(logging.INFO)
```
dstansby:
```suggestion
    logger.setLevel(logging.INFO)
```
Same reason as above

src/zarr/core/metadata/converter/cli.py
```diff
@@ -0,0 +1,180 @@
...
+                "Input Zarr to migrate - should be a store, path to directory in file system or name of zip file "
+                "e.g. 'data/example-1.zarr', 's3://example-bucket/example'..."
+            )
+        ),
+    ],
+    output_store: Annotated[
+        str | None,
+        typer.Argument(
+            help=(
+                "Output location to write generated metadata (no chunks will be copied). If not provided, "
```
dstansby:
```suggestion
                "Output location to write generated metadata (no array data will be copied). If not provided, "
```

src/zarr/core/metadata/converter/cli.py
```diff
@@ -0,0 +1,180 @@
...
+    dry_run: Annotated[
+        bool,
+        typer.Option(
+            help="Enable a dry-run: files that would be converted are logged, but no new files are actually created."
+        ),
+    ] = False,
+    overwrite: Annotated[
+        bool,
+        typer.Option(
+            help="Remove any existing v3 metadata at the output location, before migration starts."
```
dstansby: In the case where there's no valid v2 metadata, is it possible that this command could remove valid v3 metadata, and therefore result in a state where there's no valid metadata at all?

src/zarr/core/metadata/converter/cli.py
```diff
@@ -0,0 +1,180 @@
...
+        ),
+    ] = False,
+    force: Annotated[
```
dstansby: I don't understand why this option exists - if there's no valid v2 metadata, why would I want to remove v3 metadata and therefore end up with no (valid) metdata?

src/zarr/core/metadata/converter/cli.py
```diff
@@ -0,0 +1,180 @@
...
+            help=(
+                "Only used when --overwrite is given. Allows v3 metadata to be removed when no valid "
+                "v2 metadata exists at the output location."
+            )
+        ),
+    ] = False,
+    remove_v2_metadata: Annotated[
+        bool,
+        typer.Option(
+            help="Remove v2 metadata (if any) from the output location, after migration is complete."
```
dstansby:
```suggestion
            help="Remove v2 metadata from the output location, after migration is complete."
```
If there isn't any v2 metdata, won't the migration just fail?

src/zarr/core/metadata/converter/cli.py
```diff
@@ -0,0 +1,180 @@
...
+        ),
+    ] = False,
+) -> None:
+    """Migrate all v2 metadata in a zarr hierarchy to v3. This will create a zarr.json file for each level
+    (every group / array). V2 files (.zarray, .zattrs etc.) will be left as-is.
```
dstansby:
```suggestion
    (every group / array). v2 files (.zarray, .zattrs etc.) will be left as-is.
```

src/zarr/core/metadata/converter/cli.py
```diff
@@ -0,0 +1,180 @@
...
+    """
+    if dry_run:
+        _set_verbose_level()
+        logger.info(
+            "Dry run enabled - no new files will be created. Log of files that would be created on a real run:"
```
dstansby:
```suggestion
            "Dry run enabled - no new files will be created or changed. Log of files that would be created on a real run:"
```
to be clear existing files won't be deleted or edited

src/zarr/core/metadata/converter/cli.py
```diff
@@ -0,0 +1,180 @@
...
+        )
+
+    write_store = output_store if output_store is not None else input_store
```
dstansby:
```suggestion
    write_store = input_store if output_store is None else output_store
```
For some reason I can't explain, this way round makes more sense in my head. But maybe not in yours? Feel free to take or leave!

src/zarr/core/metadata/converter/cli.py
```diff
@@ -0,0 +1,180 @@
...
+            help=(
+                "Allow metadata to be deleted when no valid alternative exists e.g. allow deletion of v2 metadata, "
+                "when no v3 metadata is present."
+            )
+        ),
+    ] = False,
+    dry_run: Annotated[
+        bool,
+        typer.Option(
+            help="Enable a dry-run: files that would be deleted are logged, but no files are actually removed."
```
dstansby:
```suggestion
            help="Enable a dry-run: files that would be deleted are logged, but no files are removed or changed."
```

src/zarr/core/metadata/converter/cli.py
```diff
@@ -0,0 +1,180 @@
...
+            zarr_format=cast(Literal[2, 3], int(zarr_format[1:])),
+            force=force,
+            dry_run=dry_run,
+        )
+    )
+
+
+@app.callback()  # type: ignore[misc]
+def main(
+    verbose: Annotated[
```
dstansby: I found it a bit annoying that I couldn't do `zarr convert ... --verbose`, and had instead to do `zarr --verbose convert ...`. Perhaps this is a minor gripe though, since it's quite nice having the `--verbose` option defined for all `zarr` subcommands automatically.

src/zarr/core/metadata/converter/migrate_to_v3.py
```diff
@@ -0,0 +1,303 @@
...
+from zarr.core.sync import sync
+from zarr.registry import get_codec_class
+from zarr.storage import StoreLike
+from zarr.storage._common import StorePath, make_store_path
+
+logger = logging.getLogger(__name__)
+
+
+def migrate_v2_to_v3(
+    input_store: StoreLike,
```
dstansby:
```suggestion
    *, 
    input_store: StoreLike,
```
Going keyword only here should avoid accidentally hading the input/ouput stores the wrong way round.

src/zarr/core/metadata/converter/migrate_to_v3.py
```diff
@@ -0,0 +1,303 @@
...
+    output_store: StoreLike | None = None,
+    storage_options: dict[str, Any] | None = None,
```
dstansby: To save having to deal with `storage_options` here, I would instead just accept an actual `Store` as input to this funciton. Doing this would also avoid having to deal with the case where one wants different `storage_options` for both input and output stores.

src/zarr/core/metadata/converter/migrate_to_v3.py
```diff
@@ -0,0 +1,303 @@
...
+        chunk_key_encoding=chunk_key_encoding,
+        fill_value=metadata_v2.fill_value,
+        codecs=codecs,
+        attributes=metadata_v2.attributes,
+        dimension_names=None,
+        storage_transformers=None,
+    )
+
+
+def _convert_filters(metadata_v2: ArrayV2Metadata) -> list[ArrayArrayCodec]:
```
dstansby: I think it would make slightly more sense to pass filters, instead of the whole metadata here.
```suggestion
def _convert_filters(filters) -> list[ArrayArrayCodec]:
```
(I probably got the typing wrong)

src/zarr/core/metadata/converter/migrate_to_v3.py
```diff
@@ -0,0 +1,303 @@
...
+
+    filters_codecs = [_find_numcodecs_zarr3(filter) for filter in metadata_v2.filters]
+    for codec in filters_codecs:
+        if not isinstance(codec, ArrayArrayCodec):
+            raise TypeError(f"Filter {type(codec)} is not an ArrayArrayCodec")
+
+    return cast(list[ArrayArrayCodec], filters_codecs)
+
+
+def _convert_compressor(metadata_v2: ArrayV2Metadata) -> BytesBytesCodec | None:
```
dstansby:
```suggestion
def _convert_compressor(compressor, dtype) -> BytesBytesCodec | None:
```
Again, I think makes for a clearer API if you explicitly just pass the compressor and type instead of the whole metadata.

src/zarr/core/metadata/converter/migrate_to_v3.py
```diff
@@ -0,0 +1,303 @@
...
+    if metadata_v2.compressor is None:
+        return None
+
+    compressor_name = metadata_v2.compressor.codec_id
+
+    match compressor_name:
```
dstansby:
```suggestion
    match metadata_v2.compressor.codec_id:
```
Since you don't use the variable again

TomNicholas:
> If you have the patience for it, please keep track of the stuff we could change in the core library to make this kind of tool easier to write, and feel free to open issues to track these features.

For example, is there currently a way to use zarr-python to convert v2 to v3 metadata without the CLI? If so then that would be independently useful (e.g. in VirtualiZarr). If not then in an ideal world that would exist before a CLI is added to wrap it.

emmanuelmathot:
Just FYI, I made this [converter](https://github.com/EOPF-Explorer/data-model) for ESA EOPF datasets (Zarr v2) to EOPF GeoZarr (V3). I adopted a recursive approach to copy and keep the datatree as well as the consolidated metadata.

K-Meech:
@TomNicholas - this PR will add both a CLI + functions to convert v2 to v3 metadata. Everything possible with the CLI is also available via direct function calls. 

Although, at the moment these are added to `src/zarr/core/metadata` which I don't think is part of the public API - @d-v-b / @dstansby, any suggestions for where these could be moved to make them easy to use by downstream scripts / packages?

TomNicholas:
@K-Meech thanks. That's actually okay for my use case - I'm already importing other private metadata-related internals. But I agree a later PR to make these public would be great.

dstansby:
I'd suggest putting the metadata converter python API in a new `zarr.metadata` submodule (which can contain other bits of metadata migrated from `zarr.core.metadata` later), and hiding the the CLI wrapper in a new private `zarr._cli` submodule (or `zarr._cli.py` file)

src/zarr/core/metadata/converter/cli.py
K-Meech: I could do something like:
```
if verbose:
        logger.setLevel(logging.INFO)
    else:
        logger.setLevel(logging.WARNING)
logger.addHandler(logging.StreamHandler())
```
but the issue is this will only affect logs directly from the `cli.py` file. When using `--dry-run`, I also want to see the log of created / deleted files from `migrate_to_v3.py`, which won't be shown with this setting alone.

I could add a similar setup to the `migrate_to_v3.py` file, but this could be annoying for downstream code that uses these functions and wants to use a different logging level / setup e.g. to a file rather than the console. 

It could work if I created the `migrate_to_v3.py` logger as an explicit child of the `cli.py` logger though i.e.
```
# in cli.py
logger = logging.getLogger("cli")

# in migrate_to_v3.py
logger = logging.getLogger("cli.migrate")
```
What do you think?

src/zarr/core/metadata/converter/cli.py
K-Meech: The `--overwrite` option checks that there is valid alternative metadata before deleting. You could only create a state with no valid metadata by adding `--force` in addition.

src/zarr/core/metadata/converter/cli.py
K-Meech: There are two main scenarios for using `--overwrite` and `--force`:
- writing 'in place' i.e. `zarr migrate v3 path/to/input-zarr.zarr`. In this scenario, there is `v2` metadata present at `input-zarr.zarr` and there may be some v3 metadata there too from a previously failed conversion attempt. `--overwrite` will remove the v3 metadata and allow `migrate` to be run again.

- writing to a separate location i.e. `zarr migrate v3 path/to/input-zarr.zarr path/to/output-zarr.zarr`. In this scenario, there is `v2` metadata at `input-zarr.zarr` and potentially some partial `v3` metadata at `output-zarr.zarr`. `--overwrite` alone can't remove this, as there is no valid v2 metadata at the output location, so this is where `--force` would be needed to allow `migrate` to run again.

I agree these applications aren't super clear from these options as-is though 😅

src/zarr/core/metadata/converter/cli.py
K-Meech: This is another case for writing to a separate output location - if you run e.g. `zarr migrate v3 path/to/input-zarr.zarr path/to/output-zarr.zarr`, then there will be no v2 metadata at the output location (`output-zarr.zarr`) - hence the 'if any' qualifier.

src/zarr/core/metadata/converter/cli.py
K-Meech: I agree it's annoying to put `--verbose` before the actual command. I couldn't see a simple way to automatically add `--verbose` to all sub-commands in `typer` though. Happy to manually add this to the sub-commands - if you think it's worth it?

src/zarr/core/metadata/converter/converter_v2_v3.py
K-Meech: I think this would be possible - but do you have an example of a scenario where this would be needed? In the [numcodecs docs](https://numcodecs.readthedocs.io/en/stable/zarr3.html#filters-array-to-array-codecs), all filters seem to be listed under `ArrayArrayCodecs`. Are there codecs of a different type that are usually used in the v2 `filters` slot?

K-Meech:
I've made most of the requested changes to the implementation now. Any remaining I've responded to in the review thread above - @dstansby let me know if you have any suggestions for these, or for the refactored implementation / tests.

While changing the migration functions to accept `Store` directly (rather than `StoreLike`), I made some changes in `zarr/storage/_common.py`. This  refactored `make_store_path`, to add a new `make_store` function. I also updated the docstrings - the existing `make_store_path` one was a bit out of date (e.g. didn't mention `FSMap`).

Also, I noticed that I don't have any handling for consolidated metadata (`.zmetadata`) files at the moment. Would it be fine to just error if this is encountered, or should conversion be included?

dstansby:
> While changing the migration functions to accept Store directly (rather than StoreLike), I made some changes in zarr/storage/_common.py. This refactored make_store_path, to add a new make_store function. I also updated the docstrings - the existing make_store_path one was a bit out of date (e.g. didn't mention FSMap).

Nice! Unfortunately, in pursuit of fixing https://github.com/zarr-developers/zarr-python/issues/3295 I also did a refactor at https://github.com/zarr-developers/zarr-python/pull/3308 - I think we should probably merge my PR first (sorry!) and then rebase this one later, since this is a feature and my refactor is a pathway to fixing a bug.

> Also, I noticed that I don't have any handling for consolidated metadata (.zmetadata) files at the moment. Would it be fine to just error if this is encountered, or should conversion be included?

I think it's fine to gracefully error on consolidated metadata for now, and add support as a follow up feature in a future PR.

K-Meech:
Thanks for the info @dstansby ! I see that https://github.com/zarr-developers/zarr-python/pull/3308 has been merged now, so I'll go ahead and fix the merge conflicts with this branch.

K-Meech:
All conflicts are now fixed. I also added a line to stop conversion if consolidated metadata is detected.

src/zarr/core/metadata/converter/cli.py
dstansby: Thinking a bit, I think there are two options here:

1. Have a CLI logger and a metadata_v3.py logger (as currently implemented), and have the CLI logger configuration change functions change configuration in both logger objects.
2. Move to having a single logger for zarr-python as a whole, and just configure that appropriately when calling the CLI.

2) is a simpler option, especially when it comes to documenting to users how to configure zarr logging. It's also the option I've seen used in other projects (e.g., [matplotlib](https://matplotlib.org/stable/api/matplotlib_configuration_api.html#logging) and [sunpy](https://docs.sunpy.org/en/stable/topic_guide/logger.html)).

Can you think of any use cases to have multiple logs, as in 1) above? If there are compelling use cases then maybe we should go with multiple logging objects; otherwise, I'd be up for just going for one for the whole of zarr-python.

src/zarr/core/metadata/converter/cli.py
dstansby: That makes sense 👍

src/zarr/core/metadata/converter/cli.py
dstansby: 👍

src/zarr/core/metadata/converter/cli.py
dstansby: Hmm, I'm not sure. Unless anyone else weighs in with a strong opinion, we can leave it as is I think. It doesn't prevent someone coming along and adding the option for `--verbose` to sub commands later.

src/zarr/storage/_common.py
```diff
@@ -339,49 +322,106 @@ async def make_store_path(
 
     if isinstance(store_like, StorePath):
         # Already a StorePath
```
dstansby:
```suggestion
        # Get underlying store
```

tests/test_cli/test_migrate_v3.py
```diff
@@ -0,0 +1,632 @@
...
+    assert metadata.chunk_key_encoding == V2ChunkKeyEncoding(separator=".")
+    assert metadata.data_type == UInt16(endianness="little")
+    assert metadata.codecs == (
+        BytesCodec(endian="little"),
+        BloscCodec(typesize=2, cname="zstd", clevel=3, shuffle="shuffle", blocksize=0),
+    )
+    assert metadata.fill_value == fill_value
+    assert metadata.attributes == attributes
+    assert metadata.dimension_names is None
+    assert metadata.storage_transformers == ()
```
dstansby: Instead of asserting every attribute on metadata, it might be nicer to do `assert metadata == MetadataObject(...)`. That way if multiple attributes are wrong, they will all show up in the pytest error, as opposed to with multiple asserts pytest will just stop on the first failure.

dstansby (commented):
Looking great! I think outstanding todos here are:

- Add a release note entry
- Add a new section to the [user guide docs ](https://zarr.readthedocs.io/en/stable/user-guide/index.html) to advertise the new CLI
- Work out how to do logging (see inline thread discussion)

d-v-b:
as a general comment about converting v2 -> v3, converting the v2 `filters, compressor` to teh v3 `codecs` is hard. I suspect a general solution will not really be possible in this PR, because it relies on features that don't exist in Zarr Python yet. I am working on this (see #3162,  https://github.com/zarr-developers/zarr-python/pull/3276), but it's a big lift. Getting PRs like #3318 merged is an important part of this process. But once we do this these features in, we can generically convert filters + compressors to codecs.

src/zarr/core/metadata/converter/cli.py
K-Meech: Thanks @dstansby - I agree using a single logger is the simpler option. I've implemented a solution very similar to the [matplotlib setup](https://github.com/matplotlib/matplotlib/blob/2b0f3dc833aea9e9c9c3c9a31c67531c9dbdd52c/lib/matplotlib/__init__.py#L284) which seems to be working well with the CLI.

K-Meech:
Thanks @d-v-b - at the moment this PR:
- explicitly converts v2 `numcodecs` blosc / zstd / gzip to the v3 `zarr.codecs` equivalent.
- For any other v2 `numcodecs` codec, it tries to find a matching `numcodecs.zarr3` codec (by name), then initialises it with same settings from `.get_config`. This works fine for some codecs, but does cause issues for others e.g.  https://github.com/zarr-developers/zarr-python/issues/3256 .
 
Happy to wait on merging this PR until the other issues / PRs you mentioned are resolved.

d-v-b:
@K-Meech that approach seems good, and I don't this effort should be blocked by by codec changes in the background.

K-Meech:
@dstansby I think I've addressed all of your new comments now + I added release notes and a user guide docs page. There's still [a comment about filters](https://github.com/zarr-developers/zarr-python/pull/3257#discussion_r2221835719) from a while ago - any thoughts on that one? 

Also, let me know if you have any comments on the new changes - I had to make some small modifications to handle conflicts with the latest changes to the `main` branch.

src/zarr/core/metadata/converter/converter_v2_v3.py
d-v-b: there is no guarantee that every element of `filters` will be arrayarray codecs. 

`"filters": [{"id": "gzip", "level": 1}], "compressor": None"` and `"filters": None, "compressor": {"id": "gzip", "level": 1}` should be considered identical

d-v-b: for more background, in the zarr v2 format, all the elements of `filters`, and the value assigned to `compressor`, are the same kind of thing, from the metadata POV. 

the zarr v3 format introduced the idea of an "array array codec", "array bytes codec", "bytes bytes codec", but there was no effort to guarantee that all v2 metadata documents could be easily translated into the v3 form.

K-Meech: Thanks for the clarification @d-v-b ! At the moment I construct the new codecs list (in order from first to last) as:
- `TransposeCodec` (if the v2 order was F)
- converted filter codecs (currently errors if not `ArrayArrayCodec`)
- `BytesCodec` (with 'endianness' based on the v2 dtype)
- converted compressor codec (currently errors if not `BytesBytesCodec`)

I can change this to pool the filters / compressors together and handle them in the same way. I guess the order is still important though? Should it still be: `TransposeCodec` > filter codecs > `BytesCodec` > compressor codec?

d-v-b: > I can change this to pool the filters / compressors together and handle them in the same way. I guess the order is still important though? Should it still be: TransposeCodec > filter codecs > BytesCodec > compressor codec? 

This makes sense, and yes the order is still important, but what you have looks good.

An important case to consider is arrays that use vlen-utf8 or vlen-bytes codecs for compressing variable-length strings or variable-length bytes. For these arrays, we don't need a `BytesCodec`, because the vlen-utf8 and vlen-bytes codecs do the same job as the `bytescodec` (i.e., they are effectively `ArrayBytesCodecs`). But I think we could add support for those arrays later, if you want to ensure that this PR gets done sooner.

d-v-b: yet another situation where it would be great if we had a big collection of various zarr arrays on github, each using different compression routines, for testing this kind of thing

K-Meech: Thanks @d-v-b . Ok - I'll keep the order the same, but remove the exceptions for `ArrayArrayCodec` / `BytesBytesCodec` for filters and compressors (allowing them to be of any type). I'll also process them both in the same way, so they're treated as identical.

For the `vlen` codecs, I think the current code will error saying that there is no `numcodecs.zarr3` equivalent. My preference would probably be to add support for these in a future PR - as this is already quite large!

d-v-b: > For the `vlen` codecs, I think the current code will error saying that there is no `numcodecs.zarr3` equivalent. My preference would probably be to add support for these in a future PR - as this is already quite large!

that works! We can add this functionality later on.

K-Meech: @d-v-b I've made these changes locally, but the order of codecs seems to cause issues. 

For example, I created a v2 array with: 
- `"filters": [{"id": "zstd", "level": 3}], "compressor": None"`
Then converted it to v3: 
- `codecs` are `ZstdCodec` followed by `BytesCodec`. 

When I try to open the v3 array I get: `TypeError: Invalid codec order. ArrayBytes codec BytesCodec(endian=<Endian.little: 'little'>) must be preceded by an ArrayArrayCodec. Got <class 'zarr.codecs.zstd.ZstdCodec'> instead.` 

Opening the v2 array directly works - so I guess it is doing the equivalent of `BytesCodec` first then `ZstdCodec` after (rather than enforcing filters > BytesCodec > compressors). Should I dynamically put the `BytesCodec` in a different position based on the types of filters and compressors?

d-v-b: > * `"filters": [{"id": "zstd", "level": 3}], "compressor": None"`
>   Then converted it to v3:
> 
> * `codecs` are `ZstdCodec` followed by `BytesCodec`.

the bytes codec has to come _before_ any `BytesBytesCodec` instances, so you can only place the `BytesCodec` in the sequence of codecs only _after_ you have assigned a type to each codec.

K-Meech: Ok thanks - I'm away for the next week, but I'll give this another go when I'm back.
I'll convert all of the filters and compressors, putting them into a list in order (filters followed by compressors). At this stage they'll all have types of `ArrayArrayCodec`, `ArrayBytesCodec`, or `BytesBytesCodec`. Then I can find the first `BytesBytesCodec` and place the `BytesCodec` in the slot before this.

Seems like there may be a lot of edge cases here though - e.g. [`numcodecs.zarr3`](https://numcodecs.readthedocs.io/en/stable/zarr3.html#serializers-array-to-bytes-codecs) has some `ArrayBytesCodec` listed - so in this case I guess we wouldn't need to insert the `BytesCodec` at all. I'm curious how this works with opening an array with v2 metadata directly - is there a function somewhere in the codebase that does this logic for figuring out when/if this should happen in the filters/compressors sequence?

d-v-b: > Seems like there may be a lot of edge cases here though

yes, here's what handling those edge cases looks like: https://github.com/d-v-b/zarr-python/blob/9055a1aaff563867c11eb5ef50c1fb89767a06aa/src/zarr/core/codec_pipeline.py#L489-L701. If we get this added to zarr python, then your CLI can use it. Until then, don't worry about handling everything.

> I'm curious how this works with opening an array with v2 metadata directly - is there a function somewhere in the codebase that does this logic for figuring out when/if this should happen in the filters/compressors sequence?

We cheat and put all of the filters and compressor inside a [single array-bytes codec](https://github.com/zarr-developers/zarr-python/blob/926a52fa11845a142f65b10f13c1d9a92c754e6b/src/zarr/codecs/_v2.py#L22) :)

src/zarr/_cli/cli.py
```diff
@@ -0,0 +1,186 @@
...
+import typer
+
+import zarr
+import zarr.metadata.migrate_v3 as migrate_metadata
+from zarr.core.sync import sync
+from zarr.storage._common import make_store
+
+app = typer.Typer()
+
+logger = logging.getLogger(__name__)
```
dstansby, line 14: Is it deliberate that this is a new logger, instead of importing the logger object from `zarr`? I don't tihnk it matters too much, but re-using `zarr._logger` might save some code duplication because you could remove functions from this file for configuring the logger.

src/zarr/metadata/migrate_v3.py
```diff
@@ -0,0 +1,294 @@
...
+    metadata_files_required = {2: [ZARRAY_JSON, ZGROUP_JSON], 3: [ZARR_JSON]}
+
+    for metadata_file in metadata_files_required[zarr_format]:
+        if await (store_path / metadata_file).exists():
+            return True
+
+    return False
+
+
+def _convert_array_metadata(metadata_v2: ArrayV2Metadata) -> ArrayV3Metadata:
```
d-v-b, line 186: later on we can put this in `zarr.core.metadata`, e.g. in a `conversion` module

d-v-b:
a high-level comment about the conversion strategy. this is not a blocker for this PR, but the current conversion strategy is serial over arrays and groups. This means converting an array / group waits until the previous one is converted, and it increases the likelihood of conversion breaking in the middle, leaving a hybrid v2 / v3 hierarchy.

We could also do conversion with a concurrent API based on concurrently reading the entire input hierarchy with `dict(group.members(max_depth=None))`, preparing new metadata documents for each member, then concurrently writing those documents with [`create_hierarchy`](https://github.com/zarr-developers/zarr-python/blob/a26926cf4ec51a12ca28ff31c13ef2c8a32d51f9/src/zarr/core/sync_group.py#L51). 

This could be considered merely a performance concern, but I suspect reading and writing the hierarchy concurrently might be the difference between success and failure for this CLI when people use it on high-latency storage backends.

@K-Meech if you want to experience how your CLI tool handles IO latency, we have a [store wrapper class](https://github.com/zarr-developers/zarr-python/blob/a26926cf4ec51a12ca28ff31c13ef2c8a32d51f9/src/zarr/testing/store.py#L530) that adds latency to any other store class. Try running your conversion on a large zarr hierarchy with 100ms of latency added to `get` and `set` on a local store. If it takes more than a few seconds, then this strongly argues for moving away from the serial conversion strategy and using something concurrent.

src/zarr/_cli/cli.py
K-Meech: Yes this was intentional - most of the other libraries I've seen have their root logger (i.e. `zarr._logger`), then each individual file still calls `logger = logging.getLogger(__name__)` to create a child of that logger. 

I could use `zarr._logger` here, but there are already other files in `zarr` that use `logger = logging.getLogger(__name__)` - so I thought this would be more consistent. Also, I'm not sure changing this would remove code in this file? `_set_logging_level` uses `verbose` to determine if the logging level should be INFO or WARNING, then calls `zarr.set_log_level` and `zarr.set_format` to set this on `zarr._logger`. So it's already using settings on the overall logger?

K-Meech:
Thanks @d-v-b . I tried converting a Zarr with 30 nested groups (each with a small array inside) and this took about 10 seconds total (with a `LatencyStore` set up as `LatencyStore(local_store, get_latency=0.1, set_latency=0.1)`). I've [created a gist](https://gist.github.com/K-Meech/7c0f34fb6de4be86e9f394a33e401e94) of this small test, if you wanted to try it out.

Happy to look at changing this to use concurrent writes with `create_hierachy` - perhaps best to do in a separate PR after this one is merged? What do you think?

I've been keeping a list of potential follow up changes mentioned on this PR - happy to convert these into issues after merge:
- Handle conversion of consolidated metadata (`.zmetadata`) files
- Handle any combination of `numcodecs` codecs - [see discussion in thread](https://github.com/zarr-developers/zarr-python/pull/3257#discussion_r2221835719). Currently the code only handles filters that are `ArrayArrayCodecs` and compressors that are `BytesBytesCodecs`.
- Use concurrent writes, rather than serial

d-v-b:
> Thanks @d-v-b . I tried converting a Zarr with 30 nested groups (each with a small array inside) and this took about 10 seconds total (with a `LatencyStore` set up as `LatencyStore(local_store, get_latency=0.1, set_latency=0.1)`). I've [created a gist](https://gist.github.com/K-Meech/7c0f34fb6de4be86e9f394a33e401e94) of this small test, if you wanted to try it out.

cool, thanks for running this experiment! IMO 10 seconds for converting 30 arrays and 30 groups is fine to start with. We can treat the performance tuning as an implementation detail for a later PR.

K-Meech:
I've fixed the merge conflicts with this branch, and updated docstrings to reference the new `zarr.codecs.numcodecs` (rather than `numcodecs.zarr3`). I think I've addressed everyone's comments as well - @dstansby , see my replies to your comments above.
Do let me know if anything else needs changing / updating here - otherwise it might be ready to go in?

d-v-b:
@zarr-developers/python-core-devs I would like to merge this soon (24 hours) so now is a great time to look over this PR!

d-v-b:
it's going in once the tests are green. Thanks so much @K-Meech! If you don't mind, I'd like to make a post about your work over at image.sc, so people know about this functionality. This might be a good way to get some early feedback from users.

K-Meech:
Great - thanks @d-v-b ! I'll make some follow up issues for [additional features from the reviews on this PR](https://github.com/zarr-developers/zarr-python/pull/3257#issuecomment-3196157284).

joshmoore:
👏🏽 👏🏽 - The one thought I had while pondering this, @K-Meech, (and sorry if it was already discussed) is to what degree the CLI will follow the semver of the library itself. Not that anything needs doing but might be worth expectation management on the part of consumers.

d-v-b:
> 👏🏽 👏🏽 - The one thought I had while pondering this, @K-Meech, (and sorry if it was already discussed) is to what degree the CLI will follow the semver of the library itself. Not that anything needs doing but might be worth expectation management on the part of consumers.

the library doesn't follow semver -- we are using [effver](https://zarr.readthedocs.io/en/stable/developers/contributing.html#compatibility-and-versioning-policies)

joshmoore:
Fair enough, effver then. Can a user at this point reasonably expect the same level of stability from the CLI that they would for the API?

d-v-b:
I don't see why we should version the CLI any differently, other than the fact that it's pretty new and so it might need some polishing as people use it. To give ourselves some flexibility, we could consider adding a disclaimer to the CLI stating that it's still experimental and might experience breaking changes in future versions of zarr python.

d-v-b:
see https://github.com/zarr-developers/zarr-python/issues/3465

PR #3262, opened 2025-07-17
Add numcodecs to downstream projects to check before release

dstansby:
Zarr 3.1.0 broke numcodecs tests (see runs on https://github.com/zarr-developers/numcodecs/pull/762), so add `numcodecs` to the list of downstream packages to check before release in the future.


d-v-b:
numcodecs is both upstream and downstream. this is pretty problematic

dstansby:
pre-commit.ci autofix

PR #3264, opened 2025-07-17
improvements to `parse_dtype`

### LINKED ISSUE #3263: incomplete round-tripping of v3 data type json
`parse_data_type("variable_length_bytes", zarr_format=3)` fails because `parse_data_type` only attempts to resolve inputs as JSON if they are mappings. strings should also be supported.

d-v-b:
- Add a new function `parse_dtype`. `parse_data_type` is kept around but it just wraps `parse_dtype`. The reason for this   change is naming consistency -- the `ZDType` methods already use the "dtype" abbreviation extensively, so it's potentially confusing that `parse_data_type` does not.
- Handle strings and sequences as potential json-like inputs. Adds tests to ensure that the JSON form a of a dtype is a valid argument to `parse_dtype` (with the exception of "|O", which is ambiguous).

closes #3263


d-v-b:
cc @TomNicholas

d-v-b:
https://github.com/zarr-developers/zarr-python/pull/3264/commits/d684ada2b95f9d0b9fe2bb8c26a49cda31c432d3 adds a test to ensure that `parse_dtype` is the same as `parse_data_type`

src/zarr/core/dtype/__init__.py
```diff
@@ -188,39 +190,69 @@ def parse_data_type(
...
+    Returns
+    -------
+    ZDType[TBaseDType, TBaseScalar]
+        The ZDType corresponding to the input.
+
+    Examples
+    --------
+    >>> parse_dtype("int32", zarr_format=2)
+    Int32(endianness="little")
+    """
```
dstansby: I'd say bin the docstirng here to avoid duplication, and just point to `parse_dtype`.

src/zarr/core/dtype/__init__.py
```diff
@@ -188,39 +190,69 @@ def parse_data_type(
...
     --------
-    >>> from zarr.dtype import parse_data_type
-    >>> import numpy as np
-    >>> parse_data_type("int32", zarr_format=2)
-    Int32(endianness='little')
-    >>> parse_data_type(np.dtype('S10'), zarr_format=2)
-    NullTerminatedBytes(length=10)
-    >>> parse_data_type({"name": "numpy.datetime64", "configuration": {"unit": "s", "scale_factor": 10}}, zarr_format=3)
-    DateTime64(endianness='little', scale_factor=10, unit='s')
+    >>> parse_dtype("int32", zarr_format=2)
```
dstansby, line 256: Huh, does this not need an import? not an issue, just a question

src/zarr/dtype.py
```diff
@@ -84,4 +85,5 @@
     "data_type_registry",
     "data_type_registry",
     "parse_data_type",
```
dstansby:
```suggestion
```

dstansby (commented):
Nice - I like the name change. Having two identical functions in our API seems a bit confusing from a user POV: https://zarr--3264.org.readthedocs.build/en/3264/api/zarr/dtype/index.html#functions. Could you remove `parse_data_type` from `__all__` so it's removed from the docs, but will still be imported and work for backwards compatibility?

src/zarr/core/dtype/__init__.py
d-v-b: I think duplication is fine here. Happy to revisit this if maintaining two docstrings is a maintenance burden, but my preference is for each function to describe what it does, even if another function does the same thing.

src/zarr/core/dtype/__init__.py
d-v-b: I think it does need an import

d-v-b: fixed in https://github.com/zarr-developers/zarr-python/pull/3264/commits/c42edf6049b8c8b62d6ccf347ea4aeda0c76e7bf

d-v-b:
a notable recent change: I widened the accepted inputs even further, and now `parse_dtype` / `parse_data_type` will accept the output of `dtype.to_json(zarr_format=2)`, which allows you to unambiguously request a vlen string dtype for zarr v2 data via a dict like `{"name": "|O", "object_codec_id": "vlen-utf8"}`

src/zarr/core/dtype/__init__.py
dstansby: Do you know why the CI wasn't failing before the import was added? Perhaps something to fix in another PR.

src/zarr/core/dtype/__init__.py
```diff
@@ -188,43 +190,93 @@ def parse_data_type(
...
+        The ZDType corresponding to the input.
 
     Examples
     --------
     >>> from zarr.dtype import parse_data_type
     >>> import numpy as np
     >>> parse_data_type("int32", zarr_format=2)
     Int32(endianness='little')
-    >>> parse_data_type(np.dtype('S10'), zarr_format=2)
+    >>> parse_dtype(np.dtype('S10'), zarr_format=2)
```
dstansby:
```suggestion
    >>> parse_data_type(np.dtype('S10'), zarr_format=2)
```
Given this is the docstring for `parse_data_type` and the other examples in the block use `parse_data_type`.

src/zarr/core/dtype/__init__.py
```diff
@@ -188,43 +190,93 @@ def parse_data_type(
...
+    return parse_dtype(dtype_spec, zarr_format=zarr_format)
+
+
+def parse_dtype(
+    dtype_spec: ZDTypeLike,
+    *,
+    zarr_format: ZarrFormat,
+) -> ZDType[TBaseDType, TBaseScalar]:
+    """
+    Interpret the input as a ZDType.
```
dstansby:
```suggestion
    Convert the input as a ZDType.
```

src/zarr/core/dtype/__init__.py
```diff
@@ -188,43 +190,93 @@ def parse_data_type(
...
+
+    Parameters
+    ----------
+    dtype_spec : ZDTypeLike
+        The input to be interpreted as a ZDType. This could be a ZDType, which will be returned
```
dstansby:
```suggestion
        The input to be converted to a ZDType. This could be a ZDType, which will be returned
```

src/zarr/core/dtype/__init__.py
```diff
@@ -188,43 +190,93 @@ def parse_data_type(
...
+        directly, or a JSON representation of a ZDType, or a native dtype, or a python object that
```
dstansby:
```suggestion
        directly, or a JSON representation of a ZDType, or a numpy dtype, or a python object that
```
I think "native" isn't widely used terminology, whereas more people will understand what a "numpy dtype" is. If I misunderstand and "native" is different, I still think this should mention "numpy dtype" as an acceptable input.

src/zarr/core/dtype/__init__.py
```diff
@@ -188,43 +190,93 @@ def parse_data_type(
...
+        can be converted into a native dtype.
+    zarr_format : ZarrFormat
+        The Zarr format version. This parameter is required because this function will attempt to
```
dstansby, line 243: A stretch goal (for another issue/PR) - would it be possible to make this optional for certain inputs (e.g., converting 'int32' doesn't depend on the version of the zarr spec)

src/zarr/core/dtype/__init__.py
```diff
@@ -188,43 +190,93 @@ def parse_data_type(
...
+    >>> from zarr.dtype import parse_dtype
+    >>> import numpy as np
+    >>> parse_dtype("int32", zarr_format=2)
+    Int32(endianness='little')
+    >>> parse_dtype(np.dtype('S10'), zarr_format=2)
+    NullTerminatedBytes(length=10)
+    >>> parse_dtype({"name": "numpy.datetime64", "configuration": {"unit": "s", "scale_factor": 10}}, zarr_format=3)
+    DateTime64(endianness='little', scale_factor=10, unit='s')
+    >>> parse_dtype("int32", zarr_format=2)
+    Int32(endianness="little")
```
dstansby:
```suggestion
```
This example is already given at the top of the examples.

src/zarr/core/dtype/__init__.py
```diff
@@ -188,43 +190,93 @@ def parse_data_type(
...
     >>> import numpy as np
     >>> parse_data_type("int32", zarr_format=2)
     Int32(endianness='little')
-    >>> parse_data_type(np.dtype('S10'), zarr_format=2)
+    >>> parse_dtype(np.dtype('S10'), zarr_format=2)
     NullTerminatedBytes(length=10)
     >>> parse_data_type({"name": "numpy.datetime64", "configuration": {"unit": "s", "scale_factor": 10}}, zarr_format=3)
     DateTime64(endianness='little', scale_factor=10, unit='s')
+    >>> parse_data_type("int32", zarr_format=2)
+    Int32(endianness="little")
```
dstansby:
```suggestion
```
This example is already at the top of the doctests.

src/zarr/core/dtype/__init__.py
d-v-b: I would rather not have a function where one parameter is optional depending on the value of another parameter. That should probably be a different function entirely, e.g. one that specifically takes numpy dtype-like inputs and finds the corresponding zarr data type.

maxrjones:
I think more work here is needed for a gentle deprecation of `parse_data_type` in favor of `parse_dtype`. People using the [public function](https://zarr.readthedocs.io/en/stable/api/zarr/dtype/index.html#zarr.dtype.parse_data_type) will currently have a failure on the next release due to the change in import structure (e.g., virtualizarr - https://github.com/zarr-developers/VirtualiZarr/actions/runs/16510795789/job/46692311156?pr=736).

d-v-b:
good catch max, looks like we need to fix the import

PR #3266, opened 2025-07-17
Use `in` and `or` instead of multiple`if`'s

DimitriPapadopoulos:
TODO:


d-v-b (approved):

d-v-b (changes requested):
i think this is doomed because `data` is not guaranteed to be hashable

DimitriPapadopoulos:
Perhaps I can keep the additional `or` though.

d-v-b:
what about using a tuple instead of a set?

DimitriPapadopoulos:
Of course, you're right.

PR #3276, opened 2025-07-21
[do not merge] proof of concept for a unified v2 / v3 codecs 

d-v-b:
this PR is a proof of concept that demonstrates how we can augment our `Codec` API to support v2 or v3 codecs. 

Here are two motivating examples that run under this PR and fail in `main`:

## gzip compression

This example shows that we can use the exact same codec (`GZipCodec` OR `numcodecs.GZip`) with zarr v2 or zarr v3 
to create an array. This fails on main if you try to use `GZipCodec` with zarr v2, or `numcodecs.GZip`  with zarr v3, even though the underlying gzip compression is identical.

```python
# /// script
# requires-python = ">=3.11"
# dependencies = [
#   "zarr @ git+https://github.com/d-v-b/zarr-python.git@a2bc6555",
#   "pytest"
# ]
# ///
import json

import pytest
import numpy as np
import zarr
from zarr.codecs import GzipCodec
from numcodecs.gzip import GZip

@pytest.mark.parametrize("zarr_format", [2, 3])
@pytest.mark.parametrize('codec_cls', [GzipCodec, GZip])
def test_gzip_compression(zarr_format, codec_cls) -> None:
    store = {}
    z_w = zarr.create_array(
        store=store,
        dtype="int",
        shape=(1,),
        chunks=(10,),
        zarr_format=zarr_format,
        compressors=codec_cls(),
    )
    z_w[:] = 5

    z_r = zarr.open_array(store=store, zarr_format=zarr_format)
    assert np.all(z_r[:] == 5)

if __name__ == "__main__":
    pytest.main([__file__, f"-c {__file__}", "-s"])
```

## jpeg compression 

This example demonstrates using the `Jpeg` codec defined in `imagecodecs` as a compressor for a v2 array and a serializer for a v3 array. Data can be written and read back. 

- The codec comes in to `create_array` as an object that implements the `numcodecs.abc.Codec` API
- we internally wrap that object in a class that emulates the zarr-python `Codec` API.
- v2 and v3 JSON metadata is created by the same object, depending on the zarr format of the array
- when decoding, we check the numcodecs registry, but only if numcodecs is installed. this could be improved -- we could just register the codec directly in our own codec registry, and remove numcodecs entirely from this example.

```python
# /// script
# requires-python = ">=3.11"
# dependencies = [
#   "zarr @ git+https://github.com/d-v-b/zarr-python.git@a2bc6555",
#   "imagecodecs==2025.3.30",
#   "pytest"
# ]
# ///

#   "zarr @ git+https://github.com/zarr-developers/zarr-python.git@main",
from typing import Literal

import numcodecs
import numpy as np
import pytest
from imagecodecs.numcodecs import Jpeg

import zarr

numcodecs.register_codec(Jpeg)
jpg_codec = Jpeg()


@pytest.mark.parametrize("zarr_format", [2, 3])
def test(zarr_format: Literal[2, 3]) -> None:
    store = {}
    if zarr_format == 2:
        z_w = zarr.create_array(
            store=store,
            data=np.zeros((100, 100, 3), dtype=np.uint8),
            compressors=jpg_codec,
            zarr_format=zarr_format,
        )
    else:
        z_w = zarr.create_array(
            store=store,
            data=np.zeros((100, 100, 3), dtype=np.uint8),
            serializer=jpg_codec,
            zarr_format=zarr_format,
        )
    z_w[:] = 2
    z_r = zarr.open_array(store=store, zarr_format=zarr_format)
    assert np.all(z_r[:] == 2)
    if zarr_format == 2:
        print(z_r.metadata.to_dict()["compressor"])
    else:
        print(z_r.metadata.to_dict()["codecs"])


if __name__ == "__main__":
    pytest.main([__file__, f"-c {__file__}", "-s"])
```

## What's this requires

This functionality requires a set of changes that I would like to introduce in a series of PRs:
- Adding `to_json(zarr_format: Literal[2,3]` methods to all the codecs, so the same object can generate zarr v2 or zarr v3 metadata. [example](https://github.com/d-v-b/zarr-python/blob/a2bc6555976b57579dfcf605b44ac3abb80237c5/src/zarr/codecs/gzip.py#L70-L82).
- Adding `from_json(data, zarr_format: Literal[2,3])` methods to all the codecs, so the same object can be created from zarr v2 metadata or zarr v3 metadata. [example](https://github.com/d-v-b/zarr-python/blob/a2bc6555976b57579dfcf605b44ac3abb80237c5/src/zarr/codecs/gzip.py#L104-L114).
- Adding a protocol that models the `numcodecs,abc.Codec` API, so we can interact with numcodecs objects type-safely without a numcodecs dependency. [example](https://github.com/d-v-b/zarr-python/blob/a2bc6555976b57579dfcf605b44ac3abb80237c5/src/zarr/codecs/numcodec.py#L66-L103).
- Adding a `Codec` class that wraps a numcodec-like object and endows it with full codec powers. These numcodec-adapter objects can be cast to array-array codec, array-bytes-codec, or bytes-bytes codec as needed. [example](https://github.com/d-v-b/zarr-python/blob/a2bc6555976b57579dfcf605b44ac3abb80237c5/src/zarr/codecs/numcodec.py#L106-L200)
- Adding logic in the codec pipeline to handle codecs that are a mix of zarr python native codecs and numcodecs-adapter codecs. This ensures that we have a valid codec pipeline. [example](https://github.com/d-v-b/zarr-python/blob/a2bc6555976b57579dfcf605b44ac3abb80237c5/src/zarr/core/codec_pipeline.py#L489-L701).
... There might be more changes that I'm forgetting right now, but these are the big ones.

We also need checks to ensure that a codec claiming to be "gzip" generates gzip-compatible metadata. There are a few ways of doing this (inspect the metadata it generates, or replace it with the in-house codec with the same name), I haven't implemented either option in this PR.

## What I would like to do next

If we all agree on this strategy, I would like to start breaking this PR into separate segments and getting them merged. I think we can do all of this in a non-breaking manner, so hopefully we don't need this to be part of 3.2


d-v-b:
closing this in favor of #3332

Issue #3281, opened 2025-07-22
Enable deterministic and formatted metadata output


lkluft:
Currently, the order of entries in Zarr metadata is non-deterministic. This was previously discussed in issue #3254, where it was concluded that the Zarr specification does not prescribe a specific order—so downstream applications should not rely on one.

However, it would still be beneficial if writing **the same dataset** consistently produced **identical metadata**. This would, for example, facilitate computing dataset checksums.

Additionally, Zarr v3 now writes unformatted JSON to the `.zmetadata` file of v2 stores. In Zarr v2, human-readable (i.e., pretty-printed) metadata was preferred (see #704), a practice that v3 also adopted (see #1952). I assume that this is simply a regression. Ideally, metadata formatting would be consistent across versions and backends.

It would be great to consider enabling **deterministic** and **formatted** metadata output. This would improve reproducibility, readability, and tooling interoperability without conflicting with the specification.

/cc @d70-t

d-v-b:
this makes sense, I see two actionable items here:
- we ensure that the keys of the metadata dict are written in a consistent order (ideally the order in which the keys are defined in the zarr spec)
- We ensure that writing .zmetadata is sensitive to the global JSON formatting configuration.

evidently we were not testing either of these behaviors

d-v-b:
@lkluft can you link to any test failures caused by these regressions, or point me to the code that's sensitive to key order?

lkluft:
The following code snippet is sufficient to reproduce both issues:

```py
import xarray as xr


ds = xr.Dataset({"a": range(10), "b": range(10)})

# Creates a "non-formatted" store_v2.zarr/.zmetadata
ds.to_zarr("store_v2.zarr", zarr_format=2, mode="w")

# Creates a pretty-formatted store_v3.zarr/zarr.json
ds.to_zarr("store_v3.zarr", zarr_format=3, mode="w")
```

This directly shows that the v2 metadata is not formatted, whereas the v3 metadata is. _(In the v2 library, the v2 metadata was formatted.)_

The order of the keys is more subtle. While all Zarr-specific (top-level) keys appear to be deterministic, the order of the variables in the consolidated metadata is not. Here, "a" and "b" may appear in any order if the snippet is run repeatedly. It seems that the consolidation of the metadata occurs after the variables have been written. Consequently, a non-deterministic listing of the filesystem could cause this behaviour. However, sorting the consolidated metadata before attaching it should solve this.

[...]

PR #3288, opened 2025-07-23
Ensure deterministic ordering of consolidated metadata

lkluft:
This PR ensures that the contents of the consolidated metadata are sorted.

As a result, the consolidated metadata for the same input array becomes deterministic across all file systems, enabling, e.g., reliable checksum computation.

Keys are sorted in ascending order by path length (where a path is a sequence of strings joined by "/"). For keys with the same path length, lexicographic order is used as a tie-breaker (as suggested in #3281)

This sorting could be implemented in several places; for example, `flattened_metadata` could already be sorted. I chose to implement it in the `to_dict()` method, which seemed the least invasive and closest to the end of the internal metadata processing pipeline. However, I'm open to moving this elsewhere if there’s a better place for it.

I didn’t include a unit test because this behavior is difficult to test directly. It appears many existing tests assume sorted metadata. I suppose they pass due to the use of memory stores. That said, I'm happy to add a test if you have ideas on how best to validate this determinism.

TODO:


TomAugspurger:
Agreed.

Could you also update the docs (probably both the docs in `open_consolidated` and https://github.com/zarr-developers/zarr-python/blob/main/docs/user-guide/consolidated_metadata.rst) stating what users can rely on?

TomAugspurger:
And if you're feeling especially ambitious, you could note this as a desirable feature at https://github.com/zarr-developers/zarr-specs/pull/309. There could possibly be some indication in the Zarr metadata that the consolidated metadata has been sorted.

docs/user-guide/consolidated_metadata.rst
```diff
@@ -100,6 +100,11 @@ With nested groups, the consolidated metadata is available on the children, recu
    >>> consolidated['child'].metadata.consolidated_metadata
    ConsolidatedMetadata(metadata={'child': GroupMetadata(attributes={'kind': 'grandchild'}, zarr_format=3, consolidated_metadata=ConsolidatedMetadata(metadata={}, kind='inline', must_understand=False), node_type='group')}, kind='inline', must_understand=False)
 
+The keys in the consolidated metadata are sorted in ascending order by path
```
TomAugspurger: Can you clarify whether this is on the write side or read side (I think write?). That implies that users can rely on this when reading files written by zarr-python, but it's not necessarily safe to assume, unless we *always* sort when reading, since other systems writing consolidated metadata might not sort.

And perhaps note that this is new, by putting this in a `.. versionadded:: 3.1.1` (or whatever the next version is, I'm not sure).

TomAugspurger: Earlier in this doc, we do a `pprint(dict(sorted(consolidated_metadata.items())))`. Is that `sorted` unnecessary now?

src/zarr/core/group.py
```diff
@@ -141,7 +141,13 @@ def to_dict(self) -> dict[str, JSON]:
         return {
             "kind": self.kind,
             "must_understand": self.must_understand,
-            "metadata": {k: v.to_dict() for k, v in self.flattened_metadata.items()},
+            "metadata": {
+                k: v.to_dict()
+                for k, v in sorted(
+                    self.flattened_metadata.items(),
+                    key=lambda item: (len(item[0].split("/")), item[0]),
```
TomAugspurger:
```suggestion
                    key=lambda item: (item.count("/"), item[0]),
```

should be equivalent, and slightly faster.

lkluft:
I added a test that checks the proper sorting of a (nested) store structure. 

I also included a short paragraph in the user guide, that explains the new sorting of the consolidated metadata keys. I couldn't find a specific doc string to modify, and while thinking about it, the new behaviour seems less important on the individual function level, and more appropriate for the user guide.

src/zarr/core/group.py
```diff
@@ -141,7 +141,13 @@ def to_dict(self) -> dict[str, JSON]:
         return {
             "kind": self.kind,
             "must_understand": self.must_understand,
-            "metadata": {k: v.to_dict() for k, v in self.flattened_metadata.items()},
+            "metadata": {
+                k: v.to_dict()
+                for k, v in sorted(
```
TomAugspurger, line 147: How much should we worry about non-ascii characters here? https://zarr-specs.readthedocs.io/en/latest/v3/core/index.html#node-names just says a name is a "string of unicode code points."

Python's `sorted` on unicode data by code point, which isn't necessarily the same as alphabetical. Reading around (Chapter 4 in Fluent Python, if you have a copy) recommends [pyuca](https://github.com/jtauber/pyuca). https://realpython.com/python-sort-unicode-strings/ has an overview too. That seems complicated enough that just using the code point is fine.

We could consider normalizing `item[0]` (the second element of the tuple) using https://docs.python.org/3/library/unicodedata.html#unicodedata.normalize if you want.

lkluft: That's a fair point. So far, my main concern has been to produce deterministic output, so I haven't worried about that.

However, I am happy to normalise the key to improve robustness. If we take it a step further, we could also convert the key to lowercase so that the case does not affect the lexicographic order.

src/zarr/core/group.py
lkluft: Good catch! I think it is also more readable 👍

docs/user-guide/consolidated_metadata.rst
lkluft: As I understand it, the `to_dict()` method is only used before writing. Therefore, I assume that reading from a non-sorted store will maintain the same key order as in the store. I have updated the user guide accordingly.

I removed the `sorted()` function from the documentation, as I assume it is no longer needed. There are also instances in the tests where `sorted()` is used before matching against the expected outcome. Technically, these are also no longer needed. However, I would keep them to make those tests specific; a regression in the sort shouldn't cause those tests to fail.

I added a `versionadded` directive, but the result is rather bold – Sphinx puts the entire paragraph in a green box, as if it were a warning. I am happy to keep it that way, but personally, I don't like the aesthetic.

src/zarr/core/group.py
TomAugspurger: Sounds good to me. Having some stable sorting is good. If users want some potentially different order then they're free to sort in their application.

TomAugspurger (approved):
Thanks! I think the codecov failure is spurious.

TomAugspurger:
@dstansby are you able to take this the rest of the way? Your review is blocking the merge.

docs/user-guide/consolidated_metadata.rst
```diff
@@ -100,6 +100,14 @@ With nested groups, the consolidated metadata is available on the children, recu
    >>> consolidated['child'].metadata.consolidated_metadata
    ConsolidatedMetadata(metadata={'child': GroupMetadata(attributes={'kind': 'grandchild'}, zarr_format=3, consolidated_metadata=ConsolidatedMetadata(metadata={}, kind='inline', must_understand=False), node_type='group')}, kind='inline', must_understand=False)
 
+.. versionadded:: 3.1.1
+
+    The keys in the consolidated metadata are sorted prior to writing. Keys are
+    sorted in ascending order by path length, where a path is defined as a sequence
```
dstansby: When reading this, I interpreted "path length" as number of characters in a path, ie. `/abcde` is longer than `/a/b`, but I understand this isn't the case. I'm not sure what a better option is though, maybe "path depth" or "number of path components"?

dstansby (commented):
Implementation and docs look 👍 I have one minor doc comment inline, which I think is worth fixing before merging to avoid confusion.

docs/user-guide/consolidated_metadata.rst
lkluft: Fair point! I opted for "path depth" to keep it concise.

dstansby (approved):

PR #3292, opened 2025-07-25
Put back parse_data_type import

dstansby:
Fixes https://github.com/zarr-developers/zarr-python/pull/3264#issuecomment-3115427902.

I note that this error would have been much less likely to happen if we defined our API in one place, instead of defining it in `zarr.core` and re-exporting it. To make sure we're testing our public API, I've moved imports in the tests to come from `zarr.dtype` where possible.


d-v-b:
thanks for this fix @dstansby! I think we can skip the release notes here.

d-v-b (approved):

PR #3217, opened 2025-07-08
Fix race condition in from_array for arrays with shards

bojidar-bg:
Fixes #3169.

When passing data via `create_array(data = ...)`, the data is inserted by `from_array` after getting split by chunks. In current main, when using shards, `from_array` ends up splitting the data by sub-chunks (due to `AsyncArray.chunks` returning `Metadata.chunks` which returns the size of the chunks inside shards instead of the size of the "physical" chunk files), which ends up creating a race condition when the writes to multiple sub-chunk end up writing to the same physical chunk file.
This PR changes `from_array` to instead split the data by shard in case there are shards.

(Probably, there needs to be a lock somewhere in AsyncArray or ShardCodec which prevents multiple writes to the same shard (or, just physical chunk in general) to execute at the same time)

TODO:
* [x] Add unit tests and/or doctests in docstrings
* [x] Add docstrings and API docs for any new/modified user-facing classes and functions
* [ ] New/modified features documented in `docs/user-guide/*.rst`
* [x] Changes documented as a new file in `changes/`
* [ ] GitHub Actions have all passed
* [ ] Test coverage is 100% (Codecov passes)

Issue #3169, opened 2025-06-25
"Stored and computed checksum do not match" in case when shards large than chunks


rm1113:
### Zarr version

3.0.8

### Numcodecs version

0.16.1

### Python Version

3.13

### Operating System

Linux (WSL2)

### Installation

pip

### Description

When I try to create an array with shards size larger than chunk size, I got an error. 
```
Stored and computed checksum do not match. Stored: b'[#\xe5?'. Computed: b'"\xa7w\x8a'.
```
If the shard size equals the chunk size, the error disappears. 

Test cases:

- **Shard == chunk**  
    No error.

- **Shard and chunk share one dimension** (e.g. shards = (50, 200), chunks = (50, 100))  
    - With compressors=None: ~1% failure rate (1 failure in 100 runs)  
    - With any non-None compressor (including default): ~50% failure rate

- **Shard spans multiple chunks along each axis**  
    Almost 100% failure rate, even with compressors=None.

Sorry, if it is a duplicate

### Steps to reproduce

```python
# /// script
# requires-python = ">=3.11"
# dependencies = [
#   "zarr@git+https://github.com/zarr-developers/zarr-python.git@main",
# ]
# ///
#
# This script automatically imports the development branch of zarr to check for issues

import zarr
import numpy as np

N_ROWS, N_COLS = 10_000, 2_000
CHUNK_ROW, CHUNK_COL = 1_000, 100
SHARD_ROW, SHARD_COL = 5_000, 500
ARRAY_PATH = "/tmp/tmp.zarr"

rng = np.random.default_rng(seed=42)
x = rng.random((N_ROWS, N_COLS))

zarr.create_array(
            store=ARRAY_PATH,
            data=x,
            chunks=(CHUNK_ROW, CHUNK_COL),
            shards=(SHARD_ROW, SHARD_COL),
            overwrite=True,
        )
array = zarr.open_array(ARRAY_PATH)[:]
```

### Additional output

Full tracelog:
```
  File "/home/rm/test_zarr/test_write.py", line 32, in <module>
    zarr.create_array(
    ~~~~~~~~~~~~~~~~~^
                store=ARRAY_PATH,
                ^^^^^^^^^^^^^^^^^
    ...<3 lines>...
                overwrite=True,
                ^^^^^^^^^^^^^^^
            )
            ^
  File "/home/rm/test_zarr/.venv/lib/python3.13/site-packages/zarr/api/synchronous.py", line 888, in create_array
    sync(
[...]

PR #3299, opened 2025-07-25
fix chunk/shard iteration

d-v-b:
In `main` there are some routines for iterating over the chunks of an array, but these routines do not distinguish between chunks and shards (i.e., stored objects) for arrays with sharding. 

This PR adds a separate set of shard-specific iteration routines to complement our chunk-specific iteration routines. Various bugs related to iterating over chunks, when shards were the intended iteration target, have been fixed by these changes, notably bugs causing memory races when creating arrays via `create_array` (xref #https://github.com/zarr-developers/zarr-python/issues/3169)

I think this supersedes https://github.com/zarr-developers/zarr-python/pull/3217, @bojidar-bg I credited you as a co-author on one of these commits because your idea to change the iteration from chunks to shards was correct.

### DIFF
--- a/src/zarr/core/array.py
+++ b/src/zarr/core/array.py
@@ -5377,7 +5377,7 @@ def _iter_chunk_regions(
         The shape of the selection in grid coordinates.
 
     Returns
-    ------
+    -------
     region: tuple[slice, ...]
         A tuple of slice objects representing the region spanned by each shard in the selection.
     """


src/zarr/core/array.py
```diff
@@ -1184,15 +1185,44 @@ def basename(self) -> str:
...
+
+        Returns
+        -------
+        tuple[int, ...]
+            The shape of the chunk grid for this array.
+        """
+        return self.chunk_grid_shape
+
+    @property
+    def chunk_grid_shape(self) -> ChunkCoords:
```
d-v-b: I added a new property called `chunk_grid_shape` because `cdata_shape` is ambiguous. `cdata_shape` is still around, but it now uses `chunk_grid_shape`

src/zarr/core/array.py
```diff
@@ -1184,15 +1185,44 @@ def basename(self) -> str:
...
         -------
-        Tuple[int]
+        tuple[int, ...]
             The shape of the chunk grid for this array.
         """
-        return tuple(starmap(ceildiv, zip(self.shape, self.chunks, strict=False)))
+        return tuple(starmap(ceildiv, zip(self.shape, self.chunks, strict=True)))
+
+    @property
+    def shard_grid_shape(self) -> ChunkCoords:
```
d-v-b: this complements `chunk_grid_shape`.

src/zarr/core/array.py
```diff
@@ -1204,7 +1234,7 @@ def nchunks(self) -> int:
         int
             The total number of chunks in the array.
         """
-        return product(self.cdata_shape)
+        return product(self.shard_grid_shape)
```
d-v-b: this fixes bugs where the number of chunks (stored objects) was incorrectly counted for arrays with sharding.

bojidar-bg:
Ooh, that looks like a much more complete implementation of what I did in that other PR! Kudos; I wouldn't have been able to push things this far :sparkles::sparkles:

d-v-b:
> Ooh, that looks like a much more complete implementation of what I did in that other PR! Kudos; I wouldn't have been able to push things this far ✨✨

It turned out to be more than I expected :sweat_smile: .

src/zarr/core/array.py
bojidar-bg: Shouldn't it be called `nshards` in this case? Or at least documented as the "number of shards of the array"?

src/zarr/core/indexing.py
```diff
@@ -167,16 +167,76 @@ def _iter_grid(
...
             dimensions += (range(o, o + ss),)
-        yield from itertools.product(*(dimensions))
+        return itertools.product(*(dimensions))
 
     else:
         msg = f"Indexing order {order} is not supported at this time."  # type: ignore[unreachable]
         raise NotImplementedError(msg)
 
 
+def iter_regions(
```
d-v-b: this is a new function for iterating over contiguous regions. When we support irregular chunking, we can overload the type of `region_shape` accordingly.

src/zarr/core/array.py
d-v-b: Changing the docstring is an option. The problem with renaming it is that this an old attribute dating back to zarr python 2.x

d-v-b: for context, back in zarr python 2.x, there was no concept of shards, and the number of chunks in an array was exactly the number of objects you would create on disk. But with zarr python 3, we introduced the concept of sharding and now you can have many more chunks (regions that can be read with cheap IO) than shards (objects on disk).

dstansby: Three thoughts here:
1. From a "lets not break the users" point of view, changing the docstring would be less breaking. Given no users have complained that this is returning the wrong value, leaving the return value as is, and updating the docstring to match would be less disruption
2. Related, it's very worrying that changing/fixing this seemingly didn't break any of our tests...
3. +1 for adding an analogous `nshards` property, that gives the product of the shard shape. That makes for an easy to understand API where `nshards = prod(shard_grid_shape)`, and `nchunks = prod(chunk_grid_shape)`.

src/zarr/core/array.py
```diff
@@ -1184,15 +1185,44 @@ def basename(self) -> str:
...
+
+        Returns
+        -------
+        tuple[int, ...]
+            The shape of the shard grid for this array.
+        """
+        if self.shards is None:
+            shard_shape = self.chunks
+        else:
+            shard_shape = self.shards
```
dcherian, line 1240: love it. matches how I think about things.

src/zarr/core/array.py
d-v-b: > Related, it's very worrying that changing/fixing this seemingly didn't break any of our tests...

shards /= chunks only when sharding is turned on, and generally we are not testing everything against arrays with sharding. I fixed several extant bugs with the changes in this PR. You can see the tests I changed accordingly.

dcherian: I support @dstansby's (3). I think it's ok to be a little breaking here, sharding is pretty new to the community so we should fix conceptual errors ASAP

d-v-b: > 1. From a "lets not break the users" point of view, changing the docstring would be less breaking. Given no users have complained that this is returning the wrong value, leaving the return value as is, and updating the docstring to match would be less disruption

how would you propose changing the docstring here

src/zarr/core/array.py
```diff
@@ -1258,31 +1289,90 @@ def _iter_chunk_coords(
...
+        return iter_shard_coords(
+            array=self,
+            origin=origin,
+            selection_shape=selection_shape,
+        )
+
+    @deprecated("Use _iter_shard_keys instead")
     def _iter_chunk_keys(
         self, *, origin: Sequence[int] | None = None, selection_shape: Sequence[int] | None = None
     ) -> Iterator[str]:
```
dcherian: Can this be deleted? It seems to be a private method, so why deprecate it?

src/zarr/core/array.py
```diff
@@ -2289,6 +2383,7 @@ def nbytes_stored(self) -> int:
         """
         return sync(self._async_array.nbytes_stored())
 
+    @deprecated("Use _iter_shard_keys instead.")
     def _iter_chunk_keys(
         self, origin: Sequence[int] | None = None, selection_shape: Sequence[int] | None = None
```
dcherian: same here. Can we delete?

src/zarr/core/array.py
dstansby: I would go with simply

> The number of chunks in this array.

and complement it with a section in the user guide that explains what (to `zarr-python`) a `chunk` and `shard` is.

dstansby: > generally we are not testing everything against arrays with sharding

Indeed, this is what I find very worrying 😄

d-v-b: if there's a local consensus for reverting this change I'm happy to do it. We can complement it with `nshards`.

src/zarr/core/array.py
d-v-b: someone might be using it. It doesn't cost us anything to keep it around for at least a patch release or two.

src/zarr/core/array.py
d-v-b: we will also need to decide what to do about `nchunks_initialized`. That function is clearly scoped to chunks-qua-files, not chunks-qua-readable-atoms.

dstansby: If I have a shard with two chunks, and only assing to one chunk, does one chunk get written or two? 

- If one, then `nchunks_initialized` should stay as is
- If two, then `nchunks_initialized` should be rounded up to a multiple of the number of chunks in a shard

d-v-b: There is no way to tell if one of the chunks inside a shard has been initialized. You can only check if that chunk empty, which is not the same thing as uninitialized. When you write a shard, all the chunks are allocated space, so all the chunks exist from the start. But when you create an array, no shards (files) are created until you write some data to them. In this way chunks and shards are not analogous, because the chunks of a sharded array cannot be "missing" (unless the entire shard is missing).

d-v-b: so we could definitely report a subset of the chunks that are uninitialized (those chunks inside shards that have not been created yet), but there could still be uninitialized chunks that don't get counted by this metric (chunks in shards that were allocated but never written to).

d-v-b: in https://github.com/zarr-developers/zarr-python/pull/3299/commits/29c1cddee28b026e59e4745d559f6e0bd9136618 I deprecate `nchunks_initialized`, and recommend using `nshards_initialized` instead. Both `nshards_initialized` and  `nchunks_initialized` report  the same thing: the number of shards (stored objects) initialized in storage.

dstansby: >  When you write a shard, all the chunks are allocated space,

In that case, because the intent of `nchunks_initialized` is to report on how many chunks have been allocated space, I think it should report `nshards_initialized * chunks_per_shard`.

A [GitHub search for `nchunks_initialized`](https://github.com/search?q=nchunks_initialized&type=code) shows that it's often used in expressions like `n_chunks_initialized == nchunks`, so to maintain a coherent API and prevent breaking existing code, `nchunks_initialized` should be a measure of chunks (and `==nchunks` when a full array is written), and not a measure of shards.

src/zarr/core/array.py
dstansby: The point of marking stuff as private with an `_` is it saves us the maintenance burden of doing a deprecation dance. This is very commonly done in other scientific Python projects, so we should just delete any private API no longer used. It's also important not to signal that users can rely on deprecations happening to private API, which doing a deprecation would do.

src/zarr/core/array.py
d-v-b: > In that case, because the intent of nchunks_initialized is to report on how many chunks have been allocated space,

The intent of this property is to report how many chunks-as-files have been written to storage. I cannot see any use for reporting the product of the number of chunks per shard * the number of shards that have been written to storage, since the point is to _count the files you created_

dstansby: I think the point is to count the number of chunks you have written to storage? With the sharding codec, chunks no longer equal files, but it still makes sense to report on how many chunks have been written to storage.

d-v-b: > I think the point is to count the number of chunks you have written to storage? 

In the zarr python 2.x days, this property would be used to know _how many files have been written, out of the total possible number of files_. You would expect to be able to increment this number by 1, by writing to an unwritten chunk. I think it's still used this way today. 

Today for sharded arrays, the minimum unit if writing is not a chunk but a multiple of chunks (a shard). In this case it's simply not useful to report the number of chunks that have been written. When you are considering how much stuff you have written, you care about the smallest writable unit, which is a shard. I am struggling to find any conceivable use for reporting the number of chunks written, in a context where it's not possible to write one chunk at a time.

dstansby: >  In this case it's simply not useful to report the number of chunks that have been written.

Agreed here. How about instead erroring out on `nchunks_initialized` when there's a sharding codec, and pointing users to `nshards_initialized` in the error message? Whatever the solution, it shouldn't end up with `nchunks_initialized != nchunks` evaluating as `True` for a fully written array.

d-v-b: > but it still makes sense to report on how many chunks have been written to storage.

Just a reminder, with this API we literally cannot tell users how many chunks they have written to storage. "number of chunks written to storage" is not encoded in the state of the storage backend. But we _can_ report how many shards have been initialized -- the number of shards currently stored is literally this number.

dstansby:
Just to make sure we're on the same page here, this is where we're heading in my mind, as something that can go in the user guide:

> In zarr-python, a chunk of data represents a single file in storage. There is one exception: when a single sharding codec is part of the codecs. In this case a shard represents a single file in storage, and a chunk represents a unit of data the same size or smaller that can be read independently of other chunks in the shard. This distinction is made to allow zarr-python to optimize the creation and processing of sharded arrays.

src/zarr/core/array.py
d-v-b: > > In this case it's simply not useful to report the number of chunks that have been written.
> 
> Agreed here. How about instead erroring out on `nchunks_initialized` when there's a sharding codec, and pointing users to `nshards_initialized` in the error message? Whatever the solution, it shouldn't end up with `nchunks_initialized != nchunks` evaluating as `True` for a fully written array.

I really don't want to make some properties of the array raise an exception depending on the codec configuration. That's really bad design IMO. I think `nchunks_initialized` only makes sense if chunks are always files. Now that chunks are not always files, we have to deprecate `nchunks_initialized`, because it no longer makes any sense for a chunk to be initialized.

dstansby: But `nchunks_initialized` still makes sense and is a widely used part of the API for the non-shard case. I think a key bit of the API design is being able to ignore shards and API that has "shard" in the name entirely if you want to, which would require leaving `nchunks_initialized`.

To sum up options in the sharding case, `nchunks_initialized` could be:

1. Number of shards initialized. This is confusing though, because `nchunks_initialized` is not the same measure of chunks as other "chunk" properties.
2. Number of chunks assigned to. Not possible because we can't track which chunks in a single shard have been assigned to or not.
3. Number of chunks persisted to storage. Not neccesarily usful, but easy to keep track of (number of shards initialized * chunks per shard), and a consistent measure of chunks.
4. Raise an error. Confusing if erroring or not depends on codec configuration.

Out of those options, I still think number 3 is the best of a bad bunch, because it fits in with other "chunk" properties, doesn't break existing user code, and is easily explainable.

d-v-b: > Out of those options, I still think number 3 is the best of a bad bunch, because it fits in with other "chunk" properties, doesn't break existing user code, and is easily explainable.

I do agree with this, and I will make these changes. But I will also add a note in the documentation suggesting that people move to `nshards_initialized`, because it's much more useful for situations where you are counting stored objects

dstansby: Sorry, I missed out:

5. Deprecate/remove `*_initialized`, and replace with something like `nfiles` and `nfiles_initialized`.

Maybe it would be helpful to have `nfiles{_initialized}` anyway?

d-v-b: I think it's important that people know that shards == files. If we just had `nfiles_initialized`, it wouldn't be clear what those files correspond to. `nshards_initialized` makes this abundantly clear.

d-v-b: one wrinkle -- at the moment, an array without sharding has the `shards` attribute set to `None`. I think we want to change this so that `shards` is always a tuple of numbers

src/zarr/core/array.py
d-v-b: done

src/zarr/core/array.py
d-v-b: it's deleted!

src/zarr/core/array.py
d-v-b: as of https://github.com/zarr-developers/zarr-python/pull/3299/commits/0b9916443d555d9e762f5501314383dc828c26bf `nchunks_initialized` reports the product of shards_initialized and chunks per shard. I think for sharded arrays this value is kind of useless and potentially confusing, but we'll push that problem down the road.

dstansby:
As I understand, this has one breaking change (`nchunks_initialized` behaviour), but it's a bugfix so this should be milestoned 3.1.2?

d-v-b:
> As I understand, this has one breaking change (`nchunks_initialized` behaviour), but it's a bugfix so this should be milestoned 3.1.2?

that seems right

d-v-b:
some interesting new build errors that seem unrelated to my changes:

https://github.com/zarr-developers/zarr-python/actions/runs/16720510145/job/47323677345#step:5:33

```
3s
Run hatch build
──────────────────────────────────── sdist ─────────────────────────────────────
Setting up build environment
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/home/runner/.local/share/hatch/env/virtual/zarr/nOVTQ5-s/zarr-build/lib/python3.11/site-packages/hatchling/__main__.py", line 6, in <module>
    sys.exit(hatchling())
             ^^^^^^^^^^^
  File "/home/runner/.local/share/hatch/env/virtual/zarr/nOVTQ5-s/zarr-build/lib/python3.11/site-packages/hatchling/cli/__init__.py", line 26, in hatchling
    command(**kwargs)
  File "/home/runner/.local/share/hatch/env/virtual/zarr/nOVTQ5-s/zarr-build/lib/python3.11/site-packages/hatchling/cli/build/__init__.py", line 82, in build_impl
    for artifact in builder.build(
  File "/home/runner/.local/share/hatch/env/virtual/zarr/nOVTQ5-s/zarr-build/lib/python3.11/site-packages/hatchling/builders/plugin/interface.py", line 147, in build
    build_hook.initialize(version, build_data)
  File "/home/runner/.local/share/hatch/env/virtual/zarr/nOVTQ5-s/zarr-build/lib/python3.11/site-packages/hatch_vcs/build_hook.py", line 35, in initialize
    dump_version(self.root, self.metadata.version, self.config_version_file, **kwargs)
  File "/home/runner/.local/share/hatch/env/virtual/zarr/nOVTQ5-s/zarr-build/lib/python3.11/site-packages/setuptools_scm/_integration/dump_version.py", line 77, in dump_version
    write_version_to_path(
  File "/home/runner/.local/share/hatch/env/virtual/zarr/nOVTQ5-s/zarr-build/lib/python3.11/site-packages/setuptools_scm/_integration/dump_version.py", line 111, in write_version_to_path
    content = final_template.format(version=version, version_tuple=version_tuple)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
KeyError: 'scm_version'
```

dstansby:
I've started reviewing code, but first wanted to provide comments at a higher level:

To understand what this PR results in, I ran a little test script that compares properties for sharded and non-sharded arrays (see below). I like the concept that even when shards are not set, the shard properties still return the chunk values, but since this is a new concept it should be accomapanied with appropriate explanation in the [Sharding user guide section](https://zarr.readthedocs.io/en/stable/user-guide/arrays.html#sharding), and in the release notes (which could just be a link to the sharding user guide section).

It is a bit confusing that `.shards` now returns `None` with just chunks, where all the other shard releated properties return the equivalent chunk values. I would suggest returning `self.chunks` from `.shards` if there's no sharding to match the other new properties, and adding a new `.is_sharded: bool` property to determine whether shards are being used (previously one would have done `self.shards is None`).

It also feels like the scope of this PR continues to expand beyond just a bugfix into (very welcome!) additions and new concepts, so I would advocate for it landing in 3.2.0 instead of a micro release.

I have some pending inline comments, but I thought I'd leave them until we've hased out the higher level stuff. Thanks for this PR - I like where it's going a lot 😄 


```python
import zarr

arr = zarr.create_array(store={}, shape=(12,), chunks=(2,), dtype='uint8')
arr[0] = 1
print("For a non-sharded array")
print("-----------------------")
print(f"{arr.chunks=}")
print(f"{arr.chunk_grid_shape=}")
print(f"{arr.nchunks=}")
print(f"{arr.nchunks_initialized=}")
print()
print(f"{arr.shards=}")
print(f"{arr.shard_grid_shape=}")
print(f"{arr.nshards=}")
print(f"{arr.nshards_initialized=}")
print()

arr = zarr.create_array(store={}, shape=(12,), chunks=(2,), shards=(6,), dtype='uint8')
arr[0] = 1
print("For a sharded array")
print("-------------------")
print(f"{arr.chunks=}")
print(f"{arr.chunk_grid_shape=}")
print(f"{arr.nchunks=}")
print(f"{arr.nchunks_initialized=}")
print()
print(f"{arr.shards=}")
print(f"{arr.shard_grid_shape=}")
print(f"{arr.nshards=}")
print(f"{arr.nshards_initialized=}")
print()
```

```
For a non-sharded array
-----------------------
arr.chunks=(2,)
arr.chunk_grid_shape=(6,)
arr.nchunks=6
arr.nchunks_initialized=1

arr.shards=None
arr.shard_grid_shape=(6,)
arr.nshards=6
arr.nshards_initialized=1

For a sharded array
-------------------
arr.chunks=(2,)
arr.chunk_grid_shape=(6,)
arr.nchunks=6
arr.nchunks_initialized=3

arr.shards=(6,)
arr.shard_grid_shape=(2,)
arr.nshards=2
arr.nshards_initialized=1
```

d-v-b:
> It also feels like the scope of this PR continues to expand beyond just a bugfix into (very welcome!) additions and new concepts, so I would advocate for it landing in 3.2.0 instead of a micro release.

IMO getting the bugfix out is important enough that this should be targeted for release very soon, ideally the next release. We have a few options here but the simplest might be to make all the new API private, and then make it public for 3.2.0. 

Making the new shards API private would also defer a decision about `shards=None` vs `shards=<tuple>`

dstansby:
> IMO getting the bugfix out is important enough that this should be targeted for release very soon, ideally the next release. 

I agree, but there are lots of changes here that seem orthogonal to just fixing that bug.

d-v-b:
> > IMO getting the bugfix out is important enough that this should be targeted for release very soon, ideally the next release.
> 
> I agree, but there are lots of changes here that seem orthogonal to just fixing that bug.

They are conceptually extremely simple. we already have tools for iterating over the shape defined by a chunk. I am adding tools for iterating over the shape defined by a shard, where a shard == a chunk in many cases. if you want time to study these changes in greater detail, lets just make them private.

d-v-b:
Consider the new shard-centric API a fix not only for one specific bug, but a potentially large number of future bugs caused by iterating over chunk-sized regions when writing data. Now that we have added sharding, but not added a user-friendly API for iterating over shard-sized regions, we are setting up a massive footgun for users to break their IO routines in a concurrent context. That's why this PR adds new functions / methods.

dstansby:
> They are conceptually extremely simple.

They are not conceptually simple to me, and we've had a long discussion here before landing on a design. We shouldn't just make everything private and merge it without proper review, since private code still needs to be understood by developers. If code takes a long time to review, then it takes a long time to review - here there is the option of slimming this PR right down to only that needed to fix the bug, which would make for a quicker review.

My worry is the scope here has crept well beyond a targeted a minimal fix for a bug, to a set of wider changes that introduces new concepts and lots of new API that is not needed to fix that bug. The extra lines of code make PR review harder, and increase the surface area through which bugs could creep in, which is my reasoning for something this big to go in a minor (not micro) release.

So to move forward, I guess I can continue reviewing this, but it won't happen quickly because it's a large PR with lots of new API (public and private), or a smaller targeted just-a-bugfix PR would probably be quicker to review if there is an shorter term need to fix the bug.

d-v-b:
> since private code still needs to be understood by developers.

can you indicate which functions or routines are hard to understand here?

dstansby:
It's less a case of hard to understand, more a case of new concepts (what is a chunk? what is a shard? what do the eight chunk/shard properties mean? how does passing different values to functions affect these properties? ...). Again, I think this is really good and important work, but big enough and introduces/formalises enough concepts that it will take me a while to wrap my head around it all and review, and should in my opinion land in a minor release with comprehensive user guide documentation and examples to match the new API and concepts.

d-v-b:
> what is a chunk? what is a shard?

A chunk is the smallest region of an array that you can read in parallel. A shard is the smallest unit of an array that you can _write_ in parallel. Using this terminology, there is a 1:1 correspondence between shards and files. There is not necessarily a 1:1 correspondence between chunks and files. This confusion caused the bug that this PR fixes (along with some other bugs).

For arrays that don't use the sharding codec, the chunk shape is the same as the effective shard shape. For arrays that do use sharding, the shard shape is an integer multiple of the chunk shape.

For reference, these are not new concepts. The chunk / shard distinction already exists in `create_array` and in the `chunks` and `shards` attributes of an array. In fact I don't think this PR introduces any new concepts at all. I'm simply taking an existing chunk-centric API, and making a shard-centric version of that same API, and in the process resolving multiple bugs.

d-v-b:
as an indication of the low-complexity of these changes, read [the tests I added. ](https://github.com/zarr-developers/zarr-python/pull/3299/files#diff-d1499eae500b1036f9aa3e3c6266bc516b9f03efcd65644acac5418e406822ba). There's really nothing elaborate or deep going on here.

dstansby:
I think my earlier comments still needs addressing here:

> It is a bit confusing that .shards now returns None with just chunks, where all the other shard releated properties return the equivalent chunk values. I would suggest returning self.chunks from .shards if there's no sharding to match the other new properties, and adding a new .is_sharded: bool property to determine whether shards are being used (previously one would have done self.shards is None).

and 

> I like the concept that even when shards are not set, the shard properties still return the chunk values, but since this is a new concept it should be accomapanied with appropriate explanation in the [Sharding user guide section](https://zarr.readthedocs.io/en/stable/user-guide/arrays.html#sharding), and in the release notes (which could just be a link to the sharding user guide section).

Happy to try and review in full once we work these points out (especially the second one).

d-v-b:
> It is a bit confusing that .shards now returns None with just chunks, where all the other shard releated properties return the equivalent chunk values. I would suggest returning self.chunks from .shards if there's no sharding to match the other new properties, and adding a new .is_sharded: bool property to determine whether shards are being used (previously one would have done self.shards is None).

I agree that it's confusing, and I also agree that the `Array.shards` attribute should make sense in light of the other various shard-related properties. But an `is_sharded` property might also be confusing, especially for an array where `is_sharded == False` but `shards = <some tuple>`. What about naming this property something like `uses_sharding_codec`? IMO we should be normalizing "all arrays are sharded" as much as possible across the entire stack, and to me this argues for treating the _way_ an array is sharded (via the sharding codec, or not) as an implementation detail. But I don't feel strongly here except that whatever we do should not be confusing.

And sorry for all the pings recently but @zarr-developers/python-core-devs it would be great to get some other POVs here, since we are talking about some potential changes to the user-facing `Array` object.

For this PR, i'm totally fine making all the new `shard` methods private until we can decide on a good story for the array API. We can still use these methods to fix the various bugs related to mixing up chunks and shards. This would also address your second point, because we wouldn't need to add any new user-facing docs at this time.

dcherian:
> What about naming this property something like `uses_sharding_codec`?

This is better.

> IMO we should be normalizing "all arrays are sharded" as much as possible across the entire stack, and to me this argues for treating the way an array is sharded (via the sharding codec, or not) as an implementation detail. 

I'm 50/50 on this, given that historically Zarr has only had chunks, not shards; and that these are also kwargs to the constructor. It's going to be confusing regardless. Should we just call it `shards_or_chunks` :D 

>  i'm totally fine making all the new shard methods private until we can decide on a good story for the array API.

Let's do this ASAP, and punt on the changing the meaning of the `.shards` property.

dstansby:
Perhaps the recommendation to replace `array.shards == None` could be `array.shards == array.chunks` instead?

Certainly `.shards` shouldn't be changed in code that goes in 3.1.x now, so again I'd advocate for any new API being private, and only the minimum API added that's needed to fix the original issue.

d-v-b:
the new API is all private, let me know if we need to do anything else

d-v-b:
ping @dstansby @dcherian, let me know if we need anything else here.

src/zarr/core/indexing.py
```diff
@@ -134,11 +134,11 @@ def _iter_grid(
...
 
-    >>> tuple(iter_grid((2,3)), origin=(1,1))
-    ((1, 1), (1, 2), (1, 3), (2, 1), (2, 2), (2, 3))
+    >>> tuple(iter_grid((2,3), origin=(1,1)))
+    ((1, 1), (1, 2))
 
-    >>> tuple(iter_grid((2,3)), origin=(1,1), selection_shape=(2,2))
-    ((1, 1), (1, 2), (1, 3), (2, 1))
+    >>> tuple(iter_grid((2,3), origin=(0,0), selection_shape=(2,2)))
+    ((0, 0), (0, 1), (1, 0), (1, 1))
```
dcherian, line 141: hah that was so wrong earlier!

src/zarr/core/array.py
```diff
@@ -5007,3 +5228,160 @@ def _parse_data_params(
...
+    Parameters
+    ----------
+    array : Array | AsyncArray
+        The array to iterate over.
+    origin : Sequence[int] | None, default=None
+        The origin of the selection in grid coordinates.
+    selection_shape : Sequence[int] | None, default=None
+        The shape of the selection in grid coordinates.
+
+    Yields
```
dcherian:
```suggestion
    Returns
```

These aren't yielding any more. Correct?

d-v-b: correct, we are returning the iterator because I don't anticipate passing any signals down to this routine

src/zarr/core/indexing.py
```diff
@@ -163,14 +163,74 @@ def _iter_grid(
...
+    region_shape : Sequence[int]
+        The shape of the region to iterate over.
+    origin : Sequence[int] | None, default=None
+        The location, in grid coordinates, of the first region to return.
+    selection_shape : Sequence[int] | None, default=None
+        The shape of the selection, in grid coordinates.
+    order : Literal["lexicographic"], default="lexicographic"
+        The linear indexing order to use.
+
+    Returns
```
dcherian:
```suggestion
    Yields
```

src/zarr/core/indexing.py
d-v-b: :clown_face:

src/zarr/core/indexing.py
```diff
@@ -163,14 +163,74 @@ def _iter_grid(
...
+    """
+    grid_shape = tuple(ceildiv(d, s) for d, s in zip(domain_shape, region_shape, strict=True))
+    for grid_position in _iter_grid(
+        grid_shape=grid_shape, origin=origin, selection_shape=selection_shape, order=order
+    ):
+        out: list[slice] = []
+        for g_pos, r_shape, d_shape in zip(grid_position, region_shape, domain_shape, strict=True):
+            start = g_pos * r_shape
+            stop = start + r_shape
+            if trim_excess:
```
dcherian, line 230: just curious, is there a reason to not trim excess?

d-v-b: the chunks / shards written to storage are always fully-sized, no matter the shape of the array. trim excess lets you control whether iteration is done over array-indexing space (trim_excess=True) or stored objects indexing space (trim_excess=False), for the same shape

tests/test_array.py
```diff
@@ -375,48 +380,74 @@ def test_nchunks(test_cls: type[Array] | type[AsyncArray[Any]], nchunks: int) ->
...
+    if shard_shape is None:
+        chunks_per_shard = 1
+    else:
+        chunks_per_shard = np.prod(np.array(shard_shape) // np.array(chunk_shape))
+
+    arr = zarr.create_array(store, shape=shape, shards=shard_shape, chunks=chunk_shape, dtype="i1")
 
     # write chunks one at a time
-    for idx, region in enumerate(arr._iter_chunk_regions()):
+    for idx, region in enumerate(arr._iter_shard_regions()):
```
dcherian, line 405: these tests are a bit circular in that the `n*_initialized` property uses this function under the hood.

Not a blocker, but would be good to think about making this test more independent.

d-v-b: that's a good point, maybe later we can simplify this function and make each shard 1x1x1, then iterate over array elements directly

tests/test_indexing.py
```diff
@@ -1995,6 +1998,71 @@ def test_iter_chunk_regions():
...
+    else:
+        origin_parsed = origin
+    if selection_shape is None:
+        selection_shape_parsed = tuple(
+            ceildiv(ds, rs) - o
+            for ds, o, rs in zip(domain_shape, origin_parsed, region_shape, strict=True)
+        )
+    else:
+        selection_shape_parsed = selection_shape
+    for d_s, r_s, o, ss in zip(
```
dcherian, line 2039: Again here this is reimplementing the logic being tested.

Would be good to add some property tests in the future, for example,
1. I think we know what the last index of the last slice should be, depending on `trim_excess` being True or False.
2. we never receive a slice with `start` < corresponding index in origin.
3. `output shape == selection_shape` (with some trim_excess dependency perhaps)

Issue #3201, opened 2025-07-03
FsspecStore.from_url does not handle a swift:// URL


fjansson:
### Zarr version

3.0.9

### Numcodecs version

0.16.1

### Python Version

3.12.9

### Operating System

Linux

### Installation

using pip in virtual environment

### Description

I'm trying to write zarr data into a swift store using zarr v3.  Previously, using zarr v2, something like this used to work:
```
url = 'swift://swift.dkrz.de/some-path/'
store = zarr.storage.FSStore(url) 
ds.to_zarr(store)
```
Now I've tried different ways to use the FsspecStore, but I cannot get it to work, often meeting the error "path argument to FsspecStore must not include scheme". For example:

```
import zarr
url='swift://swift.dkrz.de/some-path/'
store = zarr.storage.FsspecStore.from_url(url, read_only=False)
```

```
File "/etc/ecmwf/nfs/dh1_home_b/nkjf/deep-botany-zarr/convert-to-zarr-0d.py", line 120, in <module>
    store = zarr.storage.FsspecStore.from_url(url, read_only=False)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/etc/ecmwf/nfs/dh1_home_b/nkjf/deep-botany-zarr/zarr_env/lib/python3.12/site-packages/zarr/storage/_fsspec.py", line 260, in from_url
    return cls(fs=fs, path=path, read_only=read_only, allowed_exceptions=allowed_exceptions)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/etc/ecmwf/nfs/dh1_home_b/nkjf/deep-botany-zarr/zarr_env/lib/python3.12/site-packages/zarr/storage/_fsspec.py", line 149, in __init__
    raise ValueError(f"path argument to FsspecStore must not include scheme ({scheme}://)")
ValueError: path argument to FsspecStore must not include scheme (swift://)
```

### Steps to reproduce

```python
# /// script
# requires-python = ">=3.11"
# dependencies = [
#   "zarr@git+https://github.com/zarr-developers/zarr-python.git@main",
# ]
# ///
#
# This script automatically imports the development branch of zarr to check for issues

import zarr
url='swift://swift.dkrz.de/some-path/'
[...]

PR #3302, opened 2025-07-28
Add exception for swift in fsspec path handling

fjansson:
Proposed as fix for #3201.

http(s) and swift filesystems need the scheme in the path. This commit adds a special case for the swift filesystem, when checking for the scheme presence and generating an exception if the scheme is present.

Also remove one if for the http(s) schemes. Here all schemes can call fs._strip_protocol(path). For the filesystems that need the scheme in the path, this function will not remove it.

TODO:


src/zarr/storage/_fsspec.py
```diff
@@ -249,9 +249,8 @@ def from_url(
 
         # fsspec is not consistent about removing the scheme from the path, so check and strip it here
         # https://github.com/fsspec/filesystem_spec/issues/1722
-        if "://" in path and not path.startswith("http"):
-            # `not path.startswith("http")` is a special case for the http filesystem (¯\_(ツ)_/¯)
-            path = fs._strip_protocol(path)
+        # for http(s) and swift, _strip_protocol() does nothing, since they need the protocol in the path
```
dstansby: I don't think the new comment is needed, and the comments above can be removed too.

src/zarr/storage/_fsspec.py
```diff
@@ -139,8 +139,8 @@ def __init__(
                 f"fs ({fs}) was not created with `asynchronous=True`, this may lead to surprising behavior",
                 stacklevel=2,
             )
-        if "://" in path and not path.startswith("http"):
-            # `not path.startswith("http")` is a special case for the http filesystem (¯\_(ツ)_/¯)
+        if "://" in path and not path.startswith("http") and not path.startswith("swift"):
+            # special cases for the http(s) and swift filesystems (¯\_(ツ)_/¯)
```
dstansby:
```suggestion
            # special cases for the http(s) and swift filesystem, which need the protocol leaving in the path.
```

dstansby (changes requested):
Thanks for the PR - could you also add a test to make sure this is working as intended? Just constructing a `FsspecStore` with a valid swift URL should be good enough.

src/zarr/storage/_fsspec.py
fjansson: ok, I've changed this.

src/zarr/storage/_fsspec.py
fjansson: changed this also

fjansson:
I tried adding a test, but I don't really know what I'm doing here.

```
def test_swift_store() -> None:
    """
    Test creating an FsspecStore with a SWIFT URL.
    """
    store = FsspecStore.from_url("swift://swift.dkrz.de/dkrz_xxx/testcontainer/test.zarr")
```
* The URL doesn't exist, I think this is fine for testing the logic of leaving swift:// in the path.
* I think the swiftspec package is optional. Does the test need to be marked as requiring this package somehow? The test now gives "ValueError: Protocol not known: swift" which I think is what happens if swiftspec is not there
* The pre-commit test complains about assigning to an unused variable, I guess I can assign to `_` to show the result will not be used?

dstansby:
> The pre-commit test complains about assigning to an unused variable, I guess I can assign to _ to show the result will not be used?

I think you could just not do any assignment at all, and it would fix this.

> The URL doesn't exist, I think this is fine for testing the logic of leaving swift:// in the path.

Yes, that's fine.

> I think the swiftspec package is optional. Does the test need to be marked as requiring this package somehow? The test now gives "ValueError: Protocol not known: swift" which I think is what happens if swiftspec is not there

This is a bit of a pain. In order to avoid us having to test different protocols, could we just get rid of the two blocks of code that read:
```python
		# fsspec is not consistent about removing the scheme from the path, so check and strip it here
        # https://github.com/fsspec/filesystem_spec/issues/1722
        if "://" in path and not path.startswith("http"):
            # `not path.startswith("http")` is a special case for the http filesystem (¯\_(ツ)_/¯)
            path = fs._strip_protocol(path)
```
and instead rely on `fsspec` to handle raising errors on invalid paths?

I've tried removing them from my local copy of zarr and running the tests, and they all seem to pass. I don't know if @jhamman has anything to add as the author of these blocks of code - is there a reason we shouldn't remove them and depend on fsspec to error on invalid paths?

fjansson:
@dstansby sounds good to me. However, I think in the case of `from_url()`, the line `path = fs._strip_protocol(path)` should be kept, to give the file systems that need it a chance to get rid of the protocol in the path. That line can safely be run for http(s) and swift (which want the protocol preserved), those file systems will just return the path unchanged.

On the other hand the error checking, e.g. generate an exception when there is a protocol in the path and the file system doesn't want it, seems fine to leave for the individual file systems.

fjansson:
@dstansby I found out that ` url_to_fs` already calls `_strip_protocol()`, so there's no need to call it again. 
E.g. [here](https://github.com/fsspec/filesystem_spec/discussions/1356) it's stated that url_to_fs splits an url into a filesystem and a path suitable for that filesystem. I think both those blocks can be deleted as you say. I'll make another PR with that to see what happens.

dstansby:
Wonderful, thanks for following up on this!

fjansson:
The problem was fixed with #3343, so I'm closing this one.

PR #3304, opened 2025-07-28
Parametrize Array with v2/v3 metadata

dstansby:
It has been a longstanding bugbear of mine that there's no easy way to specify a v2 or v3 array type. This especially came up in the context of https://github.com/zarr-developers/zarr-python/pull/3257, which deals specifically with v2/v3 arrays.

This PR ads type parametrization to the `Array` class.

After this PR, there is lots of improvements to adding overloads to functions and methods that could be made, but to keep review easier I'd like to leave that for a follow up PR.

### DIFF
--- a/src/zarr/core/array.py
+++ b/src/zarr/core/array.py
@@ -2041,7 +2041,7 @@ class Array(Generic[T_ArrayMetadata]):
     _async_array: AsyncArray[T_ArrayMetadata]
 
     @property
-    def async_array(self) -> AsyncArray[ArrayV3Metadata] | AsyncArray[ArrayV2Metadata]:
+    def async_array(self) -> AsyncArray[T_ArrayMetadata]:
         """An asynchronous version of the current array.  Useful for batching requests.
 
         Returns


src/zarr/types.py
```diff
@@ -0,0 +1,9 @@
+from typing import Any
+
+from zarr.core.array import Array
+from zarr.core.metadata.v2 import ArrayV2Metadata
+from zarr.core.metadata.v3 import ArrayV3Metadata
+
+AnyArray = Array[Any]
+Arrayv2 = Array[ArrayV2Metadata]
+Arrayv3 = Array[ArrayV3Metadata]
```
d-v-b:
```suggestion
ArrayV2 = Array[ArrayV2Metadata]
ArrayV3 = Array[ArrayV3Metadata]
```

src/zarr/types.py
```diff
@@ -0,0 +1,23 @@
...
+AnyAsyncArray: TypeAlias = AsyncArray[Any]
+"""A Zarr format 2 or 3 `AsyncArray`"""
+
+AsyncArrayV2: TypeAlias = AsyncArray[ArrayV2Metadata]
+"""A Zarr format 2 `AsyncArray`"""
+
+AsyncArrayV3: TypeAlias = AsyncArray[ArrayV3Metadata]
+"""A Zarr format 3 `AsyncArray`"""
+
+AnyArray: TypeAlias = Array[Any]
```
d-v-b, line 16: Should this be `Array[ArrayV2Metadata] | Array[ArrayV3Metadata]`

dstansby: Yep, I guess that's safer 👍

dstansby: I couldn't get that to work, beacuse of an error that I sort of but not really understand ☹️ . I think the current state with `Any` is fine though - `.metadata` still gets inferred correctly:

```python
import zarr
import zarr.storage

arr = zarr.ones(shape=(1,))
meta = arr.metadata
reveal_locals()
"""
test.py:6: note: Revealed local types are:
test.py:6: note:     arr: zarr.core.array.Array[Any]
test.py:6: note:     meta: zarr.core.metadata.v2.ArrayV2Metadata | zarr.core.metadata.v3.ArrayV3Metadata
"""
```

src/zarr/api/asynchronous.py
```diff
@@ -1096,9 +1093,7 @@ async def empty(
     return await create(shape=shape, fill_value=None, **kwargs)
 
 
-async def empty_like(
-    a: ArrayLike, **kwargs: Any
-) -> AsyncArray[ArrayV2Metadata] | AsyncArray[ArrayV3Metadata]:
+async def empty_like(a: ArrayLike, **kwargs: Any) -> AnyAsyncArray:
```
d-v-b, line 1128: if you have the energy for it, I think we could type **kwargs on one place with a typeddict and use [`Unpack`](https://docs.python.org/3/library/typing.html#typing.Unpack) in the signatures here.

this would allow us to use an `overload` to declare that `empty_like(..., zarr_format=2) -> AsyncArray[ArrayV2Metadata]`

src/zarr/core/array.py
```diff
@@ -2091,7 +2092,7 @@ def open(
             Array opened from the store.
         """
         async_array = sync(AsyncArray.open(store))
-        return cls(async_array)
+        return Array(async_array)
```
d-v-b: why this change? I think using `cls` here ensures that subclasses can re-use this method

src/zarr/api/asynchronous.py
dstansby: I would like to leave anything beyond the new parametrization and using the new `Any*` types to follow up PRs if that's okay, to keep this PR tightly scoped.

src/zarr/types.py
```diff
@@ -0,0 +1,23 @@
+from typing import Any, TypeAlias
+
+from zarr.core.array import Array, AsyncArray
+from zarr.core.metadata.v2 import ArrayV2Metadata
+from zarr.core.metadata.v3 import ArrayV3Metadata
```
d-v-b, line 5: these imports will prevent `zarr.core.array` from using anything defined in `zarr.types`, is that intentional?

d-v-b: (outside of a type-checking context, I mean)

d-v-b: thinking more about this, I think as long as _everything_ in `zarr.types` is imported exclusively in `if TYPE_CHECKING` blocks, then there's no risk of circular imports. we would only have circular imports if we tried to work with types as values outside of type-checking, for example using `get_args(ZarrFormat)`. We can always avoid this by defining a typed constant value somewhere else in the codebase.

d-v-b:
I think these changes are good, as they make it easier to track the v2-ness of v3-ness of the `Array` class. 

But since this PR changes the `Array` class, and it will also require anyone who annotated variables with the `Array` type to update that annotation, I think it's important that we get a lot of eyes on this PR.

cc @zarr-developers/core-devs

src/zarr/core/array.py
dstansby: Ah yes, the correct fix here was to keep `cls` but change the return type to `Self`. Fixed, and fixed in a couple of other places in this file too.

src/zarr/types.py
dstansby: I certinaly didn't think about this! So neither intentional or unintentional. I'm not sure I see a better solution, but open to suggestions.

d-v-b:
> I think these changes are good, as they make it easier to track the v2-ness of v3-ness of the `Array` class.
> 
> But since this PR changes the `Array` class, and it will also require anyone who annotated variables with the `Array` type to update that annotation, I think it's important that we get a lot of eyes on this PR.
> 
> cc @zarr-developers/core-devs

ping @zarr-developers/core-devs

src/zarr/types.py
d-v-b: What was the problem that necessitated defining type aliases in the top-level `types` file? I wonder if there's another way to solve that. This is my only concern about this PR btw, we can get it in soon after this is resolved

dstansby: I think there wasn't a problem, and this was a choice 😄 - the circular import mess seems a good reason to move these out and next to the array/group classes. Might be a little while until I find time, so someone else is very free to finish this PR off if I don't get there first.

d-v-b:
this looks good to me, I'm going to merge it as soon as everything is green, thanks for the work here @dstansby

d-v-b:
@dstansby I pushed some fixes to merge conflicts, let me know those changes are OK

Issue #3295, opened 2025-07-25
Using r+ mode with zarr.open creates an empty directory


K-Meech:
### Zarr version

v3.1.0

### Numcodecs version

0.16.1

### Python Version

3.13

### Operating System

Windows

### Installation

Using pip into a virtual environment

### Description

Opening a non-existent location via `zarr.open` with `r+` mode (see 'steps to reproduce') creates an empty directory at that location. Given the description in the `zarr.open` docstring: 'r+ means read/write (must exist)', I would expect no files to be changed before the `GroupNotFoundError` is thrown.

### Steps to reproduce

```python
# /// script
# requires-python = ">=3.11"
# dependencies = [
#   "zarr@git+https://github.com/zarr-developers/zarr-python.git@main",
# ]
# ///
#
# This script automatically imports the development branch of zarr to check for issues

import zarr

# creates the new directory data/images/test_r_plus.zarr, then throws GroupNotFoundError
zarr_object = zarr.open(store="data/images/test_r_plus.zarr", mode="r+")

# Using mode r, doesn't create a new directory before it throw GroupNotFoundError
zarr_object = zarr.open(store="data/images/test_r.zarr", mode="r")
```

K-Meech:
The issue seems to be inside [`make_store_path`](https://github.com/zarr-developers/zarr-python/blob/a0c56fbbba12ca4402de3d83043e5f840406d670/src/zarr/storage/_common.py#L270) in `src/zarr/storage/_common.py`:
- `_read_only` is only set for `mode == 'r'`
- then a local store is opened like: `await LocalStore.open(root=Path(store_like), read_only=_read_only)` (where `__read_only` is False for `r+` mode)
- Then inside `LocalStore` in `src/zarr/storage/_local.py` - a new directory is [made by `_open`](https://github.com/zarr-developers/zarr-python/blob/a0c56fbbba12ca4402de3d83043e5f840406d670/src/zarr/storage/_local.py#L113) in every case where `_read_only = False`

PR #3307, opened 2025-07-29
Prevent mode='r+' from creating new directories

dstansby:
Fixes https://github.com/zarr-developers/zarr-python/issues/3295. I'm not sure if I came up with the most elegant solution, but I couldn't think of a way to solve this without passing the mode to `LocalStore.open()`. Other suggestions very welcome if anyone has any better ideas?

### DIFF
--- a/changes/3307.bugfix.rst
+++ b/changes/3307.bugfix.rst
@@ -1,0 +1,1 @@
+Opening an array or group with ``mode="r+"`` will no longer create new arrays or groups.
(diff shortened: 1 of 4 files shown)


d-v-b:
this works but I find the logic hard to follow. I think that's a side effect of our failure to distinguish two separate things:

1. is the store instance allowed to modify existing objects
2. is the store instance allowed to create new objects

these are technically two separate permissions, but for zarr purposes, 2 implies 1.

We need to formalize these differences higher up in the store permissions model I think.

d-v-b (approved):

PR #3308, opened 2025-07-30
Refactor make_store_path for clarity

dstansby:
This refactors `make_store_path` to make the code easier to read by:

- Moving the `TypeError` for unused `storage_options` to the top of the funciton
- This then allows the if/else logic to be made one level deep, where each branch is for a different store type. Previously there were two levels of if/else.

There should not be any behaviour changes here.


d-v-b:
nice! the _ultimate_ refactor would be removing the `StorePath` class entirely, and adding a `path` attribute to the store class, and giving the store classes all of the `StorePath` methods. But what you have here is a much easier lift

PR #3318, opened 2025-07-31
add numcodec protocol

d-v-b:
This PR adds a protocol to model `numcodecs.abc.Codec`. The motivation for this protocol is to ensure that we can process external codecs that adhere to the numcodecs API without needing numcodecs as a dependency.

TODO:

### DIFF
--- a/src/zarr/abc/store.py
+++ b/src/zarr/abc/store.py
@@ -6,10 +6,6 @@ from dataclasses import dataclass
 from itertools import starmap
 from typing import TYPE_CHECKING, Protocol, runtime_checkable
 
-from zarr.core.buffer.core import default_buffer_prototype
-from zarr.core.common import concurrent_map
-from zarr.core.config import config
-
 if TYPE_CHECKING:
     from collections.abc import AsyncGenerator, AsyncIterator, Iterable
     from types import TracebackType
@@ -438,6 +434,9 @@ class Store(ABC):
         # Note to implementers: this default implementation is very inefficient since
         # it requires reading the entire object. Many systems will have ways to get the
         # size of an object without reading it.
+        # avoid circular import
+        from zarr.core.buffer.core import default_buffer_prototype
+
         value = await self.get(key, prototype=default_buffer_prototype())
         if value is None:
             raise FileNotFoundError(key)
@@ -476,6 +475,11 @@ class Store(ABC):
         # on to getting sizes. Ideally we would overlap those two, which should
         # improve tail latency and might reduce memory pressure (since not all keys
         # would be in memory at once).
+
+        # avoid circular import
+        from zarr.core.common import concurrent_map
+        from zarr.core.config import config
+
         keys = [(x,) async for x in self.list_prefix(prefix)]
         limit = config.get("async.concurrency")
         sizes = await concurrent_map(keys, self.getsize, limit=limit)
--- a/changes/3372.misc.rst
+++ b/changes/3372.misc.rst
@@ -1,0 +1,2 @@
+Make certain imports in ``zarr.abc.store`` local to method definitions. This minimizes the risk of
+circular imports when adding new classes to ``zarr.abc.store``.


d-v-b:
The way our code coverage works, class and function definitions are not counted as tested if they are imported by pytest prior to running actual tests. Pytest imports from a lot of our library due to fixtures defined in `conftest.py`, which is why our coverage reports in CI report function definitions as uncovered. 

This means adding new functions always reduces coverage, because the function definition is uncovered. ~Until we fix how our coverage is measured, I excluded instances of `NotImplementedError` with `#pragma: no cover` to get coverage green for this PR.~

tests/test_api.py
```diff
@@ -1282,7 +1282,7 @@ def test_gpu_basic(store: Store, zarr_format: ZarrFormat | None) -> None:
             dtype=src.dtype,
             overwrite=True,
             zarr_format=zarr_format,
-            compressors=compressors,
+            compressors=compressors,  # type: ignore[arg-type]
```
dstansby, line 1286: There's a lot of `type: ignore` comments currently dotted throughout - what's the reason for that? Is it perhaps because numcodecs.abc.Codec doesn't currently conform to the new protocol?

d-v-b: this particular example is a test, and if you look a few lines above you will see that `compressors` is either `None` or the string "auto", and mypy models this as `str | None`, leading to this error:

`tests/test_api.py:1285: error: Argument "compressors" to "create_array" has incompatible type "str | None"; expected "CompressorsLike"  [arg-type]`

I suspect the other `type: ignore` statements were added for different reasons

d-v-b:
> Is it perhaps because numcodecs.abc.Codec doesn't currently conform to the new protocol?

Spinning [this question](https://github.com/zarr-developers/zarr-python/pull/3318#discussion_r2254131010) out into the main thread: `numcodecs.abc.Codec` does not have type annotations for its methods, and even if it did, `decode` would need to be annotated as returning `object` thanks to [`Pickle.decode`](https://github.com/zarr-developers/numcodecs/blob/506c89b29d0ac5109ab6348ce5814149c1aa2b00/numcodecs/pickles.py#L41-L49), which can return literally any python object.

without type annotations on `numcodecs.abc.Codec`, I don't think we can use a type checker to check if a numcodecs codec is assignable to `Numcodec`, but this PR does add runtime checks, which are tested.

dstansby:
To make sure I understand, does passing an existing numcodecs codec to `compressors` or `filters` fail type checking in user code with this PR (because they don't conform to the new protocol)?

martindurant:
> even if it did, decode would need to be annotated as returning object thanks to [Pickle.decode](https://github.com/zarr-developers/numcodecs/blob/506c89b29d0ac5109ab6348ce5814149c1aa2b00/numcodecs/pickles.py#L41-L49)

Shouldn't it annotate our expectations, but allow for the (one?) exception to mark itself as non compliant? We probably don't want any other codecs making Object, and this is a way to say that. In fact, it raises the other question, for a different discussion: is having the pickle protocol reasonable?

d-v-b:
> Shouldn't it annotate our expectations, but allow for the (one?) exception to mark itself as non compliant? We probably don't want any other codecs making Object, and this is a way to say that.

Yes, which is why in this PR I am annotating the [`encode` and `decode` methods](https://github.com/zarr-developers/zarr-python/pull/3318/files#diff-dbd789290de692359a9581e84b215714da488bcab2fdce5419048e7cfee2081dR26-R30) as returning `Buffer | NDBuffer`. 

> In fact, it raises the other question, for a different discussion: is having the pickle protocol reasonable?

It might be reasonable for some applications (it was evidently reasonable enough to get implemented in the first place). It's also incredibly dangerous, because it performs arbitrary code execution.

d-v-b:
> To make sure I understand, does passing an existing numcodecs codec to `compressors` or `filters` fail type checking in user code with this PR (because they don't conform to the new protocol)?

this is a great question, and the answer depends on which type checker you ask.

if we take this script:

```python
# /// script
# requires-python = ">=3.11"
# dependencies = [
#   "zarr"
# ]
# ///

import zarr

from numcodecs import GZip
import numcodecs
from zarr.abc.numcodec import Numcodec

a: Numcodec = GZip()
b: numcodecs.abc.Codec = GZip()
```

mypy reports no problems:
```
bennettd@dvb-desktop-0 ➜  zarr-python git:(feat/numcodecs-protocol) ✗ uvx mypy test_3.py
Success: no issues found in 1 source file
```

but the more thorough based-pyright type checker complains about both `a` and `b`, for different reasons:

```bennettd@dvb-desktop-0 ➜  zarr-python git:(feat/numcodecs-protocol) ✗ uvx basedpyright test_3.py 
Installed 2 packages in 126ms
/home/bennettd/dev/zarr-python/test_3.py
  /home/bennettd/dev/zarr-python/test_3.py:8:8 - warning: Import "zarr" is not accessed (reportUnusedImport)
  /home/bennettd/dev/zarr-python/test_3.py:10:6 - warning: Stub file not found for "numcodecs" (reportMissingTypeStubs)
  /home/bennettd/dev/zarr-python/test_3.py:11:8 - warning: Stub file not found for "numcodecs" (reportMissingTypeStubs)
  /home/bennettd/dev/zarr-python/test_3.py:14:15 - error: Type "GZip" is not assignable to declared type "Numcodec"
    "GZip" is incompatible with protocol "Numcodec"
      "encode" is an incompatible type
        Type "(buf: Unknown) -> bytes" is not assignable to type "(buf: Buffer | NDBuffer) -> (Buffer | NDBuffer)"
          Function return type "bytes" is incompatible with type "Buffer | NDBuffer"
            Type "bytes" is not assignable to type "Buffer | NDBuffer"
      "decode" is an incompatible type
        Type "(buf: Unknown, out: Unknown | None = None) -> (Unknown | bytes)" is not assignable to type "(buf: Buffer | NDBuffer, out: Buffer | NDBuffer | None = None) -> (Buffer | NDBuffer)"
          Function return type "Unknown | bytes" is incompatible with type "Buffer | NDBuffer"
    ... (reportAssignmentType)
  /home/bennettd/dev/zarr-python/test_3.py:15:4 - warning: Type of "abc" is unknown (reportUnknownMemberType)
  /home/bennettd/dev/zarr-python/test_3.py:15:4 - warning: Type of "Codec" is unknown (reportUnknownMemberType)
  /home/bennettd/dev/zarr-python/test_3.py:15:14 - error: "abc" is not a known attribute of module "numcodecs" (reportAttributeAccessIssue)
2 errors, 5 warnings, 0 notes
```

so, tldr is that the status quo is broken because numcodecs doesn't support type checking, AND the new protocol is using a too-narrow type. i'll look into fixing this, but it's important to keep in mind that the status quo is broken.

edit: changing the numcodecs annotation from `numcodecs.abc.Codec` to `Codec`, after importing `Codec` from `numcodecs.abc` does pass pyright. but we never use that import in our codebase.

martindurant:
Sorry for a sidetrack:

I notice you include `uv` metadata in the snippet and `uvx` command to do run the procedure. It would be even nicer if you could directly pass an HTTP URL, but github doesn't generate such a thing for in-comment snippets (as opposed to a gist, for example). My question: do you know if there's a way to indicate in the matadata that you intend to run it with `mypy` rather than execute the script directly? As it is, you have two pieces of metadata, one in the file and the invocation command, where one place would be much nicer.

d-v-b:
> Sorry for a sidetrack:
> 
> I notice you include `uv` metadata in the snippet and `uvx` command to do run the procedure. It would be even nicer if you could directly pass an HTTP URL, but github doesn't generate such a thing for in-comment snippets (as opposed to a gist, for example). My question: do you know if there's a way to indicate in the matadata that you intend to run it with `mypy` rather than execute the script directly? As it is, you have two pieces of metadata, one in the file and the invocation command, where one place would be much nicer.

I'm not aware of anything like that, since it would be a bit unusual to prepare a python script exclusively for the purposes of type checking, although I think I could just declare mypy as a dependency and invoke it via this very script, e.g. `if __name__ == '__main__': run_mypy(__file__)`

edit: here's an updated version of the code snippet that will invoke the type checker when you run the script:

```python
# /// script
# requires-python = ">=3.11"
# dependencies = [
#   "uv",
#   "zarr @ git+https://github.com/d-v-b/zarr-python.git@0d0712f0974c243e376ba2fd9de0078be1978bae",
# ]
# ///

import zarr

from numcodecs import GZip
from numcodecs.abc import Codec
from zarr.abc.numcodec import Numcodec

a: Numcodec = GZip()
b: Codec = GZip()

if __name__ == "__main__":
    import subprocess
    subprocess.run(["uvx", "basedpyright", __file__])
```

martindurant:
People do all kinds of unusual stuff :) 

Using `__main__` feels wrong to me too - now you are encoding intent into the source itself, which then gets checked too. Maybe it just goes beyond what you should bundle into a single script snippet. Sorry for the diversion!

d-v-b:
> People do all kinds of unusual stuff :)
> 
> Using `__main__` feels wrong to me too - now you are encoding intent into the source itself, which then gets checked too. Maybe it just goes beyond what you should bundle into a single script snippet. Sorry for the diversion!

there is literally no purpose for this example except to demonstrate the results of the type check, so actually I think everything is as it should be.

d-v-b:
I made a change to the types of `encode` and `decode` (they are now `typing_extensions.Buffer`. This essentially moves the error reported by basedpyright to another area (the return type of `get_config`). But I don't think this matters, because the point is not to pass the type checker. it's to define a model of the various codecs defined in numcodecs. let's evaluate it against that criterion.

d-v-b:
For context, this protocol is very helpful for abstracting over v2 and v3 codecs, which I demonstrated in https://github.com/zarr-developers/zarr-python/pull/3276. This PR is the first of hopefully several separate PRs that will achieve the functionality demonstrated in #3276 (using a single set of codec classes).

see https://github.com/zarr-developers/zarr-python/issues/3162 for an outline of the broader plan.

d-v-b:
as of https://github.com/zarr-developers/zarr-python/pull/3318/commits/82992c5710e9a83a266dffdce1a20d3f5a18606c I'm using `Any` for the input / output types of `Numcodec`. We can narrow this type as needed later on.

TomAugspurger:
Something discussed briefly on the dev call today: I'd love if we could avoid the checks (`isinstance` on main`, the custom `is_*` checks on this PR).

As @d-v-b mentioned, we'd ideally have some validation at the outer layer. Once we're past that, we know we have alid metadata, so we can get from there to (what should be) a valid python codec class. At some point we should be able to avoid these checks.

But we do this already on `main` so fixing that here is out of scope for this PR.

src/zarr/abc/numcodec.py
```diff
@@ -0,0 +1,54 @@
+from typing import Any, Self, TypeGuard
+
+from typing_extensions import Protocol
+
+
+class Numcodec(Protocol):
+    """
+    A protocol that models the ``numcodecs.abc.Codec`` interface.
+    """
```
maxrjones, line 12:
```suggestion

	This protocol should be considered experimental; expect the typing for `buf`, and `out` to become stricter.
    """
```

src/zarr/abc/numcodec.py
```diff
@@ -0,0 +1,54 @@
...
+
+    codec_id: str
+
+    def encode(self, buf: Any) -> Any: ...
```
maxrjones:
```suggestion
    def encode(self, buf: Any) -> Any: ...
        """Encode data in `buf`.

        Parameters
        ----------
        buf
            Data to be encoded. May be any object supporting the new-style
            buffer protocol.

        Returns
        -------
        enc
            Encoded data. May be any object supporting the new-style buffer
            protocol.
        """
```

src/zarr/abc/numcodec.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def decode(self, buf: Any, out: Any | None = None) -> Any: ...
```
maxrjones:
```suggestion
    def decode(self, buf: Any, out: Any | None = None) -> Any: ...
        """
        Decode data in `buf`.

        Parameters
        ----------
        buf : Any
            Encoded data. May be any object supporting the new-style buffer
            protocol.
        out : Any
            Writeable buffer to store decoded data. N.B. if provided, this buffer must
            be exactly the right size to store the decoded data.

        Returns
        -------
        dec : Any
            Decoded data. May be any object supporting the new-style
            buffer protocol.
        """
```

src/zarr/abc/numcodec.py
maxrjones: Maybe it'd be helpful to at least have a alias for Any to make it more clear that this will be eventually restricted to Buffer-Like objects?

src/zarr/abc/numcodec.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def get_config(self) -> Any: ...
```
maxrjones:
```suggestion
    def get_config(self) -> Any: ...
        """
        Return a dictionary holding configuration parameters for this
        codec. Must include an 'id' field with the codec identifier. All
        values must be compatible with JSON encoding.
        """
```

src/zarr/abc/numcodec.py
```diff
@@ -0,0 +1,54 @@
...
+
+    @classmethod
+    def from_config(cls, config: Any) -> Self: ...
```
maxrjones:
```suggestion
    def from_config(cls, config: Any) -> Self: ...
        """
        Instantiate codec from a configuration object.
        """
```

src/zarr/abc/store.py
```diff
@@ -476,6 +474,9 @@ async def getsize_prefix(self, prefix: str) -> int:
         # on to getting sizes. Ideally we would overlap those two, which should
         # improve tail latency and might reduce memory pressure (since not all keys
         # would be in memory at once).
+        from zarr.core.common import concurrent_map
+        from zarr.core.config import config
```
maxrjones: were these linter fixes? I'm curious why the imports were moved as part of this PR?

maxrjones (commented):
I made some very minor suggestions around providing docstrings since the typing is now so broad.

Mostly, I'm confused about the protocol not aligning with the Zarr V3 specification. https://zarr-specs.readthedocs.io/en/latest/v3/core/index.html#id22 states:

> Logically, a codec c must define three properties:
> 
> c.compute_encoded_representation_type(decoded_representation_type), a procedure that determines the encoded representation based on the decoded representation and any codec parameters. In the case of a decoded representation that is a multi-dimensional array, the shape and data type of the encoded representation must be computable based only on the shape and data type, but not the actual element values, of the decoded representation. If the decoded_representation_type is not supported, this algorithm must fail with an error.
> 
> c.encode(decoded_value), a procedure that computes the encoded representation, and is used when writing an array.
> 
> c.decode(encoded_value, decoded_representation_type), a procedure that computes the decoded representation, and is used when reading an array

It doesn't seem like the V2 spec provided any definitions for a codec. Will external codec developers follow the numcodec protocol or the Zarr V3 specification?

d-v-b:
> Mostly, I'm confused about the protocol not aligning with the Zarr V3 specification. https://zarr-specs.readthedocs.io/en/latest/v3/core/index.html#id22 states:

This is deliberate. This protocol is a model of a thing that already exists (the numcodecs codec ABC). The numcodecs codec ABC is not a v3 codec, so neither is this model. There is no intention of using the `Numcodec` protocol as a v3-style codec in zarr python. Instead, in a later PR I will define a [`NumcodecWrapper` class that _is_ a v3 codec](https://github.com/zarr-developers/zarr-python/blob/9055a1aaff563867c11eb5ef50c1fb89767a06aa/src/zarr/codecs/numcodec.py#L107). This class will wrap the methods defined on the `Numcodec` protocol in a v3-codec-compatible way.

> It doesn't seem like the V2 spec provided any definitions for a codec. Will external codec developers follow the numcodec protocol or the Zarr V3 specification?

Although the v2 spec does define the [JSON form of a codec](https://github.com/zarr-developers/zarr-specs/blob/adba8a2c14bdea1d880add8e9ffbffc6d6e3941a/docs/v2/v2.0.rst?plain=1#L66-L68), it does not define the codec API. But the protocol I'm adding in this PR is narrowly scoped to modeling a particular API in numcodecs, not anything in the v2 spec (even though that numcodecs API does implement the v2 spec).

d-v-b:
> Will external codec developers follow the numcodec protocol or the Zarr V3 specification?

Reminder that we are adding this protocol for compatibility with existing codecs that were written for zarr python 2.x. Developers working on new codecs that target zarr python 3.x should almost certainly use the v3-style codec API defined in `zarr.abc.codec.Codec`.

src/zarr/abc/store.py
d-v-b: I think I needed to move these imports at some point in the process of this PR to avoid circular imports. See #3338, which these changes would resolve. But these changes might not be necessary in the final state of this PR (I would need to test). If they are not necessary here, we can either keep them in this PR, or I can put them in a separate PR.

src/zarr/abc/numcodec.py
```diff
@@ -0,0 +1,96 @@
+from typing import Any, Self, TypeGuard
+
+from typing_extensions import Protocol
+
+
+class Numcodec(Protocol):
```
normanrz, line 6: Could this work with https://docs.python.org/3/library/typing.html#typing.runtime_checkable?

d-v-b: I tried :( but I learned that `@runtime_checkable` doesn't work for protocols with non-methods (i.e., attributes), and the `numcodecs.abc.Codec` ABC uses `codec_id` as an attribute. I'm not 100% sure we need the `codec_id` here, but if we ever wanted to register these codecs, then it would be important.

src/zarr/abc/store.py
d-v-b: I opted for a separate PR: https://github.com/zarr-developers/zarr-python/pull/3372/files

src/zarr/abc/numcodec.py
d-v-b: important context that I forgot when I answered this question: given the structure of `Numcodec`,  `runtime_checkable` would work for `isinstance` checks, but not for `issubclass` checks, and I think there are places where we need the latter. But we should revisit this later if it turns out that we don't need `issubclass`, or if we want to use `isinstance` and rely on a custom function for `issubclass`.

Issue #2900, opened 2025-03-10
Some `numcodecs.zarr3` codecs fail with 3.0.4+


LDeakin:
### Zarr version

3.0.4+

### Numcodecs version

0.15.0

### Python Version

3.12

### Operating System

Linux

### Installation

See reproducer

### Description

The `ZFPY` and `Pcodec` codecs (and perhaps more) fail since 3.0.4 because they expect input arrays to be contiguous

### Steps to reproduce

```python
#!/usr/bin/env -S uv run
# /// script
# requires-python = ">=3.12"
# dependencies = [
#     # "zarr==3.0.2,<3.0.3", # WORKS
#     "zarr==3.0.4", # FAILS
#     "numcodecs==0.15.0",
#     "zfpy==1.0.1",
#     "pcodec==0.3.2",
# ]
# ///

import numpy as np
import zarr
from numcodecs.zarr3 import ZFPY, PCodec

for serializer in [
    ZFPY(mode = 4, tolerance=0.01),
    PCodec(level = 8, mode_spec="auto")
]:
    array = zarr.create_array(
        store=zarr.storage.MemoryStore(),
        shape=[2, 2],
        chunks=[2, 1],
        dtype=np.float32,
        serializer=serializer,
    )
    array[...] = np.array([[0, 1],[2, 3]])
```

d-v-b:
thanks for this report. I see 2 problems here: 
- the bug itself, which I cannot yet explain in terms of code changes that went in to the 3.04 release. Is it possible that this is somehow a side-effect of https://github.com/zarr-developers/zarr-python/pull/2851?
- We are not testing enough. clearly we need to be testing against all codecs in numcodecs.

mickyals:
## Zarr Version
3.0.6

## Numcodecs Version
0.15.1

## Operating System
Linux

## Description
The same issue as described previously - `ValueError: an array with contiguous memory is required` occurs due to PCodec expectation of contiguous arrays in memory. The result is corruption of the compressed arrays as shown below. Are there any updates regarding this bug? 

![Image](https://github.com/user-attachments/assets/9efb4ae5-7f6c-46c5-b073-33b7da8bc9e7)

d-v-b:
@dcherian any guesses what could be going on here?

dcherian:
Sorry for not responding.  We are now better about not copying,  so we are tripping up numcodecs

[...]

PR #3326, opened 2025-08-03
fix for failing numcodecs.zarr3 codecs

terraputix:
Closes https://github.com/zarr-developers/zarr-python/issues/2900.

I did not add a unit test because it was already mentioned in https://github.com/zarr-developers/zarr-python/issues/2900 that all numcodecs should be better tested in CI, but I guess that belongs into another PR. 

You can verify that it works by running the following script:
```bash
#!/usr/bin/env -S uv run
# /// script
# requires-python = ">=3.12"
# dependencies = [
#     # "zarr==3.0.2,<3.0.3", # WORKS
#     # "zarr==3.0.4", # FAILS
#     "zarr@git+https://github.com/terraputix/zarr-python.git@8f40aaf3f3fa1d23918520018c60dd0034997f76",  # WORKS
#     "numcodecs==0.15.0",
#     "zfpy==1.0.1",
#     "pcodec==0.3.2",
# ]
# ///
#


import numpy as np
import zarr
zarr.__version__ = "3.0.2"
from numcodecs.zarr3 import ZFPY, PCodec

for serializer in [
    ZFPY(mode=4, tolerance=0.01),
    PCodec(level=8, mode_spec="auto"),
]:
    array = zarr.create_array(
        store=zarr.storage.LocalStore("test"),
        shape=[2, 2],
        chunks=[2, 1],
        dtype=np.int64,
        serializer=serializer,
        compressors=None,
        overwrite=True,
    )
    array[...] = np.array([[0, 1], [2, 3]])
```

TODO:

### DIFF
--- a/src/zarr/core/codec_pipeline.py
+++ b/src/zarr/core/codec_pipeline.py
@@ -315,7 +315,7 @@ class BatchedCodecPipeline(CodecPipeline):
             # TODO: For the last chunk, we could have is_complete_chunk=True
             #       that is smaller than the chunk_spec.shape but this throws
             #       an error in the _decode_single
-            return chunk_value
+            return chunk_value.copy()  # make a writable copy
         if existing_chunk_array is None:
             chunk_array = chunk_spec.prototype.nd_buffer.create(
                 shape=chunk_spec.shape,


terraputix:
I improved this fix, avoiding the extra copy which I introduced before. 

Basically, this change reverts the codec pipeline modifications introduced in https://github.com/zarr-developers/zarr-python/pull/2851 and provides an alternative fix for handling chunks at the end of the array when the chunk shape does not evenly divide the array shape.
 
Do you mind taking a look at this one  @dcherian or @d-v-b ?

d-v-b:
I had a look, but I don't know this part of the code well and that function has no comments or docstring(!), so I'm not sure how much my review is worth. If the tests pass and someone who knows the code a bit better gives it a thumbs-up (cc @dcherian) then I think we can merge.

but why are we fixing this in zarr python, instead of in the individual codecs?

terraputix:
> but why are we fixing this in zarr python, instead of in the individual codecs?

I actually tried the solution proposed in https://github.com/zarr-developers/zarr-python/issues/2900#issuecomment-2778946258, but it was breaking a handful of other tests in numcodecs...

This PR is basically an improved version of https://github.com/zarr-developers/zarr-python/pull/2851, which also does not require copies in downstream codecs and (I think) therefore should be preferred.

d-v-b:
@TomAugspurger / @dcherian could you have a look at this?

d-v-b:
@terraputix could you add your snippet as a test? As minimal as possible but IMO it's fine to add dependencies as needed to our test suite to ensure that we catch this bug

crisbodnar:
Any updates on this issue?

d-v-b:
> I had a look, but I don't know this part of the code well and that function has no comments or docstring(!), so I'm not sure how much my review is worth. If the tests pass and someone who knows the code a bit better gives it a thumbs-up (cc @dcherian) then I think we can merge.
> 
> but why are we fixing this in zarr python, instead of in the individual codecs?

@dcherian @TomAugspurger could yall have a look?

d-v-b:
these changes look innocuous enough. I'm going to merge this and then put some energy into improving it in a follow-up PR

PR #3331, opened 2025-08-04
DOC: update numpy versions in test case examples

jo-mueller:
## Updated environment names in contribution guide

Hi there, first-timer here :) I went to go through the doc pages for contributing and tried running the test commands outlined there:

```
...
hatch env run --env test.py3.12-2.1-optional run-coverage
...
```

which leads to an error (`Unknown environment: test.py3.12-2.1-optional`) since the test matrix is specified with numpy 2.2:
```
[[tool.hatch.envs.test.matrix]]
python = ["3.11", "3.12", "3.13"]
numpy = ["1.26", "2.2"]
deps = ["minimal", "optional"]
```

So I just changed the example commands in the doc pages to work right away with the current `pyproject.toml`. It's a tiny little change but hopefully makes the entrance for some a little easier if they haven't worked with hatch, etc before.

Thanks for considering!

### DIFF
--- a/docs/developers/contributing.rst
+++ b/docs/developers/contributing.rst
@@ -98,7 +98,7 @@ you can do something like the following::
 To verify that your development environment is working, you can run the unit tests
 for one of the test environments, e.g.::
 
-    $ hatch env run --env test.py3.12-2.1-optional run-pytest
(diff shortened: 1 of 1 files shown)


d-v-b:
thanks for this!

d-v-b (approved):

Issue #3333, opened 2025-08-04
Mysterious code coverage drop


dstansby:
At some point between July 14th and July 21st code [coverage dropped from 95% to 60%](https://app.codecov.io/gh/zarr-developers/zarr-python). We didn't suddenly remove a load of tests or add a load of untested code, so this must be an issue withhow code coverage is being collected, perhaps due to a new version of a test dependency? This should be fixed.

@d-v-b had some potential insight in https://github.com/zarr-developers/zarr-python/pull/3318#issuecomment-3145159862: 

> The way our code coverage works, class and function definitions are not counted as tested if they are imported by pytest prior to running actual tests. Pytest imports from a lot of our library due to fixtures defined in conftest.py, which is why our coverage reports in CI report function definitions as uncovered.

d-v-b:
you can check this locally by running `coverage` before running tests e.g. `coverage run -m pytest tests`, this will show very high coverage and more significantly it will indicate that all the function definitions are covered.

dstansby:
Why isn't that being reflected in codecov then?

dstansby:
/ why did it stop working when once it was working?

dstansby:
Right, 9969a5d5aaf9bb62fdf5757a749535ec411461df is the last commit with 95% coverage, and 9ae874c5305436f5e76b97b0a54bece3e5d9ea84 is the first commit with 60% coverage

dstansby:
Aha, https://github.com/zarr-developers/zarr-python/commit/9ae874c5305436f5e76b97b0a54bece3e5d9ea84 is probably the culprit - I'm guessing adding these lines might have caused this:

https://github.com/zarr-developers/zarr-python/blob/9ae874c5305436f5e76b97b0a54bece3e5d9ea84/pyproject.toml#L460-L461

dstansby:
Yep, it looks like we're running into https://github.com/pytest-dev/pytest-cov/issues/510 which says there's a solution documented at https://pytest-cov.readthedocs.io/en/latest/plugins.html

d-v-b:
quoting @TomAugspurger  from the PR that added this line:

> so that pytest calls our code to register this marker when these tests run.

I'm guessing this is the process that's driving down the reported coverage. Maybe the simplest fix is to not use pytest-cov here, and instead measure coverage with `coverage run -m ...`. Although this was significantly slower in my experience.

d-v-b:
[...]

PR #3334, opened 2025-08-04
Fix test coverage

dstansby:
Fixes https://github.com/zarr-developers/zarr-python/issues/3333 (hopefully). To prevent me having to copy/paste environment variables for four different hatch environments, I also de-duplicated the configuration of the test environments. 🤞 this works fine, happy to revert the de-duplication and do a copy-and-paste job if folks would prefer that to keep things a bit simpler.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -150,6 +150,12 @@ dependencies = [
 ]
 features = ["test"]
 
+[tool.hatch.envs.test.env-vars]
+# Required to test with a pytest plugin; see https://pytest-cov.readthedocs.io/en/latest/plugins.html
+COV_CORE_SOURCE = "src"
+COV_CORE_CONFIG = ".coveragerc"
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -257,13 +234,22 @@ dependencies = [
...
-[tool.hatch.envs.min_deps.scripts]
-run = "pytest --verbose"
-run-hypothesis = "pytest --hypothesis-profile ci tests/test_properties.py tests/test_store/test_stateful*"
+
+[tool.hatch.envs.doctest]
+features = ["test", "optional", "remote", "remote_tests"]
+description = "Test environment for doctests"
+
+[tool.hatch.envs.doctest.scripts]
+run = "rm -r data/; pytest docs/user-guide --doctest-glob='*.rst'"
```
d-v-b, line 243: maybe for another PR but perhaps we should be using temporary directories for the docs examples

PR #3343, opened 2025-08-05
fsspec path handling

fjansson:
Proposed as fix for #3201, as an alternative to the one in #3302.

Some filesystems need the scheme as part of the path, while others don't. FsspecStore.from_url() throws an exception if the scheme is left in the path, for any filesystem except http and https. However, the swift fs also needs the scheme in the path. This commit removes the path check, rather than adding more special cases which are difficult to unit test (discussed in #3302).

TODO:

### DIFF
--- a/changes/3530.bugfix.md
+++ b/changes/3530.bugfix.md
@@ -1,0 +1,2 @@
+Fixed a bug that prevented Zarr Python from opening Zarr V3 array metadata documents that contained
+ extra keys with permissible values (dicts with a `"must_understand"` key set to `"false"`).
--- a/src/zarr/core/array.py
+++ b/src/zarr/core/array.py
@@ -103,10 +103,10 @@ from zarr.core.indexing import (
 from zarr.core.metadata import (
     ArrayMetadata,
     ArrayMetadataDict,
+    ArrayMetadataJSON_V3,
     ArrayV2Metadata,
     ArrayV2MetadataDict,
     ArrayV3Metadata,
-    ArrayV3MetadataDict,
     T_ArrayMetadata,
 )
 from zarr.core.metadata.io import save_metadata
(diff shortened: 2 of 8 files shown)


d-v-b:
@jhamman could you have a look

d-v-b:
I would _really_ like the `path` attribute of `FSSpecStore` to be consistent with the normal definition of a path in the context of a URL. This means, given the URL `"https://mydata.zarr"`, the `FSSpecStore.path` should be `"mydata.zarr"`, not `"https://mydata.zarr"`, because that's not the path of the URL!

I understand that fsspec has some design choices that make this tricky, but we should see what we can do to hide that stuff instead of propagating forward

d-v-b:
@fjansson take a look at https://github.com/d-v-b/zarr-python/tree/remove-fsspec-path-check, would that work for you? Basically I ensure that the `Fsspecstore.path` is always a valid URL path, but internally we give fsspec the full URL for operations that need it. We need to test this on lots of different fsspec backends though.

fjansson:
I like the idea: zarr internally has a path without scheme, and passes to fsspec what it wants to see.

I already made a comment on  `self._url = self.fs.to_dict()["protocol"] + "://" + path` - I think this should be `self.path` where the scheme has been removed.

Then I'm confused: we now pass a full url where it previously was a path. But some of the fsspec file systems cannot handle a scheme in their paths, others demand it. This has previously been handled by calling `fs._strip_protocol(path)` for each file system, to let them remove the scheme or not.

fjansson:
@d-v-b Would it be an idea to merge the two commits needed for SWIFT access to work, before the larger path handling effort? My two commits only remove code, fixing the path handling seems a lot more complex.

d-v-b:
I'm inclined to merge this in order to unblock SWIFT access.

PR #3345, opened 2025-08-05
fix: check for non-int 0 fill values

ilan-gold:
A bit of an "oops" from me, but hopefully this is more robust (than #3082)! It worked for my one example but testing this without using a spy object seems impossible (happy to contribute that though, or some other test).

TODO:


d-v-b:
to test this, could we perhaps patch `np.full` and `np.zeros` to raise exceptions, that you would then catch in the test? I'm thinking there would be 2 test functions, 1 testing the various inputs that should hit the `np.full` branch, and another testing the various inputs that should hit the `np.zeros` branch. This could also work for a single test function that internally checks which numpy routine the input scalar should trigger

tests/test_buffer.py
```diff
@@ -243,3 +244,29 @@ def test_empty(
         assert result.flags.c_contiguous  # type: ignore[attr-defined]
     else:
         assert result.flags.f_contiguous  # type: ignore[attr-defined]
+
+
+@pytest.mark.parametrize("dtype", [np.int8, np.uint16, np.float32, int, float])
+@pytest.mark.parametrize("fill_value", [None, 0, 1])
```
dstansby, line 250:
```suggestion
@pytest.mark.parametrize("fill_value", [None, 0, 0.0, 1])
```
Worth explicitly including a float here?

dstansby (changes requested):
Looks good to me - needs a release note entry, and I left one suggestion to improve the test too.

ilan-gold:
I've done some digging into this topic, I'll post in a bit but I would like to hold off on merging for a bit.

tests/test_buffer.py
ilan-gold: I think everything is cast anyway by `dtype`, so it shouldn't matter, no?

ilan-gold:
Ok apologies, I know that Tom's PR supercedes this in many ways but for anyone curious, my worry was about how memory is allocated.

I learned that `zeros` is only faster than `full` because it doesn't allocate memory when requested, lazily doing so later when needed (see https://stackoverflow.com/questions/70055063/how-is-memory-handled-once-touched-for-the-first-time-in-numpy-zeros/70056188#70056188).  In benchmarks, I see the times of subsequent operations on the matrix, whether something like `sum` that should touch all elements or `__setitem__` to be identical independent of initialization so I definitely think this should be merged!


```python
import numpy as np
buf = np.arange(100_000, dtype=np.float64)
# cell 1
%%timeit full = np.full((1_000_000,), 0, dtype=np.float64)

full[900_000:] = buf
# cell 2
%%timeit zeros = np.zeros((1_000_000,), dtype=np.float64)
    
zeros[900_000:] = buf
# cell 3
%%timeit full = np.full((1_000_000,), 0, dtype=np.float64)

full.sum()
# cell 4
%%timeit zeros = np.zeros((1_000_000,), dtype=np.float64)

zeros.sum()

```

Screenshot 2025-08-25 at 16 46 16

d-v-b:
@ilan-gold is this still active?

ilan-gold:
Looking back at the history here, the PR itself, and my last comment, I don't see why not. It's pretty marginal though.  Maybe @TomAugspurger has an opinion.

Issue #3352, opened 2025-08-06
`zarr.api.asynchronous.open_like` is broken


d-v-b:
[this function](https://github.com/zarr-developers/zarr-python/blob/1264a4d190a0f22a3612e302d41f4413f06edf35/src/zarr/api/asynchronous.py#L1273-L1295) is intended for creating new arrays similar to existing arrays, but it has a few problems, the most significant being that it attempts to invoke `open_array` without passing a `mode` keyword argument, which errors. It's actually hard to tell what `mode=None` does in this context? I think the ultimate behavior might be store-specific, but I suspect it defaults to `r` for memory storage. This means `open_like` can only read existing arrays, which defeats its entire purpose.

We should either fix this function or remove it.

PR #3353, opened 2025-08-06
chore/add missing tests

d-v-b:
This PR excludes some unreachable lines of code from test coverage, and adds tests for a few missing routines in `zarr.api.asynchronous`. 

I put those missing tests in a new directory to ensure that the tests are found in a location that matches the layout of the source code (unlike our current api tests). Eventually we should move _all_ of the tests in `tests/test_api.py` to `tests/test_api/test_asynchronous.py` or `tests/test_api/test_synchronous.py`.

I did not add a test for `api.asynchronous.open_like` because it's broken. see #3352 .

### DIFF
--- a/src/zarr/abc/codec.py
+++ b/src/zarr/abc/codec.py
@@ -1,11 +1,14 @@
 from __future__ import annotations
 
 from abc import abstractmethod
-from typing import TYPE_CHECKING, Generic, TypeVar
+from collections.abc import Mapping
+from typing import TYPE_CHECKING, Generic, TypeGuard, TypeVar
+
+from typing_extensions import ReadOnly, TypedDict
 
 from zarr.abc.metadata import Metadata
 from zarr.core.buffer import Buffer, NDBuffer
-from zarr.core.common import ChunkCoords, concurrent_map
+from zarr.core.common import ChunkCoords, NamedConfig, concurrent_map
 from zarr.core.config import config
 
 if TYPE_CHECKING:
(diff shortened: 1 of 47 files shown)


src/zarr/abc/codec.py
```diff
@@ -138,7 +138,7 @@ def validate(
         """
 
     async def _decode_single(self, chunk_data: CodecOutput, chunk_spec: ArraySpec) -> CodecInput:
-        raise NotImplementedError
+        raise NotImplementedError  # pragma: no cover
```
TomAugspurger, line 141: Out of curiosity, do you know why this isn't abstract if we're assuming that subclasses override it?

tests/test_api/test_asynchronous.py
```diff
@@ -0,0 +1,105 @@
...
+) -> None:
+    """
+    Test the like_args function
+    """
+    assert _like_args(observed, {}) == expected
+
+
+async def test_open_no_array() -> None:
+    """
+    Test that zarr.api.asynchronous.open attempts to open a group when is no array found, but shape was specified in kwargs.
```
TomAugspurger:
```suggestion
    Test that zarr.api.asynchronous.open attempts to open a group when no array is found, but shape was specified in kwargs.
```

TomAugspurger: And I'd be fine with deprecating this behavior (in another PR).

TomAugspurger (approved):

src/zarr/abc/codec.py
d-v-b: I was wondering the same thing, I'm not sure why.

tests/test_api/test_asynchronous.py
d-v-b: yes it should absolutely be deprecated!

src/zarr/abc/codec.py
TomAugspurger: Thanks, please don't go spelunking on my account :)

PR #3355, opened 2025-08-06
add modules for types

d-v-b:
This does a few things:
- collects a lot of type definitions and constants in `zarr.core.types`, 
- exports some of those types from `zarr.types`
- defines complete typeddict models for zarr v2 and zarr v3 metadata documents. These are useful enough on their own that we should add them to the library even if this PR goes nowhere.

### DIFF
--- a/examples/custom_dtype.py
+++ b/examples/custom_dtype.py
@@ -29,7 +29,8 @@ from zarr.core.dtype.common import (
     DTypeJSON,
     check_dtype_spec_v2,
 )
-from zarr.core.types import JSON, ZarrFormat
+from zarr.core.types import JSON
+from zarr.types import ZarrFormat
 
 # This is the int2 array data type
 int2_dtype_cls = type(np.dtype("int2"))
--- a/examples/custom_dtype.py
+++ b/examples/custom_dtype.py
@@ -22,7 +22,6 @@ import numpy as np
 import pytest
 
 import zarr
-from zarr.core.common import JSON, ZarrFormat
 from zarr.core.dtype import ZDType, data_type_registry
 from zarr.core.dtype.common import (
     DataTypeValidationError,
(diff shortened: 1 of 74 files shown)


dstansby:
The trouble with defining objects in two places (here, `zarr.types` and `zarr.core.types`) is duplication, and also means we end up importing stuff in the tests from the private place (`zarr.core.types` here) instead of the public place (`zarr.types` here), which in the past has prevented us from noticing a change in the public API (`zarr.types` here) because we're not actually using it in any of our tests.

To prevent this, could we either:

1) Make all the types public and just put them in `zarr.types`
2) Put all types in `zarr.types`, but make some private with a leading underscore in their name

?

Related, there are types that are in `zarr.core.types` that are used in the public API (at least `ShapeLike`, I'm sure many others too). If a type is used in the public API, it should be publicly documented, so there needs to be re-evaluation of what is public/private types here.

d-v-b:
> Related, there are types that are in `zarr.core.types` that are used in the public API (at least `ShapeLike`, I'm sure many others too). If a type is used in the public API, it should be publicly documented, so there needs to be re-evaluation of what is public/private types here.

that might be the first step to refining this PR -- we should figure out exactly which types are public. Definitely anything used as an annotation of a public function signature should be public. In this case, these types should be defined just once in `zarr.types`. All other private types can be defined wherever is convenient for them, which could be `zarr.core.types` but we could also have private, dtype-specific types in `zarr.core.dtype`, for example

d-v-b:
oops, I forgot that https://github.com/zarr-developers/zarr-python/pull/3304/ also adds a `types` module. ignore this PR until #3304 is sorted out

maxrjones:
I didn't see this PR when opening #3886.

> that might be the first step to refining this PR -- we should figure out exactly which types are public. 

This was the goal of #3882 - you can see the private types used in public functions here - https://github.com/zarr-developers/zarr-python/actions/runs/24146799678/job/70462721582.

one difference between this PR and  #3886 is that #3886 modified the tests to import from the public API, which might help assuage some of @dstansby's concerns.

I originally tried an approach similar to that suggested in https://github.com/zarr-developers/zarr-python/pull/3355#issuecomment-3174511460, but found that led to circular dependencies. I think it works best for the public API to export types from the private API and for the private API to never use the public API directly.

PR #3359, opened 2025-08-08
Expose LatencyStore as public API

TomNicholas:
Closes ##3358

TODO:

### DIFF
--- a/src/zarr/storage/__init__.py
+++ b/src/zarr/storage/__init__.py
@@ -6,6 +6,7 @@ from typing import Any
 from zarr.errors import ZarrDeprecationWarning
 from zarr.storage._common import StoreLike, StorePath
 from zarr.storage._fsspec import FsspecStore
+from zarr.storage._latency import LatencyStore
 from zarr.storage._local import LocalStore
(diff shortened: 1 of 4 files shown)


d-v-b:
pre-commit is mad about a docstring but otherwise this looks g2g

maxrjones:
I'm a bit confused about why this is necessary. It's extremely common for downstream libraries to use functions from the testing modules of libraries (e.g.,` np.testing.assert_...`, `xr.testing.assert_...`). Why do we need a different approach here? I question whether this change is worth it because putting `LatencyStore` in the same namespace at the other stores would add confusion for people who just want to use Zarr-Python, not develop downstream libraries.

TomNicholas:
> Why do we need a different approach here? I question whether this change is worth it because putting LatencyStore in the same namespace at the other stores would add confusion for people who just want to use Zarr-Python, not develop downstream libraries.

I actually agree - I think it would make more sense to expose `LatencyStore` within a `zarr.testing.storage` namespace. That still requires a PR though (which could be this one), because it's currently in `zarr.testing.store`, and not documented in the public API docs.

PR #3366, opened 2025-08-11
CacheStore containing source store and cache store

### LINKED ISSUE #3357: Add general CacheStore to Zarr 3.0
This is something that would be very helpful to have ported from [zarr-python 2](https://zarr.readthedocs.io/en/v2.18.5/api/storage.html#zarr.storage.LRUStoreCache)

ruaridhg:
Closes #3357 

Store containing 2 stores i.e. primary store (where the data is coming from) and a cache store (where you want to store the data as a cache).

Introduces the class `CacheStore`

This cache wraps any Store implementation and uses a separate Store instance as the cache backend. This provides persistent caching capabilities with time-based expiration, size-based eviction, and flexible cache storage options.

TODO:

### DIFF
--- a/src/zarr/experimental/cache_store.py
+++ b/src/zarr/experimental/cache_store.py
@@ -163,7 +163,7 @@ class CacheStore(WrapperStore[Store]):
 
             logger.debug("_evict_key: evicted key %s, freed %d bytes", key, key_size)
         except Exception:
-            logger.exception("_evict_key: failed to evict key %s: %s", key)
+            logger.exception("_evict_key: failed to evict key %s", key)
             raise  # Re-raise to signal eviction failure
 
     async def _cache_value(self, key: str, value: Buffer) -> None:


src/zarr/experimental/cache_store.py
```diff
@@ -0,0 +1,334 @@
...
+        key : str
+            The key to delete
+        """
+        logger.info("delete: deleting key %s from store", key)
+        await super().delete(key)
+        logger.info("delete: deleting key %s from cache", key)
+        await self._cache.delete(key)
+        self._remove_from_tracking(key)
+
+    def cache_info(self) -> dict[str, Any]:
```
d-v-b, line 333: it would be great to use a typeddict here, so that the return type is known

src/zarr/storage/_caching_store.py
```diff
@@ -0,0 +1,334 @@
...
+            if key in self._key_sizes:
+                del self._key_sizes[key]
+
+            # Update current size
+            self._current_size = max(0, self._current_size - key_size)
+            logger.info("_evict_key: evicted key %s from cache, size %d", key, key_size)
+        except Exception as e:
+            logger.warning("_evict_key: failed to evict key %s: %s", key, e)
+
+    def _cache_value(self, key: str, value: Any) -> None:
```
TomAugspurger: Why are we accepting arbitrary `value` here? Is it possible to scope this to just `Buffer` objects (or maybe `Buffer` and `NDBuffer`)?

src/zarr/experimental/cache_store.py
```diff
@@ -0,0 +1,334 @@
...
+    def _update_access_order(self, key: str) -> None:
+        """Update the access order for LRU tracking."""
+        if key in self._cache_order:
+            # Move to end (most recently used)
+            self._cache_order.move_to_end(key)
+
+    def _remove_from_tracking(self, key: str) -> None:
+        """Remove a key from all tracking structures."""
+        if key in self._cache_order:
+            del self._cache_order[key]
```
TomAugspurger: I don't know if this is called from multiple threads, but this could be done atomically with `self._cache_order.pop(key, None)`.

src/zarr/storage/_caching_store.py
TomAugspurger: At a glance, it looks like we only call this with `Buffer` so hopefully this is an easy fix.

src/zarr/experimental/cache_store.py
```diff
@@ -0,0 +1,334 @@
...
+
+            # Update current size
+            self._current_size = max(0, self._current_size - key_size)
+            logger.info("_evict_key: evicted key %s from cache, size %d", key, key_size)
+        except Exception as e:
+            logger.warning("_evict_key: failed to evict key %s: %s", key, e)
+
+    def _cache_value(self, key: str, value: Any) -> None:
+        """Cache a value with size tracking."""
+        value_size = buffer_size(value)
```
TomAugspurger: If we only accept `Buffer` here, then `buffer_size` can be removed hopefully.

docs/user-guide/cachingstore.rst
```diff
@@ -0,0 +1,304 @@
...
+   >>> import zarr.storage
+   >>> import numpy as np
+   >>>
+   >>> # Create a local store and a separate cache store
+   >>> source_store = zarr.storage.LocalStore('test.zarr')
+   >>> cache_store = zarr.storage.MemoryStore()  # In-memory cache
+   >>> cached_store = zarr.storage.CacheStore(
+   ...     store=source_store,
+   ...     cache_store=cache_store,
+   ...     max_size=256*1024*1024  # 256MB cache
```
normanrz: Thanks for this PR. I think it would be better to have the LRU functionality on the `cache_store` (in this example the MemoryStore). Otherwise the enclosing `CacheStore` would need to keep track of all keys and their access order in the inner store. That could be problematic if the inner store would be shared with other CacheStores or other code.

d-v-b: > That could be problematic if the inner store would be shared with other CacheStores or other code.

As long as one of the design goals is to use a regular zarr store as the caching layer, there will be nothing we can do to guard against external access to the cache store. For example, if someone uses a `LocalStore` as a cache, we can't protect the local file system from external modification. I think it's the user's responsibility to ensure that they don't use the same cache for separate `CacheStores`.

d-v-b: but our default behavior could be to create a fresh `MemoryStore`, which would be a safe default

normanrz: My main concern here is about the abstraction. The LRU fits better in the inner store than in the CacheStore, imo. There could even be an `LRUStore` that wraps a store and implements the tracking and eviction. 
The safety concern is, as you pointed out, something the user should take care of.

d-v-b: > My main concern here is about the abstraction. The LRU fits better in the inner store than in the CacheStore, imo.

That makes sense, maybe we could implement `LRUStore` as another store wrapper?

dstansby: > I think it would be better to have the LRU functionality on the cache_store (in this example the MemoryStore)

To understand, is the suggestion here that

1. There's no `CacheStore` class
2. Instead all the logic for caching is implemented on `Store`, and there is a `cache_store` property that can be set to a second store to enable caching?

d-v-b: @normanrz unless you have a concrete proposal for a refactor that would be workable in the scope of this PR, I would suggest we move forward with the PR as-is, and use experience to dial in the abstraction in a later PR.

But knowing that design here might change, I think we should introduce an "experimental" storage module, e.g. `zarr.storage.experimental`, or a top-level experimental module, `zarr.experimental`, and put this class there until we are sure that the API is final.

Thoughts? I would like to ship this important feature while also retaining the ability to safely adjust it later. An experimental module seems like a safe way to do that.

cmalinmayor: Just commenting that I'd love to have this included sooner rather than later, it will be immediately useful 🎉  Thanks @ruaridhg for taking the initiative and putting this together!

joshmoore: Nice use of `experimental` here! 🤩

d-v-b:
@ruaridhg I pushed a bunch of changes, please have a look before I merge!

src/zarr/experimental/cache_store.py
d-v-b: `buffer_size` is now removed, as we are using the `Buffer` type

src/zarr/experimental/cache_store.py
d-v-b: fixed

src/zarr/storage/_caching_store.py
d-v-b: fixed

ruaridhg:
> This is missing some doctests, but that's because we don't have doctests working (see #3500). I'm going to merge this as-is, and we can circle back with more fixes later. Thanks for this contribution @ruaridhg!

@d-v-b Thanks for making changes and merging. I'm on another project with a tight deadline so haven't had time to look at this, much appreciated!

PR #3367, opened 2025-08-11
Added ArrayNotFoundError

ruaridhg:
[Description of PR]

TODO:

### DIFF
--- a/src/zarr/errors.py
+++ b/src/zarr/errors.py
@@ -1,6 +1,7 @@
 from typing import Any
 
 __all__ = [
+    "ArrayNotFoundError",
     "BaseZarrError",
     "ContainsArrayAndGroupError",
     "ContainsArrayError",
@@ -32,6 +33,10 @@ class GroupNotFoundError(BaseZarrError, FileNotFoundError):
     """
 
     _msg = "No group found in store {!r} at path {!r}"
+
+
+class ArrayNotFoundError(BaseZarrError):
+    _msg = "array not found at path %r' {0!r}"
 
 
 class ContainsGroupError(BaseZarrError):
--- a/src/zarr/api/asynchronous.py
+++ b/src/zarr/api/asynchronous.py
@@ -40,6 +40,7 @@ from zarr.core.group import (
 )
 from zarr.core.metadata import ArrayMetadataDict, ArrayV2Metadata, ArrayV3Metadata
 from zarr.errors import (
+    ArrayNotFoundError,
     GroupNotFoundError,
     NodeTypeValidationError,
     ZarrDeprecationWarning,
@@ -1257,7 +1258,7 @@ async def open_array(
 
     try:
         return await AsyncArray.open(store_path, zarr_format=zarr_format)
-    except FileNotFoundError:
+    except FileNotFoundError as err:
         if not store_path.read_only and mode in _CREATE_MODES:
             overwrite = _infer_overwrite(mode)
             _zarr_format = zarr_format or _default_zarr_format()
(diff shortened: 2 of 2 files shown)


d-v-b:
thank you, this looks good. we should make sure we cover _all_ the places in the codebase where we try to read an array. maybe it's just the 1 place you found, but we should double-check.

tests/test_api.py
```diff
@@ -161,7 +166,7 @@ async def test_open_array(memory_store: MemoryStore, zarr_format: ZarrFormat) ->
     assert z.read_only
 
     # path not found
-    with pytest.raises(FileNotFoundError):
+    with pytest.raises(ArrayNotFoundError):
```
dstansby: This is currently failing the tests, because it actually raises a `GroupNotFoundError` now. I think the sensible thing to do here would be to create another new error, `NodeNotFoundError`, that is raised when a non-specific function is used (e.g., just `open`) to try and open either an array or a group.

src/zarr/errors.py
```diff
@@ -34,6 +35,10 @@ class GroupNotFoundError(BaseZarrError, FileNotFoundError):
     _msg = "No group found in store {!r} at path {!r}"
 
 
+class ArrayNotFoundError(BaseZarrError):
```
dstansby: This should also sub-class FileNotFoundError to make sure we retain backwards compatibility.

dstansby (changes requested):
Two inline comments. In addition, did you thoroughly check that there's only a single place where an `ArrayNotFoundError` should be raised?

d-v-b:
it might be helpful to check out https://github.com/zarr-developers/zarr-python/pull/3012, which unfortunately was too bloated to merge

tests/test_api.py
```diff
@@ -185,10 +192,27 @@ async def test_open_array(memory_store: MemoryStore, zarr_format: ZarrFormat) ->
     assert z.read_only
 
     # path not found
-    with pytest.raises(FileNotFoundError):
+    with pytest.raises((NodeNotFoundError, GroupNotFoundError)):
```
d-v-b: how is this raising two errors?

src/zarr/errors.py
```diff
@@ -52,14 +57,19 @@ def __init__(self, *args: Any) -> None:
...
-class NodeNotFoundError(PathNotFoundError):
-    """Raised when an array or group does not exist at a certain path."""
+    def __init__(self, *args: Any) -> None:
+        if len(args) == 1:
+            # Pre-formatted message
+            super(BaseZarrError, self).__init__(args[0])
+        else:
+            # Store and path arguments - format them
+            _msg = "No group found in store {!r} at path {!r}"
+            super(BaseZarrError, self).__init__(_msg.format(*args))
```
d-v-b, line 72: in a separate PR we should remove the pre-formatting of exception messages, and make every exception just take a single string

tests/test_api.py
ruaridhg: Yeah fair point, I was getting errors in my tests with only 1 error raised but now I've changed it so `GroupNotFoundError` and `ArrayNotFoundError` inherit from `NodeNotFoundError`

changes/3367.bugfix.rst
```diff
@@ -0,0 +1 @@
+Added `~zarr.errors.ArrayNotFoundError`, which is raised when attempting to open a zarr array that does not exist.
```
d-v-b:
```suggestion
Added `zarr.errors.ArrayNotFoundError`, which is raised when attempting to open a zarr array that does not exist, and `zarr.errors.NodeNotFoundError`, which is raised when failing to open an array or a group in a context where either an array or a group was expected.
```

PR #3369, opened 2025-08-11
feat: Implement ZEP 8 URL syntax support for zarr-python

### LINKED ISSUE #2831: Can't conveniently open zip store from path with zarr v3

### LINKED ISSUE #2943: Support ZEP 8 URL Syntax
### Summary

ZEP 8 (https://github.com/zarr-developers/zeps/pull/48) proposes a standardized URL syntax for specifying Zarr nodes across multiple implementations. Implementing this in `zarr-python` would enable more seamless interoperability between different Zarr tools, making it easier to share dataset locations across different backends.

### Motivation

- A standardized URL syntax ensures consistent dataset access across different implementations.
- The proposed syntax allows for nested storage mechanisms, including ZIP files and hierarchical Zarr groups.
- Improved integration with libraries like `xarray`, which could leverage this for seamless dataset loading.

### Proposed Implementation

I would encourage us to handle this in two phases.

#### Phase 1

Implement handling of ZEP8 URLs for stores implemented in Zarr-Python. Practically, this requires parsing URLs into _Store_, _Path_, and _ZarrFormat_ components.

#### Phase 2

Implement an entrypoint mechanism for 3rd party stores (e.g. Icechunk). This would allow external Store developers to provide a function to generate a Store given a URL.

### Example Usage in xarray

With ZEP 8 URL syntax, an `xarray.Dataset` could be opened as follows:

```python
import xarray as xr

dataset = xr.open_dataset(
    "gs://my-bucket/my-archive.zip|zip:inner-dir|zarr3:subgroup",
    engine="zarr"
)
```

This URL specifies:

- A ZipStore (`my-archive.zip`) stored in Google Cloud Storage (`gs://my-bucket/`)
- A nested hierarchy inside the Zarr store (`subgroup`)

Another example: 

```python
dataset = xr.open_dataset(
    "https://foobar.com/my-archive|icechunk|zarr3",
    engine="zarr"
)
```

This URL specifies:

- An Icechunk Store available over https

### Additional Considerations

- Backward compatibility with existing `fsspec` URLs.

cc @jbms

xref: https://github.com/earth-mover/icechunk/issues/576, https://github.com/zarr-developers/zeps/pull/48

jhamman:
This PR implements support for the ZEP 8 URL syntax in Zarr Python.

Some examples of what now works:

```python
import zarr

root = zarr.open_group('s3://bucket/data.zip|zip:|zarr3:')  # S3 → ZIP → Zarr v3
arr = zarr.create_array('memory:|zarr2:group/array', shape=(10, ), dtype='i4')  # Memory → Zarr v2

# custom adapter for icechunk
ds = xr.open_zarr('s3://icechunk-public-data/v1/glad|icechunk:')  # icechunk (from xarray)
```

TODO:

closes #2943
fixes #2831
xref: https://github.com/zarr-developers/zeps/pull/48

cc @jbms


src/zarr/storage/_builtin_adapters.py
```diff
@@ -0,0 +1,458 @@
...
+
+
+class S3Adapter(RemoteAdapter):
+    """Store adapter for S3 URLs using fsspec."""
+
+    adapter_name = "s3"
+
+    @classmethod
+    def get_supported_schemes(cls) -> list[str]:
+        return ["s3"]
```
martindurant: Historically, "s3a" has also existed. I think it was a spark/hadoop implementation detail (i.e,. another codebase/driver for the same storage backend), so it might not be relevant any more, but I thought I would mention it.

https://stackoverflow.com/a/33356421/3821154

src/zarr/storage/_builtin_adapters.py
```diff
@@ -0,0 +1,458 @@
...
+
+
+class GCSAdapter(RemoteAdapter):
+    """Store adapter for Google Cloud Storage URLs using fsspec."""
+
+    adapter_name = "gcs"
+
+    @classmethod
+    def get_supported_schemes(cls) -> list[str]:
```
jbms, line 337: It would be better, I think, to avoid supporting "gcs" in addition to "gs" since that will create implementation divergence.

src/zarr/storage/_builtin_adapters.py
```diff
@@ -0,0 +1,458 @@
...
+    """Store adapter for HTTP(S) URLs using fsspec."""
+
+    adapter_name = "https"
+
+    @classmethod
+    def get_supported_schemes(cls) -> list[str]:
+        return ["http", "https"]
+
+
+class S3Adapter(RemoteAdapter):
```
jbms, line 225: Neuroglancer also supports "s3+http://endpoint/path" and "s3+https://endpoint/path" syntax to allow custom s3 endpoints (rather than AWS) to be used.  The plain "s3://" syntax always refers to the real AWS endpoints, not a custom s3-compatible server.

src/zarr/storage/_builtin_adapters.py
jhamman: Interesting! Seeing as I've never seen thi, I'm going to ignore it until someone asks.

src/zarr/storage/_builtin_adapters.py
jhamman: great suggestion. Added in [c8941ba](https://github.com/zarr-developers/zarr-python/pull/3369/commits/c8941baf94fb42d60b62f2ca3bda710176a20bc9)

src/zarr/storage/_builtin_adapters.py
jhamman: done in [3b55281](https://github.com/zarr-developers/zarr-python/pull/3369/commits/3b552812052bccf7c2d904652e3d38104ce4bd4c)

jbms:
One tricky thing about `s3+https://endpoint/a/b` is that it is ambiguous as to whether it is using "virtual host" syntax (i.e. `endpoint` refers to a single bucket and the path is "a/b") or "path" syntax (i.e. the bucket is "a" and the path is "b").

The "path" syntax is generally the default when running a regular s3-compatible server, but the "virtual host" syntax can commonly occur when someone defines a CNAME DNS entry to map their own domain or subdomain to an AWS S3 bucket.

When designing this syntax for Neuroglancer, it seemed like it would be annoying to require users to use separate syntax to disambiguate the two cases.  Instead, for operations where it matters (namely List), Neuroglancer just automatically determines which of the two cases applies by trying both ways and seeing which one succeeds, and then caching the result so that subsequent list operations don't require two requests.

changes/3369.feature.rst
```diff
@@ -0,0 +1,27 @@
...
+
+* **URL syntax support**: Use pipe (``|``) characters to chain storage adapters, e.g., ``file:data.zip|zip``
+* **Built-in adapters**: Support for ``file``, ``memory``, ``zip``, ``s3``, ``https``, ``gcs`` schemes
+* **Store adapter registry**: New ``zarr.registry.list_store_adapters()`` function to discover available adapters
+* **Extensible architecture**: Custom store adapters can be registered via entry points
+
+Examples::
+
+    # Basic ZIP file storage
+    zarr.open_array("file:data.zip|zip", mode='w', shape=(10, 10), dtype="f4")
```
laramiel, line 16: It seems like a bad idea to break file uris.

https://datatracker.ietf.org/doc/html/rfc8089

ianhi: This is considered in the spec doc this PR is building on: https://github.com/zarr-developers/zeps/pull/48

https://github.com/jbms/zeps/blob/92bc64111c7612083560358efdd4450e061f3746/draft/ZEP0008.md?plain=1#L115-L119


And later is says:

```
  Implementations SHOULD not support `file://relative/path` since that
  is ambiguous with the `file://hostname/path` syntax defined by
  [RFC8089](https://datatracker.ietf.org/doc/html/rfc8089).
```


If you forsee serious issues here I'd encourage commenting on that PR on the standard.

docs/user-guide/storage.rst
```diff
@@ -145,6 +145,87 @@ Here's an example of using ObjectStore for accessing remote data:
...
+
+**Remote ZIP files:**
+
+   >>> # Access ZIP file from S3 (requires s3fs)
+   >>> zarr.open_array("s3://bucket/data.zip|zip", mode='r')  # doctest: +SKIP
+
+**In-memory storage:**
+
+   >>> # Create array in memory
+   >>> z = zarr.open_array("memory:", mode='w', shape=(5, 5), dtype="f4")
```
ianhi, line 186: Can I then access this from somewhere else using this syntax?

e.g. `memory:aesr80s9e8ra`?

docs/user-guide/storage.rst
```diff
@@ -145,6 +145,87 @@ Here's an example of using ObjectStore for accessing remote data:
 .. warning::
    The :class:`zarr.storage.ObjectStore` class is experimental.
 
+URL-based Storage (ZEP 8)
```
ianhi, line 148: I think what this section is missing is a showcasing of what the equivalent zarr-pyhthon code would be to put it in terms people are more familiar with. 

So each section would be:


```python
zarr.open_array("file:zep8-data.zip|zip" ....)

# is equivalent to

zarr.open_array(zarr.storage.ZipStore(...)...)
```

docs/user-guide/storage.rst
```diff
@@ -145,6 +145,87 @@ Here's an example of using ObjectStore for accessing remote data:
...
+
+   >>> import zarr
+   >>> zarr.registry.list_store_adapters()  # doctest: +SKIP
+   ['file', 'gcs', 'gs', 'https', 'memory', 's3', 'zip', ...]
+
+Additional adapters can be implemented as described in the `extending guide <./extending.html#custom-store-adapters>`_.
+
+.. note::
+   When using ZEP 8 URLs with third-party libraries like xarray, the URL syntax allows
+   seamless integration without requiring zarr-specific store creation.
```
ianhi, line 225:
```suggestion
```

This is already effectively stated above.

docs/quickstart.rst
```diff
@@ -198,6 +198,28 @@ To open an existing array from a ZIP file::
         [0.4335856 , 0.7565437 , 0.7828931 , ..., 0.48119593, 0.66220033,
             0.6652362 ]], shape=(100, 100), dtype=float32)
 
+URL-based Storage (ZEP 8)
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Zarr supports URL-based storage following the ZEP 8 specification, which allows you to specify storage locations using URLs with chained adapters::
```
ianhi, line 204: Needs a link to the more extensive docs.

docs/quickstart.rst
```diff
@@ -198,6 +198,28 @@ To open an existing array from a ZIP file::
...
+    >>> # Read it back
+    >>> z2 = zarr.open_array("file:data/example-zep8.zip|zip", mode='r')
+    >>> z2.shape
+    (50, 50)
+
+ZEP 8 URLs use pipe (``|``) characters to chain storage adapters together:
+
+- ``file:path.zip|zip`` - ZIP file on local filesystem
+- ``s3://bucket/data.zip|zip`` - ZIP file in S3 bucket
+- ``memory:`` - In-memory storage
```
ianhi, line 219:
```suggestion
```
not an example of piping

docs/quickstart.rst
```diff
@@ -198,6 +198,28 @@ To open an existing array from a ZIP file::
...
+    >>>
+    >>> # Read it back
+    >>> z2 = zarr.open_array("file:data/example-zep8.zip|zip", mode='r')
+    >>> z2.shape
+    (50, 50)
+
+ZEP 8 URLs use pipe (``|``) characters to chain storage adapters together:
+
+- ``file:path.zip|zip`` - ZIP file on local filesystem
+- ``s3://bucket/data.zip|zip`` - ZIP file in S3 bucket
```
ianhi, line 218: we should have a more deeply nested thing, to actually show chaining of adapters.

docs/quickstart.rst
```diff
@@ -198,6 +198,28 @@ To open an existing array from a ZIP file::
...
+- ``memory:`` - In-memory storage
+
+This provides a concise way to specify complex storage configurations without explicitly creating store objects.
```
ianhi, line 221: move to first paragraph of seciton

src/zarr/abc/store_adapter.py
```diff
@@ -0,0 +1,196 @@
...
+
+    def __post_init__(self) -> None:
+        """Validate the URL segment."""
+        import re
+
+        from zarr.storage._zep8 import ZEP8URLError
+
+        if not self.scheme and not self.adapter:
+            raise ZEP8URLError("URL segment must have either scheme or adapter")
+        if self.adapter and not re.match(r"^[a-zA-Z][a-zA-Z0-9_+-]*$", self.adapter):
```
ianhi, line 50: where is this regex coming from? It seems as though i could make `zip_` but not `_zip` which seems arbitrary

src/zarr/api/asynchronous.py
```diff
@@ -58,9 +58,9 @@
     from zarr.core.chunk_key_encodings import ChunkKeyEncoding
     from zarr.storage import StoreLike
 
-    # TODO: this type could use some more thought
-    ArrayLike = AsyncArray[ArrayV2Metadata] | AsyncArray[ArrayV3Metadata] | Array | npt.NDArray[Any]
-    PathLike = str
+
+ArrayLike = AsyncArray[ArrayV2Metadata] | AsyncArray[ArrayV3Metadata] | Array | npt.NDArray[Any]
+PathLike = str
```
ianhi, line 63: Is this a claude hallucination that this ok now, or has something concretely changed?

src/zarr/storage/_builtin_adapters.py
```diff
@@ -0,0 +1,562 @@
...
+from pathlib import Path
+from typing import TYPE_CHECKING, Literal, cast
+from urllib.parse import parse_qs, urlparse
+
+from zarr.abc.store_adapter import StoreAdapter
+from zarr.storage._local import LocalStore
+from zarr.storage._logging import LoggingStore
+from zarr.storage._memory import MemoryStore
+from zarr.storage._zep8 import URLStoreResolver
+from zarr.storage._zip import ZipStore
```
ianhi, line 19: Is it worth considering makign these lazy, or not a big enough gain?

src/zarr/storage/_zep8.py
```diff
@@ -0,0 +1,661 @@
...
+    Examples
+    --------
+    >>> is_zep8_url("s3://bucket/data.zip|zip:|zarr3:")
+    True
+    >>> is_zep8_url("memory:")
+    True
+    >>> is_zep8_url("s3://bucket/data.zarr")
+    False
+    >>> is_zep8_url("file:///data.zarr")
+    False
```
ianhi, line 221: I don't really follow these returning false. seems like a downgrade in functionality, also having looked a the spec a bit I don't see where this is explicitly disallowed (though very possible i misread or misunderstood)

src/zarr/storage/_zep8.py
```diff
@@ -0,0 +1,661 @@
...
+        KeyError
+            If a required store adapter is not registered.
+        """
+        # Handle simple scheme URLs (like file:/path, s3://bucket/path) by treating them as single-segment URLs
+        if not is_zep8_url(url):
+            # Check if it's a simple scheme URL that we can handle
+            if "://" in url or ((":" in url) and not url.startswith("/")):
+                # Parse as a single segment URL - the parser should handle this
+                try:
+                    segments = self.parser.parse(url)
```
ianhi, line 348: Ahh I think this answers my question above. Since these are valid urls I think ideally is_zep8_url would handle these simple cases correctly.

src/zarr/storage/_zep8.py
```diff
@@ -0,0 +1,661 @@
...
+                except Exception:
+                    raise ValueError(f"Not a valid URL: {url}") from None
```
ianhi, line 350: What are the expected exception's we are catching here. As written this might silently quash an exception that is a real bug.

I think this whole section can be consolidated and simplified a bit, especially as we don't actually do anyhthing differently here then in the `else` branch that calls the same parse

src/zarr/storage/_zep8.py
```diff
@@ -0,0 +1,661 @@
...
+
+        # Process segments in order, building preceding URL for each adapter
+        current_store: Store | None = None
+
+        # Build list of segments that create stores (excluding zarr format segments and path-only segments)
+        store_segments = []
+        for i, segment in enumerate(segments):
+            if segment.adapter in ("zarr2", "zarr3"):
+                # Skip zarr format segments - they don't create stores
+                # TODO: these should propagate to the open call somehow
```
ianhi, line 368: Calling out this TODO as importnat before merging

ianhi (commented):
Partway through the code at this point (currently following the logic of store resolution). Posting some comments now so I don't lose them (i've been burned by this before)

docs/quickstart.rst
```diff
@@ -198,6 +198,28 @@ To open an existing array from a ZIP file::
         [0.4335856 , 0.7565437 , 0.7828931 , ..., 0.48119593, 0.66220033,
             0.6652362 ]], shape=(100, 100), dtype=float32)
 
+URL-based Storage (ZEP 8)
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Zarr supports URL-based storage following the ZEP 8 specification, which allows you to specify storage locations using URLs with chained adapters::
+
+    >>> # Store data directly in a ZIP file using ZEP 8 URL syntax
+    >>> z = zarr.open_array("file:data/example-zep8.zip|zip", mode='w', shape=(50, 50), chunks=(10, 10), dtype="f4")
```
ianhi, line 207: should probably be adding a .zarr in here somewhere. so we aren't creating otherwise unopenable zip files

src/zarr/abc/store_adapter.py
```diff
@@ -0,0 +1,196 @@
...
+    """The store adapter name (e.g., 'zip', 'icechunk', 'zarr3')."""
+
+    path: str = ""
+    """Path component for the segment."""
+
+    def __post_init__(self) -> None:
+        """Validate the URL segment."""
+        import re
+
+        from zarr.storage._zep8 import ZEP8URLError
```
ianhi, line 46: Let's move this to errors.py to avoid circular import shenanigans

src/zarr/storage/_zep8.py
```diff
@@ -0,0 +1,661 @@
...
+
+
+class ZEP8URLError(ValueError):
+    """Exception raised for invalid ZEP 8 URL syntax."""
+
+
+class URLParser:
+    """Parse ZEP 8 URL syntax into components."""
+
+    def parse(self, url: str) -> list[URLSegment]:
```
ianhi, line 37: This is currently failing to error on what ought to be invalid URLs with multiple schemes. 

```
z = zarr.open_array("file:memory:example.zip|zip", mode='w', shape=(100, 100), chunks=(10, 10), dtype="i4")
```

executes and creates teh file: `memory:example.zip`


I think this is prohibited according to the rules here: https://zeps--48.org.readthedocs.build/en/48/draft/ZEP0008.html#absolute-url-syntax

ianhi: Also 
```
z = zarr.open_array("memory:memory:", mode='w', shape=(100, 100), chunks=(10, 10), dtype="i4")
```

creates a standard zarr folder on disk named `memory:memory:`

src/zarr/storage/_builtin_adapters.py
```diff
@@ -0,0 +1,562 @@
...
+    ) -> Store:
+        """Create a LocalStore from a file URL segment."""
+        # For file scheme, the preceding_url should be the full file: URL
+        if not preceding_url.startswith("file:"):
+            raise ValueError(f"Expected file: URL, got: {preceding_url}")
+
+        # Extract path from preceding URL
+        path = preceding_url[5:]  # Remove 'file:' prefix
+        if not path:
+            path = "."
```
ianhi, line 58: > For security, implementation SHOULD place limits on where this scheme is permitted.

refering to `..` which seems important here to not expose unlimited things in the filesystem.

ianhi: not really sure how to get into the business of defining what is safe here. But flagging for consideration

src/zarr/storage/_zep8.py
```diff
@@ -0,0 +1,661 @@
...
+                # Both old format "branch:main", "tag:v1.0", "snapshot:abc123"
+                # and new format "@branch.main", "@tag.v1.0", "@abc123def456"
+                if segment.adapter in ("icechunk", "ic"):
+                    # Check old format: branch:main, tag:v1.0, snapshot:abc123
+                    if ":" in segment.path and segment.path.split(":")[0] in (
+                        "branch",
+                        "tag",
+                        "snapshot",
+                    ):  # pragma: no cover
+                        continue  # Skip icechunk metadata paths  # pragma: no cover
```
ianhi, line 570: What's going on here? what is the old vs new?

can this all just be removed an handled by icechunk?

src/zarr/storage/_zep8.py
```diff
@@ -0,0 +1,661 @@
...
+                                    adapter_path = zarr_path_component  # pragma: no cover
+                                continue  # pragma: no cover
+                            # Fallback: if starts with @ and has /, extract part after first /
+                            if "/" in segment.path:
+                                _, path_part = segment.path.split("/", 1)
+                                adapter_path = path_part
+                            continue
+                        except Exception:  # pragma: no cover
+                            # If parsing fails, treat as regular path  # pragma: no cover
+                            pass  # pragma: no cover
```
ianhi, line 606: I think we should **not** do this. Or at least be much more careful about what we pass through. e.g. my `file:memory:` example I think should fail even if it's techinically valid to have a colon in a file name, i think this implementation will just make that illegal for zarr users perhaps?

ianhi: but we would need an escape hatch for whoever is already doing that :/

src/zarr/storage/_zep8.py
ianhi: Indeed this is crucial functionality:

```python
import zarr

arr = zarr.open_array("memory:|zarr2:", mode='w', shape=(5,), dtype='i4')

print(f"Array: {arr}")
print(f"Format version: {arr.metadata.zarr_format}")

```
```
Array: <Array memory://4411020480 shape=(5,) dtype=int32>
Format version: 3
```

ianhi (commented):
## Accidentally submitted too early!

Ack! Hit submit instead of a different button! Below is still being filled out! I will comment when properly done.


---
Here are some edge cases I ran into:


Throws a FileNotFoundError:
Not sure what it should do, but I think this needs a Zep8 error 
```python
import zarr
import numpy as np
import tempfile
from pathlib import Path

tmpdir = Path(tempfile.mkdtemp())
zip_path = tmpdir / "mydata.zip"

arr_write = zarr.open_array(
    f"file:{zip_path}|zip:",
    mode='w',
    shape=(10, 10),
    chunks=(5, 5),
    dtype='f4'
)

arr_write[:, :] = np.random.random((10, 10))

# throws FileNotFoundError
arr_wrong = zarr.open_array(
    f"file:{zip_path}",  # Missing |zip:
    mode='r'
)
```



### Accidentally missing an adapter. 

I think an absolutely killer feature here is to have automatic suggestions for how to correct a URL (like the python repr does for syntax errors)

e.g. 

```python
    arr_wrong3 = zarr.open_array(
        f"file:{zip_path}|log:",  # Missing |zip:
        mode='r'
    )
    print(f"Got data: {arr_wrong3[:]}")
```

gives

`ArrayNotFoundError: No array found in store logging-file:///var/folders/tc/fkgp35zn7z913f9cmsxcl6pc0000gn/T/tmpxp7e5hc3/logged_data.zip at path `

I would love to have a: "did you forget a `|zip:` as part of that error message


## Nested Zips/nested adapters

I tried to replicate a URL from the spec locally of nested zip files. It simplifies to this:

```python
import zarr, zipfile, tempfile
from pathlib import Path

# Create inner.zip with data
tmpdir = Path(tempfile.mkdtemp())
inner = tmpdir / "inner.zip"
arr = zarr.open_array(f"file:{inner}|zip:", mode='w', shape=(3,), dtype='i4')
arr[:] = [1, 2, 3]
del arr  # Close

# Put inner.zip inside outer.zip
outer = tmpdir / "outer.zip"
with zipfile.ZipFile(outer, 'w') as zf:
    zf.write(inner, "inner.zip")

# Try nested read - THIS FAILS
url = f"file:{outer}|zip:inner.zip|zip:"
arr_read = zarr.open_array(url, mode='r')  # ❌ FileNotFoundError
```

```
FileNotFoundError: [Errno 2] No such file or directory: '[/var/folders/tc/fkgp35zn7z913f9cmsxcl6pc0000gn/T/tmp9kf2z2ks/outer.zip](http://localhost:8888/var/folders/tc/fkgp35zn7z913f9cmsxcl6pc0000gn/T/tmp9kf2z2ks/outer.zip)|zip:inner.zip'
```

somewhere in the URL reconstruction we lose the zep8 ness and try to just open the whole string as a file

src/zarr/storage/_register_adapters.py
```diff
@@ -0,0 +1,50 @@
...
+    from zarr.storage._builtin_adapters import (
+        FileSystemAdapter,
+        GSAdapter,
+        HttpsAdapter,
+        LoggingAdapter,
+        MemoryAdapter,
+        S3Adapter,
+        S3HttpAdapter,
+        S3HttpsAdapter,
+        ZipAdapter,
```
ianhi, line 27: We should add `LatencyStore` to this list with some sensible defaults it would be super helpful for testing

ianhi (commented):
Hitting shift enter sending the review off is absolutely killing me. Sorry I keep opening this too early....

A weird one. Adding a log store via zep8 breaks some S3 permissions, but does not if you open with explicit stores

```python

import zarr
from zarr.storage import FsspecStore, LoggingStore

storage_options = {'anon': True}


# WORKS!
s3_store = FsspecStore.from_url(
    "s3://noaa-nwm-retro-v2-zarr-pds",
    storage_options=storage_options
)
logging_store = LoggingStore(s3_store)
group = zarr.open_group(logging_store)
keys = list(group.keys())
print(f"Keys: {keys[:3]}")


# DIES WITH PERMISSION ERROR
url = "s3://noaa-nwm-retro-v2-zarr-pds|log:"
group = zarr.open_group(url, storage_options=storage_options)
keys = list(group.keys())
print(f"Keys: {keys[:3]}")
```

i suspect that this is related the storage_options not being passed through correctly

src/zarr/storage/_zep8.py
```diff
@@ -0,0 +1,661 @@
...
+                    elif prev_segment.adapter:
+                        # Adapter segment - reconstruct format
+                        preceding_parts.append(f"{prev_segment.adapter}:{prev_segment.path}")
+
+                preceding_url = "|".join(preceding_parts)
+
+            # Create the store using the adapter with preceding URL
+            store_kwargs = kwargs.copy()
+            if storage_options:
+                store_kwargs.update(storage_options)
```
ianhi, line 459: This effectively unpacks the storage_options dict which I think is what breaks my example with S3+ log store.

needs somethign like:
```suggestion
            if storage_options:
                store_kwargs['storage_options'] = storage_options
```

src/zarr/storage/_builtin_adapters.py
```diff
@@ -0,0 +1,562 @@
...
+    Supports:
+    - Standard AWS S3: s3://bucket/path
+    - Custom S3 endpoints: s3+http://endpoint/bucket/path, s3+https://endpoint/bucket/path
+    """
+
+    adapter_name = "s3"
+
+    @classmethod
+    def get_supported_schemes(cls) -> list[str]:
+        return ["s3", "s3+http", "s3+https"]
```
ianhi, line 237: for the `s3+` urls we aren't reaching here unfortunately:

```python
zarr.open(
  "s3+https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.5/idr0062A/6001240_labels.zarr",
  storage_options={'anon': True}
)
```
```
ValueError: Protocol not known: s3+https
```

i think this relates to the prior mentioned issues with `is_zep8_url`

ianhi (commented):
I've started making some changes on a branch off of this. fixing the issues i found.

One architectural change is that in this PR as stands URLs can either go directly into fsspec (as a special case) or go through zep8 to end up in fsspec. I think it makes more sense to route all URLs through zep8 machinery. And possibly will for local stores.

I also think i found a bug with the inteded usage of `local://` bypassing fsspec stores in `is_fsspec_uri` the second part of the if condition could never be triggered

Although 
---
A weird one. Adding a log store via zep8 breaks some S3 permissions, but does not if you open with explicit stores

```python

import zarr
from zarr.storage import FsspecStore, LoggingStore

storage_options = {'anon': True}


# WORKS!
s3_store = FsspecStore.from_url(
    "s3://noaa-nwm-retro-v2-zarr-pds",
    storage_options=storage_options
)
logging_store = LoggingStore(s3_store)
group = zarr.open_group(logging_store)
keys = list(group.keys())
print(f"Keys: {keys[:3]}")


# DIES WITH PERMISSION ERROR
url = "s3://noaa-nwm-retro-v2-zarr-pds|log:"
group = zarr.open_group(url, storage_options=storage_options)
keys = list(group.keys())
print(f"Keys: {keys[:3]}")



```

jbms:
Note: I have moved the URL pipeline proposal over from the ZEP repo to a separate repository:

https://github.com/jbms/url-pipeline

PR #3374, opened 2025-08-13
chore/remove chunk coords

d-v-b:
replaces all occurences of `ChunkCoords` with `tuple[int, ...]`

### DIFF
--- a/src/zarr/core/common.py
+++ b/src/zarr/core/common.py
@@ -36,9 +36,7 @@ ZATTRS_JSON = ".zattrs"
 ZMETADATA_V2_JSON = ".zmetadata"
 
 BytesLike = bytes | bytearray | memoryview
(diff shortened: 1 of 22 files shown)


src/zarr/core/common.py
```diff
@@ -36,9 +36,7 @@
 ZMETADATA_V2_JSON = ".zmetadata"
 
 BytesLike = bytes | bytearray | memoryview
-ShapeLike = tuple[int, ...] | int
-ChunkCoords = tuple[int, ...]
-ChunkCoordsLike = Iterable[int]
+ShapeLike = Iterable[int] | int
```
d-v-b, line 39: Note this change, which better matches the intent of `ShapeLike`.

TomAugspurger (approved):

PR #3376, opened 2025-08-13
chore/handle numcodecs codecs

d-v-b:
This PR brings in all the codecs defined in `numcodecs.zarr3`. After this PR is merged, we can safely replace the `numcodecs.zarr3` module with _reexports_ from zarr python, or remove `numcodecs.zarr3` entirely, thereby fixing our circular dependency problem.

This PR also changes the default config to ensure that the locally-defined codecs take priority over the same codec found in the numcodecs registry.


docs/user-guide/config.rst
```diff
@@ -42,23 +42,43 @@ requires the value of ``codecs.bytes.name`` to be ``'custompackage.NewBytesCodec
...
+    'async': {'concurrency': 10, 'timeout': None},
+    'buffer': 'zarr.buffer.cpu.Buffer',
+    'codec_pipeline': {'batch_size': 1,
+                       'path': 'zarr.core.codec_pipeline.BatchedCodecPipeline'},
+    'codecs': {'blosc': 'zarr.codecs.blosc.BloscCodec',
                'bytes': 'zarr.codecs.bytes.BytesCodec',
                'crc32c': 'zarr.codecs.crc32c_.Crc32cCodec',
                'endian': 'zarr.codecs.bytes.BytesCodec',
                'gzip': 'zarr.codecs.gzip.GzipCodec',
+               'numcodecs.adler32': 'zarr.codecs._numcodecs.Adler32',
```
TomAugspurger: Can we reexport these classes at `zarr.codecs.numcodecs`, so that we don't have the private `_numcodecs` module in the config here?

Easiest way is probably a new directory `zarr/codecs/numcodecs` and then move `_numcodecs.py` to `_codecs.py` in there, and then a `numcodecs/__init__.py` that imports all the codecs and reexports them.

src/zarr/codecs/_numcodecs.py
```diff
@@ -0,0 +1,374 @@
...
+        return await asyncio.to_thread(
+            as_numpy_array_wrapper,
+            self._codec.decode,
+            chunk_data,
+            chunk_spec.prototype,
+        )
+
+    def _encode(self, chunk_data: Buffer, prototype: BufferPrototype) -> Buffer:
+        encoded = self._codec.encode(chunk_data.as_array_like())
+        if isinstance(encoded, np.ndarray):  # Required for checksum codecs
```
TomAugspurger: Can we know statically which are checksum codecs without the isinstance check?

src/zarr/codecs/_numcodecs.py
```diff
@@ -0,0 +1,374 @@
...
+    id = _expect_name_prefix(parsed_name)
+    return {"id": id, **parsed_configuration}
+
+
+@dataclass(frozen=True)
+class _NumcodecsCodec(Metadata):
+    codec_name: str
+    codec_config: dict[str, JSON]
+
+    def __init_subclass__(cls, *, codec_name: str | None = None, **kwargs: Any) -> None:
```
TomAugspurger: What would a codec definition look like without this magic? I'd be fine with repeating a few things if it meant we could avoid this (and IIUC some of the complexity in `__repr__` and `__init__` would go away too?).

src/zarr/codecs/_numcodecs.py
```diff
@@ -0,0 +1,374 @@
...
+    def evolve_from_array_spec(self, array_spec: ArraySpec) -> FixedScaleOffset:
+        if self.codec_config.get("dtype") is None:
+            dtype = array_spec.dtype.to_native_dtype()
+            return FixedScaleOffset(**{**self.codec_config, "dtype": str(dtype)})
+        return self
+
+
+class Quantize(_NumcodecsArrayArrayCodec, codec_name="quantize"):
+    def __init__(self, **codec_config: JSON) -> None:
+        super().__init__(**codec_config)
```
TomAugspurger: What's this do?

tests/test_codecs/test_numcodecs.py
```diff
@@ -22,3 +68,282 @@ def test_is_numcodec_cls() -> None:
...
+    a[:, :] = data.copy()
+    with codec_conf():
+        with pytest.warns(ZarrUserWarning, match=EXPECTED_WARNING_STR):
+            b = open_array(a.store, mode="r")
+    np.testing.assert_array_equal(data, b[:, :])
+
+
+@pytest.mark.parametrize("codec_class", [_numcodecs.PCodec, _numcodecs.ZFPY])
+def test_generic_bytes_codec(codec_class: type[_numcodecs._NumcodecsArrayBytesCodec]) -> None:
+    try:
```
TomAugspurger, line 262: Can you explain the different cases here? Do the pcodec zfpy codecs depend on optional numcodesc dependencies? And if so, is that the only reason we might not be able to run this test?

If so, can we maybe do `pytest.importorskip(dependency_name)` and then assume we have it and avoid the xfails.

TomAugspurger: Oh, I guess we (or numcodecs?) raises a `ValueError`...? If that's numcodecs, then whatever. But I don't think we should raise a ValueError if a dependency is missing.

src/zarr/codecs/_numcodecs.py
d-v-b: n.b., this was copy + pasted from numcodecs, but I think the answer is "no"

d-v-b:
[> Can you explain the different cases here?](https://github.com/zarr-developers/zarr-python/pull/3376#discussion_r2280971426)

Spinning this question out into the main thread -- from me, the general answer to questions like this will be "no", since I am only copy+pasting stuff from numcodecs. I haven't spent too much time figuring out what this code is doing. I do think @normanrz and @TomNicholas might be able to answer some of these questions though.

docs/user-guide/config.rst
d-v-b: sounds like a good plan

TomAugspurger:
Ah, I didn't realize this was mostly from numcodecs. I think that moots most of my comments aside from where in the public API we put these.

d-v-b:
yeah I should have made more clear that this is nearly all directly copy + pasted from [`numcodecs.zarr3`](https://github.com/zarr-developers/numcodecs/blob/main/numcodecs/zarr3.py)

docs/user-guide/config.rst
d-v-b: done in https://github.com/zarr-developers/zarr-python/pull/3376/commits/233fbbfb7f9b5295fe70d443924b4d7157d1d299

d-v-b: ok https://github.com/zarr-developers/zarr-python/commit/233fbbfb7f9b5295fe70d443924b4d7157d1d299 only partially applied the fix -- I also had to change some our codec registry logic, because we were giving each codec a name in the config based on its fully-qualified name, which would be `zarr.codecs.numcodecs._codecs.XCodec`, but we want `zarr.codecs.numcodecs.XCodec`. This alone is a great indication that representing codecs in the config by their fully-qualified names is not a great pattern. Yet more evidence that we need to rethink that API.

tests/test_codecs/test_numcodecs.py
d-v-b: > If that's numcodecs, then whatever. 

Yep, it's a numcodecs thing

d-v-b:
I think this is ready to go in (and it's necessary for #3332)

d-v-b:
and important recent addition: I moved all of the invocations of `register_codec` to `src/zarr/codecs/__init__.py`. This ensures that all codecs get registered, regardless of whether they are part of `zarr.codecs.__all__`

src/zarr/codecs/_numcodecs.py
d-v-b: now that I have rebased #3332 off of this PR, I am 100% going to demagic these codecs in that effort.

maxrjones:
@d-v-b have you tested this with different numcodecs versions to make sure there's no unexpected issues with clobbering of codec registration?

d-v-b:
> @d-v-b have you tested this with different numcodecs versions to make sure there's no unexpected issues with clobbering of codec registration?

I'm don't _think_ I expect any behavior to depend on numcodecs versions, happy to be corrected though. My understanding is that the codecs in `numcodecs.zarr3` are not registered with the numcodecs registry, and instead are exposed via the [entrypoints](https://github.com/zarr-developers/numcodecs/blob/878d2b385e4f1c6b113aca765e8632354abb3c9c/pyproject.toml#L78-L99) framework. That means we don't have to worry about anything interacting with numcodecs' own registry.

we have [a test](https://github.com/zarr-developers/zarr-python/blob/94983362f06f9bb4be3f04562b179549e36a87c4/tests/test_array.py#L1694) that checks our compatibility with these codecs, defined as dicts. In main these tests will pick up the codec class from numcodecs, but in this PR the version of the codec defined in zarr python is used instead. Is that kind of thing you are worried about?

src/zarr/codecs/numcodecs/_codecs.py
```diff
@@ -0,0 +1,325 @@
...
+        """To be used only when creating the actual public-facing codec class."""
+        super().__init_subclass__(**kwargs)
+        if codec_name is not None:
+            namespace = codec_name
+
+            cls_name = f"{CODEC_PREFIX}{namespace}.{cls.__name__}"
+            cls.codec_name = f"{CODEC_PREFIX}{namespace}"
+            cls.__doc__ = f"""
+            See :class:`{cls_name}` for more details and parameters.
+            """
```
rabernat, line 85: I vaguely remember a discussion a few months back about classes initialized this way having challenges with serialization...but I can't track down the issue.

tests/test_codecs/test_numcodecs.py
```diff
@@ -22,3 +68,290 @@ def test_is_numcodec_cls() -> None:
...
+            {},
+            shape=data.shape,
+            chunks=(16, 16),
+            dtype=data.dtype,
+            fill_value=0,
+            compressors=[codec_class()],
+        )
+
+    a[:, :] = data.copy()
+    np.testing.assert_array_equal(data, a[:, :])
```
rabernat, line 118: Love this test.

rabernat (approved):
Fantastic, great work Davis!

src/zarr/codecs/numcodecs/_codecs.py
d-v-b: those issues would have been resolved by https://github.com/zarr-developers/numcodecs/pull/745, and this PR uses the changes from that PR

PR #3379, opened 2025-08-17
Optimize getitem with empty chunks

### LINKED ISSUE #3368: Avoid calls to `np.full`
### Zarr version

main

### Numcodecs version

n/a

### Python Version

n/a

### Operating System

Linux/Mac

### Installation

source

### Description

In https://github.com/zarr-developers/zarr-python/blob/a26926cf4ec51a12ca28ff31c13ef2c8a32d51f9/src/zarr/core/buffer/cpu.py#L157-L161, we create an empty buffer. This is hit on the read path (IIRC, to allocate the buffer the data copied into).

`full` is surprisingly(?) slow.

```
In [18]: %timeit np.full(shape=125000000, fill_value=0)
74.6 ms ± 848 μs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [19]: %timeit np.empty(shape=125000000)
972 ns ± 36.2 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

It's worth confirming whether we need the Array's fill value here. My understanding is that it's just there in case the chunk happens to be missing. However, we *should* be able to limit the memset to just those regions of the output array that are actually missing. In pseduo code:

```python
out = NDBuffer.create(...)  # uses empty
for key, chunk_projection in zip(keys, chunk_projections):
    maybe_buffer = await store.get(key)
    if maybe_buffer is None:
        out[chunk_projection] = array.metadata.fill_value
    else:
        out[chunk_projection] = maybe_buffer
```

### Steps to reproduce

n/a

### Additional output

xref https://github.com/zarr-developers/zarr-python/issues/2904.

TomAugspurger:
This optimizes our read performance by using `NDBuffer.empty` instead of `NDBuffer.create` with a fill value. All chunks are filled, either with the values from the Store if that chunk was initialized, or the fill value if the Store returns `None` (the 
chunk wasn't initialized). By not memsetting values that will be overwritten anyway, things will be a bit faster.

I didn't see any documentation about CodecPipeline in the docs. Technically, users might see a behavior change, if the user's codec pipeline happened to rely on receiving an array filled with the `fill_value`. Do we know of anyone writing / using custom codec pipelines?

Here's a mini benchmark:

<details>

```python
import asyncio
import statistics
import time

import zarr.api.asynchronous
import zarr.storage


async def main():
    array = await zarr.api.asynchronous.create_array(
        store, shape=(125_000_000,), dtype="float64", chunks=(125_000,), fill_value=1.0
    )
    times = []
    for _ in range(10):
        t0 = time.monotonic()
        await array.getitem(slice(None))
        t1 = time.monotonic()
        times.append(t1 - t0)

    print("Unitialized", statistics.median(times))

    await array.setitem(slice(None), 2.0)
    times = []
    for _ in range(10):
        t0 = time.monotonic()
        await array.getitem(slice(None))
        t1 = time.monotonic()
        times.append(t1 - t0)
    print("Initialized", statistics.median(times))


if __name__ == "__main__":
    asyncio.run(main())

```

</details>

`main`:

```
❯ python bench.py
Unitialized 0.6652364450274035
Initialized 0.9079880800272804
```

HEAD:

```
❯ python bench.py
Unitialized 0.5070544500194956
Initialized 0.8395943949872162
```

Closes #3368 

TODO:

### DIFF
--- a/changes/3368.misc.rst
+++ b/changes/3368.misc.rst
@@ -1,0 +1,2 @@
+Improved performance of reading arrays by not unnecessarily using
+the fill value.
--- a/src/zarr/abc/codec.py
+++ b/src/zarr/abc/codec.py
@@ -423,6 +423,11 @@ class CodecPipeline:
             The second slice selection determines where in the output array the chunk data will be written.
             The ByteGetter is used to fetch the necessary bytes.
             The chunk spec contains information about the construction of an array from the bytes.
+
+            If the Store returns ``None`` for a chunk, then the chunk was not
+            written and the implementation must set the values of that chunk (or
+            ``out``) to the fill value for the array.
+
         out : NDBuffer
         """
         ...
--- a/src/zarr/codecs/sharding.py
+++ b/src/zarr/codecs/sharding.py
@@ -452,11 +452,10 @@ class ShardingCodec(
         )
 
         # setup output array
-        out = chunk_spec.prototype.nd_buffer.create(
+        out = chunk_spec.prototype.nd_buffer.empty(
             shape=shard_shape,
             dtype=shard_spec.dtype.to_native_dtype(),
             order=shard_spec.order,
-            fill_value=0,
         )
         shard_dict = await _ShardReader.from_bytes(shard_bytes, self, chunks_per_shard)
 
@@ -499,11 +498,10 @@ class ShardingCodec(
         )
 
         # setup output array
-        out = shard_spec.prototype.nd_buffer.create(
+        out = shard_spec.prototype.nd_buffer.empty(
             shape=indexer.shape,
             dtype=shard_spec.dtype.to_native_dtype(),
             order=shard_spec.order,
-            fill_value=0,
         )
 
         indexed_chunks = list(indexer)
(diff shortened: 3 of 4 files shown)


TomAugspurger:
@ilan-gold do you know if the zarrs-python codec pipeline relies on the `out_buffer` being an ndarray initialized with the `fill_value`? Somewhere around https://github.com/zarr-developers/zarr-python/pull/3379/files#diff-f488098fc197721bdc0ec9afd3ec04dc63a5bf1ca34078727306d6ee91c837f7L1366 is where the codec pipeline's `read` method is called.

ilan-gold:
Hmmm I'm not sure here.  I think we do the same as the zarr codec pipeline:

```
All chunks are filled, either with the values from the Store if that chunk was initialized, or the fill value if the Store returns None (the
chunk wasn't initialized). 
```

i.e., this description matches the code https://github.com/zarrs/zarrs-python/blob/main/src/lib.rs#L342-L361 so I think we're good.  Unless there were some other failure mode I could test, this looks reasonable (and our tests pass against this branch).

Issue #3389, opened 2025-08-19
DOC: Configuration value 'array.v3_default_compressors' has been removed, but is still documented


keller-mark:
### Describe the issue linked to the documentation

Hi, I am trying to disable compression (at either the store level or array level or via zarr.config) in my script.

I tried 

```py
zarr.config.set({'array.v3_default_compressors': None})
```

which is documented at https://zarr.readthedocs.io/en/stable/user-guide/config.html

but it threw an error

```
ValueError: Configuration value 'array.v3_default_compressors' has been removed
```

It seems like this is still documented in several other places as well https://github.com/search?q=repo%3Azarr-developers%2Fzarr-python%20v3_default_compressors&type=code

### Suggested fix for documentation

Beyond the `array.v3_default_compressors` configuration documentation, it would be helpful to provide an example in [Working with arrays -> Compression](https://zarr.readthedocs.io/en/stable/user-guide/arrays.html#compressors) which demonstrates how to disable compression options when using `create_array`

d-v-b:
hi @keller-mark, sorry about the confusing state of affairs with the config + docs. 

To answer your question, passing the kwargs `filters=None, compressors=None, serializer="auto"` to `create_array` should get you an array with no compression at all.

keller-mark:
Thank you for the quick response!

d-v-b:
closed via #3390

PR #3390, opened 2025-08-19
docstring cleanup

d-v-b:
This PR will do two things:
- remove outdated references to configuration options for controlling codecs (thereby closing #3389)
- make some of our function / method docstrings more consistent across the async / sync boundary.

The reason for two things in this PR, instead of 1, is that changing docstrings in zarr python is _incredibly tedious_ because we have a lot of similar functions, with very similar docstrings: we have `asynchronous.create`, `synchronous.create`, `asynchronous.create_array`, `synchronous.create_array`, `Group.create_array`, `AsyncGroup.create_array`, `init_array`, `from_array`, ... So any PR that changes docstrings needs to be sure that _all_ relevant docstrings are updated. This is more work, but it's part of the spirit of "leaving things better than you found it".

edit: this PR grew in scope, but not complexity. So I think it's fine to keep this in one PR.
Here's a list of the changes:

- remove outdated references to configuration options for controlling codecs (thereby closing #3389)
- Ensure that all docstrings for `store` parameters are consistent. They are all now correctly reporting the type of the parameter as `StoreLike` 
- Test that array creation function/methods which use the same parameters have those parameters declared identically in docstrings. We could expand these tests as needed to include more functions, but I wanted to focus on array creation routines for now because removing the outdated config content touched the docstrings for `compressor`, `serializer`, and `filters` extensively, so these needed to be tested.
- Where necessary, change docstrings to pass the above tests.

One specific change: I normalized a large number of `store` docstring declarations to `store : StoreLike or None, default=None` (obviously this is only for functions where the `store` parameter is nullable). This is consistent with NumpyDoc and more informative that just saying "StoreLike, optional", because the latter does not convey what the default value is.


d-v-b:
would be great to get some eyes on this @zarr-developers/python-core-devs

src/zarr/api/asynchronous.py
```diff
@@ -906,50 +906,28 @@ async def create(
     shape : int or tuple of ints
         Array shape.
     chunks : int or tuple of ints, optional
-        The shape of the array's chunks.
-        Zarr format 2 only. Zarr format 3 arrays should use `chunk_shape` instead.
-        If not specified, default values are guessed based on the shape and dtype.
+        Chunk shape. If True, will be guessed from `shape` and `dtype`. If
+        False, will be set to `shape`, i.e., single chunk for the whole array.
+        If an int, the chunk size in each dimension will be given by the value
+        of `chunks`. Default is True.
```
TomAugspurger:
```suggestion
        Chunk shape. If True, will be guessed from ``shape`` and ``dtype``. If
        False, will be set to ``shape``, i.e., single chunk for the whole array.
        If an int, the chunk size in each dimension will be given by the value
        of ``chunks``. Default is True.
```

src/zarr/api/asynchronous.py
```diff
@@ -958,13 +936,28 @@ async def create(
...
+        dict representations of :class:`zarr.abc.codec.ArrayArrayCodec`.
+
+        For Zarr format 2, a "filter" can be any numcodecs codec; you should ensure that the
+        the order if your filters is consistent with the behavior of each filter.
+
+        The default value of ``"auto"`` instructs Zarr to use a default used based on the data
+        type of the array and the Zarr format specified. For all data types in Zarr V3, and most
+        data types in Zarr V2, the default filters are empty. The only cases where default filters
+        are not empty is when the Zarr format is 2, and the data type is a variable-length data type like
+        `:class:zarr.dtype.VariableLengthUTF8` or `:class:zarr.dtype.VariableLengthUTF8`. In these cases,
```
TomAugspurger: I think:

```suggestion
        :class:`zarr.dtype.VariableLengthUTF8` or :class:`zarr.dtype.VariableLengthUTF8`. In these cases,
```

src/zarr/api/synchronous.py
```diff
@@ -977,24 +1024,27 @@ def from_array(
...
-        - Iterable[Codec]: List of filters to apply to the array.
-        - "auto": Automatically determine the filters based on the array's dtype.
-        - "keep": Retain the filters of the data array if it is a zarr Array.
+        The default value of ``"keep"`` instructs Zarr to infer ``filters`` from ``data``.
+        If that inference is not possible, Zarr will fall back to the behavior specified by ``"auto"``,
+        which is to choose default filters based on the data type of the array and the Zarr format specified.
+        For all data types in Zarr V3, and most data types in Zarr V2, the default filters are the empty tuple ``()``.
+        The only cases where default filters are not empty is when the Zarr format is 2, and the
+        data type is a variable-length data type like `:class:zarr.dtype.VariableLengthUTF8` or
+        `:class:zarr.dtype.VariableLengthUTF8`. In these cases, the default filters is a tuple with a
```
TomAugspurger:
```suggestion
        data type is a variable-length data type like :class:`zarr.dtype.VariableLengthUTF8` or
        :class:`zarr.dtype.VariableLengthUTF8`. In these cases, the default filters is a tuple with a
```

src/zarr/core/array.py
```diff
@@ -546,11 +539,25 @@ async def create(
...
+            dict representations of :class:`zarr.abc.codec.ArrayArrayCodec`.
+
+            For Zarr format 2, a "filter" can be any numcodecs codec; you should ensure that the
+            the order if your filters is consistent with the behavior of each filter.
+
+            The default value of ``"auto"`` instructs Zarr to use a default used based on the data
+            type of the array and the Zarr format specified. For all data types in Zarr V3, and most
+            data types in Zarr V2, the default filters are empty. The only cases where default filters
+            are not empty is when the Zarr format is 2, and the data type is a variable-length data type like
+            `:class:zarr.dtype.VariableLengthUTF8` or `:class:zarr.dtype.VariableLengthUTF8`. In these cases,
```
TomAugspurger:
```suggestion
            :class:`zarr.dtype.VariableLengthUTF8` or :class:`zarr.dtype.VariableLengthUTF8`. In these cases,
```

src/zarr/core/array.py
```diff
@@ -2086,11 +2090,25 @@ def create(
...
```
TomAugspurger:
```suggestion
            :class:`zarr.dtype.VariableLengthUTF8` or :class:`zarr.dtype.VariableLengthUTF8`. In these cases,
```

src/zarr/core/array.py
```diff
@@ -4218,24 +4236,27 @@ async def from_array(
...
-        - Iterable[Codec]: List of filters to apply to the array.
-        - "auto": Automatically determine the filters based on the array's dtype.
-        - "keep": Retain the filters of the data array if it is a zarr Array.
+        The default value of ``"keep"`` instructs Zarr to infer ``filters`` from ``data``.
+        If that inference is not possible, Zarr will fall back to the behavior specified by ``"auto"``,
+        which is to choose default filters based on the data type of the array and the Zarr format specified.
+        For all data types in Zarr V3, and most data types in Zarr V2, the default filters are the empty tuple ``()``.
+        The only cases where default filters are not empty is when the Zarr format is 2, and the
+        data type is a variable-length data type like `:class:zarr.dtype.VariableLengthUTF8` or
+        `:class:zarr.dtype.VariableLengthUTF8`. In these cases, the default filters is a tuple with a
```
TomAugspurger:
```suggestion
        data type is a variable-length data type like :class:`zarr.dtype.VariableLengthUTF8` or
        :class:`zarr.dtype.VariableLengthUTF8`. In these cases, the default filters is a tuple with a
```

src/zarr/core/array.py
```diff
@@ -4462,46 +4483,40 @@ async def init_array(
...
-        These defaults can be changed by modifying the value of ``array.v2_default_filters``
-        in :mod:`zarr.core.config`.
-        Use ``None`` to omit default filters.
-    compressors : Iterable[Codec], optional
+
+        The default value of ``"auto"`` instructs Zarr to use a default used based on the data
+        type of the array and the Zarr format specified. For all data types in Zarr V3, and most
+        data types in Zarr V2, the default filters are empty. The only cases where default filters
+        are not empty is when the Zarr format is 2, and the data type is a variable-length data type like
+        `:class:zarr.dtype.VariableLengthUTF8` or `:class:zarr.dtype.VariableLengthUTF8`. In these cases,
```
TomAugspurger:
```suggestion
        :class:`zarr.dtype.VariableLengthUTF8` or :class:`zarr.dtype.VariableLengthUTF8`. In these cases,
```

src/zarr/core/array.py
```diff
@@ -4757,6 +4773,7 @@ async def create_array(
         Ignored otherwise.
     overwrite : bool, default False
         Whether to overwrite an array with the same name in the store, if one exists.
+        If `True`, all existing paths in the store will be deleted.
```
TomAugspurger:
```suggestion
        If ``True``, all existing paths in the store will be deleted.
```

(this one might actually work, depending on the default domain...)

src/zarr/core/group.py
```diff
@@ -1062,24 +1062,25 @@ async def create_array(
...
-            If no ``filters`` are provided, a default set of filters will be used.
-            These defaults can be changed by modifying the value of ``array.v2_default_filters``
-            in :mod:`zarr.core.config`.
-            Use ``None`` to omit default filters.
+
+            The default value of ``"auto"`` instructs Zarr to use a default used based on the data
+            type of the array and the Zarr format specified. For all data types in Zarr V3, and most
+            data types in Zarr V2, the default filters are empty. The only cases where default filters
+            are not empty is when the Zarr format is 2, and the data type is a variable-length data type like
+            `:class:zarr.dtype.VariableLengthUTF8` or `:class:zarr.dtype.VariableLengthUTF8`. In these cases,
```
TomAugspurger:
```suggestion
            :class:`zarr.dtype.VariableLengthUTF8` or :class:`zarr.dtype.VariableLengthUTF8`. In these cases,
```

src/zarr/core/group.py
```diff
@@ -2448,9 +2449,149 @@ def require_groups(self, *names: str) -> tuple[Group, ...]:
...
+            dict representations of :class:`zarr.abc.codec.ArrayArrayCodec`.
+
+            For Zarr format 2, a "filter" can be any numcodecs codec; you should ensure that the
+            the order if your filters is consistent with the behavior of each filter.
+
+            The default value of ``"auto"`` instructs Zarr to use a default used based on the data
+            type of the array and the Zarr format specified. For all data types in Zarr V3, and most
+            data types in Zarr V2, the default filters are empty. The only cases where default filters
+            are not empty is when the Zarr format is 2, and the data type is a variable-length data type like
+            `:class:zarr.dtype.VariableLengthUTF8` or `:class:zarr.dtype.VariableLengthUTF8`. In these cases,
```
TomAugspurger:
```suggestion
            :class:`zarr.dtype.VariableLengthUTF8` or :class:`zarr.dtype.VariableLengthUTF8`. In these cases,
```

src/zarr/core/group.py
```diff
@@ -2891,24 +3033,25 @@ def array(
...
-            If no ``filters`` are provided, a default set of filters will be used.
-            These defaults can be changed by modifying the value of ``array.v2_default_filters``
-            in :mod:`zarr.core.config`.
-            Use ``None`` to omit default filters.
+
+            The default value of ``"auto"`` instructs Zarr to use a default used based on the data
+            type of the array and the Zarr format specified. For all data types in Zarr V3, and most
+            data types in Zarr V2, the default filters are empty. The only cases where default filters
+            are not empty is when the Zarr format is 2, and the data type is a variable-length data type like
+            `:class:zarr.dtype.VariableLengthUTF8` or `:class:zarr.dtype.VariableLengthUTF8`. In these cases,
```
TomAugspurger:
```suggestion
            :class:`zarr.dtype.VariableLengthUTF8` or :class:`zarr.dtype.VariableLengthUTF8`. In these cases,
```

d-v-b:
@TomAugspurger thanks for the fixes! tests are failing (as they should) because we only fixed one instance of a docstring that's duplicated across multiple functions. i'll get those fixes propagated, and then this should go green

Issue #3387, opened 2025-08-19
_dereference_path assertion error


zyndagj:
### Zarr version

v3.1.1

### Numcodecs version

v0.16.1

### Python Version

3.13.5

### Operating System

Linux

### Installation

conda

### Description

zarr-python v3.1.1 show incorrect file information and throws the following assertion error

```shell
$ python zarr_3.1.1_error.py        
Name        :
Type        : Group
Zarr format : 2
Read-only   : True
Store type  : LocalStore
Traceback (most recent call last):
  File "/hpc/mydata/greg.zynda/index/setup_zebrafish_czbiohub/zarr_3.1.1_error.py", line 8, i
n <module>
    print(zarr_data[0].shape)
          ~~~~~~~~~^^^
  File "/hpc/mydata/greg.zynda/anaconda/25.3.1/x86_64/envs/zarr/lib/python3.13/site-packages/
zarr/core/group.py", line 1892, in __getitem__
    obj = self._sync(self._async_group.getitem(path))                                       
  File "/hpc/mydata/greg.zynda/anaconda/25.3.1/x86_64/envs/zarr/lib/python3.13/site-packages/
zarr/core/sync.py", line 208, in _sync
    return sync(
        coroutine,
        timeout=config.get("async.timeout"),
    )
  File "/hpc/mydata/greg.zynda/anaconda/25.3.1/x86_64/envs/zarr/lib/python3.13/site-packages/
zarr/core/sync.py", line 163, in sync
    raise return_result
  File "/hpc/mydata/greg.zynda/anaconda/25.3.1/x86_64/envs/zarr/lib/python3.13/site-packages/
zarr/core/sync.py", line 119, in _runner
    return await coro
           ^^^^^^^^^^
  File "/hpc/mydata/greg.zynda/anaconda/25.3.1/x86_64/envs/zarr/lib/python3.13/site-packages/
zarr/core/group.py", line 697, in getitem
    store_path = self.store_path / key
                 ~~~~~~~~~~~~~~~~^~~~~
  File "/hpc/mydata/greg.zynda/anaconda/25.3.1/x86_64/envs/zarr/lib/python3.13/site-packages/
zarr/storage/_common.py", line 238, in __truediv__                                          
    return self.__class__(self.store, _dereference_path(self.path, other))                  
                                      ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^                   
[...]

PR #3392, opened 2025-08-20
Raise more helpful errors in _dereference_path

dstansby:
See https://github.com/zarr-developers/zarr-python/issues/3387 - this doesn't fix that bug, but it should at least raise a more helpful error message.


d-v-b (approved):
a later PR should add a docstring that explains what this function is supposed to do

PR #3394, opened 2025-08-20
Fix typing in store tests

dstansby:
More typing fixes, this time in the store tests.


dstansby:
I'll take a look and try and fix this now

PR #3395, opened 2025-08-21
reraise the `KeyError` for unknown codecs with a more specific error class

### LINKED ISSUE #3384: unhelpful error message for unknown codecs when opening a group
### Zarr version

3.1.0 (from main)

### Numcodecs version

0.16.1

### Python Version

3.12.3

### Operating System

Ubuntu 24.04

### Installation

using uv

### Description

When opening a zarr store that has a variable with an unknown codec (in my case, because I forgot to import the package defining the codec), zarr will raise a pretty unhelpful error message:
```
ZarrUserWarning: Object at <codec-name> is not recognized as a component of a Zarr hierarchy.
```

This is because https://github.com/zarr-developers/zarr-python/blob/2271067bf59a6158b93faf60b3a10c13d8691621/src/zarr/core/group.py#L3384-L3396 assumes that `KeyError` must have been thrown by the store.

However, if the codec is unknown the registry will also throw a `KeyError`, which will be caught and hidden behind the warning.

### Steps to reproduce

```python
# /// script
# requires-python = ">=3.11"
# dependencies = [
# ]
# ///

import zarr
import json
from zarr.buffer.cpu import Buffer

root = zarr.create_group(store=store)
z1 = root.create_array(
    name="baz", shape=(10000, 10000), chunks=(1000, 1000), dtype="int32"
)
print(z1)

metadata = json.loads(store._store_dict["baz/zarr.json"].to_bytes())
metadata["codecs"][0]["name"] = "b"

store._store_dict["baz/zarr.json"] = Buffer.from_bytes(json.dumps(metadata).encode())

g = zarr.open_group(store)
print(list(g.keys()))
```

This prints:
```
<Array memory://128459357805184/baz shape=(10000, 10000) dtype=int32>
.../zarr-python/src/zarr/core/group.py:3391: ZarrUserWarning: Object at b is not recognized as a component of a Zarr hierarchy.
  warnings.warn(
[]
```
where the warning is very misleading, this has nothing to do with the hierarchy.

To fix that, we'd probably have to catch the `KeyError` somewhere up the stack (in `parse_codecs`?) and convert it to a different error with a bit more information.

### Additional output

_No response_

keewis:
- [x] closes #3384

This converts the `KeyError` to a new zarr-specific `UnknownCodecError`, which since it indirectly inherits from `ValueError` won't be confused with responses by the store.

TODO:

### DIFF
--- a/src/zarr/errors.py
+++ b/src/zarr/errors.py
@@ -106,7 +106,7 @@ class UnknownCodecError(BaseZarrError):
     Raised when a unknown codec was used.
     """
 
-    _msg = "Unknown codec {!r}."
+    _msg = "{}"
 
 
 class NodeTypeValidationError(MetadataValidationError):
--- a/src/zarr/errors.py
+++ b/src/zarr/errors.py
@@ -26,6 +26,14 @@ class BaseZarrError(ValueError):
         super().__init__(self._msg.format(*args))
 
 
+class UnknownCodecError(BaseZarrError):
+    """
+    Raised when a unknown codec was used.
+    """
+
+    _msg = "Unknown codec {!r}."
+
+
 class GroupNotFoundError(BaseZarrError, FileNotFoundError):
     """
     Raised when a group isn't found at a certain path.
--- a/changes/3395.bugfix.rst
+++ b/changes/3395.bugfix.rst
@@ -1,0 +1,1 @@
+Raise a Zarr-specific error class when a codec can't be found by name when deserializing the given codecs. This avoids hiding this error behind a "not part of a zarr hierarchy" warning.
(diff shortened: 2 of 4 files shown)


src/zarr/errors.py
```diff
@@ -26,6 +26,14 @@ def __init__(self, *args: Any) -> None:
...
 
 
+class UnknownCodecError(BaseZarrError):
+    """
+    Raised when a unknown codec was used.
+    """
+
+    _msg = "Unknown codec {!r}."
+
+
```
d-v-b: I think we want to eventually move away from defining exceptions with a pre-formatted exception message, so maybe use something like 
`_msg = '{}'`

PR #3400, opened 2025-08-24
add a runtime type checker for metadata objects

### LINKED ISSUE #3285: unified runtime type checking for our json data
Reading array metadata documents means consuming JSON-formatted data and type-checking them. We are doing this in a clunky way right now: for each field `f`, we essentially have a `parse_f` function that checks if the input data is compatible with `f`. 

A big improvement would be unifying the `parse_x` functions into something with the following form:

```python
def check_literal(value, type_annotation) -> type_annotation:
    if value in get_args(type_annotation):
        return value
    raise UsefulException

def parse_union(value, type_annotation) -> type_annotation:
    # check if the value is in the union type, return it if so

def parse_tuple(value, type_annotation) -> type_annotation:
    # check if the value is consistent with the tuple type annotation, return the input as a tuple if so

...

parse_json(value, type_annotation) -> type_annotation:
    # categorize the type annotation into Mapping, tuple, Sequence, union, literal, and call out to the relevant 
    # parsing routine
    # return data 
```

i.e., functions that take a value and a type annotation, and return data assignable to that type annotation or raise a useful exception. In my thinking these are not strict type checks, because these functions are allowed to transform the input, e.g. `parse_tuple([1,2,3], tuple[int, int, int])` would return `(1,2,3)`.

This would remove a _lot_ of redundant code. We would keep the scope narrow by only concerning ourselves with the types relevant for creating array metadata documents, namely:

- primitive types None, str, int, float, bool
- Sequences (they should come out as tuples)
- unions
- TypedDict (essential for handling JSON form of dtypes, codecs, chunk grids, and metadata itself)
- Mapping[str, T], where T is any of the other types in this list

I have have had LLMs whip up implementations of this on like 3 separate occasions, and each time it wasn't more than a few hundred LOC, so I think this would not be a huge maintenance burden.

d-v-b:
This PR adds a runtime type checker specifically for checking JSON-like data against a type definition. It's currently a draft while I get the test suite happy and refine the API, but it's also ready for people to look at and try out. I'm pretty convinced of it's utility, but I also think we should have a good discussion about whether this feature is a good idea.

### Demo

The basic API looks like this:
```python
# /// script
# requires-python = ">=3.11"
# dependencies = [
#   "zarr@git+https://github.com/d-v-b/zarr-python.git@feat/type-checker",
# ]
# ///

from zarr.core.type_check import check_type
from typing import TypedDict, Literal, NotRequired

class MyChunkGridConfig(TypedDict):
   prop_a: int
   prop_b: str
   prop_c: str | int
   optional_prop: NotRequired[bool]

class MyChunkGrid(TypedDict):
   name: Literal["awesome_chunk_grid"]
   configuration: MyChunkGridConfig

valid_data = [
   {"name": "awesome_chunk_grid", "configuration": {"prop_a": 10, "prop_b": "string", "prop_c": 10}},
   {"name": "awesome_chunk_grid", "configuration": {"prop_a": 10, "prop_b": "string", "prop_c": "10"}},
   {"name": "awesome_chunk_grid", "configuration": {"prop_a": 10, "prop_b": "string", "prop_c": "10", "optional_prop": True}}
   ]

for d in valid_data:
    result = check_type(d, MyChunkGrid)
    print(result.success, result.errors)
"""
True []
True []
True []
"""

# prop_b should be a string, but it's an int
invalid_data = [
    # invalid type
   {"name": "awesome_chunk_grid", "configuration": {"prop_a": 10, "prop_b": 10, "prop_c": 10}},
   # missing key
   {"name": "awesome_chunk_grid", "configuration": {"prop_a": 10, "prop_b": "10"}}
   ]
for d in invalid_data:
    result = check_type(d, MyChunkGrid)
    print(result.success, result.errors)
"""
False ["value['configuration']['prop_b'] expected an instance of <class 'str'> but got 10 with type <class 'int'>"]
False ["value['configuration'] missing required key 'prop_c'"]
"""
```

Some aspects might evolve while this is a draft, like the nature of the error messages.

### Supported types

This is not a general-purpose type checker. It is targeted for the types relevant for Zarr metadata documents, and so It supports the following narrow set of types

- int, bool, float, str
- union
- tuple
- list
- sequence
- mapping
- typeddict

#### maintenance burden

The [type checker itself](https://github.com/d-v-b/zarr-python/blob/feat/type-checker/src/zarr/core/type_check.py) is ~530 lines of commented code, broken up into functions which are mostly easy to understand. The typeddict part, and the logic for resolving generic types, is convoluted and potentially sensitive to changes in how python exposes type annotations at runtime. Many type annotation features have been designed for static type checkers and not use within a python program, so some of this is rather fiddly. But I don't think we are relying on any brittle or private APIs here.

#### performance

As currrently implemented, the type checker will report all detectable errors:

```python
>>> check_type(tuple(range(4)), tuple[str, ...])
TypeCheckResult(success=False, errors=["value[0] expected an instance of <class 'str'> but got 0 with type <class 'int'>", "value[1] expected an instance of <class 'str'> but got 1 with type <class 'int'>", "value[2] expected an instance of <class 'str'> but got 2 with type <class 'int'>", "value[3] expected an instance of <class 'str'> but got 3 with type <class 'int'>"])
```
This is wasted compute when we don't care exactly how mismatched the data is, but it _is_ a better user experience. We might need to tune this if performance becomes a problem, e.g. by introducing a `"fail_fast"` option that returns on the first error.

### benefit

We can instantly remove a lot of special-purpose functions. Most of the [functions named `parse_*`](https://github.com/search?q=repo%3Azarr-developers%2Fzarr-python%20parse_&type=code) (~30+ functions) and essentially all of the functions named [`*check_json*`](https://github.com/search?q=repo%3Azarr-developers%2Fzarr-python+check_json&type=code) (~30 functions) could be replaced or simplified with the `check_type` function.

We can also make our JSON loading routines type-safe:

```python
# /// script
# requires-python = ">=3.11"
# dependencies = [
#   "zarr@git+https://github.com/d-v-b/zarr-python.git@feat/type-checker",
# ]
# ///

from zarr.core.type_check import guard_type
from zarr.core.common import ArrayMetadataJSON_V3

unknown_data: dict[str, object] = {
    "zarr_format": 3,
    "node_type": "array",
    "shape": (10,10),
    "data_type": "uint8", 
    "chunk_grid": {"name": "regular", "configuration": {"chunk_shape": (10,10)}},
    "chunk_key_encoding": {"name": "default", "configuration": {"separator": "/"}},
    "codecs": ({"name": "bytes"}, ),
    "fill_value": 0
    }

if guard_type(unknown_data, ArrayMetadataJSON_V3):
    reveal_type(unknown_data)
```
mypy could not infer the type correctly, but basedpyright does:
```zsh
zarr-python git:(feat/type-checker) ✗ uvx basedpyright test.py
/Users/d-v-b/dev/zarr-python/test.py
  /Users/d-v-b/dev/zarr-python/test.py:25:17 - information: Type of "unknown_data" is "ArrayMetadataJSON_V3"
0 errors, 0 warnings, 1 note
```

While we _could_ write a bespoke function that specifically checks all the possibilities for zarr v3 metadata. But then we would need to painfully modify that function by hand to support something like this:

```python
# /// script
# requires-python = ">=3.11"
# dependencies = [
#   "zarr@git+https://github.com/d-v-b/zarr-python.git@feat/type-checker",
# ]
# ///
from collections.abc import Mapping
from typing_extensions import TypedDict
from zarr.core.common import ArrayMetadataJSON_V3
from zarr.core.type_check import check_type

class GeoZarrAttrs(TypedDict):
    geozarr: Mapping[str, object]

class GeoZarrArray(ArrayMetadataJSON_V3):
    attributes: GeoZarrAttrs

unknown_data: dict[str, object] = {
    "zarr_format": 3,
    "node_type": "array",
    "shape": (10,10),
    "data_type": "uint8", 
    "chunk_grid": {"name": "regular", "configuration": {"chunk_shape": (10,10)}},
    "chunk_key_encoding": {"name": "default", "configuration": {"separator": "/"}},
    "codecs": ({"name": "bytes"}, ),
    "fill_value": 0,
    "attributes": {"not_geozarr": "bar"}
    }

result = check_type(unknown_data, GeoZarrArray)
print(result)
"""
TypeCheckResult(success=False, errors=["value['attributes'] missing required key 'geozarr'"])
"""
```

### alternatives

we could use an external JSON validation library / type checking like pydantic, attrs, msgspec, beartype, etc. But I would rather not add a dependency. With the approach in this PR, we keep control in-house, and because this PR just adds functions, it composes with the rest of our codebase at the moment. (FWIW right now this type checker doesn't do any parsing, it only validates. If you think we should [parse instead of just validating](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/), then IMO that's a job for our array metadata classes)

we could also do nothing, and continue writing JSON parsing code by hand. But I would rather not do that, because this invites bugs and makes it hard to keep up with [sneaky spec changes](https://github.com/zarr-developers/zarr-specs/issues/352). Specifically, I'm planning on writing a lot of new types to model the codecs defined in #3376, and I would rather _just_ write the type and get the type checking (and type safety) for free.

closes #3285


d-v-b:
this is pretty substantial so I would appreciate a lot of eyes @zarr-developers/python-core-devs 

if anyone has concerns about whether we should do any runtime type checking at all, maybe send those thoughts to the [issue this PR closes](https://github.com/zarr-developers/zarr-python/issues/3285)

I'm going to keep working on tests for the type checker, but so far it's working great. 

This PR does violating liskov for a few subclasses of our `Metadata` ABC, because that class requires that `to_dict` and `from_dict` use `dict[st, JSON]`, which is not very accurate. After this PR we need to make changes to class so that it supports type safety, then we won't be violating liskov any more.

Similarly, there are lots of `# type: ignores` added in various places. As you might guess, those were added because without them, mypy flagging type errors. Many of these checks will go away when we fix the lax typing of the `Metadata` ABC, and other classes.

@TomAugspurger I think you in particular will appreciate some of the effects of this PR. Since we can annotate methods like `ArrayV3Metadata.from_dict()` as taking `ArrayMetadataJSON_V3`, we don't need to do any runtime validation inside `from_dict`. The assumption is that the caller has already checked the input. This is not possible without the types defined in this PR, and it's not _practical_ without the type checker. Once we extrapolate the type safe style, we can push almost all our type checking to the IO boundary.

That being said, I think the ArrayMetadata class will still need to do some internal consistency checks, like ensuring that the number of dimension names matches the length of `shape`. I don't think we want our type checker to be smart enough to catch that kind of thing.

d-v-b:
In the dev meeting today @jhamman rightly challenged me to justify writing our own in-house runtime type checker instead of bringing in pydantic. I had a few answers:
- The type checker here is small (~500 LoC, with docstrings) and it has a very simple structure -- a bunch of functions. Pydantic is huge and complex. It relies on dynamic class creation, and bugs in in pydantic can be hard to debug. 
- pydantic is class-centric. I think a _functional_ API for type checking is a better fit for our codebase. Replacing all our classes with `BaseModel` is relatively disruptive comparing to bringing in a few functions to help with a type guard. 
- Pydantic contains a lot of stuff we don't need -- For zarr, we only need to check the types that could fit in `zarr.json`, but pydantic covers a _lot_ more. The scale of pydantic is also an upside -- it means that new features will get added, and bugs will be squashed, by a large number of developers. But as long as we test our in-house type checker extensively against the small set of types that matter to us, I don't think we would gain much from adding pydantic as a dependency.
- Pydantic makes breaking changes. If we depend on pydantic, we then become responsible for managing how those breaking changes impact our users. This is a cost that I think we can avoid.
- The changes here are very non-invasive (because it's just functions). We can always switch to pydantic later, if this type checker sucks. We could also revert and replace all instances of this type checker with a hand-written type checking function.

My calculation was, 500 LoC for an API that does exactly what we need, (and nothing else), was a better deal than bringing in a big, celebrity dependency that would require pretty substantial changes to basic Zarr Python classes. Other folks might see this calculation differently.

d-v-b:
as illustration, consider this diff enabled by this PR:

```diff
@@ -286,14 +283,7 @@ class NullTerminatedBytes(ZDType[np.dtypes.BytesDType[int], np.bytes_], HasLengt
             True if the input is a valid representation of this class in Zarr V3, False
             otherwise.
         """
-        return (
-            isinstance(data, dict)
-            and set(data.keys()) == {"name", "configuration"}
-            and data["name"] == cls._zarr_v3_name
-            and isinstance(data["configuration"], dict)
-            and "length_bytes" in data["configuration"]
-            and isinstance(data["configuration"]["length_bytes"], int)
-        )
+        return check_type(data, NullTerminatedBytesJSON_V3).success
```

 I worry that if type safety requires writing tedious, error-prone functions like what we have in main, we might not do it, and then we lose type safety. Boiling it down to a much simpler function makes type safety look a lot easier.

src/zarr/abc/codec.py
```diff
@@ -37,27 +34,6 @@
 CodecInput = TypeVar("CodecInput", bound=NDBuffer | Buffer)
 CodecOutput = TypeVar("CodecOutput", bound=NDBuffer | Buffer)
 
-TName = TypeVar("TName", bound=str, covariant=True)
-
-
-class CodecJSON_V2(TypedDict, Generic[TName]):
-    """The JSON representation of a codec for Zarr V2"""
```
TomAugspurger, line 44: Probably should reexport these types here for backwards compatibility.

src/zarr/core/array.py
```diff
@@ -945,7 +976,7 @@ def from_dict(
         ValueError
             If the dictionary data is invalid or incompatible with either Zarr format 2 or 3 array creation.
         """
-        metadata = parse_array_metadata(data)
+        metadata = parse_array_metadata(data)  # type: ignore[call-overload]
```
TomAugspurger, line 979: Is the `type: ignore` needed due to the presence of `ArrayMetadata` in the signature for `__init__`?

src/zarr/core/common.py
```diff
@@ -67,6 +68,132 @@ class NamedConfig(TypedDict, Generic[TName, TConfig]):
...
+# I.e., ConsolidatedMetadata_JSON[ArrayMetadataJSON_V2 | GroupMetadataJSON_V2]
+class ConsolidatedMetadata_JSON_V2(TypedDict):
+    """
+    A typed dictionary model for Zarr consolidated metadata.
+
+    This model is parameterized by the type of the metadata itself.
+    """
+
+    kind: Literal["inline"]
+    must_understand: Literal["false"]
```
TomAugspurger, line 155: Is this correct for v2 consolidated metadata? IIRC, it didn't have `must_understand` and maybe had a `version` field? I guess maybe I'm mixing up consolidated metadata as written by zarr-python 2.x, and the v3 spec.

src/zarr/core/type_check.py
```diff
@@ -0,0 +1,505 @@
...
+        # Special handling for single-argument generics like NotRequired, ReadOnly
+        if len(new_args) == 1:
+            return origin[new_args[0]]  # Pass single argument, not tuple
+        else:
+            return origin[new_args]  # Pass tuple for multi-argument generics
+
+    return tp
+
+
+def check_type(obj: Any, expected_type: Any, path: str = "value") -> TypeCheckResult:
```
TomAugspurger: Can we narrow the type on `expected_type` here? And do we gain anything if we do?

I'd like to minimize complexity. We know we have a relatively constrained set of expected objects we're going to be passing here (node metadata, codec configuration, ...). If there's anything we can leverage to make this simpler, let's do it.

src/zarr/core/type_check.py
```diff
@@ -0,0 +1,505 @@
...
+        tn = _type_name(expected_type)
+        return TypeCheckResult(False, [f"{path} expected {tn} but got {type(obj).__name__}"])
+    except TypeError:
+        return TypeCheckResult(False, [f"{path} cannot be checked against {expected_type}"])
+
+
+T = TypeVar("T")
+
+
+def ensure_type(obj: object, expected_type: type[T], path: str = "value") -> T:
```
TomAugspurger, line 218: `path` is unused?

TomAugspurger (commented):
Gave a Quick Look, but haven't gone through in detail. At a high level:

1. I'd prefer to avoid a dependency on pydantic (or any other runtime validators)
2. At a glance, the implementation looks reasonable, but I haven't had a chance to go through it in detail (and likely won't for a while)
3. I'd say anything we can do to limit scope, with the goal of limiting complexity, is worth considering.

src/zarr/abc/codec.py
d-v-b: done in https://github.com/zarr-developers/zarr-python/pull/3400/commits/30d48a8b7768b8fab33edef62a79eebcefaae8cf

src/zarr/core/array.py
d-v-b: we need the overload because the `data` input to `from_dict` is typed as `dict[str, JSON]`. I could change that to the union of the two typeddict metadata types, but this will incompatibly override the base class implementation of `from_dict`, and so I will need another `# type: ignore` there. The only way to clean all of this up is to redo our base `Metadata` ABC, which I deemed out of scope for this PR

src/zarr/core/common.py
d-v-b: wasn't v2 consolidated metadata exclusively via an external `.zmetadata` file, with no change to `.zattrs`?

d-v-b: you're right, the `.zmetadata` file had [version and `"metadata"` keys](https://github.com/zarr-developers/zarr-python/blob/5e1a466bd538dc02a1e56213d9858e64b32931cb/zarr/convenience.py#L1293):

```python
  out = {
        "zarr_consolidated_format": 1,
        "metadata": {key: json_loads(store[key]) for key in store if is_zarr_key(key)},
    }
```

But the above typeddict is for modelling the use of our inline consolidated metadata model for Zarr V2 arrays. I would need a totally separate type for the zmetadata contents.

tests/test_metadata/test_consolidated.py
```diff
@@ -402,7 +398,7 @@ def test_flatten(self) -> None:
                                     metadata={
                                         "array": ArrayV3Metadata.from_dict(
                                             {
-                                                **array_metadata,
+                                                **array_metadata,  # type: ignore[typeddict-item]
```
d-v-b, line 401: mypy doesn't know about [structural assignability in typeddicts](https://typing.python.org/en/latest/spec/typeddict.html#inheritance), i.e. the fact that typeddicts tolerate extra keys

src/zarr/core/type_check.py
d-v-b: i think you're kind of asking if we could define a "type-checkable type", i.e. a union of all the types we support runtime type checking for. I'd like that, but I haven't looked into setting that up yet. I don't know what the blockers would be, and I think it would be potentially helpful to more clearly define the scope of this checker.

although, in its current implementation, any type that isn't associated with a specific checking routine falls back to `isinstance`, which does work for pretty much all user-defined classes...

src/zarr/core/type_check.py
d-v-b: it's passed to `check_type` a few lines down

src/zarr/core/type_check.py
d-v-b: have a look at https://github.com/zarr-developers/zarr-python/pull/3400/commits/c7096b14e4cacd5de833cd475b8f42689e38d771, It wasn't too painful to narrow from `Any` to `type | types.UnionType | ForwardRef | None`

d-v-b:
closing as not planned for now!

Issue #3402, opened 2025-08-25
deprecate templated exception messages


d-v-b:
We currently have some exceptions that take multiple parameters and use them to generated a pre-formatted exception:

```python

class BaseZarrError(ValueError):
    """
    Base error which all zarr errors are sub-classed from.
    """

    _msg = ""

    def __init__(self, *args: Any) -> None:
        super().__init__(self._msg.format(*args))

class ContainsGroupError(BaseZarrError):
    """Raised when a group already exists at a certain path."""

    _msg = "A group exists in store {!r} at path {!r}."

ContainsGroupError('a','b')
# ContainsGroupError("A group exists in store 'a' at path 'b'.")
```

I think this is kind of pointless, because outside of our tests we are not programmatically parsing the string contents of exception messages, and it's also confusing for developers because we don't document these parameters on the exception classes, and normal python exceptions don't (to my knowledge) require multiple arguments that go into a templated string.

Since IMO this pre-formatting is of negative value, I propose deprecating it. To make this smooth, creating an exception with a single argument will treat that as a pre-formatted string (i.e., normal python exception behavior). For exceptions that previously performed special formatting, this will be triggered by creating the exception with multiple arguments.

I don't know how to deprecate this "noisily" -- emitting a deprecation warning when an exception is raised doesn't really reach the right audience, and I don't know how to issue a developer-targeted deprecation warning for a specific form of object construction. Happy to hear other ideas here.

d-v-b:
this was fixed in #3403

PR #3403, opened 2025-08-25
chore/unformatted exceptions

d-v-b:
This pr addresses #3402 by allowing all of the zarr-specific exceptions take a single argument, which will be treated as a pre-formatted message. The old behavior, in which an exception would take multiple arguments and use them to format a canned message, is still available by passing in multiple arguments. I don't know how to deprecate this visibly, but I also suspect the API of `zarr.errors` is not a load-bearing member for many downstream projects.

While changing the exceptions in `zarr.errors`, I found more exceptions defined in `zarr.core.indexing`. I moved those definitions to `zarr.errors`, because all our errors should be defined in one place, barring a strong reason otherwise. Had all the errors been defined in place, I wouldn't have needed to use a text search to discover the additional errors in `zarr.core.indexing`.

### DIFF
--- a/changes/2859.removal.rst
+++ b/changes/2859.removal.rst
@@ -1,0 +1,2 @@
+The ``Store.set_partial_writes`` method, which was not used by Zarr-Python, has been removed.
+``store.supports_partial_writes`` is now always ``False``.
--- a/src/zarr/abc/store.py
+++ b/src/zarr/abc/store.py
@@ -4,7 +4,7 @@ from abc import ABC, abstractmethod
 from asyncio import gather
 from dataclasses import dataclass
 from itertools import starmap
-from typing import TYPE_CHECKING, Protocol, runtime_checkable
+from typing import TYPE_CHECKING, Literal, Protocol, runtime_checkable
 
 if TYPE_CHECKING:
     from collections.abc import AsyncGenerator, AsyncIterator, Iterable
@@ -12,7 +12,6 @@ if TYPE_CHECKING:
     from typing import Any, Self, TypeAlias
 
     from zarr.core.buffer import Buffer, BufferPrototype
-    from zarr.core.common import BytesLike
 
 __all__ = ["ByteGetter", "ByteSetter", "Store", "set_or_delete"]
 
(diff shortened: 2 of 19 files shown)


src/zarr/api/asynchronous.py
```diff
@@ -1268,7 +1269,8 @@ async def open_array(
                 overwrite=overwrite,
                 **kwargs,
             )
-        raise ArrayNotFoundError(store_path.store, store_path.path) from err
+        msg = "No array found in store {store_path.store} at path {store_path.path}"
```
maxrjones: should this be an f-string?

src/zarr/core/group.py
```diff
@@ -3549,7 +3553,8 @@ def _build_metadata_v3(zarr_json: dict[str, JSON]) -> ArrayV3Metadata | GroupMet
     Convert a dict representation of Zarr V3 metadata into the corresponding metadata class.
     """
     if "node_type" not in zarr_json:
-        raise MetadataValidationError("node_type", "array or group", "nothing (the key is missing)")
+        msg = "Invalid value for 'node_type'. Expected 'array' or 'group'. Got nothing (the key is missing)."
```
maxrjones:
```suggestion
        msg = "Required key 'node_type' is missing from the provided metadata document."
```
I think this makes more sense

src/zarr/errors.py
```diff
@@ -146,3 +132,15 @@ class ZarrRuntimeWarning(RuntimeWarning):
...
+class VindexInvalidSelectionError(IndexError): ...
+
+
+class NegativeStepError(IndexError): ...
+
+
+class BoundsCheckError(IndexError): ...
+
+
+class ArrayIndexError(IndexError): ...
```
maxrjones, line 146: should these include the deprecated message formatters from src/zarr/core/indexing.py?

src/zarr/api/asynchronous.py
d-v-b: oops yes it should

src/zarr/errors.py
d-v-b: it's not possible because their pre-formatted messages only took 1 argument. so there's no way to support the deprecated behavior AND support a user-provided error message. I think this is OK though, because i doubt anyone was subclassing these.

src/zarr/api/asynchronous.py
d-v-b: fixed in https://github.com/zarr-developers/zarr-python/pull/3403/commits/9ce2404c505b2a1389bc54902a07a7efd186a4d4

maxrjones:
This looks good, I think a test just needs updating for my suggestion of a more informative error message. I'll approve once the tests pass

d-v-b:
@maxrjones tests are green

Issue #2582, opened 2024-12-20
Silent failure when creating an array where there is an existing node


jhamman:
### Zarr version

3.0.0b4.dev7+g122760fc.d20241213

### Numcodecs version

0.14

### Python Version

3.11

### Operating System

Mac

### Installation

hatch

### Description

In the example below, I create a root group, then a child array (name `foo`). Then I create another child array (name `foo/bar`). The resulting tree only includes the first `foo`. 

I would expect creating `foo/bar` to error because `bar` cannot have an array as a parent. 

### Steps to reproduce

```python
import numpy as np
import zarr
root = zarr.group('data/example-12.zarr')
foo = root.create_array(name='foo', shape=(1000, 100), chunks=(10, 10), dtype='f4')
bar = root.create_array(name='foo/bar', shape=(100,), dtype='i4')
foo[:, :] = np.random.random((1000, 100))
bar[:] = np.arange(100)

root.tree()
```
```
/
└── foo (1000, 100) float32
```

ricor07:
Hello, I'd like to work on the project by fixing this issue. Can you assign this to me? Thanks

K-Meech:
I'll have a go at fixing this one, as there has been no activity on it since last year. Do let me know if anyone is actively working on this though!

PR #3407, opened 2025-08-26
Prevent creation of arrays/groups under a parent array

K-Meech:
Closes https://github.com/zarr-developers/zarr-python/issues/2582

Prevents creation of arrays / groups under a parent array with `.create_array` and `.create_group` e.g. `root.create_array(name='foo/bar')` (where `foo` is an existing array)

This required changes to the `_save_metadata` function of both `zarr/core/array.py` and `zarr/core/group.py`. As both used pretty much identical code, I refactored this into a common function in `zarr/core/metadata/io.py` (along with the `_build_parents` function both relied upon). Happy to move this elsewhere - if there is a more suitable location for it!

I tried to avoid looping over the parents multiple times in `_save_metadata` for the sake of performance (potentially this path could be deeply nested). Hence looping once, and creating two sets of awaitables: one for checking if an array exists at the location + one to actually modify the metadata there. Again, happy to update this if there's a simpler solution.

TODO:

### DIFF
--- a/tests/test_indexing.py
+++ b/tests/test_indexing.py
@@ -2042,14 +2042,14 @@ class TestAsync:
 
     @pytest.mark.asyncio
     async def test_async_oindex_with_zarr_array(self, store):
-        z1 = zarr.create_array(store=store, shape=(2, 2), chunks=(1, 1), zarr_format=3, dtype="i8")
+        group = zarr.create_group(store=store, zarr_format=3)
+
+        z1 = group.create_array(name="z1", shape=(2, 2), chunks=(1, 1), dtype="i8")
         z1[...] = np.array([[1, 2], [3, 4]])
         async_zarr = z1._async_array
 
         # create boolean zarr array to index with
-        z2 = zarr.create_array(
-            store=store, name="z2", shape=(2,), chunks=(1,), zarr_format=3, dtype="?"
-        )
+        z2 = group.create_array(name="z2", shape=(2,), chunks=(1,), dtype="?")
         z2[...] = np.array([True, False])
 
         result = await async_zarr.oindex.getitem(z2)
@@ -2075,14 +2075,14 @@ class TestAsync:
 
     @pytest.mark.asyncio
     async def test_async_vindex_with_zarr_array(self, store):
-        z1 = zarr.create_array(store=store, shape=(2, 2), chunks=(1, 1), zarr_format=3, dtype="i8")
+        group = zarr.create_group(store=store, zarr_format=3)
+
+        z1 = group.create_array(name="z1", shape=(2, 2), chunks=(1, 1), dtype="i8")
         z1[...] = np.array([[1, 2], [3, 4]])
         async_zarr = z1._async_array
 
         # create boolean zarr array to index with
-        z2 = zarr.create_array(
-            store=store, name="z2", shape=(2, 2), chunks=(1, 1), zarr_format=3, dtype="?"
-        )
+        z2 = group.create_array(name="z2", shape=(2, 2), chunks=(1, 1), dtype="?")
         z2[...] = np.array([[False, True], [False, True]])
 
         result = await async_zarr.vindex.getitem(z2)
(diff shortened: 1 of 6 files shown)


d-v-b:
conceptually I don't think we want a `zarr.core.metadata.io` module. So far, the `zarr.core.metadata` module has been exclusively for definitions of the metadata classes, which is distinct from the routines for saving metadata documents to disk.

right now I think a better location would be `zarr.core.group` -- would that be possible?

K-Meech:
Thanks @d-v-b - I think the only issue with this is it may introduce a circular import? `zarr.core.group` already imports multiple functions from `zarr.core.array`, and `zarr.core.array` would need to import these common metadata saving functions from `zarr.core.group`.

I could move those functions into `zarr.core.array` instead? `_build_parents` used to be in that file, and was imported by both `zarr.core.array` and `zarr.core.group` (although only by putting an import inside that function to avoid another circular import 😅 )

d-v-b:
very good points @K-Meech. Now I do think we should have `zarr.core.metadata.io`. But I don't think it should depend on `AsyncArray` or `AsyncGroup`. Looking at `_build_parents`, all it does is save metadata. I don't think this class needs to use `AsyncGroup` for that, we can just write the metadata out directly without that class. 

In fact, we might not need the metadata-saving logic in `_build_parents`, since it's being called by _another_ routine that's already creating metadata (`save_metadata`). IMO `_build_parents` should just decide where metadata documents need to be created, and then return a `{name: GroupMetadata}` dict with the names of the groups that need to be created by the calling function (e.g., `save_metadata`).

I suspect all of this could be handled by [`create_hierarchy`](https://github.com/zarr-developers/zarr-python/blob/c9509ee4416499e0b9bf735acb0820d989a61ebe/src/zarr/core/group.py#L1451). In any case, that function should also be moved to `zarr.core.metadata.io` (i'm happy to do this in another PR if you don't want to)

edit: I mistakenly linked to the `create_hierarchy` method defined on the `AsyncGroup` class. I meant to link to the stand-alone function: https://github.com/zarr-developers/zarr-python/blob/c9509ee4416499e0b9bf735acb0820d989a61ebe/src/zarr/core/group.py#L2985

K-Meech:
Thanks @d-v-b - sounds like a good plan. I'll look into removing the dependency on `AsyncArray` + `AsyncGroup`.
I'm away for the next few weeks, but can look into this + any other comments left on the PR in the meantime once I'm back.

K-Meech:
@d-v-b I've updated the code now to match your suggestion above. The dependency on `AsyncArray` / `AsyncGroup` is removed, and `_build_parents` now returns a dict of `{name: GroupMetadata}` that `save_metadata` then creates.

I agree that `create_hierarchy` could probably handle much of this functionality, but perhaps it's best to change this in a separate PR to keep the changes minimal here? What do you think?

d-v-b:
> I agree that `create_hierarchy` could probably handle much of this functionality, but perhaps it's best to change this in a separate PR to keep the changes minimal here? What do you think?

That works for me!

tests/test_indexing.py
```diff
@@ -2110,14 +2110,14 @@ async def test_async_oindex(self, store, indexer, expected):
 
     @pytest.mark.asyncio
     async def test_async_oindex_with_zarr_array(self, store):
-        z1 = zarr.create_array(store=store, shape=(2, 2), chunks=(1, 1), zarr_format=3, dtype="i8")
```
d-v-b, line 2113: why did we need to change this test?

K-Meech: This test creates an array as a child of another array, so would throw an error (I changed it to create a group then put both arrays inside instead).

E.g. running the equivalent for a local store, shows an array inside an array:
```
store = LocalStore("data/images/arr")
z1 = zarr.create_array(store=store, shape=(2, 2), chunks=(1, 1), zarr_format=3, dtype="i8")
z1[...] = np.array([[1, 2], [3, 4]])
async_zarr = z1._async_array

# create boolean zarr array to index with
z2 = zarr.create_array(
    store=store, name="z2", shape=(2,), chunks=(1,), zarr_format=3, dtype="?"
)
z2[...] = np.array([True, False])
```

d-v-b (approved):
thank you!

PR #3412, opened 2025-08-27
Use atomic writes for new files in LocalStore

### LINKED ISSUE #3411: Atomic writes in LocalStore
Zarr v2 took care in DirectoryStore to use atomic writes by renaming files into place, which ensures that writes to disk are either complete or not written at all:
https://github.com/zarr-developers/zarr-python/blob/f1978dde52b11a7803a1a06937e69948acc8fe84/zarr/storage.py#L1146-L1161

I don't see any similar logic in Zarr v3, which just seems to using Python's file interface (via `open`):
https://github.com/zarr-developers/zarr-python/blob/c9509ee4416499e0b9bf735acb0820d989a61ebe/src/zarr/storage/_local.py#L52-L55

This would be useful to still support in Zarr v3.

Per @rabernat, this was almost certainly not intentional.

_Originally posted by @shoyer in https://github.com/zarr-developers/zarr-python/discussions/3410_

shoyer:
Fixes #3411

I use the standard strategy of writing to a temporary file in the same directory, and then renaming it to the desired name. This ensure that Zarr writes are either complete or not written at all.

It would be nice if partial writes (`_put()` with `start`) could also be atomic, but I'm honestly not sure if both atomic and efficient partial writes of files are possible. Interestingly, I don't see any uses of `set_partial_values()` inside Zarr, so maybe this feature isn't needed after all? (see https://github.com/zarr-developers/zarr-python/issues/2859 for discussion)

TODO:

### DIFF
--- a/src/zarr/storage/_local.py
+++ b/src/zarr/storage/_local.py
@@ -1,11 +1,15 @@
 from __future__ import annotations
 
 import asyncio
+from collections.abc import Iterator
+import contextlib
 import io
 import os
 import shutil
+import sys
 from pathlib import Path
-from typing import TYPE_CHECKING, Self
+from typing import TYPE_CHECKING, Literal, BinaryIO, Self
+import uuid
 
 from zarr.abc.store import (
     ByteRequest,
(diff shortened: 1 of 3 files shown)


src/zarr/storage/_local.py
```diff
@@ -41,27 +44,55 @@ def _get(path: Path, prototype: BufferPrototype, byte_range: ByteRequest | None)
...
+
 def _put(
     path: Path,
     value: Buffer,
     start: int | None = None,
     exclusive: bool = False,
 ) -> int | None:
     path.parent.mkdir(parents=True, exist_ok=True)
+    # write takes any object supporting the buffer protocol
+    view = value.as_buffer_like()
```
d-v-b, line 88: this makes me wonder why `Buffer` doesn't implement the buffer protocol!

d-v-b (approved):
we should probably get a [release note](https://zarr.readthedocs.io/en/stable/developers/contributing.html#changelog) for this change. I'm happy to write one if that works. Otherwise it looks g2g!

shoyer:
The missing code coverage is for the unused/untested `set_partial_values` (https://github.com/zarr-developers/zarr-python/issues/2859) and for `os.rename` on Windows (I guess Codecov does not include the Windows build?)

PR #3413, opened 2025-08-28
Remove Store.set_partial_writes

### LINKED ISSUE #2859: StoreTests don't test `Store.set_partial_values`

shoyer:
This feature was unused by the rest of Zarr-Python, and was only implemented for LocalStore and stores that wrap other stores.

The Zarr v3 spec still mentions partial writes, so it should probably also be updated.

Fixes #2859

TODO:

### DIFF
--- a/src/zarr/storage/_common.py
+++ b/src/zarr/storage/_common.py
@@ -163,7 +163,7 @@ class StorePath:
             prototype = default_buffer_prototype()
         return await self.store.get(self.path, prototype=prototype, byte_range=byte_range)
 
-    async def set(self, value: Buffer, byte_range: ByteRequest | None = None) -> None:
+    async def set(self, value: Buffer) -> None:
         """
         Write bytes to the store.
 
@@ -171,16 +171,7 @@ class StorePath:
         ----------
         value : Buffer
             The buffer to write.
-        byte_range : ByteRequest, optional
-            The range of bytes to write. If None, the entire buffer is written.
-
-        Raises
-        ------
-        NotImplementedError
-            If `byte_range` is not None, because Store.set does not support partial writes yet.
         """
-        if byte_range is not None:
-            raise NotImplementedError("Store.set does not have partial writes yet")
         await self.store.set(self.path, value)
 
     async def delete(self) -> None:
--- a/changes/2859.removal.rst
+++ b/changes/2859.removal.rst
@@ -1,0 +1,2 @@
+The ``Store.set_partial_writes`` method, which was not used by Zarr-Python, has been removed.
+``store.supports_partial_writes`` is now always ``False``.
--- a/src/zarr/abc/store.py
+++ b/src/zarr/abc/store.py
@@ -12,7 +12,6 @@ if TYPE_CHECKING:
     from typing import Any, Self, TypeAlias
 
     from zarr.core.buffer import Buffer, BufferPrototype
-    from zarr.core.common import BytesLike
 
 __all__ = ["ByteGetter", "ByteSetter", "Store", "set_or_delete"]
 
--- a/src/zarr/abc/store.py
+++ b/src/zarr/abc/store.py
@@ -4,7 +4,7 @@ from abc import ABC, abstractmethod
 from asyncio import gather
 from dataclasses import dataclass
 from itertools import starmap
-from typing import TYPE_CHECKING, Protocol, runtime_checkable
+from typing import TYPE_CHECKING, Literal, Protocol, runtime_checkable
 
 if TYPE_CHECKING:
     from collections.abc import AsyncGenerator, AsyncIterator, Iterable
@@ -310,25 +310,12 @@ class Store(ABC):
         ...
 
     @property
-    @abstractmethod
-    def supports_partial_writes(self) -> bool:
-        """Does the store support partial writes?"""
-        ...
+    def supports_partial_writes(self) -> Literal[False]:
+        """Does the store support partial writes?
 
-    @abstractmethod
-    async def set_partial_values(
-        self, key_start_values: Iterable[tuple[str, int, BytesLike]]
-    ) -> None:
-        """Store values at a given key, starting at byte range_start.
-
-        Parameters
-        ----------
-        key_start_values : list[tuple[str, int, BytesLike]]
-            set of key, range_start, values triples, a key may occur multiple times with different
-            range_starts, range_starts (considering the length of the respective values) must not
-            specify overlapping ranges for the same key
+        Partial writes are no longer used by Zarr, so this is always false.
         """
-        ...
+        return False
 
     @property
     @abstractmethod
@@ -499,7 +486,7 @@ class ByteSetter(Protocol):
         self, prototype: BufferPrototype, byte_range: ByteRequest | None = None
     ) -> Buffer | None: ...
 
-    async def set(self, value: Buffer, byte_range: ByteRequest | None = None) -> None: ...
+    async def set(self, value: Buffer) -> None: ...
 
     async def delete(self) -> None: ...
 
--- a/src/zarr/storage/_fsspec.py
+++ b/src/zarr/storage/_fsspec.py
@@ -89,7 +89,6 @@ class FsspecStore(Store):
     allowed_exceptions
     supports_writes
     supports_deletes
-    supports_partial_writes
     supports_listing
 
     Raises
@@ -113,7 +112,6 @@ class FsspecStore(Store):
     # based on FSSpec
     supports_writes: bool = True
     supports_deletes: bool = True
-    supports_partial_writes: bool = False
     supports_listing: bool = True
 
     fs: AsyncFileSystem
--- a/src/zarr/storage/_fsspec.py
+++ b/src/zarr/storage/_fsspec.py
@@ -26,7 +26,6 @@ if TYPE_CHECKING:
     from fsspec.mapping import FSMap
 
     from zarr.core.buffer import BufferPrototype
-    from zarr.core.common import BytesLike
 
 
 ALLOWED_EXCEPTIONS: tuple[type[Exception], ...] = (
--- a/src/zarr/storage/_fsspec.py
+++ b/src/zarr/storage/_fsspec.py
@@ -418,12 +418,6 @@ class FsspecStore(Store):
 
         return [None if isinstance(r, Exception) else prototype.buffer.from_bytes(r) for r in res]
 
-    async def set_partial_values(
-        self, key_start_values: Iterable[tuple[str, int, BytesLike]]
-    ) -> None:
-        # docstring inherited
-        raise NotImplementedError
-
     async def list(self) -> AsyncIterator[str]:
         # docstring inherited
         allfiles = await self.fs._find(self.path, detail=False, withdirs=False)
--- a/src/zarr/storage/_local.py
+++ b/src/zarr/storage/_local.py
@@ -77,23 +77,12 @@ def _atomic_write(
         raise
 
 
-def _put(
-    path: Path,
-    value: Buffer,
-    start: int | None = None,
-    exclusive: bool = False,
-) -> int | None:
+def _put(path: Path, value: Buffer, exclusive: bool = False) -> int:
     path.parent.mkdir(parents=True, exist_ok=True)
     # write takes any object supporting the buffer protocol
     view = value.as_buffer_like()
-    if start is not None:
-        with path.open("r+b") as f:
-            f.seek(start)
-            f.write(view)
-        return None
-    else:
-        with _atomic_write(path, "wb", exclusive=exclusive) as f:
-            return f.write(view)
+    with _atomic_write(path, "wb", exclusive=exclusive) as f:
+        return f.write(view)
 
 
 class LocalStore(Store):
@@ -111,14 +100,12 @@ class LocalStore(Store):
     ----------
     supports_writes
     supports_deletes
-    supports_partial_writes
     supports_listing
     root
     """
 
     supports_writes: bool = True
     supports_deletes: bool = True
-    supports_partial_writes: bool = True
     supports_listing: bool = True
 
     root: Path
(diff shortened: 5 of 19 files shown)


src/zarr/storage/_common.py
```diff
@@ -163,24 +163,15 @@ async def get(
             prototype = default_buffer_prototype()
         return await self.store.get(self.path, prototype=prototype, byte_range=byte_range)
 
-    async def set(self, value: Buffer, byte_range: ByteRequest | None = None) -> None:
+    async def set(self, value: Buffer) -> None:
```
d-v-b, line 166: I'm not sure I understand this change. Local and memory storage does support writing bytes into an existing file, and this would be very important for efficient writes to uncompressed, sharded arrays.

shoyer:
@d-v-b Yes, in theory partial writes would be helpful for writing into sharded arrays. But Cloud object stores (e.g., S3 and GCS) don't support partial writes, so Zarr-Python is never going to be able to make use of partial writes for the most performance critical stores.

As for local and memory stores, I don't think there are good use-cases for sharding, because file system access deos not have the latency issues that make using small chunks problematic.

It is possible that there are some distributed filesystems that support partial writes and would be able to realize some performance benefits. But in general filesystems don't support concurrent writes to the same files, so this could only happen from concurrent writer at a time and would require locking for safety.

Overall, I don't think niche benefits of partial writes outweigh the complexity cost of including them in the Zarr spec. A better model is to treat shards as something that need to be written all at once in all cases.

d-v-b:
> @d-v-b Yes, in theory partial writes would be helpful for writing into sharded arrays. But Cloud object stores (e.g., S3 and GCS) don't support partial writes, so Zarr-Python is never going to be able to make use of partial writes for the most performance critical stores.

For S3, a combination of multi-part upload and multi-part copy operations could be used to implement a byte-range write.

> As for local and memory stores, I don't think there are good use-cases for sharding, because file system access deos not have the latency issues that make using small chunks problematic.

The use of sharding on local storage is generally motivated by a desire to keep the number of files low, which reduces latency for any O(shards) operations, like listing the contents of a directory, or simply copying data with drag and drop. People absolutely use sharding for this, so it's the opposite of niche.

> It is possible that there are some distributed filesystems that support partial writes and would be able to realize some performance benefits. But in general filesystems don't support concurrent writes to the same files, so this could only happen from concurrent writer at a time and would require locking for safety.

> Overall, I don't think niche benefits of partial writes outweigh the complexity cost of including them in the Zarr spec. A better model is to treat shards as something that need to be written all at once in all cases.

What complexity are we talking about here? An extra parameter in `set`?

d-v-b:
also the `zarr.abc.store.Store` is public API, and there are definitely people using this interface downstream of Zarr Python. we can't just remove keyword arguments. If there's a decision to deprecate the `byte_range` parameter in `set`, we need a deprecation cycle.

shoyer:
> For S3, a combination of multi-part upload and multi-part copy operations could be used to implement a byte-range write.

I haven't used S3 in quite a while, but my understanding is that multi-part upload is quite different from partial writes. You have to supply all the partial writes to override a file. You can't upload just one part at a time.

> > Overall, I don't think niche benefits of partial writes outweigh the complexity cost of including them in the Zarr spec. A better model is to treat shards as something that need to be written all at once in all cases.
> 
> What complexity are we talking about here? An extra parameter in `set`?

The complexity is having a core method in Zarr's core `Store` inference (one of [only 7](https://zarr-specs.readthedocs.io/en/latest/v3/core/index.html#abstract-store-interface)) that Zarr doesn't actually use, and one that even in theory could only be supported by a small fraction of stores, with major cavaets around concurrency.

>  also the zarr.abc.store.Store is public API, and there are definitely people using this interface downstream of Zarr Python. we can't just remove keyword arguments.

Sure, we can keep `set_partial_values` around on `LocalStore`, `LoggingStore` and `WrapperStore` for now if you like for now with a deprecation warning, even though it has zero test coverage.

In every other case, calling `set_partial_values` or supplying a `byte_range` parameter in `set` raises `NotImplementedError`. I don't think that's much more useful than `AttributeError`.

d-v-b:
> > For S3, a combination of multi-part upload and multi-part copy operations could be used to implement a byte-range write.
> 
> I haven't used S3 in quite a while, but my understanding is that multi-part upload is quite different from partial writes. You have to supply all the partial writes to override a file. You can't upload just one part at a time.

This is true, but I think you could initiate a multi-part upload, upload the part that needs to be updated, then use [PUT part/copy](https://docs.aws.amazon.com/AmazonS3/latest/API/API_UploadPartCopy.html) for the rest. In this, you are supply some of the bytes locally, and the rest are sourced from an existing object. S3 experts should weigh in on whether this is actually possible.

> The complexity is having a core method in Zarr's core Store inference (one of [only 7](https://zarr-specs.readthedocs.io/en/latest/v3/core/index.html#abstract-store-interface)) that Zarr doesn't actually use, and one that even in theory could only be supported by a small fraction of stores, with major cavaets around concurrency.

While local and memory storage are only 2 of many possible stores, I think they are actually _extremely important_ to Zarr and we should make every effort to ensure that we have really good performance with both these storage backends. For sharded, _uncompressed_ data, separate chunks in a shard can be concurrently written, and I think this is potentially extremely valuable for a lot of applications, as it alleviates one of the big downsides to sharding (loss of write parallelism). And I think selectively writing byte ranges would also be useful for in-memory storage, when we get around to tuning the performance of that.

For this PR, i'd recommend putting a deprecation warning on `set_partial_values` because we definitely don't need that if `set` takes a byte range. My feeling is that we keep the `byte_range` parameter on `set` until it causes concrete problems for someone.

shoyer:
I would be curious what other Zarr maintainers think about the value of keeping an expanded API for this hypothetical use case (sharded stores without compression in local or memory stores).

In my opinion, local and memory stores are very niche, and mostly relevant for testing. There are better file formats than Zarr for uncompressed data that fits on a single machine (e.g., HDF5, NPZ).

Note that its only `StorePath.set` that has a `byte_range` parameter, not `Store.set`.

d-v-b:
> In my opinion, local and memory stores are very niche, and mostly relevant for testing. There are better file formats than Zarr for uncompressed data that fits on a single machine (e.g., HDF5, NPZ).

There are lots of zarr users who never touch cloud data. Such users have even told me that, for them, cloud storage is niche! So local storage is definitely used for more than testing.

And as for memory storage, there are a lot of internal operations that could be refactored as store -> store transformations, where the source store is a memory store (for example, creating a set of metadata documents in memory storage, then copying that memory store to an external store). In this case, we want memory storage to be very performant.

> Note that its only `StorePath.set` that has a `byte_range` parameter, not `Store.set`.

_This_ is a key point I had overlooked -- only `StorePath.set` (and the `ByteGetter` protocol's `set`) take the byte range. That makes the change in this PR much lower impact than I initially thought. I think we can be confident that nobody is subclassing `StorePath` and implementing `set` with a byte range, and I also think we can ultimately remove the `set_partial_values` from the base store class. The only question is whether we do this immediately, or gradually (via deprecations).

LDeakin:
> I would be curious what other Zarr maintainers think about the value of keeping an expanded API

> Overall, I don't think niche benefits of partial writes outweigh the complexity cost of including them in the Zarr spec.

That part of the spec is [not intended to be normative](https://github.com/zarr-developers/zarr-specs/issues/307), so `zarr-python` does not have to implement it.

> A better model is to treat shards as something that need to be written all at once in all cases.

It certainly is a ~~better~~ *simpler* model, but I thought it worth mentioning that I have found utility in partial writing with [`zarrs`](https://zarrs.dev), at least on *niche local stores* 🥲. The sharding codec supports incrementally (but not concurrently[^1]) updating / appending inner chunks and updating the index of a shard (via [partial encoding](https://docs.rs/zarrs/latest/zarrs/config/struct.Config.html#experimental-partial-encoding)). So, I can independently control:
- the shard shape (to keep inode usage down),
- the write granularity (to limit memory usage when processing), and
- the inner chunk shape / read granularity (for efficient "cloud visualisation" via `neuroglancer` etc).

[^1]: There are major concerns around synchronisation with partial writing. In `zarrs`, the responsibility is placed on the consumer to ensure that partial write operations are not executed concurrently on the same chunk (shard).

src/zarr/storage/_common.py
d-v-b: I was confused here -- I thought this was defined on the `Store` class but actually it's on `StorePath`. (IMO yet another reason to merge `StorePath` with `Store`). That actually makes this change an _improvement_, insofar as it makes the `StorePath` API more like `Store`, and given that we weren't using the byte_range parameter.

d-v-b:
I'm now in favor of the changes here. I think it's very unlikely that this causes any problems for downstream libraries.

If / when we do get around to supporting byte-range writes on niche filesystems :wink: , we can safely change the signature of `set` to `set(key, value, *, byte_range = None)`, i.e. making `byte_range` keyword-only, or we can add a new method as needed. This can be driven by a concrete use case, like intra-shard chunk writes.

d-v-b (approved):

dstansby:
I'm currently in the process of implementing a new store, and noticed that this PR changes the store interface (in a good, simplifying way 😄). I think this means it breaks backwards compatibility for third-party implementaitons of stores, and so it should be made sure this change happens in a minor release (3.2), not a micro release?

I also noted that the change in signature of `get()` is not documented in the release note, which seems quite important for third-parties who implement stores.

shoyer:
> I'm currently in the process of implementing a new store, and noticed that this PR changes the store interface (in a good, simplifying way 😄). I think this means it breaks backwards compatibility for third-party implementaitons of stores, and so it should be made sure this change happens in a minor release (3.2), not a micro release?

I don't think this should break any third party stores, except perhaps type checking on the store interface. If they implement partial setting, that's fine, it's just not used by Zarr (which is the same as the case perviously).

Who it might break is users who use the Zarr store interface for their own proposes, outside of Zarr. But given that only LocalStore implemented partial setting, I would be very surprised if anybody was relying on this.

nbren12:
Just saw this. I actually think writing to a shard incrementally is a nice use case that would be supported by partial writes. I was surprised to find that zarr python was not capable of this. It's quite natural when writing data from e.g. an autoregressive model to a zarr.

> In my opinion, local and memory stores are very niche, and mostly relevant for testing. There are better file formats than Zarr for uncompressed data that fits on a single machine (e.g., HDF5, NPZ).

I must disagree with this. We often use zarr in a large distributed file system. Even if ultimately uploaded to the cloud, the zarr might be created locally first. I'd even guess that this is the predominant way that zarr datasets are made in practice---especially for any simulation based data.


Related to #3560

nbren12:
Also, a main advantage of sharding is that it is safe to write concurrently from multiple workers as long as shards are partitioned across workers. AFAIK this would result in corrupt files with a single hdf5 or npz, so one would need to manually handle this partitioning.


>  But Cloud object stores (e.g., S3 and GCS) don't support partial writes, 

I hadn't realized that. FWIW it does look some some AWS buckets support appends via the [x-amz-write-offset-bytes](https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutObject.html#:~:text=for%20directory%20buckets.-,x%2Damz%2Dwrite%2Doffset%2Dbytes,-Specifies%20the%20offset) header. 

Also, I'm still provoked at the claim that POSIX filesystems are "very niche". It's hard to imagine anything less niche, it's implement in the kernel...

shoyer:
I agree that "append only" writes to parts of Zarr shards would be useful, and avoid concurrency issues as long as the user takes care to use only one worker per shards. This would be a reasonable feature to support!

The main reason why this PR removed `Store.set_partial_writes` is because _it wasn't actually being used_ inside Zarr-Python, so its existence was misleading and a potential waste of time for new `Store` authors. To be clear: No version of Zarr-Python has ever supported partial shard writes, with any storage backend!

If someone figures out how to support append-only writes to Zarr shards from Zarr-Python, we could certainly add back in a `Store` API with support for the necessary form of partial writes. I don't think it would look exactly like `set_partial_writes`, though.

One serious issue with partial writes is how to ensure the Zarr store remains valid while writing, or if a writer crashes mid-shard. When writing a complete file, we can use an atomic update, but that isn't an option when appending. The "index" in the file associated with a [Zarr shard](https://zarr.dev/zeps/accepted/ZEP0002.html#binary-shard-format) is written last, after all the chunks, so an incomplete shard would constistute a corrupt (unreadable) Zarr store. The sane way to handle this would probably be to write the index to a separate file (which can be updated atomically), similar to the [bagz](https://github.com/google-deepmind/bagz) file format, and only append it to the shards file after writing is complete. This would probably require an update to the sharding spec.

LDeakin:
> The sane way to handle this would probably be to write the index to a separate file

Unfortunately it is a complete overhaul of the codec model for a codec to touch more than one file (xref https://github.com/zarr-developers/zarr-specs/issues/346).

The topic of partial shard writing, and even doing it concurrently, was discussed a few times at the [Zarr Summit](https://zarr.dev/zarr-summit-2025/). In short, it is feasible to support concurrent shard writing with some codecs and/or an appropriate runtime flag that ensures that each subchunk has an *uncompacted* unique byte range  to write into in parallel. This needs a compaction step if using compression on the subchunks.

nbren12:
I agree that corrupted files will happen from time to time. This occasionally happens in some pipelines I've developed which shard across zip files ---which also have indexes at the end.

But this can be checked quickly, and usually a shard is not so huge that the work cannot redone. A bad shard could just be treated as missing data.

Some changes to the spec could make the format more robust even while staying with a single file. A few bytes at the beginning of the file containing the offset of the index would do.

d-v-b:
> Just saw this. I actually think writing to a shard incrementally is a nice use case that would be supported by partial writes. I was surprised to find that zarr python was not capable of this. It's quite natural when writing data from e.g. an autoregressive model to a zarr.

and we definitely want to support it! in which case the store class can regrow its support for partial writes

PR #3417, opened 2025-08-30
allow roundtripping the `BytesCodec`

### LINKED ISSUE #3416: `BytesCodec`'s `endian` does not roundtrip
### Zarr version

3.1.1.dev52+gc019a5f26.d20250830 (current main, b8dbf564d5f585e2fcecbfe124273e0dc04a80f4)

### Numcodecs version

0.16.1

### Python Version

3.12.3

### Operating System

ubuntu 24.04

### Installation

git

### Description

If `endian=None` (and thus it is omitted from the json), `BytesCodec.from_dict` replaces it with `endian=<system default>` (by unpacking a empty dict into the kwargs of `BytesCodec.__init__` we get the default for `endian`).

This means that roundtripping `BytesCodec` with `to_dict` and `from_dict` will change the `endian`. This has pretty limited impact, as for all dtypes where omitting the endianness is allowed changing that setting does not affect anything (or maybe somewhere else in the stack there's code that corrects this?). It did pop up when trying to write tests for zarr-developers/VirtualiZarr#787, though.

(I hope I'm not confusing `to_dict` and `from_dict` with "to / from JSON")

Fixing this could be as simple as calling `parsed_configuration.setdefault("endian", None)` in `BytesCodec.from_dict`.

### Steps to reproduce

```python
# /// script
# requires-python = ">=3.11"
# dependencies = [
# ]
# ///

import zarr

from zarr.codecs import BytesCodec

codec = BytesCodec(endian=None)
roundtripped = BytesCodec.from_dict(codec.to_dict())

assert codec.endian == roundtripped.endian, "did not roundtrip the endianness correctly"
```

### Additional output

_No response_

keewis:
closes #3416

Basically, we're changing the default for `endian` when constructing `BytesCodec` using `from_dict`.

(I also renamed `test_endian.py` to `test_bytes.py` as mentioned in https://github.com/zarr-developers/zarr-python/issues/3416#issuecomment-3239172086)

TODO:
* [ ] ~New/modified features documented in `docs/user-guide/*.rst`~


keewis:
looks like some tests actually relied on this behavior, i.e. choosing the default endian if none was specified for multi-byte dtypes (even though this technically violates the spec).

Since this might be potentially user facing (I'm not certain, but at least `virtualizarr` uses `ArrayV3Metadata` directly) someone might be relying on this behavior?

What we _could_ do is start a deprecation cycle by checking for this in `ArrayV3Metadata.from_dict` and warning, then adding the system default. I'll leave the decision to you.

clbarnes:
I agree that this should change (see #4073). Ideally the BytesCodec wouldn't default to the native byte order on regular instantiation but I can see that being a more breaking change; fixing it in from_dict seems a good compromise which would resolve my issue at least.

d-v-b:
@keewis do you mind if I push fixes to resolve the conflicts?

keewis:
not at all, go ahead (and the same for any future PRs I submit)

d-v-b:
@keewis have a look

keewis (commented):
looks good to me.

The only comment I have is, what to do with native endianness? Though maybe that's not an issue because `zarr` doesn't support it (I didn't check). It certainly wouldn't be easy to support for reading on a machine with different endianness.

d-v-b:
> looks good to me.
> 
> The only comment I have is, what to do with native endianness? Though maybe that's not an issue because `zarr` doesn't support it (I didn't check). It certainly wouldn't be easy to support for reading on a machine with different endianness.

there shouldn't be any issues with native endianness because the `bytes` codec describes the endianness that was used for writing bytes. Any other reader that understands the `bytes` codec can then interpret the bytes as they were written, no matter what the native endianness of the platform. Does that make sense? This is also why endianness is not part of zarr v3 data types.

keewis:
yes, it does, thanks.

d-v-b (approved):

Issue #3421, opened 2025-09-01
example that reveals inefficient sharded writes


d-v-b:
Check out how many times we call `get` in this example (writing a single shard with 10 chunks):

```python
# /// script
# requires-python = ">=3.11"
# dependencies = [
#   "zarr@git+https://github.com/zarr-developers/zarr-python.git@main",
# ]
# ///
#
# This script automatically imports the development branch of zarr to check for issues

import zarr
import numpy as np

from zarr.storage._logging import LoggingStore

store  = LoggingStore(store=zarr.storage.MemoryStore())

shape = (10,)
chunks=(1,)
shards=(10,)
data = np.ones(shape)
zarr.create_array(
            store=store,
            data=data,
            chunks=chunks,
            shards=shards,
            fill_value=0,
            overwrite=True,
        )
array = zarr.open_array(store)[:]
```

```
2025-09-01 15:50:45,109 - LoggingStore(memory://4309622656) - INFO -  Calling MemoryStore.read_only
2025-09-01 15:50:45,109 - LoggingStore(memory://4309622656) - INFO - Finished MemoryStore.read_only [0.00 s]
2025-09-01 15:50:45,109 - LoggingStore(memory://4309622656) - INFO -  Calling MemoryStore._ensure_open
2025-09-01 15:50:45,109 - LoggingStore(memory://4309622656) - INFO - Finished MemoryStore._ensure_open [0.00 s]
2025-09-01 15:50:45,110 - LoggingStore(memory://4309622656) - INFO -  Calling MemoryStore.supports_deletes
2025-09-01 15:50:45,110 - LoggingStore(memory://4309622656) - INFO - Finished MemoryStore.supports_deletes [0.00 s]
2025-09-01 15:50:45,110 - LoggingStore(memory://4309622656) - INFO -  Calling MemoryStore.delete_dir
2025-09-01 15:50:45,110 - LoggingStore(memory://4309622656) - INFO - Finished MemoryStore.delete_dir [0.00 s]
2025-09-01 15:50:45,111 - LoggingStore(memory://4309622656) - INFO -  Calling MemoryStore.set(zarr.json)
2025-09-01 15:50:45,111 - LoggingStore(memory://4309622656) - INFO - Finished MemoryStore.set(zarr.json) [0.00 s]
2025-09-01 15:50:45,112 - LoggingStore(memory://4309622656) - INFO -  Calling MemoryStore.get(c/0)
[...]

PR #3422, opened 2025-09-01
fix/shard iteration redux

d-v-b:
Completes the work of #3299 by replacing a second invocation of `_iter_chunk_coords` with `_iter_shard_coords`. 

In a separate PR, we need to refactor this code block: https://github.com/zarr-developers/zarr-python/blob/b8dbf564d5f585e2fcecbfe124273e0dc04a80f4/src/zarr/core/array.py#L4402-L4428. It's massive code smell to have special "write one array to another" logic defined in the tail end of an array creation function.

I add a test that checks how many `get` requests we make when calling `create_array(data=..)`. In `main`, it's 1 `get` per chunk (bad). In this PR, it's 1 `get` per shard (better). But we can also get to 0 `get`s per shard by introducing some special logic for full shard writes. expect this in a later PR.

### DIFF
--- a/src/zarr/core/array.py
+++ b/src/zarr/core/array.py
@@ -4421,7 +4421,7 @@ async def from_array(
 
             # Stream data from the source array to the new array
             await concurrent_map(
-                [(region, data) for region in result._iter_chunk_regions()],
+                [(region, data) for region in result._iter_shard_regions()],
                 _copy_arraylike_region,
                 zarr.core.config.config.get("async.concurrency"),
             )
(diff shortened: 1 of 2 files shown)


maxrjones (approved):
> # Note: we don't actually need one get per shard, but this is the current behavior

Could you open an issue to track this?

d-v-b:
> > Note: we don't actually need one get per shard, but this is the current behavior
> 
> Could you open an issue to track this?

https://github.com/zarr-developers/zarr-python/issues/3421 tracks this

Issue #3406, opened 2025-08-26
zarr.group raises FileNotFoundError


K-Meech:
### Zarr version

latest main

### Numcodecs version

v0.16.1

### Python Version

3.13

### Operating System

Windows

### Installation

pip install into a virtual environment

### Description

On the latest `main` branch, `zarr.group` raises a `FileNotFoundError` when creating a new group - see example code below.

Error:
```
File "C:\Users\kimme\Documents\Repos\forked\zarr-python\src\zarr\storage\_common.py", line 361, in make_store_path
    store = await LocalStore.open(root=store_like, mode=mode)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\kimme\Documents\Repos\forked\zarr-python\src\zarr\storage\_local.py", line 142, in open
    await store._open()
  File "C:\Users\kimme\Documents\Repos\forked\zarr-python\src\zarr\storage\_local.py", line 154, in _open
    raise FileNotFoundError(f"{self.root} does not exist")
FileNotFoundError: data\example.zarr does not exist
```

### Steps to reproduce

```python
# /// script
# requires-python = ">=3.11"
# dependencies = [
#   "zarr@git+https://github.com/zarr-developers/zarr-python.git@main",
# ]
# ///
#
# This script automatically imports the development branch of zarr to check for issues

import zarr
root = zarr.group('data/example.zarr')
```

d-v-b:
can you test this with 3.1.1? I wonder if somehow this slipped in via https://github.com/zarr-developers/zarr-python/pull/3307

K-Meech:
I tried with `3.1.1` and there is no error raised there.

dstansby:
😬 thanks for uncovering this, and it's not great that we don't test for it!

My PR seems to have uncovered an issue in how groups are created; I believe this is the fix, but I don't have time to write a test that fails before the change at the moment, so others please feel free to open a PR with the fix and a test!

```diff
diff --git a/src/zarr/api/asynchronous.py b/src/zarr/api/asynchronous.py
index fdbd3b34..55249236 100644
--- a/src/zarr/api/asynchronous.py
+++ b/src/zarr/api/asynchronous.py
@@ -677,7 +677,7 @@ async def group(
     if overwrite:
         mode = "w"
     else:
-        mode = "r+"
+        mode = "w-"
[...]

PR #3426, opened 2025-09-02
fix/group no error

d-v-b:
This PR adds tests for the error reported in #3406, and makes a change to the mode handling in `zarr.api.asynchronous.group` that resolves the error reported there, but causes tests elsewhere in the test suite to fail. 

I'm posting this incomplete PR so that anyone who enjoys debugging store permissions problems can pick it up.

### DIFF
--- a/src/zarr/api/asynchronous.py
+++ b/src/zarr/api/asynchronous.py
@@ -677,7 +677,7 @@ async def group(
     if overwrite:
         mode = "w"
     else:
-        mode = "a"
+        mode = "w-"
     store_path = await make_store_path(store, path=path, mode=mode, storage_options=storage_options)
 
     if chunk_store is not None:
--- a/src/zarr/api/asynchronous.py
+++ b/src/zarr/api/asynchronous.py
@@ -677,7 +677,7 @@ async def group(
     if overwrite:
         mode = "w"
     else:
-        mode = "r+"
+        mode = "a"
     store_path = await make_store_path(store, path=path, mode=mode, storage_options=storage_options)
 
     if chunk_store is not None:
--- a/tests/test_api/test_asynchronous.py
+++ b/tests/test_api/test_asynchronous.py
@@ -116,3 +116,15 @@ async def test_open_group_new_path(tmp_path: Path) -> None:
     # tmp_path exists, but tmp_path / "test.zarr" will not, which is important for this test
     store = tmp_path / "test.zarr"
     await group(store=store)
+
+
+async def test_open_group_old_path(tmp_path: Path) -> None:
+    """
+    Test that zarr.api.asynchronous.group will not overwrite an existing file.
+
+    See https://github.com/zarr-developers/zarr-python/issues/3406
+    """
+    store = tmp_path
+    await group(store=store, overwrite=True, attributes={"existing_group": True})
+    with pytest.raises(FileExistsError):
+        await group(store=store)
--- a/tests/test_api/test_asynchronous.py
+++ b/tests/test_api/test_asynchronous.py
@@ -8,10 +8,11 @@ import numpy as np
 import pytest
 
 from zarr import create_array
-from zarr.api.asynchronous import _get_shape_chunks, _like_args, open
+from zarr.api.asynchronous import _get_shape_chunks, _like_args, group, open
 from zarr.core.buffer.core import default_buffer_prototype
 
 if TYPE_CHECKING:
+    from pathlib import Path
     from typing import Any
 
     import numpy.typing as npt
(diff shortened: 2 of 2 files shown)


dstansby:
I can take a quick look now - will post back in an hour or so either way if I find any fixes or not 😄

dstansby:
As a starter for 10, de-duplicating `group()` and `create_group()` so one calls the other fixes a bunch of the test failures (and passes the new test):

```diff
diff --git a/src/zarr/api/asynchronous.py b/src/zarr/api/asynchronous.py
index 7ff73cc0..0cde9103 100644
--- a/src/zarr/api/asynchronous.py
+++ b/src/zarr/api/asynchronous.py
@@ -673,13 +673,6 @@ async def group(
 
     zarr_format = _handle_zarr_version_or_format(zarr_version=zarr_version, zarr_format=zarr_format)
 
-    mode: AccessModeLiteral
-    if overwrite:
-        mode = "w"
-    else:
-        mode = "w-"
-    store_path = await make_store_path(store, path=path, mode=mode, storage_options=storage_options)
-
     if chunk_store is not None:
         warnings.warn("chunk_store is not yet implemented", ZarrRuntimeWarning, stacklevel=2)
     if cache_attrs is not None:
@@ -689,20 +682,7 @@ async def group(
     if meta_array is not None:
         warnings.warn("meta_array is not yet implemented", ZarrRuntimeWarning, stacklevel=2)
 
-    if attributes is None:
-        attributes = {}
-
-    try:
-        return await AsyncGroup.open(store=store_path, zarr_format=zarr_format)
-    except (KeyError, FileNotFoundError):
-        _zarr_format = zarr_format or _default_zarr_format()
-        return await AsyncGroup.from_store(
-            store=store_path,
-            zarr_format=_zarr_format,
-            overwrite=overwrite,
-            attributes=attributes,
-        )
-
+    return await create_group(store=store, path=path, overwrite=overwrite, zarr_format=zarr_format, attributes=attributes, storage_options=storage_options)
 
 async def create_group(
     *,
@@ -742,9 +722,7 @@ async def create_group(
     if zarr_format is None:
         zarr_format = _default_zarr_format()
 
-    mode: Literal["a"] = "a"
-
-    store_path = await make_store_path(store, path=path, mode=mode, storage_options=storage_options)
+    store_path = await make_store_path(store, path=path, mode="a", storage_options=storage_options)
 
     return await AsyncGroup.from_store(
         store=store_path,
```

(I will PR my findings to your branch when I'm done, but thought I'd post along the way)

dstansby:
It looks like https://github.com/d-v-b/zarr-python/pull/147 does the trick. I'm not entirely sure why, but it works and deduplicates some code so 🤷

d-v-b:
i'd like to take this opportunity to flag the fact that, collectively, permissions for stores have consumed a lot of developer energy. I would LOVE to find something more literate than arcane string codes like `w-` for expressing our intent in this API.

tests/test_group.py
```diff
@@ -2159,7 +2159,7 @@ def test_group_members_performance(store: Store) -> None:
 
     latency_store = LatencyStore(store, get_latency=get_latency)
     # create a group with some latency on get operations
-    group_read = zarr.group(store=latency_store)
+    group_read = zarr.open_group(store=latency_store)
```
dstansby, line 2162: The change also caught this, which was trying to create a group on top of another group - I changed it to open the existing roup instead.

dstansby:
Related to this PR, it seems like `group()` should be deprecated (and then removed) in favour of `open_group`?

d-v-b:
> Related to this PR, it seems like `group()` should be deprecated (and then removed) in favour of `open_group`?

i'm glad you suggested this! it should absolutely be deprecated.

dstansby:
Dammit I missed some failing tests. I think they can be fixed easily enough by replacing some calls to `group()` (which promises to create a group) with `open_group()`

src/zarr/api/asynchronous.py
```diff
@@ -689,20 +682,7 @@ async def group(
...
-    except (KeyError, FileNotFoundError):
-        _zarr_format = zarr_format or _default_zarr_format()
-        return await AsyncGroup.from_store(
-            store=store_path,
-            zarr_format=_zarr_format,
-            overwrite=overwrite,
-            attributes=attributes,
-        )
-
+    return await create_group(store=store, path=path, overwrite=overwrite, zarr_format=zarr_format, attributes=attributes, storage_options=storage_options)
```
dstansby, line 685: (sorry I realise my comments are scattergun) If `group()` is actually meant to try opening a group first (it's not documented like that, but I guess it's existing behaviour we shouldn't break...?), maybe the try/except statement needs putting back, but the new `create_group` call should go in the `except` block?

d-v-b:
we don't need this any more now that #3431 is in

PR #3432, opened 2025-09-04
Pass read_only into LocalStore.open()

shoyer:
Currently, `read_only` is not passed in `open()`, which means the default value of `read_only=False` is confusingly used instead.

This mostly just surprisng, but also has minor performance implications because `read_only=False` means Zarr attempts to create directories on this line (which is a no-op because `exist_ok=True`): https://github.com/zarr-developers/zarr-python/blob/e738e2fb88dbead26a853d9982ff46eab64f313f/src/zarr/storage/_local.py#L167-L169

TODO:

### DIFF
--- a/src/zarr/storage/_common.py
+++ b/src/zarr/storage/_common.py
@@ -349,7 +349,7 @@ async def make_store_path(
 
     elif isinstance(store_like, Path):
         # Create a new LocalStore
-        store = await LocalStore.open(root=store_like, mode=mode)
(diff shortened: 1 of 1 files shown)


d-v-b:
is it feasible to add a test?

shoyer:
> is it feasible to add a test?

I don't think so, unfortunately, short of excessive mocking. This is just an internal cleanup, to avoid code smell.

d-v-b (approved):
👍

PR #3435, opened 2025-09-05
Deprecate group()

dstansby:
Recently https://github.com/zarr-developers/zarr-python/issues/3406 was a bit of a tangle where `group()` and `create_group()` had different behaviour - this was fixed by making `group()` a simple wrapper of `create_group()`, as intended.

To reduce this API surface, this PR deprecates `group()`, advising users to use either `create_group()` or `open_group()`. The intent of these other functions is clearer, and it mirrors the methods on the `Group` class, which does not have a `group()` method.

I left in one test that uses `group()` to make sure it still works during the deprecation period.

TODO:


d-v-b:
👍

src/zarr/api/asynchronous.py
```diff
@@ -654,6 +654,7 @@ async def array(data: npt.ArrayLike | AnyArray, **kwargs: Any) -> AnyAsyncArray:
     return z
 
 
+@deprecated("Use open_group() or create_group() instead")
```
d-v-b, line 657: Should we state that `group` will be removed in a future release (3.2?)

d-v-b (changes requested):
just 1 question about whether we should be more specific in the deprecation warning, otherwise looks good

src/zarr/api/asynchronous.py
dstansby: 🤷 I'm happy either way. If you'd like me to add a specific release, let me know, and I can update the developer docs too with instructions to do this when deprecating.

d-v-b:
I would like to see this merged -- @dstansby do you mind if I push to this branch?

dstansby:
Please go ahead!

Issue #3433, opened 2025-09-04
Certain extension points of zarr are not implemented


RFLeijenaar:
Zarr is built with extensibility in mind with several extension points defined in the spec. However, zarr-python does not yet implement all these extension points. Notably, chunk grid and chunk key encoding are not extendable yet, and I believe storage transformers aren't even implemented altogether.

I was interested in extending the chunk key encoder to address issues with filesystems that struggle with a large number of files in a single directory. The default V3 implementation uses `/` as the dimension separator, but this is not helpful when you have a single large dim. Sharding will large shards provides a solution, but the current implementation requires full shard rewrites upon writing, even when appending.

I couldn't find any concrete plans to add other extension points. Are there any such plans?

I could help to try and implement this for chunk key encodings, but also realize that the current setup with the registry and global mutable config is not ideal. #3360 

I am curious what your thoughts are.

d-v-b:
Thanks for writing this issue, It's definitely in scope for the project! The main reason we haven't implemented extensibility for chunk key encodings or chunks grids yet is basically just a lack of time. 

Architecturally, I _think_ chunk key encodings should be extensible on the writing side by implementing a new [ChunkKeyEncoding](https://github.com/zarr-developers/zarr-python/blob/e738e2fb88dbead26a853d9982ff46eab64f313f/src/zarr/core/chunk_key_encodings.py#L31), but this will fail on the reading side because we don't have a registry of chunk key encodings. We absolutely should, and it can be really simple -- just a key : value pairing where the keys are names, and the values are chunk key encoding classes. Happy to provide guidance here. 

Chunk grids are more complicated. I suspect our indexing machinery currently assumes regular chunking to some extent. The good news is that this absolutely will be changed when we support irregular chunking (and that's a "when", not an "if").

remcofl:
Thank you for the quick response, and for thinking along!

I can indeed confirm that `ChunkKeyEncoding` is extensible on the write side. When you read, it will parse the named config in [`ChunkKeyEncoding.from_dict()`](https://github.com/zarr-developers/zarr-python/blob/e738e2fb88dbead26a853d9982ff46eab64f313f/src/zarr/core/chunk_key_encodings.py#L41), which will obviously fail.

[...]

PR #3436, opened 2025-09-05
Add registry for chunk key encodings for extensibility

RFLeijenaar:
Following #3433 (@d-v-b), I have implemented a registry for chunk key encodings. This allows users to subclass `ChunkKeyEncoding` and create their own implementation.

The scope of [ChunkKeyEncoding.from_dict()](https://github.com/zarr-developers/zarr-python/blob/e738e2fb88dbead26a853d9982ff46eab64f313f/src/zarr/core/chunk_key_encodings.py#L41) function is reduced to what you expect it do: build from a dict. I placed the parsing function in `chunk_key_encodings.py` as it is used in both array and metadata v3 construction. I wasn't sure where else to put it.

TODO:


remcofl:
Note that the current implementation does not use the global config. This means that it only supports a single implementation (registration) for each chunk key encoding 'type' as indicated by the `name` field. Like codecs ([`parse_codecs`](https://github.com/zarr-developers/zarr-python/blob/3d0e40e171ed91d8315be7e4850c301a2093ca47/src/zarr/core/metadata/v3.py#L55)), it uses the `name` field to retrieve the correct class from the registry. This does mean that the chunk key encoding should always be registered with a qualname that matches the `name` field.

With the entrypoint plugin mechanism this currently doesn't work correctly. It uses the full class path rather than the entrypoint name `e.name` for the key.

I will update the registry to bring it in line with how codecs are implemented, with a dict of registries, that map a chunk key encoding type (name) to a registry that contains possibly multiple implementations for that type.

RFLeijenaar:
Note that the current implementation does not use the global config. This means that it only supports a single implementation (registration) for each chunk key encoding 'type' as indicated by the `name` field. Like codecs ([`parse_codecs`](https://github.com/zarr-developers/zarr-python/blob/3d0e40e171ed91d8315be7e4850c301a2093ca47/src/zarr/core/metadata/v3.py#L55)), it uses the `name` field to retrieve the correct class from the registry. This does mean that the chunk key encoding should always be registered with a qualname that matches the `name` field.

With the entrypoint plugin mechanism this currently doesn't work correctly. It uses the full class path rather than the entrypoint name `e.name` for the key.

But before fixing this, perhaps you think it is better to also use the global config for chunk key encodings?

d-v-b:
> I will update the registry to bring it in line with how codecs are implemented, with a dict of registries, that map a chunk key encoding type (name) to a registry that contains possibly multiple implementations for that type.

You don't need to copy the codec registry. A simple `{name: class}` mapping is fine for chunk key encodings.

d-v-b:
for context, the codec registry associates multiple codec classes with a single codec identifier because of need specific to codecs (running the same codec algorithm on a CPU vs GPU). Chunk key encodings will never be run on specialized hardware, so we can use a simpler mapping.

remcofl:
> for context, the codec registry associates multiple codec classes with a single codec identifier because of need specific to codecs (running the same codec algorithm on a CPU vs GPU). Chunk key encodings will never be run on specialized hardware, so we can use a simpler mapping.

Originally, I implemented it such that it would only allow one implementation, but there were some issues as noted in my other comment. That said, these can be resolved by always using the `name` field as the key, and for any entrypoint `e` setting the key to `e.name`.

Regarding different implementations, it doesn't need to be on special hardware, right? One could also make their own implementation, possibly with bindings in a faster language. Not that I think that this would be common for chunk key encodings.

Let me know if you want me to revert to the simpeler, one implementation only design.

d-v-b:
> Let me know if you want me to revert to the simpeler, one implementation only design.

Yes, if it's not too much work I think this is a better design. I don't see a situation where a chunk key encoding would require specialized hardware, so I don't think we need to copy the codec registry design.

remcofl:
I already have another branch that only allows one implementation, I can move that one into this PR!

In the meantime, I have been using a custom `ChunkKeyEncoding` in one of my repos. It works excellent to create a nested structure to reduce the number of files in a single directory in filesystems that might otherwise become slow:

```python
@dataclass(frozen=True)
class FanoutChunkKeyEncoding(ChunkKeyEncoding):
    name: str = "fanout"
    separator: SeparatorLiteral = "/"
    max_files_per_dir: int = 1000

    def _fanout_coord(self, coord: int) -> list[str]:
        # Break a coordinate into fanout parts
        parts = []
        while coord >= self.max_files_per_dir:
            coord, rem = divmod(coord, self.max_files_per_dir)
            parts.append(str(rem))
        parts.append(str(coord))
        parts.reverse()
        return parts

    def encode_chunk_key(self, chunk_coords: tuple[int, ...]) -> str:
        parts = []
        for dim, coord in enumerate(chunk_coords):
            coords = self._fanout_coord(coord)
            coords = [f"d{dim}"] + coords
            parts.extend(coords)
        parts.append("c")
        return self.separator.join(parts)
```

I believe that right now `decode_chunk_key(self, chunk_key: str)` is unused. Are there any plans to use this function in the future? Otherwise, it might be better to get rid of it now before making the extendibility public.

Are there any other things you would like me to do for this PR?

d-v-b:
> I already have another branch that only allows one implementation, I can move that one into this PR!

Great!

> In the meantime, I have been using a custom `ChunkKeyEncoding` in one of my repos. It works excellent to create a nested structure to reduce the number of files in a single directory in filesystems that might otherwise become slow:

That's very cool, if you're ever interested in sharing this chunk key encoding more widely, you might consider writing it up as a spec in https://github.com/zarr-developers/zarr-extensions/

> I believe that right now `decode_chunk_key(self, chunk_key: str)` is unused. Are there any plans to use this function in the future? Otherwise, it might be better to get rid of it now before making the extendibility public. 

I haven't thought much about this, but I think we would only need `decode_chunk_key` in a situation when we have a random chunk key and we need to know which chunk it contains. I'm pretty sure this scenario never happens in the normal operation of this library, so you're probably right that this should be eventually removed from the public API. Defining chunk decoding is useful for testing but I don't think we need it as a public method.

> Are there any other things you would like me to do for this PR?

other than moving to the simpler registry design, this looks good!

remcofl:
I have overwritten this PR with the other branch (it largely had the same commit history).

> I haven't thought much about this, but I think we would only need decode_chunk_key in a situation when we have a random chunk key and we need to know which chunk it contains. I'm pretty sure this scenario never happens in the normal operation of this library, so you're probably right that this should be eventually removed from the public API. Defining chunk decoding is useful for testing but I don't think we need it as a public method.

Yes, which could be useful to iterate/find the current chunks in the array store. But I believe it is almost always safer to then just iterate the chunk coords, encode, and see which chunk keys are present.

Anyway, I left the `decode_chunk_key` in there, but it is no longer an `@abstractmethod`, and raises `NotImplementedError` by default. Note, this will never be raised by the current library. On the other hand, a subclass that is instantiated enforces the implementation of `encode_chunk_key`. I added docstrings to make subclassing clear to the user. It can be considered later if `decode_chunk_key` should be removed altogether.

> That's very cool, if you're ever interested in sharing this chunk key encoding more widely, you might consider writing it up as a spec in https://github.com/zarr-developers/zarr-extensions/

Thanks, I will consider writing a spec for it. I was also thinking about a hashed key encoding, which could improve load balancing on distributed storage such as s3 or gcs (see [zarr-specs #115](https://github.com/zarr-developers/zarr-specs/issues/115)). But I will have to test which array/chunk/read shape configurations would actually cause a bottleneck to see how useful it is.

src/zarr/core/chunk_key_encodings.py
```diff
@@ -95,10 +80,38 @@ def encode_chunk_key(self, chunk_coords: tuple[int, ...]) -> str:
 @dataclass(frozen=True)
 class V2ChunkKeyEncoding(ChunkKeyEncoding):
     name: Literal["v2"] = "v2"
```
d-v-b: Would annotating this as a `ClassVar` work here? Then we wouldn't need to redundantly pass a `name` parameter to `__init__`

remcofl: `name` should not be an instance configurable parameter but a class constant. Agreed, `ClassVar` works perfectly here.

src/zarr/core/chunk_key_encodings.py
```diff
@@ -28,60 +29,44 @@ class ChunkKeyEncodingParams(TypedDict):
 
 
 @dataclass(frozen=True)
-class ChunkKeyEncoding(Metadata):
+class ChunkKeyEncoding(ABC, Metadata):
     name: str
     separator: SeparatorLiteral = "."
```
remcofl: It might be better to remove `separator` as a standard configuration parameter. For example, a hash-based encoding does not need a separator.

d-v-b: yes, the base ABC probably only needs to require a `name` attribute, and everything else can be methods

d-v-b:
this looks good to me @RFLeijenaar. could you write up some release notes? then we can get this in the next release

d-v-b (approved):

d-v-b:
thanks @RFLeijenaar!

PR #3443, opened 2025-09-09
Refactor `AsyncGroup.open()`

dstansby:
This is a refactor of `AsyncGroup.open()` for readbility/logic simplification:

- Factor out the code for guessing the Zarr version from a store to the top of the method
- This then allows the number of if branches to be reduced, so all the v2 code is in one branch and all the v3 code is in another branch.

### DIFF
--- a/src/zarr/core/group.py
+++ b/src/zarr/core/group.py
@@ -488,7 +488,8 @@ class AsyncGroup:
         zarr_format: ZarrFormat | None = 3,
         use_consolidated: bool | str | None = None,
     ) -> AsyncGroup:
-        """Open a new AsyncGroup
+        """
+        Create a new AsyncGroup from an existing group.
 
         Parameters
         ----------


d-v-b:
> If `zarr_format=None`, guess the Zarr format and then call `AsyncGroup.open()` recursively with the guessed format.

it's a bit weird that `AsyncGroup.open()` takes a zarr_format parameter -- `AsyncGroup` instances should be v2 or v3 only

src/zarr/core/group.py
```diff
@@ -512,6 +536,14 @@ async def open(
             to load consolidated metadata from a non-default key.
         """
         store_path = await make_store_path(store)
+        # Guess zarr_format if not passed explicitly
+        if zarr_format is None:
+            zarr_format = await _get_zarr_version(store_path)
+            return await cls.open(
+                store=store, zarr_format=zarr_format, use_consolidated=use_consolidated
```
d-v-b: this is a bit inefficient from an IO perspective, since `_get_zarr_version` does IO against metadata documents, and `cls.open` will do more IO against the same documents. 

I think it's more efficient to:
- avoid recursion here
- instead of using a function that checks if objects exist (`_get_zarr_version`), use a function that returns an existing array or group directly. This is a more efficient use of IO. I think [`get_node`](https://github.com/zarr-developers/zarr-python/blob/ee9c182a17a3851e7b360275dbf13bb323f1d922/src/zarr/core/group.py#L3652-L3678), or one of the functions it relies on, could be used here.

dstansby: Ah, good point about the IO operations. I naively assumed a `exists()` operation would be much cheaper than `get()`, but I guess on remote stores that's not the case. I've refactored to avoid recursion, and minimise IO. When working out whether a group is v2/v3 there's now two less IO operations, which is a nice win.

dstansby: > instead of using a function that checks if objects exist (_get_zarr_version), use a function that returns an existing array or group directly. This is a more efficient use of IO. I think [get_node](https://github.com/zarr-developers/zarr-python/blob/ee9c182a17a3851e7b360275dbf13bb323f1d922/src/zarr/core/group.py#L3652-L3678), or one of the functions it relies on, could be used here.

I'm not sure what you meant by this?

d-v-b: sorry if I wasn't clear. one of the outcomes of `open` is to return an existing group (and if there's an existing array, or an existing group with the wrong zarr format, we should error). So instead of using a function that _just_ infers the zarr_format of an existing array / group, it's likely simpler to simply try and open an array / group, and then handle the result of that operation.

dstansby: This is called further down the stack than the "high level" functions to open groups, so I don't think that would work? e.g., for `zarr.open_group()` the call stack is:

```python
zarr.open_group()
zarr.api.asynchronous.open_group()
AsyncGroup.open()
```

dstansby:
> it's a bit weird that AsyncGroup.open() takes a zarr_format parameter -- AsyncGroup instances should be v2 or v3 only

Agreed, but it's a `classmethod` so I guess it sort of makes sense? Maybe it should really just be a function.

TomAugspurger:
I think the raw number of I/O operations is less important that the number of synchronization points where we have to wait for I/O to complete. Thanks to concurrency, it making two reads concurrently is faster than two serial reads; making two (small) reads concurrently is about as fast as making one small read.


It looks like for the common case of `zarr_format=None`, users will need to await [one set of reads](https://github.com/zarr-developers/zarr-python/pull/3443/files#diff-2ad4cb8e97a61d216098a06921255a4be58352e51ae4a857b731b8e12080fea4R523-R529) until completion to guess the version, followed by a second await for the group metadata (and if reading this correctly, a third await for consolidated metadata with V2 data). Previously, that was all done in one shot.

IMO, minimizing the number of points where we await I/O is worth the complexity here.

dstansby:
Thanks, I didn't consider the concurrency. I think there's still room for an improved refactor, but I think it would be different enough to the current state here that I'll close this.

Issue #3114, opened 2025-06-03
Inconsistent & Unexpected Behaviour: open_group raises AssertionError if Array exists


FabricioArendTorres:
### Zarr version

3.0.8

### Numcodecs version

0.15.1

### Python Version

3.11.11

### Operating System

Linux, Dockerfile based on python:3.11-slim

### Installation

poetry

### Description

I think there is some unexpected behavior that occurs if one

1. creates an array in the root of a store
2. attempts to open a group in the root of this store in `r` or `r+` mode.

I would have expected some meaningful error, such as the `zarr.errors.ContainsArrayError`, or `FileNotFoundError`, but one just gets a very vague `AssertionError`.
I would like to avoid catching such a broad error in my code.

Below you can find a reproducible test code (tested with `uv run`)  that highlights the expected and actually occuring behaviour.

The `test_open_group_in_dir_with_arrray` fails with the following stacktrace:
```python
==================================================================== FAILURES ====================================================================
_______________________________________________________ test_open_group_in_dir_with_array ________________________________________________________

    def test_open_group_in_dir_with_array():
        store = zarr.storage.MemoryStore()
    
        # this passes
        with pytest.raises(FileNotFoundError):
            zarr.open_group(store, mode="r+")
    
        # now create an array within root
        zarr.create_array(store, shape=(2,), dtype=float)
    
        # this somehow raises an assertion error ?
        # I would expect zarr.errors.ContainsArrayError or FileNotFoundError.
        # with pytest.raises(AssertionError):
        with pytest.raises((zarr.errors.ContainsArrayError, FileNotFoundError)):
>           zarr.open_group(store, mode="r+")

zarr_bug.py:28: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/home/vscode/.cache/uv/environments-v2/zarr-bug-4f65dcfdf7520214/lib/python3.11/site-packages/zarr/_compat.py:43: in inner_f
    return f(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^
[...]

PR #3444, opened 2025-09-09
Raise helpful errors when opening a group with wrong node type.

dstansby:
Fixes https://github.com/zarr-developers/zarr-python/issues/3114


src/zarr/core/group.py
```diff
@@ -673,6 +674,11 @@ def from_dict(
         store_path: StorePath,
         data: dict[str, Any],
     ) -> AsyncGroup:
+        node_type = data.pop("node_type", None)
+        if node_type == "array":
+            raise ContainsArrayError(store_path.store, store_path.path)
```
d-v-b: i recommend specifying the exception message explicitly as a single argument instead of relying on the automatic formatting.

dstansby: I thought because of the way these errors are defined they have to be (and are designed to be) called with two arguments instead of the actual message?

d-v-b: as of #3403 these exceptions can take a single argument, in which case they act like normal exceptions. you can still do the 2-argument pre-formatted version but I don't recommend it.

dstansby: 👍 I added custom messages

d-v-b (approved):

PR #3452, opened 2025-09-14
Feature: support uvloop as a faster alternative to Python's default IO loop

### LINKED ISSUE #3451: Feature Proposal: Add optional uvloop support for improved async performance
## Summary

Add optional uvloop integration to zarr-python to improve I/O performance for async operations.

## Motivation

zarr-python performs many I/O operations, especially when working with remote storage backends (S3, GCS, etc.). uvloop is a fast, drop-in replacement for Python's default asyncio event loop that can provide significant performance improvements for I/O-intensive workloads.

## Proposed Solution

- Automatically use uvloop when available and supported (non-Windows platforms)
- Make it configurable via `async.use_uvloop` config setting
- Add uvloop as an optional dependency: `pip install 'zarr[optional]'`
- Gracefully fall back to standard asyncio when uvloop is unavailable
- No breaking changes - fully backward compatible

## Expected Benefits

- Improved performance for remote storage operations
- Better concurrent I/O handling
- Faster network-bound workloads

## Configuration

```python
# Enable (default when available)
zarr.config.set({"async.use_uvloop": True})

# Disable
zarr.config.set({"async.use_uvloop": False})
```

```shell
# Via environment variable
export ZARR_ASYNC__USE_UVLOOP=false
```

## Implementation Notes

- uvloop only works on Unix-like systems (Linux, macOS)
- Automatic fallback on Windows
- Zero configuration required - works out of the box when installed
- uvloop will be an optional dependency 

This enhancement would provide measurable performance improvements for users working with remote zarr stores while maintaining full compatibility with existing code.

jhamman:
Adds support for uvloop as a faster alternative to Python's default IO loop.

closes #3451

TODO:


src/zarr/core/config.py
```diff
@@ -107,7 +107,7 @@ def enable_gpu(self) -> ConfigSet:
                 "order": "C",
                 "write_empty_chunks": False,
             },
-            "async": {"concurrency": 10, "timeout": None},
+            "async": {"concurrency": 10, "timeout": None, "use_uvloop": True},
```
d-v-b: is uvloop the _only_ alternative event loop, or might there be others? If so, it might make more sense to have a field named `"event_loop"` which takes values `"asyncio" | "uvloop"`

d-v-b:
is there a good way to demo the performance difference between uvloop and asyncio? If so, would it make sense to include that demo in the `examples` directory?

src/zarr/core/config.py
jhamman: I asked myself this exact question and concluded we are very unlikely to have another loop anytime soon. I also asked Claude:

> Great question! Yes, there are a few other event loop implementations in the Python ecosystem, though uvloop is by far the most popular and mature:
> 
> Other Event Loop Implementations
> 
> 1. uvloop - The most popular, based on libuv (what we're implementing)
> 2. asyncio-dgram - UDP-focused event loop
> 3. winloop - Windows-specific attempt (less mature, not widely adopted)
> 4. ProactorEventLoop - Windows-specific (already built into asyncio)
> 5. SelectorEventLoop - Unix default (already built into asyncio)
> 
> However, looking at the ecosystem:
> 
> - uvloop is the dominant alternative - 99% of users who want a replacement
> use uvloop
> - Other alternatives are niche - focused on specific use cases or platforms
> - Most are experimental or abandoned - uvloop has won the "alternative event
> loop" space
> 
> Config Design Considerations
> 
> Option 1: Boolean approach (current)
> 
> zarr.config.set({"async.use_uvloop": True})
> 
> Pros: Simple, clear intent, matches the reality that uvloop is the only practical alternative
> Cons: Not extensible if other loops emerge
> 
> Option 2: String-based approach
> 
> zarr.config.set({"async.loop": "uvloop"})  # or "asyncio", "auto"
> Pros: More extensible, could support future loop implementations
> Cons: More complex, premature abstraction for a problem that may never exist
> 
> Option 3: Hybrid approach
> 
> zarr.config.set({"async.loop_policy": "uvloop"})  # or "default", "auto"
> 
> Recommendation: Stick with Boolean
> 
> I recommend keeping the current async.use_uvloop boolean approach for these
> reasons:
> 
> 1. YAGNI principle - We shouldn't over-engineer for hypothetical future needs
> 2. Ecosystem reality - uvloop has been the only serious asyncio alternative for 7+ years
> 3. Clear semantics - Boolean is unambiguous about what it does
> 4. Easy migration path - If other loops emerge, we could add async.loop_policy later and deprecate use_uvloop
> 5. Precedent - Most libraries that integrate uvloop use similar boolean flags
> 
> The boolean approach is pragmatic and matches the current ecosystem reality.
> If other event loop implementations gain traction in the future, we could
> always add a more generic async.loop_policy setting while keeping use_uvloop
> for backward compatibility.

DimitriPapadopoulos:
Note that asyncio seems to have been improved in recent versions of Python, almost filling the gap with uvloop (see https://github.com/MagicStack/uvloop/issues/566 and https://github.com/MagicStack/uvloop/pull/644).

PR #3486, opened 2025-09-23
Make ObjectStore a generic class

dstansby:
This makes `ObjectStore` generic with respect to the underlying store, mirroring similar structure in `LoggingStore` and `WrapperStore`. This allows type checkers to infer the type of the underlying store:

```python

import obstore.store

store = ObjectStore(obstore.store.MemoryStore())
reveal_type(store.store)
# Before: Revealed type is "Union[obstore.store.AzureStore, obstore.store.GCSStore, obstore.store.HTTPStore, obstore.store.S3Store, obstore.store.LocalStore, obstore.store.MemoryStore]"
# After:  Revealed type is "obstore.store.MemoryStore"
```

### DIFF
--- a/src/zarr/storage/_obstore.py
+++ b/src/zarr/storage/_obstore.py
@@ -4,7 +4,7 @@ import asyncio
 import contextlib
 import pickle
 from collections import defaultdict
-from typing import TYPE_CHECKING, TypedDict
+from typing import TYPE_CHECKING, Generic, Self, TypedDict, TypeVar
 
 from zarr.abc.store import (
     ByteRequest,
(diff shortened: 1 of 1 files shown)


d-v-b:
cc @kylebarron

kylebarron:
This is fine code-wise but why do you need to go back to the underlying object store once you've constructed the zarr store?

dstansby:
I'm not 100% sure, but at least in the tests here in `zarr-python` the underlying store is accessed. My motiviation for this was fixing typing in our test suite here, so there may well not be a real world use case. Definitely happy to be guided by those who use/develop `ObjectStore` on whether this change is worth making or not.

As a related aside, if there is no need for users to access the underlying store would it be worth deprecating `.store` and renaming to `._store` to signal that they shouldn't need to access it?

kylebarron:
I think it's fine to add the generic. I don't think it exposes too much internal details.

d-v-b (approved):

PR #3490, opened 2025-09-25
add experimental module

### LINKED ISSUE #3484: add an experimental module
We should have a module for new, additive features that are potentially useful but also susceptible to major design changes or even removal. 

I think the caching store proposed in #3366 falls into this category -- I could see many different outcomes with that feature -- maybe caching becomes basic functionality for all zarr stores, or maybe the cache store gets spun off into its own library. 

Experience with this feature would be super valuable, but we need to gain that experience _safely_, in a way that's visible to users, (ideally without more annoying warnings!). So I offer two ideas:

1. A top-level `zarr.experimental` module that contains all the experimental stuff. The cache store would be exported via `zarr.experimental.CacheStore`.
2. A storage-level `zarr.storage.experimental` module that contains experimental storage APIs. The cache store would be exported via `zarr.storage.experimental.CacheStore`

I don't feel strongly about either of these ideas, but I would like to pick one, implement it, and get #3366 shipped as soon as we can.

d-v-b:
Adds a `zarr.experimental` module. We can use this module to deliver new features that might be useful but might also evolve significantly in the future. The "experimental" designation should temper user expectations about stability.

closes #3484

### DIFF
--- a/changes/3490.feature.md
+++ b/changes/3490.feature.md
@@ -1,0 +1,1 @@
+Adds a `zarr.experimental` module for unstable user-facing features.
--- a/src/zarr/experimental/__init__.py
+++ b/src/zarr/experimental/__init__.py
@@ -1,0 +1,1 @@
+"""The experimental module is a site for exporting new or experimental Zarr features."""


d-v-b:
@zarr-developers/python-core-devs can someone please review this

Issue #3495, opened 2025-09-30
Failure to list keys on remote HTTP store


ianhi:
### Zarr version

main

### Numcodecs version

0.16.3

### Python Version

3.13

### Operating System

mac

### Installation

pep-723

### Description

Opening a remote zarr store over https can silently fails to list keys

### Steps to reproduce
```python
# /// script
# requires-python = ">=3.11"
# dependencies = [
#   "zarr@git+https://github.com/zarr-developers/zarr-python.git@main",
#   "fsspec",
#   "requests",
#   "aiohttp",
#   "ome-zarr"
# ]
# ///
"""
Minimal reproducer for Zarr remote vs local group listing issue.

This demonstrates that zarr.Group.keys() returns empty list for remote
stores (FsspecStore over HTTP) but works correctly for local stores,
even though direct access (group['0']) works in both cases.

"""

from pathlib import Path

import zarr

url = "https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.5/idr0062A/6001240_labels.zarr"
download_dir = Path("data")
local_path = download_dir / "6001240_labels.zarr"

if not local_path.exists():
    import ome_zarr.utils

    print(f"Downloading {url}...")
    download_dir.mkdir(parents=True, exist_ok=True)
    ome_zarr.utils.download(url, str(download_dir))
    print("Download complete!")
print("\n\n\n")

# actual reproducer

zarr.print_debug_info()
# Test Remote
print("\n1. REMOTE store (FsspecStore over HTTP)")
remote_group = zarr.open_group(url, mode="r")
remote_keys = list(remote_group.keys())
print(f"   keys() → {remote_keys}")
print(f"   Direct access group['0'] → {type(remote_group['0']).__name__}")

# Test Local
print("\n2. LOCAL store (LocalStore)")
print(f"   Path: {local_path}")

local_group = zarr.open_group(str(local_path), mode="r")
local_keys = list(local_group.keys())
print(f"   keys() → {local_keys}")
print(f"   Direct access group['0'] → {type(local_group['0']).__name__}")

# Show the bug
print("\n" + "=" * 80)
print("BUG: Remote keys() returns empty but direct access works!")
print("=" * 80)
print(f"Remote keys(): {remote_keys} (WRONG - should match local)")
print(f"Local keys():  {local_keys}")
print(f"\nBoth can access group['0']: ✓")
print("\nThis breaks xarray when iterating groups for DataTree.")
```

### Additional output

```


[...]

PR #3496, opened 2025-09-30
attempt to autodetect an s3 compatible URL

ianhi:
Addresses part of https://github.com/zarr-developers/zarr-python/issues/3495 by attempting to autodetect if a URL is S3 compatible or not. 
With this change the following "just works"

```python
import zarr
z = zarr.open("https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.5/idr0062A/6001240_labels.zarr", storage_options={"anon":True})
print(list(z.keys()))
```

This is equivalent to:

```python
  zarr.open(
      "s3://idr/zarr/v0.5/idr0062A/6001240_labels.zarr",
      storage_options={
          "anon": True,
          "client_kwargs": {"endpoint_url": "https://uk1s3.embassy.ebi.ac.uk"}
      }
  )
```

but now zarr handles that splitting for the user

* [NA?] Add docstrings and API docs for any new/modified user-facing classes and functions

### DIFF
--- a/src/zarr/storage/_common.py
+++ b/src/zarr/storage/_common.py
@@ -343,6 +343,9 @@ async def make_store(
     elif isinstance(store_like, str):
         # Either a FSSpec URI or a local filesystem path
         if _is_fsspec_uri(store_like):
+            # Check if this is an S3-compatible store accessed via HTTPS
+            # and convert it to use the s3:// protocol for proper support
+            store_like, storage_options = _maybe_convert_http_to_s3(store_like, storage_options)
             return FsspecStore.from_url(
                 store_like, storage_options=storage_options, read_only=_read_only
             )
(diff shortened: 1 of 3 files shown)


src/zarr/storage/_common.py
```diff
@@ -439,6 +442,99 @@ def _is_fsspec_uri(uri: str) -> bool:
...
+        return part == "s3" or part.startswith("s3-") or part.endswith("s3")
+
+    is_likely_s3 = (
+        any(has_s3_token(part) for part in hostname_parts)
+        or "object-store" in hostname  # Less likely to have false positives
+        or "objectstore" in hostname
+        or "minio" in hostname_parts  # minio.example.com
+        or "ceph" in hostname_parts  # ceph.example.com
+        or "rgw" in hostname_parts  # rgw.example.com (Ceph RADOS Gateway)
+    )
```
d-v-b, line 502: how reliable is this?

d-v-b:
@martindurant and @kylebarron relevant to your interests.

martindurant:
I don't understand: why would someone provide an HTTP URL for something that you need the S3 protocol for? The storage options only make sense to s3, not http, so the caller already knows.

d-v-b:
I'm inclined against the changes in this PR -- `"https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.5/idr0062A/6001240_labels.zarr"` is unambiguously an https URL, and so it makes sense to use an HTTP-based storage backend to access it. For example, `"https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.5/idr0062A/6001240_labels.zarr/zarr.json"` works just fine. Directory listing does _not_ work, but that's pretty common for zarr data over http.

ianhi:
I ran into this today when trying to load data from: https://idr.github.io/ome-ngff-samples/

the "copy s3 url" button  gives a one line URL: https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.5/idr0062A/6001240_labels.zarr

which I naively expected to work in zarr. So this PR is basically protection against that situation for other naive users possibly sourcing URLs that look like that. 


Ideally zarr would just work and protect me from this, or at least it should have thrown an error, rather silently failing by not returning any data but otherwise working

d-v-b:
> the "copy s3 url" button gives a one line URL: https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.5/idr0062A/6001240_labels.zarr

IMO a button that says "copy s3 url" should give you an s3 url, not an http url

ianhi:
> IMO a button that says "copy s3 url" should give you an s3 url, not an http url

I agree! but I suspect this is not the only place this is happening and we can't fix it everywhere

Also is it possible to specify the endpoint url and bucket and path in one s3: url?

martindurant:
Exactly what @d-v-b says. That URL is just wrong, it doesn't point to anything! If you poke it, the returned headers (of the 404) does contain an "x-amz-*" key, but that might be true from a real website that is hosted somewhere within AWS.

ianhi:
hmm. ok. So should zarr have raised an error for what I did here: https://github.com/zarr-developers/zarr-python/issues/3495 when it loaded as an http store that didn't fully behave like one?

and am I right that there's no one liner for that website to return becuase you can't specify an endpoint-url within the `s3:` syntax?

ianhi:
I am updating the URLS over there: https://github.com/IDR/ome-ngff-samples/pull/28

going to close this now as I hear the arguments against this change. I would sitll like to find a way to make this more transparent to users but this is clearly not the right approach.

joshmoore:
> and am I right that there's no one liner for that website to return becuase you can't specify an endpoint-url within the `s3:` syntax?

That's my understanding.

Issue #3498, opened 2025-10-01
missing doctests


d-v-b:
I don't think we have doctests working -- the doctest commands were removed from `pyproject.toml` in our switch to mkdocs, and the CI doctest job silently does nothing. We should probably fix this.

d-v-b:
this was partially addressed by #3500, but there are still many instances of code samples in our docstrings that have not been migrated to code-fenced markdown blocks. These all need to be converted before our doctests are complete.

PR #3500, opened 2025-10-02
chore/doctests

d-v-b:
Adds doctests. This PR builds atop #3366, and so should not be merged until #3366 is in.

I add a new test file in `test/test_docs.py` that collects executable examples and runs them, ensuring that no errors are raised. 

The goal here is to go from 0 doctests to some doctests. But we need to ultimately do more than "some doctests". We need tests that compare the output of the executed code against what we expected. I'm happy to take that as a contribution / addition to this PR if it's simple, or we can do that in a subsequent PR.

### DIFF
--- a/src/zarr/core/array.py
+++ b/src/zarr/core/array.py
@@ -983,7 +983,9 @@ class AsyncArray(Generic[T_ArrayMetadata]):
         Examples
         --------
         ```python
+        import asyncio
         import zarr
+        from zarr.core.array import AsyncArray
 
         async def example():
             store = zarr.storage.MemoryStore()
@@ -995,8 +997,8 @@ class AsyncArray(Generic[T_ArrayMetadata]):
             async_arr = await AsyncArray.open(store)
             return async_arr
 
-        # async_arr = await example()
-        # AsyncArray(...)
+        async_arr = asyncio.run(example())
+        # <AsyncArray memory://... shape=(100, 100) dtype=int32>
         ```
         """
         store_path = await make_store_path(store)
@@ -1313,18 +1315,21 @@ class AsyncArray(Generic[T_ArrayMetadata]):
         Examples
         --------
         ```python
+        import asyncio
         import zarr.api.asynchronous
 
         async def example():
-            arr = await zarr.api.asynchronous.create(shape=(10,), chunks=(1,), shards=(2,))
+            arr = await zarr.api.asynchronous.create(shape=(10,), chunks=(1,))
             count = await arr.nchunks_initialized()
-            # 0
+            print(f"Initial: {count}")
+            #> Initial: 0
             await arr.setitem(slice(5), 1)
             count = await arr.nchunks_initialized()
-            # 6
+            print(f"After write: {count}")
+            #> After write: 5
             return count
 
-        # result = await example()
+        result = asyncio.run(example())
         ```
         """
         if self.shards is None:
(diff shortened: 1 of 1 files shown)


d-v-b:
closes #3498

changes/3357.feature.md
```diff
@@ -0,0 +1 @@
+Adds `zarr.experimental.cache_store.CacheStore`, a `Store` that implements caching by combining two other `Store` instances. See the [docs page](https://zarr.readthedocs.io/en/latest/user-guide/cache-store) for more information about this feature.
```
maxrjones: @d-v-b can you please remove this addition? It is already properly documented using the PR number in https://github.com/zarr-developers/zarr-python/blob/main/changes/3366.feature.md

d-v-b:
> @d-v-b do you have a vision for how to match the executed code output to the expected results? Currently, I think this PR is a more complicated version of `mkdocs build --strict` so without a clear path to output comparisons I don't find it super compelling

`mkdocs build --strict` requires mkdocs and is not part of our test suite (i.e., something that you can run via `pytest`). This PR adds _actual tests_, which is the direction we need to go. Unlike checking for build failures, we can extend the approach here to compare output against expected output pretty easily. 

But you reminded me that this PR can be made much simpler with `pytest-examples`. I'll push some changes that simplify things.

d-v-b:
> > @d-v-b do you have a vision for how to match the executed code output to the expected results? Currently, I think this PR is a more complicated version of `mkdocs build --strict` so without a clear path to output comparisons I don't find it super compelling
> 
> `mkdocs build --strict` requires mkdocs and is not part of our test suite (i.e., something that you can run via `pytest`). This PR adds _actual tests_, which is the direction we need to go. Unlike checking for build failures, we can extend the approach here to compare output against expected output pretty easily.
> 
> But you reminded me that this PR can be made much simpler with `pytest-examples`. I'll push some changes that simplify things.

It was not as simple as I thought -- the fact that our docs uses code blocks organized into sessions requires the test machinery in this PR. Since the examples in our docs are not currently set up for output checking, I left those tests as-is, but included a commented code block that, once commented, will check example code output.

I also added a test for examples defined in our docstrings. We have some docstring examples that either require session logic (like the examples in our docs) or require a re-write, so I largely left those alone. I did fix #3503 though, and many other broken docstring examples.

tests/test_docs.py
```diff
@@ -0,0 +1,120 @@
...
+            continue
+        if str(example.path) == file_path and settings.get("session", "_default") == session_name:
+            examples.append(example)
+
+    # Run all examples in this session sequentially, preserving state
+    module_globals: dict[str, object] = {}
+    for example in examples:
+        # TODO: uncomment this line when we are ready to fix output checks
+        # result = eval_example.run_print_check(example, module_globals=module_globals)
+        result = eval_example.run(example, module_globals=module_globals)
```
d-v-b, line 110: @maxrjones switching `eval_example.run` to `eval_example.run_print_check` will turn on output checks for the docs.

src/zarr/core/array.py
```diff
@@ -982,10 +982,24 @@ async def open(
...
+            store = zarr.storage.MemoryStore()
+            # First create an array to open
+            await zarr.api.asynchronous.create_array(
+                store=store, shape=(100, 100), dtype="int32"
+            )
+            # Now open it
+            async_arr = await AsyncArray.open(store)
+            return async_arr
+
+        async_arr = asyncio.run(example())
```
d-v-b, line 1000: our async examples previously used top-level await, which is not generally available in python interpreters. We now use asyncio.run()

Issue #3501, opened 2025-10-02
documenting self-contained examples


d-v-b:
We should have a dedicated section in our docs page for the self-contained examples in the `examples` folder. 
We should also ensure that these examples are self-documenting as much as possible, and I think the best solution there is to put each example in its own directory, and to include a `README.md` file along with the source code. The content of the readme can be re-used in our rendered docs.

d-v-b:
closed by #3502

PR #3502, opened 2025-10-02
docs/examples

d-v-b:
This PR changes the layout of our top-level `examples` directory to be a directory-of-directories, where each sub-directory contains code + a `README.md`, which will make browsing these examples on github friendlier.

I also add a page to the user guide called `examples` that references the source code + readmes for the different examples. Right now we only have 1 example, but I would like to eventually add more, e.g. to demonstrate running benchmarks or integrating zarr with other tools, like visualization tooling.


d-v-b:
closes https://github.com/zarr-developers/zarr-python/issues/3501

mkdocs.yml
```diff
@@ -25,6 +25,8 @@ nav:
       - user-guide/extending.md
       - user-guide/gpu.md
       - user-guide/consolidated_metadata.md
+      - Examples:
+          - user-guide/examples/custom_dtype.md
       - user-guide/experimental.md
```
maxrjones:
```suggestion
      - user-guide/experimental.md
  - Examples:
      - user-guide/examples/custom_dtype.md
```

d-v-b:
> I think these would be easier to find if the examples are separate from user-guide, rather than nested, but support merging if you have a strong preference for the current layout

no strong preference! i took your suggestion.

PR #3526, opened 2025-10-15
feature: update default config setting for async.concurrency from 10->64

jhamman:
This PR updates the default async.concurrency setting in Zarr from 10 to 64. We've repeatedly seen ([example 1](https://earthmover.io/blog/xarray-open-zarr-improvements), [example 2](https://github.com/maxrjones/zarr-obstore-performance)) that Zarr performs much better at larger concurrency settings and the default value of 10 was a hold over from the very early experiments with Zarr-Python 3. 

I talked this over with @d-v-b and @normanrz here at the Zarr Summit and both were cool with the change 😎 

TODO:

### DIFF
--- a/src/zarr/core/config.py
+++ b/src/zarr/core/config.py
@@ -107,7 +107,7 @@ config = Config(
                 "order": "C",
                 "write_empty_chunks": False,
             },
-            "async": {"concurrency": 10, "timeout": None},
+            "async": {"concurrency": 64, "timeout": None},
             "threading": {"max_workers": None},
             "json_indent": 2,
             "codec_pipeline": {
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -54,7 +54,7 @@ def test_config_defaults_set() -> None:
                     "order": "C",
                     "write_empty_chunks": False,
                 },
-                "async": {"concurrency": 10, "timeout": None},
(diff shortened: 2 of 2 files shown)


rabernat:
We should also make this a global limit, not scoped to each chunk request.

dcherian:
> We should also make this a global limit, not scoped to each chunk request.

+1000. In fact i think we should revert this until that change is made; because it's going to cause errors with dask as we've learned with Icechunk.

d-v-b:
> > We should also make this a global limit, not scoped to each chunk request.
> 
> +1000. In fact i think we should revert this until that change is made; because it's going to cause errors with dask as we've learned with Icechunk.

sounds good to me. @jhamman do you want to revert?

jhamman:
I don't exactly understand what the scope of the global limit is. My preference would be to keep this change in AND document a best practice for using Zarr concurrency with Dask AND design a proper global semaphore for Zarr. I would happily lead the documentation effort but the global design is unclear to me at the moment.

dstansby:
Could this get some documentation and a changelog entry too? Recently I was quite confused by the different concurrency options when I wanted to force `zarr-python` to run with no concurrency, so changing/improving the default seems like a good oppurtunity to add some short docs on what this option affects.

PR #3528, opened 2025-10-16
Document changes to async.concurrency defaults and performance considerations.

jhamman:
As requested in https://github.com/zarr-developers/zarr-python/pull/3526, this PR adds some documentation about the change to and performance considerations of the async.concurrency config setting.

TODO:

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,7 +6,7 @@ ci:
 default_stages: [pre-commit, pre-push]
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.13.1
+    rev: v0.13.3
     hooks:
       - id: ruff-check
         args: ["--fix", "--show-fixes"]
@@ -42,7 +42,7 @@ repos:
           - hypothesis
           - s3fs
   - repo: https://github.com/scientific-python/cookie
-    rev: 2025.05.02
+    rev: 2025.10.01
     hooks:
       - id: sp-repo-review
   - repo: https://github.com/pre-commit/pygrep-hooks


d-v-b (approved):

docs/user-guide/performance.md
```diff
@@ -175,7 +175,84 @@ Coming soon.
...
+
+Higher concurrency values can improve throughput when:
+- Working with remote storage (e.g., S3, GCS) where network latency is high
+- Reading/writing many small chunks in parallel
+- The storage backend can handle many concurrent requests
+
+Lower concurrency values may be beneficial when:
+- Working with local storage with limited I/O bandwidth
+- Memory is constrained (each concurrent operation requires buffer space)
+- Using Zarr within a parallel computing framework (see below)
```
dstansby, line 204: This is exactly this use case I ran into 😄 - thanks a lot for putting this together, it's a really nice summary.

ilan-gold: @dstansby Could you share a bit of specifics as to what you saw that might have indicated this setting as the culprit?  

I like having examples, especially :)

dstansby: I wanted to run multiple parallel jobs, each one which writes to a shard, and manually run as many jobs as I had processors.

ilan-gold: Mmm, interesting, and this was causing consistency issues? I have some sharded dask writing but haven't noticed any issues.  2 jobs would try writing to the same shard?

dstansby: Ah, it wasn't causing any data issues, just performance issues. I was manually writing to one shard per process, and because my data was local on a fast SSD async concurrency wouldn't have bought me anything, and the internal multithreading that zarr-python does was causing more threads than I had processors to spin up without me wanting/needing them.

ilan-gold: Mmmm interesting, I think `zarrs` just has a global pool (at least I think that's rayon's default) so that might explain why I haven't hit this issue.  Good to know!

PR #3530, opened 2025-10-18
fix: accept extra fields in array metadata

d-v-b:
The v3 spec requires that we accept metadata documents with extra keys as long as those keys map to values that are JSON objects containing the `{"must_understand": "false"}` key: value pair.

This PR brings Zarr Python into compliance with the spec's requirements for array metadata documents with valid extra fields. I add a new field to `ArrayV3Metadata`, called `extra_fields`, that captures any valid extra fields and inserts them into the top-level key space when we convert the array metadata to dict.

In order to programmatically reason about "metadata key", "extra key", "allowed extra field", etc, this PR introduces typeddict classes that models the structure of v3 array metadata and the structure of allowed extra keys. This change required changes to type hints.

Looking forward, I think we should eventually migrate away from `dict[str, JSON]` as a type annotation and instead adopt a more strictly-typed approach based on typeddicts. 

This PR only changes things for array metadata. I do not fix group metadata here. I will handle the group class in a later PR, because we probably need to introduce a _separate class for v2 and v3 group metadata_, and that's a bit more work than what I have here.

### DIFF
--- a/changes/3532.misc.md
+++ b/changes/3532.misc.md
@@ -1,0 +1,2 @@
+Accept `"bytes"` as an alias for `"variable_length_bytes"` when parsing `JSON`-encoded Zarr V3
+data types.
--- a/changes/3535.bugfix.md
+++ b/changes/3535.bugfix.md
@@ -1,0 +1,2 @@
+Fixed a bug where the `"consolidated_metadata"` key was written to metadata documents even when
+consolidated metadata was not used, resulting in invalid metadata documents.
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -9,7 +9,7 @@
 - Add a command-line interface to migrate v2 Zarr metadata to v3. Corresponding functions are also provided under zarr.metadata. ([#1798](https://github.com/zarr-developers/zarr-python/issues/1798))
 - Add obstore implementation of delete_dir. ([#3310](https://github.com/zarr-developers/zarr-python/issues/3310))
 - Adds a registry for chunk key encodings for extensibility. This allows users to implement a custom `ChunkKeyEncoding`, which can be registered via `register_chunk_key_encoding` or as an entry point under `zarr.chunk_key_encoding`. ([#3436](https://github.com/zarr-developers/zarr-python/issues/3436))
-- Trying to open a group at a path were a array already exists now raises a helpful error. ([#3444](https://github.com/zarr-developers/zarr-python/issues/3444))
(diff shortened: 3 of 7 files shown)


d-v-b:
@zarr-developers/python-core-devs could someone have a look at this? We are violating the v3 spec until we fix this.

src/zarr/core/array.py
```diff
@@ -103,10 +103,10 @@
 from zarr.core.metadata import (
     ArrayMetadata,
     ArrayMetadataDict,
+    ArrayMetadataJSON_V3,
     ArrayV2Metadata,
     ArrayV2MetadataDict,
     ArrayV3Metadata,
-    ArrayV3MetadataDict,
```
maxrjones, line 109: why rename `ArrayV3MetadataDict` to `ArrayMetadataJSON_V3`? It expands the diff a fair bit

d-v-b: to bring the naming convention in line with what we are already doing for the [data type JSON models](https://github.com/zarr-developers/zarr-python/blob/e3ee5912ed0c40a05aca8236de7ad71f02963707/src/zarr/core/dtype/npy/time.py#L125). I've been contemplating removing the trailing underscore from these names, but I'm not sure about that yet, and it would be in a separate PR

maxrjones: I'm not a huge fan of including that change in this PR. While it's bringing consistency between `ArrayMetadataJSON_V3` and the data type JSON models, it's adding more inconsistencies in the name convention for this set of metadata objects. It seems unrelated to the goal of this PR.

d-v-b: prior to his PR, `ArrayV3MetadataDict` was useless, so I think it's safe to rename here. Sorry for all the noise in the diff, but I think it's better to rename it now rather than later

src/zarr/core/chunk_grids.py
```diff
@@ -152,7 +153,7 @@ def normalize_chunks(chunks: Any, shape: tuple[int, ...], typesize: int) -> tupl
 @dataclass(frozen=True)
 class ChunkGrid(Metadata):
     @classmethod
-    def from_dict(cls, data: dict[str, JSON] | ChunkGrid) -> ChunkGrid:
+    def from_dict(cls, data: dict[str, JSON] | ChunkGrid | NamedConfig[str, Any]) -> ChunkGrid:
```
d-v-b, line 156: This union type appears because `TypedDict` classes like `NamedConfig` are assignable to `Mapping[str, object]`, but `Mapping[str, object]` is not assignable to `dict[str, JSON]`.

Issue #3513, opened 2025-10-08
Reading data that was written with deprecated `bytes` codec


rabernat:
Prior to 3.1, it was possible to write an array that looked like this

```python
doc = {
 'shape': [],
 'data_type': 'bytes',
 'chunk_grid': {'name': 'regular', 'configuration': {'chunk_shape': []}},
 'chunk_key_encoding': {'name': 'default',
  'configuration': {'separator': '/'}},
 'fill_value': [],
 'codecs': [{'name': 'vlen-bytes', 'configuration': {}},
  {'name': 'zstd', 'configuration': {'level': 0, 'checksum': False}}],
 'attributes': {},
 'zarr_format': 3,
 'node_type': 'array',
 'storage_transformers': []
}
```

Attempting to load this data errors
```python
import zarr
zarr.core.metadata.ArrayV3Metadata.from_dict(doc)
```

```
File ~/mambaforge/envs/earthmover-demos/lib/python3.12/site-packages/zarr/core/dtype/registry.py:208, in DataTypeRegistry.match_json(self, data, zarr_format)
    206     except DataTypeValidationError:
    207         pass
--> [208](https://file+.vscode-resource.vscode-cdn.net/Users/rabernat/gh/earth-mover/demos/~/mambaforge/envs/earthmover-demos/lib/python3.12/site-packages/zarr/core/dtype/registry.py:208) raise ValueError(f"No Zarr data type found that matches {data!r}")

ValueError: No Zarr data type found that matches 'bytes'
```

The following tweaks make it loadable

```python
doc["data_type"] = "variable_length_bytes"
doc["fill_value"] = ""
```

It would be nice if we
1. Had an alias for the deprecated `bytes` dtype to `variable_length_bytes`
2. Could deal with `fill_value = []` here

Otherwise data that was written with older Zarr versions is not interoperable with new ones.

d-v-b:
we should definitely add an alias for `bytes`. The runtime fix would be to change [this function](https://github.com/zarr-developers/zarr-python/blob/6811c9483a86a0e5c835b4f3c3a2dd09254c38e5/src/zarr/core/dtype/npy/bytes.py#L1034-L1049)

d-v-b:
in fact, we should probably change the JSON signature of the vlen bytes data type to `{"name": "bytes"}` because this does [have a spec](https://github.com/zarr-developers/zarr-extensions/tree/main/data-types/bytes), which lets us get rid of an annoying warning.

PR #3532, opened 2025-10-20
Fix: allow "bytes" as an alias for "variable_length_bytes"

d-v-b:
A simple change that treats `"bytes"` as an alias for `"variable_length_bytes"`. This will address part of the problems revealed in https://github.com/zarr-developers/zarr-python/issues/3513


d-v-b:
@zarr-developers/python-core-devs could someone review this? It would be nice to have this merged today in time for a release.

maxrjones (approved):

PR #3533, opened 2025-10-20
chore: make tests faster

d-v-b:
- Reduces the maximum number of samples in hypothesis testing from 300 to 50
- Simplifies array indexing test cases
- Reduces the size of arrays used in array indexing tests

### DIFF
--- a/.github/workflows/check_changelogs.yml
+++ b/.github/workflows/check_changelogs.yml
@@ -12,7 +12,7 @@ jobs:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
 
       - name: Install uv
-        uses: astral-sh/setup-uv@2ddd2b9cb38ad8efd50337e8ab201519a34c9f24 # v7.1.1
+        uses: astral-sh/setup-uv@85856786d1ce8acfbcc2f13a5f3fbd6b938f9f41 # v7.1.2
 
       - name: Check changelog entries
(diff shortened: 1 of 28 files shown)


d-v-b:
Runtime for a sample test case  (os=ubuntu-latest, py=3.11, np=1.26, deps=minimal)
- This PR: [4m 3s](https://github.com/zarr-developers/zarr-python/actions/runs/18651989445/job/53171973821)
- Main: [4m 56s](https://github.com/zarr-developers/zarr-python/actions/runs/18652285809/job/53172878364)

tests/conftest.py
```diff
@@ -231,7 +231,7 @@ def pytest_collection_modifyitems(config: Any, items: Any) -> None:
 settings.register_profile(
     "default",
     parent=settings.get_profile("default"),
-    max_examples=300,
+    max_examples=50,
```
dcherian: do you want to change local runtime or CI runtime? the CI profile is below.


Given the rectilinear chunk grid work, i wouldn't reduce this. Indeed we might want to increase it to explore all the possibilities!

d-v-b: happy to revert this change in that case, and focus on just the indexing test speedups

dcherian: Sounds good to me!

d-v-b: reverted this change in https://github.com/zarr-developers/zarr-python/pull/3533/commits/bde752d7eb8f65e191d3328ad8a436ecb0aa60f7

Issue #1595, opened 2023-12-07
Implement Variable Chunking (ZEP003) in V3


jhamman:
https://github.com/zarr-developers/zarr-python/pull/1483 provides a proof of concept implementation of Variable Chunking 🙌 ! That implementation is on top of V2 so will obviously need to be ported to V3. This issue flags that some design considerations in the codec pipeline are going to be required and that input on #1583 (and initial PRs like #1588 and #1589) would be useful. To close this issue, we should have a working implementation of the ZEP003 in the `v3` dev branch. 

cc @martindurant, @ivirshup, @agoodm, @alex-s-gardner, @d-v-b, @normanrz

martindurant:
> That implementation is on top of V2 so will obviously need to be ported to V3

It actually works with V3 as given, for the versions of V3 coexisting with V2 in this repo.

maxrjones:
Implemented via the rectilinear chunks extension in https://github.com/zarr-developers/zarr-python/releases/tag/v3.2.0

PR #3534, opened 2025-10-20
Feature: support rectilinear chunk grid extension 

jhamman:
# Summary

Adds support for RectilinearChunkGrid extension (Zarr v3), enabling arrays with variable chunk sizes per dimension.

**Closes:** #1595 | **Replaces:** #1483 | **Related:** zarr-extensions#25

### RectilinearChunkGrid

```python
arr = zarr.create_array(
    shape=(60, 100),
    chunks=[[10, 20, 30], [25, 25, 25, 25]],
    zarr_format=3
)
```

- Zarr v3 only (not compatible with v2, sharding, or `from_array()`)
- Supports RLE in JSON metadata: `[[10, 6]]` = 6 chunks of size 10
- Stored internally in expanded format for fast indexing

### Chunk Grid Access

```python
grid = arr.chunk_grid                    # Returns ChunkGrid instance
grid.chunk_shapes                        # ((10, 20, 30), (25, 25, 25, 25))
isinstance(grid, RectilinearChunkGrid)   # Type-safe checking
```

### `.chunks` Property Behavior

```python
# RegularChunkGrid: returns tuple with FutureWarning (deprecated)
arr.chunks  # (10, 10)

# RectilinearChunkGrid: raises NotImplementedError
arr.chunks  # Use .chunk_grid instead
```

## Design Decisions

| Decision | Rationale |
|----------|-----------|
| `ChunksLike` as TypeAlias | Flexible input types without runtime overhead |
| `ResolvedChunkSpec` as frozen dataclass | Named access, immutability, IDE support |
| Standalone validation functions | Testability, clear error messages, early validation |
| `.chunks` raises for rectilinear | No sensible single-tuple representation; guides users to `.chunk_grid` |

## Removed from Earlier Designs

| Item | Reason |
|------|--------|
| `RegularChunks`/`RectilinearChunks` tuple subclasses | Rejected - unnecessary complexity |
| Named dimension access (`chunks.lat`) | Removed per review feedback |
| `ChunksType` ABC hierarchy | Not implemented - TypeAlias approach preferred |

## Deferred / TODO Items

| Item | Location | Notes |
|------|----------|-------|
| `update_shape()` optional chunks parameter | `metadata/v3.py:483` | Allow specifying new chunk sizes when resizing instead of default heuristic |
| Validation function placement | `chunk_grids.py:1513-1593` | Reviewer suggested moving to metadata module; kept for testability |

## Review Focus Areas

**High Priority:**
- `chunk_grids.py`: `RectilinearChunkGrid` class, `ChunksLike` type, RLE expansion/compression, `resolve_chunk_spec()`
- `metadata/v3.py`: `update_shape()` for rectilinear resize behavior
- `indexing.py`: Variable chunk indexing with binary search
- `array.py`: `.chunks` property behavior, `.chunk_grid` property

**Tests:**
- `test_chunk_grids/test_rectilinear.py`: Comprehensive unit tests
- `test_chunk_grids/test_rectilinear_integration.py`: End-to-end scenarios
- `testing/strategies.py`: Hypothesis strategies for property-based testing

## Breaking Changes

None. Fully backward compatible.

TODO:


docs/user-guide/arrays.md
```diff
@@ -566,6 +566,124 @@ In this example a shard shape of (1000, 1000) and a chunk shape of (100, 100) is
 This means that `10*10` chunks are stored in each shard, and there are `10*10` shards in total.
 Without the `shards` argument, there would be 10,000 chunks stored as individual files.
 
+## Variable Chunking (Zarr v3)
+
+In addition to regular chunking where all chunks have the same size, Zarr v3 supports
+**variable chunking** (also called rectilinear chunking), where chunks can have different
+sizes along each dimension. This is useful when your data has non-uniform structure or
+when you need to align chunks with existing data partitions.
+
```
d-v-b, line 612:
```suggestion
when you need to align chunks with existing data partitions.

The specification for this chunking scheme can be found [here](https://github.com/zarr-developers/zarr-extensions/tree/main/chunk-grids/rectilinear/).
```

This link doesn't resolve yet but it will when the spec is merged.

docs/user-guide/arrays.md
```diff
@@ -566,6 +566,124 @@ In this example a shard shape of (1000, 1000) and a chunk shape of (100, 100) is
...
+# Read data back
+result = z[:]
+print(f"Data matches: {np.all(result == data)}")
+print(f"Slice [10:30, 50:75]: {z[10:30, 50:75].shape}")
+```
+
+### Accessing chunk information
+
+With variable chunking, the standard `.chunks` property is not available since chunks
+have different sizes. Instead, access chunk information through the chunk grid:
```
d-v-b: I think it would be better if `.chunks` just had a different type (tuple of tuples of ints)

src/zarr/core/chunk_grids.py
```diff
@@ -200,6 +417,627 @@ def get_nchunks(self, array_shape: tuple[int, ...]) -> int:
...
+        For RegularChunkGrid, this is computed using ceildiv for each dimension.
+        """
+        return tuple(
+            ceildiv(array_len, chunk_len)
+            for array_len, chunk_len in zip(array_shape, self.chunk_shape, strict=False)
+        )
+
+
+@dataclass(frozen=True)
+class RectilinearChunkGrid(ChunkGrid):
```
d-v-b: thoughts on just calling this class `Rectilinear`, and renaming the `RegularChunkGrid` to `Regular`? We could keep around a `RegularChunkGrid` class for compatibility. But I feel like people know these are chunk grids when they import them

src/zarr/core/chunk_grids.py
```diff
@@ -200,6 +417,627 @@ def get_nchunks(self, array_shape: tuple[int, ...]) -> int:
...
+
+        return {
+            "name": "rectilinear",
+            "configuration": {
+                "kind": "inline",
+                "chunk_shapes": chunk_shapes_list,
+            },
+        }
+
+    def all_chunk_coords(self, array_shape: tuple[int, ...]) -> Iterator[tuple[int, ...]]:
```
d-v-b: we should cache this

d-v-b: nevermind, what we need to cache is the cumulative sum of the lengths along each axis, and we only get that when we know the shape

src/zarr/core/chunk_grids.py
```diff
@@ -200,6 +417,627 @@ def get_nchunks(self, array_shape: tuple[int, ...]) -> int:
...
+        ):
+            chunk_sum = sum(axis_chunks)
+            if chunk_sum != arr_size:
+                raise ValueError(
+                    f"Sum of chunk sizes along axis {axis} is {chunk_sum} "
+                    f"but array shape is {arr_size}"
+                )
+
+    @cached_property
+    def _cumulative_sizes(self) -> tuple[tuple[int, ...], ...]:
```
d-v-b: nice, this is cached

src/zarr/core/chunk_grids.py
d-v-b: i now see https://github.com/zarr-developers/zarr-python/pull/3534/files#r2446176968

src/zarr/core/indexing.py
```diff
@@ -379,35 +370,70 @@ class ChunkDimProjection(NamedTuple):
 class IntDimIndexer:
```
d-v-b, line 370: can we make this a dataclass and get rid of init?

d-v-b: nevermind it already is one :)

src/zarr/testing/strategies.py
```diff
@@ -443,3 +562,78 @@ def chunk_paths(draw: st.DrawFn, ndim: int, numblocks: tuple[int, ...], subset:
...
+    else:
+        event("using RectilinearChunkGrid (run length encoded)")
+        repeats = st.lists(
+            st.integers(min_value=1, max_value=20), min_size=nchunks, max_size=nchunks
+        )
+        chunk_shapes_rle = [
+            [[c, r] for c, r in zip(draw(dim_chunks), draw(repeats), strict=True)]
+            for _ in range(ndim)
+        ]
+        return RectilinearChunkGrid(chunk_shapes=chunk_shapes_rle)
```
dcherian: This needs fixing

tests/test_properties.py
```diff
@@ -106,11 +108,10 @@ def test_array_creates_implicit_groups(array):
 
 
 @pytest.mark.asyncio
-@settings(deadline=None)
+@settings(deadline=None, report_multiple_bugs=False)
 @pytest.mark.filterwarnings("ignore::zarr.core.dtype.common.UnstableSpecificationWarning")
-@given(data=st.data())
-async def test_basic_indexing(data: st.DataObject) -> None:
-    zarray = data.draw(simple_arrays())
+@given(data=st.data(), zarray=st.one_of([simple_arrays(), complex_chunked_arrays()]))
```
dcherian: Because the search space for the standard `arrays` strategy is so large, i made a different one `complex_chunked_arrays` that purely checks different chunk grids
with `simple_arrays()` we are only spending 10% of our time trying `RectilinearChunkGrid` so using this approach. We should boost number of examples too.

docs/user-guide/arrays.md
```diff
@@ -566,6 +566,124 @@ In this example a shard shape of (1000, 1000) and a chunk shape of (100, 100) is
...
+
+### Limitations
+
+Variable chunking has some important limitations:
+
+1. **Zarr v3 only**: This feature is only available when using `zarr_format=3`.
+   Attempting to use variable chunks with `zarr_format=2` will raise an error.
+
+2. **Not compatible with sharding**: You cannot use variable chunking together with
+   the sharding feature. Arrays must use either variable chunking or sharding, but not both.
```
shoyer, line 707: I hope this is a temporary limitation! There's a natural extension of rectilinear chunk grids to rectilinear shard grids.

jhamman: xref: https://github.com/zarr-developers/zarr-extensions/issues/34

src/zarr/core/chunk_grids.py
jhamman: 50/50. I think the more descriptive class is useful when looking at a tracebacks. Plus, this is currently in `.core` so its not meant to be used directly by users.

jhamman:
@d-v-b, @maxrjones, et al - this is ready for a real review. I'd be happy if we merged this in using an "experimental" warning of some kind and let it bake from 3.2->3.3. Suggestions on the best way to communicate the experimental nature of this are welcome.

keewis:
is it intentional that the chunk grid is `RectilinearChunkGrid` when explicitly enumerating regular chunk sizes?
```python
import zarr

arr = zarr.create_array(
    store,
    shape=(60, 100),
    chunks=[[20, 20, 20], [25, 25, 25, 25]],
    zarr_format=3,
    dtype="uint8"
)
arr.metadata.chunk_grid # RectilinearChunkGrid
```
Or is `RegularChunkGrid` only chosen if I specify a tuple of values, in this case `(20, 25)`?

Not sure how much this matters in practice (`xarray`'s code does not appear to be affected by the difference), but that's what `dask` normalizes the chunks to.

jhamman:
> is it intentional that the chunk grid is RectilinearChunkGrid when explicitly enumerating regular chunk sizes?

@keewis - yes, this is intentional. We decided this was preferable because it will allow the user to add variable lengths later (through an extend/append workflow).

src/zarr/core/chunk_grids.py
```diff
@@ -31,6 +33,489 @@
...
+    All chunk types inherit from both tuple (for backward compatibility)
+    and this ABC (for common interface and isinstance checks).
+
+    Subclasses must implement dimension_names property and provide
+    appropriate __new__ methods.
+    """
+
+    @property
+    @abstractmethod
+    def dimension_names(self) -> tuple[str | None, ...] | None:
```
d-v-b: where is this used? I feel like integer indexing should be fine for the chunk grid.

src/zarr/core/chunk_grids.py
```diff
@@ -31,6 +33,489 @@
...
+
+
+class RectilinearChunkGridConfigurationDict(TypedDict):
+    """TypedDict for rectilinear chunk grid configuration"""
+
+    kind: Literal["inline"]
+    chunk_shapes: Sequence[Sequence[ChunkEdgeLength]]
+
+
+class ChunksType(tuple[Any, ...], ABC):  # noqa: SLOT001
```
d-v-b: I'm not sold on subclassing tuple here. It's going to lead to a lot of surprises unless we carefully implement a lot of stuff that we could just use tuple for:

```python
>>> RegularChunks((10,10,1))[:2]
(10, 10) # tuple, not RegularChunks
>>> RegularChunks((10,10,1)) + (1,)
(10, 10, 1, 1) # tuple, not RegularChunks
```

I would much rather make a breaking change than introduce something super confusing

src/zarr/core/chunk_grids.py
```diff
@@ -31,6 +33,489 @@
...
+
+    @abstractmethod
+    def values(self) -> Iterator[Any]:
+        """Return chunk values, enabling dict(chunks) conversion."""
+        ...
+
+    @abstractmethod
+    def items(self) -> Iterator[tuple[str, Any]]:
+        """Return (name, value) pairs, enabling dict(chunks) conversion."""
+        ...
```
d-v-b: Why do we need key: value access?

src/zarr/core/chunk_grids.py
```diff
@@ -301,3 +1569,531 @@ def _auto_partition(
...
+        chunk_shapes = _normalize_rectilinear_chunks(chunks, shape_parsed)  # type: ignore[arg-type]
+        return RectilinearChunkGrid(chunk_shapes=chunk_shapes)
+    else:
+        # Regular tuple of ints -> RegularChunkGrid
+        chunk_shape = _normalize_chunks(chunks, shape_parsed, item_size)
+        return RegularChunkGrid(chunk_shape=chunk_shape)
+
+
+@dataclass(frozen=True)
+class ResolvedChunkSpec:
```
d-v-b: does this need to be a class? It has no methods. Seems like either a TypedDict or just a tuple would work

src/zarr/core/chunk_grids.py
```diff
@@ -301,3 +1569,531 @@ def _auto_partition(
...
+    shards : tuple[int, ...] | None
+        The shards parameter to pass to init_array/from_array.
+        None if sharding is not used.
+    """
+
+    chunk_grid: ChunkGrid
+    shards: tuple[int, ...] | None
+
+
+def _validate_zarr_format_compatibility(
```
d-v-b: why do we need this here? Can't we check when we construct the array metadata document?

src/zarr/core/chunk_grids.py
```diff
@@ -301,3 +1569,531 @@ def _auto_partition(
...
+
+        # Zarr v2 doesn't support sharding
+        if shards is not None:
+            raise ValueError(
+                f"Sharding is only supported in Zarr format 3. "
+                f"Got zarr_format={zarr_format} with shards={shards}."
+            )
+
+
+def _validate_sharding_compatibility(
```
d-v-b: shouldn't this be defined over in the array v3 metadata module, and used there to check that the array metadata document is valid?

src/zarr/core/chunk_grids.py
```diff
@@ -301,3 +1569,531 @@ def _auto_partition(
...
+
+        # Variable chunks (nested sequences) can't be used with sharding
+        if _is_nested_sequence(chunks):
+            raise ValueError(
+                "Cannot use variable chunks (nested sequences) with sharding. "
+                "Sharding requires uniform chunk sizes."
+            )
+
+
+def _validate_data_compatibility(
```
d-v-b: does this need to be stand-alone function? are we going to use this logic anywhere other than it's current usage site (resolve_chunk_spec)

src/zarr/core/metadata/v3.py
```diff
@@ -286,20 +293,43 @@ def dtype(self) -> ZDType[TBaseDType, TBaseScalar]:
         return self.data_type
 
     @property
-    def chunks(self) -> tuple[int, ...]:
+    def chunks(self) -> ChunksType:
```
d-v-b: I'd be willing to drop this attribute entirely if it's too complicated to support in a clean way.

src/zarr/core/metadata/v3.py
```diff
@@ -436,6 +468,21 @@ def to_dict(self) -> dict[str, JSON]:
         return out_dict
 
     def update_shape(self, shape: tuple[int, ...]) -> Self:
```
d-v-b, line 470: IMO this needs to take a parameter that can define the new chunks (which can default to `None` or some other sentinel flagging "default behavior" semantics)

src/zarr/core/array.py
```diff
@@ -2350,17 +2396,20 @@ def shape(self, value: tuple[int, ...]) -> None:
...
+        the uniform chunk shape. If sharding is used, the inner chunk shape is returned.
+
+        For arrays using RectilinearChunkGrid: returns a tuple of tuples, where
+        each inner tuple contains the chunk sizes along that dimension (not RLE encoded).
 
         Returns
         -------
-        tuple
-            A tuple of integers representing the length of each dimension of a chunk.
+        tuple[int, ...] | tuple[tuple[int, ...], ...]
```
dcherian: I don't like this at all. 

We need to be resilient to O(10 million) chunk arrays and converting say RLE encoded chunk lengths to full `tuple[tuple[int], ...]` would suck a lot. The user expectation today is that `.chunks` returns `tuple[int]` with O(30 elements).

Here is an example where such "normalizing" behaviour really kills things: https://github.com/pydata/xarray/issues/8902#issuecomment-2546127373

I'd strongly recommend we raise an error saying `chunks` is ill-defined for non RegularChunks arrays, adn recommend adding a `.chunk_grid -> ChunkGrid` property and adding an `ChunkGrid.as_normalized_tuple() -> tuple[tuple[int], ...]` for the fully normalized representation.

d-v-b: we should list some operations we would like this object to support. off the top of my head, we need
- iteration (lexicographic order)
- contiguous, n-dimensional slicing on chunk grid coordinates (for when we slice the array)
- concatenation? (to invert slicing)

If we want the latter operations to be well-behaved, then it looks like this chunk grid _is an array_...

d-v-b: specifically, an ND array of N-tuples of ints. If we want slicing to be reversible, then we also need to track the bounding box of this array

tomwhite:
FYI I tested this PR for [implementing rechunking with variable-sized intermediate chunks](https://github.com/cubed-dev/cubed/issues/876) in Cubed - and it worked!

I found one wrinkle in that `zarr.create_array` supports rectilinear chunk grids, but `zarr.open` (with `mode="w"`) doesn't. But that could be addressed later.

src/zarr/core/chunk_grids.py
jhamman: removed

src/zarr/core/chunk_grids.py
jhamman: removed.

src/zarr/core/array.py
jhamman: Noted. Two changes in my most recent commits: 

- `.chunks` now errors for irregular chunks, point to `.chunk_grid`
- `.chunk_grid` now exists, providing the user a path to the ChunkGrid class. 

I plan to hold other convenience methods for chunk tuple discovery, iteration, slicing, etc. for future PRs.

d-v-b:
I will give this a spin today, but assuming everything works, my question is: how can we ship this soon (this week?) while making it clear that the feature is experimental?

tinaok:
I have varient chunked array;
```
import xarray as xr
from healpix_geo import nested 
import numpy as np

da = xr.open_zarr(
    "https://data-taos.ifremer.fr/GRID4EARTH/no_chunk_healpix.zarr",
    consolidated=True,   # if metadata is consolidated
).da

depth = da.cell_ids.attrs['level']
new_depth = depth-6
parents = nested.zoom_to(da.cell_ids, depth=depth, new_depth=new_depth) 
_, chunk_sizes =np.unique(parents, return_counts=True)
da.chunk({"cell_ids": tuple(chunk_sizes.tolist())})
```
Happy to experiment to write it using this experimental feature(if it works with xarray...)

maxrjones:
@jhamman I'm finally reviewing this. I'd like to get you an approval today. The one issue that I think we should address first is a performance regression in indexing, where `IntArrayDimIndexer` and
`CoordinateIndexer` replaced a vectorized numpy operation (`dim_sel // dim_chunk_len`) with a per-element for-loop calling `array_index_to_chunk_coord()`. This affects both `RegularChunkGrid` along with
`RectilinearChunkGrid`, which is why I think this is a blocking regression. I opened a PR against your branch with a potential fix — https://github.com/jhamman/zarr-python/pull/5. This is a plot from https://github.com/maxrjones/zarr-chunk-grid-tests showing the performance impact:
image

src/zarr/core/chunk_grids.py
```diff
@@ -204,6 +495,672 @@ def get_nchunks(self, array_shape: tuple[int, ...]) -> int:
...
+            zip(array_shape, self.chunk_shapes, strict=False)
+        ):
+            chunk_sum = sum(axis_chunks)
+            if chunk_sum < arr_size:
+                raise ValueError(
+                    f"Sum of chunk sizes along axis {axis} is {chunk_sum} "
+                    f"but array shape is {arr_size}. This is invalid for the "
+                    "RectilinearChunkGrid."
+                )
+
```
maxrjones:
```suggestion
            if sum(axis_chunks[:-1]) >= arr_size:
                raise ValueError(
                    f"Axis {axis} has more chunks than needed "
                    f"The last chunk(s) would contain no valid data."
                    f"Remove the extra chunk(s) or increase the array shape."
                )

```

src/zarr/core/chunk_grids.py
```diff
@@ -301,3 +1258,531 @@ def _auto_partition(
...
+    # Validate that chunks sum to shape for each dimension
+    for i, (dim_chunks, dim_size) in enumerate(zip(chunk_shapes, shape, strict=False)):
+        chunk_sum = sum(dim_chunks)
+        if chunk_sum < dim_size:
+            raise ValueError(
+                f"Variable chunks along dimension {i} sum to {chunk_sum} "
+                f"but array shape is {dim_size}. Chunks must sum to be greater than or equal to the shape."
+            )
+
```
maxrjones:
```suggestion
        if sum(dim_chunks[:-1]) >= dim_size:
            raise ValueError(
                f"Dimension {i} has more chunks than needed "
                f"The last chunk(s) would contain no valid data."
                f"Remove the extra chunk(s) or increase the array shape."
            )

```

tests/test_chunk_grids/test_rectilinear_integration.py
```diff
@@ -0,0 +1,160 @@
...
+        chunks=[[10, 20, 30], [25, 25, 25, 25]],
+        dtype="u1",
+        zarr_format=3,
+    )
+
+    # The nchunks property now works for RectilinearChunkGrid
+    assert arr.nchunks == 12  # 3 chunks in dim 0, 4 chunks in dim 1
+
+    # Can also get nchunks from the chunk_grid directly
+    assert arr.metadata.chunk_grid.get_nchunks((60, 100)) == 12
```
maxrjones, line 160:
```suggestion
    assert arr.metadata.chunk_grid.get_nchunks((60, 100)) == 12


def test_rectilinear_validate_chunk_sum_exceeds_array_size() -> None:
    """Test that chunk sizes summing to more than array size are rejected.

    chunks=[5, 5, 5] (sum=15) for shape=(10,) should fail because the chunks
    describe more data than the array contains.
    """
    grid = RectilinearChunkGrid(chunk_shapes=[[5, 5, 5]])
    with pytest.raises(ValueError, match="more chunks"):
        grid.get_nchunks((10,))
```

tests/test_chunk_grids/test_resolve_chunk_spec.py
```diff
@@ -0,0 +1,394 @@
...
+        resolve_chunk_spec(
+            chunks=[[10, 20, 30]],  # 1D
+            shards=None,
+            shape=(60, 100),  # 2D
+            dtype_itemsize=4,
+            zarr_format=3,
+        )
+
+
+# Edge case tests
```
maxrjones, line 304:
```suggestion
def test_resolve_chunk_spec_error_chunk_sum_exceeds_array_size() -> None:
    """Test that variable chunks summing to more than array size raise error.

    shape=10 with chunks=[5, 5, 5] (sum=15) should be rejected because it
    specifies more data than the array can hold, which could cause confusing
    runtime errors or data corruption.
    """
    with pytest.raises(ValueError, match="more chunks"):
        resolve_chunk_spec(
            chunks=[[5, 5, 5]],  # sum=15 > shape=10
            shards=None,
            shape=(10,),
            dtype_itemsize=4,
            zarr_format=3,
        )

# Edge case tests
```

maxrjones (commented):
These changes add validation to reject rectilinear chunk specifications where the last chunk would contain no valid data.

For example, previously `chunks=[[5, 5, 5]]` with `shape=(10,)` would silently create a extra 3rd chunk at `slice(10, 15)` entirely outside the array.

This matters because without it, `get_nchunks()` and `all_chunk_coords()` report the extra unused chunks as real, which could cause issue downstream. Also, the code would accept (most likely) user errors rather than catching the errors at construction time.

maxrjones:
I tried this out in virtual tiff (https://github.com/virtual-zarr/virtual-tiff/pull/69) and virtualizarr (https://github.com/zarr-developers/VirtualiZarr/pull/877) and am super excited about this feature. I expect that it'll unlock a lot of downstream development and absolutely support releasing it this week as experimental. I think that the experimental status is already effectively documented.

My only additional comment beyond my two reviews earlier today is that I'm not convinced of the value of deprecating `.chunks` for `RegularChunkGrid`, but wouldn't want that concern to block merging.

Thanks for your work on this @jhamman!

src/zarr/core/chunk_grids/regular.py
```diff
@@ -0,0 +1,128 @@
...
+    parse_named_configuration,
+    parse_shapelike,
+)
+
+if TYPE_CHECKING:
+    from collections.abc import Iterator
+
+
+@dataclass(frozen=True)
+class RegularChunkGrid(ChunkGrid):
```
d-v-b, line 27: I don't think we actually need a regular chunk grid class, since the rectilinear chunk grid is more general.

src/zarr/core/chunk_grids/rectilinear.py
```diff
@@ -0,0 +1,882 @@
...
+                f"Dimension {i} has more chunks than needed. "
+                f"The last chunk(s) would contain no valid data. "
+                f"Remove the extra chunk(s) or increase the array shape."
+            )
+
+    return chunk_shapes
+
+
+@dataclass(frozen=True)
+class RectilinearChunkGrid(ChunkGrid):
```
d-v-b, line 258: many of the methods on this class take the array shape as a parameter. For an array with a constant shape, this is redundant, so we should probably define the array shape as an attribute of this class.

src/zarr/core/chunk_grids/rectilinear.py
```diff
@@ -0,0 +1,882 @@
...
+
+        return {
+            "name": "rectilinear",
+            "configuration": {
+                "kind": "inline",
+                "chunk_shapes": chunk_shapes_compressed,
+            },
+        }
+
+    def update_shape(self, new_shape: tuple[int, ...]) -> Self:
```
d-v-b, line 369: the name of this method is confusing -- it makes me think the chunk grid has a `shape` attribute. And I feel like we should allow people to define the chunking of the new part of the array. so maybe `resize` would be a better name in that case

src/zarr/api/synchronous.py
```diff
@@ -822,7 +824,7 @@ def create_array(
     shape: ShapeLike | None = None,
     dtype: ZDTypeLike | None = None,
     data: np.ndarray[Any, np.dtype[Any]] | None = None,
-    chunks: tuple[int, ...] | Literal["auto"] = "auto",
+    chunks: ChunksLike = "auto",
```
d-v-b, line 827: we declare rectilinear chunks and regular chunks both via the `chunks` attribute, but accessing the `chunks` attribute of the array raises an exception in the rectilinear case. This is a deviation from the way the rest of the parameters in this function work -- they are all attributes on the array class, and none of them raise exceptions when you access them. The closest thing is shards being `None`, but that's consistent with the signature of this function.

d-v-b: also -- we need a way for someone to say "these chunks are regular, and I want to use the regular chunk grid" vs "these chunks are regular, and I want to use the rectilinear chunk grid". I don't think the signature right now supports this.

src/zarr/core/metadata/v3.py
```diff
@@ -297,9 +320,16 @@ def chunks(self) -> tuple[int, ...]:
             else:
                 return self.chunk_grid.chunk_shape
 
+        if isinstance(self.chunk_grid, RectilinearChunkGrid):
+            msg = (
+                "The `chunks` property is not supported for arrays with variable chunk sizes "
+                "(RectilinearChunkGrid). Use `chunk_grid` instead to access chunk information."
+            )
+            raise NotImplementedError(msg)
```
d-v-b, line 328: I'd like "accessing some attributes raises exceptions" to be a last resort, even if it means breaking public API. What if we copied dask, and made `chunks` a tuple of tuples of edge lengths, one per axis?

src/zarr/core/metadata/v2.py
```diff
@@ -61,7 +61,7 @@ class ArrayV2MetadataDict(TypedDict):
 @dataclass(frozen=True, kw_only=True)
 class ArrayV2Metadata(Metadata):
     shape: tuple[int, ...]
-    chunks: tuple[int, ...]
```
d-v-b, line 64: why are we removing this? that's very confusing, from a "model the v2 metadata document" pov, where `chunks` is an attribute

src/zarr/core/metadata/v3.py
dcherian: Everyone using Zarr today expects `chunks` to be a `tuple[int]`. I think it's **significantly** worse to silently break that expectation

src/zarr/core/chunk_grids/regular.py
dcherian: I disagree. Downstream libraries may want to use this info to possibly use a fastpath.

d-v-b: How would that work? Because a rectilinear chunk grid can be regular. So if a library uses `RegularChunkGrid` to trigger a fastpath, they _also_ need to inspect the could-be-regular chunks of the `RectilinearChunkGrid` as well, and then we are back to the `RegularChunkGrid` having no use.

d-v-b: and the information about the chunk grid an array is using is already contained in the metadata for that array. clients should feel more comfortable inspecting that document than importing internal utility classes from zarr python.

src/zarr/core/metadata/v3.py
d-v-b: yes, it would be a breaking change, and so of course we can't do it silently, but it might also be a _welcome_ change to any downstream library that uses dask with zarr, so IMO we should consider this as a possibility for how we want the zarr array API to eventually look

src/zarr/core/chunk_grids/regular.py
dcherian: > they also need to inspect 

They don't _need_ to. This stuff can be expensive. My input here comes from many years of trying to write optimized dask code. Having to inspect the damned chunks tuple with potentially 10 million elements in Python is quite slow and annoying. We should strive to preserve such context through strong types.

Here is an example: https://github.com/pydata/xarray/pull/9808/changes . This core problem there was that we normalized zarr's nice `tuple[int, ...]` to dask-style `tuple[tuple[int, ...]]` and then later added a "nice performance warning" that slowed everything down because of the need to parse through half a million _identical_ ints. 🤦🏾‍♂️

d-v-b: > They don't need to. This stuff can be expensive.

Inspecting the `chunk_grid` attribute of the metadata document _will never be expensive_. We are not going to put 10 million elements in JSON.

d-v-b: for context, my suggestion is the removal of the `RegularChunkGrid` _class_, not the removal of the _information_ that the chunk grid is regular. Those are different things.

maxrjones:
I've been building on the work in this PR and the design discussion here to explore an alternative internal architecture for rectilinear chunk grids. The result is a polished proof of concept branch with a design doc, comprehensive tests, and working downstream POCs for xarray, cubed, VirtualiZarr, and virtual-tiff.

**Branch:** https://github.com/maxrjones/zarr-python/pull/5
**Design doc:** https://github.com/maxrjones/zarr-python/blob/poc/unified-chunk-grid/docs/design/chunk-grid.md

The approach reuses several components from this PR (RLE helpers, validation logic, test patterns) and tries to address the design feedback raised here. 

While the existing PR would deliver the long-requested rectilinear chunks feature, I would appreciate at least consideration of the alternative design because I think it has some clean abstractions that could enable future extensions. For example, it would be quite simple to add support for repeating chunk grids.

The key architectural differences:

- **Per-dimension decomposition** (`FixedDimension` / `VaryingDimension` behind a `DimensionGrid` protocol) instead of monolithic grid classes. Callers use polymorphic dispatch rather than `isinstance` branching.
- **Self-describing grids.** Extent stored per-dimension at construction, so `grid[coord]` works without passing `array_shape` everywhere.
- **Single `ChunkGrid` class with `is_regular` property** instead of `RegularChunkGrid` + `RectilinearChunkGrid`, with a `RegularChunkGrid` deprecation shim for backwards compatibility.
- **Explicit opt-in via feature flag**. Rectilinear chunks are gated behind `zarr.config.set({"array.rectilinear_chunks": True})`, disabled by default. This lets us ship the feature experimentally without committing to API stability, and gives a clear path to promotion once downstream usage validates the design.

The design doc covers the rationale for these choices, addresses all the review feedback from this PR, and documents what was intentionally deferred (metadata/array separation, resize chunks parameter, tile encoding).

I'd suggest reviewing the design doc first to see if the approach makes sense. If it does, I can split the implementation into 5 incremental PRs detailed at the bottom of the design doc.

Thanks, @jhamman, for the foundational work in this PR! I really appreciate everything you've done here and found it essential for making architectural trade-offs concrete.

I hope y'all will consider the alternative architecture in my branch/design doc.

PR #3545, opened 2025-10-24
fix invalid blosc defaults

### LINKED ISSUE #3427: `BloscCodec` is broken
We need to fix the following behavior of the `BloscCodec`:

```python
# /// script
# requires-python = ">=3.11"
# dependencies = [
# ]
# ///
#

from zarr.codecs import BloscCodec
b = BloscCodec()
b.to_dict()
# ValueError: `typesize` needs to be set for serialization.
```

This means picking usable defaults instead of creating a completely broken object.

Unfortunately I am not sure how to gracefully fix this, since we have load-bearing internal code that relies on certain `BloscCodec` fields being set to `None`.

d-v-b:
Ensures that the blosc codec cannot be created with invalid parameters. fixes #3427.

In `main`, the default `typesize` and `shuffle` values in `BloscCodec.__init__` are `None`.  As `None` is not a valid value for these parameters, `BloscCodec().to_dict()` fails with an exception:

```python
>>> from zarr.codecs import BloscCodec                                           
>>> BloscCodec().to_dict()
  File "/Users/d-v-b/dev/zarr-python/src/zarr/codecs/blosc.py", line 127, in to_dict
    raise ValueError("`typesize` needs to be set for serialization.")
ValueError: `typesize` needs to be set for serialization.
```

That's problematic, so in this PR the default values for `BloscCodec.__init__` are valid. You can still pass `typesize=None`, but this emits a deprecation warning, and the `None` value is replaced by the real default.

But I needed to preserve the behavior that motivated making `typesize` and `shuffle` nullable in the first place, which is the need to tune `typesize` and `shuffle` based on the data type of the array in a routine called `evolve_from_array_spec`. We might ultimately want to avoid this tuning[^1], but we should deal with that in a different PR. Preserving the old behavior required adding a new attribute `tunable_attrs` to the `BloscCodec` which tracks the names of attributes that can be tuned by `evolve_from_array_spec`. Setting `tunable_attrs` to the empty set `{}` allows someone to use default the `typesize` and `shuffle` parameters without having them tuned by `evolve_from_array_spec`. Happy to explain more but the tl;dr is that we needed this new `tunable_attrs`  parameter to keep the old behavior.

[^1]: BloscCodec is a `bytes -> bytes` codec, so there's no guarantee that input is partitioned into array-data-type-sized words. The byte stream it consumes is based on the behavior of the codecs that come before it, which might do lots of different things to the byte stream.

### DIFF
--- a/changes/3545.misc.md
+++ b/changes/3545.misc.md
@@ -1,0 +1,1 @@
+Changes the internal logic of the `BloscCodec` class to ensure that the `typesize` end `shuffle` parameters are not nullable.


normanrz (approved):

ilan-gold:
Did I do something incorrect here to get this error?

```python
# /// script
# requires-python = ">=3.11"
# dependencies = [
# ]
# ///

import numpy as np
import zarr
from zarr.codecs import BloscCodec

compressor = BloscCodec(cname="zstd", clevel=3, shuffle="bitshuffle")
data = np.arange(10)
zarr.create_array("foo.zarr", data=data, compressors=(compressor, ), overwrite=True)
f = zarr.open("foo.zarr")
assert f.compressors[0].to_dict() == compressor.to_dict(), (
    f"read back: {f.compressors[0].to_dict()} and actual {compressor.to_dict()}"
)
```

Here is the assertion (with the two configs):

```pytb
  File "/Users/ilangold/Projects/Theis/anndata/shards_type.py", line 19, in <module>
    assert f.compressors[0].to_dict() == compressor.to_dict(), (
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: read back: {'name': 'blosc', 'configuration': {'typesize': 8, 'cname': 'zstd', 'clevel': 3, 'shuffle': 'shuffle', 'blocksize': 0}} and actual {'name': 'blosc', 'configuration': {'typesize': 1, 'cname': 'zstd', 'clevel': 3, 'shuffle': 'bitshuffle', 'blocksize': 0}}
```

Apologies if this is just noise, maybe I missed something here

d-v-b:
I think this is expected -- the [evolve_from_array_spec](https://github.com/zarr-developers/zarr-python/blob/fe42655ae265e045f850e12f30726aa8668d6dde/src/zarr/codecs/blosc.py#L331-L350) method on the `BloscCodec` is used internally at array metadata creation time to create a copy of the original `BloscCodec` with the `typesize` and `shuffle` parameters tuned to the array's data type. So in your case, the information that the data has 64 bit data type leads to a new `BloscCodec` with the `typesize` set to 8. I'm not sure this behavior is right, but it's what we have been doing with since 3.0

ilan-gold:
Good point - but what of `shuffle` (which is `shuffle` in the read-back-in codec but `bitshuffle` in the setting one)?

d-v-b:
> Good point - but what of `shuffle` (which is `shuffle` in the read-back-in codec but `bitshuffle` in the setting one)?

that's also consistent with the [old implementation](https://github.com/zarr-developers/zarr-python/blob/fc8e8ad1a143b1f39a2e08b470857a5e48a38bca/src/zarr/codecs/blosc.py#L141-L154):

```python
    def evolve_from_array_spec(self, array_spec: ArraySpec) -> Self:
        item_size = 1
        if isinstance(array_spec.dtype, HasItemSize):
            item_size = array_spec.dtype.item_size
        new_codec = self
        if new_codec.typesize is None:
            new_codec = replace(new_codec, typesize=item_size)
        if new_codec.shuffle is None:
            new_codec = replace(
                new_codec,
                shuffle=(BloscShuffle.bitshuffle if item_size == 1 else BloscShuffle.shuffle),
            )

        return new_codec
```

ilan-gold:
Is there any way to set bitshuffle with an item size > 1? From reading the docs, I would think setting `tunable_attrs={}` and `shuffle="bitshuffle"`, your setting should be preserved:

```python
# /// script
# requires-python = ">=3.11"
# dependencies = [
# ]
# ///

import numpy as np
import zarr
from zarr.codecs import BloscCodec

compressor = BloscCodec(tunable_attrs={}, cname="zstd", clevel=3, shuffle="bitshuffle")
data = np.arange(10)
zarr.create_array("foo.zarr", data=data, compressors=(compressor,), overwrite=True)
f = zarr.open("foo.zarr")
assert f.compressors[0].to_dict() == compressor.to_dict(), (
    f"read back: {f.compressors[0].to_dict()} and actual {compressor.to_dict()}"
)
```

I also think it's a little strange to have the default as `bitshuffle` but then have it be automatically changed the shuffle, but maybe I'm misinterpreting.  I think I'll just [change our test](https://github.com/scverse/anndata/blob/efb9efddc51e9a9ffe72aff9e265dd096e3db381/tests/test_readwrite.py#L394-L397) that looks for "codecs on-disk written match what is requested" to not use blosc since it's a bit hairy.  Definitely support getting rid of invalid defaults!

d-v-b:
if `tunable_attrs` is the empty set, then `evolve_from_array_spec` and thereby array creation should not change any of the user-provied kwargs. so that looks like a bug. investigating now

d-v-b:
@ilan-gold see https://github.com/zarr-developers/zarr-python/pull/3553 for a better version of this PR

ilan-gold:
Awesome @d-v-b, looks like I can now set `typesize` and `shuffle` at will.  Thanks!

PR #3547, opened 2025-10-24
feat!: store-defined concurrency limits

d-v-b:

Removes Zarr's global async.concurrency limit and concurrent_map helper, using asyncio.gather for concurrent work. Concurrency control becomes the responsibility of backends and any store adapters that need it.

The current implementation does not contain the previously proposed ConcurrencyLimiter mixin or per-store limit of 50. It also does not warn when setting the removed async.concurrency key. Existing custom stores that relied on Zarr's limit can now receive more concurrent work and should be assessed accordingly.

Some recorded benchmarks improved, but performance is workload-dependent. The current CodSpeed comment reports both improvements and regressions and uses an older successful main baseline, so it does not establish a universal performance improvement or that this change is safe for every backend.


dcherian:
Nice, the other one that bugs me when profiling is https://github.com/zarr-developers/zarr-python/blob/b3e9aed305092236c5db70deee0b26dad648d3b0/src/zarr/codecs/zstd.py#L79

maxrjones:
@d-v-b do you have thoughts on how downstream libraries would interact with Zarr's concurrency limit? Would those set there own concurrency approaches (e.g., global or per function) or could there be a way to share the limit with Zarr?

d-v-b:
@maxrjones this PR is moving away from a global concurrency limit because it's actually very hard to implement a top-down concurrency control in our codebase. the new direction I'm taking with this PR is for each store to set its own concurrency limit. for stateless stores like local and remote (basically anything except zip), concurrency-sensitive applications should be able to cheaply create store instances with concurrency limits tuned to the needs of the application.

d-v-b:

Audit clarification: This describes an earlier revision. At bd836f381 the ConcurrencyLimiter mixin and per-store defaults are absent; backends/adapters must supply any limit.

Original comment (historical):

this is ready for a look. The basic idea is that we make concurrency limits an implementation detail of a particular store and thereby remove all concurrency limiting logic from array / group routines. Some stores will have no concurrency limits (e.g., memory and local storage) and so their performance will go up. Stores that actually benefit from a concurrency limit (remote stuff) has the default limit of 10 but that's something we can play with.

To cut down on boilerplate we (claude and I) define a `ConcurrencyLimiter` base class that encapsulates the concurrency limiting logic. This means taking a concurrency limit (int | none) and creating a semaphore (or a nullcontext if there's no limit). We also define a decorator that an be used to wrap a should-be-concurrency-limited method on a class that inherits from the `ConcurrencyLimiter` base class.

maxrjones (commented):
Thanks @d-v-b! I like the per-store concurrency limits idea. My main concern is that external stores implemented under the global concurrency limit model could go from limited concurrency to unbounded concurrency. It seems like there's at least a few cases of external store implementations on GitHub - https://github.com/search?q=zarr.abc.store&type=code. One option for mitigation is to make the Store ABC provide a default concurrency limit via the ConcurrencyLimiter, which store implementations like MemoryStore could opt out of rather than opt in to.

d-v-b:
> One option for mitigation is to make the Store ABC provide a default concurrency limit via the ConcurrencyLimiter, which store implementations like MemoryStore could opt out of rather than opt in to.

How would this look? Our store ABC doesn't define concrete implementations for stuff like `get` or `set` -- every store defines these anew. So I don't know how we could put a default concurrency limit in at the ABC level.

d-v-b:
also I think a large number of downstream packages import `Store` for type-checking, not because they implement their own stores. That being said, we would definitely want folks who _have_ implemented their own stores (e.g., @cgohlke, @paraseba) to have a look at these changes and vet them.

d-v-b:

Audit clarification: These were improvements in particular benchmarks, not a general guarantee. The current CodSpeed report has improvements and regressions and uses an older successful main baseline.

Original comment (historical):

looks like removing the unnecessary concurrency limits for local and memory storage gets us a very nice performance improvement, according to the automated benchmarks. I think that escalates the priority of this PR.

d-v-b:
> My main concern is that external stores implemented under the global concurrency limit model could go from limited concurrency to unbounded concurrency.

the best thing I can think of is a minor release that adds an annoying warning to `Store.__init_subclass__` like "In the next minor release of Zarr Python, concurrency limiting functionality must be defined on store classes. See `<docs page>` for information about migrating your Store subclass." I'm open to other ideas though!

src/zarr/storage/_obstore.py
```diff
@@ -40,7 +38,7 @@
 T_Store = TypeVar("T_Store", bound="_UpstreamObjectStore")
 
 
-class ObjectStore(Store, Generic[T_Store]):
+class ObjectStore(Store, ConcurrencyLimiter, Generic[T_Store]):
```
d-v-b: @kylebarron lmk if these changes make sense, or if there's a better way to do this.

d-v-b:

Audit clarification: The assertion that local-file stores never need limits was too broad. Resource and workload constraints can make limits useful. Removing Zarr limits requires assessing custom backends that depended on them; this comment does not establish merge safety.

Original comment (historical):

I think a common pattern is for concurrency-sensitive backends to impose their own concurrency limits already. For example, icechunk does this: https://icechunk.io/en/stable/performance/#concurrency. And local-file stores like tifffile don't need concurrency limits at all. 

So my current read is that this PR is actually very low risk and therefore safe to merge.

src/zarr/storage/_obstore.py
kylebarron: I think this is a good example of what I think could be provided by an obspec adapter, through composition instead of inheritance.

Some pseudocode

```py
class ObspecConcurrencyLimiter:
    client: GetRangeAsync & GetRangesAsync & ...

    def get_range_async(self):
        async with semaphore:
            return self.client.get_range_async(...)
```

But I'm still not sure how that would work with different stores that provide different methods

d-v-b:
ping @zarr-developers/python-core-devs, it would be good to get feedback on this PR.

src/zarr/abc/codec.py
```diff
@@ -243,11 +243,8 @@ async def decode_partial(
         -------
         Iterable[NDBuffer | None]
         """
-        return await concurrent_map(
-            list(batch_info),
-            self._decode_partial_single,
-            config.get("async.concurrency"),
-        )
+        # Store handles concurrency limiting internally
```
TomAugspurger: nit: these comments are helpful for review now but will look odd in a while :)

src/zarr/storage/_fsspec.py
```diff
@@ -121,11 +126,14 @@ class FsspecStore(Store):
...
         self,
         fs: AsyncFileSystem,
+        *,
         read_only: bool = False,
         path: str = "/",
         allowed_exceptions: tuple[type[Exception], ...] = ALLOWED_EXCEPTIONS,
+        concurrency_limit: int | None = 50,
     ) -> None:
-        super().__init__(read_only=read_only)
+        Store.__init__(self, read_only=read_only)
```
TomAugspurger: This didn't need to change did it? Python's mro should ensure that `super().__init__` finds `Store` first right?

src/zarr/storage/_fsspec.py
```diff
@@ -67,7 +69,7 @@ def _make_async(fs: AbstractFileSystem) -> AsyncFileSystem:
     return AsyncFileSystemWrapper(fs, asynchronous=True)
 
 
-class FsspecStore(Store):
+class FsspecStore(Store, ConcurrencyLimiter):
```
TomAugspurger: Complicating things: fsspec *backends* often have a parameter to control the max concurrency. In which case, the best thing might be to configure that to use the native backends concurrency limiting mechanism and set zarr-python's `concurrency_limit=None`.

src/zarr/storage/_fsspec.py
```diff
@@ -121,11 +126,14 @@ class FsspecStore(Store):
     def __init__(
         self,
         fs: AsyncFileSystem,
+        *,
         read_only: bool = False,
         path: str = "/",
         allowed_exceptions: tuple[type[Exception], ...] = ALLOWED_EXCEPTIONS,
+        concurrency_limit: int | None = 50,
```
TomAugspurger: This seems like a default that would be useful to set through the config.

src/zarr/storage/_utils.py
```diff
@@ -1,17 +1,83 @@
...
+        async def get(self, key, prototype, byte_range=None):
+            # This will only run when semaphore permits
+            ...
+    ```
+    """
+
+    @functools.wraps(func)
+    async def wrapper(*args: P.args, **kwargs: P.kwargs) -> T_co:
+        self = args[0]
+        async with self._limit():  # type: ignore[attr-defined]
```
TomAugspurger: Could we use a `typing.Protocol` to indicate that the first argument needs a `_limit()` method with the right type?

I guess that might be overkill...

TomAugspurger (commented):
I didn't closely review the obstore changes, but overall I think this makes sense.

My only design question is whether the default of `50` should be in Python code, or looked up from the configuration.

src/zarr/storage/_fsspec.py
d-v-b: this would be my preference!

src/zarr/storage/_utils.py
d-v-b: nah i'm down to add that

d-v-b: done in e96e7ce15

src/zarr/storage/_fsspec.py
d-v-b: cc @martindurant -- is there a consistent way to control the max concurrency of fsspec backends?

d-v-b:
quick check: are there any storage backends where we need a concurrency limit defined in zarr-python? I'm wondering if this PR should actually just be removing any reference to concurrency limits, since this seems to be a storage backend concern.

Issue #337, opened 2018-11-19
Continuous benchmarking with ASV


alimanfoo:
It would be great to implement benchmarking with [ASV](https://github.com/airspeed-velocity/asv) or similar so we catch any changes that introduce a performance regression.

jakirkham:
Agreed. This would be great.

On a related note, [this storage comparison blogpost]( https://teonite.com/blog/optimize-neural-networks-i-o-dgx-research-in-teonite/ ) is pretty interesting.

jakirkham:
Agreed. This would be great.

On a related note, [this storage comparison blogpost]( https://teonite.com/blog/optimize-neural-networks-i-o-dgx-research-in-teonite/ ) is pretty interesting. Also here's a [notebook from the TileDB folks]( https://github.com/TileDB-Inc/tiledb-benchmarks/blob/master/bcolz-zarr/slicing-benchmarks.ipynb ).

Alt-Shivam:
Hey @jakirkham , 
The link for (storage comparison blogpost) is broken.
Can you provide me with the working one, If possible
Thanks

jakirkham:
Since that is not our blog, it is not something we can really affect. Maybe something like [archive.org]( https://archive.org ) has a mirror of it (though am not seeing one atm).

Alt-Shivam:
No worries @jakirkham 
Thanks for the reply.

d-v-b:
now that https://github.com/zarr-developers/zarr-python/pull/3562 is in, I think we can close this.

PR #3554, opened 2025-10-27
set up benchmarks with asv

d-v-b:
Sets up very basic benchmarks using [asv](https://asv.readthedocs.io/en/latest/index.html).

I don't want to write up all of the benchmarks we need in this PR. this is just setting up the basic infrastructure.

### DIFF
--- a/src/zarr/core/config.py
+++ b/src/zarr/core/config.py
@@ -36,19 +36,9 @@ from donfig import Config as DConfig
 if TYPE_CHECKING:
     from donfig.config_obj import ConfigSet
 
-    from zarr.core.dtype.wrapper import ZDType
-
 
 class BadConfigError(ValueError):
     _msg = "bad Config: %r"
-
-
-# These values are used for rough categorization of data types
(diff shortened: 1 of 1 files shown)


d-v-b:
closes #337

d-v-b:
there might be better / alternative tools than asv for this purpose, like [codspeed](https://codspeed.io/). I picked asv because that's what numpy is using, but the choice of tool is definitely a degree of freedom here.

maxrjones:
> there might be better / alternative tools than asv for this purpose, like [codspeed](https://codspeed.io/). I picked asv because that's what numpy is using, but the choice of tool is definitely a degree of freedom here.

IMO it depends on your goals. If you want to run comparisons between two branches locally then asv can be fine. If you want the benchmarks integrated into CI then codspeed would work better. I personally prefer codspeed because I've found it easier to integrate with existing tests rather than needing separate a benchmarking setup.

d-v-b:
generally I'm hearing more enthusiasm for something like codspeed instead of asv, and the argument in favor of benchmarks being integrated with our tests is persuasive to me, so I'm going to close this

Issue #2909, opened 2025-03-13
`async` from `sync` in zarr


ilan-gold:
### Zarr version

v3.0.5

### Numcodecs version

0.15.1

### Python Version

3.13

### Operating System

Mac

### Installation

`uv`

### Description

Hello! I'm trying to create an api around zarr async and I came across an edge case, maybe.  I would like to be able to open the zarr store synchronously (potentially), but then use it asynchronously (hopefully).  This seems to be specific to this store and not to zarr async i.e., the following works with a local file instead of an http server.  Any advice? It would be great to write code that is synchronous at the top level, but then hooks into async at a lower level.

The weird part is that most of the [anndata tests pass](https://github.com/scverse/anndata/actions/runs/13838464598/job/38719495723?pr=1902) with this paradigm - the problem only arises in a script, or ipython sessions.  Something to do with the event loop certainly? Any advice?

### Steps to reproduce

ran in ipython/jupyter
```python
import zarr.api
import zarr.api.asynchronous

g = await zarr.api.asynchronous.open_array(store="http://127.0.0.1:8080/path/to/file/data.zarr")
await g.getitem(Ellipsis) # works
g_sync = zarr.open_array(
    "http://127.0.0.1:8080/path/to/file/data"
)
await (
    await zarr.api.asynchronous.open_array(
        store=g_sync.store
    )
).getitem(Ellipsis) # fails
```
```pytb
Traceback (most recent call last):
  File "/Users/ilangold/Projects/Theis/anndata/tester.py", line 23, in <module>
    asyncio.run(main())
    ~~~~~~~~~~~^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.13/3.13.0_1/Frameworks/Python.framework/Versions/3.13/lib/python3.13/asyncio/runners.py", line 194, in run
    return runner.run(main)
           ~~~~~~~~~~^^^^^^
  File "/opt/homebrew/Cellar/python@3.13/3.13.0_1/Frameworks/Python.framework/Versions/3.13/lib/python3.13/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
  File "/opt/homebrew/Cellar/python@3.13/3.13.0_1/Frameworks/Python.framework/Versions/3.13/lib/python3.13/asyncio/base_events.py", line 721, in run_until_complete
    return future.result()
[...]

PR #3556, opened 2025-10-27
feat: make `async_array` public

ilan-gold:
Given that [a lot of people](https://github.com/search?q=+%22._async_array%22&type=code) are using the private `Array._async_array`, I think it should be documented and put in a public namespace. This PR specifically does not make `zarr.core.sync.sync` public because [I'm not 100% sure it's necessary in the day-and-age of `obstore`](https://github.com/zarr-developers/zarr-python/issues/2909#issuecomment-3444047144) but also because I can definitely see arguments why a better API might be useful, certain one that does:

```python
some_batching_api(my_sync_array.async_array.getitem(...), my_other_sync_array.async_array.getitem(...))
```
instead of 
```python
zarr.core.sync(asyncio.gather(my_sync_array.async_array.getitem(...), my_other_sync_array.async_array.getitem(...)))
```

Another reason I think making this public is because you may want to batch together these low level requests just because of the way people's APIs are structured rather than forcing everything to go through something like

```python
some_batching_api(sync_arrays, selections)
```

Definitely open to opinions on this! Marked as draft for now!

TODO:


TomAugspurger:
I'm fine with this, as long as we're OK making the implementation detail that `Array` is backend by an `AsyncArray` part of the public API.

I would also pretty strongly encourage people using `._async_array` to consider whether they can use AsyncArray directly. Maybe they can't (because they in turn are offering a sync API).

ilan-gold:
> I'm fine with this, as long as we're OK making the implementation detail that Array is backend by an AsyncArray part of the public API.

Another option would be to create a `to_async` function that could in theory be future proofed (although the `property` does the same thing) in the sense that we can remove the "is backed by `AsyncArray`" aspect of the implementation while maintaining API compat by implementing a "reopen" scheme under the `async_array` property/`to_async` function (however it is implemented).

> I would also pretty strongly encourage people using ._async_array to consider whether they can use AsyncArray directly. Maybe they can't (because they in turn are offering a sync API).

Agreed! I think the "offering a sync API" thing is the core of it, though.

ilan-gold:
I took @TomAugspurger 's comment quite seriously and updated the doc string for the proposed `async_array` property.  I think the promise should be, independent of the current implementation, a way to get an asynchronous "version" of a synchronous array synchronously + simply i.e., for library devs who want this async-behind-sync API like zarr has. In theory, then, if we ever decouple the two classes (`AsyncArray` and `Array`), we would still hold up our promise i.e., you get an async version of the synchronous array, but not the exact one "backing" the synchronous array.

As a counter example, highlighting that something like this is not really possible at the moment, API-wise, the [`__init__` method on `AsyncArray`](https://github.com/ilan-gold/zarr-python/blob/bf3b868179c1a81a5eb6c38ce3bcdffab38df335/src/zarr/core/array.py#L327-L343) (i.e., a synchronous method) takes three arguments and then sets up its own codec pipeline.  But one of those arguments, `config`, is private on the `AsyncArray` class (setting aside the fact that `_async_array` is private) i.,e.,:

```python
AsyncArray(my_sync_array.metadata, my_sync_array._async_array.store_path, my_sync_array._async_array._config)
```

So maybe an alternative would be a synchronous way to create an `AsyncArray` easily but I like the idea more of wrapping this in the `Array` class.

d-v-b (approved):

PR #3559, opened 2025-10-28
add bytes dtype

d-v-b:

Adds the `Bytes` data type with V3 name `"bytes"`. At the current head it writes fill values as base64-encoded strings and accepts both base64 strings and integer sequences when reading them.

`Bytes` is a separate class from `VariableLengthBytes`. It accepts the legacy `"variable_length_bytes"` V3 name and uses the same V2 dtype representation, `{"name": "|O", "object_codec_id": "vlen-bytes"}`. The registry defaults to `Bytes`; helpers allow switching between the registrations.

Reading the legacy name and writing the new name can change metadata spelling on a round trip. Support for these legacy forms does not establish that every historical bytes dataset is compatible.

The previous description's integer-array output and encoding questions described an earlier revision. The current implementation emits base64. The bytes dtype tests passed: 86 passed and 2 item-size capability skips in Python 3.12 minimal Hatch.


d-v-b:

Historical discussion, with the V2 key and compatibility wording corrected below. References to an upcoming spec change and release timing describe the original discussion.

I am assuming that https://github.com/zarr-developers/zarr-extensions/pull/38 will be merged soon, as it was approved by @normanrz. 

the `Bytes` data type now reads the tuple-of-ints JSON fill value defined in the current text of the `"bytes"` dtype spec, but it reads (and writes) the base64-encoded string form that the future version of the `"bytes"` dtype spec softly recommends.

The scope of this PR has broadened somewhat because we are effectively replacing one data type (`VariableLengthBytes`) with a new one (`Bytes`). 

These two data types cannot co-exist in the data type registry, because they have the same zarr v2 JSON representation (`{"name": "|O", "object_codec_id": "vlen-bytes"}`), and the registry treats "1 JSON input matches 2 or more data type classes" as an error.

So I removed `VariableLengthBytes` from the registry, and added logic to `Bytes` to ensure that the JSON string `"variable_length_bytes"` is an alias for `"bytes"`, which supports reading these legacy dtype names; this alone does not guarantee compatibility with every historical dataset.

For any users who are committed to the `VariableLengthBytes` data type for some reason, I added a [function](https://github.com/d-v-b/zarr-python/blob/4b8763471e5cddae53122f38c7a4384ade92882d/src/zarr/core/dtype/__init__.py#L296-L306) to reset the data type registry to the old state (de-registering `Bytes` and registering `VariableLengthBytes` in its place), and a [function](https://github.com/d-v-b/zarr-python/blob/4b8763471e5cddae53122f38c7a4384ade92882d/src/zarr/core/dtype/__init__.py#L309-L319) to reverse this transformation. We need to foreground these functions in our documentation for the bytes dtype.

The intent is to preserve the supported runtime behavior, but the distinct classes and changed registry defaults deserve compatibility testing. This _is_ a metadata representation change for anyone who relies on the round-trip integrity of old data -- data saved with `"data_type" : "variable_length_bytes"` will be resaved with `"data_type": "bytes"` after this change. For that reason I think we should ship this in a 3.2 release.

src/zarr/core/dtype/__init__.py
```diff
@@ -277,6 +284,36 @@ def parse_dtype(
         # If the dtype request is one of the aliases for variable-length UTF-8 strings,
         # return that dtype.
         return VariableLengthUTF8()  # type: ignore[return-value]
+    if dtype_spec is bytes:
+        # Treat the bytes type as a request for the Bytes dtype
+        return Bytes()
```
d-v-b, line 289: flagging this change -- `parse_dtype(bytes, zarr_format = 3)` will now return an instance of `Bytes`

src/zarr/core/dtype/__init__.py
```diff
@@ -277,6 +284,36 @@ def parse_dtype(
...
+    """
+    Unregister the old VariableLengthBytes dtype from the registry, and replace it with
+    the new Bytes dtype. Used to reverse the effect of enable_legacy_bytes_dtype
+    """
+    if (
+        "variable_length_bytes" in data_type_registry.contents
+        and "bytes" not in data_type_registry.contents
+    ):
+        data_type_registry.unregister("variable_length_bytes")
+        data_type_registry.register("bytes", Bytes)
```
d-v-b, line 319: these functions let users effectively disable the changes in this PR

src/zarr/core/dtype/common.py
```diff
@@ -151,12 +151,6 @@ def unpack_dtype_json(data: DTypeSpec_V2 | DTypeSpec_V3) -> DTypeJSON:
     return data
 
 
-class DataTypeValidationError(ValueError): ...
-
-
-class ScalarTypeValidationError(ValueError): ...
-
-
```
d-v-b, line 159: these errors were moved to the main errors module

src/zarr/errors.py
```diff
@@ -144,3 +146,13 @@ class BoundsCheckError(IndexError): ...
...
 class ArrayIndexError(IndexError): ...
+
+
+class DataTypeValidationError(ValueError): ...
+
+
+class ScalarTypeValidationError(ValueError): ...
+
+
+class DataTypeResolutionError(ValueError):
```
d-v-b, line 157: this is a new error class that models problems encountered during data type resolution, i.e. the process of mapping an input to one of the available data types. This is different from data type validation, which we have defined to mean the process of mapping an input to a specific data type.

PR #3561, opened 2025-10-30
optimize shard writing

nbren12:
Writing to sharded arrays was up to 10x slower for largish chunk sizes because the _ShardBuilder object has many calls to np.concatenate. This commit coalesces these into a single concatenate call, and improves write performance by a factor of 10 on the benchmarking script in #3560.

Added a new core.Buffer.combine API

Resolves #3560

[Description of PR]

TODO:
* [x] Add unit tests and/or doctests in docstrings
* [x] Add docstrings and API docs for any new/modified user-facing classes and functions
* [x] New/modified features documented in `docs/user-guide/*.md`
* [x] Changes documented as a new file in `changes/`
* [x] GitHub Actions have all passed
* [ ] Test coverage is 100% (Codecov passes)

Issue #3524, opened 2025-10-15
Array indexing with Zarr 3 is noticeably slower than with Zarr 2


b8raoult:
### Zarr version

3.1.3

### Numcodecs version

0.15.1

### Python Version

3.12.9

### Operating System

Linux

### Installation

uv pip install

### Description

Timing comparisons between Zarr 2 and Zarr 3 of various indexing show that version 3 is always slower than version 2. I have attached the code to run the comparisons, and some results. All tests are run in memory and the compressors are disabled.

You can see that for example that accessing `data[::step]` is much slower (from 0.5s to 3.9s) . It may be due to the overhead of switching between synchronous and asynchronous code, but this is just a guess.

### Steps to reproduce

```python
import zarr
import numpy as np
import timeit
import inspect

version = int(zarr.__version__.split(".")[0])

values = np.ones(shape=(60632, 4, 1, 10))

root = zarr.group()

if version < 3:
    data = root.create_dataset("data", data=values, shape=values.shape, compressor=None)
else:
    data = root.create_array("data", data=values, compressors=None)


start = 15
end = data.shape[0] - 15
step = data.shape[0] // 10


def set_values():
    data[:] = 2


tests = [
    lambda: data[0:10, :, 0],
    lambda: data[:, 0:3, 0],
    lambda: data[0:10, 0:3, 0],
    lambda: data[:, :, :],
    lambda: data[0],
    lambda: data[0, :],
    lambda: data[0, 0, :],
    lambda: data[0, 0, 0, :],
    lambda: data[start:end:step],
    lambda: data[start:end],
    lambda: data[start:],
    lambda: data[:end],
    lambda: data[::step],
    lambda: set_values(),
]

for i, t in enumerate(tests):
    src = inspect.getsourcelines(t)[0][0].strip().replace("lambda:", "").strip(",")

    elapsed = timeit.timeit(t, number=1000)
    print(f"zarr{version}: {src:22}: {elapsed:10.4f} seconds")

```

### Additional output

Running the test above with Zarr 2.18.7:

```
zarr2:  data[0:10, :, 0]     :     0.1546 seconds
zarr2:  data[:, 0:3, 0]      :     4.1608 seconds
zarr2:  data[0:10, 0:3, 0]   :     0.1267 seconds
zarr2:  data[:, :, :]        :     6.2516 seconds
zarr2:  data[0]              :     0.1478 seconds
zarr2:  data[0, :]           :     0.1492 seconds
[...]

PR #3562, opened 2025-10-30
add benchmarks using pytest-benchmark and codspeed

d-v-b:
since #3554 was an unpopular direction I'm going instead with codspeed + pytest-benchmark. Opening as a draft because I haven't looked into how codspeed works at all, but I'd like people to weigh in on whether these initial benchmarks make sense. Naturally we can add more specific ones later, but I figured just some bulk array read / write workloads would be a good start.

### DIFF
--- a/.github/workflows/codspeed.yml
+++ b/.github/workflows/codspeed.yml
@@ -22,6 +22,8 @@ jobs:
         fetch-depth: 0  # grab all branches and tags
     - name: Set up Python
       uses: actions/setup-python@v6
+      with:
+        python-version: "3.11"
     - name: Install Hatch
       run: |
         python -m pip install --upgrade pip
@@ -30,4 +32,4 @@ jobs:
       uses: CodSpeedHQ/action@v4
       with:
           mode: walltime
-          run: hatch run test.py3.11-1.26-minimal:pytest tests/benchmarks --codspeed
+          run: hatch run test.py3.11-2.0-minimal:pytest tests/benchmarks --codspeed


d-v-b:
@zarr-developers/steering-council I don't have permission to register this repo with codspeed. I submitted a request to register it, could someone approve it?

normanrz:
> @zarr-developers/steering-council I don't have permission to register this repo with codspeed. I submitted a request to register it, could someone approve it?

done

d-v-b:
does anyone have opinions about benchmarks? feel free to suggest something concrete. Otherwise, I think we should take this as-is and deal with later benchmarks (like partial shard read / writes) in a subsequent pr

.github/workflows/codspeed.yml
```diff
@@ -0,0 +1,30 @@
...
+      uses: actions/setup-python@v6
+    - name: Install Hatch
+      run: |
+        python -m pip install --upgrade pip
+        pip install hatch
+    - name: Run the benchmarks
+      uses: CodSpeedHQ/action@v4
+      with:
+          mode: instrumentation
+          run: hatch run test.py3.11-1.26-minimal:run-benchmark
```
dcherian: can we test the latest instead? seems more appropriate...

dcherian:
> feel free to suggest something concrete

indexing please. that'll exercise the codec pipeline too. 

a peakmem metric would be good to track also, if possible.

.github/workflows/codspeed.yml
d-v-b: The latest version of python? What's the reasoning? I'd rather update this file when we drop a supported version vs when a new version of python comes out.

dcherian: Because we'd want to catch a perf regression from upstream changes too? I'm suggested latest version of released libraries `py=3.13, np=2.2`

d-v-b: [we don't have an upper bound](https://github.com/zarr-developers/zarr-python/blob/b3e9aed305092236c5db70deee0b26dad648d3b0/pyproject.toml#L36) on numpy versions, so I don't think this particular workflow will help us catch regressions from upstream changes -- we would need to update this workflow every time a new version of numpy is released. IMO that's something we should do in a separate benchmark workflow. This workflow here will run on every PR, and in that case the oldest version of numpy we support seems better.

we also don't have to use a pre-baked hatch environment here, we could define a dependency set specific to benchmarking. but my feeling is that benchmarking against older versions of stuff gives us a better measure of what users will actually experience.

d-v-b:
> > feel free to suggest something concrete
> 
> indexing please. that'll exercise the codec pipeline too.
> 
> a peakmem metric would be good to track also, if possible.

I don't think codspeed or pytest-benchmark do memory profiling. we would need https://pytest-memray.readthedocs.io/en/latest/ or something equivalent for that.

and an indexing benchmark sounds like a great idea but I don't think I have the bandwidth for it in this pr right now

d-v-b:
I added a benchmark that clearly reveals the performance improvement of #3561

d-v-b:
I added some slice-based benchmarks based on the examples from https://github.com/zarr-developers/zarr-python/issues/3524, and I updated the contributing docs with a section about the benchmarks. assuming we can resolve the discussion about which python / numpy version to use in the CI job, I think this is ready

d-v-b:
new problem: the codspeed CI benchmarks are way too slow! the benchmark suite runs in 90s locally, and It's taking over 40m to run in CI. Help would be appreciated in speeding this up.

d-v-b:
owing to the large number of syscalls in our benchmark code, codspeed recommended using the [walltime instrument](https://codspeed.io/docs/instruments/walltime) instead of their virtual CPU instrument. But to turn on the walltime benchmark, we would need to[ run our benchmarking code on codspeed's servers](https://codspeed.io/docs/instruments/walltime#usage-on-public-repositories), which is a security risk.

Given that codspeed is not turning out to be particularly simple, I am inclined to defer the codspeed CI stuff for later work. But if someone can help get the test runtime down, and / or we are OK running our benchmarks on codspeed's servers, then maybe we can get that sorted in this PR.

.github/workflows/codspeed.yml
```diff
@@ -0,0 +1,30 @@
...
+on:
+  push:
+    branches:
+      - "main"
+  pull_request:
+  # `workflow_dispatch` allows CodSpeed to trigger backtest
+  # performance analysis in order to generate initial data.
+  workflow_dispatch:
+
+jobs:
```
maxrjones, line 15:
```suggestion
permissions:
  contents: read

jobs:
```

d-v-b:
looks like the walltime instrument is working! I think this is g2g

joshmoore:
(Enabled the app)

docs/contributing.md
```diff
@@ -230,4 +230,11 @@ If an existing Zarr format version changes, or a new version of the Zarr format
...
-[https://github.com/zarr-developers/zarr-python/issues/new?template=release-checklist.md](https://github.com/zarr-developers/zarr-python/issues/new?template=release-checklist.md>). The release checklist includes all steps necessary for the release.
\ No newline at end of file
+[https://github.com/zarr-developers/zarr-python/issues/new?template=release-checklist.md](https://github.com/zarr-developers/zarr-python/issues/new?template=release-checklist.md>). The release checklist includes all steps necessary for the release.
+
+## Benchmarks
+
+Zarr uses [pytest-benchmark](https://pytest-benchmark.readthedocs.io/en/latest/) for running
+performance benchmarks as part of our test suite. The benchmarks can be are found in `tests/benchmarks`.
+By default pytest is configured to run these benchmarks as plain tests (i.e., no benchmarking). To run
+a benchmark with timing measurements, use the `--benchmark-enable` when invoking `pytest`.
```
maxrjones:
```suggestion
a benchmark with timing measurements, use the `--benchmark-enable` when invoking `pytest`.

The benchmarks are run as part of the continuous integration suite through [codspeed](https://codspeed.io/zarr-developers/zarr-python).
```

d-v-b:
> IMO it'd be better to skip the tests/benchmarks during regular test runs in the interest of speed

i think this makes sense -- on my workstation the current benchmark suite takes 40s to run as regular tests, which is a big addition to our total test runtime. The latest changes to this branch skip the `test/benchmarks` folder by default when running our main test suite and the gpu tests.

d-v-b:
I'm seeing some flakiness in the codspeed CI run: https://github.com/zarr-developers/zarr-python/actions/runs/21025818794/job/60449955113#step:5:253. Since this is a codspeed-specific issue (from what I can tell), and it doesn't detract from the utility of the benchmark suite for local benchmarking, I'm going to work around this failure by ignoring this warning.

PR #3564, opened 2025-11-01
chore: spec0 compat (python 3.14 compat, python 3.12 min)

### LINKED ISSUE #3512: SPEC 0 minimum supported python version
[SPEC 0](https://scientific-python.org/specs/spec-0000/) recommends that we drop support for a Python version for 3 years after its initial release. That means dropping support for python 3.11 ([released](https://www.python.org/downloads/release/python-3110/) on October, 24 2022) soon. 

Do we want to follow the SPEC 0 recommendation and drop support for python 3.11 soon? It's just a recommendation, so we have some leeway here. 

And if we don't want to follow the SPEC 0 minimum python version recommendation, when _would_ we drop python 3.11? 4 years after release? 5? 

I would like to be guided by a signal from downstream projects. If literally everyone downstream has already dropped python 3.11 (big doubt), then we could drop 3.11 painlessly. If _nobody_ is ready to drop python 3.11, then we incur the maximal pain if _we_ drop it. We need to put a decision boundary at some point between these two extremes.

ilan-gold:
Sorry if someone already started this! I didn't see it in the repo and I was trying to play around with python 3.14 until I realized the feature I was interested in had actually been released in 3.13.  So just pushing here if anyone wants to pick it up (or I can finish depending on the workload).

Fixes #3512

Some questions:

1. I don't think cupy is 3.14 compat yet: https://github.com/cupy/cupy/issues/9346 but does this matter? Does zarr promise the same level of python support across devices
2. Is there interest in `uv`-ifying the CI? Blockers?

Locally most tests pass, and the ones that don't looked a bit me-specific (for example, a numcodecs-related test though I had python 2.18 installed? Need to look into that).

TODO:

### DIFF
--- a/src/zarr/abc/codec.py
+++ b/src/zarr/abc/codec.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 from abc import abstractmethod
 from collections.abc import Mapping
-from typing import TYPE_CHECKING, Protocol, TypeGuard, runtime_checkable
+from typing import TYPE_CHECKING, Literal, Protocol, TypeGuard, runtime_checkable
 
 from typing_extensions import ReadOnly, TypedDict
 
@@ -35,6 +35,13 @@ __all__ = [
     "GetResult",
     "SupportsSyncCodec",
 ]
+
+
+class GetResult(TypedDict):
+    """Metadata about a store get operation."""
+
+    status: Literal["present", "missing"]
+
 
 type CodecInput = NDBuffer | Buffer
 type CodecOutput = NDBuffer | Buffer


d-v-b:
> 2\. Is there interest in `uv`-ifying the CI? Blockers?

i'm ok with uv-fying things. we are using hatch for defining a local test matrix, but we don't have to use that in ci

ilan-gold:
I tried removing the `covariant=True` stuff in https://github.com/zarr-developers/zarr-python/pull/3564/commits/98404bdc34026ab8d6e6c5c83cffaa033c69c654 but got some strange errors, which I imagine have to do with my lack of knowledge of numpy dtype inheritance i.e., things like:

```python
tests/test_metadata/test_v2.py:224: error: Argument "dtype" to "ArrayV2Metadata" has incompatible type "Int16"; expected "ZDType[dtype[generic], generic | str | bytes]"  [arg-type]
```

would love to understand this! I will try to look into it soon

joshmoore:
As a side note, recently there was a request (@d-v-b, do you remember where?) of considering using the PSF versioning (https://devguide.python.org/versions/) rather than the SPEC. Just mentioning in case it has an impact on when we would decide to drop, etc.

d-v-b:
@joshmoore see also https://github.com/zarr-developers/zarr-python/issues/3512

ilan-gold:
> I tried removing the covariant=True stuff in https://github.com/zarr-developers/zarr-python/commit/98404bdc34026ab8d6e6c5c83cffaa033c69c654 but got some strange errors

To clarify, the reason I find them strange is that the python docs make it seem like you can just get rid of `covariant` (and if you can't you're misusing it).  Maybe I misunderstood.  I'll try to look into it, but just figured I'd document my stopping point.

d-v-b:
based on [these docs](https://peps.python.org/pep-0695/#variance-inference), my feeling is that in the context of `ZDType[A, B]`, `A` and `B` should be inferred as ~~contravariant~~ covariant iff `A` and `B` only appear as return types of the `ZDType` methods, which I'm pretty sure is the case? Since I needed to do that in 3.11 to make mypy happy.

ilan-gold:
> iff A and B only appear as return types of the ZDType methods

I feel quite lost here - I would think similar but this seems to work (maybe I'm missing something) whereas the (more complicated) `zarr-python` one doesn't:

```python
class Base1: ...
class SubBase1(Base1): ...
class SubSubBase1(SubBase1): ...

class Base2: ...
class SubBase2(Base2): ...
class SubSubBase2(SubBase2): ...

class F[T: Base1, S: Base2]: ...

class FSub(F[SubBase1, SubBase2]): ...

class FSubSub(F[SubSubBase1, SubSubBase2]): ...

type GType = F

g: GType = FSubSub()
```

So maybe I'm misunderstanding.

ilan-gold:
> only appear as return types of the ZDType methods

Totally misunderstood this in our context, but I think I got it now! I made `dtype_cls` not dependent on the generic (I think) https://github.com/zarr-developers/zarr-python/pull/3564/commits/86095bbdcaf487ecb22d4518c0eb6e453758acae#diff-bb2825e8577e13fd91c0cab0696d2c2736686dc7995484285290b6bee5c4190aR73

pyproject.toml
```diff
@@ -29,11 +29,11 @@ maintainers = [
     { name = "Tom Augspurger", email = "tom.w.augspurger@gmail.com" },
     { name = "Deepak Cherian" }
 ]
-requires-python = ">=3.11"
-# If you add a new dependency here, please also add it to .pre-commit-config.yaml
+requires-python = ">=3.12"
+# If you add a new dependency here, please also add it to .pre-commit-config.yml
```
Zethson: It's still called `.yaml` actually.

Image

ilan-gold:
Ok there are a lot of things that break now with `mypy` (I'm honestly not sure why, maybe the `mypy` upgrade + required version + python > 3.12) but `NDArrayLike` seems very incompatible with `np.ndarray`.  I'll try to finish that up

I'm open to alternatives rather than manually copy-pasting the numpy API - I'm also not sure what it means to add arguments to `NDArrayLike` for compat - seems like a breaking change.

Then there's all this dtype stuff, which really confuses me.  Things like `got "dtype[signedinteger[_16Bit]]", expected "Int16DType"` seem wrong to me (I would naively think these are the same thing but apparently not). So I'll need to look into the `numpy` docs for that

UPDATE: The situation with `NDArrayLike` is fairly bleak - methods like `ravel` change the change but are currently typed as returning `Self` (same goes for `reshape`), which doesn't pick up the generified shape type that numpy now uses.

ilan-gold:
I've also checked locally and these 65 minute tests take 5minutes and pass (and I have a macbook) so that's another thing!  To summarize:

1. `NDArrayLike` typing is wrong for things like `ravel` - is there a way to turn this off in `numpy`? I am having trouble making it work using `-> NDArrayLike[...]` i.e., numpy expects `-> np.ndarray[tuple[int], dtype]` because ravel flattens but currently we return `Self` which is wrong then if `Self` has ndims > 1
2. See #3688
3. The (unreproducible locally) failures on `os=macos-latest, py=3.14, deps=optional`

dstansby:
Is it possible to keep the numpy pin in `pre-commit-config.yaml` for now, and just make the other changes independently? I'm wondering if punting the typing issues down the track is a way to get this moving again.

ilan-gold:
> Is it possible to keep the numpy pin in pre-commit-config.yaml for now, and just make the other changes independently? I'm wondering if punting the typing issues down the track is a way to get this moving again.

It's definitely possible this helps things.  This PR definitely bubbled up some issues (added #3780 as well) so now that we have those documented, we might be able to move forward incrementally. Great idea!

ilan-gold:
Ok should be all green on typing issues (at least it is locally for me).

I'd vote for taking the same approach to the unreproducible macOS tests if they are still there (will have to wait a bit to see) and then getting a review!

Thanks @dstansby for the push!

ilan-gold:
There appears to be a build error from the `uv run`-type tests with the `macos-latest` building `numcodecs`, something like:

```
2026-03-14T18:08:32.9270440Z E               clang: error: linker command failed with exit code 1 (use -v to see
2026-03-14T18:08:32.9270490Z E               invocation)
2026-03-14T18:08:32.9270620Z E               [numcodecs] command '/usr/bin/clang' failed with exit code 1
2026-03-14T18:08:32.9270780Z E               Traceback (most recent call last):
2026-03-14T18:08:32.9270830Z E                 File
2026-03-14T18:08:32.9271090Z E               "/Users/runner/.cache/uv/builds-v0/.tmp8J7NDJ/lib/python3.14/site-packages/setuptools/_distutils/spawn.py",
2026-03-14T18:08:32.9271150Z E               line 83, in spawn
2026-03-14T18:08:32.9271260Z E                   subprocess.check_call(cmd, env=_inject_macos_ver(env))
2026-03-14T18:08:32.9271350Z E                   ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2026-03-14T18:08:32.9271390Z E                 File
2026-03-14T18:08:32.9271570Z E               "/Library/Frameworks/Python.framework/Versions/3.14/lib/python3.14/subprocess.py",
2026-03-14T18:08:32.9271680Z E               line 419, in check_call
2026-03-14T18:08:32.9271760Z E                   raise CalledProcessError(retcode, cmd)
2026-03-14T18:08:32.9271920Z E               subprocess.CalledProcessError: Command '['/usr/bin/clang', '-bundle',
```
although there is 67MB of logs now attached to the latest test! Anyway, `numcodecs` [doesn't build or test against `macos-latest`](https://github.com/search?q=repo%3Azarr-developers%2Fnumcodecs%20macos-latest&type=code) it seems, so maybe we shouldn't either? Open to suggestions here!

maxrjones:
Are most of the changes here related to dropping Python 3.12, adding 3.14 support, or a mix of both? I'm interested in releasing explicit support for 3.14 in the near future and am wondering if decoupling dropping 3.12 from supporting 3.14 would help

ilan-gold:
@maxrjones I think at this point, the big blocker is what we do about the `numcodecs` build issue.  Other than that, everything seems to be kosher now that we've reverted the `numpy` type checking.  Most changes are from the type checking moving to 3.12, yes.

I can look at `numcodecs` tomorrow closer, or just disable `macos-latest` here and leave a note (pointing to an issue in `numcodecs`) that the build seems to be broken in this one case.  Presumably this will make everything green. Let me know what you think! This is a problem independent of the changes here - turning on 3.14 tests is the culprit.

maxrjones:
> @maxrjones I think at this point, the big blocker is what we do about the `numcodecs` build issue. Other than that, everything seems to be kosher now that we've reverted the `numpy` type checking. Most changes are from the type checking moving to 3.12, yes.
> 
> I can look at `numcodecs` tomorrow closer, or just disable `macos-latest` here and leave a note (pointing to an issue in `numcodecs`) that the build seems to be broken in this one case. Presumably this will make everything green. Let me know what you think! This is a problem independent of the changes here - turning on 3.14 tests is the culprit.

thanks for the overview. I think the issues in zarr-python's CI could be resolved by releasing numcodecs with a python 3.14 compatible wheel. But I'm finding numcodecs needs more care before that happens 😢

ilan-gold:
Oh wow @maxrjones I could have sworn I checked that `numcodecs` was building against `3.14` ok, yeah, definitely a blocker!

Update: https://github.com/zarr-developers/numcodecs/commit/2e7107e0f36f2ff260a06f198610bf1a1454a3cd it is running yes, I'm not crazy. And the wheels are on pypi: https://pypi.org/project/numcodecs/0.16.5/#files

The issue really does seem to be that numcodecs does not build against `macos-latest`

dstansby:
For some odd reason the failing macos test is trying to build wheels for numcodecs 0.15.0:

```
2026-03-14T18:08:27.3272750Z E            Building numcodecs==0.15.0
2026-03-14T18:08:27.3276510Z E           × Failed to build `numcodecs==0.15.0`
```

dstansby:
Ah, this is because it's trying to run the script here:

https://github.com/zarr-developers/zarr-python/blob/536ce9dbfb573f9f99a836bc1417a41566cb3426/tests/test_regression/scripts/v2.18.py#L1-L7

ilan-gold:
Ah that makes so much more sense @dstansby thank you!

dstansby:
Perhaps the most pragmatic fix is to just skip that particular test on macOS?

ilan-gold:
Awesome @dstansby everything looks like it's ready to go here then :)

maxrjones:
> Update: https://github.com/zarr-developers/numcodecs/commit/2e7107e0f36f2ff260a06f198610bf1a1454a3cd it is running yes, I'm not crazy. And the wheels are on pypi: [pypi.org/project/numcodecs/0.16.5#files](https://pypi.org/project/numcodecs/0.16.5/#files)

oops, sorry 🤦‍♂️

ilan-gold:
@dstansby do you want to review here or someone else maybe?

src/zarr/abc/codec.py
```diff
@@ -491,21 +480,21 @@ async def write(
         ...
 
 
-async def _batching_helper(
-    func: Callable[[CodecInput, ArraySpec], Awaitable[CodecOutput | None]],
-    batch_info: Iterable[tuple[CodecInput | None, ArraySpec]],
-) -> list[CodecOutput | None]:
+async def _batching_helper[CI: CodecInput, CO: CodecOutput](
```
dcherian, line 490: missed opportunity to go full on C++ with `cin` and `cout`

d-v-b:
thanks @ilan-gold and @dcherian!

Issue #2453, opened 2024-10-30
Add a "feature request" issue type


dstansby:
I just went to open a new feature request for zarr-python, but the [new issue list](https://github.com/zarr-developers/zarr-python/issues/new/choose) doesn't have the option to request a feature in the `zarr-python` library: 

![Screenshot 2024-10-30 at 17 34 32](https://github.com/user-attachments/assets/24fafd53-cfaf-4310-aff5-09899ab22835)

Not all feature requests are for the spec, so I think we should add another feature request issue type which is for requesting a feature in the `zarr-python` library specifically.

PR #3572, opened 2025-11-06
Add an issue template for feature requests

dstansby:
Previously the list of issue types only allowed for feature reqeusts to be divered to `zarr-specs`, but I think there's definitely feature requests that are valid for zarr-python. I tried to delineate between these two types of feature reqeusts, but there may well be better ways to word this.

Fixes https://github.com/zarr-developers/zarr-python/issues/2453


d-v-b:
good idea, thank you!

PR #3573, opened 2025-11-06
Add properties to get chunk and shard slices

### LINKED ISSUE #2454: Add easy way to iterate through every chunk of an array
I am finding myself wanting to apply an operation (e.g., thresholding) on every chunk in an array. It would therefore be nice if there was an easy way to iterate through each chunk of the array.

This is sort of similar to [block indexing](https://zarr.readthedocs.io/en/stable/tutorial.html#block-indexing), but iterating over `Array.blocks` only iterates over the first axis of blocks, and gives you back all the chunks along the other axes.

dstansby:
Fixes https://github.com/zarr-developers/zarr-python/issues/2454.

### DIFF
--- a/src/zarr/core/array.py
+++ b/src/zarr/core/array.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import json
 import warnings
 from asyncio import gather
-from collections.abc import Iterable, Mapping
+from collections.abc import Generator, Iterable, Mapping
 from dataclasses import dataclass, field, replace
 from itertools import starmap
 from logging import getLogger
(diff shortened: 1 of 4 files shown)


src/zarr/core/array.py
```diff
@@ -1381,6 +1382,32 @@ async def example():
     async def nbytes_stored(self) -> int:
         return await self.store_path.store.getsize_prefix(self.store_path.path)
 
+    @property
+    def chunk_slices(self) -> Generator[tuple[slice, ...]]:
```
d-v-b, line 1386: nitpicking the name: i find "slice" to be kind of ambiguous between a verb and a noun, and it also locks us in to returning slice objects, so what if we use the word `"region"` instead? 

and also I think it's helpful if the name of this routine makes it clear that it's an iterator. So what if we call it `iter_chunk_regions`, i.e. the name of the routine it wraps 😜

dstansby: re. name, I think "slices" is unambiguously a noun because it's plural? I looked at NumPy (https://numpy.org/doc/stable/user/basics.indexing.html#slicing-and-striding), and they use the terms "index", "selection tuple", or "slicing tuple". If we try and stay consistent with NumPy, how about "chunk_indices"?

I like "regions", but to me it's ambiguous whether that means the index, or the array data at that index. If we settle on it that could be fixed by documentation and consistent use though.

re. iterator, do you mean generator? A list/string etc. are also iterators, but using a `yield` makes this into more specifically a generator.

So perhaps `generate_chunk_indices`? I find that a bit clunky though,

```python
for chunk_index in arr.chunk_indices:
```

is much nicer than

```python
for chunk_index in arr.generate_chunk_indices:
```
(or `iter_chunk_indices`)

which is why I prefer simply `chunk_indices`. I don't knkow if there's prior art in other libraries for naming generators?

d-v-b: generators _are_ iterators, and my thinking was that putting "iter" in the name conveys that users should expect to _iterate_ over the value returned by calling this method. 

when I think of iterating over indices, I think of iterating over tuples of coordinates, e.g., `(0, 0, 0)`, `(0, 0, 1)`, .... which is what `_iter_chunk_coords` / `_iter_shard_coords` do right now.

d-v-b: > I like "regions", but to me it's ambiguous whether that means the index, or the array data at that index. If we settle on it that could be fixed by documentation and consistent use though.

I worry that this ambiguity will hold for any name we pick :)

PR #3574, opened 2025-11-06
feat: bounded size `auto` sharding via `zarr.config`

### LINKED ISSUE #3568: Make `auto` sharding configurable by memory size
Discussion started here around `auto` settings for sharding: https://ossci.zulipchat.com/#narrow/channel/423692-Zarr/topic/Sharding.20performance.20issues/near/553649122

At the moment, the default is [twice the number of chunks along each axis](https://github.com/zarr-developers/zarr-python/blob/eb1b6e869115925f7b653e60bd542abd410a9a4e/src/zarr/core/chunk_grids.py#L240-L247) i.e., 2 chunks per shard for one axis, 4 for 2 etc.

The proposal would be to add something like `zarr.config.max_auto_shard_size` that is used when someone passes in `shards="auto"`.

In additional, we could have min sizes as well potentially, see the [arguments to `_guess_chunks`](https://github.com/zarr-developers/zarr-python/blob/eb1b6e869115925f7b653e60bd542abd410a9a4e/src/zarr/core/chunk_grids.py#L33) which has both

ilan-gold:
closes #3568

TODO:


ilan-gold:
Not sure I fully grasp what is going on with codecov, it looks like the tests run and assert the things they should and use the funcionality promised.

melonora:
In SpatialData we could also have a use case for this. Many users are not as experienced with the whole concept of sharding and chunking, so having the option to provide a maximum size with some explanation in documentation would be very beneficial.

ilan-gold:
I was curious how `xarray` handles sharding and from what I can tell it is also manually done there:

https://github.com/pydata/xarray/blob/6da8c2c92326d05b03bffa1a022d9f87f4f6278d/xarray/tests/test_backends.py#L2891-L2907

@jhamman Is this true? Would `xarray` benefit from a feature like this or are we thinking about the problem wrong and people _should_ be doing this manually?

melonora:
@ilan-gold I think that is correct at the moment. See also https://github.com/pydata/xarray/issues/9947. In dask I am working on having the interface just provide a `shard_factor` argument. I think the interface mostly just depends on the user. Some users like to think more in sizes, particularly when they are lessed experienced with chunking / sharding as they can relate it easier to the capability of their hardware.

tests/test_config.py
```diff
@@ -53,6 +53,7 @@ def test_config_defaults_set() -> None:
                 "array": {
                     "order": "C",
                     "write_empty_chunks": False,
+                    "max_bytes_per_shard_for_auto_sharding": None,
```
d-v-b: can we find a denser name for this parameter? Something like `max_shard_size_bytes`?

src/zarr/core/chunk_grids.py
```diff
@@ -202,6 +203,22 @@ def get_nchunks(self, array_shape: tuple[int, ...]) -> int:
         )
 
 
+def _guess_num_chunks_per_axis_shard(
+    chunk_shape: tuple[int, ...], item_size: int, max_bytes: int, array_shape: tuple[int, ...]
```
d-v-b, line 207: can we get a simple docstring that explains what this function is supposed to do?

tests/test_config.py
ilan-gold: Also perhaps something more accurate? Like `target_shard_size_bytes`? It's not meant as a bound but a target

d-v-b (approved):

ilan-gold:
@d-v-b What is the take on the fact that this is pre-compression? Something I just realized messing around with this.  Should we document that more clearly?

d-v-b:
> @d-v-b What is the take on the fact that this is pre-compression? Something I just realized messing around with this. Should we document that more clearly?

Maybe just indicate in the docs that this setting governs the in-memory size? 

Also, how will this handle data types that don't have a computable size per element (like variable-length bytes)?

ilan-gold:
> Maybe just indicate in the docs that this setting governs the in-memory size?

#3599

> Also, how will this handle data types that don't have a computable size per element (like variable-length bytes)?

It takes in the same arguments as `_auto_partition` so `itemsize` is derived from the zarr data type when it inherits from `HasItemSize` as far as I can tell

https://github.com/zarr-developers/zarr-python/blob/7d8105fbc494060c4faf8e633cc80e71511582e3/src/zarr/core/array.py#L4678-L4680

The docs say for chunking:

```
If you are feeling lazy, you can let Zarr guess a chunk shape for your data by providing chunks='auto', although please note that the algorithm for guessing a chunk shape is based on simple heuristics and may be far from optimal
```

I could add an explanation of the heuristic for both sharding and chunking a bit here, or at least specify what to expect based on data type.

Issue #3583, opened 2025-11-17
Can not create Metadata for structured dtype containing subarray dtype


sehoffmann:
### Zarr version

v3.1.3

### Numcodecs version

-

### Python Version

3.12

### Operating System

Linux

### Installation

uv

### Description

Output:

```
 Traceback (most recent call last):
  File "bug.py", line 19, in <module>
    arr = zarr.create_array(store, name='test', shape=(10,), dtype=DTYPE)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "zarr/api/synchronous.py", line 962, in create_array
    sync(
  File "zarr/core/sync.py", line 159, in sync
    raise return_result
  File "zarr/core/sync.py", line 119, in _runner
    return await coro
           ^^^^^^^^^^
  File "zarr/core/array.py", line 4933, in create_array
    return await init_array(
           ^^^^^^^^^^^^^^^^^
  File "zarr/core/array.py", line 4747, in init_array
    meta = AsyncArray._create_metadata_v3(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "zarr/core/array.py", line 772, in _create_metadata_v3
    fill_value_parsed = dtype.default_scalar()
                        ^^^^^^^^^^^^^^^^^^^^^^
  File "zarr/core/dtype/npy/structured.py", line 419, in default_scalar
    return self._cast_scalar_unchecked(0)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "zarr/core/dtype/npy/structured.py", line 373, in _cast_scalar_unchecked
    res = np.array([data], dtype=na_dtype)[0]
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: a bytes-like object is required, not 'int'
```

**Case #2 `fill_value=(0, np.nan)`:**

```
zarr/core/dtype/npy/structured.py", line 371, in _cast_scalar_unchecked
    res = np.array([tuple(data)], dtype=na_dtype)[0]
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: a bytes-like object is required, not 'float'
```

**Case #3 `fill_value=0`**
TypeError: a bytes-like object is required, not 'int'

**Case #4 `fill_value=None`**
[...]

PR #3587, opened 2025-11-20
Experimental Support for Subarray DTypes

### LINKED ISSUE #3582: Subarray dtypes get lost on serialization / casted to void type
### Zarr version

v3.1.3

### Numcodecs version

v0.15.1

### Python Version

3.12.10

### Operating System

Linux

### Installation

uv / pip

### Description

Subarray dtypes are not properly serialized but are cast to raw bytes / void dtype upon serialization. Subsequent access, hence, does not yield arrays with the proper shapes.

**Output:**
```
zarr/core/dtype/npy/structured.py:318: UnstableSpecificationWarning: The data type (Structured(fields=(('a', Int32(endianness='little')), ('b', RawBytes(length=100))))) does not have a Zarr V3 specification. That means that the representation of arrays saved with this data type may change without warning in a future version of Zarr Python. Arrays stored with this data type may be unreadable by other Zarr libraries. Use this data type at your own risk! Check https://github.com/zarr-developers/zarr-extensions/tree/main/data-types for the status of data type specifications for Zarr V3.
  v3_unstable_dtype_warning(self)
zarr/core/dtype/npy/bytes.py:785: UnstableSpecificationWarning: The data type (RawBytes(length=100)) does not have a Zarr V3 specification. That means that the representation of arrays saved with this data type may change without warning in a future version of Zarr Python. Arrays stored with this data type may be unreadable by other Zarr libraries. Use this data type at your own risk! Check https://github.com/zarr-developers/zarr-extensions/tree/main/data-types for the status of data type specifications for Zarr V3.
  v3_unstable_dtype_warning(self)
Original dtype: [('a', '<i4'), ('b', '<f4', (5, 5))]
Array created with dtype: [('a', '<i4'), ('b', 'V100')]
Accessed item dtype:  [('a', '<i4'), ('b', 'V100')]
```

**zarr.json:**
```
{
  "shape": [
    10
  ],
  "data_type": {
    "name": "structured",
    "configuration": {
      "fields": [
        [
          "a",
          "int32"
        ],
        [
          "b",
          {
            "name": "raw_bytes",
            "configuration": {
              "length_bytes": 100
            }
          }
        ]
      ]
    }
  },
  "chunk_grid": {
    "name": "regular",
    "configuration": {
      "chunk_shape": [
        10
      ]
    }
  },
  "chunk_key_encoding": {
    "name": "default",
    "configuration": {
      "separator": "/"
    }
  },
  "fill_value": "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=",
  "codecs": [
    {
      "name": "bytes"
    },
    {
      "name": "zstd",
      "configuration": {
        "level": 0,
        "checksum": false
      }
    }
  ],
  "attributes": {},
  "zarr_format": 3,
  "node_type": "array",
  "storage_transformers": []
}
```

### Steps to reproduce

```python
# /// script
# requires-python = ">=3.11"
# dependencies = [
# ]
# ///
#

import zarr
import numpy as np
# your reproducer code
# zarr.print_debug_info()

DTYPE = np.dtype([('a', 'i4'), ('b', 'f4', (5,5))])

store = LocalStore('bug.zarr')
arr = zarr.create_array(store, name='test', shape=(10,), dtype=DTYPE, fill_value=bytes(DTYPE.itemsize))

print('Original dtype:', DTYPE)
print('Array created with dtype:', arr.dtype)
print('Accessed item dtype: ', arr[0].dtype)
```

### Additional output

_No response_

sehoffmann:
@d-v-b 

This PR adds experimental support for subarray dtypes (https://numpy.org/doc/stable/glossary.html#term-subarray-data-type, https://numpy.org/doc/stable/user/basics.rec.html#structured-datatype-creation) and closes #3582 and #3583.

It also fixes support for nested  (and subarray-containing) `Structured` dtypes for Zarr v2 which worked before in 2.18.* but not anymore 3.1.*. In particular, the buggy implementation forgot that a nested structured dtype is again a list of lists and not just a single flat list.

**Note 1:**
Subarray dtypes are in a very weird spot. They are a proper `np.dtype`, particular a `np.VoidDType` with unset `fields` attribute but set `subdtype` field. Hence, it makes sense to map them one-to-one to a `ZDType`. This also makes sense from an implementation standpoint wrt. serialization.

On the other hand, they do not have a proper scalar value. I.e. one can not create a `np.void` scalar for a subarray dtype (throws). Conceptually, a scalar value of a subarray dtype would be a `np.ndarray`. This, however, is not a subtype of `np.generic` despite sharing a lot of the interface. When one creates a np.ndarray with a subarray dtype directly, the result is "flat" `np.ndarray` with shape `array_shape + subarray_shape`.

I've decided to still implement them as separate `Subarray`-ZDType and not conflate them within the `Structured` class. While this works flawlessly when used within a structured dtype, the intended use case, using them directly is not fully supported. Specifically, there is no specification for standalone subarray dtypes in Zarr V2, making a lot of test cases fail. Apart from that, some tests in test_array.py do not expect an array as scalar and hence fail. I want to stress though, that I was able to successfully create and read a Subarray zarr array with V3.

Solving this conundrum adequately  is beyond my possibilities and might require significant conceptual changes in Zarr. I did not add the dtype directly to `test_dtype/contest.py` but instead added a new test case for `Structured` that uses a Subarray inside which passes.

**Note 2:** I've also added a test case for an invalid float value string which fails due to #3584. Since that test case highlights an existing bug, I've decided to leave it there.

TODO:


sehoffmann:
@d-v-b Don't want to be pushy here, but did you manage to have a look at this PR yet? Do you have any feedback or is there anything that needs to be changed or addressed?

d-v-b:
> @d-v-b Don't want to be pushy here, but did you manage to have a look at this PR yet? Do you have any feedback or is there anything that needs to be changed or addressed?

Hi @sehoffmann sorry for the long silence. I think there are 2 distinct elements in this PR: first is improving how we handle numpy structured dtypes, and the second is including sub-array data types. 

The first element looks great, but I have some concerns about the second element. So far we have tried to keep the set of supported data types as close as possible to the union of the data types zarr python v2 supported, plus the data types supported by other zarr v3 implementations (namely, zarrs and tensorstore). 

This means when we add a new data type, there are two questions to answer: is this dtype something people used in zarr python 2, (and if does adding it resolve a feature regression)? or, is this dtype something the other zarr v3 implementations are supporting? If the answer to both of those is "no", then it seems like the maintenance burden for zarr-python might not be worth it, compared to the alternative of users registering this data type themselves via the registry. And I think sub-arrays are not something people used heavily in zarr python 2.x, nor are they supported by other zarr v3 implementations (please correct me if I'm wrong on either of these points).

How important is it for your application that this data type is bundled with Zarr python? And if that outcome is very important, would you be willing to work on a data type spec in the [zarr-extensions repo](https://github.com/zarr-developers/zarr-extensions/)? I think I'd support adding the new subarray data type unreservedly if there was buy-in from other zarr implementers. Without that buy-in, I'm pretty skeptical about the addition, and I would encourage using the data type registry to register the data type instead of relying on it being shipped wit zarr-python.

these are just my thoughts though, it would be good to hear from the other devs @zarr-developers/python-core-devs

vitusbenson:
Following the implementation of the zarr-extension for structured dtypes, I've rebased this branch to the latest main and kept only the changes related to subarrays inside structured dtypes.

See https://github.com/sehoffmann/zarr-python/pull/1
And https://github.com/vitusbenson/zarr-python/tree/subarray_dtypes

sehoffmann:
Hey @d-v-b,

just to follow up on this. This PR would be ready to merge if it would add support for subarray dtypes as part of `Structured`(not standalone), following the spec from https://github.com/zarr-developers/zarr-extensions/pull/45 ?

d-v-b:
hi @sehoffmann we are still missing a spec for an "array of scalars" dtype. This should probably be a generic fixed-length data type that's configured with its length and the type of its contents. Once we have a spec for that, then it composes with the new struct dtype.

sehoffmann:
Hey @d-v-b,

Just supporting subarrays as part of `Structured` would also be completely fine from my side. I see this as a matter of taste. Given that standalone subarray dtypes are behaving rather unexpected and weird compared with other dtypes, there might actually be a good reason to not allow standalone subarrays.

I.e. `np.dtype([('x', 'f4'), ('y', np.float32), ('z', 'f4', (2, 2))])` should still be supported by zarr as part of `Structured` (as already in v2),  but it is completely ok for me if `np.dtype(('f4', (2,2)))` doesn't have a corresponding zarr datatype. Maybe this was a misunderstanding before. 

-------------

If that is fine with you, I could proceed by adapting this PR such that subarrays are only supported as part of `Structured`.

d-v-b:
hi @sehoffmann, we now have a spec for a [language-agnostic struct dtype](https://github.com/zarr-developers/zarr-extensions/tree/main/data-types/struct), and this was added to zarr-python [recently](https://github.com/zarr-developers/zarr-python/pull/3781). The spec defines the struct dtype as generic over other dtypes. 

So if we want the struct dtype to support subarrays in a way that could work for implementations outside of zarr-python, the best way forward would be a subarray dtype spec in zarr-extensions, then the struct dtype would "just work" without any changes.

I opened an issue about this in zarr-extensions: https://github.com/zarr-developers/zarr-extensions/issues/57. I don't think the spec for a subarray dtype would be too much work (but I don't have time for it at the moment).

sehoffmann:
I see, thanks for the clarification. How do you want to handle the backward compatibility to zarr v2? If I am not mistaken, subarrays were an integral part of `Structured` there and not standalone. Is this an issue and does the new subarray dtype need to concern itself with this?

PR #3593, opened 2025-11-21
fix: ensure `ZipStore` is open before acquiring lock

oelhammouchi:
Fix a bug where `ZipStore` complains that it has no attribute `_lock` when some methods are called. Any method using the lock should make sure the store is opened because that's when the `_lock` attribute gets set.


d-v-b:
how did the bug you are fixing here get past the store tests? Maybe we need to make those tests stronger?

oelhammouchi:
You're right, I didn't want to appear presumptuous but the code could do with some refactoring as well, having to repeat all those `_sync_open` calls in so many functions naturally increases the probability of forgetting some. Not sure if this should be done in this PR though.

d-v-b:
if you can find a clean way to reduce the repetition (e.g., a decorator) we'd appreciate it!

oelhammouchi:
@d-v-b just for info, I'm currently on holiday but I'm planning to implement the proposed improvements shortly

oelhammouchi:
@d-v-b I've finally had time to revisit this! I've added tests and cleanup up the open check, the proposed decorator approach turned out not very clean imo because you'd need to test for the various types of functions (async, iterator, etc.)

d-v-b:
looks good @oelhammouchi, would you mind writing a [release note](https://zarr.readthedocs.io/en/stable/contributing/#changelog) for these changes?

d-v-b (approved):

PR #3605, opened 2025-11-27
fix tests for stand-alone example scripts

d-v-b:
Since changing the layout of our standalone examples, we failed to update the tests, and so the examples are not currently getting tested. See https://github.com/zarr-developers/zarr-python/actions/runs/19712165663/job/56475393603#step:6:100 for a recent run where `test_examples.py` didn't collect any tests. 

This PR makes it a test failure if no example scripts are found, and corrects our example script discovery to find example scripts properly at test time.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -75,7 +75,7 @@ test = [
     "coverage>=7.10",
     # Pin possibly due to https://github.com/pytest-dev/pytest-cov/issues/693
     "pytest<8.4",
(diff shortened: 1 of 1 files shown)


d-v-b:
the min-deps test failure is caused by pytest-asyncio requiring a newer version of typing-extensions than our minimum supported version. That's fixed in https://github.com/zarr-developers/zarr-python/pull/3606

PR #3612, opened 2025-12-03
Adding `copy_store` convenience method for `Group`

melonora:
[Description of PR]
This PR adds a convenience method for copying the store of a `Group` to a destination store.

TODO:

@d-v-b


melonora:
@d-v-b I know right now this won't work when zarr v2 is used. I can add that functionality as well or we say the convenience function is not supported for zarr v2.

melonora:
added zarr v2 support. If in general the code is approved, I will add documentation.

src/zarr/core/group.py
```diff
@@ -697,6 +697,73 @@ def from_dict(
...
+                )
+
+                if target_zarr_format == 3:
+                    for region in member._iter_shard_regions():
+                        data = await member.getitem(selection=region)
+                        await new_array.setitem(selection=region, value=data)
+                else:
+                    for region in member._iter_chunk_regions():
+                        data = await member.getitem(selection=region)
+                        await new_array.setitem(selection=region, value=data)
```
d-v-b: iter_shard_regions [works for unsharded arrays as well](https://github.com/zarr-developers/zarr-python/blob/ee0e69a74898b9e0467df517fc55a337bbb5fb06/src/zarr/core/array.py#L5456-L5459) (it falls back to the chunk size), so I don't think we need this conditional

src/zarr/core/group.py
```diff
@@ -697,6 +697,73 @@ def from_dict(
             store_path=store_path,
         )
 
+    async def copy_store(
```
d-v-b: we should think carefully about the name -- technically this method doesn't copy the store, it only copies the arrays / groups contained under this group, so I'm not sure "copy_store" is the best description. Plain "copy" is IMO overloaded with making an in-memory copy of the group object. Alternatives: "copy_to", "resave", @zarr-developers/python-core-devs any other ideas?

src/zarr/core/group.py
```diff
@@ -697,6 +697,73 @@ def from_dict(
...
+        self,
+        store: StoreLike,
+        *,
+        overwrite: bool = False,
+        consolidate_metadata: bool | None = None,
```
d-v-b: why do we need this here? IMO it's simpler if copying is just copying. If people want to add consolidated metadata, they should do that after copying.

src/zarr/core/group.py
normanrz: I like copy_to.

src/zarr/core/group.py
melonora: ok, then there would be no reason to have `_iter_chunk_regions` and it could be deprecated straight away since it is private. I can do that in a follow up PR.

src/zarr/core/group.py
melonora: same, also with function argument it reads naturally.

src/zarr/core/group.py
```diff
@@ -697,6 +697,69 @@ def from_dict(
...
+                    **kwargs,
+                )
+
+                for region in member._iter_shard_regions():
+                    data = await member.getitem(selection=region)
+                    await new_array.setitem(selection=region, value=data)
+
+        group = await self.open(self.store, zarr_format=target_zarr_format)
+        if group.metadata.consolidated_metadata:
+            await async_api.consolidate_metadata(new_group.store)
```
d-v-b: if the original group had stale consolidated metadata, or if the new group is concurrently modified by another process, then consolidating it here will generate different consolidated metadata from the original. So IMO we should not consolidate and instead just copy the original consolidated metadata.

src/zarr/core/group.py
melonora: adjusted it. It now consolidates if the source store was consolidated, but this does not require an argument. One more thing, I only take into account when consolidation happened at the root level. It could be (although the user should not do this) that consolidation happened at the subgroup level. This could be accounted for by checking ` if member. metadata.consolidated_metadata` when looping through `members`. WDYT should I add this?

d-v-b: I would copy the original group and its children exactly, including their consolidated metadata.

melonora: What would be the purpose of copying over stale consolidated metadata?

d-v-b: it keeps the scope of this function simple -- it just copies the original group exactly. If this tool produces a different group, then it's not really copying.

melonora: So there are a few considerations here without a clear solution at the moment. I also discussed with @joshmoore. Ultimately, this work is out of spec so the way forward is not clear. I approached this initially more from the user perspective, where the users would care that their data and the metadata attributes are properly copied over, which I think with this implementation is the case. The rest I would count as store-specific. The problem here could be that this could change because of `consolidated_metadata` and particular stores like `ZipStore` being out of spec.

When it comes to copying the metadata exactly as is I am not entirely certain at the moment how to do that given that `metadata` is a `FrozenInstance`, other than to use `object.__setattr__`. Would you have suggestions?

melonora: other option that would make it more clear is than opt for name `resave`.

d-v-b: can you explain how this relates to the spec? the metadata documents are just JSON at the end of the day. To me, copying a group to a new store means _just_ copying the existing group, and copying it exactly. If people want to re-consolidate metadata after copying, then they can do that separately.

d-v-b: are you having a problem creating consolidated metadata directly (i.e., without doing IO)?

melonora: @joshmoore additional thoughts from your side?

melonora: > are you having a problem creating consolidated metadata directly (i.e., without doing IO)?

I think regarding that you could do something like `ConsolidatedMetadata.from_dict`, it is more setting it given that it is a `FrozenInstance`.

d-v-b: Is there some problem with passing `ConsolidatedMetadata` as a keyword argument to the Group metadata constructor? https://github.com/zarr-developers/zarr-python/blob/ee0e69a74898b9e0467df517fc55a337bbb5fb06/src/zarr/core/group.py#L356

d-v-b: (fwiw I think this is an area of our API that could be a lot better)

d-v-b: maybe this can be a lot simpler also on the chunk side if you think of "copying a group" as performing the following operation:
- identifying the names of all the objects to copy from the source store (the metadata documents for the subgroups / sub arrays, and the chunks)
- copying the bytes for each object from the source store to the destination store

Doing it this way avoids the need to decompress + recompress the chunk bytes, which should save a lot of time.

I'm still not sure how any of this relates to the spec though...

d-v-b: and we have routines for iterating over the names of the stored objects (chunks / shards): https://github.com/zarr-developers/zarr-python/blob/ee0e69a74898b9e0467df517fc55a337bbb5fb06/src/zarr/core/array.py#L1447

melonora: no problem there, it is more assigning it as metadata of the group. Right now, we would create by `Group.from_store` which passes on attributes, but not `ConsolidatedMetadata` or `GroupMetadata`. So far I did not find a way to achieve this with the API other than using `object.__setattribute` .

d-v-b: sounds like this is a deficiency in the `Group.from_store` method

joshmoore: tl;dr - I agree with @d-v-b's general point of making consolidated metadata a follow up 

Short version of the coffee conversation with @melonora:
 - The propagation of the consolidated metadata flag as a store concern feels quite similar to dimension separator before it was introduced into the array metadata.
 - In thinking through the API here, I could see having `source_store.copy_to(target_consolidated_metadata_aware_store)` "do the right thing" in that it follows the `super.copy_to` with generating the consolidated metadata. Perhaps `source_consolidated_metadata_aware_store.copy_to(target_consolidated_metadata_aware_store)` could even do something more custom/intelligent/performance including:
    * directly transferring the consolidated metadata file rather than regenerating
    * _only_ using the consolidated metadata to write the metadata
    * validating that the input consolidated metadata is correct
    
but I'm not sure these edge cases are well-specified enough to do something reliably correct and so would focus here on iterating through the objects which are in spec and representing them perfectly rather than adding this additional complexity to the method.

d-v-b: > The propagation of the consolidated metadata flag as a store concern feels quite similar to dimension separator before it was introduced into the array metadata.

Consolidated metadata is already in the group metadata document for zarr v3 if `"kind"` is set to `"inline"`. So if we thinking copying a group should preserve its metadata, then we are committed to copying consolidated metadata. 

I am not following the API suggestion -- the current version of this code uses the `members` method to iterate over child groups and arrays, which uses the consolidated metadata of the parent group to discover child groups and arrays.

melonora: regarding the `from_store` method, I would open a separate PR for this.

joshmoore: > Consolidated metadata is already in the group metadata document for zarr v3 if "kind" is set to "inline".

:+1: but doing something other (more or less) than just copying would require interpreting that which  is too far IMO.

d-v-b: FYI @joshmoore  the method under discussion here is a method on the `Group` class, not on stores. maybe this is a point of confusion.

d-v-b: IMO the only consolidated metadata-related parameter we need is whether iterating over the child groups uses their consolidated metadata (if present). See the relevant kwarg in the signature of [`members`.](https://github.com/zarr-developers/zarr-python/blob/ee0e69a74898b9e0467df517fc55a337bbb5fb06/src/zarr/core/group.py#L1353)

melonora: I added both, only thing remaining is probably adding `consolidated_metadata` in the async group call.

melonora:
@d-v-b @joshmoore I think I have now addressed all comments and I also added documentation, so I think this would be ready for a final round of review. @d-v-b thanks for your patience.

melonora:
@d-v-b @joshmoore , just bumping in case you have some time to see whether you approve the current state:)

src/zarr/core/group.py
```diff
@@ -697,6 +702,91 @@ def from_dict(
...
+            Note that this only affects groups loaded from the store. If the current Group already has
+            consolidated metadata, it will always be used.
+
+        Returns
+        -------
+        AsyncGroup
+            The new group in the target store.
+        """
+        target_zarr_format = self.metadata.zarr_format
+        group = await self.open(self.store, zarr_format=target_zarr_format)
```
d-v-b: isn't this the same as `self`?

src/zarr/core/group.py
```diff
@@ -697,6 +702,91 @@ def from_dict(
             store_path=store_path,
         )
 
+    async def copy_to(
+        self,
+        store: StoreLike,
+        *,
+        overwrite: bool = False,
+        use_consolidated_for_children: bool = True,
```
d-v-b, line 703: unlike [`create_group`](https://github.com/zarr-developers/zarr-python/blob/ee0e69a74898b9e0467df517fc55a337bbb5fb06/src/zarr/api/synchronous.py#L568-L576), this function doesn't take a `path` parameter, which means it's not possible to easily specify a location within a store for the group. until we make it easy to include this information in the store itself (#2272 ) we should probably take a `path` parameter.

src/zarr/core/group.py
```diff
@@ -1874,6 +1970,42 @@ def open(
         obj = sync(AsyncGroup.open(store, zarr_format=zarr_format))
         return cls(obj)
 
+    def copy_to(
+        self,
+        store: StoreLike,
+        *,
+        overwrite: bool = False,
+        use_consolidated_for_children: bool = True,
```
d-v-b, line 1918: we should add a test that checks the signature of this function against the signature of the async version. otherwise, they have a tendency to drift. Here's an example of such a test: https://github.com/zarr-developers/zarr-python/blob/ee0e69a74898b9e0467df517fc55a337bbb5fb06/tests/test_api/test_synchronous.py#L26-L41

src/zarr/core/group.py
melonora: turns out it is:)

melonora: at least I think you were refering to the `AsyncGroup.from_store` call?

d-v-b: > group = await self.open(self.store, zarr_format=target_zarr_format)

I was talking about this line

src/zarr/core/group.py
melonora: I am not certain yet, first the group is opened by calling `self.open(self.store, zarr_format=target_zarr_format)`. If we allow for `path` to be passed I interpret it as subgroup within existing zarr store to which to copy whatever store you pass. Am I following you correctly?
In other words are you saying we should allow to write a store to a subgroup destination?

d-v-b: > self.open(self.store, zarr_format=target_zarr_format)

I am still confused why this line is necessary. It will open an array or group at the root of the store, which is not what we want:

```python
>>> import zarr
>>> zg = zarr.create_group(store={}, path='group')  
>>> zg
<Group memory://133419005213760/group>
>>> zg.open(zg.store)
<Group memory://133419005213760>
```

> In other words are you saying we should allow to write a store to a subgroup destination?

this method does not write a store, it writes a group into a store. In normal group creation, you declare the store, and also the path within the store. See the signature of [`create_group`](https://github.com/zarr-developers/zarr-python/blob/ee0e69a74898b9e0467df517fc55a337bbb5fb06/src/zarr/api/asynchronous.py#L726). The question is whether `copy_to` should follow the same pattern, and take a `path` argument.

src/zarr/core/group.py
melonora: ah in that case no, but it is a good thing to point out as I forgot to ask. This is not the case when the metadata was consolidated just before, e.g. if you call `self.metadata.consolidated_metadata` it will return `None`. I had to specifically open the group in order to access the consolidated metadata.

Should this be considered a bug?

d-v-b: no it's not a bug, this is an expected consequence of how the zarr.Array / zarr.Group classes work, and its why `consolidate_metadata` returns a new `AsyncGroup` https://github.com/zarr-developers/zarr-python/blob/ee0e69a74898b9e0467df517fc55a337bbb5fb06/src/zarr/api/asynchronous.py#L188-L192

d-v-b: to keep `copy_to(self, ...)` simple, just copy `self` exactly. There's no need to check the store for a version of `self` with consolidated metadata. Users who want to copy a group with consolidated metadata to a new store should ensure that their group has consolidated metadata before invoking `copy_to`.

melonora: Thanks for clarifying, I now made it so that `consolidated_metadata` is accessed directly from `self` by also adjusting the test slightly.

src/zarr/core/group.py
melonora: so I did not do the actual docstrings itself, but I did write a test for all methods for classes and their async counterparts. I think this is generalizable to other classes as well, though currently only `Group` and `AsyncGroup` are tested. With this, already some mismatches are detected which for now are skipped but could be addressed in a follow up PR. Doing so would require deprecating some parameters. If you agree I can create an issue for it.

melonora: Please see `test_class_method_parameters_match`

src/zarr/core/group.py
melonora: will implement it also for the sake of symmetric API

melonora: On second look I think it is still a bit confusing, the signature for `create_group` in the `Group` class itself does not have `path`. However, `create_group` in the async and sync API does include this. I take it the sole reason for `create_group` in the classes is to create a subgroup so I still think implementing `path` would be good in that case to have it model more the module level functions rather than the instance methods.

d-v-b: > However, create_group in the async and sync API does include this.

IMO we should treat the async / sync api `create_group` as the reference here. We are long overdue a cleanup of these methods / functions

melonora: @d-v-b implemented, let me know what you think when you have time!

src/zarr/core/group.py
```diff
@@ -697,6 +702,99 @@ def from_dict(
             store_path=store_path,
         )
 
+    async def copy_to(
+        self,
+        store: StoreLike,
+        *,
+        path: str | None = None,
```
d-v-b, line 701: what does it mean when path is None? is it meaningfully different from defaulting to the empty string?

tests/test_group.py
```diff
@@ -63,6 +63,89 @@
...
 
+def is_deprecated(method):
+    """Check if a method is marked as deprecated."""
+    # Check for @deprecated decorator
+    return hasattr(method, "__deprecated__") or (
+        hasattr(method, "__wrapped__") and hasattr(method.__wrapped__, "__deprecated__")
+    )
+
+
+def get_method_names(cls):
```
d-v-b:
```suggestion
def get_method_names(cls) -> str:
```

tests/test_group.py
```diff
@@ -63,6 +63,89 @@
     from zarr.core.common import JSON, ZarrFormat
 
 
+def is_deprecated(method):
```
d-v-b:
```suggestion
def is_deprecated(method) -> bool:
```

tests/test_group.py
```diff
@@ -63,6 +63,89 @@
...
+def get_method_names(cls):
+    """Extract public method names from a class, excluding deprecated methods."""
+    return [
+        name
+        for name, method in inspect.getmembers(cls, predicate=inspect.isfunction)
+        if not name.startswith("_") and not is_deprecated(method)
+    ]
+
+
+def get_method_signature(cls, method_name: str):
```
d-v-b:
```suggestion
def get_method_signature(cls, method_name: str) -> dict[str, whatever]:
```

tests/test_group.py
```diff
@@ -250,6 +333,135 @@ def test_group_members(store: Store, zarr_format: ZarrFormat, consolidated_metad
...
+        dtype=arr_data.dtype,
+    )
+    src["dataset"] = arr_data
+
+    src = consolidate(src_store, src, consolidate_metadata, zarr_format)
+
+    return src, arr_data, subgroup_arr_data, zarr_format, consolidate_metadata
+
+
+def consolidate(src_store, src, consolidate_metadata, zarr_format):
```
d-v-b: please add type annotations

tests/test_group.py
```diff
@@ -250,6 +333,135 @@ def test_group_members(store: Store, zarr_format: ZarrFormat, consolidated_metad
...
+    "copy_to_test_data",
+    [
+        (2, None, False),
+        (2, None, True),
+        (3, (50,), False),
+        (3, (50,), True),
+    ],
+    indirect=True,
+)
+def test_copy_to(copy_to_test_data) -> None:
```
d-v-b:
```suggestion
def test_copy_to(copy_to_test_data: tuple[int, None | tuple[int, ...], bool]) -> None:
```

tests/test_group.py
```diff
@@ -250,6 +333,135 @@ def test_group_members(store: Store, zarr_format: ZarrFormat, consolidated_metad
...
+    "copy_to_test_data",
+    [
+        (2, None, False),
+        (2, None, True),
+        (3, (50,), False),
+        (3, (50,), True),
+    ],
+    indirect=True,
+)
+def test_copy_to_with_path(copy_to_test_data) -> None:
```
d-v-b:
```suggestion
def test_copy_to_with_path(copy_to_test_data: tuple[int, None | tuple[int, ...], bool]) -> None:
```

src/zarr/core/group.py
melonora: It is not, but I am following what seems to be the convention in this codebase. Although I did spot some heterogeneity in that. However, I think perhaps it is better if we want to change this to `path: str = ""`, to do this in a follow up PR. It makes sense to me to implement the change but it is not the scope of this PR I think.

tests/test_group.py
melonora: will do so now, just wanted to know first if in general you approve the code:)

melonora:
@d-v-b When the last test passes and if you are happy with the code changes I think this is good to go. Thanks for reviewing!

d-v-b (approved):

d-v-b:
@melonora I just realized that this implementation is decoding and re-encoding array chunks. Was there a particular reason for that choice? Because it would be far faster to copy the bytes directly

melonora:
Hmm, mostly a lack of familiarity with getting the raw bytes, but looking at the documentation, I should be able to do something like `store.get` to get them, correct?
I will change it

d-v-b:
you don't have to! I already did in a local branch. If it's OK with you, I would push those changes here.

melonora:
Ah ok, thanks! Yes I am ok with it.

melonora:
@d-v-b is this still of interest? I could resolve the conflicts or otherwise close it.

d-v-b:
it is, sorry for the delay in dealing with things here. lmk if you run into any trouble with the merge conflicts

melonora:
will do, thanks!

melonora:
@d-v-b if you are ok with the current state, it can be merged

d-v-b:
@melonora I'm going to give this a careful look today. I might identify some things in this PR that would be better off implemented in separate parts of the zarr codebase, do you mind if I push to your branch? some changes would come in that way, others might be spun off into separate PRs

PR #3613, opened 2025-12-03
Arrow ipc Array -> Bytes codec

rabernat:
Implementation of `arrow-ipc` Array Bytes codec proposed in https://github.com/zarr-developers/zarr-extensions/pull/41

TODO:


src/zarr/core/metadata/v3.py
```diff
@@ -105,7 +105,10 @@ def validate_codecs(codecs: tuple[Codec, ...], dtype: ZDType[TBaseDType, TBaseSc
     # TODO: use codec ID instead of class name
     codec_class_name = abc.__class__.__name__
     # TODO: Fix typing here
-    if isinstance(dtype, VariableLengthUTF8) and not codec_class_name == "VLenUTF8Codec":  # type: ignore[unreachable]
+    if isinstance(dtype, VariableLengthUTF8) and codec_class_name not in (
+        "VLenUTF8Codec",
+        "ArrowIPCCodec",
+    ):  # type: ignore[unreachable]
```
rabernat: This change allows us to use either the vlen-bytes or the arrow-ipc codec to encode variable length strings.

Flagging that this sort of logic for mapping codec / dtype compatibility feels quite brittle and non-scalable. But I don't have a better proposal in mind.

rabernat:
@d-v-b - resolving the typing errors here is beyond my ability. Would appreciate your help. 🙏

src/zarr/core/metadata/v3.py
d-v-b: i feel the same way! we might need a dtype x codecs compatibility matrix, not sure if it should track compatibility or incompatibility

d-v-b:
ci is passing via https://github.com/zarr-developers/zarr-python/pull/3613/commits/96273a841e8d3a6cbef3bb6d7b607360da4a2cb9

src/zarr/codecs/arrow.py
```diff
@@ -0,0 +1,78 @@
...
+    async def _decode_single(
+        self,
+        chunk_bytes: Buffer,
+        chunk_spec: ArraySpec,
+    ) -> NDBuffer:
+        record_batch_reader = read_ipc_stream(io.BytesIO(chunk_bytes.as_buffer_like()))
+        # Note: we only expect a single batch per chunk
+        record_batch = record_batch_reader.read_next_batch()
+        array = record_batch.column(self.column_name)
+        numpy_array = array.to_numpy()
```
ilan-gold, line 53: Very happy to see this happening :)

I would be very curious about the behavior of non-standard types here.  What does something like `geometry` dtype (which isn't in `pyarrow`) or `DictionaryArray` (which is in the core but has an implicit masking of sorts) do here? I can't deduce from the pyarrow docs exactly to be honest

Would it make sense to have a custom buffer class similar to what @keewis is doing for sparse (I think?)

PR #3614, opened 2025-12-04
Provide chunk_grid_space fallback in `_iter_shard_coords`

melonora:
In the codebase `_iter_shard_regions` was already falling back to chunks in case there are no shards. Another method, `iter_shard_coords` did not have this fallback option. This PR adds a fallback uption for that method in case there is no shard grid.

TODO:

@d-v-b @joshmoore

### DIFF
--- a/src/zarr/core/array.py
+++ b/src/zarr/core/array.py
@@ -1382,6 +1382,36 @@ class AsyncArray(Generic[T_ArrayMetadata]):
     async def nbytes_stored(self) -> int:
         return await self.store_path.store.getsize_prefix(self.store_path.path)
 
+    def _iter_chunk_coords(
+        self, *, origin: Sequence[int] | None = None, selection_shape: Sequence[int] | None = None
+    ) -> Iterator[tuple[int, ...]]:
+        """
+        Create an iterator over the coordinates of chunks in chunk grid space.
+
+        If the `origin` keyword is used, iteration will start at the chunk index specified by `origin`.
+        The default behavior is to start at the origin of the grid coordinate space.
+        If the `selection_shape` keyword is used, iteration will be bounded over a contiguous region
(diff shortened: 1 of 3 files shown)


d-v-b:
we need separate methods for iterating over chunks and shards because they are not the same

```python
# /// script
# requires-python = ">=3.11"
# dependencies = [
# ]
# ///

import zarr

array = zarr.create_array({} ,shape=(10,), dtype='i4', chunks=(5,), shards=(10,), fill_value=0)

print(tuple(array._iter_shard_regions()))
# ((slice(0, 10, 1),),)
print(tuple(array._iter_chunk_regions()))
# ((slice(0, 5, 1),), (slice(5, 10, 1),))
```

melonora:
I see that, but do not see where this is used internally that it matters. Otherwise I would just provide the same fallback option for `_iter_shard_coords` and leave the original `_iter_chunk` intact.

d-v-b:
we do not use these methods internally _yet_, but that's not a good reason for removing them.

also see https://github.com/zarr-developers/zarr-python/pull/3573

melonora:
> we do not use these methods internally _yet_, but that's not a good reason for removing them.
> 
> also see #3573

Then I will revert them, but provide just the same fallback option for `_iter_shard_coords'

d-v-b:
I'm confused by these changes -- is there a bug that this fixes? Because we should already fall back to the chunk grid for unsharded v2 arrays, via the [`_shard_grid_shape` property](https://github.com/zarr-developers/zarr-python/blob/ee0e69a74898b9e0467df517fc55a337bbb5fb06/src/zarr/core/array.py#L1255-L1268)

melonora:
Thanks @d-v-b for your patience and pointing it out. Was mainly based on spotting it based on the other method.

PR #3619, opened 2025-12-11
remove pytest pin

d-v-b:
removes the pin on the upper bound of pytest, as the motivating reason for that pin seems to be resolved.


d-v-b:
pre-commit failure is unrelated. and we should look at removing the pytest-asyncio pin

PR #3623, opened 2025-12-16
Some minor linter \ type-hint fixes for my own sanity

xaviernogueira:
I noticed there was some duplication of the `Literal[2,3]` typehint, when it was defined at `core.common` as `ZarrFormat`. Additionally, `ZarrFormat` is confusingly also used to refer to the CLI `StrEnum` containing `"v2"` and `"v3"`.

This fixes that. Originally I was going to move to `Literal[2] | Literal[3]` but `ruff` disagrees (which IMO I don't like but whatever I guess that is what opinionated linting is).

TODO:
* [x] Add unit tests and/or doctests in docstrings -> N/A
* [x] Add docstrings and API docs for any new/modified user-facing classes and functions -> N/A
* [x] New/modified features documented in `docs/user-guide/*.md` -> N/A

### DIFF
--- a/changes/3623.misc.md
+++ b/changes/3623.misc.md
@@ -1,0 +1,5 @@
+This PR contains minor, non-function-altering, changes to use `ZarrFormat` across the repo as opposed to duplicating is with `Literal[2,3]`.
+
+Additionally, it fixes broken linting by using a `Literal[True, False]` type hint for Numpy hypothesis testing, as opposed to `bool`.
+
+Basically improves the typehints and reduces fat-finger error surface area slightly.


d-v-b:
thanks! kind of an aside but i'm not sure we need string enums as long as literal string types exist. we should consider removing literal string enums from our codebase

xaviernogueira:
@d-v-b from what I see the `StrEnum` is used for the CLI, which may provide better CLI `--help` hints? Not sure would have to check. Or maybe it was deemed more user friendly to pass `--zarr-format v3` as opposed to `--zarr-format 3`?

Not sure what the story is there but I am inclined to agree with you. Thanks for the rapid review/approval!

Issue #3627, opened 2025-12-18
zero check in encoding less than ideal for lazy arrays


hmaarrfk:
The zero check in the encoding path is less than ideal
https://github.com/zarr-developers/zarr-python/blob/b8736917c351418dce3fb425ee16f747136da96e/src/zarr/core/codec_pipeline.py#L416

For lazy arrays, it eventually calls:

1. np.broadcast_arrays
2. np.array_equal

where the latter can instantiate a full array in memory.

The problem is that some lazy array implementation will not "keep this instantiation of the interediate full memory array" in memory for the the operation 2 lines down, that will eventually write the array.

So if the instantiation of the lazy array is costly, we are:

1. Instantiating it once for checking for zeros.
2. Instantiating a second time to write it to the store.

I would love to avoid this 'double cost' in a "pythonic way".

I would have loved to suggest 
~My proposal would be~ to refactor the "zero check" to use `np.array_equiv` https://numpy.org/devdocs/reference/generated/numpy.array_equiv.html#numpy.array_equiv

```python
>>> import numpy as np
>>> np.array_equiv(np.zeros((3, 4)), 0)
True
```

but looking at the source, it seems to recreate the same problem https://github.com/numpy/numpy/blob/main/numpy/_core/numeric.py#L2552-L2598

I'm not sure what the right solution is.

One could cache the local array.

I notice that this operation is actually considered inefficient by the original writers leaving a note like
```
        # use array_equal to obtain equal_nan=True functionality
        # Since fill-value is a scalar, isn't there a faster path than allocating a new array for fill value
        # every single time we have to write data?
```

---------------------------------------------------------------------------

Where this is a problem for me:

The numpy broadcast operation is actually really difficult to implement if you want to ensure high performance based on the chunks of the array. My experience with dask from 7 years ago reminds me that it was also difficult for them to implement efficient rechunking.

So implementing the `np.broadcast_arrays` just feels like "alot of work"....

https://github.com/zarr-developers/zarr-python/blob/b8736917c351418dce3fb425ee16f747136da96e/src/zarr/core/buffer/core.py#L543

The zero check feels like it makes sense.... but its like "alot of work to do" for some array backends.

----------------------------------------------------------------------------

Suggested patch:
```
diff --git a/src/zarr/core/buffer/core.py b/src/zarr/core/buffer/core.py
index f0d01566..54a2c09f 100644
--- a/src/zarr/core/buffer/core.py
[...]

PR #3628, opened 2025-12-18
Cache the result of the zero-check for array writing

hmaarrfk:
Closes https://github.com/zarr-developers/zarr-python/issues/3627

I ended up moving around my suggestion so that it could be more targetted to the one usecase I care about.

LMK what you think.

[Description of PR]

TODO:

### DIFF
--- a/src/zarr/core/codec_pipeline.py
+++ b/src/zarr/core/codec_pipeline.py
@@ -426,7 +426,11 @@ class BatchedCodecPipeline(CodecPipeline):
                         # If the result is not useful, we leave it for the garbage collector.
                         # We optimize this operation for the case that the GPU
                         if not hasattr(chunk_array._data, '__cuda_array_interface__'):
-                            chunk_array = NDBuffer(np.asarray(chunk_array._data))
+                            # I'm not sure why this implementation doesn't work
+                            # it seems like something is getting missed by me
+                            # chunk_array = NDBuffer(np.asarray(chunk_array._data))
+                            # This line here just feels more dirty
+                            chunk_array._data = np.asarray(chunk_array._data)
 
                         if chunk_array.all_equal(
                             fill_value_or_default(chunk_spec)


hmaarrfk:
Ok you are all sophisticated enough to catch my hack solution ^_^:

```
>           warnings.warn(
                msg,
                category=ZarrUserWarning,
                stacklevel=2,
            )
E           zarr.errors.ZarrUserWarning: Creating a zarr.buffer.gpu.Buffer with an array that does not support the __cuda_array_interface__ for zero-copy transfers, falling back to slow copy based path
```

hmaarrfk:
not sure i hvae time to troubleshoot this or to think of a worthy high performance solution.

d-v-b:
there's no way `.as_numpy_array()` isn't going to break the cpu / gpu abstraction, right?

d-v-b:
if there's some operation `X` that we need to do on cpu-backed data AND gpu-backed data, then we should look at defining `X` in the ndbuffer base class. Here `X` is "compare all elements in an NDBuffer against some scalar `s`", insofar as this is a basic part of zarr (checking if a chunk is `fill_value`) I think baking it into a core abc makes sense.

hmaarrfk:
> there's no way .as_numpy_array() isn't going to break the cpu / gpu abstraction, right?

That is my understanding, you have to start to call `np.asanyarray` which would allow lazy arrays to go through.

I don't see a way through this without help in defining "likely_not_equal"
https://github.com/data-apis/array-api/issues/985

thoughts there appreciated^^^^

If i have an MRI image, do I really need to tell you that it is exactly zero? you probably just care that it is "very likely non-zero" in this usecase, but the strict equality checks makes it hard to take any shortcuts.

d-v-b:
for zarr, we really do need to check every single element in a chunk before concluding that all elements in the chunk are equal to the fill value. but if the performance of this check is a serious concern, then at least when reading data we could check if the compressed chunk is byte-wise identical to a compress fill_value chunk

hmaarrfk:
> we really do need to check every single element in a chunk before concluding that all elements in the chunk are equal to the fill value

For the codepath that I describe in https://github.com/zarr-developers/zarr-python/issues/3627 really???? zarr will crash if a chunk is identically 0 and saved?

hmaarrfk:
I haven't really used zarr in depth before (my fork was like 6 years out of date), but doing this change, and running the tests it seems that only a few fail, and they seem to be checking that you discard zero checks.

I don't disagree that it is "nice" to ensure that things are zero, but not strictly required.

In many cases, i would be ok with compressing some extra zeros.

Your idea of "checking that the compressed data" matches that of "compressed zeros" is fine, but again, not "strictly necessary" and adds a level of complexity that might not be required in all cases.

```diff
diff --git a/src/zarr/core/codec_pipeline.py b/src/zarr/core/codec_pipeline.py
index fd557ac4..18b96911 100644
--- a/src/zarr/core/codec_pipeline.py
+++ b/src/zarr/core/codec_pipeline.py
@@ -413,12 +413,12 @@ class BatchedCodecPipeline(CodecPipeline):
                 if chunk_array is None:
                     chunk_array_batch.append(None)  # type: ignore[unreachable]
                 else:
-                    if not chunk_spec.config.write_empty_chunks and chunk_array.all_equal(
-                        fill_value_or_default(chunk_spec)
-                    ):
-                        chunk_array_batch.append(None)
-                    else:
-                        chunk_array_batch.append(chunk_array)
+                    # if not chunk_spec.config.write_empty_chunks and chunk_array.all_equal(
+                    #     fill_value_or_default(chunk_spec)
+                    # ):
+                    #     chunk_array_batch.append(None)
+                    # else:
+                    chunk_array_batch.append(chunk_array)
 
             chunk_bytes_batch = await self.encode_batch(
                 [
```


```
FAILED tests/test_array.py::test_write_empty_chunks_behavior[zarr2-0-False-memory] - assert 2 == 0
FAILED tests/test_array.py::test_write_empty_chunks_behavior[zarr2-5-False-memory] - assert 2 == 0
FAILED tests/test_array.py::test_write_empty_chunks_behavior[zarr3-0-False-memory] - assert 2 == 0
FAILED tests/test_array.py::test_write_empty_chunks_behavior[zarr3-5-False-memory] - assert 2 == 0
FAILED tests/test_codecs/test_sharding.py::test_delete_empty_shards[local] - assert <zarr.core.buffer.cpu.Buffer object at 0x76b090f2a660> is None
FAILED tests/test_codecs/test_sharding.py::test_delete_empty_shards[memory] - assert <zarr.core.buffer.cpu.Buffer object at 0x76b090e31100> is None
FAILED tests/test_examples.py::test_scripts_can_run[script_path0] - FileNotFoundError: [Errno 2] No such file or directory: 'uv'
FAILED tests/test_store/test_zip.py::TestZipStore::test_api_integration - UserWarning: Duplicate name: 'foo/c/0/0'
```

d-v-b:
the reason why we avoid saving chunks that are identical to the fill value is to avoid unnecessary IO operations and reduce the number of stored objects. this only works for chunks that are identical to the fill value because the zarr spec instructs implementations to treat missing chunks as equal to chunks filled with `fill_value`. So "very likely all `fill_value`" and "definitely all `fill_value`" are different enough in this case that we check for it. But is this PR about the fill value logic?

d-v-b:
aha didn't see that this was closed, I accidentally re-opened it. feel free to re-close it!

hmaarrfk:
> But is this PR about the fill value logic?

Exactly, and I don't know how to resolve the GPU usecase.

Today, i'm purely exploring what it would mean to convert some of my datasets from HDF5 to zarr.

Even if it is slow.... the conversion will complete....

hmaarrfk:
> So "very likely all fill_value" and "definitely all fill_value" are different enough in this case that we check for it.

I understand and believe that in practice.

But do consider the flip side:

> I can probably tell you that my images are 99.99999% not all 0. due to some noise that some scientist believe is important. is that good enough for you to skip the "bit for bit equality check?"?

> For data that is "missing" i can probably tell you also that it is surely all 0s.

I think the semantics you are looking for just aren't there for the "numpy" API.

What you are trying to do is to "optimize away one chunk of data on disk", but perhaps, I want to save 100ms. Let me skip the 0 check? or give you a quick answer?

d-v-b:
> Let me skip the 0 check? or give you a quick answer?

you can skip this check by changing the `write_empty_chunks` value in the config. see https://github.com/zarr-developers/zarr-python/blob/b8736917c351418dce3fb425ee16f747136da96e/docs/user-guide/config.md?plain=1#L32

d-v-b:
historically checking if a chunk is empty or not has not been a serious performance bottleneck compared to reading, decompression, recompression, and writing. do you have some performance metrics from your use case that suggest otherwise?

hmaarrfk:
As i described in https://github.com/zarr-developers/zarr-python/issues/3627
```
For lazy arrays, it eventually calls:

np.broadcast_arrays
np.array_equal
```

This is a problem if:

1. Reading is slow -- think really high compression algorithm. Not zstd, but lzma for example.
2. your underlying `array` implementation doesn't support these well

each might trigger a terribly slow read, on top of the read that is triggered when reading the data the final time to compress it.

hmaarrfk:
Leaving this example here since it helped me find https://mail.python.org/archives/list/numpy-discussion@python.org/thread/YY2OXO4SGCSXYSAYVOVWTLU25PCUQJT7/

Ultimately, this "zero" check just create memory that is on the order of the entire zarr store.

It may be "creating it and deleting it very fast", but its still pretty "underwhelming".

When going from "store -> store", there needs to be a faster mechanism to shortcut this
<details><summary>memory benchmark script</summary>

```python
import numpy as np
import psutil
import time
import multiprocessing
from multiprocessing import Queue

def monitor_memory(pid, queue, interval=0.01):
    proc = psutil.Process(pid)
    peak_rss = 0
    samples = []

    while True:
        try:
            rss = proc.memory_info().rss / 1024**2
            samples.append((time.time(), rss))
            if rss > peak_rss:
                peak_rss = rss
            time.sleep(interval)
        except psutil.NoSuchProcess:
            break

    queue.put(('done', peak_rss, samples))

def run_benchmark():
    a = np.full((2, 1024, 1024, 1024), fill_value=0, dtype='uint8')
    b = np.broadcast_to(
        np.full(1, fill_value=1, dtype='uint8'),
        a.shape
    )
    assert all(s == 0 for s in b.strides)
    np.array_equal(a, b)

if __name__ == '__main__':
    queue = Queue()
    child_process = multiprocessing.Process(target=run_benchmark)
    child_process.start()

    monitor_process = multiprocessing.Process(
        target=monitor_memory,
        args=(child_process.pid, queue)
    )
    monitor_process.start()

    child_process.join()
    time.sleep(0.2)
    monitor_process.terminate()
    monitor_process.join()

    status, peak_rss, samples = queue.get()

    print("\n=== Memory Monitoring Results ===")
    print(f"Peak RSS: {peak_rss:.1f} MB")
    print(f"Total samples: {len(samples)}")

    N_samples = 30
    if samples:
        print(f"\nMemory timeline (first {N_samples} and last {N_samples} samples):")
        for i, (timestamp, rss) in enumerate(samples[:N_samples]):
            print(f"  {timestamp:.2f}s: {rss:.1f} MB")
        if len(samples) > N_samples * 2:
            print("  ...")
        for i, (timestamp, rss) in enumerate(samples[-N_samples:], start=len(samples)-N_samples):
            print(f"  {timestamp:.2f}s: {rss:.1f} MB")

```
</details>


I have a few next steps:
1. Learn more about your chunk_spec,
2. See if I can engineer it.
3. Otherwise, try the idea of "comparing he encoded array to zero"

I think that an optimization for array_equal will go a long way..... and hopefully it has positive reception at numpy upstream
The problem for me relies on the fact that

hmaarrfk:
I learned about `__cuda_array_interface__` 

it is a specialization, but i think it stays in line with the spirit of zarr.

i hope this works for all.

PR #3633, opened 2025-12-25
ci: align pre-commit Python with tooling config

### LINKED ISSUE #3624: pre-commit is under-determined
in #3619 I observed that the pre-commit check that runs in CI was flagging different problems from my local pre-commit run. This is no fun for contributors. I suspect we are leaving some versions of python + deps open, when we should instead declare them explicitly.

ramanbansal1:
In #3619 and #3624, CI pre-commit checks were observed to flag different issues than local runs due to implicit Python and dependency resolution. This change removes that under-determination by:
- Explicitly pinning the Python version used by pre-commit hooks (3.11)
Fixes #3624
This ensures consistent behavior between local development and CI, reducing contributor friction.

TODO:

### DIFF
--- a/.github/workflows/check_changelogs.yml
+++ b/.github/workflows/check_changelogs.yml
@@ -12,7 +12,7 @@ jobs:
       - uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
 
       - name: Install uv
-        uses: astral-sh/setup-uv@1e862dfacbd1d6d858c55d9b792c756523627244 # v7.1.4
+        uses: astral-sh/setup-uv@681c641aba71e4a1c380be3ab5e12ad51f415867 # v7.1.6
 
       - name: Check changelog entries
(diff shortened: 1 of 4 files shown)


d-v-b:
@ramanbansal1 thanks for this! One thing I am wondering -- could we achieve the same outcome, but in a more generic way, by adding a [`.python-version`](https://docs.astral.sh/uv/concepts/python-versions/#python-version-files) file to the repo?

ramanbansal1:
Thanks for the suggestion — a `.python-version` file is definitely useful, and I agree it would be good to have a single, visible declaration of the project’s preferred Python version.

However, pre-commit itself does not consume `.python-version` (or any other project files) when selecting the interpreter for hooks. It only uses what is explicitly declared in `.pre-commit-config.yaml` via default_language_version / language_version. As a result, adding `.python-version` alone would not make pre-commit deterministic, and we would still see CI vs local mismatches depending on which Python pre-commit discovers.

ramanbansal1:
However i can create a hook that will check versions and if not found will change it .

PR #3634, opened 2025-12-30
docs: Correct grammar issues in a/an usage

star1327p:
Correct grammar issues in a/an usage

The correct usages are `"an np"` and `"an FSMap"`, because the phonetic spelling of the word starts with a vowel.
https://www.grammarly.com/blog/grammar/indefinite-articles-a-and-an/

### DIFF
--- a/src/zarr/storage/_fsspec.py
+++ b/src/zarr/storage/_fsspec.py
@@ -147,7 +147,7 @@ class FsspecStore(Store):
         allowed_exceptions: tuple[type[Exception], ...] = ALLOWED_EXCEPTIONS,
     ) -> FsspecStore:
         """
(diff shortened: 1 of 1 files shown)


src/zarr/storage/_fsspec.py
```diff
@@ -147,7 +147,7 @@ def from_upath(
         allowed_exceptions: tuple[type[Exception], ...] = ALLOWED_EXCEPTIONS,
     ) -> FsspecStore:
         """
-        Create a FsspecStore from an upath object.
+        Create an FsspecStore from an upath object.
```
TomAugspurger: Since `upath` is pronounced "you-path" I think "a" is correct:

```suggestion
        Create an FsspecStore from a upath object.
```

PR #3636, opened 2026-01-02
Bump mimimum supported version of numpy

dstansby:
Since it's been about six months since SPEC recommended removing support for NumPy 1.26, this bumps the minimum version to 2.0. This duplicates work in https://github.com/zarr-developers/zarr-python/pull/3564, but I figured that it might be easier to do the version bumps (numpy and Python) independently since https://github.com/zarr-developers/zarr-python/pull/3564 appears to have grown in scope quite a bit. ping @ilan-gold - I hope that's okay?

My motivation for doing this is the slowest test in the matrix seems to be the one with NumPy 1.26, so I'm hoping this improves the total time GH actions takes to run all the tests to completion.

### DIFF
--- a/.github/workflows/check_changelogs.yml
+++ b/.github/workflows/check_changelogs.yml
@@ -12,7 +12,7 @@ jobs:
       - uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
 
       - name: Install uv
-        uses: astral-sh/setup-uv@681c641aba71e4a1c380be3ab5e12ad51f415867 # v7.1.6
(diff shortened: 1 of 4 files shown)


ilan-gold:
No worries! As you can see that PR is mostly other things anyway

.github/workflows/test.yml
```diff
@@ -21,20 +21,20 @@ jobs:
     strategy:
       matrix:
         python-version: ['3.11', '3.12', '3.13']
-        numpy-version: ['1.26', '2.2']
+        numpy-version: ['2.0', '2.2']
```
d-v-b, line 24: After setting the minimum version to 2.0, do we still need different numpy versions in the test matrix?

PR #3638, opened 2026-01-04
Add methods for getting bytes + json to store abc

d-v-b:
Adds some convenience methods to the store abc:

- `get_bytes_async`
- `get_json_async`
- `get_bytes`
- `get_json`

The async methods have a `_async` suffix. `get_bytes` and `get_json` are synchronous.

this PR allows the following, which is a huge QOL improvement for people working with Zarr stores.

```python
import zarr

store = MemoryStore({})
zg = zarr.create_group(store)

zmeta = store.get_json('zarr.json')
print(zmeta)
# {'attributes': {}, 'zarr_format': 3, 'node_type': 'group'}
```

### DIFF
--- a/tests/benchmarks/test_indexing.py
+++ b/tests/benchmarks/test_indexing.py
@@ -22,18 +22,27 @@ indexers = (
     (slice(None), slice(0, 3, 2), slice(0, 10)),
 )
 
+shards = (
+    None,
+    (50,) * 3,
+)
 
-@pytest.mark.parametrize("store", ["memory"], indirect=["store"])
+
+@pytest.mark.parametrize("store", ["memory", "memory_get_latency"], indirect=["store"])
 @pytest.mark.parametrize("indexer", indexers, ids=str)
+@pytest.mark.parametrize("shards", shards, ids=str)
 def test_slice_indexing(
-    store: Store, indexer: tuple[int | slice], benchmark: BenchmarkFixture
+    store: Store,
+    indexer: tuple[int | slice],
+    shards: tuple[int, ...] | None,
+    benchmark: BenchmarkFixture,
 ) -> None:
     data = create_array(
         store=store,
         shape=(105,) * 3,
         dtype="uint8",
         chunks=(10,) * 3,
-        shards=None,
+        shards=shards,
         compressors=None,
         filters=None,
         fill_value=0,
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -40,6 +40,7 @@ from zarr.core.metadata.v2 import ArrayV2Metadata
 from zarr.core.metadata.v3 import ArrayV3Metadata
 from zarr.core.sync import sync
 from zarr.storage import FsspecStore, LocalStore, MemoryStore, StorePath, ZipStore
+from zarr.testing.store import LatencyStore
 
 if TYPE_CHECKING:
     from collections.abc import Generator
@@ -58,8 +59,8 @@ if TYPE_CHECKING:
 
 
 async def parse_store(
-    store: Literal["local", "memory", "fsspec", "zip"], path: str
-) -> LocalStore | MemoryStore | FsspecStore | ZipStore:
+    store: Literal["local", "memory", "fsspec", "zip", "memory_get_latency"], path: str
+) -> LocalStore | MemoryStore | FsspecStore | ZipStore | LatencyStore:
     if store == "local":
         return await LocalStore.open(path)
     if store == "memory":
@@ -68,6 +69,8 @@ async def parse_store(
         return await FsspecStore.open(url=path)
     if store == "zip":
         return await ZipStore.open(path + "/zarr.zip", mode="w")
+    if store == "memory_get_latency":
+        return LatencyStore(MemoryStore(), get_latency=0.0001, set_latency=0)
     raise AssertionError
 
 


paraseba:
What is the advantage of making these helpers part of the ABC, instead of helper functions or a new concrete Store? Is the intention that certain implementations will override these? If so, what would be the advantages of overriding?

d-v-b:
> What is the advantage of making these helpers part of the ABC, instead of helper functions or a new concrete Store?

Putting this functionality on the ABC ensures that all stores implement it, without making per-store changes. Reading JSON is not actually new functionality, and ultimately I want to use these helpers internally (specifically, the async ones for reading JSON). I didn't do that in this PR because it requires changes to the `StorePath` class. 

> Is the intention that certain implementations will override these? If so, what would be the advantages of overriding?

Yes, in this PR the localstore and memorystore implementations override these to allow `prototype` to be `None`. See https://github.com/d-v-b/zarr-python/blob/d70a5e5277ca7a225d9fb0ac22945b5e1fcb8cc3/src/zarr/storage/_memory.py#L178-L184, and I'm following the pre-existing convention on those stores with these overrides. 

The advantage in this case is to work around the cumbersome requirement that store methods take a `prototype` parameter, which IMO we should move away from, but that's something for another PR

src/zarr/abc/store.py
```diff
@@ -206,6 +209,214 @@ async def get(
         """
         ...
 
+    async def get_bytes_async(
+        self, key: str, *, prototype: BufferPrototype, byte_range: ByteRequest | None = None
```
d-v-b, line 213: since the output of this method is a `bytes` object (CPU memory), I don't see the value of specifying a `prototype` in this method. can anyone suggest a situation where specifying the `prototype` would be useful here? 

cc @TomAugspurger

TomAugspurger: I think I agree with you, that a `prototype` probably isn't needed. But I'd suggest including it for consistency with the other methods.

src/zarr/abc/store.py
```diff
@@ -206,6 +209,214 @@ async def get(
...
+
+        See Also
+        --------
+        get_bytes_async : Asynchronous version of this method.
+        get_json : Synchronous method for retrieving and parsing JSON data.
+
+        Examples
+        --------
+        >>> store = MemoryStore()
+        >>> store.set("data", Buffer.from_bytes(b"hello world"))
```
TomAugspurger: I think this needs an `await`.

src/zarr/abc/store.py
```diff
@@ -206,6 +209,214 @@ async def get(
...
+
+        This is a synchronous wrapper around ``get_bytes_async()``. It should only
+        be called from non-async code. For async contexts, use ``get_bytes_async()``
+        instead.
+
+        Parameters
+        ----------
+        key : str, optional
+            The key identifying the data to retrieve. Defaults to an empty string.
+        prototype : BufferPrototype
```
TomAugspurger, line 274: We'll want to align this with what we do in the async version.

src/zarr/storage/_local.py
```diff
@@ -306,6 +306,236 @@ async def list_dir(self, prefix: str) -> AsyncIterator[str]:
...
+        Examples
+        --------
+        >>> store = await LocalStore.open("data")
+        >>> await store.set("data", Buffer.from_bytes(b"hello"))
+        >>> # No need to specify prototype for LocalStore
+        >>> data = await store.get_bytes_async("data")
+        >>> print(data)
+        b'hello'
+        """
+        if prototype is None:
```
TomAugspurger, line 357: Can't this be done in the base class, and then all the child classes reuse the same implementation?

d-v-b: I considered this direction. It would definitely be more convenient, but coupling the abstract store definition to the behavior of the `default_buffter_prototype()`, which in turn depends on the global config, seems much worse than other options (namely, baking buffer type into the store definition, or into store creation)

d-v-b:
> Since methods like `get` are asynchronous, I'd recommend making these convenience methods async-first and add a `_sync` suffix for the sync versions. `get_bytes` / `get_bytes_async`, etc.

@jhamman had the same preference and I haven't heard anyone strongly support the current state of this PR, so I'll make this change

src/zarr/storage/_local.py
d-v-b: see #3643

d-v-b:
`get_json` and `get_bytes` are now async, and `get_json_sync` and `get_bytes_sync` are now synchronous.

d-v-b:
in the dev meeting this week @maxrjones suggested making the new methods private API, so:

```
store._get_json_sync
store._get_json
store._get_bytes
store._get_bytes_sync
```

Is that an accurate summary max?

maxrjones:
> in the dev meeting this week @maxrjones suggested making the new methods private API, so:
> 
> ```
> store._get_json_sync
> store._get_json
> store._get_bytes
> store._get_bytes_sync
> ```
> 
> Is that an accurate summary max?

this was my off-hand reaction, but I am glad to defer to those with better API design instincts

d-v-b:
If it makes this PR more palatable, I'm happy making stuff private for now!

d-v-b:
the new methods are now all private. Since "read JSON" is indisputably a core behavior of a Zarr store, I think we should make these public sooner rather than later.

d-v-b:
As the methods are private, I removed the release notes, so I think this is ready for final review

d-v-b:
> The prototype argument is awkward IMO

Agreed but that's inherited awkwardness -- we have to call `store.get` with these methods, and `store.get` (awkwardly) takes the prototype as a required positional argument. It gets a bit less awkward with #3644

PR #3644, opened 2026-01-08
default buffer logic

d-v-b:
This PR builds on #3638 to add routines to the store ABC, and thereby all other stores, for handling additional types for the `prototype` parameter passed to `store.get` and the methods that call `get`. This change makes it easier to use store methods without unnecessary imports.

The type of `prototype` is widened to `type[Buffer] | BufferPrototype | None`. `type[Buffer] | BufferPrototype` is compacted to the type alias `BufferLike`. 

When `prototype` is `type[Buffer]`, the store will use that class directly. Since the stores never use the `BufferPrototype.nd_buffer`, passing a `Buffer` class should be preferred over using `BufferPrototype`.

When `prototype` is `None`, the store class retrieves a `Buffer` class by calling its `_get_default_buffer_class` method. The default  implementation returns the `default_buffer_prototype` which makes our ABC less abstract but keeps these changes non-breaking.


normanrz:
@d-v-b Could you elaborate on the motivation for widening the Store API like this?

d-v-b:
Stores don't do anything with `NDBuffer` classes. They just use a `Buffer` class. So `prototype: BufferPrototype` is indirect -- no store will use the `BufferPrototype.nd_buffer` element.  That's why `prototype` can be a `Buffer` class in this PR. 

And multiple stores have already implemented logic for selecting a default buffer class. The changes here propagate that sensible design to all store classes.

normanrz (approved):

d-v-b:
Flagging this as a _breaking change_ for third-party stores, as `zarr-python` would after this change pass new types into `Store.get`. Assuming the mitigation to be straightforward (third-party stores have 2 handle `None` and a `Buffer` class as input), this could land in a minor release.

d-v-b:
I am closing this, because it's a breaking change to store implementations without much benefit. IMO we need to make this change, but as as part of a broader store refactor that adds features.

PR #3648, opened 2026-01-12
Fix deprecation of setting shape on a numpy array in 2.5+

mscheltienne:
Fixes deprecation observed in `numpy` 2.5+ (pre-release):

```
 .venv/lib/python3.13/site-packages/zarr/codecs/vlen_utf8.py:98: in _decode_single
    decoded.shape = chunk_spec.shape
    ^^^^^^^^^^^^^
E   DeprecationWarning: Setting the shape on a NumPy array has been deprecated in NumPy 2.5.
E   As an alternative, you can create a new view using np.reshape (with copy=False if needed).
```

### DIFF
--- a/changes/3648.misc.md
+++ b/changes/3648.misc.md
@@ -1,0 +1,1 @@
+Fix deprecation of setting a shape on an array directly in ``numpy`` 2.5+.


mscheltienne:
@d-v-b Note that this change is not fully backward compatible, if we want it backward compatible we would set `copy=False` instead of the default `copy=None`. I consider that `copy=None` is an enhancement as it will now:

- (1) Try to reshape without copy (view)
- (2) If it can't, reshape with a copy

If we set `copy=False`, it will explicitly raise if the `view` is not possible. Up-to-you, let me know if you think this added fallback mechanism is an enhancement or if hiding this copy is detrimental.

d-v-b:
As `decoded` is an object array, when would we _not_ be able to create a view?

mscheltienne:
Well it should- and personally I don't see the harm in adding this fallback- but if you prefer to have an explicit raise instead to catch eventual future changes that would somehow yield a non contiguous array of objects, then setting `copy=False` would be better.

d-v-b:
in that case the `copy=False` option seems good. Probably better to err on the side of avoiding hidden allocations wherever possible.

PR #3649, opened 2026-01-12
Set copy=False in reshape operation

mscheltienne:
Follow-up on https://github.com/zarr-developers/zarr-python/pull/3648/

@d-v-b


mscheltienne:
Exactly what I just realized- `copy` was introduced in `numpy` 2.1 which is not that old. In this case, @d-v-b I would advocate to keep the current implementation of #3648 with an implicit copy if a view is impossible- which in theory should not occur. Otherwise, I can also conditionally add the `copy` argument for `numpy >= 2.1`, let me know which solution you prefer for Zarr.

mscheltienne:
I added a version with a compat function in this PR, which retains the behavior across all `numpy` version. I'll let you decide if this is worth the added maintenance burden or if using the defaults as implemented in #3648 is sufficient. Feel free to close if you prefer #3648.

Issue #3653, opened 2026-01-15
Which versions of numpy do we need to test


d-v-b:
> Thanks @dstansby! We should decide on a testing strategy for different numpy versions, but that shouldn't stop this PR.

_Originally posted by @d-v-b in https://github.com/zarr-developers/zarr-python/pull/3636#pullrequestreview-3638532394_
            
Now that we don't support numpy 1.x, do we still need 2 versions of numpy in our test matrix, or can we just test against the oldest supported version of numpy 2 in our main tests, and add numpy to our upstream tests (if it's not there already)

PR #3654, opened 2026-01-15
Simplify testing of different numpy versions

dstansby:
Now the minimum numpy dependency is 2.0, it seems overkill to have every test run on both NumPy 2.0 and NumPy 2.2. Instead, this removes configuration of numpy in the test matrix, making it simpler in the process.

The minimum version of NumPy is still tested in the min-deps run, and the latest nightly wheel is tested in the upstream test run.

Fixes https://github.com/zarr-developers/zarr-python/issues/3653.


dcherian:
`deps=minimal` appears to install `numpy 2.4.1`? [example](https://github.com/zarr-developers/zarr-python/actions/runs/21033907878/job/60476604328?pr=3654)

dstansby:
`minimal` means the minimum set of dependencies needed to run `zarr`, not the minimum version of dependencies. In the `min_deps` environment NumPy 2.0.2 is installed: https://github.com/zarr-developers/zarr-python/actions/runs/21033907878/job/60476604299?pr=3654#step:5:55

dcherian:
confirmed min_deps installs numpy 2.0.2

PR #3658, opened 2026-01-16
chore/use prek

d-v-b:
Switches our pre-commit checker from `pre-commit` to [`prek`](https://github.com/j178/prek), because `prek` is way faster and less socially complicated than pre-commit (e.g., https://github.com/pre-commit/pre-commit.com/issues/687#issuecomment-2704318115). See [this comment](https://github.com/j178/prek/issues/73#issuecomment-3178076699) the apache airflow project detailing why they switched.

### DIFF
--- a/changes/3658.misc.md
+++ b/changes/3658.misc.md
@@ -1,0 +1,1 @@
+Switch from `pre-commit` to [`prek`](https://github.com/j178/prek) for pre-commit checks.


maxrjones (approved):
👍

Is it worth renaming the readme badge?

d-v-b:
I think the readme badge indicates whether we are passing according to pre-commit CI, which is still applicable since both prek and pre-commit use the same .pre-commit configuration?

maxrjones:
> I think the readme badge indicates whether we are passing according to pre-commit CI, which is still applicable since both prek and pre-commit use the same .pre-commit configuration?

makes sense, thanks for the explanation

PR #3663, opened 2026-01-20
Add benchmarks for sharded + local store indexing

aldenks:
Parameterize with shards and no shards. Also parameterize with local + memory store to have an example of a store which has some modest latency.

When testing performance of optimizing sharded zarr reads I found that the memory store is too fast to show any difference, but the local store makes performance improvements there very visible. The extra parameterizations do expand the test space; test_slice_indexing with benchmarking on now takes ~30 seconds vs ~7 seconds on main.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -70,7 +70,7 @@ async def parse_store(
     if store == "zip":
         return await ZipStore.open(path + "/zarr.zip", mode="w")
     if store == "memory_get_latency":
-        return LatencyStore(MemoryStore(), get_latency=0.01, set_latency=0)
+        return LatencyStore(MemoryStore(), get_latency=0.0001, set_latency=0)
     raise AssertionError
 
 


d-v-b:
if we want more control over the latency, we could wrap memory storage with a [`LatencyStore`](https://github.com/zarr-developers/zarr-python/blob/c06923d3888b07c37d90c11a665be3ca11dbd949/src/zarr/testing/store.py#L530) here. This would keep the benchmark a bit more isolated from stuff the OS is doing than localstore. Happy to implement this if you don't mind me pushing to this branch

aldenks:
@d-v-b go ahead! If you make it I can't be accused of juicing my numbers by picking high latencies ;p

In all seriousness, that would be quite useful for picking a good default coalesce_max_gap_bytes.

d-v-b:
with .0001s of latency on `get` requests against a `MemoryStore`, we get performance that's ~6x slower than a regular `MemoryStore`. This is actually much slower than local storage, but it should be a more durable metric, if we assume that local storage devices get faster, faster than system memory

aldenks:
Thank you @d-v-b!

Issue #3666, opened 2026-01-24
DOC: confusing statement about write_empty_chunks default


tlambert03:
### Describe the issue linked to the documentation

in the performance docs https://zarr.readthedocs.io/en/latest/user-guide/performance/#empty-chunks it says:

> If you know that your data will form chunks that are almost always non-empty, then there is no advantage to the optimization described above. In this case, creating an array with write_empty_chunks=True (the default) will instruct Zarr to write every chunk without checking for emptiness.

in the docs source it's [here](https://github.com/zarr-developers/zarr-python/blob/3fa19d237971baa6fadce176bdd93d88eb452e10/docs/user-guide/performance.md?plain=1#L127-L128)

but  this contradicts what I see when I check the zarr config (in a clean env, with only zarr installed and imported)

```
➜ uv run --with zarr python -c "import zarr; print(zarr.config.get('array.write_empty_chunks'))"
Installed 8 packages in 49ms
False
```

am I misunderstanding something?  or are the docs wrong about the default value for `write_empty_chunks`

d-v-b:
i think the docs are simply wrong here, and probably need to be updated to be kept in sync with the actual config. I'd love an automatic way to catch drift like this. maybe the best thing would be an inlne python code snippet in the docs that assumes the default claimed by the defaults, and which would fail when we run our doctests

tlambert03:
ok, thanks for confirming!

RituDSML:
The docs are now built with MkDocs (see .readthedocs.yaml), so the suggestion about using Sphinx doctests to catch drift may not apply in the current setup.

d-v-b:
this was addressed via #3668

PR #3668, opened 2026-01-25
Improve the usability of array runtime config

### LINKED ISSUE #3667: getting the `write_empty_chunks` attribute of an array's config is onerous
This arose while working on #3666: 

Given an array, I think the only way to get the `write_empty_chunks` attribute from the configuration is this:

```python
z = zarr.create_array({}, shape=(1,), dtype='uint8', config={"write_empty_chunks": True})
z._async_array._config.write_empty_chunks # super indirect!
# True
```

Since we declare the config with the `config` keyword argument in `create_array`, it might make sense for the array class to have a `config` attribute / property. 

And while we're at it, we should provide an option for getting an array with a _different_ config via a method like `Array.with_config(...new_config) -> Array`

d-v-b:
Fixes #3667 and #3666 by updating the docs about the default `write_empty_chunks` value, and adds some API to `Array` and `AsyncArray` so that the doc update could be done properly (i.e., without using private API).

The new API on `Array` and `AsyncArray` is as follows:
- `AsyncArray.config` is public (it was private)
- `Array.config` is a new property that exposes `._async_array.config`
- `AsyncArray` and `Array` both have a `.with_config` method that creates a new `AsyncArray` / `Array` with an updated config.  

This means users can now easily check the runtime configuration of an array, and get a new array (pointing to the same data) with a different configuration if they want.

### DIFF
--- a/docs/user-guide/arrays.md
+++ b/docs/user-guide/arrays.md
@@ -154,6 +154,32 @@ z.append(np.vstack([a, a]), axis=1)
 print(f"Shape after second append: {z.shape}")
 ```
 
+## Runtime configuration
+
+Zarr arrays are parametrized with a configuration that determines certain aspects of array behavior.
+
+We currently support two configuration options for arrays: `write_empty_chunks` and `order`.
+
+| field | type | default | description |
+| - |     - | - | - |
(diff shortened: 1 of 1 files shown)


d-v-b:
I added a section to the user guide demonstrating how to use the new attribute / method.

PR #3676, opened 2026-01-26
lift async array methods to stand-alone functions

d-v-b:
This PR takes the bulk of the `AsyncArray` methods and delegates them to stand-alone functions. This makes the core functionality like reading chunks, writing chunks, resizing, appending, etc, possible without creating an `AsyncArray`. 

A set of focused, independent functions is a much stronger API foundation than a single object.

Opening as a draft for discussion -- for example, should we have `_sync` versions of all these new functions?


d-v-b:
this is ready for review. this is a purely internal refactor that shouldn't affect anything at runtime.

PR #3677, opened 2026-01-28
feat/experimental array

d-v-b:
this PR builds on #3676 to define a new array class in `experimental.array`. 

This class has both async and sync methods, making it a combination of `zarr.core.array.Array` and `zarr.core.array.AsyncArray`. Adopting this change alone would delete a _lot_ of code from `zarr.core.array`, but the real goal for this class is to serve as a testbed for developing variable-length chunks in #3534 

Opening as a draft so we can talk about the design.


d-v-b:
I'm closing this because we can experiment on our existing array class easily enough!

PR #3698, opened 2026-02-09
Make ObjectStore backend generic over any obspec store

kylebarron:
In https://github.com/zarr-developers/zarr-python/pull/1661 we merged an [Obstore](https://github.com/developmentseed/obstore)-based backend. It sounds like the Obstore backend has become very popular, and @maxrjones found that, at least in some situations, the Obstore backend [can be significantly faster](https://github.com/maxrjones/zarr-obstore-performance).

One problem with the Obstore backend, however, is that it's **strictly tied** to Obstore. The type hinting and runtime behavior all require exact instances from the Obstore package. 

But very often people might want to insert some middleware. 

- **Caching**: disk cache, memory cache of different flavors
- **Timing**: how long are different requests taking
- **Request inspection**, e.g. for debugging, or for evaluating async performance

The goal of [Obspec](https://github.com/developmentseed/obspec) is to define generic protocols to **cleanly** enable this. As described in the [initial release post](https://developmentseed.org/obspec/latest/blog/2025/06/25/introducing-obspec-a-python-protocol-for-interfacing-with-object-storage/) from last summer, Obspec should allow downstream libraries to depend on a **protocol-based API** that works with **any implementation** that provides the given signature.

E.g. if you think of the simplest pseudocode example of a cache:

```py
from typing_extensions import Buffer
from obspec import GetRange

class SimpleCache(GetRange):
    """A simple cache for synchronous range requests that never evicts data."""

    def __init__(self, client: GetRange):
        self.client = client
        self.cache: dict[tuple[str, int, int | None, int | None], Buffer] = {}

    def get_range(
        self,
        path: str,
        *,
        start: int,
        end: int | None = None,
        length: int | None = None,
    ) -> Buffer:
        cache_key = (path, start, end, length)
        if cache_key in self.cache:
            return self.cache[cache_key]

        response = self.client.get_range(
            path,
            start=start,
            end=end,
            length=length,
        )
        self.cache[cache_key] = response
        return response
```

Then if a function expects an object implementing `GetRange`:

```py
def my_function(client: GetRange, path: str, *, start: int, end: int):
    buffer = client.get_range(path, start=start, end=end)
    # Do something with the buffer
    print(len(memoryview(buffer)))
```

Then now you can pass in _either_ the raw obstore backend _or_ the backend wrapped by the cache:

```py
from obstore.store import S3Store

store = S3Store("bucket")
caching_wrapper = SimpleCache(store)
my_function(caching_wrapper, "path.txt", start=0, end=10)
# second request will be cached by `SimpleCache`
my_function(caching_wrapper, "path.txt", start=0, end=10)
```

This architecture is much more tractable for end users than if Obstore implemented its own caching natively. Since users have full access to the cache (i.e. it isn't hidden away inside Rust), users can check methods of `SimpleCache` to track how much memory the cache is using and to manually evict cache items if they wanted.

@maxrjones has been starting to collect utilities around obspec in https://github.com/virtual-zarr/obspec-utils.

This is **backwards-compatible** (at least if you ignore the obstore version bump from 0.5.1 — released [March 2025](https://pypi.org/project/obstore/0.5.1/) — to 0.7.0 — [released June 2025](https://pypi.org/project/obstore/0.7.0/)).

### Implementation notes

- This relies on [structural subtyping](https://typing.python.org/en/latest/reference/protocols.html), i.e. that the _shape_ matters not the _name_. This works for all protocols, but we have to implement special support for exceptions, since exceptions don't support structural subtyping 🥲. As described in [`Exceptions`](https://developmentseed.org/obspec/latest/api/exceptions/) in the obspec docs, the workaround I chose is to use well-defined names, and `map_exception` will convert any external exceptions to exceptions subclassing from `obspec.exceptions`.

### TODO:


TomNicholas:
+1 Xarray could make use of this. (When reading non-icechunk stores - icechunk already has a caching layer)

kylebarron:
I'm not sure where the other locations are where I need to define obspec as a dependency

d-v-b (changes requested):
looks good, just need a release note in `changes`

kylebarron:
As discussed in https://github.com/zarr-developers/VirtualiZarr/issues/1092, this would also enable users to use the `ObjectStore` adapter class without a direct dependency on Obstore, which could be useful for users in constrained environments.

d-v-b:
sounds good, do you want me to resolve the conflicts?

d-v-b:
with the conflicts resolved there are some blockers, which I will summarize here:
- the `obspec` import is not conditional
- we still require that store classes have a FQN that starts with "obstore", so we are still doing nominal typing :/ 
- exception handling is not complete

i'm happy to deal with these

kylebarron:
If you'd like to make updates, I can review. I have a lot going on/catching up on at the moment

d-v-b:
@kylebarron have a look, i fixed the conflicts and the issues I mentioned

src/zarr/storage/_obstore.py
```diff
@@ -21,30 +20,102 @@
...
+    [obstore](https://developmentseed.org/obstore/) stores for AWS S3, Google
+    Cloud Storage and Azure Blob Storage, or a wrapper (a cache, a request logger)
+    around one of them.
 
     Parameters
     ----------
-    store : obstore.store.ObjectStore
-        An obstore store instance that is set up with the proper credentials.
+    store : ObspecInput
+        Any object implementing the ``DeleteAsync``, ``GetAsync``, ``GetRangeAsync``,
```
kylebarron, line 114: Nit: does the zarr-python codebase usually use double-backticks for code blocks? I find it overly verbose because Markdown only requires a single backtick. But I know AI always uses double :confused:

kylebarron (commented):
This looks good to me! Thanks!

src/zarr/storage/_obstore.py
d-v-b: yeah thats our llm watermark. i can remove it :)

Issue #3699, opened 2026-02-10
Wrapper stores don't support with_read_only


BorisTheBrave:
### Zarr version

v3.1.3

### Numcodecs version

N/A

### Python Version

3.11.11

### Operating System

Linux

### Installation

uv

### Description

LoggingStore, LatencyStore and CacheStore all don't support with_read_only.

This is a pain to work with. Particularly with xarray, which defaults to needing a readonly store.

### Steps to reproduce

```python
# /// script
# requires-python = ">=3.11"
# dependencies = [
#   "zarr@git+https://github.com/zarr-developers/zarr-python.git@main",
#    "xarray",
# ]
# ///
#

from zarr.storage import LocalStore, LoggingStore
import xarray as xr

store = LocalStore(root="tmp", read_only=False)
store = LoggingStore(store)
dataset = xr.open_dataset(store, engine="zarr", mode="w") # Works
dataset = xr.open_dataset(store, engine="zarr") # Fails
```

Fails with

```
ValueError: Store is not read-only but mode is 'r'. Unable to create a read-only copy of the store. Please use a read-only store or a storage class that implements .with_read_only().
```

d-v-b:
thanks for the report, looks like we need to add the with_read_only logic to the wrapper store

BorisTheBrave:
> looks like we need to add the with_read_only logic to the wrapper store

Yes, I've created a PR for you to do exactly that.

PR #3700, opened 2026-02-10
Support with read only in wrappers

BorisTheBrave:
Fixes https://github.com/zarr-developers/zarr-python/issues/3699

All three wrapper stores (Latency, Logging, Cache) now support with_read_only.

CacheStore required a bit of changes - some of the mutable state was lifted to another class, so that multiple CacheStore objects could share the same one.

* (N/A) Add docstrings and API docs for any new/modified user-facing classes and functions
* (N/A) New/modified features documented in `docs/user-guide/*.md`

### DIFF
--- a/src/zarr/experimental/cache_store.py
+++ b/src/zarr/experimental/cache_store.py
@@ -4,6 +4,7 @@ import asyncio
 import logging
 import time
 from collections import OrderedDict
+from dataclasses import dataclass, field
 from typing import TYPE_CHECKING, Any, Literal, Self
 
 from zarr.abc.store import ByteRequest, Store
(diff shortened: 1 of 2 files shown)


src/zarr/experimental/cache_store.py
```diff
@@ -15,6 +15,16 @@
     from zarr.core.buffer.core import Buffer, BufferPrototype
 
 
+class _CacheState:
```
d-v-b, line 20: if this has to be a class, it should be a dataclass (so we can initialize it with `__init__`). and if it doesn't have to be a class (e.g, it's just a collection of attributes, with no methods), we could use a typeddict.

I'm happy to make these changes if that's ok with you @BorisTheBrave

src/zarr/experimental/cache_store.py
```diff
@@ -71,13 +81,7 @@ class CacheStore(WrapperStore[Store]):
     max_size: int | None
     key_insert_times: dict[str, float]
```
d-v-b: should this be part of the cache state?

src/zarr/experimental/cache_store.py
BorisTheBrave: Sure, I'm happy to hand it over. It has no need to be a class, it's just a collection of properties.

src/zarr/experimental/cache_store.py
BorisTheBrave: Logically yes, but as it wasn't marked internal (with an underscore) I felt it was safer to not move it.

Issue #3703, opened 2026-02-12
minimize polymorphism in the codec pipeline?


d-v-b:
we have a very expensive `isinstance` check in the codec pipeline: https://github.com/zarr-developers/zarr-python/blob/2a032a8d370cf8e16097e3285e697dd70b89d717/src/zarr/codecs/bytes.py#L82. We need to remove this check, and I would suggest going further -- we should check if our strategy of supporting CPU and CUDA memory with the same codecs is compatible with getting the best performance for either type. I don't think supporting CUDA (used by a minority of users) is worth _any_ measurable performance penalty for folks using CPU memory.

PR #3704, opened 2026-02-12
perf:remove isinstance check

d-v-b:
removes an expensive `isinstance` check inside `BytesCodec._decode_single`. Isinstance on `runtime_checkable` protocols is expensive and this particular check is in a hotspot. Without the check, we are slightly less type-safe, but users who somehow get a non-ndarray into this part of the code will get an immediate a runtime error.

### DIFF
--- a/changes/3657.bugfix.md
+++ b/changes/3657.bugfix.md
@@ -1,0 +1,1 @@
+Fix obstore _transform_list_dir implementation to correctly relativize paths (removing lstrip usage).
--- a/changes/3706.misc.md
+++ b/changes/3706.misc.md
@@ -1,0 +1,1 @@
+Allow NumPy ints as input when declaring a shape.
--- a/src/zarr/core/common.py
+++ b/src/zarr/core/common.py
@@ -21,6 +21,7 @@ from typing import (
     overload,
 )
 
+import numpy as np
 from typing_extensions import ReadOnly
 
 from zarr.core.config import config as zarr_config
@@ -37,7 +38,7 @@ ZATTRS_JSON = ".zattrs"
 ZMETADATA_V2_JSON = ".zmetadata"
 
 BytesLike = bytes | bytearray | memoryview
-ShapeLike = Iterable[int] | int
+ShapeLike = Iterable[int | np.integer[Any]] | int | np.integer[Any]
 # For backwards compatibility
 ChunkCoords = tuple[int, ...]
 ZarrFormat = Literal[2, 3]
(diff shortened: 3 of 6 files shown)


d-v-b:
35% perf improvement seems good

src/zarr/codecs/bytes.py
```diff
@@ -72,20 +70,15 @@ async def _decode_single(
...
         chunk_spec: ArraySpec,
     ) -> NDBuffer:
-        assert isinstance(chunk_bytes, Buffer)
         # TODO: remove endianness enum in favor of literal union
         endian_str = self.endian.value if self.endian is not None else None
         if isinstance(chunk_spec.dtype, HasEndianness):
             dtype = replace(chunk_spec.dtype, endianness=endian_str).to_native_dtype()  # type: ignore[call-arg]
         else:
             dtype = chunk_spec.dtype.to_native_dtype()
         as_array_like = chunk_bytes.as_array_like()
```
dcherian, line 79: can we just have `as_array_like` become `as_ndarray_like`?

d-v-b: you mean just rename the variable? We have to choose where the surprise is: at the `as_array_like` call (surprising if it returns an ndarraylike) or `from_ndarray_like` (surprising if it accepts an arraylike) call. I don't see much of a difference here, but happy to rename if you feel strongly

dcherian: (this is why I shouldn't reply when half-asleep at night)

Apologies for the confusion. Can we not have `Buffer.as_ndarray_like()` that makes the bytes ready for the codec pipeline in the form the codec pipeline needs it i.e. `NDArrayLike`? That way it *is* type safe and performant.

dcherian:
Also, this will close #3703 ?

src/zarr/codecs/bytes.py
d-v-b: ah sorry so a new method, that ensures that the contents of the buffer are an ndarray? yeah I think that should be easy to add!

d-v-b:
> Also, this will close #3703 ?

it will close part of it, but not the underlying issue about our basic buffer / codec design.

src/zarr/codecs/bytes.py
d-v-b: is it OK if we spin that out into a separate issue?

dcherian: Sure, but isn't the needed change just modifying `as_array_like` to call `np.asanyarray` in the CPU buffer

d-v-b: doesn't this requires changing the `as_array_like` method for gpu buffer too, and/or the method on the abc? https://github.com/zarr-developers/zarr-python/blob/e03cfc897ac5e0a20001ee37580dce19ea3f567a/src/zarr/core/buffer/core.py#L234

dcherian: ya, up to you as to when you want to fix it...

d-v-b: Buffer is [public api](https://github.com/zarr-developers/zarr-python/blob/e03cfc897ac5e0a20001ee37580dce19ea3f567a/src/zarr/abc/buffer.py#L5) so I don't think we want to change that as part of a performance fix.

PR #3705, opened 2026-02-12
perf:improve performance of morton order iter

d-v-b:
Our morton_order_iter function uses nested O(N) loops and doesn't cache its results. These make it slow. This PR fixes both of these, so it should be faster.


normanrz (approved):

normanrz:
Really cool!

PR #3706, opened 2026-02-12
allow numpy ints in shapelike

### LINKED ISSUE #3680: Can't use numpy types for array shapes
### Zarr version

3.1.5

### Numcodecs version

0.16.5

### Python Version

3.12

### Operating System

Mac

### Installation

pip

### Description

In zarr 3, shapes can no longer be numpy int types. This used to work in zarr 2.

### Steps to reproduce

```python
# /// script
# requires-python = ">=3.11"
# dependencies = [
# ]
# ///
#

import zarr
import numpy as np


a = np.ones(2, dtype=int)
zarr.create_array("test", shape=(int(a.max())), dtype=int)  # works
zarr.create_array("test2", shape=(a.max()),
                  dtype=int)  # TypeError: Expected an integer or an iterable of integers. Got 1 instead.
```

### Additional output

_No response_

d-v-b:
closes #3680

### DIFF
--- a/changes/3705.bugfix.md
+++ b/changes/3705.bugfix.md
@@ -1,0 +1,1 @@
+Fix a performance bug in morton curve generation.
--- a/src/zarr/core/indexing.py
+++ b/src/zarr/core/indexing.py
@@ -7,7 +7,7 @@ import operator
 from collections.abc import Iterator, Sequence
 from dataclasses import dataclass
 from enum import Enum
-from functools import reduce
+from functools import lru_cache, reduce
 from types import EllipsisType
 from typing import (
     TYPE_CHECKING,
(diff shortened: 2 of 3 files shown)


src/zarr/core/common.py
```diff
@@ -185,23 +186,28 @@ def parse_named_configuration(
 
 
 def parse_shapelike(data: ShapeLike) -> tuple[int, ...]:
-    if isinstance(data, int):
+    """
+    Parse a shape-like input into an explicit shape.
+    """
+    if isinstance(data, int | np.integer):
```
maxrjones:
```suggestion
    if isinstance(data, bool):
        raise TypeError(f"Expected an integer or an iterable of integers. Got {data} instead.")
    if isinstance(data, int | np.integer):
```

src/zarr/core/common.py
```diff
@@ -185,23 +186,28 @@ def parse_named_configuration(
...
         if data < 0:
             raise ValueError(f"Expected a non-negative integer. Got {data} instead")
-        return (data,)
+        return (int(data),)
     try:
         data_tuple = tuple(data)
     except TypeError as e:
         msg = f"Expected an integer or an iterable of integers. Got {data} instead."
         raise TypeError(msg) from e
```
maxrjones, line 200: could this be consolidated with L210?

d-v-b: there's invalid stuff (dict keys, strings) that could come out of an iterable `data` and be converted to int without any error. If `data` is `["0", "1"]`, then `tuple(int(x) for x in data)`  will not error, and then we are accepting strings as shapes, which I didn't want to do here. So I think we need to end up calling tuple twice.

PR #3707, opened 2026-02-12
TEST: fix group selection for delete

ianhi:
Fix stateful test delete group precondition. This was excluding any groups at the root level, which can lead to a hypothesis error if you have these three groups `/`, `0`, `1` because all of those were being excluded. the root group is nver in `all_groups` so this change should be safe to just select from all of them.

### DIFF
--- a/changes/3713.misc.md
+++ b/changes/3713.misc.md
@@ -1,0 +1,1 @@
+Vectorize get_chunk_slice for faster sharded array writes.
--- a/src/zarr/codecs/sharding.py
+++ b/src/zarr/codecs/sharding.py
@@ -46,6 +46,8 @@ from zarr.core.dtype.npy.int import UInt64
 from zarr.core.indexing import (
     BasicIndexer,
     SelectorTuple,
+    _morton_order,
+    _morton_order_keys,
     c_order_iter,
     get_indexer,
     morton_order_iter,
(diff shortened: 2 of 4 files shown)


ianhi:
small correciton. root **should** never be there, but nothign was explcitly disallowing. so i updated the sampling to preclude it.

src/zarr/testing/stateful.py
```diff
@@ -340,13 +340,13 @@ def delete_array_using_del(self, data: DataObject) -> None:
         self.all_arrays.remove(array_path)
 
     @precondition(lambda self: self.store.supports_deletes)
-    @precondition(lambda self: len(self.all_groups) >= 2)  # fixme don't delete root
+    @precondition(lambda self: bool(self.all_groups - {"", "/"}))
```
dcherian: Why is `""` in here?

ianhi: I'm pretty sure the root group gets put there as "", or at least i think i remember that happening.

ianhi: looking again i might have been confused by this: https://github.com/zarr-developers/zarr-python/blob/23596c148087d868a07427315a4d9b05112a0015/src/zarr/testing/stateful.py#L112-L115

beceuase we strip the `/` at the start of paths i was worried that root might end up there as an empty string.

dcherian: IIRC it should be "/", so I prefer not filtering for `""`

d-v-b: the name of the root group is the empty string "". You probably get the same thing with "/", but that's technically the root group + a separator.

ianhi: so maybe we only filter for `""`?

dcherian: sounds right but please confirm what's inserted in the bundle; we should only filter for one.

ianhi: ok I just removed all the subtractino of groups. seemingly the root group doesn't ever get added so I'm not as worried as I was

Issue #3690, opened 2026-02-03
Experimntal caching for v3 does not work with ByteRequest


dsparber:
### Zarr version

3.1.2

### Numcodecs version

0.16.5

### Python Version

3.14

### Operating System

Mac

### Installation

using pip into virtual env

### Description

The current caching implementation ignores the ByteRequest parameter when storing data to cache. This results in the wrong data being returned when loading different byte ranges of the same key. 

This makes sharding incompatible with caching.

### Steps to reproduce

```python
import asyncio

from zarr.abc.store import RangeByteRequest
from zarr.core.buffer.cpu import Buffer, buffer_prototype
from zarr.experimental.cache_store import CacheStore
from zarr.storage import MemoryStore


async def main() -> None:
    source = MemoryStore()
    await source.set(key="foo", value=Buffer.from_bytes(b"bar baz"))

    cache = CacheStore(store=source, cache_store=MemoryStore())

    baz_range = RangeByteRequest(start=4, end=7)
    bar_range = RangeByteRequest(start=0, end=3)

    value_bar = await cache.get(
        key="foo",
        prototype=buffer_prototype,
        byte_range=bar_range,
    )

    value_baz = await cache.get(
        key="foo",
        prototype=buffer_prototype,
        byte_range=baz_range,
    )

    source_baz = await source.get(
        key="foo",
        prototype=buffer_prototype,
        byte_range=baz_range,
    )

    print(
        f"bar={value_bar.to_bytes()}, "
        f"baz={value_baz.to_bytes()}, "
        f"source_baz={source_baz.to_bytes()}"
    )
    # expected behaviour: bar=b'bar', baz=b'baz', source_baz=b'baz'
    # actual result: bar=b'bar', baz=b'', source_baz=b'baz'


if __name__ == "__main__":
    asyncio.run(main())

```

d-v-b:
thanks for this report! Offhand I don't know how we should fix this, but I'm sure this problem has been solved elsewhere. Unless someone else fixes this, I might have something by the end of the week.

maxrjones:
> thanks for this report! Offhand I don't know how we should fix this, but I'm sure this problem has been solved elsewhere. Unless someone else fixes this, I might have something by the end of the week.

there's some design similarities with https://github.com/virtual-zarr/obspec-utils/pull/64 in case it's helpful to know about those efforts

d-v-b:
[...]

PR #3710, opened 2026-02-15
fix: implement separate cache for byte-range-requests

d-v-b:
This PR adds a support for byte-range caching to the experimental `CacheStore`. Claude did all the heavy lifting here so we need to review carefully.

Some design notes:

The dual-store design of the `CacheStore` complicates caching range requests. I considered 3 options:
- Don't cache range requests at all. Cache the entire object, then do byte-range reads against that object. Major downside of this is the cost of fetching the entire object, which could defeat the purpose of the range request in the first place.
- Cache range requests in the regular caching layer, using a key that's a unique stringification of the byte range request, like `f"{key}.{request type}.{request params}"`. Downside of this is potentially littering the cache (which might be the local file system) with non-zarr keys.
- Use a separate in-memory cache specifically for byte ranges. This is the option I took. It's simple, but a downside is that sharded data won't end up stored in the primary cache unless the user explicitly fetches an entire shard.

I would love to hear other ideas / suggestions. Maybe we support all 3 options via a parameter on the cache store?

### DIFF
--- a/changes/3721.misc.md
+++ b/changes/3721.misc.md
@@ -1,0 +1,1 @@
+Adds synchronous (non-async) encoding and decoding methods to CPU-bound codecs. This is necessary for performance optimizations based on avoiding `asyncio` overhead. These new methods are described by a new protocol: `SupportsSyncCodec`.
--- a/src/zarr/abc/codec.py
+++ b/src/zarr/abc/codec.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 from abc import abstractmethod
 from collections.abc import Mapping
-from typing import TYPE_CHECKING, Generic, TypeGuard, TypeVar
+from typing import TYPE_CHECKING, Generic, Protocol, TypeGuard, TypeVar, runtime_checkable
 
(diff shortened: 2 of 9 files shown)


d-v-b:
closes #3690

d-v-b:
@dsparber take a look and let me know if this works for your needs

dsparber:
Looks great, thanks a lot! The double cache approach seems like a good trade-off. 

One small concern (maybe I overlooked it): Is there a way to limit the size of the in-memory cache for ranges? 
Just to avoid an unbounded RAM usage, when making many queries to the zarr.

d-v-b:
> Looks great, thanks a lot! The double cache approach seems like a good trade-off.
> 
> One small concern (maybe I overlooked it): Is there a way to limit the size of the in-memory cache for ranges? Just to avoid an unbounded RAM usage, when making many queries to the zarr.

Ranges and full fetches both contribute to the cache size limit. So there shouldn't be a risk unbounded RAM usage specifically for range queries.

PR #3712, opened 2026-02-16
perf:Add additional sharding benchmarks

mkitti:
## Summary

Added benchmarks for monitoring Morton order computation in sharded arrays. These benchmarks help assess the impact of Morton order optimizations in the context of I/O operations.

## Benchmarks Added

- `test_sharded_morton_indexing` - Sharded array indexing with power-of-2 chunks per shard
- `test_sharded_morton_indexing_large` - Large shard with 32^3 = 32,768 chunks
- `test_sharded_morton_single_chunk` - Reading a single chunk from a large shard
- `test_morton_order_iter` - Direct benchmark of `morton_order_iter` (no I/O)
- `test_sharded_morton_write_single_chunk` - Writing a single chunk to a large shard (best end-to-end test)

### Single Chunk Write (Best End-to-End Test)

Writing a single 1x1x1 chunk to a shard with 32^3 = 32,768 chunks:

| Branch | Mean Time | Improvement |
|--------|-----------|-------------|
| Main (no optimization) | 425ms | - |
| Optimized (PR #3708) | 261ms | **164ms (39% faster)** |

### Morton Order Computation (Micro-benchmark)

Direct `morton_order_iter` benchmark without I/O:

| Shape | Main Branch | Optimized | Speedup |
|-------|-------------|-----------|---------|
| (8, 8, 8) | 2.73ms | 0.85ms | **3.2x** |
| (16, 16, 16) | 25.53ms | 6.31ms | **4.0x** |
| (32, 32, 32) | 229.25ms | 51.31ms | **4.5x** |

## Profiling Analysis

Profile of single chunk write benchmark showing where time is spent:

### Main Branch (977ms total)

| Function | Time | Calls | % of Total |
|----------|------|-------|------------|
| `decode_morton` (scalar) | 289ms | 32,768 | 30% |
| `get_chunk_slice` | 104ms | 32,768 | 11% |
| `_localize_chunk` | 103ms | 32,768 | 11% |
| `_morton_order` | 99ms | 1 | 10% |
| Generator expressions | 94ms | 262k | 10% |
| `all()` / `len()` | 87ms | 263k | 9% |

### Optimized Branch (456ms total)

| Function | Time | Calls | % of Total |
|----------|------|-------|------------|
| `get_chunk_slice` | 110ms | 32,768 | 24% |
| `_localize_chunk` | 105ms | 32,768 | 23% |
| `_morton_order` | 66ms | 1 | 14% |
| Generator expressions | 38ms | 131k | 8% |
[...]

PR #3708, opened 2026-02-13
perf: Optimize Morton order with hypercube and vectorization

mkitti:
## Summary

This PR optimizes the Morton order computation in `_morton_order` and `decode_morton` functions with multiple techniques that together provide **3-5x speedup** in Morton computation, resulting in measurable end-to-end improvements for sharded array indexing.

### 1. Hypercube Optimization
Calculate the largest power-of-2 hypercube that fits within `chunk_shape`. Within this hypercube, Morton codes are guaranteed to be in bounds, eliminating the need for bounds checking.

### 2. Vectorized Morton Decoding
Created `decode_morton_vectorized` to decode multiple Morton codes at once using NumPy array operations instead of scalar bit manipulation.

### 3. Efficient Bit Counting
Replaced `math.ceil(math.log2(c))` with `(c-1).bit_length()` for more efficient and accurate bit width calculation.

### 4. Singleton Dimension Removal
For shapes with singleton dimensions (size 1) like `(1,1,32,32,32)`, remove them before Morton computation, then expand coordinates back. This enables better optimization for the non-singleton dimensions.

### Morton Order Computation (micro-benchmark, no caching)

| Shape | Original | Optimized | Speedup |
|-------|----------|-----------|---------|
| `(8, 8)` | 0.27ms | 0.13ms | **2.0x** |
| `(32, 32)` | 5.21ms | 1.37ms | **3.8x** |
| `(8, 8, 8)` | 2.57ms | 0.76ms | **3.4x** |
| `(32, 32, 32)` | 220ms | 48ms | **4.6x** |
| `(16, 16, 16, 16)` | 447ms | 107ms | **4.2x** |
| `(1, 1, 32, 32, 32)` | 237ms | 71ms | **3.3x** |

### End-to-End Sharded Array Indexing

Benchmark with 32³ = 32,768 chunks per shard (`test_sharded_morton_indexing_large`):

| Branch | Mean Time | Morton Contribution |
|--------|-----------|---------------------|
| Main (original) | 7.07s | ~220ms |
| Optimized | 6.89s | ~50ms |
| **Improvement** | **171ms (2.4%)** | **~4x faster** |

The 2.4% end-to-end improvement directly correlates with the Morton order speedup. While I/O dominates total time, the Morton optimization provides measurable benefits for workloads with large chunks-per-shard counts.

<details>
<summary>Benchmark Script</summary>

```python
#!/usr/bin/env python
"""Detailed benchmark showing contribution of each Morton order optimization."""

import time
[...]

PR #3713, opened 2026-02-17
perf: Vectorize get_chunk_slice for faster sharded writes

mkitti:
## Summary

This PR adds vectorized methods to `_ShardIndex` and `_ShardReader` for batch chunk slice lookups, significantly reducing per-chunk function call overhead when writing to shards.

### New Methods

**`_ShardIndex.get_chunk_slices_vectorized`**: Batch lookup of chunk slices using NumPy vectorized operations instead of per-chunk Python calls.

**`_ShardReader.to_dict_vectorized`**: Build a chunk dictionary using vectorized lookup instead of iterating with individual `get()` calls.

### Modified Code Path

In `_encode_partial_single`, replaced:
```python
shard_dict = {k: shard_reader.get(k) for k in morton_order_iter(chunks_per_shard)}
```

With vectorized approach:
```python
morton_coords = _morton_order(chunks_per_shard)
chunk_coords_array = np.array(morton_coords, dtype=np.uint64)
shard_dict = shard_reader.to_dict_vectorized(chunk_coords_array, morton_coords)
```

### Single Chunk Write to Large Shard

Writing a single 1x1x1 chunk to a shard with 32³ chunks (using `test_sharded_morton_write_single_chunk` from PR #3712):

| Optimization | Time | Speedup vs Main |
|--------------|------|-----------------|
| Main branch (original) | 422ms | - |
| + Morton optimization (PR #3708) | 261ms | 1.6x |
| **+ Vectorized get_chunk_slice** | **95ms** | **4.4x** |

### Profile Breakdown

| Function | Before | After |
|----------|--------|-------|
| `get_chunk_slice` + `_localize_chunk` | 215ms | 3ms |
| `to_dict_vectorized` loop | 81ms | 9ms |
| Total function calls | 299k | 37k |

## Checklist

### DIFF
--- a/src/zarr/core/indexing.py
+++ b/src/zarr/core/indexing.py
@@ -4,7 +4,7 @@ import itertools
 import math
 import numbers
 import operator
-from collections.abc import Iterable, Iterator, Sequence
+from collections.abc import Iterator, Sequence
 from dataclasses import dataclass
 from enum import Enum
 from functools import lru_cache, reduce
@@ -1504,7 +1504,7 @@ def decode_morton_vectorized(
 
 
 @lru_cache(maxsize=16)
-def _morton_order(chunk_shape: tuple[int, ...]) -> Iterable[Sequence[int]]:
+def _morton_order(chunk_shape: tuple[int, ...]) -> npt.NDArray[np.intp]:
     n_total = product(chunk_shape)
     n_dims = len(chunk_shape)
     if n_total == 0:


mkitti:
@d-v-b add the benchmark tag here as well please.

src/zarr/codecs/sharding.py
```diff
@@ -505,7 +574,10 @@ async def _encode_partial_single(
             chunks_per_shard=chunks_per_shard,
         )
         shard_reader = shard_reader or _ShardReader.create_empty(chunks_per_shard)
-        shard_dict = {k: shard_reader.get(k) for k in morton_order_iter(chunks_per_shard)}
+        # Use vectorized lookup for better performance
+        morton_coords = _morton_order(chunks_per_shard)
+        chunk_coords_array = np.array(morton_coords, dtype=np.uint64)
+        shard_dict = shard_reader.to_dict_vectorized(chunk_coords_array, morton_coords)
```
d-v-b: why does `to_dict_vectorized` take the same values as 2 different types? Can't we pass in `morton_coords` and have the array version created internally as an implementation detail?

mkitti: This is probably vestigial after some other optimizations I was trying. 

Going back and forth from a list of tuples to numpy arrays is suboptimal. Do we really need to have a list of tuples at all?

These optimizations can be summarized as "use numpy". Is there a reason we are not using numpy more?

d-v-b: >  Is there a reason we are not using numpy more?

not specifically, but tuples of ints are very cheap to create, and they are hashable unlike numpy arrays,

mkitti: I do not have a strong issue with tuples of ints. I have an issue with a list of tuples of ints, especially when that list has a length of 32,768 or greater.

d-v-b: no objection to using numpy arrays where it improves performance!

mkitti:
## Benchmark results after latest commits

Compared against `main` with cache cleared before each iteration (`_morton_order` and `_morton_order_keys`):

| Benchmark | `main` | This PR | Speedup |
|---|---|---|---|
| `test_sharded_morton_write_single_chunk[(32,32,32)]` | 76,235 µs | 6,802 µs | **11.2×** |
| `test_morton_order_iter[(16,16,16)]` | 2,092 µs | < threshold | **>8×** |
| `test_morton_order_iter[(32,32,32)]` | 17,032 µs | < threshold | **>8×** |
| `test_sharded_morton_indexing[(16,16,16)]` | 21,925 µs | 21,788 µs | ~1× (unaffected) |
| `test_sharded_morton_indexing[(32,32,32)]` | 175,284 µs | 177,995 µs | ~1× (unaffected) |

**Profiling breakdown** (`test_sharded_morton_write_single_chunk`, 32³ shard, 32,768 chunks):

| Cost | `main` | This PR |
|---|---|---|
| `_morton_order` computation | 18ms (scalar loop × 32,768) | 1ms (vectorized) |
| `get_chunk_slice` / `_localize_chunk` | 67ms (32,768 Python calls) | 1ms (`get_chunk_slices_vectorized`) |
| Total function calls | 297,531 | 166,468 |

The 11× speedup comes from two changes: vectorizing `_morton_order` (eliminating 32,768 scalar `decode_morton` calls) and vectorizing the chunk slice lookup (eliminating 32,768 per-chunk `get_chunk_slice` + `_localize_chunk` calls). Read paths are unaffected.

Claude generated this report.

src/zarr/core/indexing.py
```diff
@@ -1504,37 +1504,33 @@ def decode_morton_vectorized(
 
 
 @lru_cache(maxsize=16)
-def _morton_order(chunk_shape: tuple[int, ...]) -> tuple[tuple[int, ...], ...]:
+def _morton_order(chunk_shape: tuple[int, ...]) -> Iterable[Sequence[int]]:
```
d-v-b: can we annotate the actual return type (a numpy array)?

d-v-b:
this looks great, just 1 nit about narrowing the return type of a function and I think this is good to go

mkitti:
We seem to be stuck due to codspeed running out of time.

> The usage of Macro Runners on zarr-developers exceeded the limit of 600 minutes.

>The current period ends on February 28, 2026.

This may need to be merged manually.

Also, I have mainly optimized the power-of-2 case. I am preparing additonal benchmarks beyond a power of 2.

PR #3715, opened 2026-02-18
sketch out sync codecs + threadpool

d-v-b:
This is a work in progress with all the heavy lifting done by claude. The goal is to improve the performance of our codecs by avoiding overhead in `to_thread` and other async machinery. At the moment we have deadlocks in some of the array tests, but I am opening this now as a draft to see if the benchmarks show anything promising.

### DIFF
--- a/src/zarr/core/indexing.py
+++ b/src/zarr/core/indexing.py
@@ -1512,54 +1512,48 @@ def _morton_order(chunk_shape: tuple[int, ...]) -> npt.NDArray[np.intp]:
         out.flags.writeable = False
         return out
 
-    # Optimization: Remove singleton dimensions to enable magic number usage
-    # for shapes like (1,1,32,32,32). Compute Morton on squeezed shape, then expand.
-    singleton_dims = tuple(i for i, s in enumerate(chunk_shape) if s == 1)
-    if singleton_dims:
-        squeezed_shape = tuple(s for s in chunk_shape if s != 1)
-        if squeezed_shape:
-            # Compute Morton order on squeezed shape, then expand singleton dims (always 0)
-            squeezed_order = np.asarray(_morton_order(squeezed_shape))
-            out = np.zeros((n_total, n_dims), dtype=np.intp)
-            squeezed_col = 0
-            for full_col in range(n_dims):
-                if chunk_shape[full_col] != 1:
-                    out[:, full_col] = squeezed_order[:, squeezed_col]
-                    squeezed_col += 1
-        else:
-            # All dimensions are singletons, just return the single point
-            out = np.zeros((1, n_dims), dtype=np.intp)
-        out.flags.writeable = False
-        return out
+    # Ceiling hypercube: smallest power-of-2 hypercube whose Morton codes span
+    # all valid coordinates in chunk_shape. (c-1).bit_length() gives the number
+    # of bits needed to index c values (0 for singleton dims). n_z = 2**total_bits
+    # is the size of this hypercube.
+    total_bits = sum((c - 1).bit_length() for c in chunk_shape)
+    n_z = 1 << total_bits if total_bits > 0 else 1
 
-    # Find the largest power-of-2 hypercube that fits within chunk_shape.
-    # Within this hypercube, Morton codes are guaranteed to be in bounds.
-    min_dim = min(chunk_shape)
-    if min_dim >= 1:
-        power = min_dim.bit_length() - 1  # floor(log2(min_dim))
-        hypercube_size = 1 << power  # 2^power
-        n_hypercube = hypercube_size**n_dims
+    # Decode all Morton codes in the ceiling hypercube, then filter to valid coords.
+    # This is fully vectorized. For shapes with n_z >> n_total (e.g. (33,33,33):
+    # n_z=262144, n_total=35937), consider the argsort strategy below.
+    order: npt.NDArray[np.intp]
+    if n_z <= 4 * n_total:
+        # Ceiling strategy: decode all n_z codes vectorized, filter in-bounds.
+        # Works well when the overgeneration ratio n_z/n_total is small (≤4).
+        z_values = np.arange(n_z, dtype=np.intp)
+        all_coords = decode_morton_vectorized(z_values, chunk_shape)
+        shape_arr = np.array(chunk_shape, dtype=np.intp)
+        valid_mask = np.all(all_coords < shape_arr, axis=1)
+        order = all_coords[valid_mask]
     else:
-        n_hypercube = 0
+        # Argsort strategy: enumerate all n_total valid coordinates directly,
+        # encode each to a Morton code, then sort by code. Avoids the 8x or
+        # larger overgeneration penalty for near-miss shapes like (33,33,33).
+        # Cost: O(n_total * bits) encode + O(n_total log n_total) sort,
+        # vs O(n_z * bits) = O(8 * n_total * bits) for ceiling.
+        grids = np.meshgrid(*[np.arange(c, dtype=np.intp) for c in chunk_shape], indexing="ij")
+        all_coords = np.stack([g.ravel() for g in grids], axis=1)
 
-    # Within the hypercube, no bounds checking needed - use vectorized decoding
-    if n_hypercube > 0:
-        z_values = np.arange(n_hypercube, dtype=np.intp)
-        order: npt.NDArray[np.intp] = decode_morton_vectorized(z_values, chunk_shape)
-    else:
-        order = np.empty((0, n_dims), dtype=np.intp)
+        # Encode all coordinates to Morton codes (vectorized).
+        bits_per_dim = tuple((c - 1).bit_length() for c in chunk_shape)
+        max_coord_bits = max(bits_per_dim)
+        z_codes = np.zeros(n_total, dtype=np.intp)
+        output_bit = 0
+        for coord_bit in range(max_coord_bits):
+            for dim in range(n_dims):
+                if coord_bit < bits_per_dim[dim]:
+                    z_codes |= ((all_coords[:, dim] >> coord_bit) & 1) << output_bit
+                    output_bit += 1
 
-    # For remaining elements outside the hypercube, bounds checking is needed
-    remaining: list[tuple[int, ...]] = []
-    i = n_hypercube
-    while len(order) + len(remaining) < n_total:
-        m = decode_morton(i, chunk_shape)
-        if all(x < y for x, y in zip(m, chunk_shape, strict=False)):
-            remaining.append(m)
-        i += 1
+        sort_idx: npt.NDArray[np.intp] = np.argsort(z_codes, kind="stable")
+        order = np.asarray(all_coords[sort_idx], dtype=np.intp)
 
-    if remaining:
-        order = np.vstack([order, np.array(remaining, dtype=np.intp)])
     order.flags.writeable = False
     return order
 


docs/design/sync-bypass.md
```diff
@@ -0,0 +1,228 @@
+# Design: Fully Synchronous Read/Write Bypass
```
d-v-b: @rabernat @dcherian have a look, this is claude's summary of the perf blockers addressed in this PR

d-v-b:
performance impact ranges from "good" to "amazing" so I think we want to learn from this PR. IMO this is NOT a merge candidate but rather should function as a proof-of-concept for what we can get if we rethink our current codec API.

Some key points:
- Wrapping CPU-bound routines like gzip encode / decode with async adds needless latency. We get a lot of perf by using a sync fast path whenever possible. We need to bake this "sync is faster when available" lesson into both our codec API and store API. For example, there is no reason that reading or writing to an in-memor dict should be async. 
- We should design the chunk encoding process so that IO bound and CPU-bound routines are logically separated in the codebase. That means modelling sharding as a codec is probably wrong. Sharding is declared as a codec in array metadata, but we don't need to model it as a codec internally. Sharding changes _how_ we do IO, but it should not change _when_ we do IO. 
- I haven't looked at memory use at all. that's probably a separate effort.

d-v-b:
the current performance improvements are without any parallelism. I'm adding that now.

d-v-b:
the latest commit adds thread-based parallelism to the synchronous codec pipeline. we compute an estimated compute cost based on the chunk size, codecs, and operation (encode / code), and use that estimate to choose a parellelism strategy, ranging from no threads to full use of a thread pool.

d-v-b:
marking this as not a draft, because I think we should actually consider merging it.

d-v-b:
i added a changelog entry and made a _breaking change_: removal of the `batch_size` parameter from the `BatchedCodecPipeline`. The batch size was already limited by the concurrency limit, and the parallelism model offered by `batch_size` (applying codec X over `batch_size` chunks at once) doesn't really make performance  sense versus parallelism across chunks.

dcherian:
This is _extremely_ hard to review at the moment. Can we look at a new PR with just one affected codec (Zstd?) please?

d-v-b:
the changes here aren't really made at the granularity of a single codec. We have new codec pipeline behavior, which requires new methods on stores AND codecs. When the codec pipeline identifies that all the codecs AND the store support the fast path, then it uses the fast path. So breaking that apart is difficult.

src/zarr/experimental/sync_codecs.py
```diff
@@ -0,0 +1,34 @@
+"""Backwards-compatible alias for SyncCodecPipeline.
```
dcherian: delete?

d-v-b: yep this looks like garbage

src/zarr/core/config.py
```diff
@@ -99,11 +99,13 @@ def enable_gpu(self) -> ConfigSet:
...
             "async": {"concurrency": 10, "timeout": None},
-            "threading": {"max_workers": None},
+            "threading": {
+                "max_workers": None,
+                "codec_workers": {"enabled": True, "min": 0, "max": None},
+            },
             "json_indent": 2,
             "codec_pipeline": {
                 "path": "zarr.core.codec_pipeline.BatchedCodecPipeline",
-                "batch_size": 1,
```
dcherian, line 106: do we error if someone sets it? Would be good to ignore instead

d-v-b: good idea, we should probably remove this from the config and warn if someone tries to set it

src/zarr/codecs/gzip.py
```diff
@@ -48,23 +49,37 @@ def from_dict(cls, data: dict[str, JSON]) -> Self:
...
+
     async def _decode_single(
         self,
         chunk_bytes: Buffer,
         chunk_spec: ArraySpec,
     ) -> Buffer:
-        return await asyncio.to_thread(
-            as_numpy_array_wrapper, GZip(self.level).decode, chunk_bytes, chunk_spec.prototype
-        )
+        return await asyncio.to_thread(self._decode_sync, chunk_bytes, chunk_spec)
```
dcherian, line 75: as an aside, these `to_thread` calls are extremely annoying; they run on an independent thread pool not the one Zarr sets up (and are thus unconstrained by any config setting).

instead we need something like this:
https://github.com/earth-mover/xpublish-tiles/blob/1a800e05617d609098bbcd1a1f5ac9bbdcb531aa/src/xpublish_tiles/lib.py#L147-L152

src/zarr/codecs/crc32c_.py
```diff
@@ -64,5 +56,19 @@ async def _encode_single(
         # Append the checksum (as bytes) to the data
         return chunk_spec.prototype.buffer.from_array_like(np.append(data, checksum.view("B")))
 
+    async def _decode_single(
+        self,
+        chunk_bytes: Buffer,
+        chunk_spec: ArraySpec,
+    ) -> Buffer:
+        return self._decode_sync(chunk_bytes, chunk_spec)
```
dcherian, line 64: `to_thread` these just like gzip? And then of course, we can fix all of them later? With this, we end up blocking the event loop.

src/zarr/codecs/gzip.py
d-v-b: yes `to_thread` has serious problems: https://github.com/python/cpython/issues/136084. I will drop in your async_run idea!

tests/package_with_entrypoint/__init__.py
```diff
@@ -40,7 +40,7 @@ def compute_encoded_size(self, input_byte_length: int, chunk_spec: ArraySpec) ->
 
 
 class TestEntrypointCodecPipeline(CodecPipeline):
-    def __init__(self, batch_size: int = 1) -> None:
```
dcherian, line 43: we should ignore and raise warning instead for backcompat

src/zarr/codecs/crc32c_.py
d-v-b: the crc32c computation is not going to run long enough to block the event loop

src/zarr/experimental/sync_codecs.py
d-v-b: now deleted

src/zarr/codecs/crc32c_.py
dcherian: OK if it is indeed that fast. Also see https://github.com/awesomized/crc-fast-rust ;)

src/zarr/core/config.py
d-v-b: ok batch size is back on the batchedcodecpipeline, it doesn't do anything and emits a warning if someone tries to set it. we can remove it completely later.

tests/package_with_entrypoint/__init__.py
d-v-b: this is just an entrypoint test but I made the backwards-compat change over in the batchedcodecpipeline class

src/zarr/core/codec_pipeline.py
```diff
@@ -68,62 +63,182 @@ def fill_value_or_default(chunk_spec: ArraySpec) -> Any:
...
+    "Crc32cCodec": 0,
+    "TransposeCodec": 0,
+    "VLenUTF8Codec": 0,
+    "VLenBytesCodec": 0,
+    # Medium cost — fast C codecs, GIL released
+    "ZstdCodec": 3,
+    "BloscCodec": 2,
+    # High cost — slower C codecs, GIL released
+    "GzipCodec": 50,
+}
```
d-v-b: @dcherian here's the estimated cost of running each codec in the encode and decode path

src/zarr/core/codec_pipeline.py
```diff
@@ -68,62 +63,198 @@ def fill_value_or_default(chunk_spec: ArraySpec) -> Any:
...
+
+
+def _choose_workers(
+    n_chunks: int,
+    chunk_nbytes: int,
+    codecs: Iterable[Codec],
+    *,
+    is_encode: bool = False,
+) -> int:
+    """Decide how many thread pool workers to use (0 = don't use pool).
```
dcherian, line 88: This doesn't seem worth it to me. Why not just spin up a thread pool with maxworkers, and either submit to it or not depending on the cost model?

src/zarr/core/codec_pipeline.py
```diff
@@ -68,62 +63,198 @@ def fill_value_or_default(chunk_spec: ArraySpec) -> Any:
...
         return cls(
             array_array_codecs=array_array_codecs,
             array_bytes_codec=array_bytes_codec,
             bytes_bytes_codecs=bytes_bytes_codecs,
-            batch_size=batch_size or config.get("codec_pipeline.batch_size"),
         )
 
     @property
     def supports_partial_decode(self) -> bool:
-        """Determines whether the codec pipeline supports partial decoding.
```
dcherian, line 101: ?

src/zarr/core/codec_pipeline.py
```diff
@@ -304,7 +637,6 @@ def _merge_chunk_array(
         if (
             is_complete_chunk
             and value.shape == chunk_spec.shape
-            # Guard that this is not a partial chunk at the end with is_complete_chunk=True
```
dcherian, line 307: Claude's propensity for deleting useful comments is extremely frustrating!

src/zarr/core/codec_pipeline.py
d-v-b: no reason for this to be removed. restoring it.

src/zarr/core/codec_pipeline.py
d-v-b: I am scolding claude about this

src/zarr/core/codec_pipeline.py
```diff
@@ -68,62 +63,198 @@ def fill_value_or_default(chunk_spec: ArraySpec) -> Any:
...
+
+_CODEC_ENCODE_NS_PER_BYTE: dict[str, float] = {
+    # Near-zero cost — just reshaping/copying/checksumming
+    "BytesCodec": 0,
+    "Crc32cCodec": 0,
+    "TransposeCodec": 0,
+    "VLenUTF8Codec": 0,
+    "VLenBytesCodec": 0,
+    # Medium cost — fast C codecs, GIL released
+    "ZstdCodec": 3,
```
mkitti: Can we adjust by the compression level.? Compression level level -1000 is different compression level 22 in terms of time.

d-v-b: yes we could put this in the model. we would have to take some data first of course

src/zarr/core/codec_pipeline.py
d-v-b: those comments / docstrings are restored

src/zarr/core/codec_pipeline.py
d-v-b: good point, I simplified this in https://github.com/zarr-developers/zarr-python/pull/3715/commits/b40d53a78e1bba51a0f7d87071e140803a7054c6

src/zarr/core/codec_pipeline.py
```diff
@@ -162,30 +103,23 @@ def _choose_workers(
     if n_chunks < 2:
         return min_workers
 
-    per_chunk_ns = _estimate_chunk_work_ns(chunk_nbytes, codecs, is_encode=is_encode)
-
-    if per_chunk_ns < _POOL_OVERHEAD_NS and min_workers == 0:
+    # Only use the pool when at least one codec does real work
+    # and the chunks are large enough to offset dispatch overhead.
+    has_expensive = any(type(c).__name__ not in _CHEAP_CODECS for c in codecs)
```
dcherian: uhh... can we not `isinstance` because of cyclic imports or something?

src/zarr/core/codec_pipeline.py
```diff
@@ -64,87 +64,28 @@ def fill_value_or_default(chunk_spec: ArraySpec) -> Any:
...
+        "VLenBytesCodec",
+    }
+)
+
+# Minimum chunk size (in bytes) to consider using the thread pool.
+# Below this, per-chunk codec work is too small to offset dispatch overhead.
+_MIN_CHUNK_NBYTES_FOR_POOL = 100_000  # 100 KB
+
+
+def _choose_workers(n_chunks: int, chunk_nbytes: int, codecs: Iterable[Codec]) -> int:
```
dcherian, line 87: Can this be `def _use_thread_pool(...)->bool` instead?

src/zarr/core/codec_pipeline.py
```diff
@@ -162,30 +103,23 @@ def _choose_workers(
...
-    target_per_worker_ns = 1_000_000  # 1ms
-    workers = max(1, int(total_work_ns / target_per_worker_ns))
+    return max(min_workers, min(n_chunks, max_workers))
 
-    return max(min_workers, min(workers, n_chunks, max_workers))
 
-
-def _get_pool(max_workers: int) -> ThreadPoolExecutor:
-    """Get a thread pool with at most *max_workers* threads."""
+def _get_pool() -> ThreadPoolExecutor:
```
dcherian, line 113: hard to see why this had to change but... i"m not opposed to it.

src/zarr/core/codec_pipeline.py
```diff
@@ -68,32 +63,122 @@ def fill_value_or_default(chunk_spec: ArraySpec) -> Any:
...
+        return 0
+
+    return max(min_workers, min(n_chunks, max_workers))
+
+
+def _get_pool() -> ThreadPoolExecutor:
+    """Get the module-level thread pool, creating it lazily."""
+    global _pool
+    if _pool is None:
+        max_workers: int = config.get("threading.codec_workers").get("max") or os.cpu_count() or 4
```
dcherian: duplicated in `_choose_workers`, doesn't `donfig` have a way to do runtime defaults?

src/zarr/core/codec_pipeline.py
d-v-b: yeah we can. i switched to `isinstance` catching any bytesbytescodec

src/zarr/core/codec_pipeline.py
d-v-b: a user can still unset the param in the config even if we have a default, so we need to handle unsetness. i consolidated this logic into a single function so at least the duplication is gone.

d-v-b:
The changes here improve performance a lot, but I think we can do even better with a more comprehensive set of changes. I had claude cook up a planning document based on zarrs and tensorstore here: https://hackmd.io/A933wEUwQjOx8rJmmWo13A. Please review that document. I will use this plan to guide the next round of performance improvements. Not sure if they will be in this PR or a subsequent one.

tests/test_config.py
```diff
@@ -132,7 +133,7 @@ def test_config_codec_pipeline_class(store: Store) -> None:
     # has default value
     assert get_pipeline_class().__name__ != ""
 
-    config.set({"codec_pipeline.name": "zarr.core.codec_pipeline.BatchedCodecPipeline"})
+    config.set({"codec_pipeline.path": "zarr.core.codec_pipeline.BatchedCodecPipeline"})
```
dcherian, line 136: what's this about?

src/zarr/storage/_local.py
```diff
@@ -187,6 +187,69 @@ def __repr__(self) -> str:
...
+    # LocalStore's async get/set wrap the synchronous helpers _get() and
+    # _put() (defined at module level) in asyncio.to_thread(). These sync
+    # methods call _get/_put directly, removing the thread-hop overhead.
+    #
+    # The open-guard logic is inlined from _open(): create root dir if
+    # writable, check existence, set _is_open. We can't call the async
+    # _open() from a sync context, so we replicate its logic here.
+    # -------------------------------------------------------------------
+
+    def get_sync(
```
dcherian, line 202: are we able to share sync/async code paths at all?

src/zarr/core/codec_pipeline.py
```diff
@@ -68,32 +63,118 @@ def fill_value_or_default(chunk_spec: ArraySpec) -> Any:
         return fill_value
 
 
+# ---------------------------------------------------------------------------
+# Thread pool for parallel codec compute
+# ---------------------------------------------------------------------------
+
+# Minimum chunk size (in bytes) to consider using the thread pool.
+# Below this, per-chunk codec work is too small to offset dispatch overhead.
+_MIN_CHUNK_NBYTES_FOR_POOL = 100_000  # 100 KB
```
dcherian, line 72: let's make this a config, so its easy to experiment with

src/zarr/core/codec_pipeline.py
```diff
@@ -68,32 +63,118 @@ def fill_value_or_default(chunk_spec: ArraySpec) -> Any:
...
+
+
+_pool: ThreadPoolExecutor | None = None
+
+# Sentinel to distinguish "delete this key" from None.
+_DELETED = object()
+
+
 @dataclass(frozen=True)
 class BatchedCodecPipeline(CodecPipeline):
```
dcherian, line 129:
```suggestion
class DefaultCodecPipeline(CodecPipeline):
```

or is the name needed for backcompat

src/zarr/core/codec_pipeline.py
```diff
@@ -357,79 +642,82 @@ async def write_batch(
...
                     ],
                 )
+            return
+
+        if self._all_sync:
+            # Streaming per-chunk pipeline: each chunk flows through
+            # read_existing → decode → merge → encode → write as a single
+            # task. Running N tasks concurrently overlaps IO with compute.
+            async def _write_chunk(
+                byte_setter: ByteSetter,
```
dcherian, line 652: why is there an `async` func under `self._all_sync`? Seems like a naming issue that is very confusing to me right now

src/zarr/codecs/gzip.py
TomAugspurger: Does zarr create & own the event loop this is running? If so, we can set the default executor: https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.set_default_executor

src/zarr/core/codec_pipeline.py
d-v-b: i think we want to keep the name for compatibility

src/zarr/core/codec_pipeline.py
d-v-b: we send this function into `concurrent_map`, which is why it needs to be async

d-v-b:
superseded by https://github.com/zarr-developers/zarr-python/pull/3885

Issue #3720, opened 2026-02-25
chunk encoding performance improvements


d-v-b:
the two PRs #3715 and #3719 demonstrate that we can attain substantial performance improvements and completely new functionality by applying a series of changes to stores, codecs, and the codec pipeline logic. We get the performance via the following changes:

- Avoid async overhead when doing IO against low-latency stores
- Avoid overhead related to creating unnecessary python objects per-chunk
- Formally separate IO from compute in chunk encoding / decoding
- Use range writes when applicable to write individual subchunks inside a shard.

The two PRs that demonstrate these changes are far too large to merge on their own, and reviewing is burdened with off-target changes made inadvertantly by claude.

I think the performance wins are too great to pass up, and the cost is manageable if we break the changes into smaller PRs. This issue tracks the progress of these changes, and can serve as a discussion site as needed. 

Here's an outline of the PRs I'd like to open shortly:

- [ ] PR1  SupportsSyncCodec + _encode_sync/_decode_sync on all codecs
- [ ] PR2  CodecChain (pure-compute codec wrapper)
- [ ] PR3  Store layer: set_range + sync methods
- [ ] PR4  PreparedWrite + prepare/finalize pattern on ArrayBytesCodec
- [ ] PR5  BatchedCodecPipeline refactor (uses CodecChain, read_sync/write_sync)
- [ ] PR6  ShardingCodec refactor (prepare_write, finalize_write, set_range, dense layout)
- [ ] PR7  Vectorized shard encoding (_encode_vectorized, _encode_vectorized_sparse)
- [ ] PR8  Sync fast path in Array (opt-in config flag)

I will update this issue when I actually open these PRs.

PR #3721, opened 2026-02-25
perf: add sync methods to codecs

d-v-b:
This PR adds synchronous encode / decode methods to our codecs, and a protocol that models this behavior. These methods will be used in future improvements to the chunk encoding process. Async is a pure performance tax for CPU-limited operations like compression. These methods allow us to avoid that async overhead.

Part of #3720 

TODO:


src/zarr/codecs/zstd.py
```diff
@@ -38,7 +38,7 @@ def parse_checksum(data: JSON) -> bool:
 class ZstdCodec(BytesBytesCodec):
     """zstd codec"""
 
-    is_fixed_size = True
+    is_fixed_size = False
```
dcherian, line 41: what's this?

d-v-b: fixing a bug in the zstd codec. It's not fixed size.

src/zarr/abc/codec.py
```diff
@@ -59,6 +60,23 @@ def _check_codecjson_v2(data: object) -> TypeGuard[CodecJSON_V2[str]]:
...
+class SupportsSyncCodec(Protocol):
+    """Protocol for codecs that support synchronous encode/decode.
+
+    Codecs implementing this protocol provide ``_decode_sync`` and ``_encode_sync``
+    methods that perform encoding/decoding without requiring an async event loop.
+    """
+
+    def _decode_sync(
+        self, chunk_data: NDBuffer | Buffer, chunk_spec: ArraySpec
+    ) -> NDBuffer | Buffer: ...
```
d-v-b, line 73: this reminds me that it makes no sense for our bytes-bytes codecs to take an arrayspec parameter for encoding / decoding. we need to fix that down the road.

PR #3719, opened 2026-02-25
sketch out improved performance by refactoring codec pipeline logic

d-v-b:
This builds on top of #3715 and achieves even more perf improvements by refactoring the basic logic of codec encoding / decoding. The design document behind these changes is [here](https://hackmd.io/@d-v-b/ryKDvpFdbl). 

_I do not think this is merge-worthy_, as it's far too big. But i'm going to post the performance gains, and start figuring out how to break this into pieces.

A big feature this adds is the ability to write individual subchunks to uncompressed shards on storage backends that support range writes (local and memory).

# Benchmark comparison: perf/smarter-codecs vs main

| Test Name | perf/smarter-codecs (ms) | main (ms) | Speedup |
|-----------|-------:|--------:|--------:|
| test_write_array[memory-chunks=100,shards=1M-None] | 46.85 | 1018.60 | **21.74×** |
| test_write_array[local-chunks=100,shards=1M-None] | 68.45 | 1006.27 | **14.70×** |
| test_sharded_morton_indexing[(32,32,32)] | 22.45 | 247.49 | **11.02×** |
| test_slice_indexing[None-(0,0,0)] | 0.03 | 0.28 | **10.05×** |
| test_sharded_morton_indexing_large[(33,33,33)] | 254.85 | 2521.32 | **9.89×** |
| test_slice_indexing[(50,50,50)-full_slice] | 9.12 | 89.80 | **9.85×** |
| test_sharded_morton_indexing_large[(32,32,32)] | 226.80 | 2153.47 | **9.50×** |
| test_sharded_morton_indexing_large[(30,30,30)] | 181.86 | 1725.29 | **9.49×** |
| test_slice_indexing[(50,50,50)-(0,0,0)] | 0.06 | 0.60 | **9.45×** |
| test_write_array[memory-chunks=100,shards=1M-gzip] | 211.85 | 1978.29 | **9.34×** |
| test_slice_indexing[None-(slice(None,10,None))*3] | 0.03 | 0.28 | **9.28×** |
| test_write_array[local-chunks=100,shards=1M-gzip] | 217.00 | 1965.37 | **9.06×** |
| test_slice_indexing[(50,50,50)-strided_4] | 8.96 | 80.44 | **8.98×** |
| test_slice_indexing[(50,50,50)-strided_4_offset] | 4.83 | 43.21 | **8.95×** |
| test_sharded_morton_indexing[(16,16,16)] | 2.78 | 23.26 | **8.37×** |
[...]

PR #3722, opened 2026-02-26
perf/chunktransform

d-v-b:
This PR defines a `CodecChain` object, which is similar to `CodecPipeline` except that it performs no orchestration. It just defines array-spec-aware, pure-compute encoding and decoding routines for collections of codecs. Part of #3720 

depends on #3721

edit: `CodecChain` is now called `ChunkTransform`


d-v-b:
updating this PR to blend the codecchain logic with array_spec logic to define a `ChunkTransform` object that carries both together.

d-v-b:
this PR is necessary for the broader performance plan because it allows us to pipeline codecs that support synchronous execution (i.e., all of them, practically). We can use the `ChunkTransform` class to "fuse" the non-IO part of chunk encoding into a single callable. 

this differs from the `CodecPipeline` because that data structure does orchestration and IO, which is a separate layer of complexity.

d-v-b:
@TomAugspurger any interest in reviewing this PR? It's part of a series that culminate in the performance improvements evident in https://github.com/zarr-developers/zarr-python/pull/3719

src/zarr/core/codec_pipeline.py
```diff
@@ -69,6 +70,115 @@ def fill_value_or_default(chunk_spec: ArraySpec) -> Any:
...
+    """A synchronous codec chain bound to an ArraySpec.
+
+    Provides ``encode_chunk`` and ``decode_chunk`` for pure-compute
+    codec operations (no IO, no threading, no batching).
+
+    ``shape`` and ``dtype`` reflect the representation **after** all
+    ArrayArrayCodec transforms — i.e. the spec that feeds the
+    ArrayBytesCodec.
+
+    All codecs must implement ``SupportsSyncCodec``. Construction will
```
TomAugspurger: Idle thought without having yet looked at the implementation: make `Codec` generic over `SupportsSyncCodec` (via a protocol) so that this can be caught before runtime?

src/zarr/core/codec_pipeline.py
```diff
@@ -69,6 +70,115 @@ def fill_value_or_default(chunk_spec: ArraySpec) -> Any:
...
+            aa_codecs = (*aa_codecs, (aa_codec, spec))
+            spec = aa_codec.resolve_metadata(spec)
+
+        self._aa_codecs = aa_codecs
+        self._ab_codec = ab
+        self._ab_spec = spec
+        self._bb_codecs = bb
+
+    @property
+    def shape(self) -> tuple[int, ...]:
```
TomAugspurger: How are `shape` and `dtype` ultimately used? They're a bit complicated to understand. Presumably you need them for your full perf PR, but I wanted to confirm that.

src/zarr/core/codec_pipeline.py
```diff
@@ -69,6 +70,115 @@ def fill_value_or_default(chunk_spec: ArraySpec) -> Any:
...
+        self,
+        chunk_bytes: Buffer,
+    ) -> NDBuffer:
+        """Decode a single chunk through the full codec chain, synchronously.
+
+        Pure compute -- no IO.
+        """
+        bb_out: Any = chunk_bytes
+        for bb_codec in reversed(self._bb_codecs):
+            bb_out = bb_codec._decode_sync(bb_out, self._ab_spec)  # type: ignore[attr-defined]
```
TomAugspurger: Can you add a comment about why the `type: ignore` is needed? Presumably it relates to that `isinstance(c, SupportsSyncCodec)` check  above, which mypy can't see here in `decode`?

src/zarr/core/codec_pipeline.py
```diff
@@ -69,6 +70,115 @@ def fill_value_or_default(chunk_spec: ArraySpec) -> Any:
...
+
+        ab_out: Any = self._ab_codec._decode_sync(bb_out, self._ab_spec)  # type: ignore[attr-defined]
+
+        for aa_codec, spec in reversed(self._aa_codecs):
+            ab_out = aa_codec._decode_sync(ab_out, spec)  # type: ignore[attr-defined]
+
+        return ab_out  # type: ignore[no-any-return]
```
TomAugspurger: This `type: ignore` I don't understand. Is the `Any` up above accurate, or should this be `NDBuffer | Buffer` like I see [here](https://github.com/zarr-developers/zarr-python/blob/5d92e8539f0211f7ca301f7330b5a59bc2f84637/src/zarr/abc/codec.py#L79-L81)? And if it supposed to be `NDBuffer | Buffer` why do we declare we return `-> NDBuffer`

tests/test_sync_codec_pipeline.py
```diff
@@ -0,0 +1,222 @@
...
+            codecs=(TransposeCodec(order=(1, 0)), BytesCodec(), ZstdCodec()), array_spec=spec
+        )
+
+    def test_encode_decode_roundtrip_bytes_only(self) -> None:
+        # Minimal round-trip: BytesCodec serializes the array to bytes and back.
+        # No compression, no AA transform.
+        arr = np.arange(100, dtype="float64")
+        spec = _make_array_spec(arr.shape, arr.dtype)
+        chain = ChunkTransform(codecs=(BytesCodec(),), array_spec=spec)
+        nd_buf = _make_nd_buffer(arr)
```
TomAugspurger: FWIW, I'd be fine with consolidating these tests with the construction tests. I do like having focused construction tests when the constructors are complicated, but these seem simple enough that just seeing the traceback pointing to `__post_init__` should be enough. Either works for me though.

tests/test_sync_codec_pipeline.py
```diff
@@ -0,0 +1,222 @@
...
+
+        encoded = chain.encode(nd_buf)
+        assert encoded is not None
+        decoded = chain.decode(encoded)
+        np.testing.assert_array_equal(arr, decoded.as_numpy_array())
```
TomAugspurger: Might be worth refactoring this to a test helper, `assert_encode_decode_equal(...)`.

Also, not worth worrying about currently, `arr` is *possibly* not a NumPy array, depending on what `default_buffer_prototype()` returns, in which case `np.testing.assert_array_equal` might not work. But to solve that more generally is out of scope.

tests/test_sync_codec_pipeline.py
```diff
@@ -0,0 +1,222 @@
...
+            async def _decode_single(self, chunk_bytes: Buffer, chunk_spec: ArraySpec) -> NDBuffer:
+                raise NotImplementedError  # pragma: no cover
+
+            async def _encode_single(
+                self, chunk_array: NDBuffer, chunk_spec: ArraySpec
+            ) -> Buffer | None:
+                raise NotImplementedError  # pragma: no cover
+
+            def compute_encoded_size(self, input_byte_length: int, _chunk_spec: ArraySpec) -> int:
+                return input_byte_length  # pragma: no cover
```
TomAugspurger: Move to the top of this module, so you can reuse it in `mixed_sync_and_non_sync`?

TomAugspurger (approved):
Big +1 to the general design and the implementation looks nice. Just minor nitpicks that can be addressed / ignored as you prefer.

src/zarr/core/codec_pipeline.py
d-v-b: I'm on the fence about these attributes actually. I wanted to model the fact that the output of a `ChunkTransform` is an array, with a fixed shape and dtype, and the fact that the array -> array codecs can be thought of as layers of `ChunkTransform` objects. But I don't know if we actually need this. Maybe a richer return type annotation is a better way of conveying this information.

src/zarr/core/codec_pipeline.py
```diff
@@ -69,6 +70,115 @@ def fill_value_or_default(chunk_spec: ArraySpec) -> Any:
...
+        aa_out: Any = chunk_array
+
+        for aa_codec, spec in self._aa_codecs:
+            if aa_out is None:
+                return None
+            aa_out = aa_codec._encode_sync(aa_out, spec)  # type: ignore[attr-defined]
+
+        if aa_out is None:
+            return None
+        bb_out: Any = self._ab_codec._encode_sync(aa_out, self._ab_spec)  # type: ignore[attr-defined]
```
dcherian: This is quite hard to read/review: "is the output of `aa` after application of `ab` really `bb`?!" Doesn't help that output is Any despite all the fancy typing.

src/zarr/core/codec_pipeline.py
```diff
@@ -69,6 +70,115 @@ def fill_value_or_default(chunk_spec: ArraySpec) -> Any:
...
+
+        for bb_codec in self._bb_codecs:
+            if bb_out is None:
+                return None
```
dcherian: could be hoisted out of the loop

src/zarr/core/codec_pipeline.py
d-v-b: what do you suggest

src/zarr/core/codec_pipeline.py
```diff
@@ -69,6 +70,115 @@ def fill_value_or_default(chunk_spec: ArraySpec) -> Any:
...
+    ) -> Buffer | None:
+        """Encode a single chunk through the full codec chain, synchronously.
+
+        Pure compute -- no IO.
+        """
+        aa_out: Any = chunk_array
+
+        for aa_codec, spec in self._aa_codecs:
+            if aa_out is None:
+                return None
```
dcherian: could be moved out of the loop? Can `_encode_sync` return None?

src/zarr/core/codec_pipeline.py
dcherian: `aa_out` -> `asarray`; `bb_out` -> `asbytes`?

src/zarr/core/codec_pipeline.py
dcherian: or maybe not? I'm confused regardless.

src/zarr/core/codec_pipeline.py
d-v-b: ok so it's an issue with the variable names, I will see if I can make them more clear

dcherian: the properties are an issue too but I don't have a suggestion for that.

src/zarr/core/codec_pipeline.py
d-v-b: > Can _encode_sync return None?

unfortunately yes, in keeping with methods already defined on the `Codec` API.

src/zarr/core/codec_pipeline.py
d-v-b: [points accusingly at the v3 spec] making 3 different kinds of functions all "codecs" was maybe not the best choice. We could use the "filters, serializer, compressors" trinity used in `create_array`, but I worry that this is a bit too far from the language used by the spec.

d-v-b: I'm going to punt on this for now. I legit think the naming issues here are a basic flaw in the v3 spec.

tests/test_sync_codec_pipeline.py
d-v-b: good idea, done

tests/test_sync_codec_pipeline.py
d-v-b: good idea. I kept the encoding / decoding separate from the constructor tests, but I also made the constructor tests much more compact.

src/zarr/core/codec_pipeline.py
d-v-b: this was because the protocol for synchronous encoding / decoding was not generic over input and output types. I fixed that, so the type: ignore statement is gone

src/zarr/core/codec_pipeline.py
d-v-b: the type: ignore is gone thanks to making the protocol generic.

src/zarr/core/codec_pipeline.py
d-v-b: these attributes are gone, we can add them later if they are actually valuable

src/zarr/core/codec_pipeline.py
d-v-b: I want to avoid touching the codec ABC, so maybe we defer this for a later effort

src/zarr/core/codec_pipeline.py
d-v-b: the whole thing with encoding potentially returning none is problematic and needs to go. Unfortunately that's based on the codec ABC which I am not touching right now :(

d-v-b:
thanks for the super helpful feedback @TomAugspurger and @dcherian! I made a lot of simplifying changes. LMK if anything else needs to be done, otherwise I will merge and move on to the next step of the performance work.

PR #3725, opened 2026-02-26
perf/store sync

d-v-b:
This PR adds synchronous methods to stores that benefit from executing IO outside of an event loop, i.e. memory storage and local storage. These methods are not defined on the ABC, and are instead defined as protocols that stores can opt into. 

These methods are needed to unlock a hot path for chunk encoding / decoding that avoids event loop and `to_thread` overhead.

part of #3720 

builds on #3722 and #3721


src/zarr/codecs/blosc.py
```diff
@@ -318,8 +318,6 @@ def _encode_sync(
         chunk_bytes: Buffer,
         chunk_spec: ArraySpec,
     ) -> Buffer | None:
-        # Since blosc only support host memory, we convert the input and output of the encoding
```
dcherian, line 321: seems like this shouldn't be deleted

tests/test_codecs/test_zstd.py
```diff
@@ -23,3 +27,31 @@ def test_zstd(store: Store, checksum: bool) -> None:
...
     assert np.array_equal(data, a[:, :])
+
+
+def test_zstd_codec_supports_sync() -> None:
+    assert isinstance(ZstdCodec(), SupportsSyncCodec)
+
+
+def test_zstd_is_not_fixed_size() -> None:
+    assert ZstdCodec.is_fixed_size is False
+
```
dcherian: these seem useless

src/zarr/core/codec_pipeline.py
```diff
@@ -68,6 +69,106 @@ def fill_value_or_default(chunk_spec: ArraySpec) -> Any:
         return fill_value
 
 
+@dataclass(frozen=True, slots=True)
```
dcherian: Sorry but this is straight up unreviewable. I cant tell if there's a one character typo that breaks things here.

src/zarr/core/codec_pipeline.py
```diff
@@ -68,6 +69,106 @@ def fill_value_or_default(chunk_spec: ArraySpec) -> Any:
...
+        aa_out: Any = chunk_array
+
+        for aa_codec, spec in zip(self._aa_codecs, self._aa_specs, strict=True):
+            if aa_out is None:
+                return None
+            aa_out = cast("SupportsSyncCodec", aa_codec)._encode_sync(aa_out, spec)
+
+        if aa_out is None:
+            return None
+        bb_out: Any = cast("SupportsSyncCodec", self._ab_codec)._encode_sync(aa_out, self._ab_spec)
```
dcherian: 😵‍💫

src/zarr/storage/_local.py
```diff
@@ -187,6 +187,56 @@ def __repr__(self) -> str:
     def __eq__(self, other: object) -> bool:
         return isinstance(other, type(self)) and self.root == other.root
 
+    # -------------------------------------------------------------------
+    # Synchronous store methods
+    # -------------------------------------------------------------------
+
+    def _ensure_open_sync(self) -> None:
```
dcherian, line 194: is this necessary?

src/zarr/core/codec_pipeline.py
d-v-b: Yeah this is ugly but to set expectations, I don't think we are going to win a beauty contests with any of the stuff we have to bolt on to our codec, store, and chunk encoding routines. We are starting from a performance-unfriendly design, and trying to fix that design from the outside. In any case, this class will appear in a later PR -- I'm removing it from this one.

src/zarr/core/codec_pipeline.py
d-v-b: good catch, this got included by accident -- it's from a separate PR that needs to happen after this one.

src/zarr/storage/_local.py
d-v-b: yes. in our current `Store` design, creating an instance of a store doesn't necessarily "open" it, so we have an [async `open` method](https://github.com/zarr-developers/zarr-python/blob/b6d3ae2e184bdd5078baa991ab21f270f782fb90/src/zarr/abc/store.py#L64) that actually opens the store. Our async `get/set` methods guard against the store being un-open: https://github.com/zarr-developers/zarr-python/blob/b6d3ae2e184bdd5078baa991ab21f270f782fb90/src/zarr/storage/_local.py#L199-L200, and our sync methods have to do the same thing here.

tests/test_codecs/test_zstd.py
d-v-b: agreed, looks like claude didn't have faith in our ability to avoid making the same mistake twice

d-v-b: removed it

src/zarr/codecs/blosc.py
d-v-b: good catch, added it back

PR #3727, opened 2026-02-27
perf/prepared write

d-v-b:
- adds methods and data structures for modelling the IO phase before and after chunk encoding

depends on #3725 

part of #3720


d-v-b:
superseded by #3885

PR #3737, opened 2026-03-01
Feature/chunk grid array shape

### LINKED ISSUE #3736: chunk grid object should track array shape
## summary

With the rectilinear chunk grid incoming, we need array shape information on the internal chunk grid object. So I propose adding it.

## background

Our `ChunkGrid` class pretty narrowly models the `chunk_grid` field in zarr v3 array metadata. This means the `ChunkGrid` does _not_ have any information about the shape of the array. In array metadata that information is stored in the top-level `shape` attribute. 

For the `RegularChunkGrid` this wasn't a big deal because we can imagine the regular chunks extending to infinity. But for the `RectilinearChunkGrid`, the shape of the array is actually coupled to the chunk grid definition -- you shouldn't be able to create a chunk grid that fails to span the shape of the array (I need to check if we made this illegal in the spec...). 

## proposed changes

In #3534, there are a lot of methods on the chunk grid that take an `array_shape` parameter, and each time this paramter has to be validated against the attributes of the `RectilinearChunkGrid`. It's far simpler to define the `array_shape` once up front and bind it to an attribute on `RectilinearChunkGrid` instances. So that's what I propose. We add an `array_shape` parameter to chunk grid construction and use it internally. 

## mild complication

With the simple change I propose, we will have an awkward situation with serialization: `RegularChunkGrid(array_shape=(10,), chunk_shape=(1,)).to_dict()` will return `{"name": "regular", "configuration": {"chunk_shape": (1,)}}`, i.e. no array_shape information. That sucks. So my proposal is to widen the type of the return value of the `to_dict` method to something like this:

```python
{
    "shape": (10,),
    "chunk_grid": {"name": "regular", "configuration": {"chunk_shape": (1,)}}
 }
```

i.e., a fragment of a v3 array metadata document that includes the array shape. happy to amend this. but the basic idea is that IMO we need to put the array shape somewhere in the output.

d-v-b:
Adds an `array_shape` attribute to `ChunkGrid`. 

Depends on #3735, supports #3534 

Resolves #3736. Opening as a draft while we figure out the JSON serialization issue.

### DIFF
--- a/changes/3710.bugfix.md
+++ b/changes/3710.bugfix.md
@@ -1,0 +1,1 @@
+Add a dedicated in-memory cache for byte-range requests to the experimental `CacheStore`.
--- a/src/zarr/abc/store.py
+++ b/src/zarr/abc/store.py
@@ -19,7 +19,7 @@ if TYPE_CHECKING:
 __all__ = ["ByteGetter", "ByteSetter", "Store", "set_or_delete"]
 
 
-@dataclass
+@dataclass(frozen=True, slots=True)
 class RangeByteRequest:
     """Request a specific byte range"""
 
@@ -29,7 +29,7 @@ class RangeByteRequest:
     """The end of the byte range request (exclusive)."""
 
 
-@dataclass
+@dataclass(frozen=True, slots=True)
 class OffsetByteRequest:
     """Request all bytes starting from a given byte offset"""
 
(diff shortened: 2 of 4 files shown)


d-v-b:
this is not useful given the changes in #3802

Issue #486, opened 2019-10-19
How to prevent Zarr from returning NaN for missing chunks?


willirath:
Is there a way of preventing Zarr from returning NaNs if a chunk is missing? 

Background of my question: We're seeing problems with either copying data to GCS or with GCS having problems to reliably serve all chunks of a Zarr store.

In `arr` below, there's two types of NaN filled chunks returned by Zarr.
```python
from dask import array as darr
import numpy as np

arr = darr.from_zarr(""gs://pangeo-data/eNATL60-BLBT02X-ssh/sossheig/")
```

First, there's a chunk that is completely flagged missing in the data (chunk is over land in an Ocean dataset) but present on GCS (<https://console.cloud.google.com/storage/browser/_details/pangeo-data/eNATL60-BLBT02X-ssh/sossheig/0.0.0>) and Zarr correctly find all items marked as invalid:
```python
np.isnan(arr.blocks[0, 0, 0]).mean().compute()
# -> 1.0
```

Then, there's a chunk (<https://console.cloud.google.com/storage/browser/_details/pangeo-data/eNATL60-BLBT02X-ssh/sossheig/0.7.3>) that is not present (at the time of writing this, I get a "load failed" and a tracking id from GCS) and Zarr returns all items marked invalid as well:
```python
np.isnan(arr.blocks[0, 7, 3]).mean().compute()
# -> 1.0
```

How do I make Zarr raise an Exception on the latter?

cc: @auraoupa
related: pangeo-data/pangeo#691

willirath:
Looks like it's necessary to override https://github.com/zarr-developers/zarr-python/blob/e7708c948d2c0ff91863d675c1072b0dfe9ce2a6/zarr/core.py#L1581 et al to make zarr raise on non-existing chunks?

jakirkham:
Have you tried setting the fill value?

alimanfoo:
Yes this is not currently supported, but would be straightforward to add.
I've actually been thinking myself recently that this would be useful for
very similar reasons, i.e., when you want to ensure that an exception is
raised when trying to access data from a missing chunk.

One way to achieve this could be to add a mechanism for activating this
behaviour explicitly. E.g., something like:

z = ... # some zarr array
z.set_options(fill_missing_chunk=False)

Thoughts and suggestions welcome.

alimanfoo:
[...]

PR #489, opened 2019-10-22
Allow disabling filling of missing chunks

willirath:
This is a first stab at solving #486 by overriding filling of missing chunks.

TODO:
* [ ] Add unit tests and/or doctests in docstrings
* [ ] Add docstrings and API docs for any new/modified user-facing classes and functions
* [ ] New/modified features documented in docs/tutorial.rst
* [ ] Changes documented in docs/release.rst
* [ ] Docs build locally (e.g., run ``tox -e docs``)
* [ ] AppVeyor and Travis CI passes
* [ ] Test coverage is 100% (Coveralls passes)

PR #3748, opened 2026-03-05
Fill missing chunks

williamsnell:
Add config options for whether a missing chunk should:
- appear as a chunk filled with `fill_value` (current behaviour; retained as default)
- raise a `MissingChunkError`

This PR is entirely based on the work of @tomwhite in [this issue.](https://github.com/zarr-developers/zarr-python/issues/486#issuecomment-3385290111) I've started this PR as this an important feature that I'd like to see merged.
I've added a test (based on the demo in the issue) and a minor docs tweak.

### Questions:
- I've added an example to `config.md` - is this the right place?

TODO:
* [X] New/modified features documented in `docs/user-guide/*.md`


tomwhite:
Thanks for doing this @williamsnell!

d-v-b:
this is great @williamsnell! I'm wondering if this be exposed as part of the general array configuration, which is where `write_empty_chunks` currently sits.

williamsnell:
@d-v-b I've pushed a new commit moving this to `ArrayConfig` so we can look into the ergonomics. It does feel like a more natural fit, especially since users probably want to set `write_empty_chunks` and `fill_missing_chunks` as a pair.

One note: I'm not sure how this interacts with Sharding now - following the existing code I hardcoded `fill_missing_chunks=True` [here](https://github.com/zarr-developers/zarr-python/pull/3748/changes#diff-a10af12360a7f5329a7fabf7395daff016aa7efa81a1ae2e13786abee911c5bcL714) - does this imply there's no way to use `fill_missing_chunks` alongside sharding? If so, I can update the docs.

d-v-b:
> One note: I'm not sure how this interacts with Sharding now - following the existing code I hardcoded `fill_missing_chunks=True` [here](https://github.com/zarr-developers/zarr-python/pull/3748/changes#diff-a10af12360a7f5329a7fabf7395daff016aa7efa81a1ae2e13786abee911c5bcL714) - does this imply there's no way to use `fill_missing_chunks` alongside sharding? If so, I can update the docs.

I don't think we want this new configuration option to change the behavior of the sharding codec. A missing subchunk inside a shard is conveyed explicitly via the shard index, so from the sharding codec's POV you can't have a subchunk _appear_ missing due to a network error.

williamsnell:
> I don't think we want this new configuration option to change the behavior of the sharding codec. A missing subchunk inside a shard is conveyed explicitly via the shard index, so from the sharding codec's POV you can't have a subchunk _appear_ missing due to a network error.

If I've understood correctly, we'll want to make this tweak to `ShardingCodec._get_chunk_spec`:

```diff python
def _get_chunk_spec(self, shard_spec: ArraySpec) -> ArraySpec:
+  # Because the shard index and inner chunks should be stored
+  # together, we detect missing data via the shard index.
+  # The inner chunks defined here are thus allowed to return
+  # None, even if fill_missing_chunks=False at the array level.
+  config = replace(shard_spec.config, fill_missing_chunks=True)
   return ArraySpec(
       shape=self.chunk_shape,
       dtype=shard_spec.dtype,
       fill_value=shard_spec.fill_value,
-      config=shard_spec.config,
+      config=config,
       prototype=shard_spec.prototype,
   )
```

With this change, I think my previous point was wrong - we would be able to use `fill_missing_chunk=False` with sharding, and we would only raise an error if an entire shard (specifically its shard index) was missing.

williamsnell:
I've committed the change to `_get_chunk_spec` and rebased onto main.

I've also made two more changes:
- Added some tests to check/codify expected behaviour around:
  - when a sharded array should or shouldn't raise MissingChunkError
  - how `fill_missing_chunks` is expected to interact with `write_empty_chunks`
- Simplified the branching in `codec_pipeline` into `if/elif/else`.

williamsnell:
Is there a way to run `coverage` on this PR? I assumed it would pop up automatically, but apparently not.

src/zarr/errors.py
```diff
@@ -144,3 +144,6 @@ class BoundsCheckError(IndexError): ...
 
 
 class ArrayIndexError(IndexError): ...
+
+
+class MissingChunkError(IndexError): ...
```
d-v-b: this error should probably include a message that names the missing chunk. also i'm not sure an index error is the right base class? Indexing succeeded, but there was a problem on the storage side.

d-v-b:
```suggestion
class ChunkNotFoundError(BaseZarrError):
    """
    Raised when a chunk that was expected to be found was not found in storage.
    """
```

The rest of the PR would need to be updated after this change

src/zarr/core/codec_pipeline.py
```diff
@@ -288,8 +290,10 @@ async def read_batch(
                     if drop_axes != ():
                         tmp = tmp.squeeze(axis=drop_axes)
                     out[out_selection] = tmp
-                else:
+                elif chunk_spec.config.fill_missing_chunks:
                     out[out_selection] = fill_value_or_default(chunk_spec)
+                else:
+                    raise MissingChunkError
```
d-v-b: we should include the name of the chunk in the error message

d-v-b:
a few suggestions about the error class, if it's OK with you @williamsnell I'd be happy pushing those changes to this branch?

williamsnell:
> a few suggestions about the error class, if it's OK with you @williamsnell I'd be happy pushing those changes to this branch?

Go for it @d-v-b - thank you!

d-v-b:
- the exception is now called `ChunkNotFoundError`, which is consistent with the name convention for other exceptions in `zarr.errors`, and it inherits from `BaseZarrError`

- I added a message to the invocation of this exception that contains the chunk key and the chunk grid index for the missing chunk. _This required changing the `batch_info` object_ that gets passed around to chunk encoding  / decoding routines. The fact that we didn't know the names of chunks when we were fetching them is a cry for help from that part of the codebase. 

  An even better error message would include the store, but since we don't have a store here, only a "byte getter" protocol, we can't actually get the store! Great stuff, but not something we can fix here.

  Because this change touches some chunk encoding stuff, I added the benchmark label to ensure that we don't have any perf regressions.

- I added a note in the docs about how the spec describes the use of the fill value (for uninitialized chunks) that frames the use of the new config option (it's for chunks you _know_ are initialized). This framing means we are not violating the spec with this new error / behavior.

maxrjones:
Thanks so much for working on this @williamsnell! This is a hugely important feature 🚀 


What do folks think about returning missing chunk information from `read_batch` rather than passing through more details via `batch_info`? I'm interested in this idea because returning information about missing chunks would allow richer error messages and would keep `batch_info` shorter, which is already a bit unwieldy. This behavior is inspired by Rust's Option syntax.

I put together a short uv script to test it out:

```python
# /// script
# dependencies = [
#   "zarr @ git+https://github.com/maxrjones/zarr-python@130a91b24fed00597f77487ba9cb8bb6b3c35a80", # equivalent to the head commit on this PR
#   "numpy",
#   "obstore",
# ]
# ///

from obstore.store import from_url
import numpy as np
import zarr

url = "https://rice1.osn.mghpcc.org/carbonplan/cp-cmip/version1/rechunked_data/GARD-SV/CMIP.CCCma.CanESM5.historical.r1i1p1f1.day.GARD-SV.tasmax.zarr"

with zarr.config.set({'array.fill_missing_chunks': False}):
    store = from_url(url)
    zs = ObjectStore(store)

    root = zarr.open_group(zs)
    arr = root["tasmax"]
    data = arr[0, 0:1000, 0:1000]

```

The current head commit in this PR gives:
```console
zarr.errors.ChunkNotFoundError: chunk '0.0.4' at grid position (0, 0, 4) not found in store
```

If we instead pass the missing chunk info up the stack, we can provide information about all missing chunks and the store (e.g., 
https://github.com/maxrjones/zarr-python/commit/37a40e37f4ec57da56629c3505fdd2265b8d7937):

```console
zarr.errors.ChunkNotFoundError: 77 chunk(s) not found in store 'object_store://HTTPStore("https://rice1.osn.mghpcc.org/carbonplan/cp-cmip/version1/rechunked_data/GARD-SV/CMIP.CCCma.CanESM5.historical.r1i1p1f1.day.GARD-SV.tasmax.zarr")/tasmax'.
Set the 'array.fill_missing_chunks' config to True to fill missing chunks with the fill value.
Missing chunks:
  chunk '0.0.0' (grid position (0, 0, 0))
  chunk '0.0.1' (grid position (0, 0, 1))
  chunk '0.0.2' (grid position (0, 0, 2))
  chunk '0.0.3' (grid position (0, 0, 3))
  chunk '0.0.4' (grid position (0, 0, 4))
  chunk '0.0.5' (grid position (0, 0, 5))
  chunk '0.0.6' (grid position (0, 0, 6))
  chunk '0.1.0' (grid position (0, 1, 0))
  chunk '0.1.1' (grid position (0, 1, 1))
...
```
(77 chunks total in error message — truncated for readability)

docs/user-guide/arrays.md
```diff
@@ -158,13 +158,25 @@ print(f"Shape after second append: {z.shape}")
 
 Zarr arrays are parametrized with a configuration that determines certain aspects of array behavior.
 
-We currently support two configuration options for arrays: `write_empty_chunks` and `order`.
+We currently support three configuration options for arrays: `write_empty_chunks`, `fill_missing_chunks`, and `order`.
 
 | field | type | default | description |
 | - |     - | - | - |
 | `write_empty_chunks` | `bool` | `False` | Controls whether empty chunks are written to storage. See [Empty chunks](performance.md#empty-chunks).
+| `fill_missing_chunks` | `bool` | `True` | Controls whether missing chunks are filled with the array's fill value on read. If `False`, reading missing chunks raises a [`ChunkNotFoundError`][zarr.errors.ChunkNotFoundError].
```
ilan-gold: For symmetry, should definitely prefix this with `read_xxx`.

src/zarr/abc/codec.py
```diff
@@ -430,7 +430,9 @@ async def encode(
     @abstractmethod
     async def read(
         self,
-        batch_info: Iterable[tuple[ByteGetter, ArraySpec, SelectorTuple, SelectorTuple, bool]],
+        batch_info: Iterable[
```
ilan-gold: Ok so now codec pipelines would need to support another element in batch_info? I think @TomAugspurger had a PR moving tuple -> TypedDict or something, but would rather handle that in the release 3.2.0  TBH than another "breaking change when we change the codec pipeline" next time (again)

src/zarr/core/codec_pipeline.py
```diff
@@ -259,15 +261,19 @@ async def read_batch(
...
                 if chunk_array is not None:
                     if drop_axes:
                         chunk_array = chunk_array.squeeze(axis=drop_axes)
                     out[out_selection] = chunk_array
-                else:
+                elif chunk_spec.config.fill_missing_chunks:
                     out[out_selection] = fill_value_or_default(chunk_spec)
+                else:
+                    raise ChunkNotFoundError(
+                        f"chunk '{chunk_key}' at grid position {chunk_coords} not found in store"
```
ilan-gold: I think at least the `key` information is recoverable from `ByteGetter`: https://github.com/zarrs/zarrs-python/blob/main/python/zarrs/utils.py#L222

src/zarr/codecs/sharding.py
```diff
@@ -711,17 +715,22 @@ def _get_index_chunk_spec(self, chunks_per_shard: tuple[int, ...]) -> ArraySpec:
...
             prototype=default_buffer_prototype(),
         )
 
     def _get_chunk_spec(self, shard_spec: ArraySpec) -> ArraySpec:
+        # Because the shard index and inner chunks should be stored
+        # together, we detect missing data via the shard index.
+        # The inner chunks defined here are thus allowed to return
+        # None, even if fill_missing_chunks=False at the array level.
+        config = replace(shard_spec.config, fill_missing_chunks=True)
         return ArraySpec(
```
ilan-gold, line 736: Unfortunate that chunk spec is implemented as something called `ArraySpec` because that would be nice for handing over the `chunk_coords` if `ArraySpec` could take in a coordinate arg but of course that doesn't make sense semantically

ilan-gold:
The other thing - this would likely require some development in `zarrs` and `zarrs-python`.  Currently in `zarrs-python` full chunk (i.e., a whole file, so shard as well) reads [are filled explicitly in `zarrs-python` but partial reads are handled by `zarrs`](https://github.com/zarrs/zarrs-python/blob/main/src/lib.rs#L352-L379).

But `zarrs` for partial reads [also explicitly fills](https://github.com/zarrs/zarrs/blob/373cdb21c6a4bbfe05d79fc1b9bd7c12e6f8177e/zarrs/src/array/codec/array_to_bytes/sharding/sharding_partial_decoder_sync.rs#L369-L395) at least for fixed-size arrays (I assume the same is true elsewhere).

So I think the "real" compatibility solution is

1. If we are in a rush to implement, raise an error in `zarrs-python` when this feature is turned on in the interim 
2. Independent of the above, ensure we are ready for the breaking pipeline change
3. Implement in `zarrs` a new feature that can be turned on/bubbled up to python for matching this behavior

ilan-gold:
> What do folks think about returning missing chunk information from read_batch rather than passing through more details via batch_info? I'm interested in this idea because returning information about missing chunks would allow richer error messages and would keep batch_info shorter, which is already a bit unwieldy. This behavior is inspired by Rust's Option syntax.

I think returning all missing chunks is definitely a worthy goal! This doesn't exactly solve my "the codec pipeline breaks" since now it is just in a different way, but the direction is good IMO - If we were to implement this missing-chunk handling in `zarrs` directly, this feature of reporting all missing chunks makes sense.

maxrjones:
> This doesn't exactly solve my "the codec pipeline breaks" since now it is just in a different way, but the direction is good IMO

is this in reference to https://github.com/zarr-developers/zarr-python/pull/3748#discussion_r2945287407? the idea in https://github.com/maxrjones/zarr-python/commit/37a40e37f4ec57da56629c3505fdd2265b8d7937 proposes removing the changes to `batch_info`, but would require the read function to return chunk info in order for it to work

docs/user-guide/arrays.md
williamsnell: Were you thinking `read_empty_chunks`, `read_missing_chunks`, or something else?

src/zarr/core/codec_pipeline.py
williamsnell: Is this solving the same problem as @maxrjones proposed in the top-level review, where the info was passed further up the stack? https://github.com/maxrjones/zarr-python/commit/37a40e37f4ec57da56629c3505fdd2265b8d7937

d-v-b:
I'm also +1 on reporting all the missing chunks we know about, I think the changes in https://github.com/maxrjones/zarr-python/commit/37a40e37f4ec57da56629c3505fdd2265b8d7937 would only be breaking if you were relying on the codec pipeline `read` method returning `None`.

ilan-gold:
I think the "compatibility" issue would be raising the exact same error class but definitely a runtime version check could handle that as long as the actual error class is exported!

d-v-b:
>  as long as the actual error class is exported!

yeah `zarr.errors` is public

williamsnell:
From the discussion above, I'm planning to do the following:
1) integrate the changes from https://github.com/maxrjones/zarr-python/commit/37a40e37f4ec57da56629c3505fdd2265b8d7937
2) rename `fill_missing_chunks` to `read_missing_chunks`

Are we all happy with this approach? Anything I've missed?

d-v-b:
> From the discussion above, I'm planning to do the following:
> 
>     1. integrate the changes from [maxrjones@37a40e3](https://github.com/maxrjones/zarr-python/commit/37a40e37f4ec57da56629c3505fdd2265b8d7937)
> 
>     2. rename `fill_missing_chunks` to `read_missing_chunks`
> 
> 
> Are we all happy with this approach? Anything I've missed?

This seems like a good summary! I think max's approach doesn't require breaking any public APIs, but it does change the return type of `CodecPipeline.read`, which should probably be critically evaluated. We need `read()` to return something that conveys the missing chunks, but it might be worth checking if `list[int]` is the best return type for that purpose.

d-v-b:
> it might be worth checking if `list[int]` is the best return type for that purpose.

for example, `tuple[ReadResult, ...]`, where `ReadResult` is something like this:

```
ReadResult = Literal["missing", "present"]
```

Clients interested in the indices of the missing values can find them, but this is also open to adding additional information in the future by widening the type of `ReadResult`.

williamsnell:
> > it might be worth checking if `list[int]` is the best return type for that purpose.
> 
> for example, `tuple[ReadResult, ...]`, where `ReadResult` is something like this:
> 
> ```
> ReadResult = Literal["missing", "present"]
> ```
> 
> Clients interested in the indices of the missing values can find them, but this is also open to adding additional information in the future by widening the type of `ReadResult`.

Speaking as the obviously least qualified zarr contributor in this thread, this sounds good to me 😄. 

Before I make this change: double-checking the exact type we want to implement here? e.g. `Literal["missing", "present"]` vs an enum vs `Literal[0, 1]` vs something more complex (a `ReadResult` dataclass with a `missing` member or something?). I don't imagine this is the *most* performance critical part of the codebase, but could imagine a world in which code that interacts with the `read()` interface could be quite opinionated about this type. So double-checking.

Otherwise, I'll merge https://github.com/maxrjones/zarr-python/commit/37a40e37f4ec57da56629c3505fdd2265b8d7937 into this branch, then modify it to return `List[ReadResult]`, `ReadResult = Literal["missing", "present"]` as you've suggested @d-v-b.

maxrjones:
TBH I don't see value in widening the return type. While marginal compared to I/O,  return a "present" status for every chunk adds memory and processing overhead. I think we should only do that if there's practical value from the information, rather than speculatively suggesting that the information could be needed later.

maxrjones:
I could see value in extensibility in the return type for missing chunk information. Something like:

```python
@dataclass(frozen=True)
class MissingChunkInfo:
    key: str
    coordinates: tuple[int, ...]
```

With `read()` returning `list[MissingChunkInfo]`. This gives meaningful extensibility (e.g., a future `reason` field) without paying per-chunk overhead for present chunks. The caller could then build a comprehensive error message directly from the list without indirecting back through `batch_info`.

d-v-b:
> TBH I don't see value in widening the return type. While marginal compared to I/O, return a "present" status for every chunk adds memory and processing overhead. I think we should only do that if there's practical value from the information, rather than speculatively suggesting that the information could be needed later.

By returning a list of ints, the current design is not extensible without breaking downstream consumers. We can't predict how this might be used, so we should aim for a return type that can be extended. 

I don't think the literal strings are terribly extensible, but a typeddict would be:

```python
class ReadResult(TypedDict):
  missing: Literal[True, False]

```

This is cheap to construct and gives us as much flexibility as we could possibly need, provided we are OK sticking with a python primitive.

d-v-b:
> without paying per-chunk overhead for present chunks.

Compared to IO and compression, the overhead of bookkeeping our reads is negligible. That probably shouldn't weigh too heavily on our design here.

maxrjones:
Color me skeptical of returning info about all chunks rather than just missing ones, but I'd be glad to be proven wrong down the line. I support the solution from https://github.com/zarr-developers/zarr-python/pull/3748#issuecomment-4118903666 and won't block a dense tuple or set return type.

d-v-b:
sorry for the distraction from the core of your PR @williamsnell, we are definitely going to get this merged one way or another. It's just that we are running into a few warts in our basic IO model, which is that so far we have not associated IO operations with any kind of annotations / metadata besides the literal payload when fetching data. 

This means useful information, like the time required for the read / write, and any other information about the read / write operation the storage backend might know, is totally missing from our stack. Ideally we would have defined these semantics at the `Store` level, and then we could re-use them here, but we haven't gotten around to that yet :/

maxrjones:
💯  thank you for your patience and working on this @williamsnell 🙇 🙌

williamsnell:
I'm working on merging the changes from upstream/main and @maxrjones's branch. 

Question: `read_missing_chunks` feels off to me, now that I'm propagating it throughout the docs. It doesn't clearly convey *what* reading a missing chunk does.

The only name I've found that feels sufficiently explanatory and keeps the `read` prefix is `read_missing_chunks_as_fill_value`. This is wordy but at least explicit. 

Any strong preferences for any of the options (`fill_missing_chunks`, `read_missing_chunks`, or `read_missing_chunks_as_fill_value`)?

maxrjones:
> Question: `read_missing_chunks` feels off to me, now that I'm propagating it throughout the docs. It doesn't clearly convey _what_ reading a missing chunk does.
> 
> The only name I've found that feels sufficiently explanatory and keeps the `read` prefix is `read_missing_chunks_as_fill_value`. This is wordy but at least explicit.
> 
> Any strong preferences for any of the options (`fill_missing_chunks`, `read_missing_chunks`, or `read_missing_chunks_as_fill_value`)?

yeah, I get that. what about `raise_on_missing_chunks` or `error_on_missing_chunks`? I think that most directly describes the feature that's been missing. The default would be False, in contrast to the other options which would default to True.

specifically out of the options you suggested, I find `fill_missing_chunks` the most intuitive.

williamsnell:
I've pushed the combined changes to this branch - thanks all!

Note: codec_pipeline.py is now identical to what's on main:
- We always fill chunks with the array fill value.
- We always return a `GetResult` of either `"present"` or `"missing"`.

Only in array.py do we actually check the config for `read_missing_chunks` and decide whether to investigate the `list[GetResult]` and possibly raise an error.

williamsnell:
> yeah, I get that. what about `raise_on_missing_chunks` or `error_on_missing_chunks`? I think that most directly describes the feature that's been missing. The default would be False, in contrast to the other options which would default to True.
> 
> specifically out of the options you suggested, I find `fill_missing_chunks` the most intuitive.

With the refactors I've just pushed, the most accurate description is definitely `raise_on_missing_chunks` or `error_on_missing_chunks`.

d-v-b:
> > yeah, I get that. what about `raise_on_missing_chunks` or `error_on_missing_chunks`? I think that most directly describes the feature that's been missing. The default would be False, in contrast to the other options which would default to True.
> > specifically out of the options you suggested, I find `fill_missing_chunks` the most intuitive.
> 
> With the refactors I've just pushed, the most accurate description is definitely `raise_on_missing_chunks` or `error_on_missing_chunks`.

I think both of these names work, I will submit a third: `allow_missing_chunks`. But any one of these seems good to me.

williamsnell:
Ok, I've gone in circles myself enough with the naming and think all options have their merits. For the sake of keeping momentum, I'll leave the name as is (`read_missing_chunks`).

With that, I think this PR is ready - but if there's anything else needed, please let me know!

ilan-gold:
> The only name I've found that feels sufficiently explanatory and keeps the read prefix is read_missing_chunks_as_fill_value. This is wordy but at least explicit

I think this is fine TBH.  Just wanted to reiterate why I mentioned this in the first place (since it hasn't been mentioned): having `write_empty_chunks` + `fill_missing_chunks` *to me* is weird.  What does `fill_missing_chunks` do if not the same thing as `write_empty_chunks`?

Very pedantic, and I appear to be in the minority.

d-v-b:
i think as long as the name is not catastrophically bad we are fine :) we can always introduce a new config field in the future if we are for some reason deeply dissatisfied with the one here.

d-v-b (approved):

d-v-b:
it's going in, thanks @williamsnell

williamsnell:
Thank you all!! Great to have this merged 😁

(As a small aside, I was scrolling through PRs and noticed we might be able to close the oldest open PR in the repo: https://github.com/zarr-developers/zarr-python/pull/489 [Allow disabling filling of missing chunks](https://github.com/zarr-developers/zarr-python/pull/489). Just thought this was cool - in any case, great to have this feature now!)

PR #3761, opened 2026-03-11
fix: initialize _lock attribute in ZipStore (#3588)

### LINKED ISSUE #3588: AttributeError: 'ZipStore' object has no attribute '_lock'
### Zarr version

v3.1.3

### Numcodecs version

v0.16.3

### Python Version

3.12.10

### Operating System

Linux

### Installation

using uv

### Description

Calling the `exists` method on a `ZipStore` instance for an existing ZIP file raises the exception `AttributeError: 'ZipStore' object has no attribute '_lock'`. It seems this is because the method is not protected by a block

```python
 if not self._is_open:
    self._sync_open()
```

as seen elsewhere in the code. If this was simply an omission it should be an easy fix, I can open a PR if you want.

Thanks a lot for the great library!

### Steps to reproduce

```python
# /// script
# requires-python = ">=3.11"
# dependencies = [
# ]
# ///
#

from zipfile import ZipFile
import asyncio

ZipFile("/tmp/tmp.zip", "w").close()

async def main():
    with ZipStore("/tmp/tmp.zip", mode="r") as store:
        return await store.exists("somekey")

asyncio.run(main())
```

### Additional output

_No response_

abishop1990:
## Summary

Fixes #3588.

ZipStore._lock was initialized in `_sync_open()` but accessed by methods (get, set, exists, etc.) that can be called before the store is explicitly opened. This caused:

```
AttributeError: 'ZipStore' object has no attribute '_lock'
```

## Changes

- Move `self._lock = threading.RLock()` from `_sync_open()` to `__init__()`
- Add regression test `test_lock_initialized_before_open`

## Testing

- New test verifies `_lock` exists immediately after instantiation
- All existing ZipStore tests pass

```
tests/test_store/test_zip.py  1 passed
```

### DIFF
--- a/src/zarr/storage/_zip.py
+++ b/src/zarr/storage/_zip.py
@@ -87,12 +87,11 @@ class ZipStore(Store):
         self._zmode = mode
         self.compression = compression
         self.allowZip64 = allowZip64
+        self._lock = threading.RLock()
 
     def _sync_open(self) -> None:
         if self._is_open:
             raise ValueError("store is already open")
-
-        self._lock = threading.RLock()
 
         self._zf = zipfile.ZipFile(
             self.path,
--- a/tests/test_store/test_zip.py
+++ b/tests/test_store/test_zip.py
@@ -152,3 +152,15 @@ class TestZipStore(StoreTests[ZipStore, cpu.Buffer]):
         assert destination.exists()
         assert not origin.exists()
         assert np.array_equal(array[...], np.arange(10))
+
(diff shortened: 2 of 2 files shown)


d-v-b:
thanks @abishop1990, how does this compare to https://github.com/zarr-developers/zarr-python/pull/3593?

abishop1990:
I see PR #3593 is already addressing this issue. Closing to avoid duplicate work and let the existing PR move forward.

PR #3763, opened 2026-03-11
fix: apply drop_axes squeeze in partial decode path for sharding (#3691)

### LINKED ISSUE #3691: Shape mismatch with mixed integer/list indexing on Sharded arrays
### Zarr version

v3.1.5

### Numcodecs version

v0.16.5

### Python Version

3.11.13

### Operating System

Mac

### Installation

uv

### Description

I found a bug when using mixed indexing (combining slicing, integer indexing, lists indexing) on a sharded array. The indexing works fine on a chunked-only array.

### Steps to reproduce

Minimal example:
```python
import numpy as np
import zarr

chunked = zarr.create_array(
    "chunked.zarr",
    shape=(200, 100, 10),
    dtype=np.uint8,
    overwrite=True,
    chunks=(200, 100, 1),
)
chunked[:, :, :] = np.random.randint(0, 255, (200, 100, 10))

sharded = zarr.create_array(
    "sharded.zarr",
    shape=(200, 100, 10),
    dtype=np.uint8,
    overwrite=True,
    chunks=(200, 100, 1),
    shards=(200, 100, 10),
)
sharded[:, :, :] = np.random.randint(0, 255, (200, 100, 10))

print("chunked", chunked[0:10, 0, [0, 1]].shape)
print("sharded", sharded[0:10, 0, [0, 1]].shape)
```

### Additional output

Outputs
```
chunked (10, 2)
...
ValueError: could not broadcast input array from shape (10,1,2) into shape (10,2)
```

Workaround: replacing 0 with a slice `0:1` works:
```
print(sharded[0:10, 0:1, [0, 1]].shape)
```
although the shape is then (10, 1, 2), not (10, 2).

abishop1990:
## Summary

Fixes #3691.

Mixed integer/list indexing on sharded arrays (e.g. `arr[0:10, 0, [0, 1]]`) raised:
```
ValueError: could not broadcast input array from shape (10,1,2) into shape (10,2)
```

## Root Cause

When `OrthogonalIndexer` processes advanced indexing (slices + integers + arrays), it applies `ix_()` to `chunk_selection` to set up orthogonal numpy indexing. Integer indices become 1-element ranges (size-1 dimensions) via `ix_()`.

`CodecPipeline.read_batch()` has two paths:
1. **Non-partial decode** (regular codecs): Applies `drop_axes.squeeze()` to remove size-1 integer dims ✅
2. **Partial decode** (`ShardingCodec`): Missing `drop_axes.squeeze()` ❌

`ShardingCodec._decode_partial_single()` receives the `ix_()`-transformed `chunk_selection`, which looks like pure fancy indexing to `get_indexer()`, so it routes to `CoordinateIndexer`. The result is reshaped to the broadcast coordinate shape `(10, 1, 2)` instead of `(10, 2)`.

## Fix

Apply `drop_axes` squeeze to `chunk_array` in the partial decode branch of `read_batch()`, matching the non-partial path behaviour:

```python
if drop_axes != ():
    chunk_array = chunk_array.squeeze(axis=drop_axes)
```

## Testing

Added `test_sharding_mixed_integer_list_indexing` that verifies:
- Shape and data equality between chunked and sharded arrays for mixed indexing
- Multiple integer axes (`arr[0, 0, [0, 1, 2]]`)
- Slice + integer + slice (`arr[0:5, 1, 0:3]`)

```
tests/test_codecs/test_sharding.py  125 passed, 1 skipped
tests/test_indexing.py  149 passed, 1 skipped, 5 xfailed
```

### DIFF
--- a/src/zarr/core/array.py
+++ b/src/zarr/core/array.py
@@ -1235,24 +1235,28 @@ class AsyncArray(Generic[T_ArrayMetadata]):
     @property
     def cdata_shape(self) -> tuple[int, ...]:
         """
-        The shape of the chunk grid for this array.
+        The number of chunks along each dimension.
+
+        When sharding is used, this counts inner chunks (not shards) per dimension.
 
         Returns
         -------
         tuple[int, ...]
-            The shape of the chunk grid for this array.
+            The number of chunks along each dimension.
         """
         return self._chunk_grid_shape
 
     @property
     def _chunk_grid_shape(self) -> tuple[int, ...]:
         """
-        The shape of the chunk grid for this array.
+        The number of chunks along each dimension.
+
+        When sharding is used, this counts inner chunks (not shards) per dimension.
 
         Returns
         -------
         tuple[int, ...]
-            The shape of the chunk grid for this array.
+            The number of chunks along each dimension.
         """
         return tuple(starmap(ceildiv, zip(self.shape, self.chunks, strict=True)))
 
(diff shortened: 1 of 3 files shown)


src/zarr/core/codec_pipeline.py
```diff
@@ -263,6 +263,8 @@ async def read_batch(
                 chunk_array_batch, batch_info, strict=False
             ):
                 if chunk_array is not None:
+                    if drop_axes != ():
```
d-v-b: are we sure that drop_axes is always a tuple? maybe it is, but if not, might be safer to check if its length is 0

abishop1990: Good catch — switched to `if drop_axes:` (truthy check) throughout `codec_pipeline.py`. It's both more Pythonic and type-agnostic, even though `drop_axes` is annotated as `tuple[int, ...]` everywhere. Pushed in the latest commit.

Issue #3772, opened 2026-03-13
support scale_offset and cast_value codecs


d-v-b:
the `numcodecs.fixescaleoffset` codec is flawed for Zarr v3 and should not be used, because it has not been adapted properly to the Zarr v3 data model. To fix this we developed two new codecs in this PR https://github.com/zarr-developers/zarr-extensions/pull/43

- `scale_offset` is an array->array codec that applies a scale and offset to the values of an input array. Here's the spec: https://github.com/d-v-b/zarr-extensions/tree/feat/scale-offset/codecs/scale_offset
- `cast_value` is an array -> array codec that changes the data type of the input array to a new data type by mapping the input values to output values via a casting procedure. Here's the spec: https://github.com/d-v-b/zarr-extensions/tree/feat/scale-offset/codecs/cast_value

we should implement these codecs here or externally and direct users away from `numcodecs.fixedscaleoffset`

one concern about `cast_value` is that it contains some logic that might be hard to express in performant numpy code. We may have to choose between a memory-and-compute inefficient pure numpy solution vs a performant compiled solution.

if you or someone you know used the scale-offset transformation with zarr data, either via a codec or via a metadata convention a la CF + xarray, please have a look at the linked PR, the specs, and keep an eye on the developments related to this feature in zarr python.

d-v-b:
closed via #3774

PR #3774, opened 2026-03-13
feat/v3 scale offset cast

d-v-b:
This PR adds `scale_offset` and `cast_value` codecs. Used together, these codecs replace the `fixedscaleoffset` codec defined in numcodecs.

closes https://github.com/zarr-developers/zarr-python/issues/3772

as an experiment, I'm including some developer notes I had claude write in the commit history. We can remove them before merging if we don't want that kind of thing sticking around.


d-v-b:
cc @emmaualemalthot

src/zarr/codecs/cast_value.py
```diff
@@ -0,0 +1,387 @@
...
+    elif mode == "towards-positive":
+        return np.ceil(arr)
+    elif mode == "towards-negative":
+        return np.floor(arr)
+    elif mode == "nearest-away":
+        return np.sign(arr) * np.floor(np.abs(arr) + 0.5)
+    raise ValueError(f"Unknown rounding mode: {mode}")
+
+
+def _cast_array(
```
d-v-b, line 91: this implementation is inefficient and will probably remain so as long as we are using python for this.

d-v-b:
I'm a bit worried that this implementation is a bit too much stuff for zarr-python, and that perhaps this belongs in a separate project that manages the zarr-python compatibility. Thoughts?

emmanuelmathot:
@d-v-b For me it really depends on how we can easily release this PR. How easy is it to release it here and then move to an external package or vice-versa?

maxrjones:
> I'm a bit worried that this implementation is a bit too much stuff for zarr-python, and that perhaps this belongs in a separate project that manages the zarr-python compatibility. Thoughts?

I don't see a lot of value in bundling it in zarr-python, since it seems unlikely that we'd long-term encourage a pure python implementation. I'd recommend a new repository in zarr-developers for easy findability.

d-v-b:
> @d-v-b For me it really depends on how we can easily release this PR. How easy is it to release it here and then move to an external package or vice-versa?

this PR will not be easy to merge, so I'm going to close it and move this functionality to a separate package at https://github.com/zarr-developers/cast-value.py/

PR #3778, opened 2026-03-13
Remove `rich` as required for `tree()`

ianhi:
## Description

I often found that when I wanted to use `tree()` `rich` wasn't installed. So I started out with this as a fallback but on looking into it I think it can be fully removed in favor a small loop (which we would need anyway for a fallback) to construct the tree and just add in ansi or html bolding codes.

One thing I noticed when looking at this is that there is currently no truncation for extremely large hierarchies. I think that is bad for two reasons:

1. can massively overflow the users screen
2. Potentially bad performance doing the IO to get the full hierarchy.

So I've added a simplistic truncation. Note that this can result in nodes that are displayed having missing children.

I also added a `plain` flag to give the user explicit control on the bolding html/ansi sequences.

## Output

Per the tests the output is equivalent, and also looks so to my eyes.

To compare here is a script (runnable with [third-wheel](https://github.com/earth-mover/third-wheel?tab=readme-ov-file#-third-wheel))

` uvx third-wheel@0.3.0 run compare_trees.py`

<details>

<summary>compare_trees.py</summary>

```python
# /// script
# dependencies = [
#   "zarr_old",
#   "rich",
#   "zarr @ git+https://github.com/ianhi/zarr-python@ian/simple-tree",
# ]
# [tool.third-wheel]
# renames = [
#   {original = "zarr", new-name = "zarr_old", extras = ["optional"]},
# ]
# ///

"""Compare the old rich-based tree with the new plain tree."""

import zarr
import zarr_old

def make_group(mod):
    g = mod.group()
    g.create_group("A")
    g.create_group("B")
    g.create_group("B/C")
    g.create_group("B/C/D")
    return g

print("=== Old (rich-based) tree ===")
g_old = make_group(zarr_old)
print(repr(g_old.tree()))

print("=== New tree (styled) ===")
g_new = make_group(zarr)
print(repr(g_new.tree()))

print("=== New tree (plain) ===")
print(repr(g_new.tree(plain=True)))

```
</details>

image

or in jupyter:

image

TODO:

### DIFF
--- a/src/zarr/core/group.py
+++ b/src/zarr/core/group.py
@@ -2386,6 +2386,7 @@ class Group(SyncMixin):
         self,
         expand: bool | None = None,
         level: int | None = None,
+        *,
         max_nodes: int = 500,
         plain: bool = False,
     ) -> Any:


src/zarr/core/group.py
```diff
@@ -1588,19 +1588,29 @@ async def array_values(
         async for _, array in self.arrays():
             yield array
 
-    async def tree(self, expand: bool | None = None, level: int | None = None) -> Any:
+    async def tree(
+        self,
+        expand: bool | None = None,
+        level: int | None = None,
+        max_nodes: int = 500,
+        plain: bool = False,
```
d-v-b, line 1597: can we make these keyword-only? then we can evolve the API a bit more gracefully in the future

ianhi: good call. done

src/zarr/core/group.py
```diff
@@ -2371,26 +2382,38 @@ def array_values(self) -> Generator[AnyArray, None]:
         for _, array in self.arrays():
             yield array
 
-    def tree(self, expand: bool | None = None, level: int | None = None) -> Any:
+    def tree(
+        self,
+        expand: bool | None = None,
+        level: int | None = None,
+        max_nodes: int = 500,
```
maxrjones, line 2390:
```suggestion
        *,
        max_nodes: int = 500,
```

maxrjones:
thanks again @ianhi! I've been burned too by rich not being available, so I'm excited about dropping this optional dependency

PR #3779, opened 2026-03-13
fix: validate that dask-style chunks have regular shapes

maxrjones:
This PR addresses a TODO in normalize_chunks to validate that dask-style chunks are regular. I think this is important to add before the rectilinear chunks feature to make sure that rectilinear-chunks can be explicitly opted in to. Previously, irregular chunks like `((10, 20, 30),)` were silently coerced to `(10,)`, producing a grid that didn't match the input specifications.

TODO:

### DIFF
--- a/src/zarr/core/chunk_grids.py
+++ b/src/zarr/core/chunk_grids.py
@@ -128,7 +128,7 @@ def normalize_chunks(chunks: Any, shape: tuple[int, ...], typesize: int) -> tupl
     # handle dask-style chunks (iterable of iterables)
     if all(isinstance(c, (tuple, list)) for c in chunks):
         for i, c in enumerate(chunks):
-            if itertools.pairwise(c[:-2], c[1:-1]).map(lambda x, y: x == y).any()) or (len(c) > 1 and c[-1] > c[0]):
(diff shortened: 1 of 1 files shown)


src/zarr/core/chunk_grids.py
```diff
@@ -126,11 +126,14 @@ def normalize_chunks(chunks: Any, shape: tuple[int, ...], typesize: int) -> tupl
...
 
     # handle dask-style chunks (iterable of iterables)
-    if all(isinstance(c, (tuple | list)) for c in chunks):
-        # take first chunk size for each dimension
-        chunks = tuple(
-            c[0] for c in chunks
-        )  # TODO: check/error/warn for irregular chunks (e.g. if c[0] != c[1:-1])
+    if all(isinstance(c, (tuple, list)) for c in chunks):
+        for i, c in enumerate(chunks):
+            if len(set(c[:-1])) > 1 or (len(c) > 1 and c[-1] > c[0]):
```
dcherian:
```suggestion
            if itertools.pairwise(c[:-2], c[1:-1]).map(lambda x, y: x == y).any()) or (len(c) > 1 and c[-1] > c[0]):
```

`sret` requires hashing every element here which is potentially O(millions). Something like the above will short-circuit at the first inequality.

maxrjones: thank you!

PR #3781, opened 2026-03-16
Implement support for `structured` and `struct` zarr-extension defined dtypes

BrianMichell:
Updates for consistency with the zarr-extension for [structured types](https://github.com/zarr-developers/zarr-extensions/pull/45).

Implements legacy read-only for `structured` type and support for new `struct` definition.

TODO:


d-v-b:
this looks great, thanks for this. Instead of modifying the existing `structured` data type, can we make `struct` as a subclass of that data type, and register it? That means people who need the legacy dtype can still use it (by overriding the registry)

BrianMichell:
> this looks great, thanks for this. Instead of modifying the existing `structured` data type, can we make `struct` as a subclass of that data type, and register it? That means people who need the legacy dtype can still use it (by overriding the registry)

Should be good to go now!

BrianMichell:
@d-v-b Is there anything I can do to move this PR forward?

d-v-b:
> @d-v-b Is there anything I can do to move this PR forward?

hi @BrianMichell thanks for the ping, I will get a review in soon. since numpy is handling the actual data type, I think we will ask much less of you than the tensorstore devs :)

src/zarr/core/dtype/npy/structured.py
```diff
@@ -113,61 +140,11 @@ def __post_init__(self) -> None:
...
-            # care about the first element in either case.
-            for key, (dtype_instance, *_) in dtype.fields.items():  # type: ignore[union-attr]
-                dtype_wrapped = get_data_type_from_native_dtype(dtype_instance)
-                fields.append((key, dtype_wrapped))
-
-            return cls(fields=tuple(fields))
-        raise DataTypeValidationError(
-            f"Invalid data type: {dtype}. Expected an instance of {cls.dtype_cls}"
-        )
+        raise DataTypeValidationError(f"Use 'Struct' for native dtype matching. Got: {dtype}")
```
d-v-b: why do we need to modify the `Structured` data type at all? I would think we can keep it exactly as-is, but do not register it in the data type registry. Users who need backwards compatibility can explicitly unregister the new data type, and register the old one.

BrianMichell: I made this design choice to support transparent `r+` of `Structured` dtype and also `r+`/`w` for `Struct`. This implementation should support my usecase where we have petabyte scale `Structured` legacy data and expect to move forward with generating new data using `Struct` without requiring either a remaster or client code to register/unregister on the fly.

d-v-b: if we remove `Structured` from the registry, and `Struct` accepts `data_type` metadata that _would have_ worked for `Structured`, then doesn't this achieve your goal? The idea is that the input type for `Struct` is wider than the input type for `Structured`, so any valid `Structured` would also be a valid `Struct` data type. But if for some reason someone needs to use the old `Structured` data type, it is still there, untouched. They just have to opt in to using it.

edit, I mixed up `Struct` and `Structured`

BrianMichell: I see what you're getting at. Made that update!

d-v-b:
I think this is good to go in @BrianMichell, thanks for your patience here.

d-v-b (approved):

BrianMichell:
Thanks, @d-v-b ! Would it be possible to get this merged in before the next release? It's fine if not, just need to know whether it's worth delaying one of my releases for this change.

d-v-b:
> Thanks, @d-v-b ! Would it be possible to get this merged in before the next release? It's fine if not, just need to know whether it's worth delaying one of my releases for this change.

it's going in now!

PR #3784, opened 2026-03-17
chore: less frequent codspeed runs

maxrjones:
We keep running through our codspeed usage half way through the month, so this PR updates the triggers to run less frequently:

- use a cron job rather than every commit to main (which can be multiple per day)
- run on PRs when the label is applied rather than every commit thereafter

TODO:

### DIFF
--- a/.github/workflows/codspeed.yml
+++ b/.github/workflows/codspeed.yml
@@ -1,4 +1,4 @@
-name: Codspeed benchmarks
+name: CodSpeed Benchmarks
 
 on:
   schedule:


d-v-b:
what about running manually?

maxrjones:
> what about running manually?

what do you mean by this? do you think we should only have workflow_dispatch as the trigger?

d-v-b:
maybe workflow dispatch is more tedious than just removing and re-applying the label, which this PR would allow. And we already have workflow dispatch for these actions.

the only ergonomic improvement I can think of is a command that can be executed via a comment on a pr, e.g. "@benchmark-bot please run", but we can defer that for when / if need to revisit this.

PR #3789, opened 2026-03-19
chore/towncrier header

d-v-b:
Changes our towncrier config to use h2 header instead of h1 header for each new release notes entry


d-v-b:
this is the last change before the release notes PR :)

Issue #3796, opened 2026-03-20
`ValueError: No Zarr data type found that matches dtype 'dtype('uint32')'` on Windows


radoering:
### Zarr version

3.15

### Numcodecs version

0.14.1

### Python Version

3.13

### Operating System

Windows

### Installation

using poetry into virtual environment

### Description

On Windows, the following snippet, which worked with zarr 2, fails with zarr 3 with

```
ValueError: No Zarr data type found that matches dtype 'dtype('uint32')'
```

Repro:

```python
import zarr

a = np.array([1, 2], dtype=np.uint32) & 1
zarr.array(a)
```

The issue seems to be that the dtype is `UIntDtype` instead of `UInt32Dtype`.

When running the following snippet:

```python
dtypes = [
    (np.array([1, 2], dtype=np.uint16) & 1).dtype,
    (np.array([1, 2], dtype=np.uint32) & 1).dtype,
    (np.array([1, 2], dtype=np.uint64) & 1).dtype,
]
for dtype in dtypes:
    print(f"{dtype} {type(dtype)}")
```

we get

```
uint16 <class 'numpy.dtypes.UInt16DType'>
uint32 <class 'numpy.dtypes.UIntDtype'>
uint64 <class 'numpy.dtypes.UInt64DType'>
```

### Steps to reproduce

```python
import zarr

a = np.array([1, 2], dtype=np.uint32) & 1
zarr.array(a)
```

d-v-b:
thanks for this report! The fix is probably to robustly figure out which numpy int data type is actually a 32 bit integer.

PR #3797, opened 2026-03-20
fix: handle special uint32 arrays on Windows correctly

radoering:
Resolves: #3796

Changes are equivalent to #3151 just for `uint32`.

TODO:


tests/test_dtype/test_npy/test_int.py
```diff
@@ -216,7 +216,16 @@ class TestUInt16(BaseTestZDType):
 class TestUInt32(BaseTestZDType):
     test_cls = UInt32
     scalar_type = np.uint32
-    valid_dtype = (np.dtype(">u4"), np.dtype("<u4"))
+
+    # On Windows, this creates an UIntDType (instead of UInt32DType),
+    # similar to how np.dtype('i') creates an IntDType instead of Int32DType.
+    # However, np.dtype('u') raises a TypeError.
+    uint_dtype = (np.array([1], dtype=np.uint32) & 1).dtype
```
d-v-b, line 223: what's the `& 1` for?

radoering: To create the special dtype. Without the `& 1`, it is `UInt32Dtype`, with `& 1`, it is `UIntDtype`.

d-v-b: ok, thanks for the clarification. Why is numpy doing this?

radoering: I have no idea. It took me a while to figure out how to create an array with such a dtype after I encountered this error in more complex code.

d-v-b: yes I can imagine how frustrating that must have been. if this happens again we might need to consider relying on some kind of structural checks instead of trusting numpy's class hierarchy.

d-v-b: I'm looking for reports of this on the numpy issue tracker, this one seems related: https://github.com/numpy/numpy/issues/17351

radoering: @d-v-b I just noticed that there is a similar issue with `np.longlong` and `np.ulonglong` - this time on Linux. (By the way, I found out that the dtype that results from the `& 1` operation is `np.intc`.)

> if this happens again we might need to consider relying on some kind of structural checks instead of trusting numpy's class hierarchy.

I agree.

Issue #3795, opened 2026-03-20
public types for metadata objects


d-v-b:
For objects that end up in array or group metadata, there are 2 types that users might care about:
- an input type that defines the set of python objects that can be parsed to the metadata object
- an output type that defines the set of JSON-like objects that the metadata object serializes to

take the `shape` field of array metadata: 
- input type: something like `Iterable[int] | int | numpy integers`
- output type: `tuple[int, ...]`

for something like the `bytes` codec, we would have:
- input type: `BytesCodec | Literal["bytes"] | NamedConfig[Literal["bytes"], BytesConfig]`
- output type: `Literal["bytes"] | NamedConfig[Literal["bytes"], BytesConfig]`

the input type is generous, the output type is strict. Both of these types are useful for people using zarr-python to create codecs, chunk key encodings, chunk grids, data types, etc.

This issue tracks the creation of public types for the input ("deserializable" type) and output ("serializable") of zarr metadata objects. I considered making a big list with checkboxes, but it would be really long and the checkboxes would all be empty. Instead, we can update this issue when we add or expose types, and close it when they are all added / exposed.

maxrjones:
I think that this was implemented by https://github.com/zarr-developers/zarr-python/pull/3919

Issue #3786, opened 2026-03-17
type-safe metadata protocol


d-v-b:
To make our to-json and from-json serialization type-safe, we need a class like this:

```python
class Metadata(Generic[FromType, ToType]):

  @classmethod
  def from_json(cls, data: FromType) -> Self:
    ...

  def to_json(self) -> ToType:
    ...
```

This violates liskov because `FromType` gets narrower in subclasses. `ToType` and `FromType` effectively impose opposite variance requirements on `Metadata` here.

But it does work if we use generic protocols instead! So that's what we should do. I think it would be really nice to make all of our JSON serialization type-safe this way.

d-v-b:
here's a fleshed out example shape for a serializable object:

```python
from typing import Protocol, TypeVar, Self

T_contra = TypeVar("T_contra", contravariant=True)
T_co = TypeVar("T_co", covariant=True)

class Serializable(Protocol[T_contra, T_co]):
    @classmethod
    def from_json(cls, data: T_contra) -> Self: ...

    @classmethod
    def try_from_json(cls, data: object) -> Self: ...

    def to_json(self) -> T_co: ...
```

I feel some urgency to improve our metadata serialization as we consider adding new chunk grids and codecs. So I might put together a PR that adds this protocol, or something like it, soon.

PR #3799, opened 2026-03-20
add serializable protocol

d-v-b:

Defines `JSONSerializable[T_co]` in `zarr.abc.serializable`. The current protocol has one type parameter for the serialized representation and one method, `to_json()`.

`ArrayV3Metadata.to_json()` returns its JSON-compatible metadata representation. Deserialization is outside this protocol's current scope. Earlier proposals for two type parameters, `from_json`, `try_from_json`, and broader input-type changes were reduced during review.

Related: https://github.com/zarr-developers/zarr-python/issues/3786 and https://github.com/zarr-developers/zarr-python/issues/3795.


d-v-b:
for context, we have a `Metadata` ABC, but it has a serious flaw: if we define some class method like `from_dict(obj: T) -> Self`, no subclass can narrow the type `T` to something more specific without violating liskov substitution. So this base class approach is a dead end if we ever want separate metadata objects to clearly indicate the type the deserialize from, and IMO we do want that.

src/zarr/core/common.py
```diff
@@ -43,11 +44,12 @@
...
 ZarrFormat = Literal[2, 3]
 NodeType = Literal["array", "group"]
-JSON = str | int | float | bool | Mapping[str, "JSON"] | Sequence["JSON"] | None
+JSON: TypeAlias = str | int | float | bool | Mapping[str, "JSON"] | Sequence["JSON"] | None
 MemoryOrder = Literal["C", "F"]
 AccessModeLiteral = Literal["r", "r+", "a", "w", "w-"]
 ANY_ACCESS_MODE: Final = "r", "r+", "a", "w", "w-"
-DimensionNames = Iterable[str | None] | None
+DimensionNamesLike = Iterable[str | None] | None
+DimensionNames = DimensionNamesLike  # backwards compatibility
```
maxrjones: why change this name?

d-v-b: `XLike` is the name we are using for most of the other fields

maxrjones: Thanks for explaining. Looking at the type aliases here:

- `BytesLike` — has `Like`
- `ShapeLike` — has `Like`
- `ChunkCoords` — no `Like`
- `ZarrFormat` — no `Like`
- `MemoryOrder` — no `Like`
- `AccessModeLiteral` — no `Like`
- `DimensionNames` → `DimensionNamesLike` — being renamed

The `Like` suffix isn't consistently applied across the board, so renaming just this one creates a backwards-compat alias without actually achieving full consistency. If the goal is a naming convention, it should be done comprehensively or not at all and as a standalone change (separate PR). I don't think it's worth it here.

d-v-b: :robot: _AI text below_ :robot:

Historical naming discussion from the broader version of this PR. The current PR has been reduced to the serialization protocol. The concrete element-type typos and JSON-array representation statement are corrected below.

> The Like suffix isn't consistently applied across the board,

It is actually consistently applied, with a few exceptions (such as `DimensionNames`). The rule is that we use `XLike` to indicate a type that _narrows to `X`_ or _approximates `X`_. `AccessModeLiteral`, `ZarrFormat`, `MemoryOrder`, `ChunkCoords` are all simple types that never get narrowed in our codebase, and they don't approximate anything. Values that inhabit these types are usable as-is. 

By contrast, `ShapeLike` is a big union: `ShapeLike = Iterable[int | np.integer[Any]] | int | np.integer[Any]`. Members that inhabit this type are narrowed to a single type: `tuple[int, ...]`. In the context of Zarr metadata, `shape` is a fixed-length immutable sequence of ints, which we model as a `tuple[int, ...]`. A concrete shape is not an int, or an iterable of ints or numpy ints, etc. `tuple[int, ...]` is compact enough so we don't really need a type alias for this.

Which takes us to `DimensionNames`. Currently, `DimensionNames` is `Iterable[str | None] | None`. But the Zarr V3 spec requires that dimension names be a JSON array of strings or `null`, which we choose to represent internally as `tuple[str | None, ...]`; JSON arrays can also be represented by Python lists. The spec allows the `dimension_names` field to be unset, which we model by allowing the `dimension_names` attribute to be `None`. That means the concrete type is `tuple[str | None, ...] | None`. This is _narrower_ than `Iterable[str | None] | None`. the type `Iterable[str | None] | None` is the type we _accept_ as input for creating a concrete `dimension_names` field. That means this type should be named `DimensionNamesLike`, and the name `DimensionNames` should be reserved for the concrete type we would annotate `ArrayV3Metadata.dimension_names` with, namely `tuple[str | None, ...] | None`

> it should be done comprehensively or not at all and as a standalone change (separate PR). 

I'm happy to move these changes into a separate PR, but I firmly disagree that these changes need to be made comprehensively. As long as the change is understandable, and an improvement, we should encourage it. There are many warts in the codebase. We should be happy if they get fixed incrementally, and even if it would theoretically be better for them to be fixed all at once, that's a harder PR to review.

The change I made here is backwards compatible and an incremental improvement, motivated by the goal of moving our array metadata class to public API. I'm happy ot do this in a separate PR, but I'm also pretty sure this change does not hinder your ability to review this PR.

maxrjones: > > The Like suffix isn't consistently applied across the board,

> It is actually consistently applied, with a few exceptions 

I mean, that makes it inconsistent 😆

thanks for the semantics explanation, that's helpful

> I'm happy to move these changes into a separate PR

yes, in general I think it's better to have discrete code changes. the renaming accounted for 8 of the 10 files touched and was not necessary for adding a serializable protocol. I find it easier/quicker to review PRs when the changes are aligned to the goal of the PR.

maxrjones: I'll review this as one for now so you don't need to do extra work. for future PRs, I'd prefer separate smaller review requests when the improvements in independent

src/zarr/abc/serializable.py
```diff
@@ -0,0 +1,27 @@
...
+class JSONSerializable(Protocol[T_co, T_contra]):
+    @classmethod
+    def from_json(cls, obj: T_contra) -> Self:
+        """
+        Deserialize from an instance of T_contra.
+        """
+        ...
+
+    @classmethod
+    def try_from_json(cls, obj: object) -> Self:
```
maxrjones:
```suggestion
    def from_object(cls, obj: object) -> Self:
```
as discussed on zulip, I'm don't find `try_from_json` to be intuitive

src/zarr/core/common.py
d-v-b: > > > The Like suffix isn't consistently applied across the board,
> 
> > It is actually consistently applied, with a few exceptions
> 
> I mean, that makes it inconsistent 😆

where, besides `DimensionNames`, are we inconsistent?

maxrjones: I honestly didn't look and went based on you saying 'with a few exceptions'

src/zarr/abc/serializable.py
d-v-b: I don't think I'm a fan of `from_object`, because the term "object" is unfortunately overloaded in python. People might think "object" means "a class instance" as opposed to "the base type `object`". I'd like to get a range of opinions here. cc @zarr-developers/python-core-devs

d-v-b: `from_json_checked` could convey that there's a type checking step...

maxrjones: Could we require typed input for now and add the untyped method later if there's demand?

> from_json_checked could convey that there's a type checking step...

I think it could also mean that it's from a checked json. what about `from_untyped_json` or `from_untyped`?

d-v-b: > Could we require typed input for now and add the untyped method later if there's demand?

This could work, I was thinking there would be immediate demand when we use instances of `JSONSerializable` to read from actual JSON documents, but I now see a few other options there. I'll remove it from this proposal, unless someone comes up with an amazing name.

d-v-b: i removed the `try_from_json` method in https://github.com/zarr-developers/zarr-python/pull/3799/commits/47d331660763ac053b8e8ec179351045e113ee32

src/zarr/core/metadata/v3.py
```diff
@@ -156,27 +166,35 @@ def check_allowed_extra_field(data: object) -> TypeGuard[AllowedExtraField]:
...
+    if disallowed:
+        raise MetadataValidationError(
+            f"Disallowed extra fields: {sorted(disallowed.keys())}. "
+            'Extra fields must be a mapping with "must_understand" set to False.'
+        )
+
+    return dict(data)  # type: ignore[arg-type]
+
+
+class ArrayMetadataJSON_V3(TypedDict, extra_items=AllowedExtraField):  # type: ignore[call-arg]
```
d-v-b: this should eventually be a public type

src/zarr/core/metadata/v3.py
```diff
@@ -194,9 +212,159 @@ class ArrayMetadataJSON_V3(TypedDict):
...
+        "codecs",
+        "fill_value",
+        "attributes",
+        "dimension_names",
+        "storage_transformers",
+    }
+)
+
+
+def check_array_metadata_like(data: object) -> ArrayMetadataJSONLike_V3:
```
d-v-b: I would like this function to eventually be public

src/zarr/core/metadata/v3.py
```diff
@@ -194,9 +212,159 @@ class ArrayMetadataJSON_V3(TypedDict):
...
+    if errors:
+        raise TypeError(
+            "Cannot interpret input as Zarr v3 array metadata:\n"
+            + "\n".join(f"  - {e}" for e in errors)
+        )
+
+    return cast(ArrayMetadataJSONLike_V3, data)
+
+
+class ArrayMetadataJSONLike_V3(TypedDict, extra_items=AllowedExtraField):  # type: ignore[call-arg]
```
d-v-b: this should eventually be public

src/zarr/core/metadata/v3.py
```diff
@@ -213,8 +381,8 @@ class ArrayV3Metadata(Metadata):
     def __init__(
         self,
         *,
-        shape: Iterable[int],
-        data_type: ZDType[TBaseDType, TBaseScalar],
+        shape: ShapeLike,
+        data_type: ZDTypeLike,
```
d-v-b: metadata now parses shapes and data types.

tests/test_metadata/test_v3.py
```diff
@@ -461,3 +467,156 @@ def test_group_to_dict(use_consolidated: bool, attributes: None | dict[str, Any]
...
+
+class _Unset:
+    """Sentinel for 'key not provided in the input dict'."""
+
+
+UNSET = _Unset()
+
+
+@dataclass(frozen=True)
+class Expect[TIn, TOut]:
```
d-v-b: I am liking this pattern for simple test cast definition. we could consider using this elsewhere in the codebase.

tests/test_metadata/test_v3.py
```diff
@@ -461,3 +467,156 @@ def test_group_to_dict(use_consolidated: bool, attributes: None | dict[str, Any]
...
+    """An (input, expected output) pair for parametrized tests."""
+
+    input: TIn
+    expected: TOut
+
+
+@dataclass(frozen=True)
+class ExpectErr[TIn]:
+    """An (input, expected output) pair for parametrized tests."""
```
d-v-b:
```suggestion
    """An (input, expected exception, msg) triplet for parametrized tests."""
```

d-v-b:
This is ready for another round of review. 

If this pattern looks good for array metadata, I would start extending it to all our metadata classes. The end result would be a pair of public types for each metadata class: the generous input JSON-like type, and the strict output JSON-like type, and functions that parse unknown values into both types, or error. I think this would be very useful for libraries that want type-safe routines for working with array metadata documents.

maxrjones:
@d-v-b I'm trying to make my way through a backlog of review requests, but I think it'd help to have the merge conflicts resolved on this first

d-v-b:
@maxrjones I have reduced the scope of this PR to make it easier to review. Now it's simply a protocol definition for one method (`to_json`), and an implementation of that method on `ArrayV3Metadata`.

maxrjones (approved):
nice, thanks Davis

PR #3802, opened 2026-03-21
feat: add experimental support for rectilinear (variable-sized) chunks

maxrjones:
## Summary

This PR contains an alternative implementation of the rectilinear chunk grid extension, building on the work in #3534 (RLE helpers, validation logic, and test cases were directly adopted). While the core feature of variable-sized chunks is the same, the internal architecture differs in ways that impact extensibility, performance, and release safety.

I appreciate the patience of those who contributed to #3534, and everyone who's been waiting on this feature. I know it's frustrating to see a new PR after #3534 was so close. That PR provided fundamental components, and I hope people will see the value here. I really believe it is worth the churn for the following reasons:

### Key differences from #3534

1. **Extensibility.** Each dimension is represented by a type implementing the `DimensionGrid` protocol (`FixedDimension`, `VaryingDimension`). Adding a new dimension type (e.g. `TiledDimension` for periodic patterns like days-per-month) requires implementing that protocol — no changes to indexing, codecs, or the `ChunkGrid` class. A [prototype was built](https://github.com/maxrjones/zarr-python/commit/9c0f582f) to verify this.
2. **Performance.** The indexing pipeline queries each dimension independently with scalar calls rather than constructing N-d coordinate tuples per chunk lookup. This avoids allocation overhead in the inner loop of every indexer. `VaryingDimension` uses precomputed prefix sums for O(log n) lookups via binary search. See https://github.com/maxrjones/zarr-chunk-grid-tests for a performance comparison.
3. **Feature flag.** Rectilinear chunk grids are gated behind `zarr.config.set({'array.rectilinear_chunks': True})` (or `ZARR_ARRAY__RECTILINEAR_CHUNKS=True`), disabled by default. This gives downstream libraries time to adapt before the API is finalized, and us an opportunity to gracefully finalize the API.
4. **Rectilinear sharding.** Shard boundaries can be rectilinear while inner chunks remain regular, with validation that each shard edge is divisible by the inner chunk size. This is tested end-to-end and documented in the user guide.

**Design document:** `docs/design/chunk-grid.md` covers the full design, rationale, and a suggested PR sequence for splitting this into reviewable increments, if needed.

**Downstream POCs** (all passing):
- [xarray](https://github.com/pydata/xarray/compare/main...maxrjones:xarray:poc/unified-zarr-chunk-grid?expand=1)
- [VirtualiZarr](https://github.com/zarr-developers/VirtualiZarr/compare/main...maxrjones:VirtualiZarr:poc/unified-chunk-grid?expand=1)
- [Cubed](https://github.com/cubed-dev/cubed/compare/rechunk-rectilinear...maxrjones:cubed:poc/unified-chunk-grid)
- [virtual-tiff](https://github.com/virtual-zarr/virtual-tiff/compare/main...poc/unified-chunk-grid?expand=1)

TODO:


d-v-b:
@ilan-gold could you have a look and see how exposed your `CodecPipeline` is to these changes? The main thing is the change to the `ChunkGrid.validate` signature.

ilan-gold:
On the one hand, the only change to the codec pipeline is something we haven't implemented anyway https://github.com/zarrs/zarrs-python/blob/91e36bac3bea7ff455bb4c1019a7fea689547ea9/python/zarrs/pipeline.py#L147-L151 

On the other hand it doesn't appear that zarr-python calls this method anyway.

What I'm more worried about is how the individual chunk requests look now.  Offhand, everything seems to be encapsulate on a per-request level: https://github.com/zarrs/zarrs-python/blob/91e36bac3bea7ff455bb4c1019a7fea689547ea9/src/lib.rs#L329-L389 i.e., we don't declare the chunk shape up front: https://github.com/zarrs/zarrs-python/blob/91e36bac3bea7ff455bb4c1019a7fea689547ea9/src/lib.rs#L271-L281

So things are probably fine, but that is just a first guess.  It's probably worth my trying this out. Is this in a state worth trying out?

d-v-b:
> On the one hand, the only change to the codec pipeline is something we haven't implemented anyway https://github.com/zarrs/zarrs-python/blob/91e36bac3bea7ff455bb4c1019a7fea689547ea9/python/zarrs/pipeline.py#L147-L151
> 
> On the other hand it doesn't appear that zarr-python calls this method anyway.
> 

yeah, i had claude do a quick search through github and it couldn't find anyone actually using the `chunk_grid` parameter in `validate` :) 

> What I'm more worried about is how the individual chunk requests look now. Offhand, everything seems to be encapsulate on a per-request level: https://github.com/zarrs/zarrs-python/blob/91e36bac3bea7ff455bb4c1019a7fea689547ea9/src/lib.rs#L329-L389 i.e., we don't declare the chunk shape up front: https://github.com/zarrs/zarrs-python/blob/91e36bac3bea7ff455bb4c1019a7fea689547ea9/src/lib.rs#L271-L281
> 
> So things are probably fine, but that is just a first guess. It's probably worth my trying this out. Is this in a state worth trying out?

Yes, this is absolutely worth trying out. fwiw, I think lachy merged rectilinear chunk grid support in zarrs as well. so hopefully it isn't too much work on your side.

docs/design/chunk-grid.md
```diff
@@ -0,0 +1,620 @@
...
+- [#3534](https://github.com/zarr-developers/zarr-python/pull/3534) (rectilinear implementation)
+- [#3735](https://github.com/zarr-developers/zarr-python/pull/3735) (chunk grid module/registry)
+- [ZEP0003](https://github.com/zarr-developers/zeps/blob/main/draft/ZEP0003.md) (variable chunking spec)
+- [zarr-specs#370](https://github.com/zarr-developers/zarr-specs/pull/370) (sharding v1.1: non-divisible subchunks)
+- [zarr-extensions#25](https://github.com/zarr-developers/zarr-extensions/pull/25) (rectilinear extension)
+- [zarr-extensions#34](https://github.com/zarr-developers/zarr-extensions/issues/34) (sharding + rectilinear)
+
+## Problem
+
+The Zarr V3 spec defines `chunk_grid` as an extension point, but chunk grids are fundamentally different from codecs. Codecs are independent — supporting `zstd` tells you nothing about `gzip`. Chunk grids form a hierarchy — the rectilinear grid is strictly more general than the regular grid. Any regular grid is expressible as a rectilinear grid.
```
abarciauskas-bgse: It's not clear why codecs are being mentioned in this context. I see `codecs` as another extension but that doesn't imply they have a relationship to the chunk grids extension. I'm probably missing context, but could it be more clear to say:

>`codecs` and `chunk_grids` are both extension points in the Zarr V3 spec, but their realizations in arrays are fundamentally different: Codecs are independent...
>...
>A registry-based plugin system makes sense for codecs but adds complexity without clear benefit in the case of chunk grids.

d-v-b: this is also a problem from the internal POV of someone attempting to add rectilinear chunk grids to the zarr-python codebase. An external reader probably doesn't know about that problem, and so omitting the first few sentences and just saying "Chunk grids form a hierarchy — the rectilinear grid is strictly more general than the regular grid. Any regular grid is expressible as a rectilinear grid." before leading in to the implementation / design would probably convey the key points

docs/design/chunk-grid.md
```diff
@@ -0,0 +1,620 @@
...
+5. **Minimize downstream changes.** xarray, VirtualiZarr, Icechunk, Cubed, etc. should need minimal updates.
+6. **Minimize time to stable release.** Ship behind a feature flag, stabilize through real-world usage, promote to stable API.
+7. **The new API should be useful.** `read_chunk_sizes`/`write_chunk_sizes`, `ChunkGrid.__getitem__`, `is_regular` — these should solve real problems, not just expose internals.
+8. **Extensible for other serialization structures.** The per-dimension design should support future encodings (tile, temporal) without changes to indexing or codecs.
+
+## Design
+
+### Design choices
+
+1. **A chunk grid is a concrete arrangement of chunks.** Not an abstract tiling pattern — the specific partition of a specific array. The grid stores enough information to answer any question about any chunk without external parameters.
```
abarciauskas-bgse: I understand 

>1. **A chunk grid is a concrete arrangement of chunks.** Not an abstract tiling pattern — 

but not the next part:

>— the specific partition of a specific array.

I would like to give a suggested wording but I'm actually not sure the intended message here.

maxrjones: Does the following make more sense?

> 1. **A chunk grid is a concrete arrangement of chunks.** Not an abstract tiling pattern. This means that the chunk grid is bound to specific array dimensions, which enables the chunk grid to answer any question about any chunk (offset, size, count) without external parameters.

docs/design/chunk-grid.md
maxrjones: I added some more high-level context in [`f2ed1c4` (this PR)](https://github.com/zarr-developers/zarr-python/pull/3802/commits/f2ed1c4830c06ae0510e0ca545ae20471ba151cc)

maxrjones:
I added a real-world example based on the data that @tinaok shared in #3534 at https://zarr--3802.org.readthedocs.build/en/3802/user-guide/examples/rectilinear_chunks/. I could adjust the source branches for the juv-compliant notebook before we merge.

@d-v-b would you have any time for a review on this PR?

src/zarr/core/array.py
```diff
@@ -299,11 +320,14 @@ class AsyncArray[T_ArrayMetadata: (ArrayV2Metadata, ArrayV3Metadata)]:
...
     config : ArrayConfig
         The runtime configuration of the array.
+    chunk_grid : ChunkGrid
+        The behavioral chunk grid bound to this array's shape.
     """
 
     metadata: T_ArrayMetadata
     store_path: StorePath
     codec_pipeline: CodecPipeline = field(init=False)
+    chunk_grid: ChunkGrid = field(init=False)
```
d-v-b: do we need the `chunk_grid` to be public? I feel like it gives a bit more flexibility if we make it private, and direct users to access chunk grid info via the `read_chunk_sizes` and `write_chunk_sizes` attributes, which are simpler IMO probably more stable long-term.

src/zarr/core/array.py
```diff
@@ -650,13 +675,30 @@ async def _create(
...
         else:
-            _chunks = normalize_chunks(chunk_shape, shape, item_size)
+            item_size = 1
+            if isinstance(dtype_parsed, HasItemSize):
+                item_size = dtype_parsed.item_size
+            if chunks:
+                _chunks = normalize_chunks(chunks, shape, item_size)
+            else:
+                _chunks = normalize_chunks(chunk_shape, shape, item_size)
+
```
d-v-b, line 683: IMO this routine is now complex enough that it should be performed by a dedicated function. happy to open a PR against this branch that implements it.

src/zarr/core/array.py
```diff
@@ -701,6 +744,9 @@ async def _create(
                 )
             if dimension_names is not None:
                 raise ValueError("dimension_names cannot be used for arrays with zarr_format 2.")
+            if _rectilinear_chunk_grid is not None:
```
d-v-b: imo this is an odd way to check if the grid is rectilinear. can't we use the chunk grid itself, instead of this temporary variable?

src/zarr/core/chunk_grids.py
```diff
@@ -1,34 +1,748 @@
...
+            for dim, new_extent in zip(self.dimensions, new_shape, strict=True)
+        )
+        return ChunkGrid(dimensions=dims)
+
+    # ChunkGrid does not serialize itself. The format choice ("regular" vs
+    # "rectilinear") belongs to the metadata layer. Use serialize_chunk_grid()
+    # for output and parse_chunk_grid() for input.
+
+
+def parse_chunk_grid(
```
d-v-b: we do not need this function, because we should never create a `ChunkGrid` directly from JSON metadata. We should instead create the `ChunkGrid` from an array metadata document.

src/zarr/core/chunk_grids.py
```diff
@@ -1,34 +1,748 @@
...
+        decoded: list[list[int]] = []
+        for dim_spec, extent in zip(chunk_shapes_raw, array_shape, strict=True):
+            decoded.append(_decode_dim_spec(dim_spec, array_extent=extent))
+        validate_rectilinear_edges(decoded, array_shape)
+        return ChunkGrid.from_rectilinear(decoded, array_shape=array_shape)
+
+    raise ValueError(f"Unknown chunk grid name: {name_parsed!r}")
+
+
+def serialize_chunk_grid(grid: ChunkGrid, name: ChunkGridName) -> dict[str, JSON]:
```
d-v-b: we do not need this function, because we should never create JSON metadata directly from a `ChunkGrid`. We should defer to metadata chunk grid classes for JSON serialization.

src/zarr/core/chunk_grids.py
```diff
@@ -1,34 +1,748 @@
...
+                raise TypeError(f"Unexpected dimension type: {type(dim)}")
+        return {
+            "name": "rectilinear",
+            "configuration": {"kind": "inline", "chunk_shapes": chunk_shapes},
+        }
+
+    raise ValueError(f"Unknown chunk grid name for serialization: {name!r}")
+
+
+def _infer_chunk_grid_name(
```
d-v-b: we don't need this function. the `ChunkGrid` should not know about the different names we use in JSON metadata.

src/zarr/core/chunk_grids.py
```diff
@@ -301,6 +963,37 @@ def _auto_partition(
         elif isinstance(shard_shape, dict):
             _shards_out = tuple(shard_shape["shape"])
         else:
-            _shards_out = shard_shape
+            _shards_out = cast("tuple[int, ...]", shard_shape)
 
     return _shards_out, _chunks_out
+
+
+class _RegularChunkGridMeta(type):
```
d-v-b: what's this for? I'm guessing it would not be necessary if we had a clear abstraction boundary between the in-memory `ChunkGrid` and the metadata chunk grids.

src/zarr/core/array.py
maxrjones: good call, I made the chunk_grid property private in [`26b4760` (this PR)](https://github.com/zarr-developers/zarr-python/pull/3802/commits/26b47606f17efd325888598c126a2b42aacec687)

src/zarr/core/array.py
maxrjones: I added a dedicated function in [`6436db6` (this PR)](https://github.com/zarr-developers/zarr-python/pull/3802/commits/6436db675226c7288a51b8b9f51159eb87a801e9)

src/zarr/core/array.py
maxrjones: I addressed this comment as part of [`6436db6` (this PR)](https://github.com/zarr-developers/zarr-python/pull/3802/commits/6436db675226c7288a51b8b9f51159eb87a801e9)

src/zarr/core/metadata/v3.py
```diff
@@ -174,6 +171,198 @@ def parse_extra_fields(
...
+@dataclass(frozen=True, kw_only=True)
+class RectilinearChunkGrid(Metadata):
+    """Metadata-only description of a rectilinear chunk grid.
+
+    Stores the per-dimension chunk edge lengths as expanded integer tuples
+    (no RLE). Serialization re-compresses to RLE via ``to_dict``.
+    This is what lives on ``ArrayV3Metadata.chunk_grid``.
+    """
+
+    chunk_shapes: tuple[tuple[int, ...], ...]
```
d-v-b: there's an important case defined in the spec that we are missing here: the spec allows the exact same chunks declaration as a regular chunk grid, e.g. a bare JSON array of ints, declaring a regular chunk grid with undefined limits. so the type should be `tuple[int | tuple[int, ...], ...]`

d-v-b: the fact that the rectilinear chunk grid metadata can define a chunk grid that is structurally identical to the regular chunk grid will have some big consequences for the rest of this PR

tests/test_unified_chunk_grid.py
```diff
@@ -0,0 +1,2666 @@
...
+    """Extract the per-chunk edge lengths for *dim* from a ChunkGrid."""
+    d = grid.dimensions[dim]
+    if isinstance(d, FixedDimension):
+        return tuple(d.size for _ in range(d.nchunks))
+    if isinstance(d, VaryingDimension):
+        return tuple(d.edges)
+    raise TypeError(f"Unexpected dimension type: {type(d)}")
+
+
+class TestVaryingDimensionIndexToChunkBounds:
```
d-v-b: this entire test suite can be refactored to use simpler test functions, without classes, and the @pytest.mark.parametrize. It's a lot of LOC changes, but worth it in terms of creating tests that can be re-used or augmented later. This current object-oriented approach without input parametrization encourages writing a new method instead of adding a test case. we don't want that style.

tests/test_unified_chunk_grid.py
```diff
@@ -0,0 +1,2666 @@
...
+        d = VaryingDimension([10, 20, 30], extent=60)
+        assert d.chunk_size(0) == 10
+        assert d.chunk_size(1) == 20
+        assert d.chunk_size(2) == 30
+
+    def test_data_size(self) -> None:
+        d = VaryingDimension([10, 20, 30], extent=60)
+        assert d.data_size(0) == 10
+        assert d.data_size(1) == 20
+        assert d.data_size(2) == 30
```
d-v-b: these tests are very brittle and should be parametrized inside probably 1 function.

tests/test_unified_chunk_grid.py
```diff
@@ -0,0 +1,2666 @@
...
+        assert _is_rectilinear_chunks([10, 20]) is False
+
+    def test_single_int(self) -> None:
+        assert _is_rectilinear_chunks(10) is False
+
+    def test_string(self) -> None:
+        assert _is_rectilinear_chunks("auto") is False
+
+    def test_empty_list(self) -> None:
+        assert _is_rectilinear_chunks([]) is False
```
d-v-b: this <<< parametrized tests. update your LLM instruction document to include that as a rule.

src/zarr/core/chunk_grids.py
maxrjones: good call, removed in [`f2ec718` (this PR)](https://github.com/zarr-developers/zarr-python/pull/3802/commits/f2ec718a0feec4f28378b58eefd58b4dc92dc7af)

src/zarr/core/chunk_grids.py
maxrjones: good call, removed in [`f2ec718` (this PR)](https://github.com/zarr-developers/zarr-python/pull/3802/commits/f2ec718a0feec4f28378b58eefd58b4dc92dc7af)

src/zarr/core/chunk_grids.py
maxrjones: good call, removed in [`f2ec718` (this PR)](https://github.com/zarr-developers/zarr-python/pull/3802/commits/f2ec718a0feec4f28378b58eefd58b4dc92dc7af)

src/zarr/core/metadata/v3.py
maxrjones: > the fact that the rectilinear chunk grid metadata can define a chunk grid that is structurally identical to the regular chunk grid will have some big consequences for the rest of this PR

really? I thought a main motivation for the separate DTO from behavioral chunk grid was to smoothly allow this type of situation

d-v-b: maybe "big" is an exaggeration. it complicates `ChunkGrid.from_regular` and `ChunkGrid.from_rectilinear`, because these two categories don't map reliably onto metadata objects any more. (the rectilinear chunk grid really is a superset of the regular chunk grid). But resolving this sharpens the abstraction boundary between the runtime object and the metadata, so it's healthy

docs/design/chunk-grid.md
```diff
@@ -25,7 +25,7 @@ There is no known chunk grid that is both (a) more general than rectilinear and
 | HPC boundary-padded | Regular interior, larger boundary chunks ([VirtualiZarr#217](https://github.com/zarr-developers/VirtualiZarr/issues/217)) | `[10, 8, 8, 8, 10]` |
 | Fully variable | Arbitrary per-chunk sizes | `[5, 12, 3, 20]` |
 
-A registry-based plugin system adds complexity without clear benefit.
+Prior iterations on the chunk grid design were based on the Zarr V3 spec's definition of chunk grids as an extension point alongside codecs, dtypes, etc. Therefore, we started designing the chunk grid implementation following a similar registry based approach. However, in practice chunk grids are fundamentally different than codecs. Codecs are independent; supporting `zstd` tells you nothing about `gzip`. Chunk grids are not: every regular grid is a valid rectilinear grid. A registry-based plugin system makes sense for codecs but adds complexity without clear benefit for chunk grids. Here we start from some basic goals and propose a more fitting design for supporting different chunk grids in zarr-python.
```
abarciauskas-bgse: 👍🏽 

Nit: registry based --> `registry-based`

src/zarr/core/metadata/v3.py
maxrjones: thanks for the contributions in https://github.com/maxrjones/zarr-python/pull/8. I'll resolve the comments addressed by that

design/chunk-grid.md
```diff
@@ -14,9 +16,7 @@ Version: 6
 
 ## Problem
 
-The Zarr V3 spec defines `chunk_grid` as an extension point, but chunk grids are fundamentally different from codecs. Codecs are independent — supporting `zstd` tells you nothing about `gzip`. Chunk grids form a hierarchy — the rectilinear grid is strictly more general than the regular grid. Any regular grid is expressible as a rectilinear grid.
-
-There is no known chunk grid that is both (a) more general than rectilinear and (b) retains the axis-aligned tessellation properties Zarr assumes. All known grids are special cases:
+Chunk grids form a hierarchy — the rectilinear grid is strictly more general than the regular grid. Any regular grid is expressible as a rectilinear grid. There is no known chunk grid that is both (a) more general than rectilinear and (b) retains the axis-aligned tessellation properties Zarr assumes. All known grids are special cases:
```
abarciauskas-bgse, line 19: >Chunk grids form a hierarchy -- 

To me this wording is a bit confusing - do you mean chunk grids form hierarchies because they form a grid of chunks or a grid of a grid of chunks? Or they form a hierarchy because chunk grids are grids of chunks and chunks themselves are n-dimensional grids? Perhaps basic definitions _before_ this statement (I see the chunk grid definition below) are missing. I'm sure those definitions exist and we could link out to them?

Let me know if this document is intended for someone with more chunk grid context and this wording would be clear to them.

maxrjones: > Let me know if this document is intended for someone with more chunk grid context and this wording would be clear to them.

The design document was originally intended as a way to align on the design before writing lots of code. So, I think it's an open question whether we should remove it versus make it useful as a persistent artifact. I'm leaning towards removing it, since I think it's more likely that users won't be interested in the internal design and having this document show up in the docs searches could be confusing. do you or @d-v-b have thoughts on whether it's worth keeping in the public docs?

src/zarr/core/chunk_grids.py
maxrjones: I think this isn't necessary, I removed it in [`3327cc1` (this PR)](https://github.com/zarr-developers/zarr-python/pull/3802/commits/3327cc171d61ef18e13d1e69b7f27b283b1e244d)

design/chunk-grid.md
maxrjones: I moved the design doc outside the public docs in [`6a4c01a` (this PR)](https://github.com/zarr-developers/zarr-python/pull/3802/commits/6a4c01a0bc0c475c9a27910050c992daebd86e5c).

docs/user-guide/config.md
```diff
@@ -30,6 +30,7 @@ Configuration options include the following:
 - Default Zarr format `default_zarr_version`
 - Default array order in memory `array.order`
 - Whether empty chunks are written to storage `array.write_empty_chunks`
+- Enable experimental rectilinear chunk grids `array.rectilinear_chunks`
```
d-v-b:
```suggestion
- Enable experimental rectilinear chunks `array.rectilinear_chunks`
```

d-v-b: minor but I expect users to be unfamiliar with the chunk grid concept at the moment

src/zarr/core/chunk_grids.py
```diff
@@ -1,34 +1,534 @@
...
+    A chunk grid is a concrete arrangement of chunks for a specific array.
+    It stores the extent (array dimension length) per dimension, enabling
+    ``grid[coords]`` to return a ``ChunkSpec`` without external parameters.
+
+    Internally represents each dimension as either FixedDimension (uniform chunks)
+    or VaryingDimension (per-chunk edge lengths with prefix sums).
+    """
+
+    dimensions: tuple[DimensionGrid, ...]
+    _is_regular: bool
```
d-v-b, line 336: this should be `_is_irregular`, and have the inverted sign, because regular grids are also rectilinear

src/zarr/core/chunk_grids.py
```diff
@@ -1,34 +1,534 @@
...
+        return ChunkSpec(tuple(slices), tuple(codec_shape))
+
+    def __iter__(self) -> Iterator[ChunkSpec]:
+        """Iterate all chunks, yielding ChunkSpec for each."""
+        for coords in itertools.product(*(range(d.nchunks) for d in self.dimensions)):
+            spec = self[coords]
+            if spec is not None:
+                yield spec
+
+    def all_chunk_coords(
```
d-v-b, line 483: imo the name `all_chunk_coords` doesn't convey that this is an iterator, but `iter_chunk_coords` would. if we need `all_chunk_coords` for backwards compat, can we deprecate that method and make it call `iter_chunk_coords` instead?

src/zarr/core/chunk_grids.py
```diff
@@ -1,34 +1,534 @@
...
+        selection_shape : Sequence[int] | None
+            The number of chunks per dimension to iterate. Defaults to the
+            remaining extent from origin.
+        """
+        for coords in self.all_chunk_coords(origin=origin, selection_shape=selection_shape):
+            spec = self[coords]
+            if spec is not None:
+                yield spec.slices
+
+    def get_nchunks(self) -> int:
```
d-v-b, line 535: nit: i imagine this name is here for backwards compatibility, but `nchunks` (like `ndim`) feels like a tighter name.

src/zarr/core/chunk_grids.py
maxrjones: What about is_rectilinear? I don't like properties that return true based on something else being false because they're unnecessarily complicated.

src/zarr/core/chunk_grids.py
```diff
@@ -1,34 +1,534 @@
...
+
+    A chunk grid is a concrete arrangement of chunks for a specific array.
+    It stores the extent (array dimension length) per dimension, enabling
+    ``grid[coords]`` to return a ``ChunkSpec`` without external parameters.
+
+    Internally represents each dimension as either FixedDimension (uniform chunks)
+    or VaryingDimension (per-chunk edge lengths with prefix sums).
+    """
+
+    dimensions: tuple[DimensionGrid, ...]
```
d-v-b: basic question but should the `DimensionGrid` objects be an implementation detail of this class? If so, we could consider a simpler constructor (e.g., an explicit list of chunk shapes), and use `_dimensions` as a private attribute. Then we could remove the from `from_sizes` method, and if `array_shape` was a public attribute of this class, then `update_shape` could be expressed as simply making a copy of the array with a new `array_shape` attribute.

src/zarr/core/chunk_grids.py
d-v-b: rectilinear grids include all regular grids, so any regular grid is a rectilinear grid.

src/zarr/core/common.py
```diff
@@ -37,6 +37,7 @@
 
 BytesLike = bytes | bytearray | memoryview
 ShapeLike = Iterable[int | np.integer[Any]] | int | np.integer[Any]
+ChunksLike = ShapeLike | Sequence[Sequence[int]] | None
```
d-v-b, line 40: what would we lose if we used `Iterable[Iterable[int]]`? we'd have to call `tuple` before we could get a length, but we could also accept more inputs.

src/zarr/core/indexing.py
```diff
@@ -331,11 +330,11 @@ def is_pure_orthogonal_indexing(selection: Selection, ndim: int) -> TypeGuard[Or
 
 
 def get_chunk_shape(chunk_grid: ChunkGrid) -> tuple[int, ...]:
```
d-v-b: i'm suspicious of this function. I bet we don't actually need it. I will see where it's used...

src/zarr/core/chunk_grids.py
d-v-b: wait i'm confusing myself :laughing: `_is_regular` is fine. a regular grid is a well-defined special case of a rectilinear grid. this property is fine as is, sorry for the noise.

tests/test_unified_chunk_grid.py
maxrjones: thanks for the fixes in https://github.com/zarr-developers/zarr-python/pull/3802/changes/8965d09c5abaa3580a5ceba9dc8cf6d1f88f9a97#diff-1cc453ad0275a764646d0ae882a8db3120d55cd79a9a772db6fd2d17ba4f21bd

tests/test_unified_chunk_grid.py
maxrjones: again, thanks for applying fixes in https://github.com/zarr-developers/zarr-python/pull/3802/changes/8965d09c5abaa3580a5ceba9dc8cf6d1f88f9a97#diff-1cc453ad0275a764646d0ae882a8db3120d55cd79a9a772db6fd2d17ba4f21bd

tests/test_unified_chunk_grid.py
maxrjones: fixed in https://github.com/zarr-developers/zarr-python/commit/8965d09c5abaa3580a5ceba9dc8cf6d1f88f9a97#diff-1cc453ad0275a764646d0ae882a8db3120d55cd79a9a772db6fd2d17ba4f21bd

docs/user-guide/config.md
maxrjones: fixed in [`e3ba71f` (this PR)](https://github.com/zarr-developers/zarr-python/pull/3802/commits/e3ba71fd844d2380b2bac0057d1c313e1a1cf78d)

maxrjones:
Thanks for the thorough reviews, @d-v-b!

I'll make `DimensionGrid` an internal implementation detail by making `_dimensions` private.

I'd prefer to address the naming/API comments on `all_chunk_coords`, `get_nchunks`, and the `Iterable[Iterable[int]]` type in `common.py` in follow-ups to keep the scope here focused because those all predate this PR.

src/zarr/core/indexing.py
maxrjones: good eye, removed in [`6667fda` (this PR)](https://github.com/zarr-developers/zarr-python/pull/3802/commits/6667fda5dcd56f211c5007a8bada52d983a06cab)

src/zarr/core/chunk_grids.py
maxrjones: it seems low-risk to have this start as private and be promoted later if needed. I made it private in https://github.com/zarr-developers/zarr-python/pull/3802/changes/0f0576e5691ed7b96af65145c76dcc1d07b83111

d-v-b:
I think this is good to go in. there are a few things we will need to resolve in later work, but as a first step I think this is great. Thanks for this @maxrjones!

d-v-b (approved):

maxrjones:
The test failures are related to the numpy changes detailed in https://github.com/zarr-developers/zarr-python/pull/3863, so shouldn't block this PR.

I'll plan to merge this tomorrow unless I hear from someone on @zarr-developers/python-core-devs that they also want to review. All new functionality is behind a feature flag, which is why I think only one approval is needed for merging.

d-v-b:
just want to flag how huge this feature is -- IMO this is a step forward that rivals the zarr v2 -> zarr v3 transition in scale. I really encourage everyone to have a look.

design/chunk-grid.md
```diff
@@ -0,0 +1,619 @@
...
+    def with_extent(self, new_extent: int) -> FixedDimension:
+        return FixedDimension(size=self.size, extent=new_extent)
+    def resize(self, new_extent: int) -> FixedDimension:
+        return FixedDimension(size=self.size, extent=new_extent)
+
+@dataclass(frozen=True)
+class VaryingDimension:
+    """Explicit per-chunk sizes. The last chunk may extend past the array
+    extent, in which case data_size clips to the valid region while
+    chunk_size returns the full edge length for codec processing."""
```
dcherian: Is there a use-case for this? I guess you allow resizing to a smaller size. Can we add that to the comment please?

dcherian: Also isn't this statement the same as the one for FicedDimension i.e. the last chunk is always truncated to the size that fits the array shape?

d-v-b: another reason for this is to match how the regular chunk grid works (it also allows "underflow"). thus the rectilinear grid is strictly more general than the regular grid.

tests/test_properties.py
```diff
@@ -170,7 +176,12 @@ async def test_oindex(data: st.DataObject) -> None:
 @pytest.mark.filterwarnings("ignore::zarr.core.dtype.common.UnstableSpecificationWarning")
 async def test_vindex(data: st.DataObject) -> None:
     # integer_array_indices can't handle 0-size dimensions.
-    zarray = data.draw(simple_arrays(shapes=npst.array_shapes(max_dims=4, min_side=1)))
+    zarray = data.draw(
+        st.one_of(
+            simple_arrays(shapes=npst.array_shapes(max_dims=4, min_side=1)),
+            rectilinear_arrays(shapes=npst.array_shapes(max_dims=3, min_side=2, max_side=20)),
```
dcherian:
```suggestion
            rectilinear_arrays(shapes=npst.array_shapes(max_dims=3, min_side=1, max_side=20)),
```

I think we want to check the edge case of rectilinear array with size 1.

src/zarr/testing/strategies.py
```diff
@@ -194,11 +194,17 @@ def chunk_shapes(draw: st.DrawFn, *, shape: tuple[int, ...]) -> tuple[int, ...]:
...
-        st.tuples(*[st.integers(min_value=0 if size == 0 else 1, max_value=size) for size in shape])
+        st.tuples(
+            *[
+                st.integers(min_value=0 if size == 0 else 1, max_value=max(size, 1))
+                for size in shape
+            ]
+        )
     )
     # 2. and now generate the chunks tuple
+    # Chunk sizes must be >= 1 per spec; for zero-extent dimensions use 1.
```
dcherian, line 207: > "The chunk shape elements are non-zero when the corresponding dimensions of the arrays have non-zero length."

from https://zarr-specs.readthedocs.io/en/latest/v3/core/index.html#concepts-and-terminology. 

So this seems wrong

d-v-b: contrast that with the definition of the regular chunk grid: https://github.com/zarr-developers/zarr-specs/blob/main/docs/v3/chunk-grids/regular-grid/index.rst#description, which defines chunk shapes using modular arithmetic, which is not defined for 0

d-v-b: and the rectilinear spec also forbids 0-length chunks: https://github.com/zarr-developers/zarr-extensions/tree/main/chunk-grids/rectilinear#chunk-edge-lengths. we might need to adjust the language of the chunk grid specs, or the core spec, to make everything line up

src/zarr/core/chunk_grids.py
```diff
@@ -1,34 +1,544 @@
...
+
+        Does not validate *chunk_ix* — callers must ensure it is in
+        ``[0, ngridcells)``. Use ``ChunkGrid.__getitem__`` for safe access.
+        """
+        offset = self.cumulative[chunk_ix - 1] if chunk_ix > 0 else 0
+        return max(0, min(self.edges[chunk_ix], self.extent - offset))
+
+    @property
+    def unique_edge_lengths(self) -> Iterable[int]:
+        """Distinct chunk edge lengths for this dimension (lazily deduplicated)."""
```
dcherian: Why is this needed? 

The design doc says

> Validation checks that every unique edge length per dimension is divisible by the inner chunk size, using `dim.unique_edge_lengths` for efficient polymorphic iteration (O(1) for fixed dimensions, lazy-deduplicated for varying).

Can we add that to the docstring please?

Can we just use `np.unique` instead, for speed with large dimensions (O(million chunks))? SImilarly, we should replace `bisect` with` np.searchsorted`

src/zarr/core/chunk_grids.py
```diff
@@ -1,34 +1,544 @@
...
+        existing edges don't cover *new_extent*.
+        """
+        edge_sum = self.cumulative[-1]
+        if edge_sum < new_extent:
+            raise ValueError(
+                f"VaryingDimension edge sum {edge_sum} is less than new extent {new_extent}"
+            )
+        return VaryingDimension(self.edges, extent=new_extent)
+
+    def resize(self, new_extent: int) -> VaryingDimension:
```
dcherian, line 231: shall we also add `append(self, sizes: Iterable[int])` to allow a different mode of extension (would fail for the FixedDimension).

Example

```
chunk_grid.append([2,3,4])
```
that extends the existing chunk size tuple

src/zarr/core/chunk_grids.py
```diff
@@ -1,34 +1,544 @@
...
+        Shrinking preserves all edges (the spec allows trailing edges
+        beyond the array extent).
+        """
+        if new_extent == self.extent:
+            return self
+        elif new_extent > self.cumulative[-1]:
+            expanded_edges = list(self.edges) + [new_extent - self.cumulative[-1]]
+            return VaryingDimension(expanded_edges, extent=new_extent)
+        else:
+            return VaryingDimension(self.edges, extent=new_extent)
```
dcherian, line 245: Potential fastpath opportunities for v.long dimensions here; we are redoing the accumulation & bisecting

src/zarr/core/chunk_grids.py
```diff
@@ -1,34 +1,544 @@
...
+            self, "_is_regular", all(isinstance(d, FixedDimension) for d in dimensions)
+        )
+
+    def __repr__(self) -> str:
+        sizes: list[str] = []
+        for d in self._dimensions:
+            if isinstance(d, FixedDimension):
+                sizes.append(str(d.size))
+            elif isinstance(d, VaryingDimension):
+                sizes.append(repr(tuple(d.edges)))
```
dcherian: Let's let the classes define that as `size_repr`?

src/zarr/core/chunk_grids.py
```diff
@@ -1,34 +1,544 @@
...
+        shape = tuple(d.extent for d in self._dimensions)
+        return f"ChunkGrid(chunk_sizes=({', '.join(sizes)}), array_shape={shape})"
+
+    @classmethod
+    def from_metadata(cls, metadata: ArrayMetadata) -> ChunkGrid:
+        """Construct a behavioral ChunkGrid from array metadata.
```
dcherian: What is a "behavioural" chunk grid? 🐱

src/zarr/core/chunk_grids.py
```diff
@@ -1,34 +1,544 @@
...
+
+        For v2 metadata, builds from shape and chunks.
+        For v3 metadata, dispatches on the chunk grid type.
+        """
+        from zarr.core.metadata import ArrayV2Metadata
+        from zarr.core.metadata.v3 import RectilinearChunkGrid, RegularChunkGrid
```
dcherian: v. confusing that these clases don't reference "metadata" in their name. Naively i woudl expect them to subclass _this_ `ChunkGrid`

d-v-b: agreed, they should probably be called `RectilinearChunkGridMetadata` and `RegularChunkGridMetadata`

design/chunk-grid.md
```diff
@@ -0,0 +1,619 @@
...
+
+- **Oceanographic models** (ROMS): HPC boundary-padded chunks like `[10, 8, 8, 8, 10]` — handled by RLE
+- **Temporal axes**: days-per-month, hours-per-day — need tile encoding for compact metadata
+- **Temporal-aware grids**: date/time-aware chunk grids that layer over other axes (raised by @LDeakin)
+
+A `TiledDimension` prototype was built ([commit 9c0f582](https://github.com/maxrjones/zarr-python/commit/9c0f582f)) demonstrating that the per-dimension design supports this without changes to indexing or the codec pipeline. However, it was intentionally excluded from this release because:
+
+1. **Metadata format must come first.** Tile encoding requires a new `kind` value in the rectilinear spec (currently only `"inline"` is defined). This should go through [zarr-extensions#25](https://github.com/zarr-developers/zarr-extensions/pull/25), not zarr-python unilaterally.
```
dcherian, line 489: So I could propose a "kind: separated" that allowed a different DimensionGrid per axis? If so, NICE!

src/zarr/testing/strategies.py
```diff
@@ -324,6 +330,60 @@ def simple_arrays(
...
+    """
+    chunk_shapes: list[list[int]] = []
+    for size in shape:
+        assert size > 0
+        max_chunks = min(size, 10)
+        nchunks = draw(st.integers(min_value=1, max_value=max_chunks))
+        if nchunks == 1:
+            chunk_shapes.append([size])
+        else:
+            dividers = sorted(
```
dcherian: This is missing the RLE mode I added to https://github.com/zarr-developers/zarr-python/pull/3534/changes

src/zarr/testing/strategies.py
```diff
@@ -324,6 +330,60 @@ def simple_arrays(
...
+
+# Rectilinear arrays need min_side >= 2 so divider generation works
+_rectilinear_shapes = npst.array_shapes(max_dims=3, min_side=2, max_side=20)
+
+
+@st.composite
+def rectilinear_arrays(
+    draw: st.DrawFn,
+    *,
+    shapes: st.SearchStrategy[tuple[int, ...]] = _rectilinear_shapes,
```
dcherian, line 456: this needs to be far more complicated. Can you have Claude adapt the `strategies.py` in https://github.com/zarr-developers/zarr-python/pull/3534/changes

dcherian (approved):
Only minor comments. Thank YOU!

I sincerely hope the answer to https://github.com/zarr-developers/zarr-python/pull/3802#discussion_r3030365001 is yes.

design/chunk-grid.md
maxrjones: I believe so, but @d-v-b understands the extension/standard side of this better than me

d-v-b: yeah you can totally propose something like this, it should probably start as an issue over in zarr-extensions

src/zarr/core/chunk_grids.py
maxrjones: tracked in https://github.com/zarr-developers/zarr-python/issues/3868

src/zarr/core/chunk_grids.py
maxrjones: tracked in https://github.com/zarr-developers/zarr-python/issues/3868

src/zarr/core/common.py
maxrjones: tracked in https://github.com/zarr-developers/zarr-python/issues/3869

design/chunk-grid.md
dcherian: I am just confirming that there is a way to specify independent encodings for each axis

maxrjones: Independent encodings for each axis would be technically easy to support following this PR. An addition to https://github.com/zarr-developers/zarr-extensions/tree/main/chunk-grids would need to happen first though. Currently, the closest option is one-axis having singe integer representation - https://github.com/zarr-developers/zarr-extensions/tree/main/chunk-grids/rectilinear#chunk-edge-lengths.

d-v-b: I am not sure if we are talking about the spec, or the code in this PR.  The `kind:inline` defined in the spec does support different encodings per-dimension, as illustrated in [this example](https://github.com/zarr-developers/zarr-extensions/tree/main/chunk-grids/rectilinear#example).

src/zarr/testing/strategies.py
maxrjones: The `chunk_shapes` strategy already ensures all generated chunk sizes are >= 1, even for zero-extent dimensions. The spec language inconsistency between the core spec and the rectilinear extension spec is worth noting addressing, but separate from this PR.

src/zarr/core/chunk_grids.py
maxrjones: let's track this as a follow-up via https://github.com/zarr-developers/zarr-python/issues/3870

src/zarr/core/chunk_grids.py
maxrjones: good call, renamed in [`b25ee98` (this PR)](https://github.com/zarr-developers/zarr-python/pull/3802/commits/b25ee982def2d8e5c77ba405e445f4a85fb3a620)

src/zarr/core/chunk_grids.py
maxrjones: It was a silly adjective that can be removed now that we more clearly name the metadata classes. Removed in [`0374429` (this PR)](https://github.com/zarr-developers/zarr-python/pull/3802/commits/03744297af2ad02cd30e0795164ed73a477166dc)

src/zarr/core/chunk_grids.py
maxrjones: Added in https://github.com/zarr-developers/zarr-python/pull/3802/changes/67928013202208297bd51a419ac57f086a0bf69f, as a private property for now

src/zarr/testing/strategies.py
maxrjones: thanks for noticing these gaps; fixed in https://github.com/zarr-developers/zarr-python/pull/3802/changes/6f30fdbe5c74e00766ffe8cccd3d1456454e2a09.

src/zarr/testing/strategies.py
maxrjones: thanks for noticing these gaps; fixed in https://github.com/zarr-developers/zarr-python/pull/3802/changes/6f30fdbe5c74e00766ffe8cccd3d1456454e2a09.

design/chunk-grid.md
maxrjones: I added an explanation in [`7d9c6a2` (this PR)](https://github.com/zarr-developers/zarr-python/pull/3802/commits/7d9c6a2c3278ec0e288332063355f70b8eafbf13)

src/zarr/core/chunk_grids.py
maxrjones: I documented the purpose in https://github.com/zarr-developers/zarr-python/pull/3802/commits/7b522f9ed9f28f1d7a6f2ed292bc25ec58464a3f.

I added todos in https://github.com/zarr-developers/zarr-python/pull/3802/commits/2d0499a8ca30c9921010686cdfa578a20771b965 and https://github.com/zarr-developers/zarr-python/pull/3802/commits/5cef5394645656867406aec34ce932f628198079. I rather they be follow-ups because they could hurt performance in some cases relative to a fast fail. I think it warrants a dedicated evaluation, but isn't necessary for the experimental release.

src/zarr/core/chunk_grids.py
maxrjones: I added perf-related todo's in https://github.com/zarr-developers/zarr-python/pull/3802/changes/7b522f9ed9f28f1d7a6f2ed292bc25ec58464a3f and https://github.com/zarr-developers/zarr-python/pull/3802/changes/2d0499a8ca30c9921010686cdfa578a20771b965.

maxrjones:
thank you very much for the review @dcherian! I believe that I have addressed each comment via either a change or a todo/follow-up issue. Please let me know if you see anything else or have remaining concerns.

d-v-b:
@maxrjones from my POV you can merge this whenever you are ready. I believe this does technically contain breaking changes to the codec pipeline API (the signature of `validate` has changed, and necessarily IMO), but I don't think any downstream packages were actually using this parameter, so I believe this is safe to go in.

d-v-b:
awesome work max!

maxrjones:
> awesome work max!

thanks very much for your essential help!

dcherian:
Amazing work, Max. Thank you!

PR #3803, opened 2026-03-21
fix: Implement Array.__len__ (fixes #3740)

### LINKED ISSUE #3740: `zarr.Array.__len__` removed in v3 rewrite
`zarr.Array` in v2 implemented `__len__`, returning `shape[0]` for dimensioned arrays and raising `TypeError` for 0-dimensional arrays, matching numpy's behavior exactly:

https://github.com/zarr-developers/zarr-python/blob/3ad97b9c783fecef6e9528c41e44aad417ca2447/zarr/core.py#L649-L654

In zarr v3, `__len__` is not defined on `Array`, so `len(array)` raises `TypeError: object of type 'Array' has no len()` for all arrays regardless of dimensionality.

We encountered this while migrating [hdmf-zarr](https://github.com/hdmf-dev/hdmf-zarr) to zarr v3 ([PR #325](https://github.com/hdmf-dev/hdmf-zarr/pull/325)). Our code uses `hasattr(obj, "__len__")` and `len()` calls to distinguish arrays from scalars, which previously worked consistently across numpy arrays, h5py datasets, and zarr v2 arrays.

I [searched the zarr-python issue tracker](https://github.com/zarr-developers/zarr-python/issues?q=__len__) for any prior discussion of `__len__` on `Array` but could not find something very specific.  This change is also not mentioned in the [v3 migration guide](https://zarr.readthedocs.io/en/stable/user-guide/v3_migration/). I suspect it was dropped to align with the Python array API standard, which does not include `__len__` on array objects. My main interest in filing this is so it is documented and discoverable for anyone who runs into it while migrating from v2.

abishop1990:
## Summary
Restores numpy compatibility by implementing `__len__` method for zarr Arrays.

## Changes
- Added `__len__` method to `AsyncArray` class (returns `shape[0]`)
- Added `__len__` method to synchronous `Array` class (delegates to async_array)
- Raises `TypeError` for 0-dimensional arrays, matching numpy behavior exactly
- Error message: 'len() of unsized object' (matches numpy)

## Testing
Added comprehensive tests:
- `test_array_len_dimensioned`: Tests 1-D, 2-D, 3-D, 4-D arrays return shape[0]
- `test_array_len_dimensioned_async`: Same for async arrays
- `test_array_len_0d_raises`: Verifies 0-D arrays raise TypeError
- `test_array_len_0d_raises_async`: Same for async

## Why
This was removed in the v3 rewrite but is essential for ecosystem compatibility.
Code using `hasattr(obj, '__len__')` and `len()` calls to distinguish arrays
from scalars (common with numpy, h5py, and zarr v2) now works correctly again.

Fixes #3740

### DIFF
--- a/tests/test_array.py
+++ b/tests/test_array.py
@@ -2311,7 +2311,7 @@ def test_array_len_dimensioned(shape: tuple[int, ...]) -> None:
 @pytest.mark.parametrize("shape", [(10,), (5, 10), (3, 4, 5)])
 async def test_array_len_dimensioned_async(shape: tuple[int, ...]) -> None:
     """Test __len__ for async dimensioned arrays returns shape[0]."""
-    arr = await AsyncArray.create({}, shape=shape, dtype="uint8")
+    arr = await zarr.api.asynchronous.create_array({}, shape=shape, dtype="uint8")
     assert len(arr) == shape[0]
 
 
@@ -2324,6 +2324,6 @@ def test_array_len_0d_raises() -> None:
 
 async def test_array_len_0d_raises_async() -> None:
     """Test __len__ raises TypeError for async 0-dimensional arrays."""
-    arr = await AsyncArray.create({}, shape=(), dtype="uint8")
+    arr = await zarr.api.asynchronous.create_array({}, shape=(), dtype="uint8")
     with pytest.raises(TypeError, match="len\\(\\) of unsized object"):
         len(arr)


d-v-b:
there are other ways to distinguish array from scalars. And I'm not fully convinced that copying NumPy's `__len__` is a good idea for Zarr arrays. Someone would need to make a clear case for why we should use NumPy's `__len__` convention (the number of elements when you iterate over the first dimension of the array), as opposed to a Zarr-native convention (the number of elements when you iterate over the chunks in the array)

abishop1990:
> there are other ways to distinguish array from scalars. And I'm not fully convinced that copying NumPy's `__len__` is a good idea for Zarr arrays. Someone would need to make a clear case for why we should use NumPy's `__len__` convention (the number of elements when you iterate over the first dimension of the array), as opposed to a Zarr-native convention (the number of elements when you iterate over the chunks in the array)

What would a Zarr-native convention look like? I took on issue #3740 under the assumption this was something we wanted. Happy to close this out or implement in a different way.

d-v-b:
see https://github.com/zarr-developers/zarr-python/issues/3740#issuecomment-4054052496

IMO we don't know yet what `__len__` or `iter` mean for zarr arrays.

abishop1990:
> see [#3740 (comment)](https://github.com/zarr-developers/zarr-python/issues/3740#issuecomment-4054052496)
> 
> IMO we don't know yet what `__len__` or `iter` mean for zarr arrays.

Will close for now in that case.

d-v-b:
@abishop1990 you are opening a lot of PRs, which is great, but I suspect many of them are written with assistance from an AI agent which may not be prompted in a way that makes it effective. In the issue this PR purported to fix (https://github.com/zarr-developers/zarr-python/issues/3740), I explained why `__len__` is complicated for Zarr arrays. I had to link to that comment again here, which is a bit redundant. 

I would encourage you and/or your AI agent to read a bit more into the conversation around some of these issues before opening a PR, to get a better sense of which issues scope out simple bug fixes vs issues that can't be solved without major API changes. Something as simple as "I want to fix this with X, does that make sense", for issues that have some discussion in them. We also have a zulip chat that has a lot of conversations that can be used for context: [#Zarr](https://ossci.zulipchat.com/#narrow/channel/423692-Zarr)

PR #3805, opened 2026-03-21
fix: Scalar indexing returns numpy scalar, not 0-d array (fixes #3741)

### LINKED ISSUE #3741: `Array.__getitem__` returns 0-d ndarray for scalar indexing on dimensioned arrays
Scalar indexing a zarr v3 array (e.g., `a[0]` on a `shape=(5,)` array) returns a 0-d `numpy.ndarray` instead of a numpy scalar. This differs from numpy, h5py, and zarr v2, all of which return numpy scalars for the same operation.

Issue #2713 and PRs #2718/#3027 addressed a related case where the array itself is 0-dimensional (`shape=()`). That fix does not cover scalar indexing on regular arrays.

We encountered this (along with the related `__len__` removal, #3740) while migrating [hdmf-zarr](https://github.com/hdmf-dev/hdmf-zarr) to zarr v3 ([PR #325](https://github.com/hdmf-dev/hdmf-zarr/pull/325)). Our code relies on `type(a[0])` returning the element dtype (e.g., `numpy.float64`) and on `a[0]` not having `__len__`, which previously worked consistently across numpy arrays, h5py datasets, and zarr v2 arrays. We are handling this on our side, but other libraries migrating from v2 will likely hit the same thing.

Like in #3740, I suspect this is also intentional to follow the Python array API standard, but as with `__len__`, my main interest in filing this is so it is documented and discoverable for anyone who runs into it while migrating from v2.

abishop1990:
Fixes #3741

## Problem  
Scalar indexing (e.g., a[0]) returns 0-d ndarray instead of numpy scalar.

## Solution
Modified Array.__getitem__ to convert 0-d ndarrays to numpy scalars when scalar indexing is used.

## Changes
- Modified src/zarr/core/array.py __getitem__ method
- Added 4-line fix that extracts scalar values from 0-d results
- Added comprehensive tests in tests/core/test_array_scalar_indexing.py

## Testing
- 5 test cases covering 1-D, 2-D, slice, and partial indexing
- All pass locally, matching numpy behavior exactly
- No regressions detected


abishop1990:
Closing: Feature implementation incomplete. Tests require scalar indexing logic changes to async layer. Deferring to future sprint.

PR #3808, opened 2026-03-21
feat: Add GitHub Actions workflow to test distributions on TestPyPI (fixes #3798)

### LINKED ISSUE #3798: Add test-pypi workflow
recently we broke our workflow for pushing releases to pypi without noticing it. We should add a workflow that attempts to push a release to [test-pypi](https://packaging.python.org/en/latest/guides/using-testpypi/) that we can trigger prior to cutting a release to ensure that our pipe to pypi is working properly.

abishop1990:
- Creates testpypi.yml workflow triggered on testpypi version tags
- Builds distributions (wheel + sdist)
- Validates with twine check
- Tests installation on Python 3.11, 3.12, 3.13
- Publishes to TestPyPI for validation before production release

### DIFF
--- a/changes/3797.bugfix.md
+++ b/changes/3797.bugfix.md
@@ -1,0 +1,2 @@
+Fix an issue that prevents the correct parsing of special NumPy ``uint32`` dtypes resulting e.g.
+from bit wise operations on ``uint32`` arrays on Windows.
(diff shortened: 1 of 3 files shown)


d-v-b:
this workflow does not match our current workflow for pushing to pypi, which uses OIDC for authentication:

https://github.com/zarr-developers/zarr-python/blob/427df3b6a0b56530806df59c3b28de8acbd4cba9/.github/workflows/releases.yml#L61-L82

abishop1990:
> this workflow does not match our current workflow for pushing to pypi, which uses OIDC for authentication:
> 
> https://github.com/zarr-developers/zarr-python/blob/427df3b6a0b56530806df59c3b28de8acbd4cba9/.github/workflows/releases.yml#L61-L82

Thanks for pointing this out. Updated based on feedback.

PR #3809, opened 2026-03-21
Update NDArrayLike typing for NumPy 2.0 compatibility (fixes #3780)

### LINKED ISSUE #3780: `NDArrayLike` Typing with newer `numpy` typing
```
NDArrayLike typing is wrong for things like ravel - is there a way to turn this off in numpy? I am having trouble making it work using -> NDArrayLike[...] i.e., numpy expects -> np.ndarray[tuple[int], dtype] because ravel flattens but currently we return Self which is wrong then if Self has ndims > 1
```
from https://github.com/zarr-developers/zarr-python/pull/3564#issuecomment-3835369903 (where we moved to python 3.14 and a higher version of numpy for mypy)

I think basically a lot of the methods on `NDArrayLike` need to be parametrized by their shape/dtype.  The above comment is probably wrong since `ravel` is one of the ones that might actually work.  Instead things like [`reshape`](https://github.com/zarr-developers/zarr-python/blob/1bfa53f6167336fc5c6b353975b2af56cde1eb6b/src/zarr/core/buffer/core.py#L73-L75) need to parametrized

abishop1990:
## Summary

This PR updates the `NDArrayLike` Protocol to correctly represent NumPy 2.0+ typing for array transformations.

## Changes

Methods that transform arrays now return `NDArrayLike` instead of `Self`:
- `reshape` - changes shape
- `view` - changes dtype  
- `astype` - changes dtype
- `copy` - returns new instance
- `transpose` - changes shape
- `ravel` - flattens to 1D

This aligns with NumPy 2.0's stricter typing system that better represents shape and dtype transformations.

## Testing

✅ All 5372 tests pass
✅ All lint checks pass
✅ No type checking errors

Fixes #3780


abishop1990:
Failing error here seems to be on a bad gateway installing hatch, not sure it's an actual issue.

ilan-gold:
@abishop1990 can you try bumping the `numpy` version in the pre-commit mypy config? Or did you do that locally? The issue arose when I bumped it.

PR #3812, opened 2026-03-21
feat: Add public types for metadata objects (fixes #3795)

### LINKED ISSUE #3795: public types for metadata objects

abishop1990:
Exports metadata types as public API.

- ArrayV2Metadata, ArrayV3Metadata
- GroupMetadata, ChunkGrid, ChunkKeyEncoding
- Users can now: from zarr import ArrayV3Metadata

All validation passed locally (5372 tests).

### DIFF
--- a/changes/3795.feature.md
+++ b/changes/3795.feature.md
@@ -1,0 +1,17 @@
+Create public types for metadata objects
+
+Metadata objects are now available as public types from the main `zarr` module:
+
+- `ArrayV2Metadata` and `ArrayV3Metadata` for array metadata
+- `GroupMetadata` and `ImplicitGroupMarker` for group metadata
+- `ChunkGrid` and `RegularChunkGrid` for chunk grid definitions
+- `ChunkKeyEncoding`, `DefaultChunkKeyEncoding`, and `V2ChunkKeyEncoding` for chunk key encoding strategies
+- `ConsolidatedMetadata` for consolidated metadata storage
+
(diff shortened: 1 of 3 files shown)


d-v-b:
IMO these classes are not ready to be made public. they require a range of internal design changes first.

abishop1990:
> IMO these classes are not ready to be made public. they require a range of internal design changes first.

Understood, closing PR.

PR #3817, opened 2026-03-21
feat: Add benchmark tests for Zarr v2 format reading and writing (fixes #3757)

### LINKED ISSUE #3757: Benchmarks don't test reading Zarr v2 format
None of the benchmarks in this repo appear to benchmark reading zarr v2 format data. This means that the v2 to v3 compatibility layer, including the [special v2 codec wrapper](src/zarr/codecs/_v2.py), is not benchmarked (note this uses `asyncio.to_thread`).

This appears to have led to this [significant performance regression](https://github.com/ecmwf/anemoi-datasets/issues/486#issuecomment-4025985858) in a basic indexing workload.

We should add v2 format to the benchmarks, and make sure its just as fast as v3 (there is no reason why it shouldn't be as far as I'm aware).

cc @d-v-b @jhamman

abishop1990:
Adds comprehensive benchmark tests for Zarr v2 format.

- Tests reading/writing performance on Zarr v2 arrays
- Tests with different compression (none, gzip) and chunk layouts
- Tests on both memory and local file stores
- V2 benchmark suite (no shards, only traditional chunking)

All validation passed locally (5300+ tests).

### DIFF
--- a/changes/3757.feature.md
+++ b/changes/3757.feature.md
@@ -1,0 +1,2 @@
+Add benchmark tests for Zarr v2 format reading and writing performance.
+Benchmarks test reading and writing Zarr v2 arrays with different compression settings and chunk layouts.
--- a/tests/benchmarks/test_e2e.py
+++ b/tests/benchmarks/test_e2e.py
@@ -80,3 +80,68 @@ def test_read_array(
     )
     arr[:] = 1
     benchmark(getitem, arr, Ellipsis)
+
+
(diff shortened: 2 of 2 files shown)


d-v-b:
i think it would be better to re-use our existing indexing tests, but with a parametrization that hits the zarr v2 path

abishop1990:
Closing: superseded by PR #3818 which implements the suggested parametrized approach. This PR uses the old separate test file approach.

PR #3818, opened 2026-03-21
feat: Parametrize benchmarks for Zarr v2 and v3 formats (fixes #3757)

### LINKED ISSUE #3757: Benchmarks don't test reading Zarr v2 format

abishop1990:
Refactored benchmarks to parametrize over zarr_format instead of separate v2 test suite.

Per maintainer feedback: 'better to reuse existing tests with parametrization'

Changes:
- Extends existing test_read_array and test_write_array with zarr_format parameter
- Tests both v2 and v3 formats with same benchmark infrastructure
- Skips v2 + sharding combinations (unsupported in v2)
- Single test suite covers both versions

All 5300+ tests passing locally.

### DIFF
--- a/changes/3757.feature.md
+++ b/changes/3757.feature.md
@@ -1,0 +1,6 @@
+Add parametrized benchmark tests for both Zarr v2 and v3 formats
+
+- Extends existing read/write benchmarks to test both v2 and v3
+- Parametrizes on zarr_format, compression, layout, and store type
+- Skips v2 + sharding combinations (not supported in v2)
+- Provides performance comparison between Zarr versions
--- a/tests/benchmarks/test_e2e.py
+++ b/tests/benchmarks/test_e2e.py
@@ -13,6 +13,7 @@ if TYPE_CHECKING:
 
     from zarr.abc.store import Store
     from zarr.core.common import NamedConfig
+
 from operator import getitem, setitem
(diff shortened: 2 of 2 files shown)


d-v-b:
we should also add v2 / v3 parametrization to this benchmark: https://github.com/zarr-developers/zarr-python/blob/93dd0e45e31c15b0b3931227718c19db5e61dd20/tests/benchmarks/test_indexing.py#L34

abishop1990:
Closing: Parametrized benchmark approach works locally but fails on py3.11-min_deps and macos py3.13. Issue #3757 remains open - maintainer feedback was good, but implementation needs debugging on CI matrix. Can reopen with better environmental understanding.

Issue #3755, opened 2026-03-09
Release Zarr-Python v3.1.6


maxrjones:
**Release**: 3.1.6
**Scheduled Date**: 2026/03/17

**Priority PRs/issues to complete prior to release**

Add any here!

**Before release**:

- [x] Make sure the release branch (e.g., `3.1.x`) is up to date with any backports.
- [x] Make sure that all pull requests which will be included in the release have been properly documented as changelog files in the [`changes/` directory](https://github.com/zarr-developers/zarr-python/tree/main/changes).
- [x] Run ``towncrier build --version x.y.z`` to create the changelog, and commit the result to the release branch.
- [ ] Check [SPEC 0](https://scientific-python.org/specs/spec-0000/#support-window) to see if the minimum supported version of Python or NumPy needs bumping.
- [x] Check to ensure that:
  - [ ] Deprecated workarounds/codes/tests are removed. Run `grep "# TODO" **/*.py` to find all potential TODOs.
  - [x] All tests pass in the ["Tests" workflow](https://github.com/zarr-developers/zarr-python/actions/workflows/test.yml).
  - [x] All tests pass in the ["GPU Tests" workflow](https://github.com/zarr-developers/zarr-python/actions/workflows/gpu_test.yml).
  - [x] All tests pass in the ["Hypothesis" workflow](https://github.com/zarr-developers/zarr-python/actions/workflows/hypothesis.yaml).
  - [x] Check that downstream libraries work well (maintainers can make executive decisions about whether all checks are required for this release).
    - [x] numcodecs
    - [x] Xarray (@jhamman @dcherian @TomNicholas)
        - Zarr's upstream compatibility is tested via the [Upstream Dev CI worklow](https://github.com/pydata/xarray/actions/workflows/upstream-dev-ci.yaml).
        - Click on the most recent workflow and check that the `upstream-dev` job has run and passed. `upstream-dev` is not run on all all workflow runs.
        - Check that the expected version of Zarr-Python was tested using the `Version Info` step of the `upstream-dev` job.
        - If testing on a branch other than `main` is needed, open a PR modifying https://github.com/pydata/xarray/blob/90ee30943aedba66a37856b2332a41264e288c20/ci/install-upstream-wheels.sh#L56 and add the `run-upstream` label.
    - [ ] Titiler.Xarray (@maxrjones)
[...]

PR #3821, opened 2026-03-23
add release trigger that takes a tag as input

d-v-b:
This change to `release.yml` should allow us to use `workflow-dispatch`, i.e., manual triggering,
to invoke the release workflow, with an explicit tag as input. This supports publishing a tagged
release outside of the commit that creates that tag. This is useful for situations where we issued a
release, but the github actions workflow was broken in some way that prevented us from uploading the
release to pypi.

We can also try pushing to test pypi via the same `workflow-dispatch` mechanism.

### DIFF
--- a/.github/workflows/releases.yml
+++ b/.github/workflows/releases.yml
@@ -6,6 +6,16 @@ on:
   pull_request:
     branches: [main]
   workflow_dispatch:
+    inputs:
+      tag:
+        description: 'Git tag to build and publish (e.g. v3.1.6)'
+        required: true
+        type: string
+      test_pypi:
+        description: 'Publish to Test PyPI instead of PyPI'
+        required: false
+        type: boolean
+        default: true
 
 permissions:
   contents: read
@@ -25,6 +35,7 @@ jobs:
     steps:
       - uses: actions/checkout@v6
         with:
+          ref: ${{ inputs.tag || github.ref }}
           submodules: true
           fetch-depth: 0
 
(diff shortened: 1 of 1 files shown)


.github/workflows/releases.yml
```diff
@@ -80,3 +93,10 @@ jobs:
           subject-path: dist/*
       - name: Publish package to PyPI
         uses: pypa/gh-action-pypi-publish@v1.13.0
+        if: ${{ !(github.event_name == 'workflow_dispatch' && inputs.test_pypi) }}
+
+      - name: Publish package to Test PyPI
+        uses: pypa/gh-action-pypi-publish@v1.13.0
+        if: ${{ github.event_name == 'workflow_dispatch' && inputs.test_pypi }}
+        with:
+          repository-url: https://test.pypi.org/legacy/
```
maxrjones, line 102: I don't think this will work without a token added or OIDC setup. Auth for test PyPI is distinct from PyPI

maxrjones:
I think this is a clever solution to https://github.com/zarr-developers/zarr-python/issues/3755#issuecomment-4094570320, but would rather not have this generally available. For example, I wouldn't want anyone to be able to easily re-publish the prior releases that we've yanked from PyPI.

Would you consider just adding an option to deploy 3.1.6 via a workflow dispatch, which we will remove after?

The test PyPI additions are helpful.

d-v-b:
> I think this is a clever solution to [#3755 (comment)](https://github.com/zarr-developers/zarr-python/issues/3755#issuecomment-4094570320), but would rather not have this generally available. For example, I wouldn't want anyone to be able to easily re-publish the prior releases that we've yanked from PyPI.
> 
> Would you consider just adding an option to deploy 3.1.6 via a workflow dispatch, which we will remove after?
> 
> The test PyPI additions are helpful.

I assumed you can't unyank a release, but either way I'm totally fine reverting these changes after we use them to publish 3.1.6. That's the only real reason for doing this PR, once it has fulfilled its purpose the changeset here can be reverted.

PR #3826, opened 2026-03-24
feat: subchunk write order

ilan-gold:
In order to encourage ecosystem compatibility + reserve runtime setting strings/enums (see https://github.com/zarrs/zarrs-python/pull/160), subchunk write order is expanded from `morton` to include `lexicographic`, `colexicographic`, and `unordered` (which is randomized).

TODO:


src/zarr/codecs/sharding.py
```diff
@@ -78,10 +79,27 @@ class ShardingCodecIndexLocation(Enum):
     end = "end"
 
 
+class SubchunkWriteOrder(Enum):
```
d-v-b: advantage of an enum over `Literal["morton", "unordered", "lexicographic", "colexicographic"]`?

ilan-gold: Just copied what was done for `ShardingCodecIndexLocation`!

d-v-b: I'm not a huge fan of enums in python (including `ShardingCodecIndexingLocation`), so unless you object I think it would be better to use a simple `Literal` + a final tuple of strings, like:

```
SubchunkWriteOrder = Literal["morton", "unordered", "lexicographic", "colexicographic"]
SUBCHUNK_WRITE_ORDER: Final[tuple[str, str, str, str]] = ("morton", "unordered", "lexicographic", "colexicographic")
```

ilan-gold: Done (hopefully)!

src/zarr/codecs/sharding.py
```diff
@@ -78,6 +79,15 @@ class ShardingCodecIndexLocation(Enum):
     end = "end"
 
 
+SubchunkWriteOrder = Literal["morton", "unordered", "lexicographic", "colexicographic"]
+SUBCHUNK_WRITE_ORDER: Final[tuple[str, str, str, str]] = (
```
ilan-gold, line 85: maybe `get_args` here as well, like in the test?

d-v-b: I don't think the result of `get_args` is typed! There might be cleaner annotation for the `SUBCHUNK_WRITE_ORDER` constant, but defining them right next to each other basically means we don't need any functions to ensure consistency.

src/zarr/codecs/sharding.py
```diff
@@ -523,6 +540,31 @@ async def _decode_partial_single(
...
+        match self.subchunk_write_order:
+            case "morton":
+                subchunk_iter = morton_order_iter(chunks_per_shard)
+            case "lexicographic":
+                subchunk_iter = np.ndindex(chunks_per_shard)
+            case "colexicographic":
+                subchunk_iter = (c[::-1] for c in np.ndindex(chunks_per_shard[::-1]))
+            case "unordered":
+                subchunk_list = list(np.ndindex(chunks_per_shard))
+                random.shuffle(subchunk_list)
```
d-v-b: do we want a seed here, e.g. to get deterministic output?

docs/user-guide/performance.md
```diff
@@ -113,6 +113,13 @@ bytes within chunks of an array may improve the compression ratio, depending on
 the structure of the data, the compression algorithm used, and which compression
 filters (e.g., byte-shuffle) have been applied.
 
+### Subchunk memory layout
+
+The order of chunks **within each shard** can be changed via the `subchunk_write_order` parameter of the `ShardingCodec` using the `SubchunkWriteOrder` enum or a corresponding string.
```
d-v-b:
```suggestion
The order of chunks **within each shard** can be changed via the `subchunk_write_order` parameter of the `ShardingCodec`. That parameter is a string which must be one of `["morton", "lexicographic", "colexicographic", "unordered"]`.
```

src/zarr/codecs/sharding.py
ilan-gold: That's a good question.  To me, there are/were three options:

1. as-is `random.shuffle`
2. Use `numpy`'s recommended [`shuffle`](https://numpy.org/devdocs/reference/random/generated/numpy.random.Generator.shuffle.html#numpy.random.Generator.shuffle) via `np.random.default_rng().shuffle` (no seed, always fresh entropy)
3. Accept a `rng: numpy.random.Generator | None` argument to the `ShardingCodec` (and a corresponding attribute)

I'm ambivalent between the above 3 options. If we want "a seed", I think 3. makes the most sense.  It also opens up the possibility of reusing the rng elsewhere.

I think that allowing `seed: AllOrSomeOfTheOptionsFor_default_rng` is a bit silly because we're just going to call `np.random.default_rng` anyway (which the user may or may not have wanted i.e., for some compat reasons they may construct their own generator that has its own properties).  `seed: AllOrSomeOfTheOptionsFor_default_rng | Generator` is ok but complicated since then we have to handle both cases.

d-v-b: is there something local to this function call we could use as a seed, to make things deterministic? I'm just thinking that testing is a bit easier with determinism. But I also don't know what kind of randomness people will expect -- will users expect multiple writes of the exact same chunk to produce different subchunk orders?

ilan-gold: > will users expect multiple writes of the exact same chunk to produce different subchunk orders?

I would think that's expected or acceptable, yes, given the name `unordered`.  I actually accidentally made the mistake of going with `NoGuarantee` in the `zarrs` enum for this reason but then went back to `Unordered`.  Now that might be looking better.

But both names also don't vibe with the idea of you passing in a RNG - that would be `random`.

> is there something local to this function call we could use as a seed, to make things deterministic? 

We could always make a hash of some sort like subchunk shape + dtype or something. Hashing may have performance issues though...product of the shape + `itemsize`?

d-v-b: for successful shard writes, we can reconstruct the order from the index. For failed writes, however, we wouldn't be able to reconstruct the intended order from the input without determinism. This doesn't matter at all if true randomness (i.e., non-determinism) is the point. But I don't actually know what the point of this write order mode is. Who needs it, and do they care about true randomness, or would a deterministic, jumbled order be OK?>

ilan-gold: > Who needs it, and do they care about true randomness, or would a deterministic, jumbled order be OK?

I think the later is fine.  There are two cases I have in mind:

1. `zarrs` implements this "unorderedness" dues its threading and there may be performance gains to be had here by making no guarantee on write order and doing the same. In `zarrs`, you can do `encode -> write_to_buffer` instead of `encode -> write_to_intermediate -> enforce_ordering -> write_to_buffer`.  The middle step of `enforce_ordering` means you have to leave the parallelization while in the first case, you don't and can do both steps in one go without a middle synchronization step.
2. If someone was ever whacky enough, being able to check the write order ahead of time could affect data access patterns (or simply knowing the write order because you "own" the data).  If you knew your data were "random", you could do sequential i/o on the shard (i.e., go over the shard in the `index` order which might be `(1, 1), (0, 1), (0, 0), (1, 0)` and not the chunk order like `(0, 0), (0, 1), (1, 0), (1, 1)`) and actually get random access without paying the penalty on local filesystems.  This is nice for ML systems (certainly very intriguing for me) or subsampling.

Both of these seem compatible with "deterministic, jumbled."

What do you have in mind for seeding? `prod(shard_shape) + dtype.itemsize`? Or the same seed for all shards?

ilan-gold: > For failed writes, however, we wouldn't be able to reconstruct the intended order from the input without determinism.

Why is this important?

d-v-b: it's a very vague, generic concern about random routines in code -- if a shard write failed, we might want to know what the shard writer was trying to do. the intended subchunk write order is part of that picture. If that order is deterministic, we can create the exact same state of the failed write in a test. If the order is not deterministic, we can't reproduce exactly what the shard writer was trying to do, and so we would need to debug with a sampling approach.

ilan-gold: In this case, I would think a way of specifying an `rng` is the way to go. Otherwise, I am not sure how we don't end up writing all shards in the same order, which isn't very `unordered` (since the seed-per-shard, at least local to this function, would be based on `chunks_per_shard`).

ilan-gold: So `ShardingCodec` needs to be picklable and so RNGs won't match by `==` after pickling it seems.  On top of that, there is also the fact that shuffling occurs more than once for writing i.e., there are two creations of the iterator which means two calls to `shuffle`, which I don't find very intuitive (also not clear how necessary that is TBH).

I am going to punt on it and remove it.  I think it only makes sense anyway with the performance improvement that comes from "direct" writing.

ilan-gold: Just to be clear, they will have the same "state" AFAIK, it's just that we have a test that does `assert pickle.loads(pickle.dumps(codec)) == codec`

ilan-gold: Ah, will jsut use `rng: None | np.random.Generator`, duh, seems better for completeness (i.e., matching `zarrs`)

src/zarr/codecs/sharding.py
```diff
@@ -582,7 +616,7 @@ async def _encode_partial_single(
         )
 
         if self._is_complete_shard_write(indexer, chunks_per_shard):
-            shard_dict = dict.fromkeys(morton_order_iter(chunks_per_shard))
+            shard_dict = dict.fromkeys(np.ndindex(chunks_per_shard))
```
ilan-gold: cc @mkitti 

Here and below, I don't think there is any need to construct the `dict` in morton order, right? There should be no correctness or performance hit here?

@d-v-b This now ensures we only shuffle in the `unordered` case once so the test is nice and clean - write once + get order, create a new codec with the same seed + create the iterator from that codec, match orders

mkitti: In Python, dicts are ordered and I think the optimal iteration order may need to be encoded in the dict the last time I examined the situation. I was just trying to preserve the situation before my edits.

src/zarr/codecs/sharding.py
d-v-b: What does "unordered" guarantee? is it actually important that the order be random, or just the lack of any guaranteed order? Because IMO adding a rng to the sharding codec for no performance benefit is not looking like a good deal to me. What if we just use an internally-defined order when "unordered" is picked, with no promise that this order is stable?

src/zarr/codecs/sharding.py
ilan-gold: https://github.com/zarr-developers/zarr-python/pull/3826/commits/027c4699bb543a7c08d0499ed3981691cea27114

So this wasn't about dictionary order, but instead in the vectorized case, the order to `ShardReader.to_dict_vectorized` had to match that of what `ShardReader` was internally generating, as it turned out morton order.  So I'm glad I caught this because I think it means the data was being corrupted for the other orders (which weren't getting hypothesis-tested).

So I'm going to add something to the hyptothesis tests for this.

I had the same feeling initially that the dictionary order mattered, but it turns out the final call to `_encode_shard_dict` [actually handles the ordering for us to the output buffer](https://github.com/zarr-developers/zarr-python/blob/ea13af5a3f7fc489e37a62f043c4b45a9ae55027/src/zarr/codecs/sharding.py#L634) while writing to the intermediate `shard_dict` can be done in any order, as long as the final buffer is done in the correct order

d-v-b (approved):
this looks good, thanks! 

we have some interesting follow-up work:

- finding a nice way to expose the sharding codec settings in the signature of `create_array`
- writing a chunk writing routine that can incrementally stream out a shard
- formalizing the distinction between codec attributes that get serialized to JSON and runtime-only attributes

PR #3857, opened 2026-03-31
fix: convert zarr Array to numpy before __setitem__ async dispatch

### LINKED ISSUE #3611: Error trying to assign slice of a Zarr array to another Zarr array
### Zarr version

3.1.6.dev5+gee0e69a74

### Numcodecs version

0.6.15

### Python Version

3.12.8

### Operating System

macOS

### Installation

From source

### Description

Trying to assign to a slice of a Zarr array with another Zarr array fails. Assigning a `float` or a NumPy array to the same slice works fine.

### Steps to reproduce

```python
# /// script
# requires-python = ">=3.11"
# dependencies = [
# ]
# ///
#

import zarr
arr1 = zarr.zeros((5, 5))
arr2 = zarr.ones((1, 1))

slc = (slice(1, 2), slice(2, 3))
arr1[slc] = arr2
```

### Additional output

```python
  File "/Users/dstansby/software/zarr/zarr-python/test.py", line 7, in <module>
    arr1[slc] = arr2
    ~~~~^^^^^
  File "/Users/dstansby/software/zarr/zarr-python/src/zarr/core/array.py", line 2975, in __setitem__
    self.set_orthogonal_selection(pure_selection, value, fields=fields)
  File "/Users/dstansby/software/zarr/zarr-python/src/zarr/core/array.py", line 3458, in set_orthogonal_selection
    return sync(
           ^^^^^
  File "/Users/dstansby/software/zarr/zarr-python/src/zarr/core/sync.py", line 159, in sync
    raise return_result
  File "/Users/dstansby/software/zarr/zarr-python/src/zarr/core/sync.py", line 119, in _runner
    return await coro
           ^^^^^^^^^^
  File "/Users/dstansby/software/zarr/zarr-python/src/zarr/core/array.py", line 1736, in _set_selection
    await self.codec_pipeline.write(
  File "/Users/dstansby/software/zarr/zarr-python/src/zarr/core/codec_pipeline.py", line 488, in write
    await concurrent_map(
  File "/Users/dstansby/software/zarr/zarr-python/src/zarr/core/common.py", line 116, in concurrent_map
    return await asyncio.gather(*[asyncio.ensure_future(run(item)) for item in items])
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/dstansby/software/zarr/zarr-python/src/zarr/core/common.py", line 114, in run
    return await func(*item)
           ^^^^^^^^^^^^^^^^^
  File "/Users/dstansby/software/zarr/zarr-python/src/zarr/core/codec_pipeline.py", line 392, in write_batch
    self._merge_chunk_array(
  File "/Users/dstansby/software/zarr/zarr-python/src/zarr/core/codec_pipeline.py", line 325, in _merge_chunk_array
    chunk_value = value[out_selection]
                  ~~~~~^^^^^^^^^^^^^^^
  File "/Users/dstansby/software/zarr/zarr-python/src/zarr/core/buffer/cpu.py", line 186, in __getitem__
    return self.__class__(np.asanyarray(self._data.__getitem__(key)))
                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/dstansby/software/zarr/zarr-python/src/zarr/core/array.py", line 2868, in __getitem__
    return self.get_orthogonal_selection(pure_selection, fields=fields)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/dstansby/software/zarr/zarr-python/src/zarr/core/array.py", line 3339, in get_orthogonal_selection
    return sync(
           ^^^^^
  File "/Users/dstansby/software/zarr/zarr-python/src/zarr/core/sync.py", line 146, in sync
    raise SyncError("Calling sync() from within a running loop")
zarr.core.sync.SyncError: Calling sync() from within a running loop
sys:1: RuntimeWarning: coroutine 'AsyncArray._get_selection' was never awaited
```

nuglifeleoji:
Assigning a `zarr.Array` as the value in a slice assignment fails with:

```
zarr.core.sync.SyncError: Calling sync() from within a running loop
```

The root cause is that the codec pipeline calls `value[out_selection]` from inside an already-running async event loop. When `value` is a zarr `Array`, that indexing triggers another `sync()` call, which raises the error.

The fix is to convert the value to a NumPy array in `Array.__setitem__` before dispatching to the async codec pipeline. `np.asarray(value)` calls `Array.__array__`, which reads data synchronously and safely before we enter the async context.

```python
src = zarr.array(np.arange(10))
dst = zarr.zeros(10, dtype=src.dtype)
dst[:] = src  # previously raised SyncError, now works
```

Closes #3611

Made with [Cursor](https://cursor.com)

### DIFF
--- a/changes/3611.bugfix.md
+++ b/changes/3611.bugfix.md
@@ -1,0 +1,1 @@
+Fix `SyncError` raised when assigning a `zarr.Array` as the value in a `__setitem__` call (e.g. `dst[:] = src` where `src` is a zarr array). The source array is now converted to a NumPy array before entering the async codec pipeline.
--- a/src/zarr/core/array.py
+++ b/src/zarr/core/array.py
@@ -2929,6 +2929,12 @@ class Array[T_ArrayMetadata: (ArrayV2Metadata, ArrayV3Metadata)]:
         [blocks][zarr.Array.blocks], [__getitem__][zarr.Array.__getitem__]
 
         """
+        # Converting a zarr Array to numpy here avoids a SyncError that occurs when
+        # value.__getitem__ is called inside the async codec pipeline (which already
+        # runs within a running event loop).  np.asarray triggers Array.__array__,
+        # which reads the data synchronously before we enter the async context.
+        if isinstance(value, Array):
+            value = np.asarray(value)
         fields, pure_selection = pop_fields(selection)
         if is_pure_fancy_indexing(pure_selection, self.ndim):
             self.vindex[cast("CoordinateSelection | MaskSelection", selection)] = value
--- a/tests/test_indexing.py
+++ b/tests/test_indexing.py
@@ -443,7 +443,6 @@ def test_orthogonal_indexing_fallback_on_getitem_2d(
     np.testing.assert_array_equal(z[index], expected_result)
 
 
-@pytest.mark.skip(reason="fails on ubuntu, windows; numpy=2.2; in CI")
 def test_setitem_zarr_array_as_value() -> None:
     # Regression test for https://github.com/zarr-developers/zarr-python/issues/3611
     # Assigning a zarr Array as the value used to raise
@@ -462,6 +461,7 @@ def test_setitem_zarr_array_as_value() -> None:
     assert_array_equal(dst2[2:7], np.arange(2, 7))
 
 
+@pytest.mark.skip(reason="fails on ubuntu, windows; numpy=2.2; in CI")
 def test_setitem_repeated_index():
     array = zarr.array(data=np.zeros((4,)), chunks=(1,))
     indexer = np.array([-1, -1, 0, 0])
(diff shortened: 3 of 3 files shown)


d-v-b:
this fix will resolve the exception but is extremely sub-optimal from a performance POV. I think we can accept this fix but we need to be aware that collecting the entire array in memory, then writing it out again, is a waste of compute, IO bandwidth, and memory.

nuglifeleoji:
Totally agree — this is a short-term fix. The proper solution would be to make the codec pipeline's chunk-iteration async-aware so a zarr Array source can be read chunk-by-chunk without triggering a nested sync(). Happy to follow up with that once this lands, or we could close this in favour of a more complete async-pipeline approach if that's preferred.

d-v-b:
this is good for now, we will just need an issue to track the opportunity for performance improvements on this path.

nuglifeleoji:
will open a tracking issue for that

d-v-b (approved):

PR #3863, opened 2026-04-01
fix: numpy default NaT handling

ilan-gold:
Upstream CI started failing because of `numpy` `NaT` requiring a unit, not sure if this is intended on their side since `NaT` would be the one time I might not expect a unit..  ~~I am going to open an issue~~ Judging by https://github.com/numpy/numpy/pull/29619#issuecomment-3587365654, this is the intention.  

I don't really get this as a newbie to datetimes since:

```python
import numpy as np
np.testing.assert_equal(np.timedelta64("NaT", "m"), np.timedelta64("NaT", "ns"))
```
works.

See: https://github.com/zarr-developers/zarr-python/actions/runs/23841178099/job/69496895274?pr=3826 for the failures

TODO:

### DIFF
--- a/src/zarr/core/dtype/npy/time.py
+++ b/src/zarr/core/dtype/npy/time.py
@@ -545,7 +545,7 @@ class TimeDelta64(TimeDTypeBase[np.dtypes.TimeDelta64DType, np.timedelta64], Has
         raise a TypeError.
         """
         if self._check_scalar(data):
-            if isinstance(data, np.timedelta64) and np.isnan(data):
+            if isinstance(data, np.timedelta64) and np.isnat(data):
                 return np.timedelta64("NaT", self.unit)
             return self._cast_scalar_unchecked(data)
         msg = (
--- a/src/zarr/core/dtype/npy/time.py
+++ b/src/zarr/core/dtype/npy/time.py
@@ -559,7 +559,7 @@ class TimeDelta64(TimeDTypeBase[np.dtypes.TimeDelta64DType, np.timedelta64], Has
         This method provides a default value for the timedelta64 scalar, which is
         a 'Not-a-Time' (NaT) value.
         """
-        return np.timedelta64("NaT")
+        return np.timedelta64("NaT", self.unit)
 
     def from_json_scalar(self, data: JSON, *, zarr_format: ZarrFormat) -> np.timedelta64:
         """
--- a/src/zarr/core/dtype/npy/time.py
+++ b/src/zarr/core/dtype/npy/time.py
@@ -545,6 +545,8 @@ class TimeDelta64(TimeDTypeBase[np.dtypes.TimeDelta64DType, np.timedelta64], Has
         raise a TypeError.
         """
         if self._check_scalar(data):
+            if isinstance(data, np.timedelta64) and np.isnan(data):
+                return np.timedelta64("NaT", self.unit)
             return self._cast_scalar_unchecked(data)
         msg = (
             f"Cannot convert object {data!r} with type {type(data)} to a scalar compatible with the "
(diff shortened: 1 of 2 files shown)


d-v-b (approved):

d-v-b:
I'm not actually sure what the endgame is here. It started with this issue: https://github.com/numpy/numpy/issues/28287, now it looks like the direction is requiring that `NaT` values be associated with units, which seems conceptually a bit odd to me. We should probably keep an eye on this and see how it shakes out.

src/zarr/core/dtype/npy/time.py
```diff
@@ -545,6 +545,8 @@ def cast_scalar(self, data: object) -> np.timedelta64:
         raise a TypeError.
         """
         if self._check_scalar(data):
+            if isinstance(data, np.timedelta64) and np.isnan(data):
+                return np.timedelta64("NaT", self.unit)
```
ilan-gold: Not really sure what to do here since the following strikes me as a bug that won't be fixed (i.e., the below is the codepath of the next line `return self._cast_scalar_unchecked(data)`):

```python
# /// script
# requires-python = "==3.12"
# dependencies = [
#   "numpy==2.0",
# ]
# ///
import numpy as np

# AssertionError
assert np.isnan(
    np.dtype("timedelta64[10s]").type(
        np.timedelta64("NaT", "s"), "10s"
    )
)
```

ilan-gold:
> which seems conceptually a bit odd to me

Agreed

src/zarr/core/dtype/npy/time.py
d-v-b: either this is a bug in numpy that we should upstream, or there's some equivalent `isNaT` routine? Alternatively, we check the int64 value of the timedelta, i think all the NaT values land on the same value.

ilan-gold: > i think all the NaT values land on the same value

I observed this as well

> we should upstream

My "bug that won't be fixed" point was that this appears to be `numpy<2.4` specific (i.e., the above script fails on all numpy versions I tried less than 2.4)

ilan-gold:
```python
# /// script
# requires-python = "==3.12"
# dependencies = [
#   "numpy==2.0", # or 2.3
# ]
# ///
import numpy as np

# AssertionError
assert np.isnat(
    np.dtype("timedelta64[10s]").type(
        np.timedelta64("NaT", "s"), "10s"
    )
)
```
also fails. 

The. output of 

```python
np.dtype("timedelta64[10s]").type(
    np.timedelta64("NaT", "s"), "10s"
)
```
is this magic integer you mentioned

d-v-b: gotcha, I didn't see the numpy version pin. looks like numpy does have an `isnat` function so we should probably just use that instead

d-v-b: wait what do we expect here? this isn't actually a NaT value, it's just a huge timedelta:

```python
>>> np.dtype("timedelta64[10s]").type(np.timedelta64("NaT", "s"), "10s")
np.timedelta64(922337203685477579,'10s')
```

ilan-gold: Before this change, the call would have been unit-less:

```python
assert np.isnat(np.dtype("timedelta64[10s]").type(np.timedelta64("NaT"), "10s"))
```
which works.  Bringing in the units is what is causing the non-NaT value in pre-numpy 2.4 versions but now:

```python
# /// script
# requires-python = "==3.12"
# dependencies = [
#   "numpy>=2.4",
# ]
# ///
import numpy as np

assert np.isnat(np.dtype("timedelta64[10s]").type(np.timedelta64("NaT", "ms"), "10s"))
assert np.isnat(np.dtype("timedelta64[10s]").type(np.timedelta64("NaT", "ns"), "10s"))
```  
works so our old code path works fine although now that I'm writing this, is no longer hit when `NaT`.

d-v-b:
@ilan-gold is this ready?

ilan-gold:
Yea, I agree we should keep an eye on this.

PR #3874, opened 2026-04-06
feat: add cast_value and scale_offset codecs

d-v-b:
Defines two new codecs that together provide a v3-native replacement for the existing `numcodecs.fixedscaleoffset` codec.

The `cast_value` codec requires an optional dependency on the [`cast-value-rs` package](https://github.com/zarr-developers/cast-value.rs/). The reason for bringing in rust here is because it gives us better average speed and _much_ better memory usage than a numpy implementation. Happy to show this with benchmarks if needed.

fwiw, I would rather not define these codecs in `zarr-python`. In fact I already defined these codec classes in [another package](https://github.com/zarr-developers/cast-value.py). But it's not possible to make zarr-python depend on an externally-defined codec without creating a circular dependency, so here we are. See discussion in the related issue.

related, and possibly closes https://github.com/zarr-developers/zarr-python/issues/3867

edit: if depending on a new optional package for `cast_value` is too much, then we should revisit #3774. The conclusion there was that defining the `cast_value` codec in zarr-python itself was too much. And if the conclusion is that both defining the codec externally is too much, and defining it internally is too much, then we should do something like deprecate `numcodecs.fixedscaleoffset` and  direct people to the python packages that implement its replacement.


d-v-b:
cc @emmanuelmathot

d-v-b:
the coverage gaps are spurious because some of these routines only run when the optional `cast-value-rs` package is available. this is ready for review.

vincentsarago:
👋 This is great.

With this codec we could reduce the size of our stored dataset 🙏

src/zarr/codecs/scale_offset.py
```diff
@@ -0,0 +1,151 @@
...
+    def _decode_sync(
+        self,
+        chunk_array: NDBuffer,
+        chunk_spec: ArraySpec,
+    ) -> NDBuffer:
+        arr = chunk_array.as_ndarray_like()
+        offset = self._to_scalar(self.offset, chunk_spec.dtype)
+        scale = self._to_scalar(self.scale, chunk_spec.dtype)
+        if np.issubdtype(arr.dtype, np.integer):
+            result = (arr // scale) + offset  # type: ignore[operator]
```
maxrjones, line 110: per the spec, should this use `/` and error if the result isn't representable in the data type?

pyproject.toml
```diff
@@ -191,6 +192,16 @@ run-benchmark = "pytest --benchmark-enable tests/benchmarks"
...
 
+[tool.hatch.envs.cast-value]
+template = "test"
+features = ["cast-value-rs"]
+
+[[tool.hatch.envs.cast-value.matrix]]
+python = ["3.12"]
+
+[tool.hatch.envs.cast-value.scripts]
+run = "pytest tests/test_codecs/test_cast_value.py {args:}"
```
maxrjones, line 203: these should be tested in the CI

maxrjones (changes requested):
Nice, thanks @d-v-b! A few comments:

## protect against data corruption

I think you need to protect against integer overflow/underflow. Here's a couple tests that demonstrate corrupted data:

```
def test_encode_uint8_underflow_should_error():
    """uint8(0) - offset(1) wraps to 255, then * 2 = 254. Decode gives 128, not 0."""
    arr = zarr.create_array(
        store={},
        shape=(3,),
        dtype="uint8",
        chunks=(3,),
        filters=[ScaleOffset(offset=1, scale=2)],
        compressors=None,
        fill_value=1,
    )
    data = np.array([0, 1, 2], dtype="uint8")
    with pytest.raises((ValueError, OverflowError)):
        arr[:] = data


def test_encode_int8_overflow_should_error():
    """int8(50) * scale(3) = 150 wraps to -106. Decode gives -36, not 50."""
    arr = zarr.create_array(
        store={},
        shape=(1,),
        dtype="int8",
        chunks=(1,),
        filters=[ScaleOffset(offset=0, scale=3)],
        compressors=None,
        fill_value=0,
    )
    data = np.array([50], dtype="int8")
    with pytest.raises((ValueError, OverflowError)):
        arr[:] = data
```

## user-guide documentation

I think it'd be worth having some documentation page for users coming from other data formats, since they could easily mix up the scale offset parameters. Here's a summary:

The main ones across scientific/geospatial data:

**CF conventions** (NetCDF): `unpacked = packed * scale_factor + add_offset`
- Scale means "multiply when unpacking"

**GDAL/GeoTIFF**: `value = raw * scale + offset`
- Same formula as CF. GDAL uses this for raster band transformations.

**HDF-EOS5**: `value = scale_factor * (stored - offset)`
- Subtract offset *before* scaling — different from CF which adds offset *after* scaling. This is a common source of confusion between HDF-EOS and CF.

**GRIB (WMO)**: `value = reference_value + (raw * 2^binary_scale_factor) / 10^decimal_scale_factor`
- Two scale factors (binary and decimal) plus a reference value. Much more complex.

**numcodecs FixedScaleOffset** (zarr v2): `encode: round((x - offset) * scale)`, `decode: x / scale + offset`
- Same formula as the new zarr spec, plus explicit rounding on encode.

**Zarr v3 scale_offset**: `encode: (x - offset) * scale`, `decode: x / scale + offset`
- Same as numcodecs but without built-in rounding (that's now delegated to `cast_value`).

**ML quantization** (e.g., TensorFlow Lite): `real = (quantized - zero_point) * scale`
- Scale means "multiply when unpacking" (like CF). `zero_point` is the integer that maps to real zero.

The key axes of variation:

| | Scale means | Order | Rounding |
|---|---|---|---|
| CF / GDAL / ML quant | multiply to unpack | scale then offset | implicit |
| HDF-EOS5 | multiply to unpack | offset then scale | implicit |
| Zarr v3 / numcodecs | multiply to pack | offset then scale | explicit (cast_value) |
| GRIB | two scale factors | its own thing | N/A |

The biggest gotcha is that CF's `scale_factor=0.1` means the same thing as zarr's `scale=10`. Users migrating CF data need to invert the scale value.

## sustainability

Will you add me or the zarr team as a maintainer on PyPI for cast-value-rs if this is merged?

pyproject.toml
d-v-b: good catch, I added `cast-value-rs` to the optional branch of our test matrix, which should cover this. And I removed the cast-value-rs-specific hatch env.

src/zarr/codecs/scale_offset.py
d-v-b: yes, I fixed this is recent commits. the checks are much more stringent, and the code is quite a bit slower. We are going against the grain of numpy here, as numpy defaults to mathematically sensible operations at the expense of silent casting. If this is a performance problem worth fixing, we might be looking at a `scale-offset-rs` situation :/

d-v-b:
> Will you add me or the zarr team as a maintainer on PyPI for cast-value-rs if this is merged?

absolutely! I made you (via the zarr-python core devs team) admins of `cast-value-py` and `cast-value-rs`.

d-v-b:
> I think it'd be worth having some documentation page for users coming from other data formats, since they could easily mix up the scale offset parameters. Here's a summary:

I totally agree, but I wonder if this should be a separate PR? I don't know those conventions well at all, and I worry I would mess something up.

d-v-b:
@maxrjones have a look at the latest changes, I think I addressed everything but lmk if anything else needs to be done

maxrjones (approved):

TomNicholas:
Guys I tried using this in VirtualiZarr (for https://github.com/zarr-developers/VirtualiZarr/issues/1004) and hit a possible problem when writing roundtrip tests. Claude:

> `scale_offset` can't bit-match CF's decode for non-power-of-two factors.
>
> CF decodes by multiplying: unpacked = packed * scale_factor + add_offset. scale_offset decodes by dividing: x / scale + offset with scale = 1/scale_factor. Division and multiplication round differently, and 1/scale_factor usually isn't exactly representable, so the two disagree at the ~1 ULP level:
> 
>  3 * 0.1   # 0.30000000000000004   (CF: packed * scale_factor)
>   3 / 10.0  # 0.3                    (scale_offset: packed / scale)
>  
> So migrating an existing CF-packed dataset to scale_offset and reading it back doesn't reproduce xarray.decode_cf bit-for-bit — decoded values shift by < 1 ULP. This is well below the quantization floor that packing already imposes (at scale_factor=0.01 the stored int16 has already discarded everything finer than 0.01), so it's not a correctness problem in any practical sense. But it does mean "round-trips CF exactly" is not a property this codec can offer, which is worth being explicit about for anyone migrating CF data or comparing against xarray's `decode_cf`.

Why was scale inverted here compared to `scale_factor`?

d-v-b:
> Guys I tried using this in VirtualiZarr (for [zarr-developers/VirtualiZarr#1004](https://github.com/zarr-developers/VirtualiZarr/issues/1004)) and hit a possible problem when writing roundtrip tests. Claude:
> 
> > `scale_offset` can't bit-match CF's decode for non-power-of-two factors.
> > CF decodes by multiplying: unpacked = packed * scale_factor + add_offset. scale_offset decodes by dividing: x / scale + offset with scale = 1/scale_factor. Division and multiplication round differently, and 1/scale_factor usually isn't exactly representable, so the two disagree at the ~1 ULP level:
> > 3 * 0.1   # 0.30000000000000004   (CF: packed * scale_factor)
> > 3 / 10.0  # 0.3                    (scale_offset: packed / scale)
> > So migrating an existing CF-packed dataset to scale_offset and reading it back doesn't reproduce xarray.decode_cf bit-for-bit — decoded values shift by < 1 ULP. This is well below the quantization floor that packing already imposes (at scale_factor=0.01 the stored int16 has already discarded everything finer than 0.01), so it's not a correctness problem in any practical sense. But it does mean "round-trips CF exactly" is not a property this codec can offer, which is worth being explicit about for anyone migrating CF data or comparing against xarray's `decode_cf`.
> 
> Why was scale inverted here compared to `scale_factor`?

Because nobody involved in developing the codec was trying to exactly replicate this aspect of CF. See the conversation in the PR that added this codec: https://github.com/zarr-developers/zarr-extensions/pull/43

d-v-b:
with respect to the source data, the scale-offset transformation is lossy in nearly every application. Is it important to exactly replicate that lossiness?

TomNicholas:
Okay so that wasn't a design goal, I see. 

But given my use case in VirtualiZarr (i.e. replacing the existing xarray decoding mechanism with a zarr-native one), is this a problem?

> with respect to the source data, the scale-offset transformation is lossy in nearly every application. Is it important to exactly replicate that lossiness?

I don't really know, that's partly why I'm asking. Is this something I should care about? 

At the very least we would have to document that `xr.assert_identical` cannot be expected to work on the result @maxrjones.

d-v-b:
If exactly replicating the old behavior is desirable and there's no configuration of `scale_offset` that works, then you could define a new codec `inverse_scale_offset` or something, that flips division for multiplication. Implementations could re-use a lot of the existing code.

PR #3877, opened 2026-04-06
Revert "chore: update pre-commit hooks"

maxrjones:
Reverts zarr-developers/zarr-python#3876

zarr-developers/zarr-python#3876 caused the CI to fail, since it bumped the linters without applying fixes. I also think we should apply a policy to not bump dependencies that are <1 week old, given the number of recent supply chain attacks. I've opened #3878 to discuss that proposal.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -11,13 +11,13 @@ default_language_version:
 
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.15.9
+    rev: v0.15.4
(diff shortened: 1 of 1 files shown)


d-v-b (approved):

d-v-b:
thanks for fixing this!

maxrjones:
> thanks for fixing this!

no problem. I'm going to add the pre-commit/lint workflows to the required passing checks in the branch protection rules for main, so that auto-merge doesn't ignore it. It'll be easy to change back if you or others disagree.

PR #3883, opened 2026-04-08
chore: no warning if `array.target_shard_size_bytes` is set with `shards="auto"`

ilan-gold:
I started a zulip discussion [here](https://ossci.zulipchat.com/#narrow/channel/423692-Zarr/topic/How.20do.20people.20do.20sharding.20from.20a.20user-perspective.3F/with/583594752) and I feel like the following PR is reasonable.  Using sharding simply to make file systems more performant while viewing the subchunks as the real "units" seems to be a nice use-case and if you opt-in to "auto" with a target size, maybe you shouldn't get a warning?

We could in theory also document the strategy of "pack chunks in equally along all dimensions" but is there really any appetite to change that? That stands in contrast to the current "auto" behavior (without a target size) of 2 chunks per dimension in the shard, which does seems like the sort of thing that could change.

TODO:


ilan-gold:
Closing because we discussed that users can turn this off.

PR #3885, opened 2026-04-08
perf: use sync methods for chunk encoding / decoding

d-v-b:
# perf: synchronous codec pipeline (`FusedCodecPipeline`), opt-in

## Summary

This PR adds `FusedCodecPipeline`, an **opt-in** codec pipeline. It removes the per-chunk **async scheduling overhead** of the default `BatchedCodecPipeline` — roughly one coroutine per chunk, which on real workloads dominates the actual codec work — by running codec compute synchronously and in bulk.

The default pipeline is **unchanged** for standard configurations: existing users see no behavior change unless they opt in. One caveat — sharded arrays whose inner codec chain changes the dtype now evolve that chain correctly (the #2179 fix under _Notable internal changes_); that is a bugfix and it also runs on the default pipeline. Standard inner chains (`[BytesCodec]`, `[BytesCodec, ZstdCodec/BloscCodec]`, transpose + bytes) are byte-identical to before. We want to move slowly here — let early adopters exercise the new pipeline before considering it as a default.

> **Note on the original framing.** This PR started as `PhasedCodecPipeline`, pitched around "separate IO from compute." That pitch did not pan out: indexing / Python-object-creation overhead swamped the benefit. The win that remained — and what this PR now delivers — is simpler: **avoid async for storage backends that don't need it.** See [this comment](https://github.com/zarr-developers/zarr-python/pull/3885#issuecomment-4642056937) for the history.

## What it does

1. **Synchronous codec chain.** When every codec implements `SupportsSyncCodec`, a `ChunkTransform` runs the codec chain synchronously — no event loop, no per-chunk coroutine — optionally across a thread pool for CPU-heavy decode/encode.
2. **Sharded reads use synchronous IO.** Byte-range reads are coalesced via a new `Store.get_ranges_sync`, and dense, fixed-size, uncompressed shards take a vectorized whole-shard bulk decode. Sharded writes go through the codec's synchronous full-shard-rewrite path.
3. **Graceful fallback.** When the store lacks synchronous IO (e.g. `ZipStore`), or a codec isn't sync-capable, the pipeline falls back to the async path — behavior-equivalent to `BatchedCodecPipeline`.

IO ownership is unchanged: the sharding codec still holds the byte getter/setter and reads/writes storage directly (the same model as before; this PR does **not** move toward storage-free codecs).

## Performance

Large speedups for sharded arrays — up to ~24× writes / ~14× reads on many-chunks-per-shard layouts, more with compression — and no regressions on compute-bound workloads.

## Opting in

- The default `codec_pipeline.path` is unchanged (`BatchedCodecPipeline`). Enable the new pipeline with:
  ```python
  zarr.config.set({"codec_pipeline.path": "zarr.core.codec_pipeline.FusedCodecPipeline"})
  ```
- `codec_pipeline.max_workers` defaults to `None`, which means `FusedCodecPipeline` runs **fully threaded** (`os.cpu_count()` workers). It is the only pipeline that reads this setting — `BatchedCodecPipeline` ignores it entirely. Set it to `1` for sequential, single-threaded execution (often faster for memory-backed stores, where store reads are GIL-locked and thread-pool overhead dominates), or to a fixed integer to cap the pool.

These config additions are **non-breaking**: the new pipeline and its threading only take effect once you opt in via `codec_pipeline.path`. The default pipeline never reads `max_workers` and never spins up a thread pool.

## Public API additions

- `SyncByteGetter` / `SyncByteSetter` — runtime-checkable protocols letting custom byte getters/setters opt into the sync fast path for in-memory IO (used by the sharding codec for its inner chunks).
- `Store.get_ranges_sync` — synchronous, coalescing counterpart of `get_ranges`, with the same coalescing policy.

## Correctness

Both pipelines are held to the same behavior, with new equivalence/parity suites asserting that each fast path produces results identical to the general path it bypasses:

- `tests/test_pipeline_parity.py` — Fused vs Batched parity across sharded/unsharded, compressed/uncompressed, read/write/overwrite scenarios.
- `tests/test_codec_pipeline_suite.py` — a shared suite both pipelines must pass (incl. zarr v2 + nested sharding).
- `tests/test_fastpath_equivalence.py` — property tests pinning each fast path (complete-chunk merge view, bulk shard decode, scalar-broadcast write, byte-range coalescing) equal to the general path.
- `tests/test_fused_pipeline.py` — Fused-specific behavior incl. thread-pool and async-fallback paths.

Three fast-path correctness issues were found, reproduced, and fixed with regression tests: bulk shard decode could serve a reordering `vindex`/`oindex` selection in natural order on uncompressed crc-free shards (now gated on `sel_shape`); the `ChunkTransform` spec cache was not torn-read-safe under `max_workers > 1` (now a single atomically-assigned entry); and sharded reads with variable-length / numcodecs inner codecs crashed because `Codec.is_fixed_size` had no default (now defaults to `False`).

## Notable internal changes

- Shared single-chunk helpers in `src/zarr/core/chunk_utils.py` (merge/encode, decode/scatter, empty-chunk normalization, spec evolution) used by both pipelines and the sharding codec, so the sync and async paths can't drift.
- `V2Codec` and the numcodecs wrappers gained `_decode_sync`/`_encode_sync` cores; the async methods are now thin `asyncio.to_thread` wrappers.
- The sharding codec's inner sub-chunk pipeline now follows the configured `codec_pipeline.path` (restoring #2179 behavior) and is evolved/memoized per spec.

## Lineage

Builds on prior work in #3719. Depends on changes from #3907 and #3908. Mostly authored with Claude.


d-v-b:
@TomAugspurger how would this design work with CUDA codecs?

src/zarr/core/codec_pipeline.py
```diff
@@ -679,3 +720,504 @@ def codecs_from_list(
...
+            return transform.decode_chunk(raw, chunk_spec)
+
+        specs = [cs for _, cs, *_ in batch]
+        if n_workers > 0 and len(batch) > 1:
+            with ThreadPoolExecutor(max_workers=n_workers) as pool:
+                decoded_list = list(pool.map(_decode_one, raw_buffers, specs))
+        else:
+            decoded_list = [
+                _decode_one(raw, spec) for raw, spec in zip(raw_buffers, specs, strict=True)
+            ]
```
ilan-gold, line 962: Why isn't this all multi-threaded i.e., the I/O as well?

d-v-b: I should benchmark this, but my expectation was that IO against memory storage and local storage is not compute-limited, and so threads wouldn't remove a real bottleneck. for memory storage i'm sure this is true, not sure about local storage though

src/zarr/core/codec_pipeline.py
```diff
@@ -679,3 +762,506 @@ def codecs_from_list(
...
+            byte_getter, chunk_spec, chunk_selection, out_selection, _ = item
+            raw = byte_getter.get_sync(prototype=chunk_spec.prototype)
+            if raw is None:
+                out[out_selection] = fill
+                return _missing
+            decoded = transform.decode_chunk(raw, chunk_spec)
+            selected = decoded[chunk_selection]
+            if drop_axes:
+                selected = selected.squeeze(axis=drop_axes)
+            out[out_selection] = selected
```
ilan-gold, line 1010: It might be worth experimenting with moving this setting operation `out[out_selection] = selected` outside the threadpool execution since, IIRC, it holds the GIL and is probably non-trivial time-wise.

ilan-gold: The memory usage will probably go up a bit though....

ilan-gold (commented):
The first thing that jumps out at me is the potential for a performance regression because there is no whole-shard special casing in the new fused codec pipeline.  I guess your benchmarks cover that, but they also might not.

As you point out in #3925, that PR will then bring nested concurrency because the pipeline will have "outer" concurrency (that also controls the decompression) while there will be some inner concurrency from coalesced ranges.

I'd like to understand how these two PRs will fit together!

ilan-gold (commented):
Hi @d-v-b maybe in the interest of pruning this a bit, I don't actually see a dependency on `set_range_sync` being used in the pipeline - could it be removed from this PR? 

In fact, if we want to be able to do what `zarrs` does for write performance, we mayb likely need #3826 + special casing for `unordered` first as this would then be mixed with `set_range_sync` to achieve the "first to compress, first to write" paradigm that appears in zarrs for `unordered` subchunk writing:

https://github.com/zarrs/zarrs/blob/82876988f0ef25cb45950320c75f9ef591a4359c/zarrs/src/array/codec/array_to_bytes/sharding/sharding_codec.rs#L822-L871

Otherwise, you need to hold the shard in-memory AFAICT to be able to create the ordering ahead of time (i.e., morton) to write.

d-v-b:
first apologies for the messy state and second yes consider the set_range stuff extra credit. I do want to ensure that we _can_ support range writes, but as long as we are confident that we aren't blocking that path, then it's totally fine to slim this PR down to "whatever it takes to speed up local + memory storage"

d-v-b:
and you should absolutely feel free to push ad libitum to this branch. I'm not actively working on it, and I'm confident that the twin constraints of our test suite and the benchmarks can keep us sane

d-v-b:
(claude wrote this)

## Current state of this PR (reconciled against the original framing)

This branch has evolved substantially. Posting a current-state summary so the description can be updated.

### ⚠️ The original "hard IO/compute boundary" framing no longer describes this PR

This PR was opened as a **phased** pipeline built on separating IO from compute (preparatory IO → pure compute → final IO), with the suggestion that the spec be rewritten around that framing. **That is not what shipped.** The class was renamed `PhasedCodecPipeline` → `FusedCodecPipeline`, and its docstring now explicitly disclaims the original thesis:

> The win is **NOT** "separating IO from compute" — the codecs (notably `ShardingCodec`) still perform their own storage IO.

The actual win is **synchronous, fused per-chunk execution**: eliminating `BatchedCodecPipeline`'s ~one-coroutine-per-chunk async scheduling, which was the real dominant overhead — not IO/compute mixing. IO ownership stays with the codecs (the *zarrs* model). The `PreparedWrite`/phase abstractions from the first commit are not in the live code path. The "fetch exactly what we need" idea survives only as a read-path optimization inside the sharding codec (`get_ranges_sync` coalescing, vectorized whole-shard bulk decode), not as a pipeline-level phase.

**So the PR description should be rewritten** around "synchronous execution" rather than "separating IO from compute."

### What it actually is now

- `FusedCodecPipeline` runs the codec chain synchronously when all codecs are sync-capable (`SupportsSyncCodec`); falls back to the async path (≡ `BatchedCodecPipeline`) for non-sync stores (ZipStore, fsspec).
- Intended to be behavior-identical to Batched; a pipeline-parity test battery asserts Fused ≡ Batched (same results, same store keys, byte-identical shards).

### User-facing changes

- **Default pipeline flipped** to `FusedCodecPipeline` (the one genuinely user-facing change). One-line revert: `zarr.config.set({"codec_pipeline.path": "zarr.core.codec_pipeline.BatchedCodecPipeline"})`.
- **Threading off by default** (`codec_pipeline.max_workers=1`, sequential). Threading-by-default is deferred (downstream thread-safety; oversubscription on many-core nodes). Opt in with `max_workers > 1`.
- **No public API additions**; import surface unchanged.

### Scope changes since opening

- **Byte-range writes were removed** from this PR (pending a store-interface decision); partial shard writes take the full-rewrite path. ⇒ **the "depends on #3907" note is stale** — #3907 (`SupportsSetRange`) is still open, but this PR no longer depends on it.
- #3908 is merged.

### Performance (latest CodSpeed)

- **×3.1 overall, 0 regressed**, 31 improved, 11 untouched. Sharded morton reads ~×11–12.
- A mid-development write regression (`test_sharded_morton_write_single_chunk`, −38–39%) was fixed (vectorized the partial-shard dict build), which also lifted the overall figure ×2.7 → ×3.1.

### Hardening highlights

- Fixed a spec-threading bug where `FusedCodecPipeline.evolve_from_array_spec` evolved codecs against the original (not forward-threaded) spec, stripping `BytesCodec.endian` for dtype-widening cast_value chains. Root cause was duplicated logic drifting between the two pipelines; a follow-up refactor extracted the shared logic (`evolve_codecs`, `pipeline_supports_partial_decode/encode`) into freestanding functions so the class of bug can't recur. Added a dependency-free regression test on both pipelines.
- Shared `CodecPipelineTests` suite now runs every pipeline-agnostic behavior on **both** pipelines × **sync/async stores**.

### Open items for review

1. Update this description (IO/compute framing → synchronous-execution framing).
2. Default flip: opt-in for a release first, or explicit risk sign-off (mitigated by parity tests + one-line revert).
3. Downstream: numcodecs zarr3 passes; xarray backend suite needs a clean run, we are waiting on https://github.com/zarr-developers/zarr-python/pull/4047 for that.
4. `supports_partial_decode/encode` divergence between the pipelines (deferred; behavior question).
5. Pre-existing Batched transpose+sharding bug (on `main` too; deferred to a separate PR).

d-v-b:
@ilan-gold this is ready for you to dig into

floriankrb:
@d-v-b This looks promising. Is this branch ready to be tested regarding the regression we saw here https://github.com/ecmwf/anemoi-datasets/issues/486?

d-v-b:
@floriankrb yes if you have time to test it, that would be great

src/zarr/codecs/sharding.py
```diff
@@ -416,6 +527,395 @@ def validate(
...
+
+    def _encode_sync(
+        self,
+        shard_array: NDBuffer,
+        shard_spec: ArraySpec,
+    ) -> Buffer | None:
+        """Encode a full shard synchronously.
+
+        Sync counterpart to `_encode_single`. This is reached when a
+        `ShardingCodec` is an *inner* codec of another sharding codec (nested
```
ilan-gold, line 693: Is this true? I hit this path when I ran `test_create_array_with_data_num_gets` which looks pretty un-nested

src/zarr/codecs/sharding.py
```diff
@@ -380,8 +483,16 @@ def to_dict(self) -> dict[str, JSON]:
         }
 
     def evolve_from_array_spec(self, array_spec: ArraySpec) -> Self:
+        # Thread the spec through the inner chain (`evolve_codecs`): each codec
```
ilan-gold: Should this and others be real docstrings?

src/zarr/codecs/sharding.py
```diff
@@ -416,6 +527,395 @@ def validate(
...
+                # matching the async _encode_partial_single path. A per-coordinate
+                # __getitem__ loop here is O(n_chunks) Python overhead that dominates
+                # partial writes into shards with many inner chunks.
+                shard_dict = shard_reader_fb.to_dict_vectorized(
+                    np.indices(chunks_per_shard).reshape(len(chunks_per_shard), -1).T
+                )
+            else:
+                shard_dict = dict.fromkeys(morton_order_iter(chunks_per_shard))
+
+        from zarr.core.codec_pipeline import merge_and_encode_chunk
```
ilan-gold: What do you think about moving these helpers that currently exist in and are shared from `codec_pipeline` into their own file?

src/zarr/codecs/sharding.py
```diff
@@ -416,6 +527,395 @@ def validate(
...
+        buffer_prototype: BufferPrototype,
+    ) -> Buffer | None:
+        """Sync version of _encode_shard_dict.
+
+        Layout via the shared `_build_shard_layout` (offsets already absolute),
+        then a single index encode and concatenation.
+
+        Returns `None` for an all-empty shard (no chunks present).
+        """
+        layout = self._build_shard_layout(shard_dict, chunks_per_shard)
```
ilan-gold, line 934: Instead of materializing `layout` as a `list`, does it make sense to have it be an iterator? This could help save on memory, but not sure.

src/zarr/codecs/sharding.py
```diff
@@ -548,6 +1048,192 @@ def _subchunk_order_iter(
                 raise ValueError(f"Unrecognized subchunk write order: {subchunk_write_order!r}.")
         return subchunk_iter
 
+    def _decode_full_shard_bulk(
```
ilan-gold:
```suggestion
    def _decode_full_shard_bulk_if_uncompressed(
```
or something.  Then you probably don't need such a giant comment in `_decode_partial_sync`

ilan-gold (commented):
A few small comments, may do them myself.

src/zarr/codecs/sharding.py
d-v-b: yeah I think it should be

src/zarr/codecs/sharding.py
d-v-b: I think that's a good idea

src/zarr/codecs/sharding.py
d-v-b: as long as we are sure we aren't consuming the iterator, then passing the (dead, empty) iterator off to a consumer, then sure

src/zarr/codecs/sharding.py
d-v-b: lgtm!

src/zarr/core/codec_pipeline.py
```diff
@@ -54,123 +133,316 @@ def resolve_batched(codec: Codec, chunk_specs: Iterable[ArraySpec]) -> Iterable[
...
 
-    def compute_encoded_size(self, byte_length: int, array_spec: ArraySpec) -> int:
-        for codec in self.codecs:
-            byte_length = codec.compute_encoded_size(byte_length, array_spec)
-            array_spec = codec.resolve_metadata(array_spec)
-        return byte_length
-
 
 @dataclass(frozen=True)
 class BatchedCodecPipeline(CodecPipeline):
```
ilan-gold, line 522: Should we deprecate this? I think so.

tests/test_store/test_local.py
```diff
@@ -8,6 +8,10 @@
 
 import zarr
 from zarr import create_array
+
+# SupportsSetRange import disabled with the byte-range-write tests below
+# (removed from this PR pending a store-interface decision).
+# from zarr.abc.store import SupportsSetRange
 from zarr.core.buffer import Buffer, cpu
```
ilan-gold:
```suggestion

from zarr.core.buffer import Buffer, cpu
```

tests/test_store/test_local.py
```diff
@@ -108,6 +112,58 @@ async def test_move(
...
+    # def test_set_range_sync(
+    #     self, store: LocalStore, start: int, patch: bytes, expected: bytes
+    # ) -> None:
+    #     """set_range_sync should overwrite bytes at the given offset."""
+    #     sync(store.set("test/key", cpu.Buffer.from_bytes(b"AAAAAAAAAA")))
+    #     store.set_range_sync("test/key", cpu.Buffer.from_bytes(patch), start=start)
+    #     result = store.get_sync(key="test/key", prototype=cpu.buffer_prototype)
+    #     assert result is not None
+    #     assert result.to_bytes() == expected
+
```
ilan-gold, line 162:
```suggestion
```

tests/test_store/test_memory.py
```diff
@@ -8,6 +8,10 @@
 import pytest
 
 import zarr
+
+# SupportsSetRange import disabled with the byte-range-write tests below
+# (removed from this PR pending a store-interface decision).
+# from zarr.abc.store import SupportsSetRange
```
ilan-gold:
```suggestion
```

tests/test_store/test_memory.py
```diff
@@ -76,6 +80,59 @@ async def test_deterministic_size(
...
+    #     self, store: MemoryStore, start: int, patch: bytes, expected: bytes
+    # ) -> None:
+    #     """set_range_sync should overwrite bytes at the given offset."""
+    #     store._is_open = True
+    #     store._store_dict["test/key"] = cpu.Buffer.from_bytes(b"AAAAAAAAAA")
+    #     store.set_range_sync("test/key", cpu.Buffer.from_bytes(patch), start=start)
+    #     result = store.get_sync(key="test/key", prototype=cpu.buffer_prototype)
+    #     assert result is not None
+    #     assert result.to_bytes() == expected
+
```
ilan-gold, line 131:
```suggestion
```

ilan-gold (commented):
In general, I don't know how I feel about all these added tests. I'm not a huge fan of testing non-public APIs unless absolutely necessary, so I wonder how much of this can be captured by codspeed. 

That being said, they are all super clear, I just worry about other reviewers.

src/zarr/core/codec_pipeline.py
d-v-b: since we are keeping it as the default for now, we should deprecate in the next round of revisions, when we make the new codec pipeline the default

d-v-b:
@ilan-gold some changes for you to note:
- the batched codec pipeline is now the default
- I added a documentation page entry explaining the motivation for the new codec pipeline, and how to opt in.
- i removed the commented tests for the partial write stuff
- i updated the PR description

docs/user-guide/experimental.md
```diff
@@ -2,6 +2,103 @@
...
+
+`FusedCodecPipeline` is an experimental alternative that runs codec compute *synchronously*, avoiding
+that per-chunk async scheduling overhead. On real workloads the scheduling cost can dominate the
+actual codec work, so removing it is a significant speedup -- especially for **sharded arrays**, where
+a single shard read or write involves many inner chunks.
+
+> **Note:** The win is *not* a faster compressor or a different on-disk format -- the bytes written are
+> identical. It is purely the removal of async scheduling overhead, plus a few vectorized fast paths
+> for dense, uncompressed shards.
+
```
ilan-gold:
```suggestion
A *codec pipeline* is the machinery that turns chunks of array data into stored bytes and back, by running the configured codecs (filters, serializer, compressors) and performing the storage IO.
The default pipeline, `BatchedCodecPipeline`, schedules both the IO and codec work asynchronously -- roughly one coroutine per chunk operation.

`FusedCodecPipeline` is an experimental alternative that runs codec compute and synchronous IO *synchronously*, avoiding that per-chunk async scheduling overhead and nasty [`asyncio.to_thread` overhead](https://github.com/python/cpython/issues/136084).
On real workloads the scheduling cost can dominate the actual codec work, so removing it is a significant speedup -- especially for **sharded arrays**, where a single shard read or write involves many inner chunks.

> **Note:** The win is *not* a faster compressor or a different on-disk format -- the bytes written are
> identical. It is purely the removal of async scheduling overhead, plus a few vectorized fast paths
> for dense, uncompressed shards i.e., removing compute where it is not needed.

```

ilan-gold: I try to do one line per sentence in markdown

docs/user-guide/experimental.md
```diff
@@ -2,6 +2,103 @@
...
+The benefit shows up when storage IO is fast enough that the *scheduling* overhead, not the IO itself,
+is the bottleneck. That means **low-latency stores** -- in particular
+[`zarr.storage.MemoryStore`][] and [`zarr.storage.LocalStore`][].
+
+For high-latency stores (e.g. S3 or HTTP), where each access is dominated by network round-trips, the
+pipeline transparently **falls back to the asynchronous path**, so there is no benefit but also no
+regression. Compute-bound workloads (heavy compression on large chunks) see no regression either,
+since the codec work is the same.
+
```
ilan-gold:
```suggestion
There are two main benefits in this new pipeline:

1. When storage IO is fast enough that the *scheduling* overhead, not the IO itself, is the bottleneck. That means **low-latency stores** that are themselves synchronous -- in particular [`zarr.storage.MemoryStore`][] and [`zarr.storage.LocalStore`][].

2. Whenever codec work that is truly synchronous will not need the overhead of `async` scheduling i.e., inner-chunk codec work in sharding using something like `zstd`. We also now make use of `asyncio.as_completed` so that IO from asynchronous sources can begin decompression immediately.

```

docs/user-guide/experimental.md
```diff
@@ -2,6 +2,103 @@
...
+zarr.config.set({"codec_pipeline.max_workers": 8})
+
+# Or "auto", sized to the number of CPUs.
+zarr.config.set({"codec_pipeline.max_workers": None})
+```
+
+Threading is off by default because it is not universally a win: it slows small chunks (where the
+per-task overhead outweighs the work), and on many-core nodes a pool sized to `cpu_count` can
+oversubscribe workloads that already parallelize at a higher level (e.g. Dask). `max_workers` only
+affects `FusedCodecPipeline`; the default batched pipeline ignores it.
```
ilan-gold: I could do some work on this.  I'll have to check the benchmarks to understand more. But any guidance here would be appreciated

src/zarr/codecs/sharding.py
```diff
@@ -389,8 +459,48 @@ def from_dict(cls, data: dict[str, JSON]) -> Self:
 
     @property
     def codec_pipeline(self) -> CodecPipeline:
+        # Resolve against the configured pipeline (registry default), matching the
+        # rest of this module's use of get_pipeline_class — NOT a hard-coded
+        # BatchedCodecPipeline. This restores main's behavior (#2179) that the
+        # branch had reverted: the inner sub-chunk pipeline follows the same
+        # codec_pipeline.path config as the outer array.
         return get_pipeline_class().from_codecs(self.codecs)
```
ilan-gold:
```suggestion
return get_pipeline_class().from_codecs(self.codecs)
```

src/zarr/codecs/sharding.py
```diff
@@ -439,6 +568,384 @@ def validate(
...
+        `__init__`/`__setstate__` (the single cache mechanism for these
+        builders — do not add another layer inside the body).
+
+        Codecs are evolved with the spec THREADED forward (`evolve_codecs`):
+        each inner codec is evolved against the spec produced by the previous
+        one, not the original chunk spec. Evolving every codec against the same
+        unthreaded spec is the bug shape that stripped `BytesCodec.endian` at
+        the pipeline level (see `evolve_codecs`) — the inner chain must use the
+        same single source of truth.
+        """
```
ilan-gold, line 579: Not sure we really need to state this every time
```suggestion
```

src/zarr/codecs/sharding.py
```diff
@@ -439,6 +568,384 @@ def validate(
...
+
+        chunk_spec = self._get_chunk_spec(shard_spec)
+        return ChunkTransform(codecs=evolve_codecs(self.codecs, chunk_spec))
+
+    def _get_index_chunk_transform(self, chunks_per_shard: tuple[int, ...]) -> Any:
+        """The synchronous transform for the index codec chain.
+
+        Memoized via instance-local `lru_cache` and spec-threaded for the same
+        reasons as `_get_inner_chunk_transform`.
```
ilan-gold:
```suggestion
        Memoized via instance-local `lru_cache`.
```

ilan-gold (commented):
Just little things at this point!

ilan-gold:
I have found Lachi's https://github.com/zarrs/zarr_benchmarks to be the most reflective benchmark of real-world data.  Here are the most relevant results IMO on my 16-core, SSD-backed linux VM from https://www.denbi.de/:

## Read-only

single-threaded, new pipeline:
```python
| Image                              |   Time (s)<br>zarrs<br>rust |   <br>zarr<br>python |   <br>zarrs<br>python |   Memory (GB)<br>zarrs<br>rust |   <br>zarr<br>python |   <br>zarrs<br>python |
|:-----------------------------------|----------------------------:|---------------------:|----------------------:|-------------------------------:|---------------------:|----------------------:|
| data/benchmark_compress_shard.zarr |                        0.98 |                 3.88 |                  1.32 |                           2.16 |                 2.20 |                  2.23 |
```

fully-threaded, new pipeline:
```python
| Image                              |   Time (s)<br>zarrs<br>rust |   <br>zarr<br>python |   <br>zarrs<br>python |   Memory (GB)<br>zarrs<br>rust |   <br>zarr<br>python |   <br>zarrs<br>python |
|:-----------------------------------|----------------------------:|---------------------:|----------------------:|-------------------------------:|---------------------:|----------------------:|
| data/benchmark_compress_shard.zarr |                        1.00 |                 1.60 |                  1.37 |                           2.16 |                 2.76 |                  2.22 |
```

`3.2.1`:
```python
| Image                              |   Time (s)<br>zarrs<br>rust |   <br>zarr<br>python |   <br>zarrs<br>python |   Memory (GB)<br>zarrs<br>rust |   <br>zarr<br>python |   <br>zarrs<br>python |
|:-----------------------------------|----------------------------:|---------------------:|----------------------:|-------------------------------:|---------------------:|----------------------:|
| data/benchmark_compress_shard.zarr |                        0.98 |                 4.00 |                  1.37 |                           2.16 |                 3.97 |                  2.24 |
```

## Roundtrip

single-threaded, new pipeline:

```python
| Image                              |   Time (s)<br>zarrs<br>rust |   <br>zarr<br>python |   <br>zarrs<br>python |   Memory (GB)<br>zarrs<br>rust |   <br>zarr<br>python |   <br>zarrs<br>python |
|:-----------------------------------|----------------------------:|---------------------:|----------------------:|-------------------------------:|---------------------:|----------------------:|
| data/benchmark_compress_shard.zarr |                        1.49 |                16.32 |                  1.94 |                           0.22 |                 2.20 |                  2.24 |
```

fully-threaded, new pipeline:
```python
| Image                              |   Time (s)<br>zarrs<br>rust |   <br>zarr<br>python |   <br>zarrs<br>python |   Memory (GB)<br>zarrs<br>rust |   <br>zarr<br>python |   <br>zarrs<br>python |
|:-----------------------------------|----------------------------:|---------------------:|----------------------:|-------------------------------:|---------------------:|----------------------:|
| data/benchmark_compress_shard.zarr |                        1.52 |                 6.07 |                  2.06 |                           0.22 |                 2.79 |                  2.24 |
```

`3.2.1`:
```python
| Image                              |   Time (s)<br>zarrs<br>rust |   <br>zarr<br>python |   <br>zarrs<br>python |   Memory (GB)<br>zarrs<br>rust |   <br>zarr<br>python |   <br>zarrs<br>python |
|:-----------------------------------|----------------------------:|---------------------:|----------------------:|-------------------------------:|---------------------:|----------------------:|
| data/benchmark_compress_shard.zarr |                        1.44 |                14.55 |                  1.97 |                           0.22 |                 5.35 |                  2.24 |
```

## Smaller subchunks than Presented Here (i.e., `<64x64x64`)

If I alter the benchmarks to have even smaller subchunks in the shards, say `32x32x32`, the difference between rust and python grows, but the difference between the new pipeline and the old one also grows by a similar order of magnitude

## Conclusion

I think we should turn threading on by-default.  Most people are going to be seeing the benefit, if I had to guess, on the local-file-system level. That being said, it would be great if someone could benchmark this on cloud-backed data!

ilan-gold (approved):
This PR should contain no breaking changes and simply add a new pipeline option that is more performant to the current stack. That pipeline includes a bunch of desired features that have been sitting around in our heads for a while and are now materialized, centered around hte notion that "sync should be sync".  It has been tested downstream in the CI here and in https://github.com/scverse/anndata/pull/2529 where everything is passing that would be expected to pass (we have a flag that detects pre-releases so that's what's failing there).

In short, I am confident this PR is ready for prime-time!

d-v-b:
thanks for your help here ilan! this is getting merged. I will follow-up with some docs / QOL changes in separate PRs

PR #3891, opened 2026-04-09
[do not merge] benchmarks + tests for phased codecpipeline

d-v-b:
This PR should not be merged. It contains changes necessary to make the codec pipeline developed in https://github.com/zarr-developers/zarr-python/pull/3885 the default, which allows us to run our full test suite + benchmarks against that codec pipeline class.


d-v-b:
this PR is not needed any more.

PR #3895, opened 2026-04-09
fix: simplify and fix codspeed workflow

d-v-b:
codspeed workflows are [failing](https://github.com/zarr-developers/zarr-python/actions/runs/24181463848/job/70575233873) because the workflow file refers to a nonexistent hatch environment. Also, the workflow installs python, which is not necessary at all. This PR fixes both things.


d-v-b:
looks like something added to main between the last successful benchmark run and this PR has caused a real but small regression in our morton order benchmarks.

maxrjones (approved):

PR #3897, opened 2026-04-10
chore: clean up metadata tests

d-v-b:
Simplifies our v3 metadata tests, and makes the v3 metadata typeddict a bit more accurate.


src/zarr/core/metadata/v3.py
```diff
@@ -411,25 +414,43 @@ def parse_chunk_grid(
...
-    fill_value: object
-    codecs: tuple[str | NamedConfig[str, Mapping[str, object]], ...]
+    chunk_grid: str | NamedConfig[str, Mapping[str, JSON]]
+    chunk_key_encoding: str | NamedConfig[str, Mapping[str, JSON]]
+    fill_value: JSON
+    codecs: tuple[str | NamedConfig[str, Mapping[str, JSON]], ...]
     attributes: NotRequired[Mapping[str, JSON]]
-    storage_transformers: NotRequired[tuple[NamedConfig[str, Mapping[str, object]], ...]]
+    storage_transformers: NotRequired[tuple[str | NamedConfig[str, Mapping[str, JSON]], ...]]
     dimension_names: NotRequired[tuple[str | None]]
```
maxrjones:
```suggestion
    dimension_names: NotRequired[tuple[str | None, ...]]
```
I think the existing type is inaccurate, this may need to change elsewhere too

src/zarr/core/metadata/v3.py
```diff
@@ -411,25 +414,43 @@ def parse_chunk_grid(
...
+    "data_type",
+    "shape",
+    "chunk_grid",
+    "chunk_key_encoding",
+    "fill_value",
+    "codecs",
+    "attributes",
+    "storage_transformers",
+    "dimension_names",
+}
```
maxrjones, line 453: Might be worth adding a test for to make sure`ARRAY_METADATA_KEYS` and `ArrayMetadataJSON_V3` stay in sync:

```
  def test_array_metadata_keys_matches_typeddict():                                                                                                                                                                                                           
      assert ARRAY_METADATA_KEYS == set(ArrayMetadataJSON_V3.__annotations__.keys())  
```

maxrjones (approved):

src/zarr/core/metadata/v3.py
d-v-b: good catch, I will check where else we need to fix this

src/zarr/core/metadata/v3.py
d-v-b: that's a great idea, I will add it

src/zarr/core/metadata/v3.py
d-v-b: i didn't find any other places that refer to this type -- we have a `DimensionNamesLike` type that's wider and it doesn't have this issue, because it's `Iterable`

src/zarr/core/metadata/v3.py
d-v-b: https://github.com/zarr-developers/zarr-python/pull/3897/commits/1ead02835440d46b511dc257481341af3ae5551e

PR #3906, opened 2026-04-14
feat: add IndexTransform library for composable, lazy coordinate mappings

d-v-b:

Adds an experimental TensorStore-inspired coordinate-transform library and lazy array indexing. At the current head the accessor is `.lazy`, and `.result()` materializes a selected view. Ordinary `__getitem__` remains immediate.

```python
from zarr import create_array
arr = create_array(store={}, shape=(100,), dtype="uint8")
arr.lazy[0] = 10
value = arr.lazy[0].result()
```

The selected scalar view must be materialized to obtain that scalar; materializing the original array returns the whole array. Negative coordinates in the transform layer are literal coordinates rather than NumPy end-relative indices.

The transform layer defines domains, output maps, composition, intersections, and translation. It supports planning chained selections before reading. This prototype does not establish universal compatibility or performance parity with existing indexers or TensorStore, and it does not provide lazy writes.

The previous `.z`/`.resolve()` examples described an earlier revision and have been replaced. Historical test counts and broad claims about Dask maintenance have been omitted.


d-v-b:
cc @vincentsarago

vincentsarago:
@d-v-b I'm new to zarr-python indexing, does `.z[0]` select the first value of the first dimension? 

My use case is mainly if I have an array of shape `(n, m, x, y)`,  I want to be able to `select` on the first/second dimension to reduce the size of my array. How would I be able to do it with `.z[]` ?

d-v-b:
> @d-v-b I'm new to zarr-python indexing, does `.z[0]` select the first value of the first dimension?
> 
> My use case is mainly if I have an array of shape `(n, m, x, y)`, I want to be able to `select` on the first/second dimension to reduce the size of my array. How would I be able to do it with `.z[]` ?

on this branch:

```python
>>> arr = create_array(store={}, shape=(100,100), dtype="uint8")
>>> arr.z[0]      
<Array memory://136980013828096 shape=(100,) dtype=uint8 domain={ 0, [0, 100) }>
>>> arr.z[0].shape                 
(100,)
```

the `domain={ 0, [0, 100) }` means "this array is what you get when you take the first index for the first dimension, and all values along the other dimension", which I think is what you want?

vincentsarago:
What is I have 
```
# create a 4d array (time, band, y, x)
arr = create_array(store={}, shape=(2, 3, 100,100), dtype="uint8")
```
and I want to select the arrays for band 1 (second dim) but for all the times, would `arr.z[:, 0]` gives what I want?

d-v-b:
> What is I have
> 
> ```
> # create a 4d array (time, band, y, x)
> arr = create_array(store={}, shape=(2, 3, 100,100), dtype="uint8")
> ```
> 
> and I want to select the arrays for band 1 (second dim) but for all the times, would `arr.z[:, 0]` gives what I want?

yeah it should!

psobolewskiPhD:
I played with this a bit today and this type of slicing is really great:
`data.z[:,:,:,0:100,0:100]`

One thing that tripped me up -- probably just shows my naivety (PBCAK) -- is that e.g. data.z.shape is `AttributeError`. So you need to get your attributes from `data`, but do your slicing with `data.z`.

d-v-b:
> I played with this a bit today and this type of slicing is really great:
> 
> `data.z[:,:,:,0:100,0:100]`
> 
> 
> 
> One thing that tripped me up -- probably just shows my naivety (PBCAK) -- is that e.g. data.z.shape is `AttributeError`. So you need to get your attributes from `data`, but do your slicing with `data.z`.
> 
> 
> 
> 

We should probably have a shape attribute there. Thanks for reporting that, and thanks for trying this branch out. Let me know if you find anything else we need to fix!

psobolewskiPhD:
I guess from my PoV it would be nice if I could just use `data.z` for everything, so it was basically array-like.
Or to flip it around: if `data` acted like `data.z` -- but that would be a big breaking change, so I understand :P

Edit: Don't get me wrong, this is pretty great -- as an example, with this I can remove the dask.array wrapper here:
https://github.com/napari/napari/pull/8260/changes#diff-418087900d622739ab769c612c09fbf012e7555d28d0e75840d145eb055785f0

instead, just check if my array is zarr and if so use `array.z` for slicing.

d-v-b:
> I guess from my PoV it would be nice if I could just use `data.z` for everything, so it was basically array-like. Or to flip it around: if `data` acted like `data.z` -- but that would be a big breaking change, so I understand :P
> 
> Edit: Don't get me wrong, this is pretty great -- as an example, with this I can remove the dask.array wrapper here: https://github.com/napari/napari/pull/8260/changes#diff-418087900d622739ab769c612c09fbf012e7555d28d0e75840d145eb055785f0
> 
> instead, just check if my array is zarr and if so use `array.z` for slicing.

I'm glad it's useful! I agree that this is probably how slicing should work by default. But that would be a big breaking change, so it's far down the road.

The `.z` thing, however, is purely additive so I hope we can ship that in a minor release soon! Ping @zarr-developers/python-core-devs, it would be good to get some feedback on this PR.

psobolewskiPhD:
Makes sense. So maybe in followup _LazyIndexAccessor could have more of the array properties and support more of the numpy protocol, so that array.z could be used as a drop-in for array, but be lazy about slicing, etc.

normanrz:
I like this a lot!
Why did you pick `.z` as a name for the key? I think something like `.lazy` would be more descriptive, but 3 more chars.
tensorstore uses `.result()` to materialize a read, because it models the lazy slicing as futures. Maybe it makes sense to use the same terms? Or even making this API based on futures as well?

d-v-b:
> I like this a lot!
> Why did you pick `.z` as a name for the key? I think something like `.lazy` would be more descriptive, but 3 more chars.

I just wanted something really short. Since this is (IMO) a strictly better slicing API, I wanted to make accessing it as low-friction as possible. But I will totally bow to the will of the crowd here, we can use whatever people find intuitive. 

> tensorstore uses `.result()` to materialize a read, because it models the lazy slicing as futures. Maybe it makes sense to use the same terms? Or even making this API based on futures as well?

I use `resolve` here but I don't feel strongly about it. Changing to `.result()` would be fine! And I think using a futures-based API is a great idea, especially when we get into lazy writes, which are currently not supported in this PR. If we do go for a futures API, we should think about how to use this comprehensively, e.g. with our stores as well.

vincentsarago:
I agree with @normanrz, the `.z` keys feels a bit weird to me. I think something like `.lazy_select(...) -> _LazyIndexAccessor` makes things a bit more descriptive. 

the length of `.z` make me feel like it is an attribute not a method.

d-v-b:
what about `.select` or `.sel`?

jsignell:
Just for reference: xarray uses `.sel`, `.isel` and `.loc` (https://docs.xarray.dev/en/latest/user-guide/indexing.html#quick-overview). I guess in that naming scheme what you have in this PR would more like `.iloc`

d-v-b:
ping @ilan-gold for visibility

src/zarr/core/transforms/transform.py
```diff
@@ -0,0 +1,932 @@
...
+  maps. No I/O occurs. This is how lazy slicing works.
+
+- **intersect(output_domain)** — restrict to storage coordinates within a
+  region. This is chunk resolution: "which of my coordinates fall in this
+  chunk?"
+
+- **translate(shift)** — shift all output coordinates. This makes coordinates
+  chunk-local: "express my coordinates relative to the chunk origin."
+
+- **compose(outer, inner)** — chain two transforms. See ``composition.py``.
```
ilan-gold: I think it could be cool to have different kinds of composition like union and I don't see that anywhere intthe PR - it seems that only `array.z[my_transform][my_other_transform_relative_to_the_first]` is supported wheres it could be cool to have `array.z[my_transform + my_other_transform]`.  Concretely, if I wanted `(slice(0, 10), slice(0, 10))` and then also `(slice(50, 60), slice(50, 60))` from an array, how would I do that? `np.concatenate` + `np.arange` to generate outer indexers is the only option AFAICT but that kind of stinks.

ilan-gold (commented):
> use data structures compatible with tensorstore. tensorstore's indexing machinery serializes to JSON. if we adopt the same patterns, we are closer to using tensorstore as an optional backend.

I'm sure `zarrs` could also support this but I wonder about JSON performance for things like vindex/oindex where you could easily have 10's of thousands of individual coordinates.

src/zarr/core/transforms/transform.py
d-v-b: that's a very cool idea, not sure about overloading the addition operator, but some method for combining indexing expressions seems nice. I'll see what we need for this

d-v-b:
I could be wrong but I strongly suspect this PR is too large to get reviewed in a reasonable amount of time 😅 . I'm going to close it, and start working on breaking the components into smaller PRs, based on the conversation here. Thanks to everyone for suggestions! 

A few concrete changes I'm making based on what yall said:
- instead of `.z` it's going to be `.lazy`
- instead of `.lazy[...].resolve()` it's going to be `.lazy[...].result()`

anything else I should fold in to the next iteration?

.claude/hooks/.logs/hook-log.jsonl
```diff
@@ -0,0 +1,203 @@
+{"ts":"2026-06-30T08:02:48.679Z","hook":"stop-notify","projectName":"hopeful-euclid-d75809","duration":"2m 17s","iterations":0}
```
maxrjones: should we .gitignore .claude?

d-v-b: i should tell claude never to go git add -A!

d-v-b: but updating .gitignore seems like a good idea. I assume codex has a similar private directory we should exclude?

maxrjones: > but updating .gitignore seems like a good idea. I assume codex has a similar private directory we should exclude?

done in https://github.com/zarr-developers/zarr-python/pull/4041

d-v-b: thanks max!

emfdavid:
Following this closely as a downstream consumer and I'd like to help get it over the line — it's a big PR and the coordinate algebra is exactly the primitive a data loader wants under it. A few concrete ways I could contribute; happy to take whichever is most useful to you.

**Context:** [insitubatch](https://github.com/emfdavid/insitubatch) plans a training batch as a set of chunk reads and gathers decoded chunks into one buffer — i.e. it hand-rolls a sample→chunk coordinate mapping today. `IndexTransform` + the `chunk_projections` partition API is the thing I'd rather build on than maintain myself, so I have a real incentive to see this land.

**Ways I could help:**

1. **Dogfood it against a real consumer.** I can map our planner onto `IndexTransform` (`.intersect`/`.translate`/`chunk_projections`) and report back friction, gaps, and — relevant to the `benchmark` label — whether the ~18% read regression actually hits a consumer that drives the *transform layer directly* vs. the eager `__getitem__` path. A second real user exercising the resolver may surface or de-risk things the test matrix doesn't.

2. **Take the serialization perf item** (the large vindex/oindex JSON concern). My read is it's two separate problems and worth splitting:
   - *Small/common case:* `output_index_map_to_json` does `index_array.tolist()` (N boxed Python ints) before encode. An optional numpy-native fast encoder (e.g. orjson's `OPT_SERIALIZE_NUMPY`) behind a stdlib fallback skips the boxing — a constant-factor win, no hard dependency added.
   - *Genuinely-large `ArrayMap`:* the real problem is representational — a big index array as inline JSON decimals is huge to produce *and* to reparse, on both encode and decode, and no encoder speed fixes O(n)-huge text. Worth noting TensorStore stores these inline too (and its docs acknowledge index-array representation size growing under composition), so this isn't a parity gap — but an optional out-of-line / binary encoding (base64-packed buffer or sidecar reference) would fix size, encode, and decode at once. The trade-off is a deliberate divergence from the TensorStore-compatible JSON, so it'd want to be opt-in rather than the default wire format.

   I'm happy to prototype either tier if that's a useful thing to peel off.

3. **Feed a consumer requirement for the union / multi-disjoint-window work** that's noted as future. Our gather is N sample-windows resolved together per batch, so multi-window union isn't hypothetical for us — I can write it up as a concrete use case (or contribute) whenever that piece is in scope.

Let me know what's actually useful vs. noise.

d-v-b:
> Dogfood it against a real consumer. I can map our planner onto IndexTransform (.intersect/.translate/chunk_projections) and report back friction, gaps, and — relevant to the benchmark label — whether the ~18% read regression actually hits a consumer that drives the transform layer directly vs. the eager __getitem__ path. A second real user exercising the resolver may surface or de-risk things the test matrix doesn't.

Great idea, and thanks for offering your help. I'm working on moving the lazy indexing logic here out into a separate package in the zarr-python workspace. That PR should be much easier to land, since it wouldn't change anything about zarr python. Once that package is up and running, we can publish it on pypi and you can start trying it out. How does this sound?

> Take the serialization perf item (the large vindex/oindex JSON concern). My read is it's two separate problems and worth splitting:

    Small/common case: output_index_map_to_json does index_array.tolist() (N boxed Python ints) before encode. An optional numpy-native fast encoder (e.g. orjson's OPT_SERIALIZE_NUMPY) behind a stdlib fallback skips the boxing — a constant-factor win, no hard dependency added.
    Genuinely-large ArrayMap: the real problem is representational — a big index array as inline JSON decimals is huge to produce and to reparse, on both encode and decode, and no encoder speed fixes O(n)-huge text. Worth noting TensorStore stores these inline too (and its docs acknowledge index-array representation size growing under composition), so this isn't a parity gap — but an optional out-of-line / binary encoding (base64-packed buffer or sidecar reference) would fix size, encode, and decode at once. The trade-off is a deliberate divergence from the TensorStore-compatible JSON, so it'd want to be opt-in rather than the default wire format.

I'm happy to prototype either tier if that's a useful thing to peel off.

I think we can sidestep this for now, because the JSON thing is only relevant when / if we want to interchange array selections expressions over the wire. We don't need a performant JSON serialization inside Zarr-Python itself.

> 3\. **Feed a consumer requirement for the union / multi-disjoint-window work** that's noted as future. Our gather is N sample-windows resolved together per batch, so multi-window union isn't hypothetical for us — I can write it up as a concrete use case (or contribute) whenever that piece is in scope.

this is a shared need -- see https://github.com/zarr-developers/zarr-python/pull/4028. I think the lazy-indexing-oriented API would look like this: you declare the N regions of the array you want to read, then you arrange those N regions into a concatenated array, and then you submit the request to fetch that array to a zarr backend that can efficiently do all the IO you need. This is just an idea right now, but I think it would be a very clean API.

PR #3907, opened 2026-04-15
feat: add SupportsSetRange protocol and store implementations

d-v-b:

Adds `SupportsSetRange` for synchronous and asynchronous writes into an existing stored value. The current implementation covers MemoryStore, LocalStore, and prefixed ManagedMemoryStore operations; GpuMemoryStore explicitly opts out.

A write must fit within the existing value. These operations provide a building block for updating shard byte ranges without replacing the entire object. They do not by themselves provide transactional shard updates or crash recovery; callers must coordinate overlapping writes and writes racing with set/delete.

The branch contains per-key and filesystem locking work. The review discussion records remaining limitations around rename-based locks and interactions with other operations; this description does not assert that these limitations have been resolved.


maxrjones:
should this get the same design pivot as https://github.com/zarr-developers/zarr-python/pull/3925#issuecomment-4425310058 started in https://github.com/zarr-developers/zarr-python/pull/3925, regarding protocols vs. abc methods?

Do you remember where you were previously pointed to using protocols over methods despite the weight of our existing store API? It might be helpful to quickly jot down our decisions here (use methods for now, plan for a better protocol-based store API in the future) in either https://zarr.readthedocs.io/en/stable/contributing/ or a CLAUDE/AGENTS.md for future reference.

d-v-b:
byte-range writes are an optional behavior that only a handful of "niche" stores support (local and memory). There's not really a sensible fallback or default implementation, (unlike `get_ranges`). So it makes sense for stores to opt in rather than opt out.

And if we made this a method on the `Store` abc, callers would need to check for `NotImplementedError` to figure out of the store _really_ supports it, and the method would clutter the signatures of most stores that will never support it (cloud storage).

I don't think we can categorically say "no" to adding functionality to stores or codecs via protocols. There's already a precedent for defining extra functionality with semi-structural mixins: see https://github.com/zarr-developers/zarr-python/blob/7e58df003f528470f06d6fc689099c4f04ef0c95/src/zarr/abc/codec.py#L256. Arguably this should have been a protocol from the start.

maxrjones:
I'd prefer someone whose work is more oriented towards local/HPC filesystems review this PR if they're available and willing (@LDeakin and @ilan-gold come to mind). 

I'm not fully prepared to discuss tradeoffs, but lack of a concurrency/atomicity contract in the docstring raised a few questions for me:

  1. Is parallel `set_range` to disjoint ranges of the same key supposed to be safe? The motivating sharded-write use case suggests yes, but `LocalStore` doesn't seem to have locking.
  2. Is `set_range` racing against `set` defined?
  3. Should crash-mid-write atomicity be a protocol requirement?

d-v-b:

Disjoint byte ranges are the intended safe case for this protocol. Overlapping range writes and range writes racing with set/delete require caller coordination.

Correction/clarification for the current branch: LocalStore's whole-value `_put` already uses `_atomic_write` with a temporary file and replacement. That does not make the new in-place `_put_range` operation crash-atomic. The current SupportsSetRange contract explicitly permits a partially updated value after a crash; transactional shard updates and durability require additional coordination or recovery machinery.

The intended caller is the sharding codec when inner chunks have deterministic encoded sizes. Adding a range-write method is a building block for that use case, not by itself a guarantee of atomic subchunk updates.

src/zarr/storage/_local.py
```diff
@@ -77,6 +78,13 @@ def _atomic_write(
         raise
 
 
+def _put_range(path: Path, value: Buffer, start: int) -> None:
+    """Write bytes at a specific offset within an existing file."""
+    with path.open("r+b") as f:
+        f.seek(start)
+        f.write(value.as_numpy_array().tobytes())
```
mkitti: Do we need a lock here? Or maybe use `pwrite` when possible and this shim otherwise?

```
import threading

file_lock = threading.Lock()

def pwrite_cross_platform(file_object, data, offset):
    with file_lock:
         file_object.seek(offset)
            file_object.write(data)
        finally:
            file_object.seek(current_pos)
```

mkitti (changes requested):
I think we may need to try some kind of file or thread locking. At the very least, we should ensure that the partial writes are atomic.

src/zarr/storage/_local.py
d-v-b: if we introduce locking we need to be careful about the guarantees we make. thread-based locks will only block races in the same python process, which means dask can still set up races. and file-based locks are only reliable on a subset of storage backends.

mkitti: The bare minimum is document the need for some kind of concurrency control when doing this.

d-v-b: sounds good, I can add that. the only anticipated consumer of this API will be the sharding codec when targeting uncompressed data in memory or local storage, so the burden of coordination will be on that call site.

mkitti:
Thinking about this a little more, I'm not sure if we should expose all of this as public API.

For the public API, what we could do in lieu of file locking is use a context manager that does the following:
1. On entering the context, moves, copies, or creates a hard link of the current file being edited to a temporary "lock" file. This is similar to the atomic file write mechanisms that are currently implemented.
2. Allows for partial writes to the file of interest within the context
3. On exit, flushes the partial writes to disk and then "moves" the current file back to the orignal location.

What we have now here increasingly looks to me like low-level helper functions than something that should not be exposed to external client software directly. Also, this encourages clients to aggregate partial chunk updates which would increase the potential overhead of opening and closing the file many times.

Under parallel executation situation, we should encourage a single-writer multiple-reader pattern (SWMR), a concept that I'm borrowing from HDF5:
* https://support.hdfgroup.org/documentation/hdf5-docs/advanced_topics/intro_SWMR.html
* https://docs.h5py.org/en/stable/swmr.html

In this case, there should be a single writer that holds the context and lock for doing partial writes to the file. Other parallel units then should communicate with the single writer if they have data that needs to be written. This allows the single writer to coordinate concurrency and ensure consistent atomic partial writes. In some cases, the single writer might even be able to coalesce writes into a single operation. This might involve some buffering of the writes in memory before actually flushing these to disk. If we are smart about memory paging, we could write in units of "super-chunks" that provide more granularity than an entire shard file but that are coarser than individual chunks. At the moment, the I/O bottleneck is often not shear write speed but rather the number of system calls and the required context switching.

d-v-b:
@mkitti thanks, that's insightful. I'm perfectly happy making this private API for now. But I would caution that what you propose sounds like a major (albeit helpful) change to the way stores work today, whereas just adding this method here as-is is closer to the minimal required functionality to support atomic subchunk writes inside a shard. 

We should probably agree on scope. IMO, in the short term it would be useful to expose range writes via an opt-in mechanism. In the medium-to-long term, we should figure out an elegant and safe way to express this functionality.

mkitti:
The problem my suggestion is trying to fix is actually to make this operation look more like the prior store operations. I started to have to worry about locks and concurrency because with these partial writes it became unclear if the key-value pair had a single owner or multiple owners.

The prior store API were either pure functions (`get_*`) or operated on the entire value associated with a key. These partial write operations break that because they manipulate part of the value, allowing the possibility of multiple partial updates which may have to be deconflicted because it is not clear if the partial updates are coordinated.

My suggestion above is more consistent with the prior API because it forces a macro operation on the entire value. It just implements that operation as a series of micro operations within the macro operation. The important aspect here is that the ownership of the whole value must be clear.

Once we establish atomic ownership of the whole key-value, I think there is now a clear path to extend this operation to other stores. These partial updates _could_ be implemented as whole value updates. The partial write operation is just an optimization available to certain stores.

d-v-b:
> I started to have to worry about locks and concurrency because with these partial writes it became unclear if the key-value pair had a single owner or multiple owners.

I am not sure thinking about things in terms of ownership helps us much here. When Zarr writes to files / objects, there's always the possibility of a third party accessing the same file / object at the same time. In fact, that's kind of the whole point of the feature in this PR: we want multiple loosely coordinated writers to write subchunks to the same shard. Those writers might be separate threads, or even separate processes. This is only safe under special conditions, but under those conditions it would be a very useful feature.

So in general a single store instance doesn't own the objects it describes. That's unrealistic given the open nature of general storage backends, and also the specific goal of this PR. It might be more realistic to focus on failure modes: e.g., when can a store operation leave a file / object in an undefined state, and what can we do about that?

mkitti:
Prior to this pull request though, I don't think you would expect that a value to have orignated partly from one write and partly from another write leaving the combined value in perhaps in an inconsistent state. It would be one or the other. There indeed may still be a race condition, but at least the chunk would be internally consistent and decodable.

The owner could partition that ownership out to concurrent processes by issuing licenses to certain byte ranges. That could occur by perhaps providing a buffer or a dict-like store structure to write into.

The array provided by TensorStore's virtual chunked to user functions is one example of this:
https://google.github.io/tensorstore/python/api/tensorstore.virtual_chunked.html#

Another example is how [ImarisWriter](https://github.com/imaris/ImarisWriter) works.

d-v-b:
> Prior to this pull request though, I don't think you would expect that a value to have orignated partly from one write and partly from another write leaving the combined value in perhaps in an inconsistent state. It would be one or the other. There indeed may still be a race condition, but at least the chunk would be internally consistent and decodable.
> 
> The owner could partition that ownership out to concurrent processes by issuing licenses to certain byte ranges. That could occur by perhaps providing a buffer or a dict-like store structure to write into.
> 
> The array provided by TensorStore's virtual chunked to user functions is one example of this: https://google.github.io/tensorstore/python/api/tensorstore.virtual_chunked.html#
> 
> Another example is how [ImarisWriter](https://github.com/imaris/ImarisWriter) works.

Agreed, I think our sharding write path is missing a key : value model for the contents of a shard. If they keys are strings / bytes and the values are buffers and / or arrays, we would need a transformation on that key : valure that resolves each key to a byte range (with space for the index). We have something roughly like this today, but IMO the abstraction is not complete. Partitioning these keys (byte ranges) among workers serves as the substrate for a coordination / allocation mechanism.

But I think this is out of scope for this PR. I think we need to get to that final state incrementally, and IMO the first step is simply defining a low-level routine stores can use for writing a byte range into an object, which is the goal of this PR. 

If that makes sense, then:
- what needs to change in this PR
- what should the follow-up PRs contain?

mkitti:
# Local

1. Check for the existence of key specific lock file, wait if it exists. Consider checking for stale lock.
2. Obtain a (per-key?) thread lock
3. Check for the existence of key specific lock file. If it exists, release lock and return to step 1.
4. Move file to key specific lock file location
5. Release thread lock, file lock will maintain lock

(Above steps could be aggregated later into a single context manager enter)

6. Perform partial write on lock file location

(Below steps could be aggregated later into a single context manager exit)
7. Obtain (per-key?) thread lock
8. Move lock file back to original location
9. Release thread lock

Note: A slower but more robust version makes a copy of the file to lock file, then moves it back on exit.

# Memory (not shared)

We could follow the same procedure as above, but the key lock may not be necessary.

1. Obtain (per-key?) lock
2. Perform partial write
3. Release (per-key?) lock

mkitti:
The above procedure is the test and test-and-set mutual exclusion pattern:
https://en.wikipedia.org/wiki/Test_and_test-and-set?wprov=sfla1

d-v-b:
Do you have time to oversee an implementation of this? If so feel free to push to this branch

src/zarr/storage/_memory.py
```diff
@@ -194,6 +194,26 @@ async def delete(self, key: str) -> None:
...
+    async def set_range(self, key: str, value: Buffer, start: int) -> None:
+        self._check_writable()
+        await self._ensure_open()
+        self._set_range_impl(key, value, start)
+
+    def set_range_sync(self, key: str, value: Buffer, start: int) -> None:
+        self._check_writable()
+        if not self._is_open:
+            self._is_open = True
+        self._set_range_impl(key, value, start)
```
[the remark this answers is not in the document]
d-v-b: good catch, I found this locally as well. IMO it was a mistake for gpumemorystore to subclass memorystore. If there's a safe way to implement this feature for cuda-backed memory, we should implement it. otherwise, the gpu memory store needs to opt out

mkitti:
## Elaboration of the `set_range` Concurrency Plan

### Background

This elaborates on my [earlier outline](https://github.com/zarr-developers/zarr-python/pull/3907#issuecomment-4583507433). The current `_put_range` in `LocalStore` (and `_set_range_impl` in `MemoryStore`) have no concurrency control. This plan provides it.

---

### Why the Current Implementation Is Unsafe

`f.seek(start); f.write(view)` is **not atomic** — it is two separate syscalls. Two concurrent callers can interleave their seek/write pairs, corrupting each other's data even for non-overlapping ranges (the file offset is per-file-description, not per-range). The memory store's numpy array slice assignment has the same problem for concurrent threads.

---

### Local Store — Step-by-Step Elaboration

#### Lock File Naming
Each key `k` gets a lock file at `root / (k + ".__lock__")`. The name must be deterministic and not a valid Zarr key so it cannot collide with real data.

#### Step 1 — Check for existing lock file, wait if present
```python
lock_path = root / (key + ".__lock__")
while lock_path.exists():
    await asyncio.sleep(POLL_INTERVAL)
```
This is the **outer spin-wait**. The stale-lock check belongs here: if `lock_path.stat().st_mtime` is more than `STALE_TIMEOUT` seconds old, assume the previous holder crashed and proceed (optionally logging a warning). Stale detection is essential — without it, a crashed writer permanently blocks the key.

#### Step 2 — Acquire the per-key in-process lock
```python
async with _key_lock(key):  # asyncio.Lock keyed by `key`
```
This is an **in-process serialization point**. Multiple coroutines running in the same event loop can all pass the step 1 existence check simultaneously (they all see `exists() == False` before any of them creates the lock file). A per-key `asyncio.Lock` (stored in a `weakref`-keyed dict) ensures only one coroutine proceeds to the filesystem rename at a time.

#### Step 3 — Re-check for lock file inside the in-process lock
```python
if lock_path.exists():
    # release in-process lock, loop back to step 1
    continue
```
This is the **double-checked locking** pattern. Between step 1 and step 2, another coroutine or process may have acquired the file lock. After acquiring the in-process lock we must re-verify before doing the filesystem rename.

#### Step 4 — Move the file to the lock location (atomic acquire)
```python
key_path.rename(lock_path)  # POSIX: atomic
```
On Linux/macOS, `rename(2)` is atomic — it either moves the file or it doesn't. This is the **actual lock acquisition**. Any other process that checks for `lock_path.exists()` after this point will block in step 1.

On Windows, `rename` raises `FileExistsError` if the destination exists, so this is safe-by-exception there too. The existing `_safe_move` helper in `_local.py` already encodes this platform distinction.

#### Step 5 — Release the in-process lock; file lock is now held
The per-key `asyncio.Lock` can now be released. The filesystem state (lock file exists) takes over as the concurrency signal visible across processes. The in-process lock was only needed to prevent duplicate rename races within the same process.

#### Step 6 — Perform the partial write at `lock_path`
```python
with lock_path.open("r+b") as f:
    f.seek(start)
    f.write(value.as_buffer_like())
    f.flush()
    os.fsync(f.fileno())  # optional; required if crash-safety is needed
```
All writes happen against `lock_path`. The original path is temporarily absent — callers that do `get(key)` during this window will get a `KeyError`. This is an acceptable trade-off for the sharding use case where the shard file is pre-allocated and readers are expected to wait.

#### Step 7 — Re-acquire the in-process lock for the rename-back
Same `asyncio.Lock`. Needed to serialize the release rename against any concurrent in-process coroutine that passed step 1, is now waiting at step 2, and would otherwise proceed to step 3 the moment this lock is released.

#### Step 8 — Move lock file back to original path (atomic release)
```python
lock_path.rename(key_path)  # atomic on POSIX
```
The file is back at its canonical location. The lock is released. The write is now visible.

#### Step 9 — Release the in-process lock

---

### Slower/More Robust Variant (Copy Instead of Move)

Instead of step 4 being a rename, it is a `shutil.copy2`:
- The original file remains at `key_path` throughout, so readers never see a `KeyError`.
- On crash between steps 6 and 8, the original (pre-write) file is still intact.
- Downside: double the disk I/O and requires enough free space for two copies.

This is preferred when partial reads during an ongoing write are acceptable (e.g., intermediate shard states are still valid).

---

### Context Manager API

Steps 1–5 map cleanly to `__aenter__` and steps 7–9 to `__aexit__`. This suggests a richer protocol method:

```python
@asynccontextmanager
async def set_range_context(self, key: str) -> AsyncIterator[None]:
    # __aenter__: steps 1–5 (acquire lock once)
    ...
    yield  # caller performs one or more set_range calls inside the context
    # __aexit__: steps 7–9 (release lock once)
    ...
```

This is important for the sharding use case: a shard codec writing N subchunks should acquire the lock **once** and batch all N partial writes inside the context, rather than paying the full lock-acquire/release cost N times. This also motivates the note about aggregating steps into a context manager.

---

### Interaction With `set`

The plan as written does not address `set` vs `set_range` races. Worth noting: `set` (via `_atomic_write`) writes to a temp file and renames it to `key_path`. If `set` finishes during an active `set_range` lock (steps 4–8), the new file ends up at `key_path`. When step 8 runs, `lock_path.rename(key_path)` atomically **overwrites** that new file with the partially-written version — silently discarding the `set`. A complete solution requires `set` to be aware of the lock file (either wait for it, or also take ownership of it via rename to `lock_path` before writing).

---

### Memory Store — Elaboration

Since `MemoryStore` is never shared across processes, the file-based lock is unnecessary. A per-key `asyncio.Lock` (stored in a `dict[str, asyncio.Lock]`) is sufficient:

```python
async with self._key_locks.setdefault(key, asyncio.Lock()):
    self._set_range_impl(key, value, start)
```

The existing `_set_range_impl` already manipulates the backing numpy array correctly; it just needs the lock wrapper. The `set_range_sync` path would use a `threading.Lock` from a parallel dict.

---

### Implementation Staging

Given the preference for a minimal footprint in the near term:

1. **Now**: Add in-process locking to `MemoryStore.set_range` (trivial, low-risk).
2. **Soon**: Add the file-lock mechanism to `LocalStore.set_range` with the double-check pattern above.
3. **Later**: Introduce `set_range_context` as a separate method/protocol entry point for the batched sharding use case.
4. **Future**: Coordinate `set` with the lock file to close the `set`-vs-`set_range` race.

d-v-b:
(this is claude writing)

I pushed some follow-up fixes (e02437cd5) on top of @mkitti's per-key locking from #178:

- **`GpuMemoryStore` no longer falsely advertises `set_range`.** `MemoryStore`/`LocalStore` now satisfy `SupportsSetRange` *structurally* (dropped the nominal base class) so `GpuMemoryStore` can disclaim the methods (`set_range = None`); `isinstance(gpu_store, SupportsSetRange)` is now `False` instead of a false positive. The inherited impl would have mutated a host copy of the GPU buffer and silently lost the write.
- **Uniform out-of-bounds handling.** Added `_check_set_range_bounds` so a too-large write raises a clear `ValueError` in both stores, instead of `LocalStore` silently extending the file and `MemoryStore` raising an opaque numpy error. (Matches `zarrs`, which validates ranges rather than extending.)
- **`ManagedMemoryStore.set_range`/`set_range_sync`** now apply the store's path prefix like the other key-bearing methods.
- Expanded the `SupportsSetRange` docstring with the caller-owns-consistency contract and recorded the opt-in-protocol vs. universal-method design decision.

Verified on GPU hardware (cupy) + full store suite.

---

A code review of the branch flagged some correctness gaps in the rename-based file locking from #178 — @mkitti, these are your call on how (or whether) to address:

**High — concurrent `set` is silently clobbered.** Since the lock file *is* the data file (renamed `path` ↔ `lock_path`), while a `set_range` is in progress the key disappears from `get`/`exists`/`list`, and a concurrent `set` writes to `path` and is then unconditionally overwritten when `set_range`'s `finally` renames the lock file back. The "racing against `set` is a race" disclaimer doesn't quite capture that one write is *guaranteed* lost. A sidecar lock file (0-byte marker, data written in place) would avoid this.

**Medium — stale-lock detection is effectively dead + can busy-spin.** `_is_stale_lock` uses the lock file's mtime, but `rename`/`link` don't refresh mtime, so any chunk file older than `_LOCK_STALE_TIMEOUT` (60s — i.e. essentially all stored chunks) reads as stale the instant a second writer arrives. The stale branch takes no recovery action and the outer `while True` has no sleep on the failed-acquire path, so it can tight-loop until the holder's first write bumps mtime (indefinitely if a process crashed mid-lock).

**Medium — `ManagedMemoryStore` per-key locks aren't shared across same-name instances.** Two `ManagedMemoryStore` handles for the same `name` share `_store_dict` via the registry but each builds its own `_key_locks`/`_key_locks_sync`, so `set_range` from the two handles can race on the same backing dict. The registry shares the data; it should share the locks too.

**Low — missing-key error type.** `_safe_move` raises a raw `FileNotFoundError` for a missing key; `MemoryStore` raises `KeyError`. Normalizing would be friendlier.

(There's also a separate Medium on `FsspecStore.from_url` unconditionally setting `_owns_fs = True` even though `fsspec.url_to_fs` returns instance-cached filesystems — closing one store can break another sharing the cached fs. Likely wants `skip_instance_cache=True` or a cache check.)

I deliberately did **not** touch the `#178` locking or the fsspec ownership code here since those are your design decisions — happy to help with any of them if useful.

mkitti:
The concurrent `set` issue is pre-existing. We don't guarantee that a `set` operation will not be lost if you are using concurrency and have multiple set operations. Here we're just guaranteeing that that two partial sets will not occur since two partial sets without concurrency controls could result in an inconsistent value when combined. If this is an issue, I think that is best explored in its own pull request.

raphaeljolivet:
Hi, 

We needed something like this to ensure only partial writes of shard files on disk when updating (append in our case) a a single chunk.

Our concern was it might quickly "burn" SD cards : We use Zarr storage directly on our device (fisheye cameras).

I vibe coded an implementation of a partial update based on this branch : 
- If the layout of existing chunks have not changed, it only writes the required chunks instead of the whole block.

I probed the actual write of pages/disk block and it works fine.

I understand that we lose atomicity by doing so (no more write in a separate file then move it) but we are happy with this trade off.

You can find it there : 
https://github.com/zarr-developers/zarr-python/commit/da9e9fd571e54eb9e9037e0df0c6b1822d5f4c87

PR #3908, opened 2026-04-15
perf: cache default ArraySpec for regular chunk grids

d-v-b:
For regular grids, all chunks have the same codec_shape, so we can
build the ArraySpec once and reuse it for every chunk — avoiding the
per-chunk ChunkGrid.__getitem__ + ArraySpec construction overhead.

Adds _get_default_chunk_spec() and uses it in _get_selection and
_set_selection. Saves ~5ms per 1000 chunks.

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>


src/zarr/core/array.py
```diff
@@ -5370,6 +5370,37 @@ def _get_chunk_spec(
     )
 
 
+def _get_default_chunk_spec(
+    metadata: ArrayMetadata,
+    chunk_grid: ChunkGrid,
+    array_config: ArrayConfig,
+    prototype: BufferPrototype,
+) -> ArraySpec | None:
```
maxrjones: given the name of `_get_default_chunk_spec`, should this be a `ChunkSpec`?

d-v-b: it's the arrayspec for a chunk, not a chunkspec. And the consumer needs an `ArraySpec`, so we can't change the return type.

maxrjones: ok, still a bit confusing but I guess that might relate to the fact that a better design wouldn't need a per-chunk ArraySpec.

My only other question is why build a new function rather than making `chunk_coords: tuple[int, ...]` optional in `_get_chunk_spec`, such that it could return a default?

d-v-b: what would the default chunk coordinates be? the "origin" chunk coordinate depends on the dimensionality of the array

d-v-b: Part of the goal here is to avoid object creation overhead inside `_get_chunk_spec`. Adding default parameters to `_get_chunk_spec` would not help us, because we would still create many identical `ArraySpec` objects. The only change to `_get_chunk_spec` that would help is adding a caching layer via `@lru_cache`, I should see if that works

maxrjones: yeah, I agree with the goal. It might just be me, but I find the code confusing which was the reason for all my comments. I think a flow that would be more interpretable but accomplish the same could look like:

```python
regular_grid = chunk_grid.is_regular:
if regular_grid:
	regular_chunk_spec = _get_chunk_spec(metadata, chunk_grid, _config, prototype)
results = await codec_pipeline.read(
            [
                (
                    store_path / metadata.encode_chunk_key(chunk_coords),
                    regular_chunk_spec
                    if regular_grid
                    else _get_chunk_spec(metadata, chunk_grid, chunk_coords, _config, prototype),
                    chunk_selection,
                    out_selection,
                    is_complete_chunk,
	...
```

d-v-b: good idea! I implemented this in https://github.com/zarr-developers/zarr-python/pull/3908/commits/ff088912c80cf66ddecd7aa90008f8e584595b8a

maxrjones (approved):

PR #3909, opened 2026-04-15
TESTS: don't force rectilinear config true in the test

ianhi:
The new array strategy for stateful testing was ignoring the config setting and forcibly changing it - which gives no way for a downstream consumer to opt of out. This makes it respect the config. There is a small worry that now one of our strategies depends on a global state (potential cause of flakyiness) but I'm not really sure there's a clean way around that. 

@dcherian 

TODO:

### DIFF
--- a/src/zarr/testing/strategies.py
+++ b/src/zarr/testing/strategies.py
@@ -1,7 +1,6 @@
 import math
 import sys
 from collections.abc import Callable, Mapping
-from contextlib import nullcontext
 from typing import Any, Literal
 
 import hypothesis.extra.numpy as npst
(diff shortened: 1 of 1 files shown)


src/zarr/testing/strategies.py
```diff
@@ -426,6 +422,7 @@ def chunk_grids(
     draw: st.DrawFn, *, shape: tuple[int, ...]
 ) -> RegularChunkGridMetadata | RectilinearChunkGridMetadata:
     """Generate either a RegularChunkGridMetadata or RectilinearChunkGridMetadata.
+    Depending on whether the config is set (WARN: look here if we get a flaky error)
```
dcherian: huh?

dcherian (approved):

src/zarr/testing/strategies.py
ianhi: That was all me - so any confusion is my fault.

This makes this strategy dependent on global state, i.e. a recipe for a flaky strategy down the line. I wanted to leave a note for the future in case that happens. Don't think I really hit the mark here though

dcherian: ok let's write in a longer comment then :)

d-v-b:
i'm confused -- don't we want to turn on rectilinear chunks explicitly before this test?

d-v-b:
or is the idea to leave that detail to the configuration at test time

d-v-b:
also... we should rethink putting the config check inside the chunk grid class itself. If someone wants to import the class and construct it, that should be allowed.

ianhi:
> i'm confused -- don't we want to turn on rectilinear chunks explicitly before this test?

Yes for the zarr-python tests, but not within the the strategy. The strategy is public API and might be run on a system that doesn't support rectilinear chunks. For example icechunk format 1 just straight up doesn't support it, so I need to be able to turn it off via the config and have that be respected.

In the actual zarr-python stateful tests it gets turned on by default:

https://github.com/zarr-developers/zarr-python/blob/2fbd30aa0eb7b57a73d698b935cbd1e88c25faab/tests/test_store/test_stateful.py#L21-L25

src/zarr/testing/strategies.py
ianhi:
```suggestion

    This strategy depends on the global state of the config having rectilinear chunk grids enabled or not.
    This means that it may be a possible source of a hypothesis FlakyStrategy error due dependence
    on global state. However, in practice this seems unlikely to happen.
```

ianhi: @dcherian how is that?

maxrjones:
thanks for catching this before the release, @ianhi!

PR #3914, opened 2026-04-16
TEST: include `icechunk`s zarr tree strategies here

ianhi:
Up streaming our zarr tree hypothesis strategy from icechunk.

generates trees like these:

```
/
├── EtTWAYY/
│   ├── EtTWAYY.jEm  [array shape=(1,)]
│   ├── EtTWAYY/
│   │   ├── 7ek  [array shape=(1,)]
│   │   ├── g  [array shape=(1,)]
│   │   ├── 7H7_NaN  [array shape=(1,)]
│   │   └── NaN  [array shape=(1,)]
│   ├── NaN  [array shape=(1,)]
│   └── 91  [array shape=(1,)]
└── NaN/
    └── srtB  [array shape=(1,)]
```

Two commits 
- 398ecb48dac077b1b9b0506d146cf29f1b141f1f adds the models and strategies - nearly a straight lift from icehcunk. added a bit more logic around materialize write mode, and zarr format + case sensitivity otherwise the same as in icechunk.
- bb62db741981663c9d4b1e9e1c11e985aa9c9084 uses them in some tests - mostly as an example for anyone who wants to add more in the future. 

### DIFF
--- a/src/zarr/testing/models.py
+++ b/src/zarr/testing/models.py
@@ -1,0 +1,149 @@
+"""Models for comparison testing.
+
+The tree descriptors (GroupNode / ArrayNode) are pure data structures.
+Materialization writes it into any zarr store.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+from typing import TYPE_CHECKING, Any, Literal
+
+import numpy as np
+
+import zarr
+import zarr.abc.store
+import zarr.api.asynchronous
+from zarr.core.buffer import default_buffer_prototype
+from zarr.core.sync import sync
+
+if TYPE_CHECKING:
+    from collections.abc import Iterator
+
+_PROTOTYPE = default_buffer_prototype()
+
+
+@dataclass(frozen=True)
+class ArrayNode:
+    shape: tuple[int, ...]
+    dtype: np.dtype
+
+
+@dataclass(frozen=True)
+class GroupNode:
+    children: dict[str, ArrayNode | GroupNode] = field(default_factory=dict)
+
+    def walk(self, prefix: str = "") -> Iterator[tuple[str, Node]]:
+        """Yield ``(path, child)`` for every node, depth-first."""
+        for name, child in self.children.items():
+            p = f"{prefix}/{name}" if prefix else name
+            yield p, child
+            if isinstance(child, GroupNode):
+                yield from child.walk(p)
+
+    def nodes(self, prefix: str = "", *, include_root: bool = False) -> list[str]:
+        """Return paths of all nodes, optionally including root."""
+        root = [prefix] if include_root else []
+        return root + [p for p, _ in self.walk(prefix)]
+
+    def groups(self, prefix: str = "", *, include_root: bool = False) -> list[str]:
+        """Return paths of all group nodes, optionally including root."""
+        root = [prefix] if include_root else []
+        return root + [p for p, c in self.walk(prefix) if isinstance(c, GroupNode)]
+
+    def arrays(self, prefix: str = "") -> list[str]:
+        """Return paths of all array nodes."""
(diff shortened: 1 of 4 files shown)


src/zarr/testing/models.py
```diff
@@ -0,0 +1,149 @@
...
+class ArrayNode:
+    shape: tuple[int, ...]
+    dtype: np.dtype
+
+
+@dataclass(frozen=True)
+class GroupNode:
+    children: dict[str, ArrayNode | GroupNode] = field(default_factory=dict)
+
+    def walk(self, prefix: str = "") -> Iterator[tuple[str, Node]]:
```
d-v-b, line 36: i feel like "iter_nodes"or similar is a bit more expressive than "walk"

d-v-b:
Thanks for this! I feel like if these array / group hierarchy methods are useful for testing, then they are just useful, and we should consider integrating them into the codebase. With that in mind I'm gonna leave some comments but they are just advisory, feel free to ignore.

src/zarr/testing/models.py
```diff
@@ -0,0 +1,149 @@
...
+        """Yield ``(path, child)`` for every node, depth-first."""
+        for name, child in self.children.items():
+            p = f"{prefix}/{name}" if prefix else name
+            yield p, child
+            if isinstance(child, GroupNode):
+                yield from child.walk(p)
+
+    def nodes(self, prefix: str = "", *, include_root: bool = False) -> list[str]:
```
d-v-b, line 44: all these iteration routines can do IO. so returning `list[stuff]` requires doing all the IO before the first value is available. `Iterator[stuff]` or `Generator[stuff]` gives us a bit more flexibility if we want to stream results instead of doing all the IO up-front.

d-v-b: (this applies in a world where we put these routines on the group class that's bound to a store)

src/zarr/testing/models.py
```diff
@@ -0,0 +1,149 @@
...
+
+
+@dataclass(frozen=True)
+class ArrayNode:
+    shape: tuple[int, ...]
+    dtype: np.dtype
+
+
+@dataclass(frozen=True)
+class GroupNode:
```
d-v-b, line 33: this approach to "structurally model a group" reminds me of [`GroupSpec`](https://github.com/zarr-developers/pydantic-zarr/blob/13b1a2c2091fe06a4e7464cfa6b293b40c736c7a/src/pydantic_zarr/v3.py#L497) in pydantic-zarr. this is clearly a useful representation, so we should figure out how we can fit it into zarr-python proper

src/zarr/testing/models.py
```diff
@@ -0,0 +1,149 @@
...
+
+    @classmethod
+    def from_store(cls, store: zarr.abc.store.Store) -> GroupNode:
+        """Build a GroupNode by reading a zarr store's structure.
+
+        Example::
+
+            GroupNode.from_store(some_memory_store)
+        """
+        return sync(cls.from_store_async(store))
```
d-v-b, line 146: using `sync` here forces any consumer to use the zarr-python event loop. flagging this for visibility

tests/test_group.py
```diff
@@ -673,6 +677,38 @@ def test_group_child_iterators(store: Store, zarr_format: ZarrFormat, consolidat
     assert sorted(group.array_values(), key=lambda x: x.name) == expected_array_values
 
 
+@given(
+    tree=trees(),
+    consolidate=st.booleans(),
+    zarr_format=st.sampled_from([2, 3]),
+)
+@pytest.mark.filterwarnings("ignore:Consolidated metadata:zarr.errors.ZarrUserWarning")
+def test_groups_and_arrays_split_children(
```
d-v-b, line 686: was there bugged behavior that this test was necessary to fix?

PR #3917, opened 2026-04-20
test: add tests for version info in zarr

NITHIVARSHATP-07:
What does this PR do?
This pull request adds a new test file, test_version.py, to check that the __version__ attribute in Zarr is always present, is a string, and follows the expected format. This helps make sure version info is always available for bug reports and package management.

Why is this needed?
I noticed there weren’t direct tests for version info. Having these tests will help catch any accidental changes to version reporting and improve reliability.

How did I test this?
I ran the new tests with pytest and confirmed they all pass.

Checklist:
 Added new tests
 All tests pass locally

### DIFF
--- a/changes/3917.misc.md
+++ b/changes/3917.misc.md
@@ -1,0 +1,1 @@
+Add tests to verify that the Zarr version attribute (`__version__`) is always present, is a string, and follows the expected format.
--- a/tests/test_version.py
+++ b/tests/test_version.py
@@ -1,26 +1,25 @@
 """Tests for zarr._version module"""
-import pytest
-import importlib
-import sys
 
 
-def test_version_is_available():
+def test_version_is_available() -> None:
     """Test that __version__ is available and is a string."""
     from zarr import __version__
+
     assert __version__ is not None
     assert isinstance(__version__, str)
     assert len(__version__) > 0
 
 
-def test_version_format():
+def test_version_format() -> None:
     """Test that version follows basic format."""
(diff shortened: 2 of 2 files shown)


maxrjones:
Hi @NITHIVARSHATP-07, thank you for your effort in this PR. However, I think that this functionality is already tested in https://github.com/zarr-developers/zarr-python/blob/dd5a32196ab909807cb8595909da8ec095b84b88/tests/test_zarr.py#L16-L29, which checks that `__version__` appears in `print_debug_info()`, which implicitly covers "exists and is a non-empty string". Since the `_version.py` is generated by `hatch-vcs`, I think the format/fallback behavior is not necessary for zarr-python to test.

NITHIVARSHATP-07:
> Hi @NITHIVARSHATP-07, thank you for your effort in this PR. However, I think that this functionality is already tested in
> 
> https://github.com/zarr-developers/zarr-python/blob/dd5a32196ab909807cb8595909da8ec095b84b88/tests/test_zarr.py#L16-L29
> 
> , which checks that `__version__` appears in `print_debug_info()`, which implicitly covers "exists and is a non-empty string". Since the `_version.py` is generated by `hatch-vcs`, I think the format/fallback behavior is not necessary for zarr-python to test.

Thank you so much  for explaining how this is already covered in the existing tests! I really appreciate you taking the time to point me in the right direction. This was a great learning experience for me, especially getting familiar with the contribution process and the project’s CI.

PR #3919, opened 2026-04-21
feat:metadata package

d-v-b:
This PR creates a new subpackage called `zarr-metadata` just for JSON metadata. it's stored in `packages/zarr_metadata`. It contains typeddict classes that model the spec-defined JSON forms for v2 array + group metadata, and v3 array + group metadata, including data types, codecs, chunk key encodings and chunk grids. I only included type definitions for metadata that has an external spec. So zarr-python will need to define some types internally for e.g. unspecified data types or codecs.

I would like to publish this subpackage to pypi. These types useful to any python tool that works with zarr data, even if that tool doesn't use zarr-python. It is also useful to zarr-python, because it means we can remove and resolve some lingering questions about publishing types.

If we adopted the changes here, adding a new data type / codec / chunk grid, etc, would require adding types to `zarr-metadata`, then adding the implementation in zarr-python that work with those types. We wouldn't need to do these 2 operations in the same PR, but I expect that would be the normal practice.

This change does add complexity to our publishing workflow: we need to ensure that `zarr-metadata` changes are published at or before new `zarr-python` releases. We should add some checks to ensure that this happens.

Docs for the new package are missing from this PR. I would handle that in a follow-up.

I would appreciate feedback at all levels, including the following topics:
- are we aligned on the idea of a subpackage?
- are the types correct (do they match the spec documents)
- what needs to change in our CI?

closes #3355 and #3795


d-v-b:
if this gets merged, we would need the following changes in subsequent PRs:
- setting up CI / CD for `zarr-metadata`
- publishing `zarr-metadata` to pypi
- docs
- changes to zarr-python: 
  - depending on zarr-metadata 
  - removing internally-defined types 
  - deprecating re-exported public types
  - integrating zarr-python's docs with zarr-metadata's docs

d-v-b:
@zarr-developers/python-core-devs I'd love feedback on this.

maxrjones:
> @zarr-developers/python-core-devs I'd love feedback on this.

I really like the idea of having a public interface for metadata classes with minimal dependencies, and am fine with a sub-package as the mechanism for doing that.

The order of operations seems a little off to me. In particular, I'm concerned about drift over the course of the follow-up PRs, and us realizing that we're not happy with the design of zarr-metadata while migrating the core library. The following sequence would make more sense to me:

- Open a PR with a design document for public metadata classes
- Refactor the internal metadata classes to agree with the agreed upon API
- Move internal metadata classes to a subpackage
- Release subpackage
- Depend on subpackage internally

As an aside, I'd recommend reserving the package names that you're hoping to use on PyPI in anticipation of these changes

d-v-b:
> Open a PR with a design document for public metadata classes
> Refactor the internal metadata classes to agree with the agreed upon API

The purpose of this PR is not to define metadata classes that zarr-python uses. It is narrowly scoped to define typeddicts that model the _contents_ of zarr metadata documents. So this package doesn't affect anything about the shape of metadata-modelling classes in zarr-python, except by defining the type of their JSON form.

This package would only replace _typeddict / literal types_ that we currently define inside zarr-python. Allowing this package to own those types requires no behavioral changes from zarr-python.

That's why my timeline is roughly "make this package stand alone (exist on pypi), then make zarr-python depend on it"

d-v-b:
> n particular, I'm concerned about drift over the course of the follow-up PRs, and us realizing that we're not happy with the design of zarr-metadata while migrating the core library. 

What aspects of the code added here do you think might drift? The only changes I anticipate would be fixes for the following issues:
- failure to match the spec 
- failure to use python types ergonomically
- name changes / bikeshedding

All of these can be handled gracefully by downstream packages.

> As an aside, I'd recommend reserving the package names that you're hoping to use on PyPI in anticipation of these changes

good idea, this is done under my own pypi account. if this PR is approved I would transfer ownership to zarr-developers.

maxrjones:
> The purpose of this PR is not to define metadata classes that zarr-python uses.

It seemed from your earlier request for feedback that this subpackage was intended to eventually be useful for zarr-python. 

> What aspects of the code added here do you think might drift?

Most visibly, I'd like some more thought to be put into the names of the classes. It's inconsistent that some include Metadata and others don't.

d-v-b:
> It seemed from your earlier request for feedback that this subpackage was intended to eventually be useful for zarr-python.

The types and typeddicts in this package _are_ useful for zarr python. These types define the JSON structure of the array metadata, including all the chunk grids, codecs, data types, etc. We use this information to model the JSON metadata documents we read and write. 

So the impact of this package would basically be upstreaming many typeddict and literal types that zarr-python defines, e.g. [this one](https://github.com/zarr-developers/zarr-python/blob/029c376000c0cf3b9c311966ce8a1f4387112797/src/zarr/core/metadata/v3.py#L417).

> Most visibly, I'd like some more thought to be put into the names of the classes. It's inconsistent that some include Metadata and others don't.

The array metadata, group metadata, and consolidated metadata typddicts are named `XMetadata`. That's tracking how we talk about these data structures. I'd be fine with using the metadata suffix for other types, e.g. `BytesCodec -> BytesCodecMetadata`.

packages/zarr-metadata/pyproject.toml
```diff
@@ -0,0 +1,51 @@
+[build-system]
+requires = ["hatchling>=1.29.0"]
+build-backend = "hatchling.build"
+
+[project]
+name = "zarr-metadata"
+version = "0.1.0"
+description = "Spec-defined metadata types for Zarr v2 and v3."
+readme = "README.md"
+requires-python = ">=3.11"
```
maxrjones, line 10: will this cause any issues with zarr-python being 3.12+? Should they be consistent?

packages/zarr-metadata/src/zarr_metadata/v2/array.py
```diff
@@ -0,0 +1,59 @@
...
+
+    Spec-faithful: `datatype` is a numpy-style dtype string; `shape` is
+    present only when the field is a subarray field.
+
+    See https://zarr-specs.readthedocs.io/en/latest/v2/v2.0.html#data-type-encoding
+    """
+
+    fieldname: str
+    datatype: str
+    shape: NotRequired[tuple[int, ...]]
```
maxrjones: wouldn't the spec-faithful representation just have a list of lists? the V2 spec doesn't have JSON keys for fieldname or datatype

packages/zarr-metadata/src/zarr_metadata/v3/array.py
```diff
@@ -0,0 +1,52 @@
...
+    zarr_format: Literal[3]
+    node_type: Literal["array"]
+    data_type: MetadataField
+    shape: tuple[int, ...]
+    chunk_grid: MetadataField
+    chunk_key_encoding: MetadataField
+    fill_value: JSON
+    codecs: tuple[MetadataField, ...]
+    attributes: NotRequired[Mapping[str, JSON]]
+    storage_transformers: NotRequired[tuple[MetadataField, ...]]
```
maxrjones: there are more restrictions on data_type, chunk_grid, chunk_key_encoding, and codecs than implied by `MetadataField`. This would be more useful if it were more strictly typed. In addition, `MetadataField` allowing a string shorthand is wrong in the chunk grid case.

packages/zarr-metadata/src/zarr_metadata/v3/codec/crc32c.py
```diff
@@ -0,0 +1,36 @@
...
+
+class Crc32cCodec(TypedDict):
+    """`crc32c` codec metadata.
+
+    Per spec the codec has no configuration fields. `configuration` is
+    optional and, if present, should be an empty mapping.
+    """
+
+    name: Crc32cCodecName
+    configuration: NotRequired[dict[str, JSON]]
```
maxrjones: this typing allows erroneous metadata (e.g., `"configuration": {"foo": 1}}`)

packages/zarr-metadata/src/zarr_metadata/v3/codec/blosc.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+class BloscCodecConfiguration(TypedDict):
+    """Configuration for the Zarr v3 `blosc` codec."""
+
+    cname: BloscCName
+    clevel: int
+    shuffle: BloscShuffle
+    blocksize: int
+    typesize: int
```
maxrjones: typesize should be optional because it is not required with `noshuffle`

packages/zarr-metadata/src/zarr_metadata/v2/consolidated.py
```diff
@@ -0,0 +1,37 @@
...
+    """
+    `.zmetadata` file contents.
+
+    The `metadata` map uses flat path keys (`"foo/bar/.zarray"`,
+    `"foo/.zattrs"`, etc.) pointing to the JSON contents of the file at
+    that path. The keys include the filename suffix, not just the node path.
+    """
+
+    zarr_consolidated_format: int
+    metadata: Mapping[str, GroupMetadataV2 | ArrayMetadataV2 | JSON]
```
maxrjones: why allow arbitrary JSON as the values?

packages/zarr-metadata/src/zarr_metadata/v3/array.py
maxrjones: storage_transformers should be empty array until one is defined as an extension

packages/zarr-metadata/src/zarr_metadata/v3/codec/__init__.py
```diff
@@ -0,0 +1,19 @@
+"""
+Zarr codec metadata types.
+"""
+
+from collections.abc import Mapping
+
+Codec = str | Mapping[str, object]
+"""
+The widest JSON shape that can specify a codec (v2 or v3).
```
maxrjones: why have a v2 or v3 codec under the v3 namespace?

packages/zarr-metadata/src/zarr_metadata/v3/array.py
```diff
@@ -0,0 +1,52 @@
...
+
+from collections.abc import Mapping
+from typing import Literal, NotRequired
+
+from typing_extensions import TypedDict
+
+from zarr_metadata.common import JSON, NamedConfig
+
+
+class AllowedExtraField(TypedDict, extra_items=JSON):  # type: ignore[call-arg]
```
maxrjones: I'd slightly prefer `OptionalExtension` as the name here

packages/zarr-metadata/src/zarr_metadata/v3/data_type/int32.py
```diff
@@ -0,0 +1,23 @@
...
+"""Literal type of the `data_type` field for `int32`."""
+
+Int32FillValue = int
+"""Permitted JSON shape of the `fill_value` field for `int32`: a JSON integer in [-2**31, 2**31 - 1]."""
+
+
+__all__ = [
+    "INT32_DTYPE_NAME",
+    "Int32DTypeName",
+    "Int32FillValue",
```
maxrjones: do we really need three exports for each data type/codec? Could we start more minimal and add more if they are needed?

packages/zarr-metadata/src/zarr_metadata/v3/data_type/numpy_datetime64.py
```diff
@@ -0,0 +1,61 @@
...
+NUMPY_DATETIME64_DTYPE_NAME: Final = "numpy.datetime64"
+"""The `name` field value of the `numpy.datetime64` data type."""
+
+NumpyDatetime64DTypeName = Literal["numpy.datetime64"]
+"""Literal type of the `name` field of the `numpy.datetime64` data type."""
+
+DateTimeUnit = Literal[
+    "Y", "M", "W", "D", "h", "m", "s", "ms", "us", "μs", "ns", "ps", "fs", "as", "generic"
+]
+"""Time unit codes used by numpy.datetime64."""
```
maxrjones, line 20: this should be declared once and shared between this and timedelta

packages/zarr-metadata/src/zarr_metadata/common.py
```diff
@@ -0,0 +1,47 @@
...
+    Named-config envelope with required configuration.
+
+    Generic with two parameters: name literal and configuration mapping.
+
+    Uses the PEP 484 ``Generic[T]`` form rather than PEP 695 ``[T]`` syntax
+    so the package supports Python 3.11.
+    """
+
+    name: ReadOnly[TName]
+    configuration: ReadOnly[TConfig]
```
maxrjones: why include this?

maxrjones (commented):
I like the direction here. I think we should have a very high bar for accuracy if we're promoting these as the canonical representation for public use.

For naming (recognizing this is bike-shedding), I'd prefer consistent suffix conventions:
- Codecs all end in Codec
- DataTypes all end in DataType
- Chunk grids all end in ChunkGrid

I think this functionality would be best tested with some round-trip property tests

packages/zarr-metadata/pyproject.toml
d-v-b: this should be fine with zarr-python's more stringent requirement. zarr-metadata users on python 3.11 won't be able to use zarr-python, but that's OK. We want zarr-metadata to be as useful as possible, even to users who are not using zarr-python.

packages/zarr-metadata/src/zarr_metadata/v3/array.py
d-v-b: which restrictions are you thinking of? per the spec they can all be strings or `{name: <str>}` or `{name: <str>, config: <object>}`

packages/zarr-metadata/src/zarr_metadata/common.py
d-v-b: this is a type currently used in zarr-python, but we don't actually need it here. I will remove it.

packages/zarr-metadata/src/zarr_metadata/v2/array.py
d-v-b: good catch, this should just be `tuple[str, str] | tuple[str, str, tuple[int, ...]]`

packages/zarr-metadata/src/zarr_metadata/v3/array.py
d-v-b: why?

packages/zarr-metadata/src/zarr_metadata/v3/codec/__init__.py
d-v-b: that's a bug, and the `Codec` type is also wrong. this will be fixed.

packages/zarr-metadata/src/zarr_metadata/v3/array.py
maxrjones: I was thinking of strict type definitions for `<object>`

packages/zarr-metadata/src/zarr_metadata/v2/consolidated.py
d-v-b: this is a side-effect of not including the `attributes` field in the array v2 metadata dict, because in zarr v2 the attributes are stored in a separate file (`.zattrs`). But I actually think it's OK if the metadata typeddicts don't match exactly the contents of specific JSON documents, so I'm going to add `attributes` to ArrayMetadataV2 and remove `JSON` from here

packages/zarr-metadata/src/zarr_metadata/v3/codec/__init__.py
d-v-b: no reason, and this file has other issues that I'm going to fix

packages/zarr-metadata/src/zarr_metadata/v3/array.py
d-v-b: such as? we don't know anything more about the `configuration` field except that it's `Mapping[str, JSON]`

packages/zarr-metadata/src/zarr_metadata/v3/codec/crc32c.py
d-v-b: good point, I need to figure out how to provide a type annotation for an empty dict, since I think the spec technically allows just that

packages/zarr-metadata/src/zarr_metadata/v3/data_type/numpy_datetime64.py
d-v-b: We have a choice between repetition and introducing coupling between data types. My feeling in this case is that the repetition is manageable (we don't expect this code to change rapidly), and it gives us a model where each data type is independent. So I'm inclined to keep the repetition.

packages/zarr-metadata/src/zarr_metadata/v3/codec/blosc.py
d-v-b: fixed in https://github.com/zarr-developers/zarr-python/pull/3919/commits/a732fb22689c5b8319e2c6d8caabaf76446e9e57

packages/zarr-metadata/src/zarr_metadata/v3/codec/crc32c.py
d-v-b: fixed in https://github.com/zarr-developers/zarr-python/pull/3919/commits/82e10d65b66629fa13783945af0d17f9b0f5706f

packages/zarr-metadata/src/zarr_metadata/v3/codec/__init__.py
d-v-b: fixed in https://github.com/zarr-developers/zarr-python/pull/3919/commits/84d5ca177078012bcc61054d3fb8241a1a0d849d

d-v-b:
I restored `src/zarr` to its current state, to keep the changes minimal. this PR now just adds the new package. a subsequent PR will be needed to make `zarr` depend on it.

packages/zarr-metadata/src/zarr_metadata/v3/array.py
d-v-b: how about just `ExtraField`?

packages/zarr-metadata/src/zarr_metadata/v3/data_type/int32.py
d-v-b: IMO we should default to exporting all the types and constants declared in the spec. The point here is to be comprehensive. Consumers can always choose what they import.

packages/zarr-metadata/src/zarr_metadata/v3/array.py
d-v-b: maybe it helps to clarify that, because data types, codecs, etc are extensible, we don't want to restrict the set of allowed data types, codecs etc to the ones defined in the spec. All we should require is that these fields have the right shape.

packages/zarr-metadata/src/zarr_metadata/v3/array.py
maxrjones: That works

packages/zarr-metadata/src/zarr_metadata/v3/array.py
maxrjones: This is the same discussion as in https://github.com/zarr-developers/zarr-python/pull/3919#discussion_r3173383819, which boils down to whether we restrict types to metadata structures formalized in zarr-extensions or allow non-formalized metadata. Let's discuss further there

maxrjones (approved):
Let's try this out. Since it starts at v0.1.0, it seems very low risk

d-v-b:
> Let's try this out. Since it starts at v0.1.0, it seems very low risk

thanks max! I'm adding some tests, and then I will merge + publish to pypi. After that, we will need some PRs to wire this up to zarr-python

PR #3925, opened 2026-04-24
feat:get_ranges

d-v-b:
this PR adds a `get_ranges` protocol for stores. the protocol defines the shape of a function that fetches multiple byte ranges within the same stored object. The purpose is to define a method stores can opt into if they offer an efficient way to fetch multiple byte ranges from the same object, which would be immediately useful for the sharding codec. The protocol looks like this:

```python
async def get_ranges(
    key: str,
    byte_ranges: Iterable[ByteRequest | None],
    *,
    prototype: BufferPrototype,
) -> AsyncIterator[Sequence[tuple[int, Buffer | None]]]:
```

the return type is an async iterator over sequences, where each sequence is the result of an IO operation the store implemented. this provides some observability to the caller about the actual coalescing, if any, that occurred. Results are returned in computed order, so the inner result type is `tuple[int, Buffer | None]`, where the `int` is the index into the input `byte_ranges` for that result.

Only byte range requests that declare an explicit interval (`RangeByteRequest`) are coalesced. Any other byte range, or `None`, results in no coalescing and so the ranges will be fetched separately. I assume here that we do not care about coalescing overlapping suffix or prefix range requests, but we could add support for that if we need to.

In addition to this protocol, there's a freestanding function that takes: 
- a `f(byte range) -> Awaitable[Buffer]` function (which we would generate by combining `Store.get` with `functools.partial`)
- an iterable of byte range requests
- options (max gap bytes, max total bytes, etc)

this function contains basic byte range coalescing logic, and it can be re-used for multiple stores. This is a non-abstract-base-class alternative to a default implementation on an abc.

That freestanding function is used to implement `get_ranges` on the `FsspecStore`. This is probably not useful for local- or memory-backed storage, but is useful for remote storage. The actual implementation is lightweight:

```python
    async def get_ranges(
        self,
        key: str,
        byte_ranges: Iterable[ByteRequest | None],
        *,
        prototype: BufferPrototype,
    ) -> AsyncIterator[Sequence[tuple[int, Buffer | None]]]:
        """Read many byte ranges from ``key``, coalescing nearby ranges and fetching concurrently.

        See :class:`zarr.storage._protocols.SupportsGetRanges` for the contract and
        :func:`zarr.core._coalesce.coalesced_get` for the full semantics.
        """
        fetch = partial(self.get, key, prototype)
        async for group in coalesced_get(fetch, byte_ranges, options=self.coalesce_options):
            yield group
```

cc @aldenks, the idea here is to build a basis for your range coalescing work for the sharding codec
cc @kylebarron, would love your feedback on this design.

related issues/PRs:

#1758 
#3004


d-v-b:
for context, we do already have a `get_partial_values` method, but it returns eagerly (no streaming), and it handles fetching from multiple keys. `get_ranges` only targets a single key, and it supports streaming. This is IMO a more targeted and performance-friendly design, and it's narrowly what the sharding codec needs for IO against a single shard.

kylebarron:
I think making it iterable adds complexity and adds confusion to whether request coalescing is expected to be applied here or not.

If you have requests:

- 0-1000
- 1000-2000
- 2000-3000
- 100_000 - 100_005

Then of course we should coalesce all the first 3 requests into one. But then the async iterable implies that we might want to use one of the first responses before the last one arrives... but the last request would've arrived first.

Either, if you _really_ want the response type to be async iterable, the responses should probably have an `index` for which of the input requests it's associated to.

But I think it would be much simpler to take in a sequence of byte ranges and return a sequence of results. Just like [object-store](https://docs.rs/object_store/latest/object_store/trait.ObjectStore.html#method.get_ranges)/[obstore](https://developmentseed.org/obstore/latest/api/get/#obstore.get_ranges)/obspec.

d-v-b:
> If you have requests:
> 
>     * 0-1000
> 
>     * 1000-2000
> 
>     * 2000-3000
> 
>     * 100_000 - 100_005

In the design in this PR we are iterating over IO calls the reader actually did, which is less than or equal to the number of byte ranges requested. So, assuming the first three are fused, we would either see:
```
yields ((0, <payload>), (1, <payload>), (2, <payload>))
yields ((3, <payload>),)
```
or 
```
yields ((3, <payload>),)
yields ((0, <payload>), (1, <payload>), (2, <payload>))
```

Which, for sharding is useful -- you can start decoding chunks immediately while you wait for the rest of the sub-chunks to come in. Does that make sense? or am I misunderstanding something.

d-v-b:
my thinking is that the possibility of coalescing the individual range requests makes the response effectively paginated, and so we should model this in the response type.

kylebarron:
> Either, if you _really_ want the response type to be async iterable, the responses should probably have an `index` for which of the input requests it's associated to.

I didn't see that you've already implemented this.

Seems fine, if a little more complex to manage than sequence return type

d-v-b:
> Seems fine, if a little more complex to manage than sequence return type

totally agree, my reasoning for the complexity here is that this API is used inside a very performance-sensitive routine, and so we should pay the mild complexity cost for perf payoffs

src/zarr/core/_coalesce.py
```diff
@@ -0,0 +1,195 @@
...
+            merged_start = min(x[1].start for x in last)
+            prospective_end = max(last_end, r.end)
+            prospective_size = prospective_end - merged_start
+            if (
+                gap <= options["max_gap_bytes"]
+                and prospective_size <= options["max_coalesced_bytes"]
+            ):
+                last.append(pair)
+                continue
+        groups.append([pair])
```
ilan-gold, line 129:
```suggestion
    mergeable.sort(key=lambda pair: pair[1].start)
    iter_mergeable = iter(mergeable)
    groups: list[list[tuple[int, RangeByteRequest]]] = [[next(iter_mergeable)]]
    for pair in iter_mergeable:
        _i, r = pair
        last = groups[-1]
        last_end = max(x[1].end for x in last)
        gap = r.start - last_end
        merged_start = min(x[1].start for x in last)
        prospective_end = max(last_end, r.end)
        prospective_size = prospective_end - merged_start
        if (
            gap <= options["max_gap_bytes"]
            and prospective_size <= options["max_coalesced_bytes"]
        ):
            last.append(pair)
            continue
        groups.append([pair])
```

src/zarr/core/_coalesce.py
```diff
@@ -0,0 +1,195 @@
...
+
+    # Build a uniform list of work items. Each work item is a list of
+    # (input_index, ByteRequest | None) pairs. Merged groups have multiple
+    # members (all RangeByteRequest); uncoalescable items have a single member.
+    work_items: list[list[tuple[int, ByteRequest | None]]] = [
+        [(idx, r) for idx, r in g] for g in groups
+    ]
+    work_items.extend([(idx, single)] for idx, single in uncoalescable)
```
ilan-gold:
```suggestion
    work_items: list[list[tuple[int, ByteRequest | None]]] = groups
    work_items.extend([(idx, single)] for idx, single in uncoalescable)
```
if you want the type hints, or just ged rid of `work_items` and append to `groups`?

src/zarr/core/_coalesce.py
```diff
@@ -0,0 +1,195 @@
...
+                group_start = min(starts)
+                group_end = max(ends)
+                big = await fetch(RangeByteRequest(group_start, group_end))
+                if big is None:
+                    await completion_queue.put(("missing", None))
+                    return
+                ordered = sorted(members, key=lambda pair: pair[1].start)  # type: ignore[union-attr]
+                sliced: list[tuple[int, Buffer | None]] = []
+                for idx, r in ordered:
+                    sliced.append((idx, big[r.start - group_start : r.end - group_start]))  # type: ignore[union-attr]
```
ilan-gold:
```suggestion
                sliced: list[tuple[int, Buffer | None]] = [
                     (idx, big[r.start - group_start : r.end - group_start])
                    for idx, r in ordered
                ]
```
maybe for performance reasons if this is hit in a hot loop with lots of requests?

ilan-gold (commented):
Nice!

At an architecture level, it's not immediately clear to me how this will be exposed to users.  But maybe that is out of the scope of this PR? I think we don't want to turn this on by default without some rough estimates.

d-v-b:
@ilan-gold have another look, I cleaned things up here a bit. one notable change that I plan on re-using: define subfunctions that explicitly depend on a shared state, instead of relying implicitly on variables defined in the outer scope (like the queue and the semaphore)

chuckwondo (commented):
Is there any reason you chose not to implement a range coalescing function independent of fetching?  Perhaps I'm missing something, particularly since I'm still very much getting my feet wet with this codebase, but it seems to me that having an independent range coalescing function would allow for far simpler and easier testing.

As currently written, your tests must include fetching since you have coupled coalescing and fetching, which seems to have added quite a bit of verbosity/complexity to your tests that you might otherwise be able to avoid, since existing tests should already be covering the fetching.

d-v-b:
No good reason! I think factoring the range coalescing into a separate routine is a great idea for the reasons you mention. I'll implement that

chuckwondo:
I would also argue that `OffsetByteRequest`s _are_ coalescable.

Although you may not know how many total bytes are available, you certainly know, by definition, where the range starts (the offset), so as you coalesce earlier `RangeByteRequest`s, when you bump into your first `OffsetByteRequest`, and you still haven't hit your `max_coalesced_bytes` limit, you can pull as many bytes from the front of the `OffsetByteRequest` as you need to reach the limit, and shift the offset by the corresponding number of bytes.

You can also coalesce beyond that point.  For example, all `RangeByteRequest`s and `OffsetByteRequest`s starting after the first `OffsetByteRequest` are subsumed by that first `OffsetByteRequest`.

This, of course, would seem to add a bit of complexity to the coalescing logic, so I suspect you could exclude such logic from this PR, and separately consider whether or not it would be worth the additional complexity at a later date (or never).

d-v-b:
Agree, but the only consumer of this method is the sharding codec which exclusively requests explicit byte ranges. So if we implemented coalescing for other byte range requests, it would never be used.

chuckwondo:
I'm a bit confused then. Why do you accept all types of byte requests as input if you expect only byte range requests?

d-v-b:
The coalescing is a performance optimization, and we don't want to push branching on the request type to the caller. A non-coalesced request falls through to a regular fetch, which is submitted in the same batch as the coalesced fetches. The alternative requires two batches.

And we can't predict the requirements of future codecs. We know today what sharding needs, so we optimize for that. If a future caller needs to coalesce offset range requests, we can implement it inside this function without changing the signature.

chuckwondo:
That clarifies a bit for me, but I suspect I might need more time getting familiar with the library before that will be much more clear to me.

After looking at the PR code a bit more, I see that you group the range requests, but don't actually _collapse_ them (except when you actually fetch the bytes) so that once the bytes are fetched, you can slice them up according to the original range requests that you preserved.

src/zarr/core/_coalesce.py
```diff
@@ -0,0 +1,208 @@
...
+    max_coalesced_bytes: int
+    """Upper bound on the size of a single merged fetch (ignored for an already-oversized single request)."""
+    max_concurrency: int
+    """Maximum number of merged fetches in flight at once."""
+
+
+DEFAULT_COALESCE_OPTIONS: CoalesceOptions = {
+    "max_gap_bytes": 1 << 20,  # 1 MiB
+    "max_coalesced_bytes": 16 << 20,  # 16 MiB
+    "max_concurrency": 10,
```
ilan-gold: Why does this need its own maximal concurrency? How will this interact with the global setting?

src/zarr/core/_coalesce.py
```diff
@@ -0,0 +1,208 @@
...
+    ``members`` must already be sorted by ``start``; callers in this module
+    build it from the sorted mergeable list.
+    """
+    from zarr.abc.store import RangeByteRequest
+
+    try:
+        start = members[0][1].start
+        end = max(r.end for _, r in members)
+        async with ctx.semaphore:
+            big = await ctx.fetch(RangeByteRequest(start, end))
```
ilan-gold: Reuse `_fetch_single`?

src/zarr/storage/_protocols.py
```diff
@@ -0,0 +1,34 @@
...
+    Private / unstable. Shape may change before being made public.
+    """
+
+    def get_ranges(
+        self,
+        key: str,
+        byte_ranges: Iterable[ByteRequest | None],
+        *,
+        prototype: BufferPrototype,
+    ) -> AsyncIterator[Sequence[tuple[int, Buffer | None]]]:
```
chuckwondo: Given that this produces a sequence of tuples where the first element of each tuple is the index of the input byte range, I suggest typing `byte_ranges` as a `Sequence` as well, rather than an `Iterable`, otherwise it doesn't completely make sense to return indexes for something declared as a non-indexable type (i.e., `Iterable`).

src/zarr/core/_coalesce.py
d-v-b: this needs a max concurrency because implementations may launch concurrent range reads. Right now this setting doesn't interact at all with the global setting, and that's hard to support given our current concurrency limit design -- codecs like sharding call `concurrent_map` *recursively*, so the global concurrency limit is alrady on shaky ground for sharding. We should do some testing with this parameter to see if there are any pathologies possible with too much concurrency.

src/zarr/storage/_protocols.py
d-v-b: done in https://github.com/zarr-developers/zarr-python/pull/3925/commits/17f8d64d475b4eb9e4e1b4ff06ffb3faecc324d8

src/zarr/core/_coalesce.py
d-v-b: fetch_single doesn't do what we need here, unfortunately

ilan-gold: Ah I see both of these methods include putting the range on the queue.

src/zarr/core/_coalesce.py
```diff
@@ -0,0 +1,251 @@
...
+    ctx = _WorkerCtx(
+        fetch=fetch,
+        semaphore=asyncio.Semaphore(options["max_concurrency"]),
+        queue=asyncio.Queue(),
+    )
+
+    # Launch all work as tasks. The semaphore bounds actual I/O concurrency.
+    tasks: set[asyncio.Task[None]] = set()
+    for group in groups:
+        tasks.add(asyncio.create_task(_fetch_group(ctx, group)))
```
maxrjones:
```suggestion
        # A one-member "group" is a RangeByteRequest that did not merge with a
        # neighbor; route it through `_fetch_single` so it skips the redundant
        # slice-by-zero in `_fetch_group`.
        if len(group) == 1:
            idx, single = group[0]
            tasks.add(asyncio.create_task(_fetch_single(ctx, idx, single)))
        else:
            tasks.add(asyncio.create_task(_fetch_group(ctx, group)))```

File: src/zarr/storage/_fsspec.py [hunk 8]
```diff
@@ -124,11 +131,14 @@ def __init__(
         read_only: bool = False,
         path: str = "/",
         allowed_exceptions: tuple[type[Exception], ...] = ALLOWED_EXCEPTIONS,
+        *,
+        coalesce_options: CoalesceOptions = DEFAULT_COALESCE_OPTIONS,
```
maxrjones:
```suggestion
        coalesce_options: CoalesceOptions | None = None,
```

src/zarr/storage/_fsspec.py
```diff
@@ -124,11 +131,14 @@ def __init__(
...
     ) -> None:
         super().__init__(read_only=read_only)
         self.fs = fs
         self.path = path
         self.allowed_exceptions = allowed_exceptions
+        self.coalesce_options = coalesce_options
```
maxrjones:
```suggestion
        # Copy so per-instance mutation of `self.coalesce_options` does not
        # leak into the module-level `DEFAULT_COALESCE_OPTIONS` singleton
        # (or into other stores constructed without an explicit kwarg).
        self.coalesce_options: CoalesceOptions = (
            DEFAULT_COALESCE_OPTIONS.copy() if coalesce_options is None else coalesce_options.copy()
        )```

File: src/zarr/storage/_protocols.py [hunk 10]
```diff
@@ -0,0 +1,34 @@
...
+        """Read many byte ranges from `key`.
+
+        Each yield corresponds to one underlying I/O operation.
+
+        See `zarr.core._coalesce.coalesced_get` for full semantics.
+        """
+        ...
```
maxrjones: we should either put this in `zarr/abc/store.py` with the other protocols or more the other protocols here. It's confusing to have this separate from similar protocols.

src/zarr/storage/_protocols.py
```diff
@@ -0,0 +1,34 @@
...
+    from zarr.abc.store import ByteRequest
+    from zarr.core.buffer import Buffer, BufferPrototype
+
+
+@runtime_checkable
+class SupportsGetRanges(Protocol):
+    """Stores that satisfy this protocol can efficiently read many byte ranges
+    from a single key in a single call, typically via coalescing and concurrent fetch.
+
+    Private / unstable. Shape may change before being made public.
```
maxrjones: given that this is intended to be private/unstable, should the methods be prefixed with a `_` and the addition not included in the changelog as a feature release?

maxrjones (commented):
Nice, I think this is well-implemented and tested. I have one correctness issue and a few comments.

While I don't think this is the responsibility of this PR, I still want to register my general dissatisfaction with our API design that this PR inherits and extends, in particular with mixing ABC and protocol-based abstraction mechanisms. The issue here is that the new methods are not available in wrapper stores such as LoggingStore. Let's continue to work on a better architecture for the storage API.

src/zarr/storage/_fsspec.py
chuckwondo: `coalesce_options` should be immutable (or at least typed as such), so that any erroneous attempt to mutate them should fail (either during static analysis or at runtime), so this should be unnecessary.

To aid in this effort, I suggest considering a frozen dataclass rather than a TypedDict for CoalesceOptions.

d-v-b: > To aid in this effort, I suggest considering a frozen dataclass rather than a TypedDict for CoalesceOptions.

we can annotate the fields as `ReadOnly`

chuckwondo: Could do, but in this case, I would lean towards a kw-only, frozen dataclass for the following reasons:

1. no real motivation for using a dict here (we're not marshaling between JSON)
2. we can set defaults for every field, thus eliminating the need for a separate default instance (simply using `CoalesceOptions()` as the param default would do the trick)
3. allows specifying only what we want to override rather than having to specify everything (e.g., passing `CoalesceOptions(max_concurrency=20)` will use the default values for the other fields). Of course, we could mark the TypedDict as total=False, but that would require a bit more code to fill in the default values for the user (not much, but a bit less convenient than using a dataclass, IMO).

d-v-b:
> While I don't think this is the responsibility of this PR, I still want to register my general dissatisfaction with our API design that this PR inherits and extends, in particular with mixing ABC and protocol-based abstraction mechanisms. The issue here is that the new methods are not available in wrapper stores such as LoggingStore. Let's continue to work on a better architecture for the storage API.

I don't mind adding this method to the `Store` abc! I think I opted for the protocol initially because I assumed it would be easier to get a new protocol added than change the base `Store` class. But if we are all on board with changes to the base `Store` class, then IMO that's simpler.

src/zarr/core/_coalesce.py
```diff
@@ -0,0 +1,251 @@
...
+        return
+
+    ctx = _WorkerCtx(
+        fetch=fetch,
+        semaphore=asyncio.Semaphore(options["max_concurrency"]),
+        queue=asyncio.Queue(),
+    )
+
+    # Launch all work as tasks. The semaphore bounds actual I/O concurrency.
+    tasks: set[asyncio.Task[None]] = set()
```
chuckwondo: nitpick: Why is this a set? Seems a little odd to be adding tasks to a set since there cannot be "duplicate" tasks, as far as I know.

src/zarr/storage/_fsspec.py
chuckwondo: I've added a suggestion to the `CoalesceOptions` definition.

src/zarr/core/_coalesce.py
```diff
@@ -0,0 +1,251 @@
...
+
+    All fields required. See DEFAULT_COALESCE_OPTIONS for a sensible default.
+    """
+
+    max_gap_bytes: int
+    """Two RangeByteRequests separated by at most this many bytes may be merged into one fetch."""
+    max_coalesced_bytes: int
+    """Upper bound on the size of a single merged fetch (ignored for an already-oversized single request)."""
+    max_concurrency: int
+    """Maximum number of merged fetches in flight at once."""
```
chuckwondo:
```
@dataclass(frozen=True, kw_only=True)
class CoalescingOptions:
    """Knobs for coalescing contiguous byte ranges into fewer I/O requests."""
    max_gap_bytes: int = 1 << 20  # 1 MiB
    """Two RangeByteRequests separated by at most this many bytes may be merged into one fetch."""
    max_coalesced_bytes: int = 16 << 20  # 16 MiB
    """Upper bound on the size of a single merged fetch (ignored for an already-oversized single request)."""
    max_concurrency: int = 10
    """Maximum number of merged fetches in flight at once."""
```

src/zarr/core/_coalesce.py
```diff
@@ -0,0 +1,251 @@
...
+
+
+DEFAULT_COALESCE_OPTIONS: CoalesceOptions = {
+    "max_gap_bytes": 1 << 20,  # 1 MiB
+    "max_coalesced_bytes": 16 << 20,  # 16 MiB
+    "max_concurrency": 10,
```
chuckwondo:
```suggestion
```

src/zarr/core/_coalesce.py
```diff
@@ -0,0 +1,251 @@
...
+
+    return groups, uncoalescable
+
+
+async def coalesced_get(
+    fetch: Callable[[ByteRequest | None], Awaitable[Buffer | None]],
+    byte_ranges: Sequence[ByteRequest | None],
+    *,
+    options: CoalesceOptions,
+) -> AsyncGenerator[Sequence[tuple[int, Buffer | None]], None]:
```
chuckwondo: Consider simplification:

```suggestion
) -> AsyncIterator[Sequence[tuple[int, Buffer | None]]]:
```

src/zarr/core/_coalesce.py
```diff
@@ -0,0 +1,251 @@
...
+      (one fetch, one single-tuple yield per input).
+    - If any fetch returns `None` the iterator stops scheduling further fetches
+      and completes without yielding the missing group. Groups completed before
+      the miss remain observable.
+    - If a fetch raises, the exception propagates on the yield that produced the
+      failing group; earlier-completed groups remain observable.
+    """
+    groups, uncoalescable = coalesce_ranges(byte_ranges, options=options)
+    if not groups and not uncoalescable:
+        return
```
chuckwondo: The only way both `groups` and `uncoalescable` are empty is if `byte_ranges` is empty, so just check `byte_ranges` instead:

```suggestion
    if not byte_ranges:
        return

    groups, uncoalescable = coalesce_ranges(byte_ranges, options=options)
```

src/zarr/core/_coalesce.py
chuckwondo: @maxrjones, that's a bit of a micro-optimization, but if you want to make that logic change, I suggest pushing that into `_fetch_group` directly, where it will check the length of the list, and defer to `_fetch_single` for a singleton list, leaving the code here cleaner.

src/zarr/core/_coalesce.py
```diff
@@ -0,0 +1,251 @@
...
+
+    ctx = _WorkerCtx(
+        fetch=fetch,
+        semaphore=asyncio.Semaphore(options["max_concurrency"]),
+        queue=asyncio.Queue(),
```
chuckwondo: Why did you choose to use a queue? Perhaps I'm missing something, but as far as I can tell, this is perhaps adding unnecessary complexity.

src/zarr/core/_coalesce.py
```diff
@@ -0,0 +1,251 @@
...
+    total_work = len(tasks)
+
+    try:
+        pending_error: BaseException | None = None
+        for _ in range(total_work):
+            entry = await ctx.queue.get()
+            if entry[0] == "ok":
+                yield entry[1]
+                continue
+            # "missing" or "error": stop scheduling and cancel pending work.
```
chuckwondo: Why do you not raise an exception in the case of "missing"? Since this cancels all remaining work just like the "error" case does, unless I'm missing something, it seems that the caller will end up with incomplete results, but no error, which seems potentially confusing (and perhaps even error-prone for the user).

src/zarr/core/_coalesce.py
```diff
@@ -0,0 +1,251 @@
...
+            # Late arrivals that raced to enqueue before cancellation took
+            # effect sit in the completion queue and are discarded by the
+            # finally block (the queue is local and will be garbage-collected).
+            for t in tasks:
+                if not t.done():
+                    t.cancel()
+            if entry[0] == "error":
+                pending_error = entry[1]
```
chuckwondo: Why don't you simply raise the error here? This would eliminate the need for `pending_error`.

chuckwondo (changes requested):

src/zarr/core/_coalesce.py
d-v-b: good idea! I applied this in 5731be6b

src/zarr/core/_coalesce.py
```diff
@@ -0,0 +1,234 @@
...
+COALESCE_DEFAULT_MAX_GAP_BYTES: Final = 1 << 20
+COALESCE_DEFAULT_MAX_COALESCED_BYTES: Final = 16 << 20
+COALESCE_DEFAULT_MAX_CONCURRENCY: Final = 10
+
+
+def coalesce_ranges(
+    byte_ranges: Sequence[ByteRequest | None],
+    *,
+    max_gap_bytes: int | None = None,
+    max_coalesced_bytes: int | None = None,
```
chuckwondo: Why are you not specifying the defaults here? Doing so would make it obvious in and IDE what the defaults are, without having to go look for them.  It also allows you to simplify the function by eliminating the `is None` checks.

src/zarr/core/_coalesce.py
```diff
@@ -0,0 +1,234 @@
...
+    return groups, uncoalescable
+
+
+async def coalesced_get(
+    fetch: Callable[[ByteRequest | None], Awaitable[Buffer | None]],
+    byte_ranges: Sequence[ByteRequest | None],
+    *,
+    max_concurrency: int | None = None,
+    max_gap_bytes: int | None = None,
+    max_coalesced_bytes: int | None = None,
```
chuckwondo: Again, I suggest using the defaults.  I realize it may get a bit redundant to repeat the defaults everywhere, but doing so is more pleasant for the user experience.

src/zarr/core/_coalesce.py
```diff
@@ -0,0 +1,234 @@
...
+) -> Sequence[tuple[int, Buffer | None]]:
+    """Fetch one merged byte range and slice it back into per-input buffers.
+
+    `members` must already be sorted by `start`; callers in this module
+    build it from the sorted mergeable list. Raises `FileNotFoundError`
+    if the key is absent.
+    """
+    from zarr.abc.store import RangeByteRequest
+
+    start = members[0][1].start
```
chuckwondo: This is the logic I suggested moving out of `coalesced_get`.

```suggestion
    from zarr.abc.store import RangeByteRequest

    if len(members) == 1:
        solo_idx, solo_req = members[0]
        return await _fetch_single(ctx, solo_idx, solo_req)

    start = members[0][1].start
```

src/zarr/core/_coalesce.py
```diff
@@ -0,0 +1,234 @@
...
+    end = max(r.end for _, r in members)
+    async with ctx.semaphore:
+        big = await ctx.fetch(RangeByteRequest(start, end))
+    if big is None:
+        raise FileNotFoundError
+    sliced: list[tuple[int, Buffer | None]] = [
+        (idx, big[r.start - start : r.end - start]) for idx, r in members
+    ]
```
chuckwondo: No need for the cluttered, explicit annotation here.  Type checkers can infer this:

```suggestion
    sliced = [(idx, big[r.start - start : r.end - start]) for idx, r in members]
```

src/zarr/core/_coalesce.py
```diff
@@ -0,0 +1,234 @@
...
+            yield await fut
+    finally:
+        # Best-effort cancellation for any task still in flight. Covers the
+        # consumer-break path as well as exception propagation from `await fut`,
+        # where the remaining tasks must not be left running.
+        for t in tasks:
+            if not t.done():
+                t.cancel()
+        if tasks:
+            await asyncio.gather(*tasks, return_exceptions=True)
```
chuckwondo: I suggest simplifying as follows, which includes moving the singleton check into _fetch_group (as I mentioned in another comment). I have commented on the _fetch_group function with the corresponding change there.
The suggestion below also change the name of uncoalescable to singles (shorter and corresponds better as the non-groups values -- also easier to say 😅)

Using TaskGroup as a context manager automatically safely handles cancellation/completion for us, and also uses an ExceptionGroup to group all exceptions other than cancellation exceptions.  This still initiates cancellation as soon as a task fails, but if any tasks complete with failure in the interim (between the start of cancellation and final cleanup), they are captured nicely for us.

```suggestion
    groups, singles = coalesce_ranges(
        byte_ranges,
        max_gap_bytes=max_gap_bytes,
        max_coalesced_bytes=max_coalesced_bytes,
    )

    ctx = _WorkerCtx(fetch=fetch, semaphore=asyncio.Semaphore(max_concurrency))

    # Launch all work as tasks. The semaphore bounds actual I/O concurrency.

    async with asyncio.TaskGroup() as tg:
        tasks = [
            *(tg.create_task(_fetch_group(ctx, group)) for group in groups),
            *(tg.create_task(_fetch_single(ctx, i, single)) for i, single in singles),
        ]

        for fut in asyncio.as_completed(tasks):
            yield await fut
```

chuckwondo (changes requested):

d-v-b:
@maxrjones and @chuckwondo thanks for the great feedback, I rolled a lot of your points into a series of changes, summarized in this bulleted list:
- the protocol is gone and we have a new method on the store abc with a default implementation. get_ranges should work for all stores now
- the queue is gone, replaced with iteration over `as_completed`
- we abort as soon as any of the `get` requests returns `None`, and raise `FileNotFoundError`. We can't handle the results from an inconsistent read, so an exception makes sense here.
- The typeddict is gone, and we just use plain keyword arguments. `None` is a sentinel value that means "look up the default". When we set up a call site for `get_ranges` in the sharding codec operations, we will need to ensure that these parameters are passed in (e.g., from the per-array configuration).

...I see chunk has more suggestions, so the above list might already be stale xD

src/zarr/core/_coalesce.py
d-v-b: I think that's a good idea. In order to preserve the `None` sentinel behavior in an ergonomic way, we need to bring back a typeddict that models these kwargs so callers can construct such a dict and omit any keys that should take the default value.

chuckwondo: I don't think a typeddict is necessary. I'm fine with the individual kwargs, but I would prefer to eliminate the `None`s and defaulting to the default values explicitly.

The problem with a typeddict in this case is that it makes it _more_ awkward_ to deal with default values, or at least more verbose.

src/zarr/core/_coalesce.py
d-v-b: it's not a set any more!

src/zarr/core/_coalesce.py
d-v-b: the nones are gone, and the defaults are set in just the store `get_ranges` method. This is less convenient for callers of the internal routines, but IMO balances convenience for people calling the store method with avoiding redundant default parameters at different levels of detail. no need for typeddicts

src/zarr/abc/store.py
```diff
@@ -616,6 +617,59 @@ async def _get_many(
...
 
+    async def get_ranges(
+        self,
+        key: str,
+        byte_ranges: Sequence[ByteRequest | None],
+        *,
+        prototype: BufferPrototype,
+        max_concurrency: int = 10,
+        max_gap_bytes: int = 1 << 20,
+        max_coalesced_bytes: int = 16 << 20,
```
chuckwondo: Nice. I think this makes good sense. Perhaps just add some comments for clarity:

```suggestion
        max_concurrency: int = 10,
        max_gap_bytes: int = 1 << 20,  # 1 MiB
        max_coalesced_bytes: int = 16 << 20,  # 16 MiB
```

src/zarr/core/_coalesce.py
```diff
@@ -0,0 +1,222 @@
...
+
+
+async def coalesced_get(
+    fetch: Callable[[ByteRequest | None], Awaitable[Buffer | None]],
+    byte_ranges: Sequence[ByteRequest | None],
+    *,
+    max_concurrency: int,
+    max_gap_bytes: int,
+    max_coalesced_bytes: int,
+) -> AsyncIterator[Sequence[tuple[int, Buffer | None]]]:
```
chuckwondo: So we don't have to `cast` in order to invoke `aclose()`:

```suggestion
) -> AsyncGenerator[Sequence[tuple[int, Buffer | None]]]:
```

src/zarr/core/_coalesce.py
```diff
@@ -0,0 +1,222 @@
...
+            for fut in asyncio.as_completed(tasks):
+                yield await fut
+    except BaseExceptionGroup as eg:
+        # Unwrap: prefer GeneratorExit, then a single inner exception, otherwise raise group.
+        for exc in eg.exceptions:
+            if isinstance(exc, GeneratorExit):
+                raise exc from None
+        if len(eg.exceptions) == 1:
+            raise eg.exceptions[0] from None
+        raise
```
chuckwondo: We do NOT want to reraise `GeneratorExit`:

```suggestion
        # Unwrap: prefer a single inner exception, otherwise raise group.
        if subgroup := eg.subgroup(lambda e: not isinstance(e, GeneratorExit)):
            e = subgroup.exceptions[0] if len(subgroup.exceptions) == 1 else subgroup
            raise e from None
```

tests/test_coalesce.py
```diff
@@ -0,0 +1,671 @@
...
+    ranges: list[ByteRequest | None] = [RangeByteRequest(i * 1000, i * 1000 + 1) for i in range(6)]
+
+    agen = coalesced_get(fetch, ranges, **opts)
+    async for _group in agen:
+        break
+    # Explicitly close the generator so its finally block runs (cancelling
+    # in-flight tasks) before we make assertions. The narrow AsyncIterator
+    # return type does not expose `.aclose()`, but the runtime object is an
+    # async generator and supports it.
+    await cast("AsyncGenerator[Any, None]", agen).aclose()
```
chuckwondo:
```suggestion
    # Explicitly close the generator so its finally block runs (cancelling
    # in-flight tasks) before we make assertions.
    await agen.aclose()
```

tests/test_coalesce.py
```diff
@@ -0,0 +1,671 @@
...
+
+    assert cancelled_calls >= 1
+    assert completed_calls >= 1
```
chuckwondo:
```suggestion
    assert completed_calls >= 1
    assert cancelled_calls == len(ranges) - completed_calls
```

chuckwondo (changes requested):
We're getting very close!

src/zarr/abc/store.py
```diff
@@ -616,6 +617,59 @@ async def _get_many(
         for req in requests:
             yield (req[0], await self.get(*req))
 
+    async def get_ranges(
+        self,
+        key: str,
+        byte_ranges: Sequence[ByteRequest | None],
+        *,
+        prototype: BufferPrototype,
+        max_concurrency: int = 10,
```
ilan-gold, line 626: I was actually just thinking about this last night - I agree this is the way to go.

src/zarr/abc/store.py
```diff
@@ -616,6 +617,59 @@ async def _get_many(
...
+        max_gap_bytes: int = 1 << 20,
+        max_coalesced_bytes: int = 16 << 20,
+    ) -> AsyncIterator[Sequence[tuple[int, Buffer | None]]]:
```
ilan-gold, line 629: Just out of curiosity, why not return `AsyncIterator[tuple[int, Buffer | None]]`? Any reason to make the caller deal with iterating though the sub-iterable that is returned here? Maybe this is good for batched decompression?

d-v-b: see https://github.com/zarr-developers/zarr-python/pull/3925#issuecomment-4315968090

the idea is that N byte ranges are declared, M actual byte range requested are made, where N <= M. The output iterates over the M actual requests.

ilan-gold: I guess my point is that AFAICT (maybe I'm misunderstanding), the caller will still have to do:

```python
async for coalesced_sequence in store.get_ranges(...)
    for bytes in coalesced_sequence:
        out[subset] = decompress(bytes)
```
no?

> requested are made, where N <= M

Isn't this reversed, where N byte ranges are coalesced into `M<=N` byte requests?

d-v-b: haha yes, I flipped the inequality sign. the number of actual requests is less than or equal to the number of byte ranges 🤦

d-v-b: btw @kylebarron also found the `AsyncIterator[Sequence[TResult]]` pattern confusing / low value compared to `AsyncIterator[TResult]`. If yall think the marginal benefit of exposing the actual cadence of fetches to the caller is too marginal, I'm happy making the signature simpler (and thereby masking the order of actual fetches).

chuckwondo: I don't yet have enough insight to have a preference.  I'm fine with the additional complexity if you think it can help eke out better performance.

d-v-b: I am now pretty sure there's no difference in performance -- we always yield results as soon as they are available. the only difference is whether the caller is burdened / enlightened with the knowledge of which results came together in a batch.

src/zarr/core/_coalesce.py
```diff
@@ -0,0 +1,219 @@
...
+                *(tg.create_task(_fetch_single(ctx, i, single)) for i, single in singles),
+            ]
+
+            for fut in asyncio.as_completed(tasks):
+                yield await fut
+    except BaseExceptionGroup as eg:
+        # Unwrap: prefer a single inner exception, otherwise raise group.
+        if subgroup := eg.subgroup(lambda e: not isinstance(e, GeneratorExit)):
+            e = subgroup.exceptions[0] if len(subgroup.exceptions) == 1 else subgroup
+            raise e from None
```
chuckwondo: I'm not convinced that we should distinguish between single and multiple errors.  If the caller wants to handle errors, it becomes more convoluted to have to distinguish between a singular exception and an exception group. _Always_ raising an exception group is consistent, regardless of how many individual exceptions are in the group, allowing for simpler handling on the caller's end.

```suggestion
        # Filter out GeneratorExits, which should NOT be reraised.
        if subgroup := eg.subgroup(lambda e: not isinstance(e, GeneratorExit)):
            raise subgroup from None
```

src/zarr/abc/store.py
d-v-b: I don't feel confident predicting all future uses of this API, so I think I'm +.00001 on keeping the more expressive, albeit more nested, return type.

chuckwondo: Oh, I see. I thought you were also wanting to yield them in the same order as the input, which I thought Kyle was also indicating (i.e., not only yielding the same number of items as input, but also preserving order), which would seem to prevent opportunity for speed improvement by not allowing the results to be processed as they become available (which you previously mentioned as a concern).

In that case, I'm leaning the other way. That is, I'm not seeing the need for yielding the batches, which would seem to only add complexity on the calling end.  Yielding the individual ranges as they become available, rather than preserving the input ordering would seem simpler, while also allowing for performance.

However, I suppose we could keep the more complex function and provide another function (perhaps in a separate PR) that would yield the individual items by calling the more complex function that yields the batches, and would then do the simplification for the caller.

chuckwondo: Other than coming to a consensus on this point, I think everything else looks good to me, so from my perspective, this is the last point to resolve before approving this PR.

d-v-b: I would be fine taking an empirical approach -- if the baroque `get_ranges` signature turns out to be a burden, we can discover this through experience and mitigate it (e.g., with a convenience function). That seems easier than recovering performance or functionality from a prematurely simplified signature.

chuckwondo: That's agreeable to me. I'm approving.

chuckwondo (approved):
Great work @d-v-b! Thanks for the great review/refinement collaboration.

d-v-b:
@chuckwondo have a look at the latest changes. I'm new to exception groups!

src/zarr/core/_coalesce.py
```diff
@@ -0,0 +1,227 @@
...
+                yield await fut
+    except BaseExceptionGroup as eg:
+        # Strip GeneratorExits (consumer aclose()) and propagate whatever
+        # remains. `split` is used instead of `subgroup` because the latter
+        # short-circuits on the group object itself, returning the unchanged
+        # group when a predicate lambda happens to be true for the wrapper.
+        _, rest = eg.split(GeneratorExit)
+        if rest is None:
+            return  # only GeneratorExits — clean close
+        raise rest from None
```
chuckwondo: Not sure I follow the logic in your comment, but support for using a general function with `subgroup` was added in python 3.13, so changing to `split` makes sense (since we support earlier versions, apologies for that).

I might tweak your code slightly though, but take it or leave it:


```suggestion
        # Strip GeneratorExits (consumer aclose()) and propagate whatever remains.
        _, other_errors = eg.split(GeneratorExit)

        if other_errors is not None:
            raise other_errors from None
```

src/zarr/core/_coalesce.py
```diff
@@ -0,0 +1,227 @@
...
+    -----
+    Only `RangeByteRequest` inputs participate in coalescing. Two ranges
+    merge when both: their gap (next `start` minus current group's running
+    `end`) is `<= max_gap_bytes`, and the resulting merged span is
+    `<= max_coalesced_bytes`.
+    """
+    indexed: list[tuple[int, ByteRequest | None]] = list(enumerate(byte_ranges))
+    mergeable: list[tuple[int, RangeByteRequest]] = [
+        (i, r) for i, r in indexed if isinstance(r, RangeByteRequest)
+    ]
```
chuckwondo: We can eliminate a bit of noise, as these are properly inferred types:

```suggestion
    indexed = list(enumerate(byte_ranges))
    mergeable = [(i, r) for i, r in indexed if isinstance(r, RangeByteRequest)]
```

tests/test_coalesce.py
```diff
@@ -0,0 +1,674 @@
...
+# ---------------------------------------------------------------------------
+# Key-missing semantics.
+# ---------------------------------------------------------------------------
+
+
+async def test_key_missing_from_first_call_raises() -> None:
+    """If the very first fetch returns None, the iterator raises an ExceptionGroup containing FileNotFoundError."""
+    fetch = FakeFetch(b"x" * 100, key_exists=False)
+    ranges: list[ByteRequest | None] = [RangeByteRequest(0, 10), RangeByteRequest(20, 30)]
+    with pytest.RaisesGroup(pytest.RaisesExc(FileNotFoundError)):
```
chuckwondo, line 402: Nice!

chuckwondo (commented):
@d-v-b, per your request, I took another scan. Looks good. Only a couple of minor suggestions -- take 'em or leave 'em at your pleasure.

d-v-b:
> @d-v-b, per your request, I took another scan. Looks good. Only a couple of minor suggestions -- take 'em or leave 'em at your pleasure.

thanks, I took them :)

d-v-b:
@aldenks does everything here look good to you? we are waiting for another maintainer to sign off, and then this should ready for wiring up to the sharding codec

aldenks:
Awesome! Thank you everyone. I reviewed and this will drop in to #3004 easily. Default concurrency, max gap, and max coalesce values all seem good to me.

maxrjones (approved):
great work!

d-v-b:
thanks for help yall!

PR #3935, opened 2026-05-02
chore: cd workflow

d-v-b:
this adds a cd workflow for `zarr-metadata`

we either publish to pypi, in which case the version declared in the package is used, or we publish to testpypi, in which case a synthetic, unique version is created to ensure that we never have a conflict on testpypi.

this workflow only supports workflow dispatch right now. later PRs will add a tag-based release workflow.

TODO:


ilan-gold:
If I understand this right here, we can get rid of this "make a unique version script" since the version should be unqiuely identified by a commit and it doesn't really make sense to run this twice on the same commit

d-v-b:
@ilan-gold good point, that script is now gone

ilan-gold:
Ok so my comment [here](https://github.com/zarr-developers/zarr-python/pull/3949/changes#r3197987873) was right.  My above comment assumes you are using `hatch-vcs`.  Would you be interested in that @d-v-b? I could apply it to both this and the top-level package since we seem to be using hatch here.

ilan-gold:
I become so used to our own dev setup, I forget what doesn't happen by default!

d-v-b:
> Ok so my comment [here](https://github.com/zarr-developers/zarr-python/pull/3949/changes#r3197987873) was right.  My above comment assumes you are using `hatch-vcs`.  Would you be interested in that @d-v-b? I could apply it to both this and the top-level package since we seem to be using hatch here.

Yes, vcs versioning for zarr metadata would be great! I think we can define a regex that matches git tags like `zarr-metadata/<version>`

ilan-gold:
Ok will have a look in the morning!

ilan-gold (commented):
What immediately jumps out at me (I think claude mentioned this) is documenting this convention, probably in the `README.md` for this package.

packages/zarr-metadata/README.md
```diff
@@ -47,6 +47,26 @@ Zarr v2 and v3 specs, consolidated metadata, and a subset of the metadata
 defined in `zarr-extensions`. We are generally open to contributions that
 add types for Zarr metadata with a published spec.
 
+## Releasing
```
d-v-b, line 50: @ilan-gold how does this look

ilan-gold (approved):

d-v-b:
A change worth noting: to ensure that the github releases page only shows zarr-python releases, zarr-metadata is released by pushing tags. this is documented in the zarr-metadata readme.

Issue #3939, opened 2026-05-03
`memory://` URL scheme no longer routes to fsspec, breaking interop with `MemoryFileSystem`


maxrjones:
### Zarr version

v3.2.0

### Numcodecs version

v0.16.5

### Python Version

3.13

### Operating System

macOS (also reproduced on Linux in xarray's CI)

### Installation

via pixi (conda-forge `zarr=3.2.0`)

### Description

In zarr-python 3.2.0, `zarr.storage._common.make_store` special-cases the `memory://` URL scheme to construct a `ManagedMemoryStore` instead of falling through to `FsspecStore.from_url(...)`. This silently changes the meaning of `memory://...` URLs.

Previously, a `memory://...` string was treated as an fsspec URI and resolved against fsspec's process-wide `MemoryFileSystem`. Now it resolves to zarr's own in-process `ManagedMemoryStore`, which is a separate, isolated key/value store with no connection to fsspec.

As a result, the pattern of writing to fsspec's `MemoryFileSystem` via a mapper and then opening the resulting store via its `memory://...` URL is broken.

This is what `make_store` does in 3.2.0:

```python
elif isinstance(store_like, str) and parsed is not None:
    if parsed.scheme == "memory":
        # Create or get a ManagedMemoryStore
        return ManagedMemoryStore(name=parsed.name, path=parsed.path, read_only=_read_only)
```

The previous behaviour was that, when fsspec is installed, a `memory://...` URL passed as a string continues to be routed through fsspec (so writers and readers using the same URL share state). If `ManagedMemoryStore` is the new default, then either:

- it should share state with fsspec's `MemoryFileSystem`, or
- the breaking change should be documented in release notes with a deprecation/warning path, and ideally an opt-in scheme (e.g. `managed-memory://`) so existing fsspec-backed users aren't silently broken.

Concrete impact: xarray's `test_open_fsspec` fails against zarr 3.2.0 with `GroupNotFoundError: No group found in store 'memory://out2.zarr' at path ''`, even though the data is present in fsspec's `MemoryFileSystem`. Any downstream library doing the same write-via-mapper / read-via-URL pattern will silently read from an empty store.

### Steps to reproduce

```python
# /// script
# requires-python = ">=3.12"
# dependencies = [
#   "zarr@git+https://github.com/zarr-developers/zarr-python.git@main",
#   "fsspec",
# ]
# ///
#
# This script automatically imports the development branch of zarr to check for issues

import fsspec
import zarr

[...]

PR #3942, opened 2026-05-04
release : 3.2.1 release notes

d-v-b:
release notes for 3.2.1

TODO:

### DIFF
--- a/changes/3938.bugfix.md
+++ b/changes/3938.bugfix.md
@@ -1,4 +1,0 @@
-Fixed a `CastValue` validation bug where the "can we use an out-of-range mode" check
-inspected the source dtype instead of the target dtype. This meant arrays with a
-float source dtype and an integer target dtype incorrectly raised a `ValueError`
-when configured with a `wrap` out-of-range mode.
--- a/changes/3941.bugfix.md
+++ b/changes/3941.bugfix.md
@@ -1,6 +1,0 @@
-Fixed a bug where the codec pipeline evolved each codec against the original
-array spec instead of the spec produced by upstream array-to-array codecs. This
-caused failures whenever an upstream codec changed the dtype between codec
-boundaries — e.g. arrays using `CastValue` to convert a single-byte source dtype
-(`int8`) to a multi-byte target dtype (`int16`) raised a `ValueError` from
-`BytesCodec` about a missing `endian` configuration.
--- a/changes/3944.bugfix.md
+++ b/changes/3944.bugfix.md
@@ -1,4 +1,0 @@
-Fixed breakage in existing fsspec-dependent workflows caused by associating the "memory" URL scheme with
-instances of `ManagedMemoryStore` instead of fsspec's memory-backed store. After this change, store URLs with a "memory" scheme are handled differently when `fsspec` is installed:
-with `fsspec`, a `FsspecStore` backed by a `MemoryFileSystem` is used. Without `fsspec`,
-a `ManagedMemoryStore` is used.
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -2,6 +2,22 @@
 
 <!-- towncrier release notes start -->
 
+## 3.2.1 (2026-05-04)
+
+### Bugfixes
+
+- Fixed a `CastValue` validation bug where the "can we use an out-of-range mode" check
+  inspected the source dtype instead of the target dtype. This meant arrays with a
+  float source dtype and an integer target dtype incorrectly raised a `ValueError`
+  when configured with a `wrap` out-of-range mode. ([#3938](https://github.com/zarr-developers/zarr-python/issues/3938))
+- Fixed a bug where the codec pipeline evolved each codec against the original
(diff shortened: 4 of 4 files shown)


maxrjones:
@d-v-b would you have any time to fix the regression from https://github.com/zarr-developers/zarr-python/issues/3939 for 3.2.1? I could take a look tomorrow if you don't have time

d-v-b:
the two changes in this release are pretty simple bugfixes without which there is no mitigation of the underlying problem, whereas #3939 does have a mitigation and fixing the underlying issue in a satisfactory way will require some design work. so if possible i'd like to quickly release the two merged bugfixes and then get a fix for #3939 when we are happy with the design. how does that sound?

maxrjones:
> the two changes in this release are pretty simple bugfixes without which there is no mitigation of the underlying problem, whereas #3939 does have a mitigation and fixing the underlying issue in a satisfactory way will require some design work. so if possible i'd like to quickly release the two merged bugfixes and then get a fix for #3939 when we are happy with the design. how does that sound?

Couldn't we just revert the two problematic lines referenced in #3939 to resolve the regression and only release the automatic usage of ManagedMemoryStore when it's properly integrated after the more robust design effort?

d-v-b:
> > the two changes in this release are pretty simple bugfixes without which there is no mitigation of the underlying problem, whereas #3939 does have a mitigation and fixing the underlying issue in a satisfactory way will require some design work. so if possible i'd like to quickly release the two merged bugfixes and then get a fix for #3939 when we are happy with the design. how does that sound?
> 
> Couldn't we just revert the two problematic lines referenced in #3939 to resolve the regression and only release the automatic usage of ManagedMemoryStore when it's properly integrated after the more robust design effort?

I'm happy to implement a fix in a new PR but I don't see why that should delay releasing a bugfix version with the 2 fixes we already merged. IMO we should release quickly, as soon as fixes are done. So in this case we have 2 fixes merged, so we can release them now, and tomorrow I can open a PR that deals with the memory store thing, and we release _that_ quickly when it's done.

d-v-b:
Patching over the breakage reported in #3939 _is_ very simple. see #3944. If that PR is approved, I would probably have time to bring those changes in to this PR and we could include them with 3.2.0. later this evening.  But I think my broader point about not blocking releases because of other stuff we want to release still holds -- we should normalize releasing more, rather than less.

maxrjones (approved):

maxrjones:
> Patching over the breakage reported in #3939 _is_ very simple. see #3944. If that PR is approved, I would probably have time to bring those changes in to this PR and we could include them with 3.2.0. later this evening.  But I think my broader point about not blocking releases because of other stuff we want to release still holds -- we should normalize releasing more, rather than less.

I feel like I'm enabling frequent releases by quickly reviewing several PRs. I was asking if you have time to unblock some of my work by quickly bundling a PR that blocks downstream work by fixing a regression. It's fine if you don't have time and need a release immediately. I don't think it's productive to say I'm blocking releases. Specifically, asking to delay a release by a single day to reduce total work seems fine. It's just an ask, not a block.

d-v-b:
> I feel like I'm enabling frequent releases by quickly reviewing several PRs. I was asking if you have time to unblock some of my work by quickly bundling a PR that blocks downstream work by fixing a regression. It's fine if you don't have time and need a release immediately. I don't think it's productive to say I'm blocking releases. Specifically, asking to delay a release by a single day to reduce total work seems fine. It's just an ask, not a block.

I totally agree that it's not productive to talk about "blocking" releases, and I'm sorry for using that terminology, and more broadly for implying that you are doing anything other than being super helpful in this process! 

That being said, in principle I don't think we should delay a release for (completed) fix A while we wait for (pending) fix B, when A and B are independent. In this case, fix B is really simple so the maximum delay would not be very long, but as long as releases are cheap, I think our _default_ should be to address this situation by releasing each repaired feature as soon as the fixes are complete.

d-v-b:
as soon as #3944 is merged I will include it in this branch

maxrjones:
> > I feel like I'm enabling frequent releases by quickly reviewing several PRs. I was asking if you have time to unblock some of my work by quickly bundling a PR that blocks downstream work by fixing a regression. It's fine if you don't have time and need a release immediately. I don't think it's productive to say I'm blocking releases. Specifically, asking to delay a release by a single day to reduce total work seems fine. It's just an ask, not a block.
> 
> I totally agree that it's not productive to talk about "blocking" releases, and I'm sorry for using that terminology, and more broadly for implying that you are doing anything other than being super helpful in this process!
> 
> That being said, in principle I don't think we should delay a release for (completed) fix A while we wait for (pending) fix B, when A and B are independent. In this case, fix B is really simple so the maximum delay would not be very long, but as long as releases are cheap, I think our _default_ should be to address this situation by releasing each repaired feature as soon as the fixes are complete.

thanks for the clarification. That default sounds good to me

d-v-b:
and thank you for fixing the release notes!

PR #3944, opened 2026-05-04
fix: defer to fsspec for memory url scheme

### LINKED ISSUE #3939: `memory://` URL scheme no longer routes to fsspec, breaking interop with `MemoryFileSystem`

d-v-b:
Only associate the "memory" scheme with `ManagedMemoryStore` when fsspec is not installed.

closes #3939 

TODO:

### DIFF
--- a/changes/3944.bugfix.md
+++ b/changes/3944.bugfix.md
@@ -1,0 +1,4 @@
+Fixed breakage in existing fsspec-dependent workflows caused by associating the "memory" URL scheme with
+instances of `ManagedMemoryStore` instead of fsspec's memory-backed store. After this change, store URLs with a "memory" scheme are handled differently when `fsspec` is installed:
+with `fsspec`, a `FsspecStore` backed by a `MemoryFileSystem` is used. Without `fsspec`,
+a `ManagedMemoryStore` is used.
--- a/src/zarr/storage/_common.py
+++ b/src/zarr/storage/_common.py
@@ -369,7 +369,7 @@ async def make_store(
         return await LocalStore.open(root=store_like, mode=mode, read_only=_read_only)
 
     elif isinstance(store_like, str) and parsed is not None:
-        if parsed.scheme == "memory":
+        if parsed.scheme == "memory" and not _has_fsspec:
             # Create or get a ManagedMemoryStore
             return ManagedMemoryStore(name=parsed.name, path=parsed.path, read_only=_read_only)
         elif parsed.scheme == "file" or not parsed.scheme:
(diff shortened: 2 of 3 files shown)


maxrjones (approved):

d-v-b:
I added a test, when it passes I will merge this.

maxrjones:
Thanks for fixing this, @d-v-b!

PR #3947, opened 2026-05-05
Allow s3 testing in face of ambient AWS credentials

chuckwondo:
This PR allows tests that use s3fs to run successfully in environments where there are "ambient" AWS credentials (i.e., either `AWS_*` env vars set or presence of `~/.aws/config`) and the AWS region is set to something other than `"us-east-1"`.  In such cases, tests using s3fs fail with `IllegalLocationConstraintException`.

This PR allows tests to succeed in such environments (i.e., typically a local dev setup).  Further, since mypy was failing during pre-commit for code in the module `tests.test_store.test_fsspec`, it also adds that module to the mypy "ignore" list in `pyproject.toml` until mypy configuration can be addressed to avoid the problem (to be addressed in a separate PR that will make mypy execution more robust/consistent).  In the process of diagnosing the mypy error, I updated the mypy configuration to provide more informative output.

Finally, I adjusted a test fixture to eliminate the "noise" of the "AssertionError: Session was never entered" message (and traceback) upon completion of test execution, which is disconcerting and confusing (made me wonder how the tests all passed while showing a traceback).

TODO:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -365,7 +365,9 @@ ignore = [
 python_version = "3.12"
 ignore_missing_imports = true
 namespace_packages = false
-
+pretty = true
+show_error_code_links = true
+show_error_context = true
 strict = true
 warn_unreachable = true
 enable_error_code = ["ignore-without-code", "redundant-expr", "truthy-bool"]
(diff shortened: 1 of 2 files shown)


d-v-b (approved):

d-v-b:
thanks for this @chuckwondo!

PR #3949, opened 2026-05-06
chore(zarr-metadata): bump version to 0.1.1

d-v-b:
version 0.1.0 is burned on pypi in order to namesquat the package. so the first version we can publish will be 0.1.1.

TODO:

### DIFF
--- a/packages/zarr-metadata/pyproject.toml
+++ b/packages/zarr-metadata/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "zarr-metadata"
-version = "0.1.0"
+version = "0.1.1"
 description = "Spec-defined metadata types for Zarr v2 and v3."
 readme = "README.md"
 requires-python = ">=3.11"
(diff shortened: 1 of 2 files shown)


packages/zarr-metadata/pyproject.toml
ilan-gold, line 2: we use https://github.com/ofek/hatch-vcs with a decent amount of success! This might explain why I think dev commit hashes are suffixed by default on to versions, hence my comment in https://github.com/zarr-developers/zarr-python/pull/3935#issuecomment-4389801488.

ilan-gold (approved):

packages/zarr-metadata/pyproject.toml
d-v-b: We are also using vcs versioning but for zarr python. As this package is versioned independently, we will to add some logic to the git tag -> version string workflow to ensure that zarr python release tags are distinct from zarr metadata release tags

ilan-gold: Got it, I see that now.

PR #3956, opened 2026-05-08
[lazy indexing] transforms 

d-v-b:
This is a claude-assissted PR derived from #3906 that breaks out the index transformation functionality into its own module.

Ultimately we will model indexing an array as a mapping from the _source indices_ to a set of _output indices_ via an _index transformation_. The data structures here define the pieces of that model.

- `IndexDomain` is a model of a contiguous domain of n-dimensional array indices. This tracks the domain of an array that is being indexed. It's basically the set of valid coordinates for an indexable array, represented implicitly.
- `OutputIndexMap` is a union of types that model where, in an output (like a chunk), the indexing operation will send array elements.
- An `IndexTransform` is a mapping from an `IndexDomain` to a collection of `OutputIndexMap` elements. In a later PR, every Zarr array will be associated with an `IndexTransform`, which tracks how the array's indices are "projected" to stored indices.


d-v-b:
these data structures are not wired up to indexing yet, so we could safely merge this without changing any behavior.

d-v-b:
closing, as this is superseded by https://github.com/zarr-developers/zarr-python/pull/4196

PR #3961, opened 2026-05-10
refactor: bring in types from zarr-metadata

d-v-b:
replaces some of our types with exports from zarr-metadata. I expect a few related PRs, alternating between ones like this (importing types) and ones that add missing types to zarr-metadata.

TODO:


d-v-b:
cc @chuckwondo

pyproject.toml
```diff
@@ -38,6 +38,7 @@ dependencies = [
     'google-crc32c>=1.5',
     'typing_extensions>=4.13',
     'donfig>=0.8',
+    'zarr-metadata>=0.1.1',
```
ilan-gold: So development of these packages will proceed entirely independently and one is not developed against another? Or is there a secrete setting somewhere that enables this to be linked?

d-v-b: can you explain what you mean? `zarr-metadata` is on pypi, which is how we expect people who install `zarr-python` to get that package. `zarr-metadata` will be developed with the goal of primarily supporting `zarr-python`, and when new versions of `zarr-metadata` are released, we would update `zarr-python` to depend on them.

ilan-gold: My concern is that without an upper bound, something could be developed in `zarr-metadata` that breaks `zarr-python`. So maybe it makes sense to have a hatch job added to the test suite that tests `zarr-python` against what is in `zarr-metadata`?

d-v-b: yeah that's a good idea! this should run whenever we make changes to zarr-metadata. I'll add that to this PR.

d-v-b: i made a few changes in this area:
1. we put an upper bound on the supported version of zarr-metadata 
2. we ensure that the default `zarr-python` tests run on the package-local `zarr-metadata` (so we don't need an explicit upstream test)
3. we ensure that our min deps tests run on the pypi-installed `zarr-metadata` (so we know if we break compatibility with the oldest version on pypi). We use a pre-commit check to ensure that the version of `zarr-metadata` pinned in the min deps environment matches the lowest version `zarr-python` is compatible with (i.e., if we bump the minimum version of `zarr-metadata` in our main deps, we see a pre-commit failure if we don't also bump the version pinned in the min deps environment)

d-v-b:
we should not merge this until #3972 is sorted out. that PR switches our pre-commit mypy check to run in a locked environment, which can include the package-local copy of `zarr-metadata`. Without that PR, mypy in pre-commit would run against the version of `zarr-metadata` on pypi, which would generate false alarms when we pushed pre-release changes to `zarr-metadata` + `zarr-python`.

packages/zarr-metadata/src/zarr_metadata/v3/codec/bytes.py
```diff
@@ -29,18 +29,25 @@ class BytesCodecConfiguration(TypedDict):
...
+
+    `configuration` is itself optional — when no configuration fields are
+    set, the entire `configuration` key may be omitted. This matches the
+    bare-string short-hand form (`BytesCodecName`) at the canonical data
+    level; both encodings describe a `bytes` codec with default settings.
+    """
 
     name: BytesCodecName
-    configuration: BytesCodecConfiguration
+    configuration: NotRequired[BytesCodecConfiguration]
```
d-v-b, line 44: this fixed a bug in the type definition

packages/zarr-metadata/src/zarr_metadata/v3/codec/gzip.py
```diff
@@ -20,11 +20,15 @@ class GzipCodecConfiguration(TypedDict):
...
-    a default.
+    is slowest with the best compression ratio. The codec's compressed
+    output depends on `level`, so metadata that omits it cannot
+    reproducibly identify the chunk bytes produced by a writer — `level`
+    is required for the metadata to fulfill its reproducibility role,
+    even though the spec text does not mark it required with RFC 2119
+    keywords.
     """
 
-    level: NotRequired[int]
```
d-v-b: this fixed a bug in the type definition

src/zarr/core/group.py
```diff
@@ -441,6 +443,12 @@ def to_dict(self) -> dict[str, Any]:
         else:
             # Leave consolidated metadata unset if it's None
             result.pop("consolidated_metadata")
+        # `node_type` is a v3-only field. v2 group metadata (.zgroup) has
+        # only `zarr_format`; attributes live in a sibling .zattrs file.
+        # The dataclass carries `node_type` for in-memory use; strip it
+        # from the serialized v2 form.
+        if self.zarr_format == 2:
+            result.pop("node_type", None)
```
d-v-b, line 447: `node_type` is not a valid field in v2 group metadata, so arguably this was a bug. but we also don't use `to_dict` for creating our v2 metadata, so while correct, this change doesn't fix any broken behavior.

tests/test_array.py
```diff
@@ -326,47 +329,47 @@ def test_serializable_sync_array(store: LocalStore, zarr_format: ZarrFormat) ->
...
-            "zarr_format": zarr_format,
+        # Plain v2 array metadata; no v3-only fields (no codecs,
+        # storage_transformers, chunk_grid, etc.).
+        v2_metadata: ArrayMetadataV2 = {
+            "zarr_format": 2,
             "shape": (10,),
             "chunks": (1,),
             "dtype": "|u1",
             "dimension_separator": ".",
-            "codecs": (BytesCodec().to_dict(),),
```
d-v-b, line 353: this was a bug -- the metadata was supposed to be v2 metadata, for which the `codecs` field is invalid.

d-v-b:
this is ready for final review! 

key notes for reviewers:

- nothing should change at runtime
- some tests have changed because they used invalid metadata
- the `struct` dtype used lists instead of tuples for its JSON metadata as declared in zarr-python. That type is changed to use tuples, following our broader metadata convention. I do not think this will affect any consumers.
- There's a fair number of infrastructure / build changes, which try to ensure that we don't ship a version of zarr that depends on unreleased features in zarr-metadata.

.github/workflows/releases.yml
```diff
@@ -63,9 +63,119 @@ jobs:
           ls
           ls dist
 
+  # ---------------------------------------------------------------------------
```
ilan-gold, line 66: Could we add a test similar to https://github.com/zarr-developers/zarr-python/pull/3961/changes#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R267-R271 that runs the tests against the maximum version allowed from `pypi`? It feels like that would also catch any bugs present in that version.  If there is a test for the minimum `zarr_metadata` version, how could a user add a feature that hasn't been released yet and also have fully-passing CI?

src/zarr/core/metadata/v3.py
```diff
@@ -139,14 +140,12 @@ def parse_storage_transformers(data: object) -> tuple[dict[str, JSON], ...]:
     )
 
 
-class AllowedExtraField(TypedDict, extra_items=JSON):  # type: ignore[call-arg]
-    """
-    This class models allowed extra fields in array metadata.
-    They must have ``must_understand`` set to ``False``, and may contain
-    arbitrary additional JSON data.
-    """
+AllowedExtraField = ExtensionFieldV3
```
ilan-gold: Does it make sense to deprecate these sorts of aliases? A custom `__getattr__` that checks if the caller is trying to import one of these?

d-v-b: i think we need to patch the module object for this. lmk if you think that's high enough value in this case -- iirc these types were never exported at the top level

.github/workflows/releases.yml
d-v-b: after this PR, our test suite will run against the in-tree version of zarr-metadata. that's always going to be the latest version

src/zarr/core/metadata/v3.py
ilan-gold: > i think we need to patch the module object for this.

Yes, I think so, something like https://github.com/scverse/anndata/blob/b0746ac7db220a3428ab93d3d1d92c1c1b0122c8/src/anndata/_io/__init__.py#L8-L17

d-v-b:

Updated the branch to zarr-metadata **0.4.0** (merged current `main`, which contains the 0.4.0 release):

- Renamed the removed 0.3.0 document-type names to their 0.4.0 spellings across `src/` and `tests/` (`ArrayMetadataV2` → `ZarrV2ArrayMetadataJSON`, `GroupMetadataV3` → `ZarrV3GroupMetadataJSON`, `ZAttrsMetadata` → `ZarrV2ZAttrsJSON`, etc.).
- Bumped the published range to `zarr-metadata>=0.4.0,<0.5` and the `min_deps` floor pin to `zarr-metadata==0.4.0`.
- Added a `[tool.uv] conflicts` set so zarr-metadata's new standalone `docs` group (mkdocs-material 9.7.6) doesn't have to co-resolve with zarr's `dev`/`docs` groups (mkdocs-material 9.7.7) in the workspace lock.
- zarr-metadata CI workflow: kept the standalone-venv steps (the workspace's `requires-python >=3.12` blocks the package's 3.11 floor, so the justfile's `uv run` recipes can't be used here), adopted the `pyright==1.1.404` pin, and ported the new docs job to the same standalone pattern.
- Merge-resolution note: struct dtype `to_json` keeps this branch's tuple emission of `fields` while adopting main's removal of the unstable-dtype warning for the now-stable `struct` spec.

Verified locally: full test suite (7062 passed), zarr-metadata package tests (590 passed), `mypy src tests` clean.

PR #3962, opened 2026-05-10
fix(zarr-metadata): model stored metadata more closely

d-v-b:
Zarr V2 uses a separate JSON document named `.zattrs` for the attributes of an array or group.
This package was inconsistent about how it modelled this fact. The array metadata document type modelled
array fields (`shape`, `dtype`, etc), which would be stored in `.zarray`,  AND the `attributes` field,
which would be stored in `.zattrs`. Thus the array metadata model matched the representation
of an array that a program might use, rather than the stored layout. But the group metadata type didn't
follow this pattern -- it has no `attributes` field.

This PR addresses that inconsistency by adding an `attributes` field to `GroupMetadataV2`. That field is
not required. To model the stored representation of V2 data, this PR adds 3 new types: `ZArrayMetadata`,
`ZGroupMetadata`, and `ZAttrsMetadata`, that closely model the contents of the `.zarray`, `.zgroup`, and
`.zattrs` documents, respectively.

This change makes the V2 consolidated metadata type more accurate, as consolidated metadata for Zarr V2
is comprised of inlined metadata documents.


d-v-b:
cc @chuckwondo

packages/zarr-metadata/src/zarr_metadata/v2/array.py
```diff
@@ -62,7 +85,7 @@ class ArrayMetadataV2(TypedDict):
     order: ArrayOrderV2
     filters: tuple[CodecMetadataV2, ...] | None
     dimension_separator: NotRequired[ArrayDimensionSeparatorV2]
-    attributes: Mapping[str, object]
+    attributes: NotRequired[Mapping[str, object]]
```
ilan-gold, line 88:
```suggestion
    attributes: NotRequired[ZAttrsMetadata]
```
just for consistency

packages/zarr-metadata/src/zarr_metadata/v2/group.py
```diff
@@ -3,24 +3,46 @@
...
+    separate files; this type folds them so a single TypedDict represents
+    the complete in-memory state of a v2 group node. Consumers that read
+    or write the real on-disk files should use `ZGroupMetadata` (strict
+    `.zgroup`) plus `ZAttrsMetadata` directly.
 
     See https://zarr-specs.readthedocs.io/en/latest/v2/v2.0.html
     """
 
     zarr_format: Literal[2]
+    attributes: NotRequired[Mapping[str, object]]
```
ilan-gold, line 42:
```suggestion
    attributes: NotRequired[ZAttrsMetadata]
```

packages/zarr-metadata/tests/v2/consolidated/with_array_and_group.json
```diff
@@ -10,8 +11,8 @@
       "compressor": null,
       "fill_value": 0,
       "order": "C",
-      "filters": null,
-      "attributes": {}
-    }
+      "filters": null
+    },
+    "data/.zattrs": {"units": "meters"}
```
ilan-gold, line 16: Was there a bug here?

d-v-b: yes, `attributes` is not a field in `.zarray`

packages/zarr-metadata/src/zarr_metadata/v2/group.py
d-v-b: i'm not sure I see the value in obfuscating `Mapping[str, object]` here. IMO the model of the array metadata shouldn't depend on a serialization detail (the use of the `.zattrs` key) of one of its fields (`attributes`).

packages/zarr-metadata/tests/v2/consolidated/with_array_and_group.json
ilan-gold: IT also seems like something was not written that should have been?

ilan-gold: Ah I see this fixture is used just to check validity

ilan-gold (approved):

Issue #3457, opened 2025-09-15
deprecate / remove enums


d-v-b:
We use enums in a few places in the codebase, and I think in every instance a simple string literal is a better solution. We should remove or deprecate the use of `Enum` from the following modules:

- bytes.py: https://github.com/zarr-developers/zarr-python/blob/bce30dd3cc92a35038b88f80c4c42130670d4100/src/zarr/codecs/bytes.py#L21-L27
- indexing.py: https://github.com/zarr-developers/zarr-python/blob/bce30dd3cc92a35038b88f80c4c42130670d4100/src/zarr/core/indexing.py#L697C7-L720
- sharding.py: https://github.com/zarr-developers/zarr-python/blob/bce30dd3cc92a35038b88f80c4c42130670d4100/src/zarr/codecs/sharding.py#L68-L74
- blosc.py: https://github.com/zarr-developers/zarr-python/blob/bce30dd3cc92a35038b88f80c4c42130670d4100/src/zarr/codecs/blosc.py#L25-L56

PR #3963, opened 2026-05-10
deprecate blosc enums

d-v-b:
This PR deprecates the string enums used for the blosc codec. `Literal` strings are used instead. We only use enums in a few places -- Literal strings are much more abundant for representing constant string types -- and I think that's because enums in python are rather cumbersome compared to `Literal[<str>]`.

If this PR has legs, I would follow up with similar deprecations for the other enums (used in the bytes and sharding codecs).

~~closes~~ partially addresses https://github.com/zarr-developers/zarr-python/issues/3457


src/zarr/codecs/blosc.py
```diff
@@ -29,6 +30,8 @@
 CName = Literal["lz4", "lz4hc", "blosclz", "snappy", "zlib", "zstd"]
 """The codec identifiers used in the blosc codec """
 
+CNAME: Final = ("lz4", "lz4hc", "blosclz", "snappy", "zlib", "zstd")
```
ilan-gold: https://docs.python.org/3/library/typing.html#typing.get_args or?

src/zarr/codecs/blosc.py
```diff
@@ -148,12 +209,13 @@ class BloscCodec(BytesBytesCodec):
         The data type size in bytes. This affects how the shuffle filter processes
         the data. If None, defaults to 1 and the attribute is marked as tunable.
         Default: 1.
-    cname : BloscCname or {'lz4', 'lz4hc', 'blosclz', 'snappy', 'zlib', 'zstd'}, optional
-        The compression algorithm to use. Default: 'zstd'.
+    cname : {'lz4', 'lz4hc', 'blosclz', 'snappy', 'zlib', 'zstd'}, optional
```
ilan-gold: Why use a `set` here and not one for `cname`? Or just use the `Literal` type?

tests/test_codecs/test_blosc.py
```diff
@@ -135,3 +146,57 @@ def test_blosc_codec_sync_roundtrip() -> None:
...
+def test_blosc_shuffle_member_access_warns() -> None:
+    with pytest.warns(DeprecationWarning, match="BloscShuffle.shuffle"):
+        value = BloscShuffle.shuffle
+    assert value == "shuffle"
+
+
+def test_blosc_cname_member_access_warns() -> None:
+    with pytest.warns(DeprecationWarning, match="BloscCname.zstd"):
+        value = BloscCname.zstd
+    assert value == "zstd"
```
ilan-gold: Parametrization?

tests/test_codecs/test_blosc.py
```diff
@@ -135,3 +146,57 @@ def test_blosc_codec_sync_roundtrip() -> None:
...
+
+
+def test_blosc_enum_classes_import_silently() -> None:
+    with warnings.catch_warnings():
+        warnings.simplefilter("error")
+        from zarr.codecs.blosc import BloscCname as _BloscCname  # noqa: F401
+        from zarr.codecs.blosc import BloscShuffle as _BloscShuffle  # noqa: F401
+
```
ilan-gold, line 229: What is this test for?

tests/test_codecs/test_blosc.py
```diff
@@ -135,3 +146,57 @@ def test_blosc_codec_sync_roundtrip() -> None:
...
+def test_blosc_codec_rejects_unknown_cname() -> None:
+    with pytest.raises(ValueError, match="cname must be one of"):
+        BloscCodec(cname=cast(CName, "not-a-codec"))
+
+
+def test_blosc_codec_rejects_unknown_shuffle() -> None:
+    with pytest.raises(ValueError, match="shuffle must be one of"):
+        BloscCodec(shuffle=cast(Shuffle, "not-a-shuffle"))
+
```
ilan-gold:
```suggestion
@pytest.mark.parametrize("param", ["cname", "shuffle"])
def test_blosc_codec_rejects_unknown() -> None:
    with pytest.raises(ValueError, match=f"{param} must be one of"):
        BloscCodec(**{param: f"not-a-{param}"}) # type: ignore[arg-type]
```
modulo my type:ignore, which probably needs something different

src/zarr/codecs/blosc.py
```diff
@@ -56,38 +59,66 @@ class BloscJSON_V3(NamedRequiredConfig[Literal["blosc"], BloscConfigV3]):
     """
 
 
-class BloscShuffle(Enum):
+class _DeprecatedStrEnumMeta(type):
```
ilan-gold, line 62: https://docs.python.org/3/library/enum.html#enum.EnumType or?

Also why not just warn on import instead of warn on usage? Just warn when people even import the class.  It's not like it's used anywhere else.

ilan-gold (commented):
> closes https://github.com/zarr-developers/zarr-python/issues/3457

Don't think this Pr does that.

Not sure why we can't use `Literal` in the strings https://zarr--3963.org.readthedocs.build/en/3963/api/zarr/codecs/#zarr.codecs.BloscCodec for the docs.

Any reason we are not using https://github.com/tox-dev/sphinx-autodoc-typehints?

tests/test_codecs/test_blosc.py
d-v-b: testing that the classes can be imported without a warning (e.g., testing that the module that defines the classes doesn't also instantiate them).

src/zarr/codecs/blosc.py
d-v-b: `get_args` is only useful at runtime, it doesn't help you with type checking. for a small number of literal strings that won't change, writing them out explicitly as a final collection _does_ help with runtime and type checking.

src/zarr/codecs/blosc.py
d-v-b: Warning on use means fewer, more targeted warnings. I should see how many warnings we would get inside zarr-python if we warned on import.

d-v-b:
> Any reason we are not using https://github.com/tox-dev/sphinx-autodoc-typehints?

we aren't using sphinx!

tests/test_codecs/test_blosc.py
d-v-b: i added a docstring that explains it in https://github.com/zarr-developers/zarr-python/pull/3963/commits/9eebe01c8c2d2d442593a02cd75a13b92582c2d8

tests/test_codecs/test_blosc.py
d-v-b: done in https://github.com/zarr-developers/zarr-python/pull/3963/commits/9eebe01c8c2d2d442593a02cd75a13b92582c2d8

d-v-b:
> > closes #3457
> 
> Don't think this Pr does that.

good catch, I updated the PR summary to say "partially addresses"

src/zarr/codecs/blosc.py
d-v-b: Good call, I switched to the literal type itself

ilan-gold:
> we aren't using sphinx!

Oh right, mkdocs is completely separate - for a moment I was thinking it used sphinx as a backbone, but I guess not

src/zarr/codecs/blosc.py
ilan-gold: But this `CNAME` all-caps isn't used for anything except runtime checking unless I am missing something, no?

ilan-gold: Like, why even define it? `get_args(CName)` + `lru_cache` if there is a performance concern

d-v-b: it's 6 strings that are defined in a spec. given the choice between `get_args` + `lru_cache` (two imports, produces something untyped) and just binding the 6 strings to a constant, the static constant is far simpler (no imports, gets typed correctly).

Anyone who wants to iterate over the values defined in the literal type needs a way of making those values. We _should_ have been doing this in our tests, but we were not. I added some tests that exercise this.

tests/test_codecs/test_blosc.py
```diff
@@ -135,3 +155,134 @@ def test_blosc_codec_sync_roundtrip() -> None:
...
+    to the same value on the stored attribute. Adding a new value to the
+    BloscCnameLiteral type alias without also adding it to BLOSC_CNAME (or vice
+    versa) is caught here.
+    """
+    codec = BloscCodec(cname=cname)
+    assert codec.cname == cname
+
+
+@pytest.mark.parametrize("shuffle", BLOSC_SHUFFLE)
+def test_blosc_codec_accepts_all_shuffles(shuffle: BloscShuffleLiteral) -> None:
```
d-v-b, line 173: @ilan-gold this demonstrates the utility of a value that matches the literal values. making this type-safe with `get_args` requires casting boilerplate, because `get_args` returns `tuple[Any,...]`.

src/zarr/codecs/blosc.py
```diff
@@ -133,12 +192,14 @@ class BloscCodec(BytesBytesCodec):
         Always False for Blosc codec, as compression produces variable-sized output.
     typesize : int
         The data type size in bytes used for shuffle filtering.
-    cname : BloscCname
-        The compression algorithm being used (lz4, lz4hc, blosclz, snappy, zlib, or zstd).
+    cname : BloscCnameLiteral
```
d-v-b, line 195: Once we fully deprecate the enums, the `BloscCname` identifier will be available, and I would bind it to the literal type.

d-v-b:
contextual point I should have mentioned earlier: now that we have [code](https://github.com/zarr-developers/zarr-python/blob/main/packages/zarr-metadata/src/zarr_metadata/v3/codec/blosc.py) in `zarr-metadata` dedicated to defining the types + constants that are relevant to the codecs defined in the core spec, we should aim to eventually remove all definitions of those types in zarr-python itself. that will _first_ require the de-enumification we have here.

tests/test_codecs/test_blosc.py
ilan-gold: https://github.com/python/mypy/pull/17784 would be awesome to see this resolved but understood!

src/zarr/codecs/blosc.py
ilan-gold: So I guess the answer here is "too many?"

tests/test_codecs/test_blosc.py
d-v-b: yeah if `get_args` was typed properly we would totally use that

tests/test_codecs/test_blosc.py
```diff
@@ -135,3 +155,134 @@ def test_blosc_codec_sync_roundtrip() -> None:
...
+    """
+    Every shuffle mode in BLOSC_SHUFFLE is accepted by BloscCodec and round-trips
+    to the same value on the stored attribute. Adding a new value to the
+    BloscShuffleLiteral type alias without also adding it to BLOSC_SHUFFLE (or
+    vice versa) is caught here.
+    """
+    codec = BloscCodec(shuffle=shuffle)
+    assert codec.shuffle == shuffle
+
```
ilan-gold, line 182: I would say these should be parametrized and be one test

tests/test_codecs/test_blosc.py
```diff
@@ -135,3 +155,134 @@ def test_blosc_codec_sync_roundtrip() -> None:
...
+    JSON serialization (to_dict / from_dict) preserves every shuffle mode.
+    Guards against drift in the codec's V3 JSON form for any shuffle option.
+
+    The non-shuffle fields are fully specified so the codec has no tunable
+    attributes; tunability is not part of the JSON form and would otherwise
+    cause spurious round-trip mismatches.
+    """
+    codec = BloscCodec(typesize=1, cname="zstd", clevel=5, shuffle=shuffle, blocksize=0)
+    restored = BloscCodec.from_dict(codec.to_dict())
+    assert restored == codec
```
ilan-gold: Same, these are completely identical

ilan-gold (approved):
Not sure what the status of warning on importing is, but it seems like it would be a bit less code and not change the `BloscShuffle` type, for example, but if the warnings would be too aggressive because of how things are being imported, I understand.  I don't fully follow, though - what in the code base would import these values? Seems like they should be removed.

Other than that, looks good!

tests/test_codecs/test_blosc.py
d-v-b: agreed! and same for the next ones

maxrjones:
> > we aren't using sphinx!
> 
> Oh right, mkdocs is completely separate - for a moment I was thinking it used sphinx as a backbone, but I guess not

FYI mkdocstrings + griffe provides the same functionality as sphinx-autodoc-typehints. We still have the types manually included in the docstrings because many of the types are not public yet.  see "Run uv run python ci/check_unlinked_types.py" in https://github.com/zarr-developers/zarr-python/actions/runs/25682116308/job/75396132267

PR #3968, opened 2026-05-13
deprecate more enums

### LINKED ISSUE #3457: deprecate / remove enums

d-v-b:
continues the work in #3963 by applying the same deprecation to the enums used in the sharding and bytes codecs.

this will close #3457


src/zarr/codecs/_deprecated_enum.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+def _coerce_enum_input(value: object, param_name: str, codec_name: str) -> object:
+    """
+    If `value` is a real `enum.Enum` instance, emit a deprecation warning
+    naming `codec_name` and return `value.value`. Otherwise return `value`
+    unchanged. The third argument lets the warning text name the actual
+    codec (e.g. `BloscCodec`, `BytesCodec`, `ShardingCodec`).
+    """
+    if isinstance(value, Enum):
```
ilan-gold, line 51: Something I missed was that this line is not called by instantiating a class with a now-deprecated former `Enum` because nothing actually inherits from `Enum` anymore. What is this for then? Am I misreading this?

d-v-b: this is very defensive, to cover a situation where someone defined their own str enum outside of zarr, which would have worked before.

ilan-gold (approved):

PR #3972, opened 2026-05-14
chore: run mypy from developer environment

d-v-b:
On main, mypy gets its own python environment for pre-commit checks. This is problematic:
- it's a development burden because we need to update that environment every time we add a new optional dependency, or every time we raise the minimum version of a dependency. 
- It's host to some quirks: we have numpy pinned to a specific version simply because we do not have correct type annotations for the newer version of numpy, even through we support both versions.
- As we want `zarr` to depend on `zarr-metadata`, we need mypy to type-check against the repo-local version of zarr-metadata, not the version on pypi. this is tedious to set up in the isolated pre-commit environment.

for these reasons, this claude-authored PR removes the mypy-specific environment setup in pre-commit.yml, and instead uses `hatch run dev:mypy` in pre-commit to simply invoke mypy in the pre-existing developer environment. To ensure consistency, this PR defines the developer environment more precisely by specifying a python version and committing a lockfile to ci. 

because our dev environment does _not_ pin numpy to a specific version, this change reveals type-checking failures on the newer version of numpy installed in the dev environment. so this PR addresses those failures, often by simply ignoring them. These ignore statements are temporary -- I will revisit this in a follow-up PR that actually fixes the underlying type-checking issue.


d-v-b:
another point: because the `pre-commit.ci` CI job doesn't clone our repo locally, it can't use the new mypy setup. so mypy is being skipped in pre-commit ci (it happens in the lint workflow)

src/zarr/api/asynchronous.py
```diff
@@ -1033,6 +1033,11 @@ async def create(
         mode = "a"
     store_path = await make_store_path(store, path=path, mode=mode, storage_options=storage_options)
 
+    # Legacy v2 behavior: an unspecified dtype defaults to float64. Resolve it here so the
+    # value forwarded to `_create` is a real dtype rather than `None`.
+    if dtype is None:
+        dtype = "float64"
+
```
chuckwondo: nitpick/personal pref: I'm not a fan of reassignment (not that I never use it, but I avoid it when it's easy/sensible). In this case, I suggest simply inlining this in the call to `_create` (i.e., simply change the argument there to `dtype=dtype or "float64"`, and move the new comment above that line).

src/zarr/core/dtype/npy/int.py
```diff
@@ -600,7 +600,7 @@ def to_native_dtype(self) -> np.dtypes.Int16DType:
             The np.dtype('int16') instance.
         """
         byte_order = endianness_to_numpy_str(self.endianness)
-        return self.dtype_cls().newbyteorder(byte_order)
+        return self.dtype_cls().newbyteorder(byte_order)  # type: ignore[return-value]  # numpy 2.x stub: newbyteorder widens to base dtype, runtime preserves the concrete subclass
```
chuckwondo: nitpick: can we split off the numpy stub comment and move it to the line above to avoid excessively long lines (to be applied to all similar comments in this PR)?

chuckwondo (approved):
Fantastic @d-v-b!

I like that although you had to add a few "ignore" comments, you were also able to remove some existing ones, particularly since some of those existing ignores were head-scratchers.

I made a couple of nitpicky suggestions, but non-blocking, so feel free to ignore. Approving.

src/zarr/api/asynchronous.py
d-v-b: yes! another "dislikes reassigning" person! I also dislike reassignment. happy to apply this suggestion here.

d-v-b:
thanks for the comments @chuckwondo, I integrated that feedback in the latest changes. let me know if anything else needs work!

.github/workflows/lint.yml
```diff
@@ -22,4 +22,12 @@ jobs:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           persist-credentials: false
+      - name: Set up Python
+        uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
+        with:
+          python-version: "3.12"
+      - name: Install Hatch
+        uses: pypa/hatch@257e27e51a6a5616ed08a39a408a21c35c9931bc
```
chuckwondo: Version number in a comment?

chuckwondo (commented):
When I run `hatch run dev:mypy`, I get errors (same when running `hatch run dev:pre-commit run -a`, so at least we have consistency now):

```plain
tests/test_examples.py: note: In function "set_dep":
tests/test_examples.py:42: error: Argument 1 to "tuple" has incompatible type "Item | Container"; expected "Iterable[Any]"  [arg-type]
        for idx, dep in enumerate(tuple(config["dependencies"])):
                                        ^~~~~~~~~~~~~~~~~~~~~~
tests/test_examples.py:44: error: Unsupported target for indexed assignment ("Item | Container")  [index]
                config["dependencies"][idx] = dependency
                ^~~~~~~~~~~~~~~~~~~~~~~~~~~
tests/test_examples.py:44: note: See https://mypy.rtfd.io/en/stable/_refs.html#code-index for more info
tests/test_examples.py:44: error: Invalid index type "int" for "Container"; expected type "Key | str"  [index]
                config["dependencies"][idx] = dependency
                                       ^~~
tests/test_array.py: note: In function "test_from_array":
tests/test_array.py:1795: error: Statement is unreachable  [unreachable]
        assert result.chunks == new_chunks
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
tests/test_array.py:1795: note: See https://mypy.rtfd.io/en/stable/_refs.html#code-unreachable for more info
```

Oddly, VSCode does _not_ flag these, even after restarting.

d-v-b:
@chuckwondo the issue was dependency skew -- hatch doesn't use lockfiles (yet), so there was nothing keeping our two `dev` envs aligned. the recent changes use uv, which should keep us in sync via the lockfile.

chuckwondo:
> @chuckwondo the issue was dependency skew -- hatch doesn't use lockfiles (yet), so there was nothing keeping our two `dev` envs aligned. the recent changes use uv, which should keep us in sync via the lockfile.

Great! I get success now via pre-commit.

d-v-b:
it looks like lockfile support is coming to hatch soon: https://github.com/pypa/hatch/issues/2176. until then, using uv + uv.lock is our best bet

pyproject.toml
```diff
@@ -154,6 +153,7 @@ version.source = "vcs"
 hooks.vcs.version-file = "src/zarr/_version.py"
 
 [tool.hatch.envs.dev]
+python = "3.12"
```
maxrjones:
```suggestion
```

given that this is using uv, this change doesn't relate to the goal of running mypy from a developer environment and could be removed. If you want a specific version for uv, you could add a [.python-version file](https://docs.astral.sh/uv/concepts/python-versions/#python-version-files). otherwise uv will pick up based on the lockfile.

src/zarr/api/asynchronous.py
```diff
@@ -1049,7 +1049,8 @@ async def create(
         store_path,
         shape=shape,
         chunks=chunks,
-        dtype=dtype,
+        # Legacy v2 behavior: an unspecified dtype defaults to float64.
+        dtype=dtype or "float64",
```
maxrjones: This should at least have it's own changelog entry because on main,` dtype="", {}, 0, False, []` raised loudly while this PR silently produces float64. or use `"float64" if dtype is None else dtype` if you want to retain error raising on `"", {}, 0, False, []` but use `float64` as a default if `dtype=None`.

maxrjones (approved):
Thanks @d-v-b! I think the `dtype=dtype or "float64",` comment is blocking (did you mean for the default to apply to all Falsy values or just None?). The other is a nit. Both are minor enough that another review isn't needed.

src/zarr/api/asynchronous.py
d-v-b: good catch, that behavioral change is not intentional, and I will do this via reassignment to a temp variable

d-v-b: eh or an inline conditional statement

d-v-b: done in https://github.com/zarr-developers/zarr-python/pull/3972/commits/fff75cf821372907d335304293a6490a3f33709f

PR #3973, opened 2026-05-14
prune old string dtype class

d-v-b:
When we supported NumPy versions < 2, we had to implement the variable-length string dtype on top of the numpy "object" dtype. For NumPy versions > 2, we use the [`StringDType`](https://numpy.org/devdocs/user/basics.strings.html#variable-width-strings). 

Since we no longer support numpy < 2, we can remove the variable-length dtype class that used the "object" dtype. that's this PR.


tests/test_array.py
```diff
@@ -1981,23 +1980,14 @@ def test_array_repr(store: Store) -> None:
     assert str(arr) == f"<Array {store} shape={shape} dtype={dtype}>"
 
 
-class UnknownObjectDtype(UTF8Base[np.dtypes.ObjectDType]):
-    object_codec_id = "unknown"  # type: ignore[assignment]
-
-    def to_native_dtype(self) -> np.dtypes.ObjectDType:
-        """
-        Create a NumPy object dtype from this VariableLengthUTF8 ZDType.
+class UnknownObjectCodecDtype(VariableLengthUTF8):
```
d-v-b, line 1983: this test didn't actually depend on the codec using the "object" codec. it just depends on having an unknown object codec id

changes/3973.removal.md
```diff
@@ -0,0 +1 @@
+Removed the NumPy 1.x implementation of the `VariableLengthUTF8` data type.
```
maxrjones:
```suggestion
Removed the NumPy 1.x implementation of the `VariableLengthUTF8` data type because NumPy 1.x is no longer supported under [SPEC0](https://scientific-python.org/specs/spec-0000/).
```

maxrjones (approved):

PR #3978, opened 2026-05-15
misc(zarr-metadata): typed constants and fixes

d-v-b:
fixes inaccuracies in the `bytes` and `gzip` codec metadata types, and adds some typed constants

discovered while working on #3961


packages/zarr-metadata/tests/v3/codec/gzip/cases.json
```diff
@@ -2,10 +2,5 @@
   "with_level": {
     "name": "gzip",
     "configuration": {"level": 5}
-  },
-  "no_level": {
-    "name": "gzip",
-    "configuration": {}
-  },
```
ilan-gold, line 9: This data on disk is impossible, right? https://github.com/zarr-developers/zarr-python/blob/2e58a7a44c5388ebfa81100fb94f6c215e584dce/src/zarr/codecs/gzip.py#L49-L50 seems like it would always write a `configuration`.  Maybe link to this line in the comment about requirements by the spec to show that we have always written `conifguration` with a `level`?

d-v-b: that case is being removed because it's invalid. the gzip spec requires that level be set.

ilan-gold: Then what is https://github.com/zarr-developers/zarr-python/pull/3978/changes#diff-83847cf9f053159528992bc39f5333b74364994241ea45ec251c9996e3bdf782R23-R28 talking about?

d-v-b: >  Configuration for the Zarr v3 `gzip` codec.

    `level` is an integer in the range 0-9; 0 disables compression and 9
    is slowest with the best compression ratio. The spec does not mandate
    a default.
    is slowest with the best compression ratio. The codec's compressed
    output depends on `level`, so metadata that omits it cannot
    reproducibly identify the chunk bytes produced by a writer — `level`
    is required for the metadata to fulfill its reproducibility role,
    even though the spec text does not mark it required with RFC 2119
    keywords.

that's because the gzip spec doesn't explicitly mark the `level` field as required using the word MUST, but it is in fact required

d-v-b: this is the key change: https://github.com/zarr-developers/zarr-python/pull/3978/changes#diff-83847cf9f053159528992bc39f5333b74364994241ea45ec251c9996e3bdf782R31

previously `level` was marked `NotRequired`. now it's correctly marked as required.

ilan-gold (approved):

PR #3987, opened 2026-05-20
Control sharding codec read coalescing with ArrayConfig and runtime config options

aldenks:
Follow up #3004 by adding `ArrayConfig` and runtime configuration options for the thresholds that control how requests are coalesced when reading in the sharding codec.

Two new fields on `ArrayConfig` control how the sharding codec coalesces partial-shard reads: `sharding_coalesce_max_gap_bytes` (default 1 MiB) and `sharding_coalesce_max_bytes` (default 16 MiB). When reading multiple chunks from the same shard, nearby byte ranges are merged into a single request to the store if separated by no more than `sharding_coalesce_max_gap_bytes` and the merged read stays within `sharding_coalesce_max_bytes`. Defaults are seeded from the matching `array.sharding_coalesce_max_gap_bytes` / `array.sharding_coalesce_max_bytes` keys in `zarr.config` at array-creation time, and can be overridden per array by passing `config={...}` to `zarr.create_array`.

TODO:


d-v-b:
disclaimer: I'm not a big fan of our global config object, so I'd like to explore some alternative ways for the sharding reads to access this configuration.

A few options:
- new attributes on the sharding codec
  I'm not a big fan of this, because declaring the sharding codec explicitly from `create_array` is tedious, and also because we want to move away from the codecs knowing too much about IO operations.
- new fields on [`ArrayConfig` ](https://github.com/zarr-developers/zarr-python/blob/79d5b8fde084b66bfe3c3c2a70fe8a013eb9c55b/src/zarr/core/array_spec.py#L35)
  probably the best option. we still use the global config for setting the defaults, but the sharding codec gets these parameters from the array config object, which is tied to the array, not a mutable global.

aldenks:
@d-v-b I also like new fields on ArrayConfig. Thinking that through:
- This allows you to set coalesce options differently per array
- If you only use global configs, whatever global setting you have _at the time of array open_ is what is used for the life of that array.
- The global config field becomes something like `array.sharding_coalesce_max_gap_bytes` and `array.sharding_coalesce_max_bytes` matching the ArrayConfig convention of pulling from a singly nested field under `array.`.
- For the time being, if you're interacting with zarr via xarray then you still can only set these via the global config but that's a pre-existing inability to specify the ArrayConfig via when opening via xarray.

That sound alright?

d-v-b:
> @d-v-b I also like new fields on ArrayConfig. Thinking that through:
> 
>     * This allows you to set coalesce options differently per array
> 
>     * If you only use global configs, whatever global setting you have _at the time of array open_ is what is used for the life of that array.
> 
>     * The global config field becomes something like `array.sharding_coalesce_max_gap_bytes` and `array.sharding_coalesce_max_bytes` matching the ArrayConfig convention of pulling from a singly nested field under `array.`.
> 
>     * For the time being, if you're interacting with zarr via xarray then you still can only set these via the global config but that's a pre-existing inability to specify the ArrayConfig via when opening via xarray.
> 
> 
> That sound alright?

yeah, that sounds right. the array config object is designed to make it easy to get a cheap copy of an array with a new config, using the `with_config` method. Unfortunately, xarray makes it very hard to use this, because xarray doesn't give access to the base zarr array. So until xarray adds a zarr array-config-aware API, the global config is the only knob xarray users have, without re-creating the dataarray entirely.

aldenks:
`ArrayConfig` options implemented, pulling from the global config if not set explicitly. This is ready for review!

src/zarr/core/array_spec.py
```diff
@@ -45,22 +48,42 @@ class ArrayConfig:
...
+    sharding_coalesce_max_bytes: int
 
     def __init__(
-        self, order: MemoryOrder, write_empty_chunks: bool, *, read_missing_chunks: bool = True
+        self,
+        order: MemoryOrder,
+        write_empty_chunks: bool,
+        *,
+        read_missing_chunks: bool = True,
+        sharding_coalesce_max_gap_bytes: int = 1 << 20,  # 1 MiB
```
d-v-b, line 71: flagging this as an old problem we need to deal with, not a blocker: the default values appear twice, once here, and once in the global config object. that's a latent bug. the defaults should be defined in exactly 1 place.

d-v-b (approved):

src/zarr/core/array_spec.py
aldenks: +1 For now I added `test_array_config_init_defaults_match_global_config` which makes sure they don't drift.

aldenks:
@d-v-b ready for merge! thank you for the review

d-v-b:
thanks for this @aldenks!

aldenks:
Woohoo! may sharded reads be fast forever

PR #3988, opened 2026-05-20
Do not install CuPy on darwin

chuckwondo:
CUDA isn't available on macOS, so don't attempt to install it. This allows running "uv sync --all-extras" without error on macOS.


d-v-b (approved):

d-v-b:
thanks chuck!

Issue #3993, opened 2026-05-20
zarr-metadata tags break zarr tag-based versioning


d-v-b:
zarr-metadata releases are associated with a special git tag that starts with `zarr-metadata-<version>`. When generating a runtime version string from a git tag, `zarr` reads the `zarr-metadata` tag and reports that as its version, leading to `zarr` reporting it's current version as `0.2.0.dev...`. 

To fix this, we need `zarr` to exclude `zarr-metadata-<version>` tags in its hatch-vcs configuration.

d-v-b:
see https://github.com/zarr-developers/zarr-python/actions/runs/26187664671/job/77047478983?pr=3992

d-v-b:
closed by #3994

PR #3994, opened 2026-05-20
fix: match v* when extracting a zarr version from git tags

d-v-b:
This is high priority, as failing to apply this fix before our next release will ship a zarr that pretends to be much younger than it is.

TODO:

### DIFF
--- a/tests/test_version_derivation.py
+++ b/tests/test_version_derivation.py
@@ -1,162 +1,34 @@
-"""Tests for the git-tag-based version derivation configured in `pyproject.toml`.
+"""Sanity check that ``zarr.__version__`` looks like a v3-or-newer release.
 
-zarr-python derives its version from git tags via `hatch-vcs`, which runs a
-`git describe` command configured under `[tool.hatch] version.raw-options`.
-Because this repository is a monorepo that also releases the `zarr-metadata`
-subpackage using `zarr_metadata-v*` tags, the describe command must be
-restricted to zarr-python's own `v*` tags. Otherwise a newer
-`zarr_metadata-v*` tag hijacks the derived version (e.g. zarr reports
-`0.2.x` instead of `3.x`), which silently breaks downstream version gates.
+Background: zarr-python derives its version from ``git describe`` via
+hatch-vcs. The repo also publishes a separate ``zarr-metadata`` subpackage
+that uses ``zarr_metadata-v*`` tags. Without the ``--match v*`` filter in
+``[tool.hatch] version.raw-options.git_describe_command``, ``git describe``
+walks back to those subpackage tags and reports a version like ``0.2.0`` for
+a from-source build of zarr-python itself — see
+https://github.com/zarr-developers/zarr-python/pull/3994.
 
-These tests pin that behavior by running the *actual configured command*
-against synthetic repositories, so they would catch a regression in the
-config without depending on the ambient repository's tag state.
+This test catches that class of regression: anything that makes zarr-python
+report a version lower than the v3 release line. When 4.0 is released,
+bump the floor; that's a deliberate, one-line edit at a planned boundary.
 """
 
 from __future__ import annotations
 
-import shlex
-import subprocess
-import tomllib
-from pathlib import Path
-from typing import TYPE_CHECKING
+from packaging.version import Version
 
-if TYPE_CHECKING:
-    from collections.abc import Sequence
-
-REPO_ROOT = Path(__file__).parent.parent
+import zarr
 
 
-def _configured_describe_command() -> list[str]:
-    """The `git describe` command zarr-python configures for version derivation.
-
-    Read from `pyproject.toml` so this test exercises the real config rather
-    than a hand-retyped copy that could drift from it.
-    """
-    pyproject = tomllib.loads((REPO_ROOT / "pyproject.toml").read_text())
-    raw_options = pyproject["tool"]["hatch"]["version"]["raw-options"]
-    command = raw_options["git_describe_command"]
-    if isinstance(command, str):
-        return shlex.split(command)
-    # setuptools_scm also accepts a pre-split list
-    return list(command)
-
-
-def _git(repo: Path, *args: str) -> str:
-    result = subprocess.run(
-        ["git", "-C", str(repo), *args],
-        check=True,
-        capture_output=True,
-        text=True,
+def test_version_is_v3_or_newer() -> None:
+    # Use packaging.Version so we transparently handle hatch-vcs dev suffixes
+    # like "3.2.2.dev30+gdc5e1825" that appear on any source build past the
+    # latest v* tag — Version.major returns 3 for that string.
+    parsed = Version(zarr.__version__)
+    assert parsed.major >= 3, (
+        f"zarr.__version__={zarr.__version__!r} is not on the v3 (or newer) "
+        f"release line. If this fires on a from-source build, check that "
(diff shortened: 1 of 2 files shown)


d-v-b:
closes #3993

maxrjones:
Is there a way to test this in the CI?

maxrjones:
> Is there a way to test this in the CI?

Seems like we could use the test structure that numcodecs uses in https://github.com/zarr-developers/zarr-python/actions/runs/26187664671/job/77047478983?pr=3992. Could you please add a test @d-v-b?

d-v-b:
The problem here is that zarr derives its version from the wrong git tag at runtime. That numcodecs test job fails because no numcodecs tests are collected, because the entire test file is skipped on zarr versions < 3.1.3: https://github.com/zarr-developers/numcodecs/blob/fa92ee7ec031479e48098f57604fed313cc72986/tests/test_zarr3.py#L10

I don't think we want to codify this behavior -- failing if zarr's runtime version is < "3.1.3" -- in a test.

d-v-b:
an automated test would be valuable if we were planning on making a lot of changes to our git tag conventions. but I don't think we are, so I'm not really sure a test makes sense here, vs manually checking that the zarr version is correctly inferred on this branch.

d-v-b:
`uv run python -c "import zarr; print(zarr.__version__)"`

main: 

```
0.2.0
```

this branch:
```
3.2.2.dev30+gdc5e1825
```

maxrjones (approved):
I generally think we should avoid minimalism in testing and manual testing whenever possible. I think it's not good that we only caught this error because we were running numcodec's tests in our CI. That suggests to me that our version testing infrastructure is insufficient. Testing for some reasonable version structure like a regex similar `v3.*.*` seems like a small addition with few downsides that would prevent future issues and rarely need changing. But I'm not going to block merging on that.

d-v-b:
> Testing for some reasonable version structure like a regex similar v3.*.* seems like a small addition with few downsides that would prevent future issues and rarely need changing. But I'm not going to block merging on that.

what specific test do you propose? Testing for 3.x would break when we release 4.x we could e.g. test that the current version is greater than the previous version, but that would break if we choose a different versioning convention.

tests/test_version_derivation.py
```diff
@@ -0,0 +1,162 @@
+"""Tests for the git-tag-based version derivation configured in `pyproject.toml`.
```
d-v-b: @maxrjones have a look, this is a test that would have failed without the fix in this PR. 

I'm not really sure it's worth the code, and it feels brittle because it requires parsing our pyproject.toml file. Open to ideas for simplifying this and / or finding a better way to to test the invariant (that tags starting with `v`, and not tags starting with `zarr-metadata-v`, contribute to the output of `git describe`, given the extra parameters to that command defined in `pyproject.toml`)

maxrjones:
> what specific test do you propose? 

```python

"""Sanity check that ``zarr.__version__`` looks like a v3-or-newer release.
 
Background: zarr-python derives its version from ``git describe`` via
hatch-vcs. The repo also publishes a separate ``zarr-metadata`` subpackage
that uses ``zarr_metadata-v*`` tags. Without the ``--match v*`` filter in
``[tool.hatch] version.raw-options.git_describe_command``, ``git describe``
walks back to those subpackage tags and reports a version like ``0.2.0`` for
a from-source build of zarr-python itself — see
https://github.com/zarr-developers/zarr-python/pull/3994.
 
This test catches that class of regression: anything that makes zarr-python
report a version lower than the v3 release line. When 4.0 is released,
bump the floor; that's a deliberate, one-line edit at a planned boundary.
"""
 
from packaging.version import Version
 
import zarr
 
 
def test_version_is_v3_or_newer() -> None:
    # Use packaging.Version so we transparently handle hatch-vcs dev suffixes
    # like "3.2.2.dev30+gdc5e1825" that appear on any source build past the
    # latest v* tag — Version.major returns 3 for that string.
    parsed = Version(zarr.__version__)
    assert parsed.major >= 3, (
        f"zarr.__version__={zarr.__version__!r} is not on the v3 (or newer) "
        f"release line. If this fires on a from-source build, check that "
        f"[tool.hatch] version.raw-options.git_describe_command in "
        f"pyproject.toml still includes ``--match v*`` so the "
        f"``zarr_metadata-v*`` subpackage tags are excluded. See PR #3994."
    )
 ```

tests/test_version_derivation.py
maxrjones: what about the suggestion in https://github.com/zarr-developers/zarr-python/pull/3994#issuecomment-4510337276? It seems much less fragile and would catch bugs of this class

d-v-b:
> > what specific test do you propose?
> 
> ```python
> """Sanity check that ``zarr.__version__`` looks like a v3-or-newer release.
>  
> Background: zarr-python derives its version from ``git describe`` via
> hatch-vcs. The repo also publishes a separate ``zarr-metadata`` subpackage
> that uses ``zarr_metadata-v*`` tags. Without the ``--match v*`` filter in
> ``[tool.hatch] version.raw-options.git_describe_command``, ``git describe``
> walks back to those subpackage tags and reports a version like ``0.2.0`` for
> a from-source build of zarr-python itself — see
> https://github.com/zarr-developers/zarr-python/pull/3994.
>  
> This test catches that class of regression: anything that makes zarr-python
> report a version lower than the v3 release line. When 4.0 is released,
> bump the floor; that's a deliberate, one-line edit at a planned boundary.
> """
>  
> from __future__ import annotations
>  
> from packaging.version import Version
>  
> import zarr
>  
>  
> def test_version_is_v3_or_newer() -> None:
>     # Use packaging.Version so we transparently handle hatch-vcs dev suffixes
>     # like "3.2.2.dev30+gdc5e1825" that appear on any source build past the
>     # latest v* tag — Version.major returns 3 for that string.
>     parsed = Version(zarr.__version__)
>     assert parsed.major >= 3, (
>         f"zarr.__version__={zarr.__version__!r} is not on the v3 (or newer) "
>         f"release line. If this fires on a from-source build, check that "
>         f"[tool.hatch] version.raw-options.git_describe_command in "
>         f"pyproject.toml still includes ``--match v*`` so the "
>         f"``zarr_metadata-v*`` subpackage tags are excluded. See PR #3994."
>     )
> ```

I'm fine taking this but just not that it will fail to catch the case where the tagged version of a subpackage is greater than zarr's version, and that subpackage's git tag interferes with zarr's version inference. It will also fail if we ever change our versioning scheme.

maxrjones:
> I'm fine taking this but just not that it will fail to catch the case where the tagged version of a subpackage is greater than zarr's version, and that subpackage's git tag interferes with zarr's version inference.

yeah, I think that's an ok risk.

> It will also fail if we ever change our versioning scheme.

I think this is a good thing; it means our change in versioning scheme must be intentional and involve updating the test.

d-v-b:
> > I'm fine taking this but just not that it will fail to catch the case where the tagged version of a subpackage is greater than zarr's version, and that subpackage's git tag interferes with zarr's version inference.
> 
> yeah, I think that's an ok risk.
> 
> > It will also fail if we ever change our versioning scheme.
> 
> I think this is a good thing; it means our change in versioning scheme must be intentional and involve updating the test.

to be more precise, it _might_ fail if we change our versioning scheme, but it also might not, because it doesn't test for a specific versioning scheme, just that whatever version we use sorts after "3". But I think this risk is pretty low, so I'm going to merge this pr with this test. Thanks for your help max!

PR #3997, opened 2026-05-22
feat: top-level  function for running awaitables in zarr's event loop

d-v-b:

Adds public `zarr.run`, a synchronous bridge that runs a coroutine on Zarr's managed event loop by default, or on an explicitly supplied loop, and returns its result.

It translates the internal SyncError to RuntimeError when called from the target loop's thread. Unlike asyncio.run, it does not reject every running loop: a different active loop on the calling thread permits dispatch, although that thread is blocked. Async callers should normally await the async API directly.

Documentation and a regression test distinguish those cases. Audit validation: `hatch run test.py3.12-minimal:pytest tests/test_sync.py -q` — 27 passed; commit hooks passed. Existing timeout tests emit pending-task destruction diagnostics because this branch's sync timeout does not cancel the submitted coroutine.


d-v-b:
@TomNicholas, @aladinor let me know how this looks

PR #3998, opened 2026-05-22
refactor/consolidated JSON IO routines

d-v-b:
This is an internal refactor that does a few things related to JSON IO:

- removes the _get_json / _get_bytes methods from the store ABCs. we never used them, and that functionality has proven to be better served via helper functions that take a store as a parameter.
- adds helper functions that take a store as a parameter for doing useful stuff with JSON (reading it and writing it)
- adds helper functions that handle buffer -> JSON and vice versa

none of these things touch public APIs. the goal is removing dead code and boilerplate.


src/zarr/core/_json.py
```diff
@@ -0,0 +1,130 @@
...
+    # json.loads is typed as returning Any; the result is by definition JSON.
+    return cast("JSON", json.loads(buffer.to_bytes()))
+
+
+def buffer_to_json_object(buffer: Buffer) -> dict[str, JSON]:
+    """Parse the contents of a `Buffer` as a JSON object (a `dict`).
+
+    Every metadata document zarr reads is a JSON object, so this narrows the
+    `JSON` union to `dict[str, JSON]` once, here, instead of at each call site.
+
```
chuckwondo, line 49: Seems trivial, but do you want to add a Parameters section?

src/zarr/core/_json.py
```diff
@@ -0,0 +1,130 @@
...
+    *,
+    prototype: BufferPrototype | None = None,
+    indent: int | None = None,
+    allow_nan: bool = True,
+) -> Buffer:
+    """Serialize a JSON value into a `Buffer`.
+
+    Parameters
+    ----------
+    obj : JSON
```
chuckwondo: In all new code, do you want to stop including types in docstrings?

src/zarr/core/_json.py
```diff
@@ -0,0 +1,130 @@
...
+
+    Returns
+    -------
+    JSON or None
+        The parsed JSON value, or `None` if `key` does not exist.
+    """
+    buffer = await store.get(key, default_buffer_prototype(), byte_range)
+    if buffer is None:
+        return None
+    return buffer_to_json(buffer)
```
chuckwondo:
```suggestion
    return None if buffer is None else buffer_to_json(buffer)
```

chuckwondo (approved):
LGTM! I made some trivial/non-blocking comments.

maxrjones:
It looks like this will break Icechunk's tests (https://github.com/earth-mover/icechunk/blob/6c504cf377db2e9559703467adfa435a845ce14d/icechunk-python/tests/test_zarr/test_store/test_icechunk_store.py#L573-L583), but likely not its functionality. I'd like to delegate my review request to an Icechunk dev if possible since there's an an impact on downstream maintenance and I hold no opinion about this PR. @dcherian would you mind reviewing this one?

d-v-b:
> It looks like this will break Icechunk's tests (https://github.com/earth-mover/icechunk/blob/6c504cf377db2e9559703467adfa435a845ce14d/icechunk-python/tests/test_zarr/test_store/test_icechunk_store.py#L573-L583), but likely not its functionality. I'd like to delegate my review request to an Icechunk dev if possible since there's an an impact on downstream maintenance and I hold no opinion about this PR. @dcherian would you mind reviewing this one?

ouch, we _really_ shouldn't have downstream projects relying on private methods like this

d-v-b:
looking more closely, those tests are gated by checking the zarr version. so the icechunk devs can just cap the version range on whatever version of zarr removes the methods under test. So I don't think this should slow us down here.

maxrjones (approved):

PR #3999, opened 2026-05-22
fix: compare metadata objects by JSON form

### LINKED ISSUE #2929: Metadata comparison fails for NaN fill_values

d-v-b:
this PR makes `__eq__` and `__hash__` use the JSON form of array / group metadata objects, which means two metadata objects with `fill_value=NaN` will compare as equal, because the string form of the fill value will be used in the comparison.

This change makes our metadata objects hashable on the string form of the JSON representation. We can use a simpler form of hashing in the future if we go all-in on immutable data structures like frozendict here.

(finally) closes https://github.com/zarr-developers/zarr-python/issues/2929

### DIFF
--- a/changes/2929.bugfix.md
+++ b/changes/2929.bugfix.md
@@ -1,0 +1,3 @@
+Fix equality comparison of `ArrayV2Metadata` and `ArrayV3Metadata` objects with a
+`NaN` fill value. Such objects are now compared by their JSON-serialized form, so two
+otherwise-identical metadata objects with a `NaN` (or infinite) fill value compare equal.
--- a/src/zarr/core/metadata/v2.py
+++ b/src/zarr/core/metadata/v2.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import json
 import warnings
 from collections.abc import Iterable, Sequence
 from functools import cached_property
@@ -25,7 +26,6 @@ if TYPE_CHECKING:
         ZDType,
     )
 
-import json
 from dataclasses import dataclass, field, fields, replace
 
 import numpy as np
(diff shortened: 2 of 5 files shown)


TomNicholas (approved):

dcherian (approved):

maxrjones:
@d-v-b are you surprised by the codspeed analysis results from https://github.com/zarr-developers/zarr-python/runs/77699725053?

d-v-b:
I am surprised! we should see if it's due to changes in this PR (which I doubt) or instead changes that we merged recently without a baseline benchmark.

d-v-b:
the cause was loss of morton order caching in #3826, we should get our perf back when we bring back the caching

PR #4000, opened 2026-05-22
Run doctests via pytest and fix broken doctests.

### LINKED ISSUE #3498: missing doctests

chuckwondo:
Runs doctest via pytest and fixes all broken doctests.

Fixes #3498

TODO:
* ~~[ ] Add unit tests and/or doctests in docstrings~~
* ~~[ ] Add docstrings and API docs for any new/modified user-facing classes and functions~~
* ~~[ ] New/modified features documented in `docs/user-guide/*.md`~~


src/zarr/core/array.py
```diff
@@ -847,10 +847,13 @@ def read_chunk_sizes(self) -> tuple[tuple[int, ...], ...]:
 
         Examples
         --------
-        >>> arr = zarr.create_array(store, shape=(100, 80), chunks=(30, 40))
+        >>> import zarr.storage
+        >>> store = zarr.storage.MemoryStore()
+        >>> arr = zarr.create_array(store, dtype="i1", shape=(100, 80), chunks=(30, 40))
```
d-v-b: for these kinds of examples, we can use the URL form of the memorystore: 
```
arr = zarr.create_array("memory:///", dtype="i1", shape=(100, 80), chunks=(30, 40))
```

chuckwondo: Oh, nice! I'll tweak.

d-v-b: or `"memory:/"`

d-v-b: out of scope ofc but we should labor to comply with the URL rules defined here: https://en.wikipedia.org/wiki/File_URI_scheme, specifically as regards number of slashes

d-v-b: you can also pass an empty dict `{}` as the store, that's probably even shorter, but it means you can't refer to the store later

chuckwondo: Unfortunately, that doesn't work. Making that change causes the test to fail:

> ContainsArrayError("An array exists in store <FsspecStore(AsyncFileSystemWrapper, )> at path ''.")

chuckwondo: Sorry, not at this spot, but elsewhere. Let me get more specifics for you.

d-v-b: that reveals that the docstring tests are running in a shared session, which we probably don't want

chuckwondo: Oh, I think I see what's happening. It succeeds on the first test it hits that uses "memory:///", but all subsequent tests that use "memory:///" produce that error.  Seems like they all use the same instance.  Caching?

d-v-b: i wouldn't call it caching, they just use the same in-memory store. But I don't think separate docstring snippets should run in the same sessions -- they should be run in completely separate sessions. is that something we can control?

chuckwondo: In this case, I don't think we get the isolation you expect. If using "memory:///" causes some global state creation, it will be seen across all tests.  Only tests running in separate interpreters would be fully isolated.

If you really would prefer use of the string, I propose opening a separate issue, as I don't think it's worth blocking this PR, especially since these are all pre-existing doctests, adjusted in this PR only to resolve pre-existing problems with them.

chuckwondo:
@d-v-b, how would you like me to proceed here? Is this good to go? Shall we address the use of "memory:///" in a separate issue?

d-v-b:
> @d-v-b, how would you like me to proceed here? Is this good to go? Shall we address the use of "memory:///" in a separate issue?

that's up to you! the goal of using `"memory:///"` was to make the examples shorter, because we can skip the store import + construction lines. Since `"memory:///"` doesn't work, we can get the same shortening by using a dict literal, e.g. `open(store={}, ...)` instead of `open(store=store, ...)`. But if that feels out of scope i'm happy to deal with in in a follow-up

chuckwondo:
> > @d-v-b, how would you like me to proceed here? Is this good to go? Shall we address the use of "memory:///" in a separate issue?
> 
> that's up to you! the goal of using `"memory:///"` was to make the examples shorter, because we can skip the store import + construction lines. Since `"memory:///"` doesn't work, we can get the same shortening by using a dict literal, e.g. `open(store={}, ...)` instead of `open(store=store, ...)`. But if that feels out of scope i'm happy to deal with in in a follow-up

I've updated to use the empty dict literal and that works fine.

d-v-b (approved):
thanks chuck!

PR #4003, opened 2026-05-23
fix(FsspecStore): close owned async filesystem on store.close()

josh-ag2:
## Summary

\`FsspecStore.from_url()\` and \`from_mapper()\` create their own async filesystem instance, but \`Store.close()\` never cleaned it up. This left the aiohttp \`ClientSession\` inside s3fs open until garbage collection, producing:

\`\`\`
ResourceWarning: Unclosed client session
ResourceWarning: coroutine 'ClientCreatorContext.__aexit__' was never awaited
\`\`\`

The test file already had a \`# TODO: fix these warnings\` comment and a module-level \`pytestmark\` filter suppressing these warnings — this PR resolves the underlying cause for the \`close()\`-triggered path.

## Changes

**\`src/zarr/storage/_fsspec.py\`**

- Add \`_close_fs(fs)\` — async helper that calls \`fs.set_session()\` then \`client.close()\` for filesystems that expose \`set_session()\` (i.e. s3fs / aiobotocore); silently skips all others.
- Add \`_owns_fs: bool\` to \`FsspecStore.__init__\` (default \`False\`). Set \`True\` in \`from_url()\` unconditionally (zarr created the fs). Set \`True\` in \`from_mapper()\` only when \`_make_async()\` produced a new instance (sync→async conversion). Direct construction and \`from_upath()\` leave \`_owns_fs=False\` — the caller supplied the fs and remains responsible for its lifecycle.
- Override \`close()\` to call \`zarr_sync(_close_fs(self.fs))\` before \`super().close()\` when \`_owns_fs is True\`. Wrapped in \`contextlib.suppress(Exception)\` so it can never raise from a destructor context.

## Tests

Eight new tests in \`tests/test_store/test_fsspec.py\`:

- \`test_from_url_owns_filesystem\` — \`_owns_fs=True\` after \`from_url()\`
- \`test_from_url_close_releases_store\` — \`close()\` completes without error and sets \`_is_open=False\`
- \`test_direct_construction_does_not_own_filesystem\` — \`_owns_fs=False\` when fs supplied directly
- \`test_from_upath_does_not_own_filesystem\` — \`_owns_fs=False\` for \`from_upath()\`
- \`test_from_mapper_does_not_own_already_async_filesystem\` — \`_owns_fs=False\` when mapper's fs is already async (identity case in \`_make_async\`)
- \`test_from_mapper_owns_wrapped_sync_filesystem\` — \`_owns_fs=True\` when sync fs is wrapped
- \`test_close_fs_closes_s3_client\` — unit test (mock): \`_close_fs()\` calls \`set_session()\` then \`client.close()\`
- \`test_close_fs_no_op_for_fs_without_set_session\` — unit test (mock): \`_close_fs()\` is a no-op for fs without \`set_session\`

All 105 existing + new tests pass; 6 skipped (version-gated).

## Scope

This fix closes the S3 client session that is **active at the time \`store.close()\` is called**. It does not prevent warnings that may arise from sessions abandoned earlier — for example, s3fs with \`cache_regions=True\` internally refreshes and replaces its S3 client during I/O, discarding prior sessions before \`close()\` is ever invoked. Those intermediate sessions are an upstream s3fs issue addressed in [fsspec/s3fs#1028](https://github.com/fsspec/s3fs/pull/1028), which fixes:

- stale \`S3BucketRegionCache\` not closed on refresh
- vacuous-truth stale-session check on an empty \`_sessions\` dict
- concurrent \`asyncio.gather\` callers each creating a separate creator and tearing down each other's in-flight sessions

## Notes

- The \`pytestmark\` \`ResourceWarning\` filter is kept: it still applies to stores GC'd without an explicit \`close()\` call (e.g. \`test_basic\`) and to the residual aiobotocore \`ClientCreatorContext.__aexit__\` issue. The filter comment is updated to reflect what's now fixed vs what's still outstanding.
- \`from_upath()\` and direct \`FsspecStore(fs=..., ...)\` construction leave \`_owns_fs=False\` — the caller owns the fs in those cases, matching the existing ownership model.


d-v-b:
thanks @josh-ag2, there are a few improvements I'd like to make to this, is it OK if I push some commits?

josh-ag2:
> thanks @josh-ag2, there are a few improvements I'd like to make to this, is it OK if I push some commits?

Hi @d-v-b, please do!

d-v-b:
@josh-ag2 I pushed a few small but important changes in https://github.com/zarr-developers/zarr-python/pull/4003/commits/96d786a81902cd0b5411c598cfe708bd8f26b58e, lmk if these are objectionable otherwise I will merge this soon

josh-ag2:
> @josh-ag2 I pushed a few small but important changes in [96d786a](https://github.com/zarr-developers/zarr-python/commit/96d786a81902cd0b5411c598cfe708bd8f26b58e), lmk if these are objectionable otherwise I will merge this soon

These look good to me! Appreciate the help here, I'm a total newbie to this repo.

d-v-b (approved):

d-v-b:
thanks @josh-ag2!

d-v-b:
fyi @josh-ag2 we are discussing reverting these changes in #4165. As it turns out, there are good reasons for zarr stores to not try to handle the life cycle of the fsspec file systems.

Issue #4005, opened 2026-05-25
sharding codec equality broken due to new rng field


d-v-b:
the sharding code has a new rng field, which breaks equality between sharding codecs because two identical rngs aren't equal in the eyes of `__eq__`

d-v-b:
not needed any more now that https://github.com/zarr-developers/zarr-python/issues/4011 is in

Issue #4004, opened 2026-05-25
sharding codec subchunk write order does not round-trip through pickling


d-v-b:
the subchunk write order is not included in `__getstate__`. This should be a priority fix because otherwise sharding codecs will silently revert to the default write order when folks use dask.

maxrjones:
@d-v-b was this fixed by #4008?

d-v-b:
via https://github.com/zarr-developers/zarr-python/pull/4011! closing

PR #4006, opened 2026-05-25
fix: post-sharding codec order fixes

d-v-b:
This PR contains two fixes for the sharding codec following the changes in #3826 

- The subchunk order and rng seed now survives pickling
- Equality between codec instances now uses the rng state

@ilan-gold do you think we need `rng` to be a parameter to the constructor, or would just exposing the seed work as well? Because that would simplify things a bit. I don't think we really care which random number generator does the work, just that there's a way to make things repeatable. 

TODO:


d-v-b:
closes #4005 and #4004

d-v-b:
this is superseded by #4011

PR #4011, opened 2026-05-26
refactor: remove random subchunk ordering

d-v-b:
This PR reverts the addition of random subchunk ordering in the sharding codec which was introduced in #3826. After this PR, choosing "unordered" will write chunks lexicographically but only as an implementation detail.


src/zarr/testing/strategies.py
```diff
@@ -308,30 +308,38 @@ def arrays(
...
+                    # create_array build a second, outer ShardingCodec and nest the two,
+                    # leaving `subchunk_write_order` governing only a 1-element inner grid.
                     serializer = ShardingCodec(
                         subchunk_write_order=subchunk_write_order,
                         codecs=[BytesCodec(), ZstdCodec()],
                         index_codecs=[BytesCodec(), Crc32cCodec()],
                         chunk_shape=chunks_param,
                     )
                     compressors_unsearched = None
+                    chunks_param = shard_shape
```
ilan-gold: So the `chunks` param controls shard layout when the `serializer` is used?

ilan-gold (approved):

src/zarr/testing/strategies.py
d-v-b: good question, I need to double-check this interaction! this is a weakness of the create_array api.

d-v-b: the changes were not necessary for the work in this PR so I reverted them. we can fix the bug here in a separate PR.

PR #4012, opened 2026-05-26
perf: cache lexicographic chunk coords in sharding codec

d-v-b:
Fixes a performance regression introduced in #3826 that made morton-order-indexing slower.

for details, see claude's summary below. 

The subchunk_write_order feature (#3826) regressed sharded write performance: _encode_partial_single rebuilt the full per-shard chunk coordinate grid on every write via
`np.array(list(_subchunk_order_iter(..., "lexicographic")))`, and `to_dict_vectorized` rebuilt a tuple key per row with `tuple(coords.ravel())`. For a single-chunk write into a shard with tens of thousands of chunks this roughly doubled write time (~22ms -> ~40ms on test_sharded_morton_write_single_chunk, matching the -44% CodSpeed regression).

Add cached `_lexicographic_order` (array) and `_lexicographic_order_keys` (tuples) helpers in indexing.py, mirroring `_morton_order`/`_morton_order_keys`, and pass the cached keys into `to_dict_vectorized` instead of deriving them row-by-row. This restores write throughput to the pre-#3826 baseline while preserving identical chunk ordering (verified equal to np.ndindex across shapes including 0-d and empty).

[Description of PR]

TODO:


src/zarr/codecs/sharding.py
```diff
@@ -266,13 +269,19 @@ def __iter__(self) -> Iterator[tuple[int, ...]]:
     def to_dict_vectorized(
```
maxrjones: I think the following reads cleaner than adding another parameter. The reader already knows its own `chunks_per_shard`, so it can fetch both cached structures itself instead of having them threaded in. With this, the call site simplifies to just `shard_reader.to_dict_vectorized()`.

I tested it still fixes the regression — same ~1.6–1.8× speedup over `main` on `test_sharded_morton_write_single_chunk`, and the difference vs the committed version is within run-to-run noise.

```python
    def to_dict_vectorized(self) -dict[tuple[int, ...], Buffer | None]:
        """Build a dict of chunk coordinates to buffers using vectorized lookup.

        The full per-shard chunk coordinate grid (both the array used for the
        vectorized index lookup and the plain tuples used as dict keys) is
        cached on ``chunks_per_shard``, so neither is rebuilt on every call.
        For a shard with tens of thousands of chunks this avoids reconstructing
        that many tuples on every partial write.

        Returns
        -------
        dict mapping chunk coordinate tuples to Buffer or None
        """
        chunks_per_shard = self.index.chunks_per_shard
        chunk_coords_array = _lexicographic_order(chunks_per_shard)
        chunk_coords_keys = _lexicographic_order_keys(chunks_per_shard)
        starts, ends, valid = self.index.get_chunk_slices_vectorized(chunk_coords_array)

        result: dict[tuple[int, ...], Buffer | None] = {}
        for i, coords in enumerate(chunk_coords_keys):
            if valid[i]:
                result[coords] = self.buf[int(starts[i]) : int(ends[i])]
            else:
                result[coords] = None

        return result
```

d-v-b: that's great feedback! have a look at https://github.com/zarr-developers/zarr-python/pull/4012/commits/6e02c8f833fd64aaf106c84b540324e9aefc71b1 and see if things look better to you

src/zarr/core/indexing.py
```diff
@@ -1584,6 +1584,33 @@ def morton_order_iter(chunk_shape: tuple[int, ...]) -> Iterator[tuple[int, ...]]
...
+    return order
+
+
+@lru_cache(maxsize=16)
+def _lexicographic_order_keys(chunk_shape: tuple[int, ...]) -> tuple[tuple[int, ...], ...]:
+    return tuple(tuple(int(x) for x in row) for row in _lexicographic_order(chunk_shape))
+
+
+def lexicographic_order_iter(chunk_shape: tuple[int, ...]) -> Iterator[tuple[int, ...]]:
+    return iter(_lexicographic_order_keys(tuple(chunk_shape)))
```
ilan-gold:
```suggestion
    return iter(_lexicographic_order_keys(chunk_shape))
```
or?

ilan-gold (approved):
Apologies for the regression!

src/zarr/core/indexing.py
```diff
@@ -1584,6 +1584,33 @@ def morton_order_iter(chunk_shape: tuple[int, ...]) -> Iterator[tuple[int, ...]]
...
+    return order
+
+
+@lru_cache(maxsize=16)
+def _lexicographic_order_keys(chunk_shape: tuple[int, ...]) -> tuple[tuple[int, ...], ...]:
+    return tuple(tuple(int(x) for x in row) for row in _lexicographic_order(chunk_shape))
+
+
+def lexicographic_order_iter(chunk_shape: tuple[int, ...]) -> Iterator[tuple[int, ...]]:
+    return iter(_lexicographic_order_keys(chunk_shape))
```
chuckwondo: Not sure this makes a lot of sense. Why would we provide a "convenience" function for producing an "lazy" iterator over something that has already been eagerly constructed? Shouldn't we expose the lazy option to the user and then the user can eagerly collect the results, if that's what they want.  Otherwise, we simply provide no means for the caller to get lazy results.

chuckwondo: Although, in this case, does being lazy even matter? I'm not familiar enough with what's going on here to tell if chunk_shape is always something relatively small, in which case laziness is likely unnecessary.

ilan-gold: I didn't want to open this can of worms, but I agree things are funny here. I don't think the iter function here is super important but 

> user can eagerly collect the results, if that's what they want.

The idea here is definitely to cache the eager results and then let the user create an iterator. But I don't even think an iterator is necessary `_lexicographic_order_keys` should satisfy the APIs that `lexicographic_order_iter` is meant for e.e.,g

```python
dict.fromkeys((1, 2, 3))
```
 works just fine (which is one place where the iterator is used).

mkitti: It really should be the other way around. We should define the lazy version and the eager version should just be "collect" the lazy version.

The problem is that pay an enormous penalty for generating the entire list of all chunks when we might only just want the first one.

src/zarr/core/indexing.py
```diff
@@ -1581,11 +1581,47 @@ def _morton_order_keys(chunk_shape: tuple[int, ...]) -> tuple[tuple[int, ...], .
...
+    return tuple(tuple(int(x) for x in row) for row in _lexicographic_order(chunk_shape))
+
+
+def lexicographic_order_iter(chunk_shape: tuple[int, ...]) -> Iterator[tuple[int, ...]]:
+    # Iterate the chunk grid in lexicographic (row-major / C) order. The full
+    # coordinate tuple is built once and cached on `chunk_shape` (see
+    # _lexicographic_order_keys), so repeated calls for the same shape — e.g. one
+    # per shard write — reuse it rather than rebuilding tens of thousands of
+    # tuples each time.
+    return iter(_lexicographic_order_keys(chunk_shape))
```
mkitti:
```suggestion
def _lexicographic_order_keys(chunk_shape: tuple[int, ...]) -> tuple[tuple[int, ...], ...]:
    return tuple(lexicographic_order_iter(chunk_shape))


def lexicographic_order_iter(chunk_shape: tuple[int, ...]) -> Iterator[tuple[int, ...]]:
    # Iterate the chunk grid in lexicographic (row-major / C) order. The full
    # coordinate tuple is built once and cached on `chunk_shape` (see
    # _lexicographic_order_keys), so repeated calls for the same shape — e.g. one
    # per shard write — reuse it rather than rebuilding tens of thousands of
    # tuples each time.
    return (tuple(int(x) for x in row) for row in _lexicographic_order(chunk_shape))
```

src/zarr/core/indexing.py
d-v-b: agreed! this pattern is what you get when you optimize for benchmarks. we might need to relax that concern, or add a separate benchmark that checks "time to fetch 1 chunk"

src/zarr/core/indexing.py
mkitti: The lazy "lexicographic_order_iter" uses a generator comprehension. The eager "_lexicographic_order_keys " just pulls that into a tuple.

d-v-b:
something I added here: I realized that we already had a `c_order_iter` function, but `lexicographic_order_iter` is a much better name for this functionality. so `c_order_iter` is now just a wrapper for `lexicographic_order_iter`.

src/zarr/core/indexing.py
```diff
@@ -1581,11 +1581,51 @@ def _morton_order_keys(chunk_shape: tuple[int, ...]) -> tuple[tuple[int, ...], .
...
+def _lexicographic_order_keys(chunk_shape: tuple[int, ...]) -> tuple[tuple[int, ...], ...]:
+    # Eagerly collect `lexicographic_order_iter` into a tuple, cached per shape so
+    # the per-shard chunk grid is built once and reused across writes (the hot
+    # path feeds this straight into `dict.fromkeys`).
+    return tuple(lexicographic_order_iter(chunk_shape))
 
 
 def c_order_iter(chunks_per_shard: tuple[int, ...]) -> Iterator[tuple[int, ...]]:
-    return itertools.product(*(range(x) for x in chunks_per_shard))
+    # Soft-deprecated alias for `lexicographic_order_iter`. "C order" names the
```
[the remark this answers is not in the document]
d-v-b: there are no more callers of `c_order_iter` in the codebase, so this is not an issue.

d-v-b:
thanks for the careful review all! I think I've addressed everything

src/zarr/core/indexing.py
```diff
@@ -1581,11 +1581,51 @@ def _morton_order_keys(chunk_shape: tuple[int, ...]) -> tuple[tuple[int, ...], .
 
 
 def morton_order_iter(chunk_shape: tuple[int, ...]) -> Iterator[tuple[int, ...]]:
-    return iter(_morton_order_keys(tuple(chunk_shape)))
+    return iter(_morton_order_keys(chunk_shape))
```
mkitti:
```suggestion
@lru_cache(maxsize=16)
def _morton_order_keys(chunk_shape: tuple[int, ...]) -> tuple[tuple[int, ...], ...]:
    return tuple(morton_order_iter)


def morton_order_iter(chunk_shape: tuple[int, ...]) -> Iterator[tuple[int, ...]]:
    return (tuple(int(x) for x in row) for row in _morton_order(chunk_shape))
```

This is also backwards.

mkitti: The suggestion is not applying to the correct lines. Manual intervention is likely required.

d-v-b: done in https://github.com/zarr-developers/zarr-python/pull/4012/commits/68402491328347b820b9f13c61496070026b1c7c

mkitti (approved):

mkitti:
@codspeedbot explain why this is faster

mkitti:
@codspeedbot explain why this is faster

mkitti:
Claude review incoming...

src/zarr/core/indexing.py
```diff
@@ -1575,17 +1575,61 @@ def _morton_order(chunk_shape: tuple[int, ...]) -> npt.NDArray[np.intp]:
     return order
 
 
+def morton_order_iter(chunk_shape: tuple[int, ...]) -> Iterator[tuple[int, ...]]:
```
mkitti: **Issue 1 (high): this de-optimizes a hot write path — the opposite of this PR's goal.**

Previously `morton_order_iter` returned `iter(_morton_order_keys(...))`, i.e. it handed back an `lru_cache`d tuple-of-tuples, so repeated calls were essentially free. Now it rebuilds every coordinate tuple from the (cached) array on every call.

`morton_order_iter` is consumed in `_encode_shard_dict`, the per-shard serialization path that runs on **every shard write** — so this reintroduces exactly the per-write tuple-reconstruction cost this PR removes for the lexicographic case. Two knock-on effects:
- `_morton_order_keys` is now **dead in `src/`** (only the benchmark's `cache_clear` references it); its cache is never populated in production.
- The asymmetry is backwards: the lexicographic hot path uses cached `_lexicographic_order_keys`, while the morton hot path uses the *un*cached `morton_order_iter`.

Suggest reverting the morton inversion (keep `morton_order_iter` delegating to cached `_morton_order_keys`), or route the `_encode_shard_dict` call site through `_morton_order_keys`. Source-shape symmetry isn't worth a hot-path de-optimization in a perf PR.

_— Claude (Claude Code)_

src/zarr/codecs/sharding.py
```diff
@@ -686,7 +689,8 @@ def _is_total_shard(
         self, all_chunk_coords: set[tuple[int, ...]], chunks_per_shard: tuple[int, ...]
     ) -> bool:
         return len(all_chunk_coords) == product(chunks_per_shard) and all(
-            chunk_coords in all_chunk_coords for chunk_coords in c_order_iter(chunks_per_shard)
+            chunk_coords in all_chunk_coords
+            for chunk_coords in lexicographic_order_iter(chunks_per_shard)
```
mkitti: **Issue 2 (medium): this is strictly worse than the code it replaced for the early-exit case, and the commit message claims the reverse.**

This previously iterated `c_order_iter` = `itertools.product` — genuinely lazy, O(1) memory, true early-exit. `lexicographic_order_iter` calls `_lexicographic_order`, which **eagerly materializes the whole `np.indices` grid** before yielding the first tuple. So the "laziness" only skips tuple *conversion* of unconsumed rows, not the expensive grid build.

Concretely, commit f6d91e6's claim ("on a cold cache for a 32³ shard … ~15.8ms → ~24µs, the lazy generator builds one coordinate and bails") can't hold here — `np.indices((32,32,32))` runs in full on the first `next()`. For the pure early-exit case on a cold cache, the original `itertools.product` was strictly better and used less memory.

In practice this is mostly masked because the `dict.fromkeys` hot path warms `_lexicographic_order` for the same shape. But if the early-exit win is actually wanted, make `lexicographic_order_iter` a true generator (e.g. `itertools.product`) and have `_lexicographic_order_keys` collect *that* — otherwise drop the early-exit framing from the commit history.

_— Claude (Claude Code)_

src/zarr/core/indexing.py
```diff
@@ -1575,17 +1575,61 @@ def _morton_order(chunk_shape: tuple[int, ...]) -> npt.NDArray[np.intp]:
...
+
+@lru_cache(maxsize=16)
+def _lexicographic_order_keys(chunk_shape: tuple[int, ...]) -> tuple[tuple[int, ...], ...]:
+    # Eagerly collect `lexicographic_order_iter` into a tuple, cached per shape so
+    # the per-shard chunk grid is built once and reused across writes (the hot
+    # path feeds this straight into `dict.fromkeys`).
+    return tuple(lexicographic_order_iter(chunk_shape))
 
 
 def c_order_iter(chunks_per_shard: tuple[int, ...]) -> Iterator[tuple[int, ...]]:
```
mkitti: **Issue 3 (low-med): the alias quietly changes operational characteristics.**

`c_order_iter` went from pure-lazy `itertools.product` (no allocation, no cache) to an alias that materializes and `lru_cache`s a full numpy grid. For any large-`chunks_per_shard` caller this is a cold-start + memory regression, and it now retains up to 16 large arrays. In practice `chunks_per_shard` is modest so it's likely fine — but it's a real semantic change presented as a no-op "soft-deprecated alias." Worth calling out explicitly in the PR description.

_— Claude (Claude Code)_

mkitti (commented):
_This review was authored by Claude (Claude Code), at the request of @mkitti._

Reviewed the diff critically. **The core fix is sound** — caching `_lexicographic_order`/`_lexicographic_order_keys` and feeding the cached keys straight into `to_dict_vectorized`/`dict.fromkeys` removes exactly the two per-write costs #3826 introduced (`np.array(list(...))` and per-row `tuple(coords.ravel())`). The `to_dict_vectorized()` signature simplification is a genuine improvement (the reader already knows `chunks_per_shard`), order equivalence with `np.ndindex` holds across 0-d/empty/N-d, and `chunk_coords_array[i]`/`chunk_coords_keys[i]` are guaranteed aligned.

My concern is that the PR has accreted a lazy/eager + naming-symmetry refactor that overshoots the goal, and one piece works *against* it. Details inline, summarized by severity:

1. **(high)** The morton "symmetry" commit de-optimizes a hot write path — see comment on `morton_order_iter`.
2. **(medium)** `_is_total_shard` is actually made worse than the `itertools.product` it replaced, and the commit rationale claims the reverse — see comment on `_is_total_shard`.
3. **(low-med)** `c_order_iter` quietly changes operational characteristics (lazy/no-alloc → eager + cached numpy grid) — see comment on `c_order_iter`.

Minor: dropping `tuple(chunk_shape)` coercion in `morton_order_iter` means a non-tuple arg now fails inside `lru_cache` rather than being coerced (acceptable per type hints); `_lexicographic_order` caches a non-contiguous transposed view that pins its larger `np.indices` base alive; and the rename/soft-deprecation is scope creep relative to a perf-regression fix — worth a line in the PR description so the `c_order_iter` behavior change is intentional and visible.

Net: I'd approve the lexicographic caching + `to_dict_vectorized()` change as-is, but push back on item 1 and trim/correct the item 2 narrative before merge.

src/zarr/core/indexing.py
d-v-b: I think @chuckwondo's point indicates something more basic -- there's no reason to expose an _iteration_ interface, lazy or not, for something that has a defined size, like a dense, finite, coordinate grid. We should be exposing an indexable data structure from the start.

d-v-b:
thanks for everyone's patience here. Some recent changes:
- I added explanatory comments for _why_ we need two representations (tuples of ints and numpy arrays of ints) of the same data in `to_dict_vectorized`. Just using Numpy arrays of ints would be more performant, but fully exploiting that requires changes to the calling code, which models chunk indices as tuples of ints. IMO we need to consider the chunk index representation a target for performance optimization.
- I removed the premise of laziness from the `*_order` routines. Now they are not iterators, they return full collections of coordinates, and they have names that make this clear: `morton_order_coords` and `lexicographic_order_coords`. I also changed the name of the parameters from `chunk_shape` to just `shape`, because these functions are not just for chunks, and `chunk_shape` is actually rather ambiguous.

@mkitti feel free to throw your agents at the latest changes.

d-v-b:
@claude please review

d-v-b:
I'm going to merge this today, but I will register that I think we can improve our API here a bit. What I suggest is defining some simple objects that lightly wrap numpy arrays, and making the strings "morton" or "lexicographic" parameters that define the construction of these arrays. Then we would have something like `indices=grid_order(shape, order="morton")`, which could be very re-usable across the codebase.

mkitti:
Do we have a way of catching performance regressions in the future?

d-v-b:
Nothing shaped like a test. we just have the benchmark suite, but for PR granularity this requires a maintainer opt in explicitly to running it. Otherwise benchmarks are weekly, and without much fanfare. I think the story here will be that we periodically check for regressions, and deal with them as needed.

Issue #4016, opened 2026-05-29
DOC: create_array() got an unexpected keyword argument 'mode'


sdebionne:
### Describe the issue linked to the documentation

The [Persistent Storage](https://zarr.readthedocs.io/en/stable/quick-start/#persistent-storage) section of the  Getting Started chapter reads:

```
import s3fs

z = zarr.create_array("s3://example-bucket/foo", mode="w", shape=(100, 100), chunks=(10, 10), dtype="f4")
z[:, :] = np.random.random((100, 100))
```

But `mode="w"` gives an "unexpected keyword" RuntimeError.

```
$ pixi list zarr
Name  Version  Build               Size  Kind   Source
zarr  3.2.1    pyhc364b38_0  359.10 KiB  conda  https://conda.anaconda.org/conda-forge
```

### Suggested fix for documentation

Drop the `mode="w"` parameter?

d-v-b:
thanks for this report, I'll get a fix out soon!

d-v-b:
this was closed by https://github.com/zarr-developers/zarr-python/pull/4018

PR #4018, opened 2026-05-29
docs: test more code snippets in docs

### LINKED ISSUE #4017: run all code snippets in docs
#4016 revealed that we have a code example in our docs that doesn't work because it uses a fictional function parameter. As that code example demonstrates how to write to cloud storage, it's not tested as part of our doctest suite, which seems like something we need to fix. If we run this test with the same mock s3 environment we use for our s3 tests, everything should work.

There are a few other tests that don't run because our doctest environment doesn't have the right resources (GPU tests). These should also be runnable in the right pytest environment.

d-v-b:
Fixed an invalid `zarr.create_array` example in the quick-start documentation (it passed an unsupported `mode` argument) and made the cloud-storage example execute against a mock S3 backend in CI. Added a test ensuring every Python code block in the documentation is either executed or explicitly opted out with a documented reason, so an invalid example can no longer go untested.

The changes here touch on our s3 tests because we needed to re-use the mock s3 backend to ensure that the s3 example in the docs runs properly.

closes #4017 and #4016


docs/contributing.md
```diff
@@ -253,6 +253,64 @@ renders as:
 print("Hello world")
 ```
 
+#### Validating code blocks: `exec` vs `test`
```
d-v-b, line 256: this addition to the docs explains the logic added in this PR

maxrjones (approved):
thanks!

Issue #4027, opened 2026-06-03
fold `AsyncArray` methods into `Array`


d-v-b:
It would be convenient if we had a single array class instead of two (`Array` and `AsyncArray`). 

`Array` exists because users expect an object that has blocking IO methods. `AsyncArray` exists because under the hood, those blocking IO methods need to use async routines. Lurking in the background is the `sync` function, which we use to run coroutines in an event loop running in a zarr-owned daemon thread. 

schematically, this is what we have today:

```python
class AsyncArray:
     async def do_stuff() -> coroutine[result]: 
        # actually do stuff

class Array:
    _async_array: AsyncArray

    def do_stuff() -> result:
        return sync(self._async_array.do_stuff())
```

An alternative approach: 

```python

class Runner(Protocol):
    def run(coro: Awaitable[T]) -> T: ...

class DefaultRunner:
    def run(coro: Awaitable[T]) -> T:
        return sync(coro)

class Array:
    _runner: Runner
    def __init__(self, runner: Runner | None = None) -> None:
        if runner is None: runner = DefaultRunner()
        self._runner = runner

    def do_stuff_async() -> coroutine[result]:
       # actually do stuff

    def do_stuff() -> result:
        return self._runner.run(self.do_stuff_async())
```

As long as python async works the way it does, we will be dealing with 2 versions of the same function. so IMO we should accept that, and endeavor to put the two versions as close together as possible in the codebase. This proposal achieves that. And we get to _eventually_ remove an entire class! (NOTE! I'm not saying we will do that immediately. the goal of removing dead code is not a promise to immediately do breaking changes).

maxrjones:
> the goal of removing dead code is not a promise to immediately do breaking changes

While the decision is deferred, the goal to _eventually_ remove the class is pretty clearly influencing the design decisions in https://github.com/zarr-developers/zarr-python/pull/4034. So it's worth discussing now whether that class is worthwhile to remove, because it influences whether we use a design that's optimal for the case that we won't remove that class or one that's optimal for the case that we will eventually remove that class.

I want to separate two things the proposal bundles together: the refactor (folding the two methods into one class) and the eventual removal of the class. Those are different decisions with different costs. Right now it's the _option_ to remove — not any actual removal — that's distorting #4034, which is why I don't think we should leave it as a deferred maybe.

[...]

PR #4034, opened 2026-06-04
refactor: one array class

d-v-b:

Adds a configurable `Runner` to `Array` and async counterparts for its I/O operations. Synchronous wrappers use the runner to execute coroutines; the default `SyncRunner` uses Zarr's shared background event loop through `sync`.

This is not a one-to-one async duplication of every property, special method, or synchronous helper. `AsyncArray` remains available while the unified `Array` interface develops.

Compatibility is a goal rather than a claim that every observable behavior is unchanged: constructing `Array(async_array)` now emits a deprecation warning. Direct construction accepts metadata, a store path, and optional configuration and runner.


d-v-b:
the main breaking change I was worried about here is `Array.__init__`, which previously took a single `AsyncArray` argument, e.g. `Array(my_async_array)`. This still works on this branch, but it's deprecated. The preferred constructor for `Array` looks much more user-friendly now -- you pass metadata, a storepath, etc.

src/zarr/core/array.py
```diff
@@ -200,6 +202,20 @@ def _chunk_sizes_from_shape(
     return tuple(result)
 
 
+def _sharding_codec(metadata: ArrayMetadata) -> ShardingCodec | None:
+    """Return the array's sharding codec, or `None` if the array is not sharded.
+
+    An array is considered sharded when its metadata declares exactly one codec
+    and that codec is a `ShardingCodec`.
+    """
```
maxrjones, line 206:
```suggestion
    """Return the array's sole sharding codec, or `None`.

      The gate used by the chunk/shard accessors: sharding is reported only
      when the sharding codec is the only declared codec, because any other
      codec (e.g. ``codecs=[sharding, gzip]``) makes the inner chunks not
      independently addressable. Not a general sharded-ness predicate; see #4036.
    """
```
I think it's important to document that this won't always return the sharding codec to prevent misuse.

maxrjones (commented):
Why did you choose this approach over using `AsyncArray` as the single source of truth, and making `Array` a facade holding a reference to it? The regression section details the most critical issue with this PR, and would be prevented by having `AsyncArray` as the single source of truth. Using `AsyncArray` as the single source of truth would reduce the amount of code needed by a few hundred lines (see duplication section). A facade `Array` can still expose every `*_async` method and the runner-dispatched sync wrappers by delegating to the held `AsyncArray`, so the user-facing API of this PR is identical either way.

 The regression and the code duplication below are two symptoms of the same root cause: after this PR,  `Array` and `AsyncArray` each independently own array state and derive behavior from it. Caching the handle would fix the regression and delegating to the shared helpers would fix the method duplication. Even with those changes, the state and the derived properties remain mirrored across both classes, and every future change must land on both sides. The facade removes the second source of truth instead of patching its symptoms one at a time.

Also, what's the use-case for a synchronous Array having a custom runner? I would think that most people who want to bring an event loop would just use the async methods. It seems like a YAGNI case, and not worth exposing publicly in this PR since it's not usable via `zarr.open_array`, `zarr.create_array`, etc.

## regression 

On `main`, `Array.async_array` returned the shared backing `_async_array`, and all `Array` properties read through it — so `arr.async_array.resize((N,))` updated `arr.shape` too. On this branch the property constructs a **fresh, throwaway** `AsyncArray(self.metadata, self.store_path, self.config)` on every access. Three consequences, all verified:

- `arr.async_array.resize((N,))` — a previously working pattern — now runs `_resize` against the detached temporary: the store metadata **is** updated (`array.py:6683`) but `arr.metadata`/`arr.shape` stay stale. Subsequent `arr[...]` reads/writes index against the old shape while the store has the new one.
- `aa = arr.async_array; arr.resize(...)` leaves `aa` holding the pre-resize metadata, since `resize` rebinds `arr.metadata` via `object.__setattr__` on `arr` only (`array.py:4634`, `6686`).
- `arr.async_array is arr.async_array` is now `False`, breaking identity-based caching.

The changelog frames `async_array` as "deprecated but still works for now"; for mutating operations it does not work — it silently desynchronizes the handle from the array.

## code duplication avoided by using `AsyncArray` as the single source of truth

`AsyncArray`'s selection methods are one-line delegations to shared module-level helpers:

```python
# AsyncArray.get_orthogonal_selection (array.py:1544)
return await _get_orthogonal_selection(
    self.store_path, self.metadata, self.codec_pipeline, self.config, self._chunk_grid,
    selection, out=out, fields=fields, prototype=prototype,
)
```

The new `Array.get_orthogonal_selection_async` re-inlines the body of that same helper instead of calling it:

```python
# Array.get_orthogonal_selection_async (array.py:2865)
if prototype is None:
    prototype = default_buffer_prototype()
indexer = OrthogonalIndexer(selection, self.shape, self._chunk_grid)
return await self._get_selection(indexer=indexer, out=out, fields=fields, prototype=prototype)
```

```python
# _get_orthogonal_selection — the existing shared helper (array.py:6336)
if prototype is None:
    prototype = default_buffer_prototype()
indexer = OrthogonalIndexer(selection, metadata.shape, chunk_grid)
return await _get_selection(..., indexer=indexer, out=out, fields=fields, prototype=prototype)
```

The same pattern repeats for the mask, coordinate, and block selection getters and setters. Beyond the methods:

- `Array._info` (array.py:1954) is a byte-for-byte copy of `AsyncArray._info` (array.py:1825) — a 20-line `ArrayInfo` construction duplicated verbatim.
- The `__init__` state-derivation block (`parse_array_metadata` / `ChunkGrid.from_metadata` / `create_codec_pipeline`) is duplicated between `AsyncArray.__init__` (array.py:415–426) and `Array.__init__` (array.py:1892–1902).
- The derived properties (`order`, `read_only`, `filters`, `serializer`, `compressors`, `nchunks`, …) are mirrored across both classes.

With the docstrings these twins carry, that's the few hundred lines. Each pair is a place where a future fix can land on one side and silently miss the other. Sync and async selection silently returning different results for the same call. A facade `Array` delegating to its held `AsyncArray` would have exactly one copy of each.

maxrjones:
Here's Claude's evaluation of the pros/cons of either class being the facade, which I think is the cleanest way to avoid duplicate state. It seems to come down to how much diff/compatibility risk we accept now in exchange for `Array` being the real implementation and `AsyncArray`'s eventual removal being trivial (i.e., whether #4027's removal goal is a commitment or an aspiration) My preferred goals are to reduce the duplication between the two classes, make whatever duplication remains mechanically checkable (see "Keeping the mirrored pairs in sync" below), advise users toward the single class, but keep the other around in perpetuity so people's code keeps working. By the criterion at the end, this position lands on Option A, with a parity check making the permanent mirroring affordable.

Note, a third shape exists where both classes hold a shared mutable state object, with methods delegating to the module-level helpers this branch already has, but it adds a third class without changing the analysis below much.

## Option A — review's direction: `AsyncArray` owns implementation, `Array` facades it

**Pros**

- **Smallest, safest diff.** This is `main`'s architecture (where `Array` is a dataclass holding a single field, `_async_array`, with all properties reading through it) plus new `*_async` delegating methods on `Array`. The shape is battle-tested; downstream (xarray, etc.) has run against it for years.
- **Regression fixed for free.** `async_array` returns the held object — identity, shared state, mutation coherence all come from there being only one stateful object, with zero new machinery.
- **Async API untouched.** `AsyncGroup.getitem`, `zarr.api.asynchronous` keep constructing self-contained `AsyncArray`s; no sync machinery (runner, facade plumbing) enters the async path.
- **Clean layering.** Async is the primitive, sync wraps it — the conventional direction; every reviewer's prior matches it.
- **No dataclass-semantics churn.** `AsyncArray` keeps its real fields; anything using `dataclasses.replace`/`fields()` on it keeps working.

**Cons**

- **Fights #4027's end goal.** "Eventually remove an entire class" requires a second migration later — moving every implementation off `AsyncArray` onto `Array`, roughly this PR's diff again, plus another deprecation cycle.
- **`Array` stays second-class.** The flagship API of this PR is a pure wrapper: every `*_async` method is a delegation whose signature + docstring must still be mirrored on `Array` (the few hundred lines of signatures/docs still exist; only the bodies shrink to one line). The "one array class" goal isn't structurally advanced — it's `main` with new method names.
- **Two-class mental model persists indefinitely**, with the docs burden of explaining both.

## Option B — inverted: `Array` owns implementation (async impls as `*_async`), `AsyncArray` facades it

**Pros**

- **Achieves #4027's end state now.** Sync and async versions of each operation are co-located on one class ("put the two versions as close together as possible"); deleting `AsyncArray` later is removing a shim, not a migration.
- **The new public API is first-class.** Real implementations, real docstrings on `Array`; no delegation hop for new-API users; deprecation pressure points at the legacy class, which is the thin one.
- **Same regression fix, same deduplication.** A cached `Array.async_array` view sharing state fixes all three symptoms; every duplication item in the review (re-inlined selection bodies, `_info`, `__init__` derivation, mirrored properties) collapses to one copy — the line savings are identical to Option A.
- **One-time cost.** The read-through plumbing you write now is the *last* restructuring; Option A's ease is partly deferred cost.

**Cons**

- **Larger, riskier diff now.** `AsyncArray`'s fields (`metadata`, `store_path`, `config`, `codec_pipeline`, `_chunk_grid`) become read-through properties over a held `_array`; all its construction sites (group.py:791, group.py:3506, asynchronous.py:369, ~77 references in group.py) flow through the facade.
- **Dataclass breakage.** `AsyncArray` is `@dataclass(frozen=True)` with real fields today; converting fields to properties changes `dataclasses.fields()`, `replace()`, generated `__eq__`/`repr`. Needs an audit of internal and downstream `replace()` use — this is the most concrete compat hazard.
- **Layering inversion.** Every `AsyncArray` now wraps an `Array` carrying a `_runner` it never uses; the async path depends on the unified class's sync machinery. Mostly aesthetic, but it's the part reviewers will push on.
- **The shim still costs its lines — and dynamic delegation doesn't rescue it yet.** One might hope the facade could be generated dynamically (`__getattr__`/class decorator), but `AsyncArray` is *not* deprecated for async users — it's the typed return value of the entire public async API. Until that changes, the facade must be fully typed and documented, so it's hand-written one-line delegations with mirrored signatures, not `__getattr__` magic. Dynamic delegation only becomes acceptable in a later phase, once the async API returns `Array` (or equivalent) and `AsyncArray` is deprecated end-to-end.
- **Property-hop overhead** on `AsyncArray.metadata`-style access in internal async hot paths. Almost certainly negligible, but it's new indirection where Option A adds none.

## The symmetric parts (don't discriminate)

- Duplication eliminated: identical in both.
- Single-source-of-truth / "changes land once": identical in both.
- The mirrored-signatures+docstrings burden: identical in both — it just sits on `Array` in A and on `AsyncArray` in B. In B it's temporary (deleted with the shim); in A it's permanent (though see the next section — "permanent" is cheap if it's CI-enforced).
- Runner YAGNI: orthogonal — keep `_runner` out of the public constructor signature in either design.

## Keeping the mirrored pairs in sync mechanically

Whichever class is the facade, the residual duplication is mirrored signatures + docstrings, and that can be guarded by tooling rather than vigilance. The key observation: once the facade's method bodies are one-line delegations, *behavior* can't drift between the pairs — signature and docstring drift is the only remaining risk, and it's mechanically checkable. In order of weight:

- **Parity test (cheapest — no new infrastructure).** A pytest test that pairs `foo`/`foo_async` across the two classes by introspection and asserts `inspect.signature` equality, with an explicit exceptions table for the pairs that legitimately differ (e.g. `resize(new_shape)` vs `resize_async(new_shape, delete_outside_chunks=True)`). Drift becomes a CI failure instead of a silent bug, and everything stays hand-written, fully typed, and documented.
- **Docstring sharing.** The "This is the asynchronous variant of …" boilerplate can be generated by a small decorator that copies the source-of-truth docstring and prepends the variant note, so the prose is written once.
- **Full codegen (heaviest).** Generate the entire facade as checked-in source from the implementation class, with a CI step that fails if regeneration differs — the `unasync` pattern (psycopg 3 maintains its whole sync API this way; urllib3/httpcore have used `unasync`). Zero hand-maintenance with full typing and docs, but the generator needs the same exceptions table and is real infrastructure to own.

To be clear, this is a complement to the facade, not a substitute for it: the parity test only guards signatures. The regression above (shared-state desync) and the re-inlined helper bodies are behavioral problems that only the facade fixes.

## Decision criterion

It reduces to one question: **is #4027's "eventually remove `AsyncArray`" a commitment or an aspiration?** If there's consensus to actually deprecate and remove it (including changing what the async API returns — a bigger conversation than this PR), B's extra diff is a down payment on work that must happen anyway, and A means doing this restructuring twice. If removal is indefinite, B's costs (dataclass churn, facade plumbing, layering inversion) buy nothing that A doesn't deliver more cheaply — especially with a parity test making A's permanent mirroring a CI-enforced non-problem rather than a vigilance problem — and A is simply the better engineering call. A defensible middle path: take A now (it's the minimal fix for the review's confirmed regression and duplication), and treat B as the mechanical follow-up PR that executes the removal decision once it's actually made — A doesn't foreclose B, it just doesn't prepay for it.

d-v-b:
> It reduces to one question: is https://github.com/zarr-developers/zarr-python/issues/4027's "eventually remove AsyncArray" a commitment or an aspiration?

Yes, the whole point is to have a single class that has the sync and async apis. That class should be the `Array` class, not the `AsyncArray` class, because the `Array` class is the primary user-facing class. Once the `Array` class does everything the `AsyncArray` class used to do, there is no reason for the `AsyncArray` class to exist, and we can remove it. Consolidating functionality on the `Array` class and eventually removing the `AsyncArray` class is the entire goal of this effort.

src/zarr/core/array.py
d-v-b: Does the code today actually do the check you are describing?

d-v-b:
the regressions are fixed and I (via claude) added tests to ensure that the `Array` and `AsyncArray` have consistent API surfaces.

kylebarron:
I'm missing some context I'm sure, but some opinions that came out of Slack discussion with @d-v-b:

I have strong opinions to separate sync and async code paths **completely** at low and medium API levels. _Only_ the highest-level API should paper over differences between sync and async APIs, for users who just want _something_ to work, regardless of the backend, and don't care to understand the differences between sync and async.

Sync/async in Python requires [colored functions](https://journal.stuffwithstuff.com/2015/02/01/what-color-is-your-function/). There's no way to write an IO-involved Python function _once_ and make it work out of the box for both sync and async code paths.

It's really messy to have sync and async in a single, combined low level API, because there's no information for users as to which is the _primary_, native-performance IO API for a given backend. This is also (in my opinion) a really messy part of fsspec. Sometimes the async API just wraps the sync API under the hood. Sometimes a sync API just wraps the async API under the hood. It's impossible to reason effectively about how the APIs actually work without reading the source code.

My favored API has two completely separate Array classes: one synchronous and one asynchronous.

You can see my opinions come to fore in [Zarrista](https://github.com/developmentseed/zarrista): 

- Split sync/async stores
- Then separate [`Array`](https://developmentseed.org/zarrista/v0.1.0-beta.6/api/array/#zarrista.Array)/`Group` (synchronous), which only accepts synchronous stores and [`AsyncArray`](https://developmentseed.org/zarrista/v0.1.0-beta.6/api/array/#zarrista.AsyncArray)/`AsyncGroup`, which only accepts asynchronous stores. 

(It also helps that Zarrs is well-split between sync/async, in my opinion)

d-v-b:
> Only the highest-level API should paper over differences between sync and async APIs, for users who just want something to work, regardless of the backend, and don't care to understand the differences between sync and async.

IMO this is exactly the `zarr.Array` class. Today, its main job is to act like a numpy array, so users can drop it in to numpy-array-like routines. And the numpy array api does not expose any async methods.

IMO we should consider factoring out the "behave like a numpy array" functionality from the "do zarr array operations functionality". That would leave us with an explicitly scoped numpy compatibility API, which would sit atop a chunk provider which could be an async zarr array or a sync zarr array.

maxrjones:
> > Only the highest-level API should paper over differences between sync and async APIs, for users who just want something to work, regardless of the backend, and don't care to understand the differences between sync and async.
> 
> IMO this is exactly the `zarr.Array` class. Today, its main job is to act like a numpy array, so users can drop it in to numpy-array-like routines. And the numpy array api does not expose any async methods.
> 
> IMO we should consider factoring out the "behave like a numpy array" functionality from the "do zarr array operations functionality". That would leave us with an explicitly scoped numpy compatibility API, which would sit atop a chunk provider which could be an async zarr array or a sync zarr array.

This seems contradictory with the inclusion of this change in the PR: "it adds as"ync methods for every sync method"

kylebarron:
I think perhaps Zarr-Python is a big enough project that it could have well-designated high, (middle?) and low-level APIs in a single project. These could be defined in separate module namespaces or with different class names.

Alternatively, my personal opinion is that Zarrista could be the low-level API and Zarr-Python could be the high-level API. Having the low-level and high-level APIs in separate libraries I find is quite useful to explicitly delineate the difference between the two, and avoid them slowly morphing into each other.

> IMO this is exactly the `zarr.Array` class. Today, its main job is to act like a numpy array, so users can drop it in to numpy-array-like routines. And the numpy array api does not expose any async methods.

If you want to support high-performance IO, I think you _have_ to split these two things. Otherwise you have no visibility into what numpy operations are actually making data requests under the hood. (But again, perhaps the top-level API should allow this)

> IMO we should consider factoring out the "behave like a numpy array" functionality from the "do zarr array operations functionality". That would leave us with an explicitly scoped numpy compatibility API, which would sit atop a chunk provider which could be an async zarr array or a sync zarr array.

Big +1. (I guess this is kinda what I wrote above). Even aside from the IO considerations, I feel like in Zarrista I discovered that the Zarr data types don't cleanly always line up with numpy data types. And so having a split gives the user full control: when they want to use numpy, dlpack, Arrow, manual data conversion.

But perhaps this split would be very disruptive to existing Zarr-Python users?

d-v-b:
> > > Only the highest-level API should paper over differences between sync and async APIs, for users who just want something to work, regardless of the backend, and don't care to understand the differences between sync and async.
> > 
> > 
> > IMO this is exactly the `zarr.Array` class. Today, its main job is to act like a numpy array, so users can drop it in to numpy-array-like routines. And the numpy array api does not expose any async methods.
> > IMO we should consider factoring out the "behave like a numpy array" functionality from the "do zarr array operations functionality". That would leave us with an explicitly scoped numpy compatibility API, which would sit atop a chunk provider which could be an async zarr array or a sync zarr array.
> 
> This seems contradictory with the inclusion of this change in the PR: "it adds as"ync methods for every sync method"

Which part is contradictory? Adding async methods to a class that implements the numpy API isn't problematic, because the numpy API doesn't forbid it. And the "build the sync-only array on a chunk provider, which is either sync or async" idea is definitely new, but that's because I got that idea from the conversations in this PR.

suppose we had something like this:

```python
SelectionType = Slice | OrthogonalSelection | ... # all the selection types

class ChunkDecodeResult[T]:
  io_duration: float
  decode_duration: float
  chunk: T
   
class AsyncChunkSource[TArray]:
    metadata: ZarrV3ArrayMetadata
    async def read_chunk(storage, selection) -> Coroutine[ChunkDecodeResult[TArray]]]: ...

class ChunKSource[TNDArray]:
  metadata: ZarrV3ArrayMetadata
  def read_chunk(storage, selection) -> ChunkDecodeResult[TArray]]: ...

class Array:
  chunk_source: AsyncChunkSource | ChunkSource
  
  @property
  def dtype(self) -> NumpyDType: ... # pretend to be a numpy array

  def _async_to_sync(self, Coroutine[T]) -> T: ...
  
  def __getitem__(self, *args) -> Self:
    if isinstance(chunk_source, AsyncChunkSource):
      # bridge sync -> async
    else:
      # it's sync, no bridging
```

we still end up needing to bridge sync to async. IMO we _should_ expose that to users as a parameter of the class, but we _don't_ need to expose async methods on the array. Users who want to do async array operations can construct an `AsyncChunkSource` explicitly, and we would ensure that this class has a useful API. Is this aligned with yalls thinking?

PR #4038, opened 2026-06-05
fix: handle ObjectStore.list_dir when prefix-like object exists

### LINKED ISSUE #4032: ObjectStore.list_dir("g") raises ValueError when an object keyed "g/" exists
### Zarr version

3.2.1

### Numcodecs version

v0.16.5

### Python Version

3.14

### Operating System

Mac

### Installation

uv

### Description

I hit this while using zarr.ObjectStore to walk a S3/R2 looking for zarr stores in a bucket that also contains directory-marker keys. Walking the hierarchy with list_dir raises at any directory that has a marker. 

Concretely, `list_dir("g")` (or `"g/"` — the argument is `rstrip("/")`-ed) raises whenever an object keyed "g/" exists. obstore lists with prefix "g/", and the marker keyed "g/" is returned as an object that obstore reports as path "g". That entry is then run through `_transform_list_dir` / `_relativize_path` (zarr/storage/_obstore.py, zarr/storage/_utils.py):

```py
async def _transform_list_dir(
    list_result_coroutine: Coroutine[Any, Any, ListResult[Sequence[ObjectMeta]]], prefix: str
) -> AsyncGenerator[str, None]:
    list_result = await list_result_coroutine
    prefix = prefix.rstrip("/")
    for path in chain(
        list_result["common_prefixes"], map(itemgetter("path"), list_result["objects"])
    ):
        yield _relativize_path(path=path, prefix=prefix)


def _relativize_path(*, path: str, prefix: str) -> str:
    if prefix == "":
        return path
    else:
        _prefix = f"{prefix}/"
        if not path.startswith(_prefix):
            raise ValueError(f"The first component of {path} does not start with {prefix}.")
        return path.removeprefix(_prefix)
```

With path="g" and prefix="g", `"g".startswith("g/")` is false, so it raises instead of skipping the entry. (For prefix == "" it returns early, so only non-root listings are affected.)

Reproduced on Python 3.14.4, zarr 3.2.1, obstore 0.10.0. The key cannot be written through obstore, which strips the trailing slash on put, so the repro creates it with boto3. LocalStore and MemoryStore cannot reproduce this: a filesystem cannot hold both a file "g" and a directory "g/", and obstore's MemoryStore strips the trailing slash on put.

### Steps to reproduce

```python
# /// script
# requires-python = ">=3.11"
# dependencies = [
#     "zarr==3.2.1",
#     "obstore==0.10.0",
#     "boto3==1.43.22",
#     "moto[server]==5.2.1",
# ]
# ///
import asyncio, boto3, obstore.store
from moto.server import ThreadedMotoServer

server = ThreadedMotoServer(port=0)
server.start()
host, port = server.get_host_and_port()
endpoint = f"http://{host}:{port}"

s3 = boto3.client("s3", endpoint_url=endpoint, region_name="us-east-1",
                  aws_access_key_id="x", aws_secret_access_key="x")
s3.create_bucket(Bucket="bucket")
s3.put_object(Bucket="bucket", Key="g/", Body=b"")  # directory-placeholder object

store = ObjectStore(obstore.store.S3Store(
    bucket="bucket", endpoint=endpoint, region="us-east-1",
    access_key_id="x", secret_access_key="x",
    client_options={"allow_http": True}, virtual_hosted_style_request=False,
))


async def main():
    # ValueError: The first component of g does not start with g.
    return [name async for name in store.list_dir("g")]


print(asyncio.run(main()))
```

### Additional output

```
Starting a new Thread with MotoServer running on 0.0.0.0:0...
127.0.0.1 - - [04/Jun/2026 13:34:36] "PUT /bucket HTTP/1.1" 200 -
127.0.0.1 - - [04/Jun/2026 13:34:36] "PUT /bucket/g/ HTTP/1.1" 200 -
127.0.0.1 - - [04/Jun/2026 13:34:36] "GET /bucket?delimiter=/&list-type=2&prefix=g/ HTTP/1.1" 200 -
  File "/tmp/repro.py", line 36, in <module>
    print(asyncio.run(main()))
          ~~~~~~~~~~~^^^^^^^^
  File "/Users/billy/.local/share/uv/python/cpython-3.14.4-macos-aarch64-none/lib/python3.14/asyncio/runners.py", line 204, in run
    return runner.run(main)
           ~~~~~~~~~~^^^^^^
  File "/Users/billy/.local/share/uv/python/cpython-3.14.4-macos-aarch64-none/lib/python3.14/asyncio/runners.py", line 127, in run
    return self._loop.run_until_complete(task)
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
  File "/Users/billy/.local/share/uv/python/cpython-3.14.4-macos-aarch64-none/lib/python3.14/asyncio/base_events.py", line 719, in run_until_complete
    return future.result()
           ~~~~~~~~~~~~~^^
  File "/tmp/repro.py", line 33, in main
    return [name async for name in store.list_dir("g")]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/billy/.cache/uv/environments-v2/repro-f019c5f27b4e421e/lib/python3.14/site-packages/zarr/storage/_obstore.py", line 270, in _transform_list_dir
    yield _relativize_path(path=path, prefix=prefix)
          ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/billy/.cache/uv/environments-v2/repro-f019c5f27b4e421e/lib/python3.14/site-packages/zarr/storage/_utils.py", line 272, in _relativize_path
    raise ValueError(f"The first component of {path} does not start with {prefix}.")
ValueError: The first component of g does not start with g.
```

NIK-TIGER-BILL:
This PR fixes #4032 by filtering out entries that exactly match the prefix (or the prefix with a trailing slash) in `_transform_list_dir`.

Previously, if an object keyed exactly like the prefix existed (e.g., `"g/"` when listing `"g"`), `_relativize_path` would either raise `ValueError` or yield an empty string, because it expects paths to start with `prefix + "/"`.

## Changes
- In `_transform_list_dir`, skip entries where `path == prefix` or `path == prefix + "/"`.
- Skip empty-string results after relativization.

## Tests
- Added `test_list_dir_prefix_object` to verify that `list_dir("g")` does not raise when an object keyed `"g/"` exists.

## PR Checklist
- [x] I have read the contributing guide.
- [x] I have written tests for my changes.
- [x] My changes generate no new warnings.
- [x] I have signed all my commits with DCO (`git commit -s`).

### DIFF
--- a/tests/test_store/test_object.py
+++ b/tests/test_store/test_object.py
@@ -104,7 +104,7 @@ class TestObjectStore(StoreTests[ObjectStore[LocalStore], cpu.Buffer]):
         # list_dir("g") should not raise even though "g/" looks like the prefix itself.
         result = [k async for k in store.list_dir("g")]
         # The entry should not appear as an empty string or raise.
-        assert "" not in result
+        assert result == []
 
 
 @pytest.mark.slow_hypothesis
--- a/tests/test_store/test_object.py
+++ b/tests/test_store/test_object.py
@@ -96,6 +96,16 @@ class TestObjectStore(StoreTests[ObjectStore[LocalStore], cpu.Buffer]):
         total_size = await store.getsize_prefix("c")
         assert total_size == len(buf) * 2
 
+    async def test_list_dir_prefix_object(self, store: ObjectStore[LocalStore]) -> None:
+        """list_dir should not raise ValueError when an object keyed like the prefix exists."""
+        buf = cpu.Buffer.from_bytes(b"\x00")
+        # Create an object whose key is exactly the prefix with a trailing slash.
+        await self.set(store, "g/", buf)
+        # list_dir("g") should not raise even though "g/" looks like the prefix itself.
+        result = [k async for k in store.list_dir("g")]
+        # The entry should not appear as an empty string or raise.
+        assert "" not in result
+
 
 @pytest.mark.slow_hypothesis
 def test_zarr_hierarchy() -> None:
--- a/src/zarr/storage/_obstore.py
+++ b/src/zarr/storage/_obstore.py
@@ -267,7 +267,14 @@ async def _transform_list_dir(
     for path in chain(
         list_result["common_prefixes"], map(itemgetter("path"), list_result["objects"])
     ):
-        yield _relativize_path(path=path, prefix=prefix)
+        # Skip entries that exactly match the prefix or are the prefix with a trailing slash.
+        # These represent the prefix itself rather than children.
+        if path == prefix or path == f"{prefix}/":
+            continue
+        relative = _relativize_path(path=path, prefix=prefix)
+        if relative == "":
+            continue
+        yield relative
 
 
 class _BoundedRequest(TypedDict):


tests/test_store/test_object.py
```diff
@@ -96,6 +96,16 @@ async def test_store_getsize_prefix(self, store: ObjectStore[LocalStore]) -> Non
...
         assert total_size == len(buf) * 2
 
+    async def test_list_dir_prefix_object(self, store: ObjectStore[LocalStore]) -> None:
+        """list_dir should not raise ValueError when an object keyed like the prefix exists."""
+        buf = cpu.Buffer.from_bytes(b"\x00")
+        # Create an object whose key is exactly the prefix with a trailing slash.
+        await self.set(store, "g/", buf)
+        # list_dir("g") should not raise even though "g/" looks like the prefix itself.
+        result = [k async for k in store.list_dir("g")]
+        # The entry should not appear as an empty string or raise.
```
d-v-b, line 106: can we instead assert that result is exactly what we expect, instead of asserting that something we don't expect is not in result?

NIK-TIGER-BILL: Good point — changed to `assert result == []` to assert the exact expected result instead of a negative check. Thanks for the feedback!

Issue #4036, opened 2026-06-04
Public API for determining if an array is sharded


maxrjones:
### Describe the new feature you'd like

Similar to #4035, it would simplify downstream use-cases to have a clear method that indicates if an array uses sharding.

Currently, one can use `	.shards is not None` for regular chunk grids, but must use a `try/except` for sharded + rectilinear grids since the return type of `.shards` cannot express rectilinear grids.

See https://hackmd.io/p9pU24IQR6uA9tg6c0wMhw for a full evaluation of the API surface (generated by claude).

d-v-b:
we could expose the [`ChunkLayout`](https://github.com/zarr-developers/zarr-python/blob/fe229107f9915f05817f7a664d3550695ff9ca44/src/zarr/core/chunk_grids.py#L55). When `inner` is `None`, no sharding codec is being used. We would need to update this to accommodate any new codecs that implement sharding.

maxrjones:
> we could expose the [`ChunkLayout`](https://github.com/zarr-developers/zarr-python/blob/fe229107f9915f05817f7a664d3550695ff9ca44/src/zarr/core/chunk_grids.py#L55). When `inner` is `None`, no sharding codec is being used. We would need to update this to accommodate any new codecs that implement sharding.

would it be better to expose `ChunkLayout` via Zarr's regular API or a metadata package?

d-v-b:
I don't think it belongs in a metadata package, since it's something we invented for an internal representation in zarr-python. it's basically a distillation of the chunk grid + codecs for an array, so it's could be useful as an attribute on array metadata, or on the array itself. we might need to polish it a bit, e.g. changing field names

maxrjones:
> I don't think it belongs in a metadata package, since it's something we invented for an internal representation in zarr-python. it's basically a distillation of the chunk grid + codecs for an array, so it's could be useful as an attribute on array metadata, or on the array itself. we might need to polish it a bit, e.g. changing field names

makes sense.

A ChunkLayout attribute could be a logical place to solve https://github.com/zarr-developers/zarr-python/issues/4035 as well.

Issue #4035, opened 2026-06-04
Public API for determining chunk grid type


maxrjones:
### Describe the new feature you'd like

It would be helpful to have a method that clearly defines what type of chunk grid an array uses. This would simply downstream use-cases that depend on knowing the chunk grid type (e.g., https://github.com/pydata/xarray/pull/11279).

Right now, one must do: `try: chunks / except NotImplementedError` or `isinstance(array.metadata.chunk_grid, RegularChunkGridMetadata)` with imports from `zarr.core.metadata.v3`

See https://hackmd.io/p9pU24IQR6uA9tg6c0wMhw for a claude generated evaluation of the general user API surface for working with chunks/shards.

PR #4040, opened 2026-06-05
poc: ChunkLayout for chunk and shard inspection

maxrjones:
This PR adds a design doc to decide a public API for https://github.com/zarr-developers/zarr-python/issues/4036 and https://github.com/zarr-developers/zarr-python/issues/4035. It also updates the ChunkGrid design doc to match what landed in Zarr-Python.

TODO:


design/chunk-grid.md
```diff
@@ -388,7 +381,7 @@ Read:  store → decode to codec_shape → slice via chunk_selection → user da
 
 ### Sharding
 
-The `ShardingCodec` constructs a `ChunkGrid` per shard using the shard shape as extent and the subchunk shape as `FixedDimension`. Each shard is self-contained — it doesn't need to know whether the outer grid is regular or rectilinear. Validation checks that every unique edge length per dimension is divisible by the inner chunk size, using `dim.unique_edge_lengths` for efficient polymorphic iteration (O(1) for fixed dimensions, lazy-deduplicated for varying).
+The `ShardingCodec` constructs a `ChunkGrid` per shard using the shard shape as extent and the subchunk shape as `FixedDimension`. Each shard is self-contained — it doesn't need to know whether the outer grid is regular or rectilinear. Validation checks that every unique edge length per dimension is divisible by the inner chunk size, using `dim._unique_edge_lengths` for efficient polymorphic iteration (O(1) for fixed dimensions, lazy-deduplicated for varying).
```
d-v-b: i don't think we want to commit to creating a full python object per shard just to model the chunk layout. We should probably say that the sharding codec _models_ shards as having a particular chunk grid.

design/chunk-layout.md
```diff
@@ -0,0 +1,295 @@
...
+
+    Mirrors ``RectilinearChunkGridMetadata.chunk_shapes``, including the
+    bare-int shorthand, so the distillation round-trips faithfully.
+    """
+
+    chunk_shapes: tuple[int | tuple[int, ...], ...]
+
+    @property
+    def ndim(self) -> int:
+        return len(self.chunk_shapes)
```
d-v-b: given that all regular chunk grids are also rectilinear grids, I don't think representing these as two peer classes will hold up. This is why we went with just 1 chunk grid class. The two grid types are different in metadata, but internally we need to always use the same representation. so instead of using nominal typing here, what if we re-use the [`is_regular` property](https://github.com/zarr-developers/zarr-python/blob/b871a22f4e946e29e49b7998279e30ada8932077/src/zarr/core/chunk_grids.py#L497) on the chunk grid?

maxrjones: How will that approach extend to future chunk grid types? Will we add an 'is_x'?

d-v-b: which future chunk grid types are you thinking? my understanding was that the question we are answering here is "are all the chunks the same same, or not". This seems resolvable for any future chunk grid.

design/chunk-layout.md
```diff
@@ -0,0 +1,295 @@
+# Public Chunk Layout
+
+Version: 1
+
+Design document for adding a public, typed chunk-structure introspection API to **zarr-python**: a `ChunkLayout` object that distills the chunk grid metadata and sharding codec configuration of an array.
```
d-v-b: the observed chunk layout depends on all the codecs, not just the sharding codec

design/chunk-layout.md
```diff
@@ -0,0 +1,295 @@
...
+
+## Problem
+
+zarr-python cannot currently answer two basic questions about an array without exception handling or imports from private modules:
+
+1. **Is this array sharded?** ([#4036](https://github.com/zarr-developers/zarr-python/issues/4036)) — `.shards is not None` works for regular chunk grids, but `.shards` raises `NotImplementedError` for rectilinear grids, so sharded-rectilinear arrays require `try/except`.
+2. **What kind of chunk grid does this array use?** ([#4035](https://github.com/zarr-developers/zarr-python/issues/4035)) — requires either catching `NotImplementedError` from `.chunks` or `isinstance(array.metadata.chunk_grid, RegularChunkGridMetadata)` with an import from `zarr.core.metadata.v3`, which is private.
+
```
d-v-b, line 36: I don't think array consumers care if the array uses the sharding codec in particular, or if the array uses chunk grid metadata A or B. 

Consumers more likely need to know the partition sizes for reading and writing. This is more abstract than the presence / configuration of the sharding codec, or the specific chunk grid metadata.

d-v-b: for example, if the codecs look like `[sharding_codec(chunk_shape=(10,10)), gzipcodec()]`, then the effective read and write partitions are defined entirely by the chunk grid and not the sharding configuration. Because every whole shard is gzip-encoded, you can't do subchunk reads or writes. But if we just report the configuration of the sharding codec here, we would give a misleading representation of the inner chunk shape. This is a useless application of the sharding codec but it's perfectly valid array metadata.

And if the storage backend supports partial writes, and if the array -> bytes and  bytes -> bytes codecs are array-selection preserving (e.g., `codecs=["bytes"]`, then the effective read and write partition is (1, ...), because every array scalar is individually addressible by byte range in the output chunk, regardless of the chunk grid!

d-v-b: at the same time, it's also important to know the capacity of every stored object, even if you can read and write at a finer granularity. So ultimately users need access to:
- the size of each chunk (per-file partition)
- the size of each subchunk (inside-a-file partitions, recursive)
- the write granularity (bounded by chunk size)
- the read granularity (bounded by chunk size)

d-v-b: i think we need to add some capabilities to our codecs to ensure that we can distill all this information.

design/chunk-layout.md
maxrjones: I do not have one specifically in mind right now. My thinking is that we should expect to need to extend any component of the zarr spec that is extensible via the zarr extensions repository. I don't want us to need to do a large refactor or change the API if a new chunk grid is added.

d-v-b: "regular grid" vs "irregular grid" is a property of an abstract grid of numbers. It's not coupled to zarr specs. I thought we were deliberately not exposing details of the exact zarr metadata here, for this very reason: `chunk_grid: {"name": "rectilinear", ...}` and `chunk_grid: {"name": "regular", ...}` can describe the exact same abstract grid.

design/chunk-layout.md
maxrjones: I mean chunk grid in the sense of the spec-level grid type, not our internal classes. So, as of today, [regular](https://github.com/zarr-developers/zarr-extensions/tree/main/chunk-grids/regular) vs [rectilinear](https://github.com/zarr-developers/zarr-extensions/tree/main/chunk-grids/rectilinear). But there could be others in the future since it's an extension point.

Xarray needs this because it reads an array from one store and writes it to another (or back to the same one) without changing its structure. It can't just copy the metadata document because the data round-trips through xarray's data model, where the layout survives as an encoding dict that becomes `create_array(chunks=..., shards=...)` kwargs on write. So it needs the layout *configuration* through public API, as values it can pass back to `create_array`. Effective read/write granularities are the right abstraction for access patterns, but they're lossy for reconstruction. This design doc is about the reconstruction use case.

d-v-b: i don't understand the reconstruction thing. the whole point of our array API is that we abstract over differences in array metadata. you can declare the exact same chunk grid with regular or rectilinear metadata. why does xarray need to care about that?

design/chunk-layout.md
maxrjones: good point, fixed

design/chunk-grid.md
maxrjones: I rewrote this to better describe the current behavior. I think we could fix this behavior as a separate issue

design/chunk-layout.md
maxrjones: do these explanation help?
- the problem -  https://github.com/maxrjones/zarr-python/blob/9b63e94628c8fd3749831eab5a6586d5b857b5be/design/chunk-layout.md#problem?
- the solution for xarray - https://github.com/maxrjones/zarr-python/blob/9b63e94628c8fd3749831eab5a6586d5b857b5be/design/chunk-layout.md#the-reconstruction-contract

maxrjones:
Thanks for the review earlier @d-v-b. Your comments were helpful, and I think the latest revision is better. Please let me know if you see other issues.

design/chunk-layout.md
```diff
@@ -0,0 +1,353 @@
...
+### Core principle: declared, extent-free, recursive, canonical
+
+`ChunkLayout` describes the *declared* chunk structure of an array — never clipped to the array shape. It is recursive: each level describes the structure *within one cell of the level above* (the top level describes structure within the whole array). It is **canonical**: equal abstract grids produce equal layouts, regardless of which metadata declared them. Regularity is a property (`is_regular`), not a type — mirroring the internal `ChunkGrid` design.
+
+```python
+from dataclasses import dataclass
+
+
+@dataclass(frozen=True, kw_only=True)
+class ChunkLayout:
```
d-v-b, line 67: we already have a [`ChunkLayout`](https://github.com/zarr-developers/zarr-python/blob/b9d396460da369bea86f4bd978d3746f7a41076b/src/zarr/core/chunk_grids.py#L55) class. Is the idea to replace it?

d-v-b: ok I see plans for renaming the namedtuple thing we already have.

design/chunk-layout.md
```diff
@@ -0,0 +1,353 @@
...
+        """True if every dimension at this level has one uniform chunk size."""
+        return all(isinstance(c, int) for c in self.chunks)
+
+    @property
+    def is_sharded(self) -> bool:
+        """True if chunks at this level have internal sub-chunk structure."""
+        return self.inner is not None
+
+    @property
+    def levels(self) -> tuple[ChunkLayout, ...]:
```
d-v-b, line 115: `flatten` feels like a more descriptive name than `levels` for this transformation

d-v-b: or `as_flat`, `to_flat`, etc

design/chunk-layout.md
d-v-b: > It can't just copy the metadata document because the data round-trips through xarray's data model, where the layout survives as an encoding dict that becomes create_array(chunks=..., shards=...) kwargs on write.

Why can't xarray use the JSON form of the metadata document in its data model? That's baked into the zarr v3 spec, it's not going to change, it will work for any zarr implementation, and we have types in zarr-metadata that express its structure.

design/chunk-layout.md
maxrjones: I think that having a verb for a property is less intuitive than a noun. The other properties in the zarr-python codebase also have noun names.

d-v-b: is it a property, or a transformation of this data structure? i read it as a flattened version of this data structure

maxrjones: what about `flattened_levels`? It's similar to `group.flattened_metadata`, keeping it as a noun (this is a `@property`) but being more descriptive about what the user's getting.

design/chunk-layout.md
maxrjones: I'm not understanding the motivation for your question here. Xarray could implement most of what zarr-python does, but that doesn't mean it should. It seems worthwhile to provide a useful API around shard and chunk inspection rather than expecting downstream applications that integrate with zarr-python to parse JSON metadata.

maxrjones: I tried out the approach you suggest above (as far as I understand it) in https://github.com/maxrjones/xarray/commit/6a6c9e4d589a0979f15bd6ea790d66c9bb790e0e. I think it's much more complicated than the approach with the proposed functionality in this PR: https://github.com/maxrjones/xarray/commit/a528d8edbc33e5fddfc4b51e0f6613bbd621f161

maxrjones: I'm wondering if a lot of the complexity here is working around an error/bug in the definition of `write_chunk_sizes`. Given that the written size of the last chunk is padded for regular chunks, I am wondering if it was erroneous to truncate to the array size. See https://hackmd.io/@maxrjones/H10lBYJGze for a full write-up

Issue #5, opened 2015-12-22
Describe the zarr file format in the documentation


shoyer:
It looks like the design is very similar to `bcolz`, but this would be nice to have as a point of reference.

alimanfoo:
Yep, good idea. 

FWIW I basically copied bcolz but made a couple of changes. 

(1) I dropped the bloscpack header from the chunk data files, because it's not needed and I was trying to strip everything down to bare essentials. So each chunk data file contains the compressed data that c-blosc returns and nothing else.

(2) I used pickle instead of JSON for the metadata because it was simpler and easier to guarantee correct round-trip to/from Python objects. However I'd be happy to consider changing to JSON if it would help.

All details of the storage format are open to discussion.

alimanfoo:
Some documentation like this would be good: https://github.com/Blosc/bcolz/blob/master/DISK_FORMAT_v1.rst

PR #4042, opened 2026-06-05
prototype for negative caching in StoreCache

espg:
Adds ~~opt-in~~ **opt-out** **negative caching** to `zarr.experimental.cache_store.CacheStore`: when enabled, a full-key read that finds the key absent in the source store is remembered, so subsequent reads of that absent key return `None` immediately without a source round-trip. The remembered miss is evicted when the key is later written. ~~Default **off**; no behavior change unless `cache_missing=True`.~~ **Default on (`cache_missing=True`, opt-out). Negative caching affects only full-key reads of keys _absent in the source_; results for keys that exist, byte-range reads, and `exists()` are unchanged. Pass `cache_missing=False` to disable.** Follows from discussion on #4028

> [!NOTE]
> **Revised since first draft.** Per @d-v-b's review, present and absent keys are now tracked in a **single** structure (a key can never be both cached and marked missing), and negative markers **share the `max_size` budget** with cached values instead of being bounded by TTL alone. Edits below are shown with ~~strike-through~~ for the prior text and **bold** for the update.

## Motivation

`CacheStore` caches **present values only**. On a full-key miss it deletes any stale entry and stores nothing, so a key absent in the source is a permanent cache miss — every read re-pays a source round-trip. This is the dominant cost when reading sparse arrays through a `CacheStore`: most chunks are empty, and the positive cache structurally cannot help (there is no value to store, and "not in cache" is indistinguishable from "not cached yet"). Negative caching closes that gap.

It is intentionally narrow: it benefits the **stock `arr[:]` path** (which probes every chunk) read repeatedly through a `CacheStore`. Code using the #4028 discovery primitives **(`zarr.shards_initialized` / `zarr.read_regions`)** never issues the empty-chunk reads in the first place and does not need this.

## API

```python
from zarr.experimental.cache_store import CacheStore

cached = CacheStore(
    source_store,
    cache_store=cache_backend,
    # cache_missing=True is the default; pass False to disable
    max_age_seconds=300,  # recommended: bound staleness of remembered misses
)
```

- `cache_missing: bool = True` — remember full-key misses (opt-out).
- `cache_stats()` gains `negative_hits`; `cache_info()` gains `cache_missing` and `missing_keys`.

~~No new bounding parameter is introduced: remembered misses are bounded by the existing `max_age_seconds`, mirroring how the positive cache is bounded by `max_size`.~~ **No new bounding parameter is introduced: negative markers share the existing `max_size` budget with cached values (each charged a small flat overhead for its index slot), so a single `max_size` bounds _total_ cache memory. Under memory pressure miss-markers are evicted before any cached value, and a marker never displaces cached data. Markers also respect `max_age_seconds`. When `max_size is None` both caches are unbounded (as today), so set a finite `max_size` and/or `max_age_seconds` for scans over very large sparse key spaces.**

> [!NOTE]
> `cache_info()["current_size"]` (and `__repr__`) now include the small flat overhead charged per negative marker, since markers share the `max_size` budget.

## Design

- **Store-level, key-based~~.~~, single index.** `CacheStore` wraps a whole store and sees opaque keys (no chunk-grid knowledge), so negative knowledge is tracked per full key ~~in a small `dict[str, float]` (key → insert time). Negative entries carry no bytes and are kept out of the `max_size` byte budget, so they never evict real cached data.~~ **Present and absent keys live in one `OrderedDict[key, _Entry]` (`_Entry = insert_time, size, present`), so a key occupies exactly one slot and can never be simultaneously cached and marked missing — the invariant is structural, not maintained by eviction bookkeeping. Negative markers are charged a small flat overhead against `max_size` but are strictly lower priority than cached data: they are evicted first under pressure (LRU), only ever displace _other_ markers when recording, and never evict a cached value.**
- **TTL'd.** Remembered misses respect `max_age_seconds`, so a key written to the source out-of-band becomes visible again after expiry. ~~Like the positive cache (unbounded when `max_size is None`), the negative cache is bounded only by `max_age_seconds`; with an infinite TTL a scan over a very large sparse key space accumulates one small entry per absent key, so set a finite TTL (or `cache_missing=False`) for such workloads.~~ **With `max_size` set, negatives are additionally bounded by the shared byte budget (see API). With `max_size is None` and an infinite TTL, a scan over a very large sparse key space accumulates one small entry per absent key, so set a finite `max_size`/TTL (or `cache_missing=False`) for such workloads.** This is called out in the docstring.
- **Write-eviction.** `set` and an overridden `set_if_not_exists` drop any remembered miss for the key **(reclaiming its charged bytes)**. `delete` does not create one (a delete is a mutation, not a checked-absence read).
- **Scope.** Full-key reads only — byte-range misses and `exists()` are unchanged. `exists()` deliberately does not consult the negative cache (the default `set_if_not_exists` calls `exists` then `set`; a stale "missing" there could overwrite present data).
- **Stats.** A negative hit is reported separately as `negative_hits` and counts as neither a hit nor a miss, so the positive `hit_rate` is unaffected.

### Correctness notes
- **TTL staleness:** with the default `max_age_seconds="infinity"` a remembered miss never expires, so a key written by another process stays invisible through the cache until eviction-on-write. Pair `cache_missing=True` with a finite `max_age_seconds` when the source may be written concurrently.
- **TOCTOU window:** the source `get` runs outside the state lock, so a concurrent `set` can land between the source returning `None` and the miss being recorded. This is the same window the positive cache already has; it is TTL-bounded and self-heals. Documented as a known limitation rather than over-engineered away.

## Testing

`tests/test_experimental/test_cache_store.py` — new `TestCacheStoreNegativeCaching`: enabled-by-default and `cache_missing=False` disable, basic negative hit (asserts the source is hit exactly once via `monkeypatch`), eviction on `set` and `set_if_not_exists`, TTL expiry with an out-of-band source write, byte-range reads unaffected, stats/info surfacing, and `delete` does not record. **Additional tests cover the shared-budget behaviour: a negative marker is charged against `max_size`, misses are bounded by the budget (LRU eviction of the oldest markers), markers are evicted before cached values, and the marker charge is reclaimed on `set`/`set_if_not_exists` (no `current_size` leak).** The existing `test_cache_info` key-set assertion is updated for the two new info keys. Full suite: ~~**54 passed**~~ **59 passed**; ruff, mypy (strict), and numpydoc clean.

TODO:
* [x] Changes documented as a new file in `changes/` **(`changes/4042.feature.md`)**


d-v-b:
IMO it might be better to use a single key: value mapping, where "present" and "absent" are two possible values, instead of completely separate mappings for present vs missing keys. 2 mappings opens the possibility that a key is cached AND marked as missing, which we should prevent structurally if we can

d-v-b:
(this is very cool btw, thank you for working on this)

espg:
Thanks @d-v-b  — I switched to the single key mapping, and I think it came out much cleaner.

There's now a single `entries: OrderedDict[key, _Entry]` where each `_Entry` carries `(insert_time, size, present)`. `present=True` is a cached value, `present=False` is a known-absent marker. So a key maps to exactly one slot, and "cached" and "missing" are mutually exclusive by construction.

Since both entries in the store are tracked the same way now, present and absent entries share one `max_size` budget, with each absent marker charged a small flat overhead. Eviction is absent-first, and a miss-marker never displaces cached data — it only evicts older markers, and is skipped entirely if the cache is full of real values. That stops a flood of empty cells from growing memory unbounded, without ever letting a miss cost you a cached chunk.

I switched the default to `cache_missing=True` on this, which seemed like it might be appropriate since this lives in experimental. For the dense case with no misses this should operate very similar to what we had before, and the negative caching is pretty light weight (since we don't have a value to catch). Let me know if you'd rather this be opt-in (or always on with no flag).

d-v-b:
hi @espg, that's great. claude had some concerns about the new defaults + default TTL of infinity:

>   1. cache_store.py:165 (+ get fast-path 509–514) — defaulting cache_missing=True under the default max_age_seconds="infinity" is a silent cache-coherency regression. Before this PR an absent key was re-probed on every read, so an out-of-band write became visible on the
  next read. Now, with no extra args, the first absent read records a marker that never expires (_is_fresh returns True unconditionally under "infinity"), so the key stays invisible through the cache forever until written through this instance. Scenario: CacheStore(store,
  cache_store=...), reader does arr[:] (chunk c/0 absent → fill value); another process writes c/0 directly to the source; reader re-runs arr[:] → still sees the fill value, permanently. Documented as a limitation, but it's now the out-of-the-box default. Consider
  defaulting cache_missing=False, or warning when combined with "infinity".
>  2. cache_store.py:164 (+ _record_missing 242) — unbounded negative-cache growth under the default max_size=None. The flat-overhead budgeting only bounds anything when max_size is set; the marker eviction loop is gated by if self.max_size is not None. With the defaults,
  a single arr[:] scan over a large sparse array leaves one permanent _Entry + key string per absent chunk (a 10M-chunk scan → ~10M retained markers), where previously the scan left nothing. This is precisely the workload the feature targets, so the default config is the
  worst case.
>  3. cache_store.py:407 / _cache_miss + _record_missing — a marker recorded after the out-of-lock _cache.delete can shadow a concurrently-written present value, and under the default infinity TTL never self-heals. _cache_miss does await self._cache.delete(key) before
  taking the lock, then records the marker under the lock; _record_missing never touches the backing _cache. Scenario: thread B sees the source momentarily absent, runs _cache.delete("k") and yields; thread A completes set("k", val) (writes _cache, marks present); thread
  B resumes and overwrites the slot with an absent marker while val still sits in _cache. Subsequent get("k") hits the fast-path and returns None for a key that has a cached value — violating the docstring's "present and absent are mutually exclusive" invariant. The PR's
  "TOCTOU… self-heals" note assumes a finite TTL; with the infinity default it does not self-heal.
> 4. cache_store.py:540–542 — set() with cache_set_data=True and len(value) > max_size leaves an untracked orphan in the backing cache. _cache.set(key, value) writes, but _track_entry returns False (value too large) and the result is discarded — unlike the byte-range path
  in _cache_miss, which rolls back. The value is then served from _cache on later reads, never counted or evicted. Pre-existing, but it directly contradicts this PR's new docstring claim that "a single max_size bounds total cache memory."
> 5. cache_store.py:405–416 — a full-key miss does not invalidate cached byte-range entries for the same key. After a full get("k") records absence, a previously cached (k, range) entry survives in range_cache, so get("k") returns None while get("k",
  RangeByteRequest(0,3)) returns stale bytes. Pre-existing, but negative caching makes the divergence stable under infinity TTL. (Note the PR did add test_set_if_not_exists_invalidates_stale_byte_range for the analogous case — the miss path has the same gap.)
> 6. **cache_store.py:509–514 — negative-cache hits never refresh LRU recency.** The fast-path increments negative_hitsand returns withoutmove_to_end(key), unlike positive hits (_update_access_order). So eviction is least-recently-*inserted*, not least-recently-*used* as
  the class / _next_eviction_candidate/_accommodate_valuedocstrings claim — a frequently-read absent key is evicted before a cold but newer one. Low severity; fix is a one-linemove_to_end` or a docstring correction.

Some of these concerns are about the _change_ in behavior. Since this is marked as experimental, I think that's not as big a deal, but worth an impact assessment I think

d-v-b:
I think a finite default max age is probably a good idea in any case, just to keep things bounded.

espg:
Thanks for running that review — the claude concerns were all legitimate, small fixes. I set the default max age to 5 minutes / 300 seconds, and had claude fable do a pass on the review findings-- all addressed, here's summary report:

---

**1. Coherency regression from `cache_missing=True` + infinite default TTL.** Fixed by
taking your suggestion: the default `max_age_seconds` is now **300 seconds** rather than
`"infinity"`. That bounds the staleness of both remembered misses *and* cached values, so
out-of-band writes become visible within the TTL out of the box; `"infinity"` remains
available as an explicit opt-in. (Impact assessment below.)

**2. Unbounded negative-cache growth under `max_size=None`.** Fixed. When `max_size` is
set, markers share the byte budget as before. When `max_size is None` — where the shared
budget couldn't bound anything — the marker *count* is now capped at an internal limit
(100,000, ~13 MB of index overhead worst-case), evicting the least-recently-used marker
first. The state carries an O(1) `negative_count` so the cap costs nothing on the hot
path. A scan over a 10M-chunk sparse array now leaves at most 100k markers (the hottest
ones), not 10M. I kept it an internal constant rather than a new constructor parameter to
avoid growing the API for a bound most users should never hit; easy to promote to a
parameter later if someone needs to tune it.

**3. Write/miss TOCTOU: a stale "absent" observation shadowing a concurrent write.**
Fixed structurally rather than relying on TTL self-healing. The miss handler now records
*when the source fetch began* (monotonic), and on the absent path it re-checks the key's
slot under the lock: if a present entry exists that was inserted at-or-after the fetch
began, a concurrent `set` won the race and the stale absent observation is dropped —
no marker is recorded and the cached value is left untouched. The backing-cache delete
also moved inside that lock/check, which closes the same window for the pre-existing
positive-cache variant of the race (a concurrent `set`'s cached value being deleted by a
stale miss). There's a regression test that replays the interleaving deterministically.

**4. Oversized `set()` leaving an untracked orphan in the backing cache.** Pre-existing,
but you're right that this PR's "a single `max_size` bounds total cache memory" claim
made it load-bearing — fixed in both places it occurred: `set()` with
`cache_set_data=True`, and the full-key read-miss fill path. If `_track_entry` declines
the value (larger than `max_size`), the just-written backing-cache entry is deleted,
mirroring the byte-range path's existing rollback.

**5. Full-key miss not invalidating cached byte-range entries.** Fixed: observing a key
absent now invalidates that key's byte-range entries in the same locked section that
records the miss, so `get(key)` and `get(key, byte_range)` can no longer diverge (this
matches what `set`/`set_if_not_exists` already did; there's a test for the miss path
now too).

**6. Negative hits not refreshing LRU recency.** Fixed with the one-liner: a negative-
cache hit now `move_to_end`s the marker, so eviction is genuinely least-recently-*used*
and a hot absent key outlives cold markers (test asserts the eviction order).

## Impact assessment of the behavior changes

Two defaults change, both experimental-tier:

- **`max_age_seconds`: `"infinity"` → `300`.** Reads of keys not seen by this instance
  route to the source once and re-validate at most every 5 minutes thereafter — for
  workloads that relied on never-expiring entries (static data, immutable stores), this
  adds a source round-trip per key per TTL window; those users can pass `"infinity"`
  explicitly. One subtlety: with a finite TTL, a *fresh* `CacheStore` instance pointed at
  a pre-populated persistent cache backend re-fetches each key once before trusting it
  (an unseen key has no tracked entry, so it isn't considered fresh). That is the
  pre-existing finite-TTL semantic, but it's newly the default.
- **`cache_missing`: absent → `True`.** Absent keys are now remembered (bounded by TTL +
  the budget/cap above). With the finite default TTL the worst case for an out-of-band
  write is 5 minutes of invisibility, identical to the staleness bound on positive
  entries — the asymmetric "absent forever" mode from the earlier revision is gone
  unless explicitly configured (`cache_missing=True` + `max_age_seconds="infinity"`,
  which the docstring now warns about specifically).

Existing tests all pass with the new defaults (one assertion updated to expect the
finite default), and the six fixes each carry a dedicated regression test (71 total in
the cache-store suite now). Also merged `main` into the branch.

d-v-b:
hi @espg, thanks for these improvements. This addition exposed some underlying structural problems with the cachestore, and it's a bit involved to explain so i'm going to have claude do it (see text below). The tl;dr is that there's a few things remaining to be fixed in this PR, and then someone (me probably) will open a follow-up PR that effects the core improvement into the cachestore that makes it simpler to work on.

Hi @espg — the single-slot restructure and the round-2 fixes held up well under re-review, and one point in your favor we hadn't credited: the `set_if_not_exists` override fixes a *pre-existing* staleness bug. On `main`, `CacheStore` doesn't override it at all, so `WrapperStore` delegates straight to the source with no cache invalidation — a successful `set_if_not_exists` left any stale cached value being served indefinitely. Your unconditional-invalidate override closes that.

Testing the branch (merged onto current `main`) did turn up one more coherence gap, and I think it points at a structural issue worth solving once rather than patch-by-patch — proposal below.

## One more coherence bug

A successful **byte-range** read doesn't clear the key's negative marker, so the same instance answers differently for the same key depending on how you ask:

```python
cs = CacheStore(src, cache_store=MemoryStore())   # all defaults

await cs.get("k", proto)                                      # → None; marker recorded
await src.set("k", cpu.Buffer.from_bytes(b"HELLO WORLD"))     # out-of-band write to the source
await cs.get("k", proto, byte_range=RangeByteRequest(0, 5))   # → b"HELLO" (source consulted, range cached)
await cs.get("k", proto)                                      # → None (served from the marker!)

cs.cache_info()   # tracked_keys=2, missing_keys=1  ← present range entry AND absent marker, same key
```

This is the mirror image of your fix #5 (full-key miss now invalidates range entries): the reverse edge — range success clears the marker — was never written. The "a key can never be both cached and marked missing" invariant only covers the full-key slot; byte-range entries live under `(key, ByteRequest)` tuple keys, outside it. The divergence window is the full TTL (300 s by default), and mixed full/range reads on one key are realistic with sharding, where shard-index reads are ranged.

The point fix is one line — `_evict_missing(key)` in `_cache_miss`'s byte-range success path, plus a test. But it would be the *seventh* hand-maintained cross-invalidation edge, which is the real problem.

## Root cause

Knowledge about one key lives in two structures keyed differently. The source-side truth per key is a single fact — *absent*, or *present with some bytes* — but the cache encodes it in a full-key slot (value or marker) **plus** per-range slots. Coherence then depends on hand-written edges between them: `set` invalidates ranges; fix #5 made full-key misses invalidate ranges; this bug is the edge nobody wrote. Every new path × state combination is another edge someone has to remember.

## Proposal: one record per key; a read is a query with a range parameter

Make the unit of caching the **key's state**, not the request. One record per key; a full read is just the range `[0, ∞)`:

```python
@dataclass
class _KeyState:
    exists: bool               # last observed existence; False = negative entry
    observed_at: float
    intervals: IntervalMap     # known byte-intervals of the value;
                               # a full value is the single interval [0, size)
    size: int | None           # total object size, if known
```

```python
async def get(self, key, prototype, byte_range=None):
    want = _as_interval(byte_range)            # full read → [0, ∞)
    state = self._entries.get(key)
    if state is not None and fresh(state):
        if not state.exists:
            return None                        # negative hit — same path for ranged reads
        if state.intervals.covers(want):
            return state.slice(want)
    result = await super().get(key, prototype, byte_range)
    # ONE atomic update under one lock:
    #   result is None → state = _KeyState(exists=False, intervals={})
    #   result is data → state.exists = True; state.intervals.merge(want, result)
    return result
```

A byte-range read becomes literally "read + interval restriction." Every bug class we've been fixing becomes structurally impossible rather than checklist-prevented:

- **Marker/range divergence** (this bug): can't exist — `exists` and the data intervals are one record mutated atomically; observing data forces `exists=True` in the same assignment.
- **Fix #5**: automatic — recording `exists=False` replaces the record, and the intervals go with it because they're inside it.
- **Negative caching**: no longer bolted-on machinery (`_record_missing` / `_evict_missing` / `negative_count`) — it's just the `exists=False` variant of the one state type. A large fraction of this PR's code dissolves.
- **Single budget**: a record has one size (Σ interval bytes + flat overhead); LRU over keys. "A marker never displaces cached data" stops being an eviction-policy promise and becomes trivial — an absent record *is* the cheap record.

And it buys things the current design can't do at all:

- A cached **full value serves subsequent range reads** by slicing — today a range read of a fully-cached key still goes to the source. For sharded arrays (read a shard, then ranged index/chunk reads of the same key) this is the dominant access pattern.
- **Ranged reads get negative caching too** — a ranged probe of an absent shard is remembered identically, instead of being scoped out.
- **`delete` can record absence** — after a delete you *know* the key is absent; today that knowledge is discarded because the marker machinery is read-path-only.
- Interval **coalescing** (`[0,100)` + `[100,200)` serves `[50,150)`) becomes an optional upgrade to `IntervalMap`, not a redesign.

One contract note: this wants `get(key, byte_range) is None ⟺ key absent`. That holds for `MemoryStore` and `FsspecStore` (KeyError / FileNotFoundError → `None`) and I believe it's the intended ABC semantics; worth pinning down in the `Store.get` docstring. If some store can't promise it, the fallback is just "don't record absence from ranged misses" — you lose an optimization, but coherence still holds because there's only one record to update.

This is well-trodden ground: TensorStore's kvstore cache (per-key entries, `byte_range` as a read option), HTTP range caching (RFC 9111 — all cached ranges hang off one resource + validator), and the OS page cache (key = file, intervals = pages) all have this shape. None of them keep "does it exist" and "which bytes do we have" in separate structures.

## What I'd suggest for this PR vs. a follow-up

I don't think the full interval design belongs in this PR — it's a rewrite of the read path. Two levels:

1. **This PR (incremental):** nest the range cache inside the per-key entry — `entries: dict[str, _KeyState]` where the state holds present/absent *and* that key's range buffers (today's `range_cache[key]` moves inside). Tuple keys disappear from `entries`; every mutation for a key goes through one slot under one lock, and `absent ⇒ no ranges` becomes an assertion at the single mutation point instead of a distributed invariant. Per-range recency can live inside the record, so the LRU/budget behavior you built is preserved. This is a modest refactor of what's already here and closes the whole divergence class, including the bug above. (If you'd rather land the one-line point fix + test now and leave the restructure to a follow-up, I'm open to that too — say the word.)
2. **Follow-up PR:** the full interval design (serve ranges from cached full values, ranged negative hits, coalescing). Experimental-tier status gives us room to do this without deprecation ceremony. I'm happy to collaborate on it.

## Smaller items from the same review pass

- **`_next_eviction_candidate` is O(n) per eviction, even with `cache_missing=False`** — it walks the whole dict to prove no absent marker exists before returning the LRU present entry (old code was `next(iter(...))`). Measured on evicting `set`s with `cache_missing=False`: ~6 µs flat on `main` vs. 10/22/68/261 µs at 100/400/1600/6400 resident entries on this branch — a linear regression on the plain positive-cache path. The fix is one line, using state you already maintain: `if self._state.negative_count == 0: return next(iter(self._state.entries), None)`. The same shortcut makes `_record_missing`'s worst case (budget saturated with cached values, zero markers — currently an O(n) scan that then records nothing) O(1). Under the incremental restructure above, both scans get cheaper still since range entries leave the top-level dict.
- **Changelog fragment is misnumbered**: `changes/4040.feature.md` → should be `4042.feature.md` (#4040 is the ChunkLayout PR).
- **Merge conflict with `main` is trivial**: both sides appended tests at EOF (`TestCacheStoreNegativeCaching` here, the sync-IO tests from #4206 on `main`) — keep both; `cache_store.py` itself auto-merges cleanly. I ran the merged result locally: 73 tests pass, mypy strict clean.
- **TOCTOU note**: the `prior_entry` identity guard can't fire when the concurrent writer leaves no present entry — `cache_set_data=False`, and also `set_if_not_exists` (its override removes tracking rather than inserting a present entry). TTL-bounded and narrow; I'd just extend the existing comment rather than add machinery.
- Cosmetic: `cache_info()` / `__repr__` scan all entries to count present/missing though `negative_count` is already maintained; and the `if self.cache_missing:` gate before `_evict_missing` in `set` is redundant (the following `_track_entry` / `_remove_from_tracking` clears the slot either way).

espg:
Thanks — took option 1; all on the branch (see details below)

Happy to collaborate on the interval follow-up — the slicing rule from this last pass is a step toward it rather than away.  I'd love to supply some benchmarks on real data once the redesign and #4228  land; our sparse/sharded workloads (shard-index ranged reads plus empty-chunk probes at scale) will be reproducible against a public store within the next week or so as we reprocess to our final schema (will be ~300+ GB on S3, sparse geophysical data over California). 

@d-v-b let me know what else I can do to help 

🤖 AI text below 🤖

---

The restructure is as you sketched it: `entries: OrderedDict[str, _KeyState]` — the full-key slot (value or absent marker) and that key's byte-range buffers in one record; the tuple-keyed entries and separate range cache are gone; eviction stays absent-first LRU, with per-range recency inside the record.

Claude's self-review found the "cannot diverge" claim wasn't earned yet, so two more commits landed the rule in one line: **a ranged read is served by slicing a fresh cached full value when one exists, and any source observation — bytes or absence, full or ranged — replaces the key's whole record under the lock.** A record holds a full slot *or* ranges, never both. One deliberate scope call: a ranged miss invalidates the record but records no absent marker — whether `get(key, byte_range) is None` implies absence is a store-contract question better settled in the interval design; invalidation needs no answer to it and costs at most one extra round-trip.

Same class, same path, so also fixed: `get_ranges`/`get_partial_values`/`_get_many` went around the cache entirely (shard index cached, chunk reads not) — they now route through `self.get` (the `LatencyStore` pattern), so a cached shard serves its chunk ranges by slicing. `delete_dir`/`clear` likewise delegated to the source and left the cache serving deleted chunks for a full TTL; both now drop the affected records.

Testing: your repro, its stale-present polarity, and both sibling edges are regression tests; two-coroutine race tests drive the public API; a seeded fuzz calls `_assert_invariants()`, a whole-state sweep the tests use since `-O` strips asserts. The sweep caught one more real bug: the oversized-`set` rollback deleted the backing value but left the previous entry tracked, phantom-charging the budget — fixed under the lock, tested from a non-empty cache. The residual blind spot is stated exactly: a concurrent mutation that leaves no present slot behind (`cache_set_data=False`, `set_if_not_exists`, oversized `set`, `delete`) has no identity to guard against — TTL bounds it, no machinery added.

The O(1) candidate shortcut is its own commit (`negative_count == 0 → next(iter(entries), None)`, same check in `_record_missing`'s saturated case). Your measurement's shape reproduced: **9.3/21.5/66.5/259.8 µs** at 100/400/1600/6400 resident before, **flat 5.4–5.6 µs** after — flat also with `cache_missing=True` and with a hot marker pinned at MRU. Also done: `main` merged (both test blocks kept), changelog fragment renamed to 4042, `cache_info`/`__repr__` O(1). One CI catch after pushing: within-key LRU recency was wall-clock-ordered, and Windows' ~15.6 ms monotonic granularity tied the timestamps and mis-evicted — recency is now a monotonic use-tick (real time stays for TTL only), pinned by a frozen-clock test. 93 tests, mypy strict clean.

PR #4046, opened 2026-06-06
refactor: use backported sentinel type

### LINKED ISSUE #4045: upgrade home-built sentinel to `typing_extensions.Sentinel`
We define a sentinel class [here](https://github.com/zarr-developers/zarr-python/blob/b871a22f4e946e29e49b7998279e30ada8932077/src/zarr/core/array.py#L173), and we use it throughout the codebase where want to allow `None` to work as a fill value declaration.

`typing_extensions>=4.14` has a backport of the [stdlib `Sentinel` type ](https://docs.python.org/3.15/library/functions.html#sentinel), which does the same thing as our home-built sentinel. So we should swap our implementation over to that.

d-v-b:
closes #4045


ilan-gold (commented):
Oh wow I did not know this existed, very cool!

ilan-gold (approved):

PR #4048, opened 2026-06-07
fix: allow opening non-existent ZipStore in read mode

### LINKED ISSUE #2450: A fresh `ZipStore` cannot be opened with mode `r`
You cannot open a non-existent `ZipStore` in `r` mode, because `zipfile.ZipFile.__init__` will fail to find the zip file on disk and raise an exception. This is obvious given how files work, but surprising in the context of the other `Store` classes, which can all be opened in `r` mode without any IO (I think).

NIK-TIGER-BILL:
## Summary

Currently, opening a non-existent ZipStore in mode `r` fails because `zipfile.ZipFile` raises an exception when the file does not exist. This is inconsistent with other Store classes which can be opened in `r` mode without issues.

## Changes

- Handle the case where a ZipStore is opened in `r` mode but the file does not yet exist.

## Testing

- Added test verifying that a fresh ZipStore can be opened in `r` mode.

Fixes zarr-developers/zarr-python#2450

### DIFF
--- a/src/zarr/storage/_zip.py
+++ b/src/zarr/storage/_zip.py
@@ -94,6 +94,18 @@ class ZipStore(Store):
 
         self._lock = threading.RLock()
 
+        if self._zmode == "r" and not self.path.exists():
+            # zipfile.ZipFile requires the file to exist in read mode,
+            # but other stores can be opened in read mode without IO.
+            # create an empty zip file so that the store can be opened.
+            with zipfile.ZipFile(
+                self.path,
+                mode="w",
+                compression=self.compression,
+                allowZip64=self.allowZip64,
+            ):
+                pass
+
         self._zf = zipfile.ZipFile(
             self.path,
             mode=self._zmode,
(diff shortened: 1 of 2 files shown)


d-v-b:
i feel like it's a bit surprising if `mode=r` has side effects (creating a zip file). This is a special thing for the zipstore because a zip archive exists in some _other_ storage system.

NIK-TIGER-BILL:
That is a fair point. I chose creating an empty zip as the least-surprising way to keep the store API consistent — other stores can be opened in `r` mode even when the backing path does not exist yet, and the caller gets an empty read-only store.

The alternative I considered was to raise a custom `FileNotFoundError`, but that breaks the uniformity of the `Store.open` contract. Another option would be to hold a no-op `ZipFile`-like object internally, but that would require re-implementing parts of `zipfile` just to avoid the side effect.

I'm happy to switch to a different approach if the maintainers prefer — e.g., raise a clear `FileNotFoundError` in `r` mode, or document the side effect explicitly. Let me know what you think.

PR #4049, opened 2026-06-07
refactor: one Array class with sync+async APIs

maxrjones:
Alternative to #4034 that delivers every #4027 goal except removing the public `AsyncArray` class.

- Array gains a user-friendly constructor: `Array(metadata, store_path, config=...)` builds the wrapped `AsyncArray` internally; `Array(async_array)` still works for back-compat.
- `Array` gains the full public async surface (`*_async` twins for every sync method), so callers no longer reach into the private _async_array.
- Each operation has a single implementation: the async twin holds the logic and the sync method delegates via `sync()`. No drift between the pairs; a parity test guards their signatures.
- No `Runner` protocol and no `_AsyncArrayView`/`_rebind_state` shared-state machinery: `Array` still wraps one `AsyncArray`, so `async_array` is the held object and mutation coherence is free.
- `getitem_async`/`setitem_async` mirror the full basic-selection parameter surface (`out=`, `fields=`, `prototype=`).

Tests: integrates the async/sync boundary suite (shared-state coupling, codec-pipeline-built-once, selection parity, out= zero-copy, full parameter surface). The loop-affinity test is xfail (strict) as a design-independent limitation of mixing zarr's background loop with a user-owned loop on stores with event-loop affinity.

TODO:


src/zarr/core/array.py
```diff
@@ -2271,7 +2307,7 @@ def nchunks_initialized(self) -> int:
         >>> arr.nchunks_initialized
         6
         """
-        return sync(self.async_array.nchunks_initialized())
+        return sync(self.nchunks_initialized_async())
```
mkitti, line 2310: If all the "synchronous" calls just call `sync` on their asynchronous versions, then we still incur all the overhead of scheduling an async task.

src/zarr/core/array.py
```diff
@@ -3801,6 +3778,10 @@ def resize(self, new_shape: ShapeLike) -> None:
         ----------
         new_shape : tuple
             New shape of the array.
+        delete_outside_chunks : bool, optional
+            If True (default), chunks that fall entirely outside the new array
+            shape are deleted from the underlying store. If False, those chunks
+            are left in place.
```
mkitti, line 3784: Should we be exposing new keyword arguments in the synchronous API at this time? Could we defer this change to a later pull request?

mkitti (changes requested):
I think we have the pattern backwards here. If the objective is to lower overhead for the synchronous API,
* Instead of calling `sync` on the asynchronous API ...
* Use `asyncio.to_thread` to call a synchronous function.

This way the synchronous API has as little overhead as possible. Meanwhile, we still have an asynchronous API that is still asynchronous.

maxrjones:
thanks for your review @mkitti! I should've marked this PR as a draft. It's one of a few possible solutions related to https://github.com/zarr-developers/zarr-python/issues/4027, including #4034 and [generated sync code](https://hackmd.io/@maxrjones/BJFwYD4bMg). I'd be curious for your overall take on the different options.

mkitti:
My understand of the situation is there is a technical design flaw currently in zarr-python in that it uses asynchronous tasks excessively creating a bottleneck in the scheduler. The objective to resolve this is to reduce the overhead of asynchronous calls by providing a low-overhead synchronous API.

As this StackOverflow answer demonstrates the act of simply creating an asynchronous task has some significant overhead, especially when the asynchronous task is small. In the case of the SO example, the async task is just a `noop`:
https://stackoverflow.com/a/55766474

If we want to eliminate code duplication, rather than implementing an asynchronous function and calling `sync` on it, we should implement a low-overhead synchronous function and use `asyncio.to_thread` to turn it into an asynchronous function. This is not appropriate for all cases, but at least will preserve the async functionality.

In the medium term, we would want to aggregate async functionality into larger tasks in order to create less burden on the async scheduler.

In the long term, I recommend exploring how the API could be best adopted for true free threading. As free threading becomes more popular, using a thread pool via [`ThreadPoolExecutor`](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor) is the superior approach. The key here is to minimize the cost of thread creation by using the thread pool, and then achieving true parallelization by submiting tasks to threads such that the tasks can actually occur concurrently.

PR #4051, opened 2026-06-08
docs: add a deprecation policy to the contributing guide

### LINKED ISSUE #2924: define our deprecation policy
An update to our versioning scheme is pending in #2910. That PR spawned a [necessary discussion](https://github.com/zarr-developers/zarr-python/pull/2910#discussion_r2004400243) about what deprecation policy we should have. This also relates to the question of our release schedule. We can use this issue to figure these things out.

cc @dstansby

maxrjones:
Closes #2924.

This PR adds a deprecation policy to the contributing guide, so there's a shared framework all contributors and maintainers can apply rather than re-deriving one for each PR. Making it a public doc also gives users visibility into how we handle breaking changes.

The policy is consistent with our EffVer versioning and covers scope, decision framework, mechanics, minimum period, and exceptions to the policy.

TODO:


docs/contributing.md
```diff
@@ -369,6 +369,49 @@ The Zarr library is an implementation of a file format standard defined external
...
+- **The cost of keeping the API.** What do we pay by not removing it -- maintenance burden, bug surface, duplicated logic, or a worse design that we can't improve while the old API exists? A high carrying cost strengthens the case for removal.
+- **Who benefits.** Distinguish benefit to users from benefit to maintainers. A change that helps only maintainers (for example, deleting code that is inconvenient to keep) is a weak case on its own, because users pay the migration cost and get nothing. A change that *also* gives users something they want is a much stronger case.
+- **Whether the goal can be met without removal.** Often a refactor, an alias, or an additive change achieves the same end without breaking anyone. Prefer those. Removing public API is never free; the deprecation cycle is the price, and it is mandatory.
+
+If, after weighing these factors, a removal is not clearly worthwhile, prefer to keep the API. Conversely, do not leave the decision to remove an API indefinitely deferred: a vague intention to "eventually" break something is itself a form of technical debt. Either commit to the deprecation and begin the cycle, or decide to keep the API and design around it.
```
d-v-b: This reads a bit like an effort to discourage developers from wantonly deprecating code they don't like, without regard for how it impacts users. I don't think we have a lot of that in our codebase, so maybe a more neutral tone would work better here. Even better would be just linking to or copying text from the NumPy deprecation guidelines: https://numpy.org/neps/nep-0023-backwards-compatibility.html#general-principles

maxrjones: thanks for the review. I revised the text to better capture the intended neutral tone in [`88faa7a` (this PR)](https://github.com/zarr-developers/zarr-python/pull/4051/commits/88faa7ad8936288b77413c457c387cdb1d0dd913) and ran the section through a [tone analyzer](https://goblin.tools/Judge) to validate the changes. 

I think it's better to adapt NumPy's principles for Zarr-Python rather than copying/linking directly. NEP23 is from 2018, and I expect even they would write it slightly differently today.

docs/contributing.md
```diff
@@ -398,6 +398,51 @@ The Zarr library is an implementation of a file format standard defined external
...
+3. **Document it.** Mark the deprecation in the docstring (via the `deprecated` decorator or a "Deprecated" admonition) so it appears in the rendered API documentation, and update any affected user-guide prose.
+4. **Add a changelog entry.** Add a `removal` changelog fragment in `changes/` (for example, `changes/1234.removal.md`) so the deprecation is announced under "Deprecations and Removals" in the release notes. When the API is finally removed, add a second `removal` fragment for that release.
+
+When an API is being replaced rather than simply dropped, deprecate the old name and introduce its replacement in a single step, rather than repeatedly adjusting the same signature across releases. Users should have to migrate only once.
+
+#### How long a deprecation lasts
+
+A deprecated API must remain available, emitting its warning, for **at least 6 months and at least one minor release** before it is removed. These are minimums, not targets: for widely-used API, prefer a longer cycle, and communicate the upcoming removal beyond the warning itself (for example, in release notes or a migration guide).
```
ilan-gold, line 434: I would note that effver means we can do this, but that this is not real semver.  This is actually not immediately stated anywhere in this doc AFAICT but this new section of the doucment really sharpens the need i.e., "breaking changes are allowed on the minor / patch if needed/warranted"

docs/contributing.md
```diff
@@ -398,6 +398,51 @@ The Zarr library is an implementation of a file format standard defined external
...
+2. **Write an actionable message.** The warning message must state the release the deprecation first appeared in, the planned removal (a specific version if known, otherwise "a future release"), and the migration path -- what the user should do instead. A deprecation warning with no replacement guidance is incomplete.
+3. **Document it.** Mark the deprecation in the docstring (via the `deprecated` decorator or a "Deprecated" admonition) so it appears in the rendered API documentation, and update any affected user-guide prose.
+4. **Add a changelog entry.** Add a `removal` changelog fragment in `changes/` (for example, `changes/1234.removal.md`) so the deprecation is announced under "Deprecations and Removals" in the release notes. When the API is finally removed, add a second `removal` fragment for that release.
```
ilan-gold, line 428: What about for changed, not deprecated, behavior i.e., what you state in point 1?

docs/contributing.md
```diff
@@ -398,6 +398,51 @@ The Zarr library is an implementation of a file format standard defined external
...
 
+### Deprecation policy
+
+Our versioning policy (above) commits us to minimizing the effort users spend upgrading. A deprecation cycle is the main tool we use to honor that commitment: rather than removing or changing public behavior abruptly, we first ship a release that keeps the old behavior working while warning that it is going away, and only remove it in a later release. This section defines when a deprecation cycle is required, how to decide whether a removal is worth it, and the concrete steps a deprecation must follow.
+
+This policy governs the user-facing Python API of `zarr-python`: importable names, function and method signatures, and observable runtime behavior. It does not govern the Zarr data format, whose compatibility is described under [Data format compatibility](#data-format-compatibility) above, nor the `zarr.experimental` namespace, which is governed by the [Experimental API policy](#experimental-api-policy) below. Private API (names prefixed with `_`, and anything not documented as public) may change at any time without a deprecation cycle.
```
ilan-gold, line 405: This should be clear to include exceptions as well IMO I get that counts as "runtime behavior" but it may not come to people's minds

PR #4052, opened 2026-06-08
Modify custom dtype example to use only public API

chuckwondo:
Modify custom dtype example to use only public API.  As part of this effort, move/deprecate any types or functions currently inaccessible from the public API, but should be.

- Internally change all imports of `DataTypeValidationError` to import from `zarr.errors` and deprecate imports of `DataTypeValidationError` from all other modules.

TODO:

### DIFF
--- a/src/zarr/core/dtype/__init__.py
+++ b/src/zarr/core/dtype/__init__.py
@@ -3,10 +3,6 @@ from __future__ import annotations
 from collections.abc import Sequence
 from typing import TYPE_CHECKING, Final
 
-from zarr.core.dtype.common import (
-    DataTypeValidationError,
-    DTypeJSON,
-)
 from zarr.core.dtype.npy.bool import Bool
 from zarr.core.dtype.npy.bytes import (
     NullTerminatedBytes,
@@ -39,6 +35,7 @@ from zarr.core.dtype.npy.time import (
 
 if TYPE_CHECKING:
     from zarr.core.common import ZarrFormat
+    from zarr.core.dtype.common import DTypeJSON
 
 from collections.abc import Mapping
 
@@ -61,7 +58,6 @@ __all__ = [
     "Complex64",
     "Complex128",
     "DataTypeRegistry",
-    "DataTypeValidationError",
     "DateTime64",
     "DateTime64JSON_V2",
     "DateTime64JSON_V3",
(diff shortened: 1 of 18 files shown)


chuckwondo:
@d-v-b, I believe I've covered everything, except for renaming `check_dtype_spec_v2` to `is_dtype_spec_v2`.  Do you want me to do that too? If so, do you want me to deprecate `check_dtype_spec_v2`, or is that unnecessary since prior to this PR it is technically not public?

chuckwondo:
> @d-v-b, I believe I've covered everything, except for renaming `check_dtype_spec_v2` to `is_dtype_spec_v2`. Do you want me to do that too? If so, do you want me to deprecate `check_dtype_spec_v2`, or is that unnecessary since prior to this PR it is technically not public?

nudge

d-v-b:
> @d-v-b, I believe I've covered everything, except for renaming `check_dtype_spec_v2` to `is_dtype_spec_v2`. Do you want me to do that too? If so, do you want me to deprecate `check_dtype_spec_v2`, or is that unnecessary since prior to this PR it is technically not public?

yes, we can safely rename, as it wasn't public, despite appearing in the example.

d-v-b:
out of abundance of caution I can open a follow-up PR that keeps the old `check_...` name around, with a deprecation warning.

chuckwondo:
> out of abundance of caution I can open a follow-up PR that keeps the old `check_...` name around, with a deprecation warning.

Actually, I propose that I revert the rename in the PR, particularly since there are a number of other `check_*` functions that should be renamed to `is_*`, so perhaps they can all be addressed together in a separate PR.

Given that this PR is meant to focus solely on replacing use of non-public things in the example with only public things, the rename is perhaps out of scope.

chuckwondo:
@d-v-b, is there anything you would like me to add to other parts of the docs describing some of the changes in this PR, particularly the deprecation of the exception from its previous location?

src/zarr/core/dtype/__init__.py
```diff
@@ -287,3 +283,19 @@ def parse_dtype(
...
+
+
+def __getattr__(name: str) -> object:
+    if name == "DataTypeValidationError":
+        import warnings
+
+        from zarr.errors import DataTypeValidationError, ZarrDeprecationWarning
+
+        warnings.warn(
+            "Importing DataTypeValidationError from zarr.core.dtype is deprecated. "
```
d-v-b, line 295: flagging for myself: we should direct readers of this message to a timeline for when the import will break. That could be either be a reference to a specific future release, or a link to a general deprecation policy.

d-v-b (approved):
Thank you chuck!

PR #4053, opened 2026-06-08
docs: add core-developer self-merge policy

maxrjones:
The contributor guide currently says "Ideally, pull requests submitted by a core developer should be reviewed and approved by at least one other core developer before being merged." It doesn't define when that "ideally" can be relaxed.

This PR defines circumstances under which a core developer should feel empowered to self-review and merge a PR. The goal is to focus reviewer energy on the changes where it matters, and to reduce review load on routine, low-risk ones.

We should continue to evaluate the self-review boundary as the split between writing code and reviewing AI-generated code evolves. This is just a starting point.

As an example, I think https://github.com/zarr-developers/zarr-python/pull/4047 should be self-mergable.

The new content lives in a `Project governance` section that groups the maintainer-facing process (PR merging, the new self-merge policy, and the release procedure, which moved up from the bottom of the page into this section).

TODO:


dcherian:
+10_000!

Can we add "low-blast-radius bug fixes" and "emergency bug fixes" to the list too? We already trust maintainers with the commit bit; i think we can by default trust them to make reasonable decisions regarding when to self-merge.

I'd like to add "low-blast-radius new features" too, but am having trouble defining it.

maxrjones:
> +10_000!
> 
> Can we add "low-blast-radius bug fixes" and "emergency bug fixes" to the list too? We already trust maintainers with the commit bit; i think we can by default trust them to make reasonable decisions regarding when to self-merge.
> 
> I'd like to add "low-blast-radius new features" too, but am having trouble defining it.

thanks for your feedback! I took a stab at addressing your comment in [`3fb243d` (this PR)](https://github.com/zarr-developers/zarr-python/pull/4053/commits/3fb243ddef28c98d9f2d06665eaf5cd8fc51df62)

docs/contributing.md
```diff
@@ -327,14 +327,79 @@ Alternatively, you can manually create the files in the `changes` directory usin
...
+
+#### Conditions
+
+A core developer may self-merge their own pull request only when **all** of the following hold:
+
+1. The change falls into one of the eligible categories listed below.
+2. The standard merge requirements above are met — all CI checks have passed (GitHub Actions, Codecov) against code that has had the latest `main` merged in, and the milestone is set.
+3. The pull request has been open for at least **24 hours, not counting Saturdays or Sundays (UTC)**, with no unresolved objection from another core developer. This ensures the change is visible during at least one working day, so other core developers have a realistic chance to object.
+4. The author has applied the `self-merge-eligible` label. Any core developer may remove this label, which returns the pull request to the standard review process. Removing the label does not require justification.
+
```
dcherian: I don't like how prescriptive this is, but I have not been very closely involved recently to know if a looser policy would be problematic.

My perspective: By adding members of the maintenance team, we are implicitly placing a lot of trust in them. I think we can trust people to make decisions, and we want to bias towards action. This kind of prescriptive rule writing biases against action which is sad. 

FWIW Xarray switched (implicitly) to a far more permissive merge-within-3-days-if-you-think-it-is-ok and the project is far better off for it. It also tends to encourage newer maintainers to get comfortable merging changes. Reverts are cheap.

dcherian (approved):
Approving since it is an improvement.

I left a comment arguing for a looser policy: https://github.com/zarr-developers/zarr-python/pull/4053/changes#r3375928403

docs/contributing.md
maxrjones: I support a more permissive policy. Since no one else has commented yet, I'll adjust the PR to use a less prescriptive approach

TomAugspurger (commented):
I was going to leave a comment similar to Deepak's. I'm glad he beat me to it.

docs/contributing.md
maxrjones: done in [`6234621` (this PR)](https://github.com/zarr-developers/zarr-python/pull/4053/commits/6234621b48f02e3226fb22f561c5328399ffd310)

maxrjones:
fyi @dcherian @TomAugspurger the current language is more trusting/less prescriptive than when Deepak approved the PR, but in line with your preferences IIUC

d-v-b:
thanks for writing this up max, I think it's a great idea. Since this reduces the formality / oversight for commits on main, maybe we could afford to increase the formality level / oversight for _releases_. Offhand I can't think of any particular rules or procedures to change, so maybe we can leave the language as-is. But my feeling is that when it's time for a release we should probably aim to have ~two contributors review all the PRs merged since the previous release. That's an aspiration, and it depends on dev time etc etc, so maybe just something to mention as a note.

maxrjones:
> Since this reduces the formality / oversight for commits on main, maybe we could afford to increase the formality level / oversight for releases. Offhand I can't think of any particular rules or procedures to change, so maybe we can leave the language as-is. But my feeling is that when it's time for a release we should probably aim to have ~two contributors review all the PRs merged since the previous release. That's an aspiration, and it depends on dev time etc etc, so maybe just something to mention as a note.

Thanks for raising this for discussion. I think the bar that we're unlikely to meet is to have two contributors "review" all the PRs, where review means understand and agree with all the changes. It would help to understand more specifically what problem(s) you want to avoid via higher formality/oversight for releases (e.g., bugs, poorly designed features, ...)? That would help establish whether we can use CI and/or AI to address some of the concerns without moving the human bottleneck from merging PRs to making releases.

docs/contributing.md
```diff
@@ -327,14 +327,43 @@ Alternatively, you can manually create the files in the `changes` directory usin
...
 
 Before merging, the milestone must be set to decide whether a PR will be in the next patch, minor, or major release. The next section explains which types of changes go in each release.
 
+### Self-merging pull requests
+
+The default is that a pull request opened by a core developer is reviewed and approved by at least one other core developer before it is merged. We trust core developers to use their judgment, though, and we would rather bias toward action than make routine changes wait on review they do not really need.
+
+So a core developer may merge their own pull request whenever they judge the change to be low-risk, provided the standard merge requirements are met — CI is green against code that has had the latest `main` merged in, and the milestone is set — and other core developers have had a fair chance to weigh in. As a rule of thumb, leave the pull request open for a few days before self-merging, unless it is genuinely trivial or time-sensitive. If you are confident a change is fine, merge it; if you have real doubts, ask for a review.
```
ilan-gold: So I assume, to some degree, the goal here is to accelerate development.  In this sense, could we require review requested but not received? This way, another developer would be _aware_ of a change at least but not necessarily have to spend time reviewing it. I think disseminating this knowledge of what is being fixed and how is nice i.e., not just one person knows what is going on even if the person whose review is requested is not necessarily doing a review.

d-v-b:
>  It would help to understand more specifically what problem(s) you want to avoid via higher formality/oversight for releases (e.g., bugs, poorly designed features, ...)?

All types of problems are in scope here. I'm thinking in very simple terms: today, we have a certain rate of review-per-change `r_1`. If we apply the changes recommended here, we will end up with a new rate of review-per-change `r_2`, where `r_2 < r_1` (most likely). That means less review, and so scoped only the PRs, this means more PRs with problems of any kind slipping through. On its own, that would be problematic. But if we are deliberate about applying compensatory review prior to _releases_, then we can get back to our old effective review rate `r_1`, and then we don't expect a higher rate of bugs.

I don't think we need to literally review every single merged PR in detail prior to releases. Just that, to retain our total review rate, a decrease in PR review must be compensated by review at some other time (namely, releases).

d-v-b:
we could also decide that the reduction in review here will only negligibly increase the rate of problematic stuff getting merged, and that this is largely outweighed by higher dev velocity (which I desperately want).

docs/contributing.md
maxrjones: I like the direction of making sure multiple people are aware of what is going on in the code base.

I'm wondering if there's a different mechanism than review requests that could accomplish this, so that it's more clear whether the author is seeking awareness vs. a [true review](https://github.com/zarr-developers/governance/blob/main/CORE_DEV_GUIDE.md#how-to-conduct-a-good-review). What do you think about adding a statement that authors should feel encouraged to ping one or more other core developers for feedback on the scope of the change without formally requesting a review?

maxrjones:
> we could also decide that the reduction in review here will only negligibly increase the rate of problematic stuff getting merged, and that this is largely outweighed by higher dev velocity (which I desperately want).

I think we need non-human bottlenecked ways of making sure that a reduction in PR review numbers doesn't harm the quality of the library. This could include:

- CI improvements (improving our test's ability to catch regressions, etc. should be quicker than ever right now) 
- Agentic skills for self-review and/or pre-release

My thesis in opening this PR (which really just makes our trust model more explicit, rather than changing it) is that it's not possible to address the core-dev author review bottleneck via human effort for a couple reasons:

- we won't be able to increase the reviewer pool to match the rate of core-dev driven code development, given the current balance of AI speed/quality/cost. In order to prevent an active core-dev code writer from becoming prohibitively frustrated by the pace, or the reviewer pool from burning out, the role of the core-dev PR author needs to shift into a dual author/reviewer role.
- there might be a part of the codebase that one core dev wants to improve but cannot motivate any other core devs to invest time in.

docs/contributing.md
ilan-gold: Yup, that seems like the route.  Just ping someone who you maybe know has touched this part of the code

maxrjones: I modified this section accordingly in [`f169d9b` (this PR)](https://github.com/zarr-developers/zarr-python/pull/4053/commits/f169d9b415b38944695cb9630f478b22efbb58a1)

d-v-b (approved):

d-v-b:
image

d-v-b:
thanks max! I made this change in the configuration of `main`

maxrjones:
> thanks max! I made this change in the configuration of `main`

I think there's an option to allow administrators to bypass requirements that would work better, since the configuration above applies to both core devs and non-core dev PR, but the trust model here just applies to core devs

d-v-b:
I'll look into that, unless someone gets there first

d-v-b:
I deselected the box indicated
> Do not allow bypassing the above settings The above settings will apply to administrators and custom roles with the "bypass branch protections" permission. 

This also allows merging when tests / checks are failing, but that seems OK as an extension of our new trust principle.

PR #4064, opened 2026-06-16
[do not merge] zarrs bindings 

d-v-b:

Adds pluggable array engines and a zarrista-backed engine. At the current head this is wired into Zarr's array creation/opening APIs through `engine=`, including `engine="zarrista"`; it is no longer only a standalone CRUD prototype.

The zarrista engine uses the separate `zarrista` package and `zarr-indexing` to resolve supported selections over the backend's subset reads and writes. Dense forward writes can use a subset call; other supported selections use chunk-grouped read/modify/write. Request normalization, planner capabilities, supported dtypes, and storage adapters limit this integration. It does not establish full NumPy indexing compatibility.

Validity-mask tensor layouts are currently rejected because this engine does not preserve those masks. That is an implementation limitation, not a claim that Python or NumPy cannot represent masked data.

Development dependencies are declared in the `zarrista` dependency group, with zarrista tracked from Git and pinned in the lockfile. Build requirements depend on how that dependency is installed; the former description of an in-repository Rust bindings subpackage is obsolete.

The earlier benchmark table and approximately 15x claim described a prior prototype and are omitted from this current-head description. No current end-to-end speedup is established by this factual documentation audit.


d-v-b:
@clbarnes related to your interests

docs/superpowers/specs/2026-06-11-zarrs-functional-api-design.md
```diff
@@ -0,0 +1,210 @@
...
+async def create_new_group(metadata, store, path) -> None          # error if node exists
+async def create_overwrite_group(metadata, store, path) -> None
+async def create_new_array(metadata, store, path) -> None
+async def create_overwrite_array(metadata, store, path) -> None
+async def read_metadata(store, path) -> dict[str, JSON]            # array or group doc
+async def delete_node(store, path) -> None
+async def list_children(store, path) -> list[tuple[str, dict]]     # (path, metadata)
+
+# chunk-level I/O
+async def decode_chunk(metadata, store, path, chunk_coords, *, selection=None) -> np.ndarray
```
kylebarron: Returning an ndarray will force a copy out of rust memory. Longer term, I'd perhaps suggest instead making a wrapper class on the Rust side that will expose an ndarray. At least whenever the data type is supported by the buffer protocol that should remove a memory copy. See for example [`async_tiff.Array`](https://developmentseed.org/async-tiff/latest/api/array/)

docs/superpowers/specs/2026-06-11-zarrs-functional-api-design.md
```diff
@@ -0,0 +1,210 @@
...
+
+## Store bridge
+
+A Rust-side `StoreConfig` resolver, tried in priority order:
+
+1. `zarr.storage.LocalStore` → native `zarrs_filesystem` store.
+2. obstore-backed `ObjectStore` → `zarrs_object_store` (Phase 3).
+3. **Anything else** → generic `PyStore`: a Rust struct implementing
+   `ReadableStorageTraits` / `WritableStorageTraits` /
+   `ListableStorageTraits` over a Python callback object.
```
kylebarron: Nit: as you might expect from our zulip discussion I'm hesitant to lump all of these together. I'm interested in spending a little time with this thinking about how to represent these traits through a Python API

docs/superpowers/specs/2026-06-11-zarrs-functional-api-design.md
```diff
@@ -0,0 +1,210 @@
...
+
+The callback path: the async API function wraps the user's `Store` in a small
+sync Python shim whose methods submit coroutines to zarr-python's existing
+sync event-loop thread (`zarr.core.sync`,
+`asyncio.run_coroutine_threadsafe(...)` + blocking result). Rust calls the
+shim while holding no locks of its own. This makes any conformant `Store`
+(Memory, Zip, Logging, Wrapper, user-defined) work without Rust knowing its
+type. Deadlock safety relies on the existing invariant that code running on
+the zarr sync loop never blocks on these Rust entry points.
```
kylebarron: You know I don't love this 😅. I'd like to specialize Sync and Async stores on the rust side.

docs/superpowers/specs/2026-06-11-zarrs-functional-api-design.md
```diff
@@ -0,0 +1,210 @@
...
+
+## Array construction cache
+
+`Array::new_with_metadata` (serde-parsing the metadata document and building the
+codec chain) is the dominant per-call cost on the native path — measured at
+~20µs for a bytes-only array up to ~80µs for sharded+blosc, against single-digit
+µs of actual chunk I/O on a warm filesystem. To amortize it across the common
+"open one array, then do many chunk operations" pattern, the chunk/region
+routines memoize the constructed `Array` in a process-global LRU cache
+(capacity 128) keyed on `(filesystem root, node path, metadata JSON)`.
```
kylebarron: I think it's pretty messy to do this in a global LRU cache. There's no reason to have a public LRU cache when you could just cache the metadata in... a class instance 😄

docs/superpowers/specs/2026-06-15-crud-backend-abstraction-design.md
```diff
@@ -0,0 +1,259 @@
...
+  `_zarrs_bindings` and the `_bridge.StoreShim`/`resolve_store` plumbing.
+
+The public surface (`zarr.zarrs.decode_region`, etc.) is unreleased on this
+branch, so it can move without backward-compatibility constraints.
+
+zarr-python already contains everything a pure-Python backend needs:
+`BatchedCodecPipeline` (`src/zarr/core/codec_pipeline.py`), `BasicIndexer`
+(`src/zarr/core/indexing.py`), `save_metadata` (`src/zarr/core/metadata/io.py`),
+metadata parsing (`ArrayV3Metadata.from_dict` / `ArrayV2Metadata.from_dict`),
+and chunk-key encoding (`src/zarr/core/chunk_key_encodings.py`).
```
kylebarron: I need to learn more about codec pipelines to understand how they integrate with zarrs

docs/superpowers/specs/2026-06-11-zarrs-functional-api-design.md
```diff
@@ -0,0 +1,210 @@
...
+- `chunk_coords`: `tuple[int, ...]` grid coordinates.
+- `selection`: numpy-style basic indexing — integers, slices (including steps; strided/reversed selections fetch the step-1 bounding box in one call and apply numpy views), and `Ellipsis`. Fancy indexing (integer/boolean arrays) and `np.newaxis` are not supported.
+- `options`: every function also accepts keyword-only
+  `options: ZarrsOptions | None = None` (omitted from the signatures below for
+  brevity) — a dataclass holding concurrency limits and checksum validation
+  flags. Defaults are applied when omitted; in Phase 1 the dataclass exists
+  but carries only defaults (fields become meaningful in Phase 3).
+
+```python
+# node lifecycle
```
kylebarron: I don't like the functional API because it means that you're reparsing metadata that makes sense to just store in a class instance

docs/superpowers/specs/2026-06-11-zarrs-functional-api-design.md
d-v-b: we have very adjacent prior art in https://github.com/zarrs/zarrs-python, I should see what they do!

docs/superpowers/specs/2026-06-11-zarrs-functional-api-design.md
d-v-b: haha I'm not a fan either! ideal scenario is that a store connection has a declarative form that rust can use to reconstruct it (like a URL). Things tied to python memory will remain hard.

docs/superpowers/specs/2026-06-11-zarrs-functional-api-design.md
d-v-b: >  in... a class instance 😄

storing the metadata in a class instance isn't the same as globally caching it. The point here is to avoid performing the same computation that is keyed by exactly the metadata document. Even if we were using OOP, I think it would make sense for the procedure that takes metadata and emits chunk encoding / decoding machinery to use a cache.

kylebarron: > The point here is to avoid performing the same computation that is keyed by exactly the metadata document

You're keying on filesystem - node path - metadata JSON... how would that key ever be shared across more than one Array? Seems like an `Array` class is a logical place to put that metadata

docs/superpowers/specs/2026-06-11-zarrs-functional-api-design.md
clbarnes: You can instantiate numpy arrays on the rust side, either from a vec directly or with an ndarray: https://docs.rs/numpy/latest/numpy/convert/trait.IntoPyArray.html#tymethod.into_pyarray

I imagine it would get more complicated when custom buffer prototypes are passed, because they need to be instantiated on the python side.

d-v-b: I am kind of hoping to put the custom buffer prototype thing in the rear view window. It has not worked well.

d-v-b: I am open to all ideas here, but "every routine that allocates memory takes a python callable" is probably not going to work for cross-language interop. Another complication is that "contiguous stream of bytes" is a lower-level abstraction than "N-dimensional array".

My rough feeling is that we should partition the chunk encoding / decoding routines by the memory access pattern needed for the array side of the array -> bytes transformation. For dense arrays, rust can hand back a contiguous buffer, or write into a pre-allocated buffer. For sparse arrays, we will need to settle on an memory layout, and we probably can't support writing into a pre-allocated sparse array.

docs/superpowers/specs/2026-06-11-zarrs-functional-api-design.md
d-v-b: it's possible that we aren't disagreeing! I want a functional API because I think that's actually a very natural way to interact with zarr data -- for a given operation, you specify the metadata document you want to use, the location, and any other parameters.

It _also_ definitely makes sense to provide users with objects that persistently bind the metadata and the storage backend. That's how zarr-python works today. But if you _start_ with the object oriented API, it can be hard to add the functional API later. I think the functional API is actually really important for doing a lot of cool things with Zarr! so i want to see if we can bake it in foundationally, and add the object-oriented API on top.

docs/superpowers/specs/2026-06-11-zarrs-functional-api-design.md
kylebarron: Fwiw that's how obstore works when passed into a pyo3 binding. It recreates the store on the rust side so no requests need to go through Python

kylebarron:
In case it's useful for evaluation, I started writing a direct, standalone binding of zarrs to Python: https://github.com/developmentseed/zarrista

d-v-b:
> In case it's useful for evaluation, I started writing a direct, standalone binding of zarrs to Python: https://github.com/developmentseed/zarrista

this is awesome kyle! Extremely useful to have a comparison point for the work here.

changes/+zarrista-backend.feature.md
```diff
@@ -0,0 +1,8 @@
+Added `zarr.zarrista`, a `zarr.crud` backend backed by the Rust
+[zarrs](https://zarrs.dev) crate through the
+[zarrista](https://developmentseed.org/zarrista) package. It accelerates
+chunk-level I/O and, unlike the in-repo `zarr.zarrs` backend, has no generic
+Python-store callback bridge: it operates only on stores it can map to a
+zarrista store (currently `LocalStore`) and raises `UnsupportedStoreError` for
```
kylebarron: Why not `ObjectStore` too?

src/zarr/zarrista/_backend.py
```diff
@@ -0,0 +1,171 @@
...
+    backend currently maps a `LocalStore` to a zarrista `FilesystemStore` and
+    rejects everything else. Use the ``"reference"`` or ``"zarrs"`` backend for
+    stores zarrista cannot ingest.
+    """
+
+
+def _node_path(path: str) -> str:
+    """Convert a zarr path (`""`, `"foo/bar"`) to a zarrista node path
+    (`"/"`, `"/foo/bar"`)."""
+    return f"/{path.strip('/')}"
```
kylebarron: Do zarrista paths currently have to start with `/`? Should they have to? Should `a/b/c` be an allowed input?

src/zarr/zarrista/_backend.py
```diff
@@ -0,0 +1,171 @@
...
+def _resolve_store(store: Store) -> zarrista.FilesystemStore:
+    """Map a zarr `Store` to the zarrista store that backs it, or raise.
+
+    zarrista exposes Rust-native store types rather than a callback into an
+    arbitrary zarr `Store`, so only stores with a direct zarrista equivalent are
+    supported. A `LocalStore` maps to a zarrista `FilesystemStore`; any other
+    store raises `UnsupportedStoreError`.
+    """
+    if isinstance(store, LocalStore):
+        return zarrista.FilesystemStore(str(store.root))
```
kylebarron: You can pass the underlying `store` inside an `ObjectStore` directly into Zarrista.

src/zarr/zarrista/_backend.py
```diff
@@ -0,0 +1,171 @@
...
+
+    zarrista accelerates the chunk-level I/O paths (`read_chunk`, `read_subset`,
+    `write_chunk`, `delete_chunk`). Node metadata documents are written, read,
+    listed and deleted with zarr-python's own machinery (delegated to the
+    `ReferenceBackend`); zarrista has no "write this exact metadata document"
+    primitive for arrays, and these operations are not performance-critical.
+
+    All methods first resolve the store with `_resolve_store`, so the backend
+    consistently rejects stores it cannot ingest (raising
+    `UnsupportedStoreError`) rather than half-working on them.
```
kylebarron: With a CRUD backend this means that zarrista is only doing chunk-by-chunk operations? I'm thinking perhaps zarrista should own more of the indexing strategy than just chunk-by-chunk reading

src/zarr/zarrista/_backend.py
kylebarron: Although I guess that does force you into the async backend, and maybe you want to focus on sync operations for now.

src/zarr/zarrista/_backend.py
```diff
@@ -0,0 +1,171 @@
...
+
+    Limitations:
+
+    - Only `LocalStore` is ingestable today (maps to a zarrista
+      `FilesystemStore`). obstore- and icechunk-backed stores are future work.
```
kylebarron: And fwiw it should be one line of code to pass an icechunk session, though that's also async

src/zarr/zarrs/_backend.py
```diff
@@ -0,0 +1,150 @@
...
+def _translate_errors() -> Iterator[None]:
+    try:
+        yield
+    except _zb.NodeNotFoundError as err:
+        raise NodeNotFoundError(str(err)) from err
+    except _zb.NodeExistsError as err:
+        raise NodeExistsError(str(err)) from err
+
+
+class ZarrsBackend:
```
kylebarron: I'd recommend subclassing from `CrudBackend`. That gives the reader more information over the intent here (for `ZarrsBackend` to implement `CrudBackend`) as well as telling the type checker that it should give you type errors if you don't accurately implement `CrudBackend`

d-v-b: good idea, I'll go with that

changes/+zarrista-backend.feature.md
d-v-b: :robot: _AI text below_ :robot:

No reason beyond increment size — `LocalStore` was the smallest store that let the differential suite run against the other backends. `zarr.storage.ObjectStore` exposes the underlying obstore store as `.store`, and the `CrudBackend` contract is already async, so the natural next step is mapping it onto zarrista's async API (`AsyncArray.open_async(store.store, ...)`). Planned as a follow-up, and an icechunk `Session` mapping has the same shape.

src/zarr/zarrista/_backend.py
d-v-b: :robot: _AI text below_ :robot:

Checked against `0.1.0b5`: yes, the leading slash is required today — `Array.open(store, "a")` raises `ValueError: invalid node path a`, and `"/a/"` (trailing slash) fails too; only `"/a"` opens. That matches zarrs' `NodePath`, but zarr-python's convention is store-key style (`""` for the root, `"a/b/c"` for children), hence this shim. If zarrista accepted bare `a/b/c` and normalized internally, this adapter would disappear — weak vote for that from the interop side, but the shim is cheap either way.

src/zarr/zarrista/_backend.py
d-v-b: :robot: _AI text below_ :robot:

Yes — and the async constraint is actually a non-issue here: the `CrudBackend` methods are all `async def` already. The sync zarrista API + `asyncio.to_thread` in this file was just the smallest first increment for `LocalStore`; the obstore path would use `zarrista.AsyncArray` natively with no thread offload, which is a better fit for the contract, not a worse one. `.store` on `zarr.storage.ObjectStore` is the unwrap.

src/zarr/zarrista/_backend.py
d-v-b: :robot: _AI text below_ :robot:

Reads are not chunk-by-chunk: `read_subset` hands the whole region to zarrista's `retrieve_array_subset`, so zarrista owns the multi-chunk read decomposition. Writes are chunk-level today because the CRUD write contract is `write_chunk` — the shared facade does the region decomposition and read-modify-write for partial boundary chunks in Python.

Agreed the backend should be able to own more of this. The natural evolution is an optional backend-level `write_region`, with the facade falling back to chunk-wise RMW for backends that don't provide it. zarrista would need a region-write primitive to exploit that (the stubs expose `store_chunk`/`store_encoded_chunk` but no subset write) — is exposing zarrs' `store_array_subset` something you'd entertain on the zarrista side?

src/zarr/zarrista/_backend.py
d-v-b: :robot: _AI text below_ :robot:

Good to know. Once the obstore/async mapping lands, a `Session` mapping is the same shape — keeping it in scope for that follow-up. (One caveat from the zarrista docs worth flagging: sessions backed by `in_memory_storage()` won't work since the Rust side reconstructs the session as a separate instance, so tests will need file- or S3-backed storage.)

src/zarr/zarrs/_backend.py
d-v-b: :robot: _AI text below_ :robot:

Done in be59fb4a9 (and 6c76610d3 for `ZarristaBackend`): all three backends now subclass `CrudBackend` explicitly, and the protocol docstring notes that conformance is checked at the definition site.

changes/+zarrista-backend.feature.md
d-v-b: :robot: _AI text below_ :robot:

Done in 6c76610d3: `zarr.storage.ObjectStore` now unwraps to its inner obstore store and runs through `zarrista.AsyncArray` — natively async, no thread offload. Verified round-trips (chunk read/write + cross-chunk subset) against an obstore `LocalStore`, cross-checked with zarr-python and the reference backend.

One empirical finding: zarrista `0.1.0b5` rejects an obstore `MemoryStore` at open time with `TypeError: expected an async compatible storage object`, while other obstore stores pass through fine — so the backend rejects memory-backed `ObjectStore`s at its gate for consistency. Is that expected on the zarrista side, or worth an issue?

kylebarron: A `MemoryStore` won't work because the underlying memory can't be extracted

src/zarr/zarrista/_backend.py
kylebarron: @d-v-b in terms of API design, should Zarrista support strings without a leading `/` or require it?

src/zarr/zarrista/_backend.py
kylebarron: I think separately there's a question of _should_ the `CrudBackend` methods all be async?

d-v-b: yeah that's a good q, i'm not sure. i suspect we ultimately want async and sync versions, because the best choice is genuinely contingent.

src/zarr/zarrista/_backend.py
d-v-b: this is actually a subtle point in the spec. the spec states that, in the _hierarchy representation_, [node names are absolute "/"-delimited paths](https://github.com/zarr-developers/zarr-specs/blob/dc3e95ed36060d9533361364ab7f54fe3e53f82b/docs/v3/core/index.rst?plain=1#L214-L229). But when you address the location of an array / group _inside a storage backend_, IMO it's more clear to use a relative path, i.e. no leading "/" character.

src/zarr/zarrista/_backend.py
```diff
@@ -0,0 +1,225 @@
...
+        start: Sequence[int],
+        shape: Sequence[int],
+    ) -> bytes:
+        target = _resolve_store(store)
+        np_dtype = _native_dtype(metadata)
+        selection = tuple(slice(s, s + length) for s, length in zip(start, shape, strict=True))
+        if isinstance(target, zarrista.FilesystemStore):
+
+            def _read() -> bytes:
+                array = zarrista.Array.open(target, _node_path(path))
```
kylebarron: Here you should use `Array.from_metadata` because you already have the metadata available. That will avoid reading the metadata again from source.

src/zarr/zarrista/_backend.py
```diff
@@ -0,0 +1,225 @@
...
+                return _to_bytes(array.retrieve_array_subset(selection), np_dtype)
+
+            return await asyncio.to_thread(_read)
+        array = await _open_async(target, path)
```
kylebarron: The above comment means that this is no longer async to open, whenever you already have the metadata (ditto to all the other methods in the backend as well)

src/zarr/zarrista/_backend.py
```diff
@@ -0,0 +1,225 @@
...
+
+def _native_dtype(metadata: Mapping[str, JSON]) -> np.dtype[Any]:
+    """Numpy dtype in native byte order, matching the facade's reassembly."""
+    return parse_array_metadata(metadata).dtype.to_native_dtype().newbyteorder("=")
+
+
+def _to_bytes(decoded: Any, np_dtype: np.dtype[Any]) -> bytes:
+    """Reinterpret a zarrista decoded array as C-contiguous native-dtype bytes."""
+    arr = np.asarray(decoded.to_numpy(), dtype=np_dtype)
+    return np.ascontiguousarray(arr).tobytes()
```
kylebarron: It's important to note that because `bytes` is the contract you have, this **necessarily copies**.

In zarrista I'm exposing [`Tensor`](https://developmentseed.org/zarrista/v0.1.0-beta.6/api/decoded_array/#zarrista.Tensor) as a core class to minimize data copies. So numpy-compatible fixed-width dtypes can directly be exposed from rust memory without a copy

d-v-b: `decoded: Any` 🙃 we can definitely do better than this; this operation should absolutely be zero-copy. I'll look into wrapping the tensor class!

kylebarron: > `decoded: Any` 🙃

What's your point here?

d-v-b: just that the type annotation is wrong -- the decoded object isn't `Any`

d-v-b:
This branch will be re-directed along the following lines:
- no rust code written here. we defer entirely to `zarrista` for the `zarrs` bindings.
- no more crud API. instead, we define `ArrayEngine` and `AsyncArrayEngine` protocols. These protocols define the behavior of an object that supports the operations necessary for working with arrays in zarr python. We define an implementation of these protocols using the existing python code as well as the `zarrista` array class.

d-v-b:

## Refactored against current `zarrista` + `zarr-indexing`

Force-pushed: this branch is now rebuilt on current `main`. The previous history is preserved at [`backup/zarrs-bindings-pre-refactor`](https://github.com/d-v-b/zarr-python/tree/backup/zarrs-bindings-pre-refactor). Both dependencies moved underneath this PR since it was opened.

Diff went from **+10,906 / −642 across 49 files** to **+3,705 / −133 across 40 files**.

### zarrista was broken, not just outdated

Three things failed on *every* call against `0.1.0-rc.1`:

| This PR (before) | Current zarrista |
|---|---|
| `self._arr.store` | renamed `.storage` — `.store` no longer exists |
| `type(x).__name__ != "Tensor"` → raise | concrete class is `FixedLengthTensor`; `Tensor` is now a union alias, so this rejected every read |
| `zarr_metadata.ArrayMetadataV3` | renamed `ZarrV3ArrayMetadataJSON` |

Both hand-rolled read-modify-write chunk loops (~40 lines each) are gone — `store_array_subset` does that natively now. Also picked up: `to_numpy()` on all tensor types, and `MemoryStore`/`ZipStore` as valid sync stores.

### zarr-indexing replaced the hand-rolled selection code

`_normalize.py` (213 lines) and its 140-line test are deleted. zarrista natively accepts only ints, step-1 slices and `Ellipsis` — exactly `UnitStepReader`'s contract — so `LazyArray` supplies the full NumPy dialect for free.

The engine went from **basic indexing only** to basic, orthogonal, coordinate, mask and block, reads *and* writes, including strided and negative-step.

This also removes the accepted regression. Serving every selection via its bounding box forced a `test_accessed_chunks` relaxation; partitioning by the chunk grid means `oindex[[0, 999]]` reads two chunks, not a thousand rows. **The indexing suite now passes unmodified.**

### Two design changes

**The boundary carries a `SelectionRequest` (raw selection + dialect), not a `Region` or an `IndexTransform`.** I had written the transform-from-indexer adapter before finding that `CoordinateIndexer.selection` is stored *chunk-sorted* (`indexing.py:1300`) — deriving a transform from it would have silently reordered `vindex` results. Deriving both views from the raw selection avoids that, and keeps `zarr_indexing` out of `zarr/core` entirely, so it stays an optional dependency.

**`_default.py` delegates instead of duplicating.** It forwards to `main`'s module-level `_get_selection`/`_set_selection`, so the built-in engine is the identity engine by construction. The old copy had already silently drifted from `main` (#3885, #4205, the v2/v3 `metadata.dtype` unification) — invisible to git, since it is a new file. `array.py` grew **+317 lines instead of +1446**.

### Three bugs found and fixed

Each caught by something executable, not by review:

1. **The async engine deadlocked** — it called `sync()` from inside a coroutine already on the loop. zarrista's pyo3 futures bind to the loop at *creation*, so calls are now deferred onto the owning loop and the `LazyArray` resolves on a worker thread.
2. **The built-in engine rebuilt the codec pipeline** `AsyncArray` had already built — double work per array, and it double-fired the sharding advisory warning, breaking `test_pipeline_parity`.
3. **Scalar writes** broadcast to 0-stride arrays that NumPy reports as C-contiguous but zarrista rejects.

### Verification

- **7196 passed, 0 failures** (full suite, including 134 engine/zarrista tests)
- ruff, mypy, and all pre-commit hooks clean
- Differential suite extended to 17 read + 13 write cases × both engines, with numpy as oracle
- The `open_with_engine` example runs against both engines

Dropped the ~6,500 lines of `docs/superpowers/` planning docs, which described the superseded design.

### Not done

**The ~15x benchmark is unverified against this path.** The write path in particular is structurally different (single `store_array_subset` for dense boxes, pointwise chunk-grouped scatter otherwise) — worth re-running before trusting the number in the description above, which still describes the original design.

Follow-ups, deliberately out of scope: `list_engines()` is backed by a `Literal` rather than an entry-point group, so third-party engines can only be passed as instances; and the hierarchy cache in `_resolve.py` is keyed on `id(store)`.

kylebarron:
> * **The async engine deadlocked** — it called `sync()` from inside a coroutine already on the loop. zarrista's pyo3 futures bind to the loop at _creation_, so calls are now deferred onto the owning loop and the `LazyArray` resolves on a worker thread.

This is interesting. I don't understand it well but I'm curious if it's something that zarrista could change. It might also be a limitation of https://github.com/PyO3/pyo3-async-runtimes/

kylebarron:
> 3\. **Scalar writes** broadcast to 0-stride arrays that NumPy reports as C-contiguous but zarrista rejects.

Feel free to open an issue for this

PR #4066, opened 2026-06-16
harden developer environment

d-v-b:
We have failing tests caused by upstream breakage in pytest. This PR pins our true development dependencies (pytest, etc) to exact versions. That will shield us from breaking changes in pytest.

Library dependencies that are installed for tests are left unpinned, because we want to catch breaking changes for these packages in our test suite.


d-v-b:
the changes in this PR:
- make our CI work
- do not alter shipped code

so I'm going to bypass review and merge as soon as tests are green. I will also omit the release notes, because this is a purely internal change.

PR #4074, opened 2026-06-16
Claude/storage codec bugfixes

d-v-b:
_I had claude do a comprehensive review of zarr-python and look for bugs, optimizations, etc. The [tracking issue](https://github.com/d-v-b/zarr-python/issues/181) is over on my fork. I will be opening claude-authored PRs that fix the issues._

## Summary

This PR addresses five pre-verified bugs (B1–B5 of [#181](https://github.com/d-v-b/zarr-python/issues/181)):

**B1 — MemoryStore SuffixByteRequest clamp** (`src/zarr/storage/_utils.py`)
When `SuffixByteRequest.suffix > len(data)`, the computed `start` index went
negative, causing numpy slicing to return too few bytes. Fixed by clamping:
`start = max(0, len(data) - byte_range.suffix)`. This matches how `ZipStore`
and `LocalStore` already handle this case.

**B2 — LoggingStore.get_partial_values generator exhaustion** (`src/zarr/storage/_logging.py`)
The log-hint string was built by iterating over `key_ranges` with a list
comprehension, which exhausted a one-shot generator *before* passing it to
the wrapped store. Fixed by materialising `key_ranges = list(key_ranges)`
first.

**B3 — getsize_prefix substring over-count** (`src/zarr/abc/store.py`)
`getsize_prefix("foo")` called `list_prefix("foo")`, which matched both
`"foo/a"` and `"foobar/b"` in prefix-matching stores. Fixed by normalising
the prefix to end with `"/"` (matching `delete_dir` / `is_empty`), so only
keys under the exact directory are counted.

**B4 — ZipStore.close() AttributeError when never opened** (`src/zarr/storage/_zip.py`)
`_lock` is only created in `_sync_open()`. If `close()` was called on a
store that was never opened (e.g. `with ZipStore(...): pass`), it raised
`AttributeError: _lock`. Fixed by returning early from `close()` when
`_is_open` is `False`.

**B5 — Missing raise in codecs_from_list** (`src/zarr/core/codec_pipeline.py`)
In the `BytesBytesCodec` branch, when `prev_codec` is an `ArrayArrayCodec`,
the error message was assigned to `msg` but never raised. The sibling
branches (`ArrayArrayCodec` and `ArrayBytesCodec` cases) both do
`raise TypeError(msg)`; this one was missing the `raise`. Fixed by adding
`raise TypeError(msg)`.

## Notes

- A towncrier changelog entry should be added before un-drafting this PR.
- Addresses B1–B5 of [#181](https://github.com/d-v-b/zarr-python/issues/181).


d-v-b:
@dcherian do you have any ideas for better property-based tests that we could add, given these bugs?

tests/test_store/test_utils.py
```diff
@@ -95,3 +97,32 @@ def test_drive_letter_not_special_on_non_windows(self, url: str) -> None:
             result = parse_store_url(url)
         # urlparse interprets the drive letter as a scheme
         assert result.scheme == "c"
+
+
+class TestNormalizeByteRangeIndex:
```
dcherian, line 102: Looks like a missing property test

dcherian (approved):

tests/test_store/test_utils.py
d-v-b: :+1: I will add one

pyproject.toml
```diff
@@ -393,7 +393,7 @@ show_error_code_links = true
 show_error_context = true
 strict = true
 warn_unreachable = true
-enable_error_code = ["ignore-without-code", "redundant-expr", "truthy-bool"]
+enable_error_code = ["ignore-without-code", "redundant-expr", "truthy-bool", "truthy-iterable"]
```
d-v-b, line 396: flagging this change to our mypy config. the fsspecstore codebase was relying on `if Iterator: ...` which I think we should avoid. so now mypy helps us avoid it.

PR #4076, opened 2026-06-17
chore: CI worfklows for linting docs

maxrjones:
This PR adds a few scripts/workflows to:

- Check that all top-level functions are included in the mkdocs API documentation
- Check that docstrings use mkdocs rather than sphinx styles admonitions and links
- Check that all links resolve

TODO:


maxrjones:
@d-v-b @chuckwondo this adds a fair amount of config/workflow files, so a review on the direction here would be welcome but not necessarily required

ci/check_documented_exports.py
```diff
@@ -0,0 +1,131 @@
...
+    Scans the indented option block immediately following the `:::` line, stopping at
+    the first non-indented line (the end of this directive's block)."""
+    for line in text[directive_start:].splitlines()[1:]:
+        if line.strip() == "":
+            continue
+        if not line.startswith((" ", "\t")):
+            break
+        if MEMBERS_DISABLED_RE.match(line):
+            return True
+    return False
```
chuckwondo: This is perhaps more readable:

```suggestion
    return any(
        MEMBERS_DISABLED_RE.match(line)
        for line in text[directive_start:].splitlines()[1:]
        if line.startswith((" ", "\t")) and line.strip()
    )
```

However, I cannot tell (yet, I'll keep reviewing) if `text` is the entirety of a python file, in which case, slicing it from some point to the end of the file for every directive encountered in the file seems highly inefficient (but perhaps isn't a noticeable performance problem, or I need to keep looking at your code to see this is not the case).

Although, I suspect that the suggestion above doesn't address the potentially larger issue of how this script is architected. In the end, however, this might be sufficient since this is simply a CI script, but I'll make more significant suggestions elsewhere.

ci/check_documented_exports.py
```diff
@@ -0,0 +1,131 @@
...
+import importlib
+import re
+from pathlib import Path
+from types import ModuleType
+from typing import Any
+
+import zarr
+
+REPO_ROOT = Path(__file__).parent.parent.resolve()
+API_DOCS_ROOT = REPO_ROOT / "docs" / "api"
```
chuckwondo: I suggest you make this a CLI argument instead of hard-coding it here. Then modify the github workflow to call this script with "docs/api" as the argument since it will be invoked from the root of the repo.

ci/check_documented_exports.py
```diff
@@ -0,0 +1,131 @@
...
+            documented.add(id(obj))
+            if isinstance(obj, ModuleType) and not members_disabled(text, match.start()):
+                member_names = getattr(obj, "__all__", None) or [
+                    name for name in dir(obj) if not name.startswith("_")
+                ]
+                for name in member_names:
+                    member = getattr(obj, name, None)
+                    if member is not None:
+                        documented.add(id(member))
+    return documented
```
chuckwondo: This is nearly indecipherable, so it may be annoying to maintain.

I suggest refactoring this to take a pipeline approach, which I believe will not only make it more readable, but also allow for greater efficiency/speed (related to my comment on the `members_disabled` function), which seems to be a motivating factor based on your comments elsewhere.

A pipeline approach should allow you to avoid repeatedly splitting each file's text into individual lines every time a directive is encountered.

ci/check_documented_exports.py
```diff
@@ -0,0 +1,131 @@
...
+
+def find_undocumented_exports() -> list[str]:
+    documented = documented_object_ids()
+    missing = []
+    for name in zarr.__all__:
+        if name in EXEMPT_EXPORTS:
+            continue
+        if id(getattr(zarr, name)) not in documented:
+            missing.append(name)
+    return sorted(missing)
```
chuckwondo: Use the ~force~comprehension, ~Luke~Max!

```suggestion
    return sorted(
        name
        for name in zarr.__all__
        if name not in EXEMPT_EXPORTS
        and id(getattr(zarr, name)) not in documented
    )
```

ci/check_documented_exports.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+def main() -> None:
+    if not API_DOCS_ROOT.exists():
+        raise FileNotFoundError(f"{API_DOCS_ROOT} does not exist.")
```
chuckwondo: As I mentioned in another comment, take the docs root as an argument, rather than hard-coding it.

ci/check_documented_exports.py
```diff
@@ -0,0 +1,131 @@
...
+    lines.append(
+        "\nAdd a `::: zarr.<name>` page under docs/api/zarr/ (and register it in "
+        "mkdocs.yml and docs/api/zarr/index.md), or -- if the export is intentionally "
+        "undocumented -- add it to EXEMPT_EXPORTS in this script with a reason."
+    )
+    raise ValueError("\n".join(lines))
+
+
+if __name__ == "__main__":
+    main()
```
chuckwondo: I suggest following the pattern you used in `lint_docs.py`, where you print to stderr and use a non-zero exit code, rather than raising.

ci/lint_docs.py
```diff
@@ -0,0 +1,258 @@
...
+        if not docstring:
+            continue
+        # node.body[0].value is the docstring literal; its lineno is the line the string
+        # opens on, so content line i maps to source line (start + i).
+        start = node.body[0].value.lineno  # type: ignore[attr-defined]
+        for offset, line in enumerate(docstring.splitlines()):
+            findings.extend(
+                Finding(path, start + offset, category, line) for category in _scan_line(line)
+            )
+    return findings
```
chuckwondo: Haven't tested this, but I would argue this is more readable. I encourage you to embrace comprehensions over nested for loops with embedded conditionals.  While the logic is essentially the same, the comprehension syntax tends to be more readable (even more natural-language-like).

To avoid having the comprehensions themselves become incomprehensible, I've also created 2 comprehensions in place of the single nested for loops.

There are several other places in this script where I encourage you to do similar refactorings to aid in readability and maintainability.

```suggestion
    doc_nodes = (ast.Module, ast.ClassDef, ast.FunctionDef, ast.AsyncFunctionDef)
    docstring_start_pairs = [
        # node.body[0].value is the docstring literal; its lineno is the line the string
        # opens on, so content line i maps to source line (start + i).
        (docstring, start = node.body[0].value.lineno)
        for node in ast.walk(tree)
        if isinstance(node, doc_nodes)
        if docstring := ast.get_docstring(node, clean=False)
    ]
    
    return [
        Finding(path, start + offset, category, line)
        for docstring, start in docstring_start_pairs
        for offset, line in enumerate(docstring.splitlines())
        for category in _scan_line(line)
    ]
```

chuckwondo (changes requested):

maxrjones:
thanks for the review @chuckwondo. I applied your fixes in [`fd36246` (this PR)](https://github.com/zarr-developers/zarr-python/pull/4076/commits/fd36246b1e09b73071e6b2930b37101fbd72e97d) and some other readability improvements. I chose to make the directory an optional argument.

d-v-b (approved):

PR #4083, opened 2026-06-19
feat![zarr-metadata]: extensive top-level exports

d-v-b:
Adds a bunch of top-level exports to zarr-metadata. It ensures that you can import everything you need for basic zarr metadata operations from the outermost namespace.

This change also renames many types, to ensure that they are unambiguous in the same namespace. This will be a breaking change to zarr-metadata.


d-v-b:
head's up: I am going to self-merge this quickly, because I think I am the only person who cares about this package right now.

PR #4085, opened 2026-06-21
Implement ZipStore delete via archive rewrite (fixes #828)

### LINKED ISSUE #828: Implementing `ZipStore`'s `__delitem__` via overwrite

Akash-t25:
## Summary

Implements `delete()` and `delete_dir()` for `ZipStore`, resolving #828.

ZIP files don't support in-place deletion, so this PR uses an archive-rewrite strategy: copy every surviving member into a fresh temporary ZIP, drop the deleted entries, then atomically swap it in using `os.replace`.

### `src/zarr/storage/_zip.py`
- Flipped `supports_deletes: bool = False` → `True`
- Added `_rewrite_without(should_delete)` private helper that rewrites the archive atomically
- Rewrote `delete(key)` — removes the key (missing key = no-op) instead of raising `NotImplementedError`
- Rewrote `delete_dir(prefix)` — single archive rewrite dropping everything under the prefix

### `tests/test_store/test_zip.py`
- Updated `test_api_integration` — chunk deletion and `del root["bar"]` now succeed
- Added `test_store_supports_deletes`, `test_delete_compacts_duplicates`, `test_delete_then_set`
- Inherited `StoreTests.test_delete`, `test_delete_dir`, `test_delete_nonexistent_key_does_not_raise` now run and pass

### `tests/test_codecs/test_sharding.py`
- Added `filterwarnings("ignore:Duplicate name")` marker for the now-active zip case

### `changes/828.feature.md`
- Changelog entry in towncrier format

## Test Results
- `test_zip.py`: 74 passed, 5 skipped
- `test_sharding.py` (zip): 49 passed
- `test_core.py` + `test_api.py` (zip): 28 passed, 3 skipped

Note: `*_sync` delete tests are skipped — `ZipStore` does not implement `SupportsDeleteSync`, which is out of scope for this PR.

## Checklist

Closes #828


d-v-b:
I worry that this will have counter-intuitive performance characteristics, especially since we don't have a transactional API that would support batching many deletes into one atomic archive resurrection. IF we want to support this delete implementation, it should probably be exposed via runtime store configuration. 

cc @mkitti

Akash-t25:
Hi @d-v-b, thank you for the feedback! I completely understand the concern — rewriting the entire archive on every delete is O(n) in archive size, which could be very expensive for large stores.

I'd like to implement this the right way. Could you point me to an example of "runtime store configuration" in the zarr codebase? I want to make sure I follow the existing pattern rather than inventing something new. 

For example, would something like an `allow_deletes: bool = False` parameter on `ZipStore.__init__` be in the right direction, or is there a different configuration mechanism you have in mind?

Happy to iterate on this — just want to make sure I'm aligned with the project's conventions before making changes.

PR #4090, opened 2026-06-22
chore: dependabot only updates direct dependencies

d-v-b:
I don't think we need dependabot to update our transitive dependencies.


d-v-b:
This change does not affect our code, and we can easily revert. I will self-merge as soon as tests are passing.

PR #4094, opened 2026-06-22
chore: test correctness fixes

d-v-b:
_claude-generated fixes of claude-discovered issues in our test suite_

### Downstream `gpu` marker was never registered
`src/zarr/testing/conftest.py` defined `pytest_configure` to register the `gpu` marker, but the `pytest11` entry point is `zarr.testing` (`src/zarr/testing/__init__.py`), which has no `pytest_configure` — so downstream projects consuming `StoreTests` under `--strict-markers` never got the marker.
- Add `pytest_configure` to `src/zarr/testing/__init__.py` (registers `gpu`, comment kept).
- Delete the now-dead `src/zarr/testing/conftest.py` and its `[tool.coverage.run] omit` entry.
- Import `assert_bytes_equal` only when pytest is available (it imports pytest), matching `StoreTests`.

zarr-python's own suite still registers `gpu` via the main `markers` list, so local behavior is unchanged. Verified the marker now appears via the plugin in an isolated empty directory.

### Pin real exception messages on broad `pytest.raises`
Added verified `match=` regexes to ~16 high-value broad-builtin `pytest.raises` in `tests/test_api.py`, `tests/test_group.py`, and `tests/test_store/test_core.py`. Every message was confirmed against the actual code, not guessed.

### Remove silently no-op assertions
In `tests/test_experimental/test_cache_store.py`, dropped the `if hasattr(cached_store, "_is_key_fresh")` guards (the method exists on `CacheStore`) so the TTL/freshness checks always run.

### xfail/skip hygiene
Added `reason=` to two reason-less `xfail` params in `tests/test_indexing.py`. The env-specific unconditional `skip` on `test_setitem_repeated_index` was left as-is (repeated-index assignment has undefined ordering and fails broadly, so no correct `skipif` exists).

### Verification
- `pytest tests/test_api.py tests/test_group.py tests/test_store/test_core.py tests/test_experimental/test_cache_store.py tests/test_indexing.py`: **1413 passed, 25 skipped, 5 xfailed**.
- Full-suite collection: **6479 tests**, no errors. ruff/mypy pre-commit pass.

Part of [#195](https://github.com/d-v-b/zarr-python/issues/195)


maxrjones (approved):
thanks!

PR #4096, opened 2026-06-24
chore: replace hatch environments with uv + ci scripts

maxrjones:
This PR switches to using uv + just + hatchling rather than uv + hatch + hatchling.

For context, I think it's sub-optimal to have environments managed by both hatch (for CI and in contributing guide) and uv (for typing). This PR shows how we could use only uv via a small set of bash scripts that make up for the missing task runner and environment names features in uv relative to hatch. The benefit is that all of our CI and developer workflows would be based on the exact same lock files for improved reproducibility.

I tried hatch's new lockfile support first, but it's platform specific so it doesn't help us much given we want cross-platform consistency.

I don't feel strongly about this PR, so totally fine being told that we should stick with the status quo.

TODO:


d-v-b:
i can understand how some folks might not like the churn but I'm +1 on this. we can always go back to hatch if we don't like plain uv.

d-v-b:
@maxrjones did you consider putting some of these scripts into a runner like a [justfile](https://github.com/casey/just)?

maxrjones:
> @maxrjones did you consider putting some of these scripts into a runner like a [justfile](https://github.com/casey/just)?

I considered nox/tox, but decided those were overkill after a chat with @chuckwondo. Just might be a better fit

d-v-b:
i mention `just` or something similar because I _do_ think there's real value in a one-stop place for the verbs we need as developers. hatch never really delivered on this, but I think the goal is good. See https://github.com/earth-mover/icechunk/blob/main/Justfile

pyproject.toml
```diff
@@ -30,10 +30,14 @@ maintainers = [
...
 requires-python = ">=3.12"
+# The `>=` lower bounds below (and in [project.optional-dependencies]) are the
+# minimum versions we support. The `min_deps` CI job tests them via the matching
+# pins in `ci/min-deps-constraints.txt` — keep the two in sync when you bump a
+# floor (per the SPEC 0 drop schedule: https://scientific-python.org/specs/spec-0000/).
 dependencies = [
     'packaging>=22.0',
     'numpy>=2',
-    'numcodecs>=0.14',
+    'numcodecs>=0.14.1',
```
maxrjones, line 40: Zarr-Python was actually not compatible with numcodecs 0.14.0; the lowest supported version is 0.14.1 in reality due to the need for the soft dependency fixes in https://github.com/zarr-developers/numcodecs/releases/tag/v0.14.1.

ilan-gold:
At first glance I'm opposed to it because now we're maintaining a script.  There's also now a cool vscode plugin for hatch environments: https://github.com/pypa/hatch-code

I will set this to unread and look later though

d-v-b:
would it make sense to keep the justfile and defer the "drop or keep hatch" decision? I think a single file for all the verbs is a nice idea on its own

PR #4108, opened 2026-06-29
docs: document authentication for private remote stores via storage_options

### LINKED ISSUE #2995: DOC: Show how to authenticate with `fsspec`'s `storage_options` dict

In the [Zarr storage guide](https://zarr.readthedocs.io/en/latest/user-guide/storage.html) there is an example showing how to connect to an S3 bucket anonymously and read data. This is great! Yet I think a lot of people experimenting with this feature on their local machines using `minio` will have problems authenticating and testing using this example. I've combed through massive amounts of `fsspec` documentation yet I can't find what keys are acceptable for this in `storage_options`.

I've tried:
```
import zarr
import numpy as np

store = zarr.storage.FsspecStore.from_url(
    'http://localhost:9000',
    storage_options={
        'key': 'minioadmin',
        'secret': 'minioadmin'
        }
)
zarr.open_group(store=store, mode='w')
```
along with:
```
storage_options={
    'username': 'minioadmin',
    'password': 'minioadmin'
}
```
I know I can get a solution working with `s3fs`:
```
import zarr
import s3fs
import numpy as np

bucket = "buck"
name = "sample"

s3 = s3fs.S3FileSystem(
    client_kwargs={"endpoint_url": "http://localhost:9000"},
    key="minioadmin",
    secret="minioadmin",
    use_ssl=False,
)
s3store = s3fs.S3Map(root=f'{bucket}/{name}', s3=s3)
```
However, this is nowhere near as sleek and integrated as the Zarr docs make it seem.

We're looking to avoid errors like:
`TypeError: ClientSession._request() got an unexpected keyword argument 'secret' `
Suggest a way we can add documentation to let users know what keys are valid in `storage_options` or where auth should otherwise be specified.

deodharaditi:
## Summary

Closes #2995.

The storage user-guide only showed anonymous (`anon=True`) access examples for remote
stores, leaving users to guess which keys to put inside `storage_options` when accessing
a private S3 bucket or an S3-compatible service like MinIO. A common wrong guess — nesting
credentials inside `client_kwargs` — routes them to the underlying `aiohttp`/`aiobotocore`
`ClientSession`, which produces a cryptic `TypeError: ClientSession._request() got an
unexpected keyword argument 'secret'`.

This PR adds a short **Authentication** subsection under `### Remote Store` that:

1. Shows how to pass `key`, `secret`, `token`, and `endpoint_url` as **top-level** keys in
   `storage_options` (they map directly to `s3fs.S3FileSystem` arguments, not to the inner
   client session).
2. Explains with a `!!! note` admonition exactly why `client_kwargs` is the wrong place for
   credentials (aiohttp `ClientSession` does not accept them), so the connection between the
   error message and the fix is clear.
3. Notes that credentials can be omitted if the standard AWS auto-discovery path is used
   (env vars, `~/.aws/credentials`, instance/role profile).

The code example is a plain (non-executed) `python` block deliberately — the surrounding
examples use `exec="true"` (markdown-exec runs them at build time), but we cannot connect
to a private bucket in CI.

## For reviewers

The main thing I'd like a second look at is whether the `s3fs.S3FileSystem` API description
is accurate and complete enough. I verified the top-level keys (`key`, `secret`, `token`,
`endpoint_url`) by inspecting the installed signature of `s3fs` v2026.4.0 in a local dev
environment, but I'd welcome a check from someone more familiar with the fsspec/s3fs
interaction. I'm confident the prose style matches the existing admonition usage in the doc,
and that the non-executed block won't break the CI build.

## Author attestation

- [x] I am a human, these are my changes, and I have reviewed and understood every change
  and can explain why each is correct.

## TODO

* [x] ~~Add unit tests and/or doctests in docstrings~~ — N/A, docs-only change
* [x] ~~Add docstrings and API docs for any new/modified user-facing classes and functions~~ — N/A
* [x] ~~Test coverage is 100% (Codecov passes)~~ — N/A, no code changed

### DIFF
--- a/changes/2995.doc.md
+++ b/changes/2995.doc.md
@@ -1,0 +1,1 @@
+Document how to authenticate to private remote stores (e.g. S3 / MinIO) by passing credentials through `storage_options`, and clarify that credentials such as `key` and `secret` are top-level keys rather than `client_kwargs` entries.
--- a/docs/user-guide/storage.md
+++ b/docs/user-guide/storage.md
@@ -151,6 +151,43 @@ store = zarr.storage.FsspecStore(fs)
 print(store)
 ```
 
+#### Authentication
+
+The examples above use anonymous access (`anon=True`) to a public bucket. To read from or write
+to a **private** bucket — for example a private AWS S3 bucket or a self-hosted, S3-compatible
+service such as [MinIO](https://min.io/) — pass your credentials through `storage_options`. These
+options are forwarded to the underlying fsspec filesystem (for S3 that is
+[`s3fs`](https://s3fs.readthedocs.io/)), so the accepted keys are the ones that filesystem
+understands; for S3 the most common are `key`, `secret`, `token`, and `endpoint_url`:
+
+```python
+import zarr
+
+store = zarr.storage.FsspecStore.from_url(
+    's3://my-private-bucket/path',
+    storage_options={
+        'key': '<ACCESS_KEY_ID>',
+        'secret': '<SECRET_ACCESS_KEY>',
+        # 'token': '<SESSION_TOKEN>',  # optional, for temporary credentials
+        # 'endpoint_url': 'https://my-minio.example.com',  # for S3-compatible services like MinIO
+    },
+)
+group = zarr.open_group(store=store, mode='r')
+```
+
+!!! note
+    Credentials are **top-level** keys in `storage_options`; they map directly to
+    [`s3fs.S3FileSystem`](https://s3fs.readthedocs.io/en/latest/api.html) arguments. Do not nest
+    them inside `client_kwargs` — that dictionary is forwarded to the underlying
+    `aiobotocore`/`aiohttp` client session, which does not accept `key`/`secret` and will raise an
+    error such as `TypeError: ClientSession._request() got an unexpected keyword argument 'secret'`.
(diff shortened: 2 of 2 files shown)


deodharaditi:
Hi, this is my first contribution. I've added an Authentication subsection to the storage guide to address #2995. Would appreciate a review when you have time!

d-v-b:
@deodharaditi thanks for this addition to the docs. There's a check in our tests that's failing -- it's designed to ensure that every code example in the docs is annotated as runnable as part of the docs build and / or checkable as part of our doctests pass. This example is probably neither, unless you want to go the extra mile to configure our test-only s3 backend to use credentials. So you need to annotate the code example with exec=false and a reason. See https://zarr.readthedocs.io/en/latest/contributing/#validating-code-blocks-exec-vs-test

deodharaditi:
Thanks for the heads-up and the pointer to the contributing guide! I've added exec="false" reason="requires private S3 credentials not available in the docs build environment" to the code block in [9be3d548](https://github.com/deodharaditi/zarr-python/commit/9be3d548). Let me know if the reason string needs adjustment or if there's anything else to address.

d-v-b:
thanks! I'm going to approve this, and ask @maxrjones to review, since he has a bit more experience using zarr + s3 than I do, and he might have some additional suggestions. Great work!

d-v-b (approved):

PR #4109, opened 2026-06-30
test: comprehensive eager indexing parity oracle + reusable indexing strategies

d-v-b:

Standalone test-infrastructure improvement extracted from the lazy-indexing work (#3906). It hardens the **existing (eager)** indexing API and has no dependency on the lazy `IndexTransform` work, so it can land independently. **No lazy-indexing tests are included here** — those live with the lazy PR; this PR only sets up the harness so that work can reuse it.

## What this adds

**Reusable Hypothesis strategy** (`zarr.testing.strategies`)
- `numpy_array_indexers(mode, shape)` → a `(zarr_selection, numpy_selection)` pair for any of `basic` / `oindex` / `vindex` / `mask` — the element-space modes that have a direct NumPy-array oracle — so an indexing test is written once and parametrized over mode. (Block indexing is deliberately out of scope: it addresses the chunk grid, not array elements, and has no NumPy equivalent; it keeps its own `block_indices` strategy.)

**NumPy-oracle parity tests** (`tests/test_properties.py`), folding the old per-mode tests into a small set:
- `test_indexing_read_parity` — every mode, for sync read, read into an `out=` buffer, and async read (adds `out=` coverage the per-mode tests lacked). The read oracle is dtype-strict (asserts dtype, not just values).
- `test_indexing_write_parity` — every mode; ill-defined writes (a cell hit more than once after negative-index normalization) are filtered via `_write_is_unambiguous`, and the known sharded multi-axis-`oindex` limitation is an `xfail`, not a silent skip.
- `test_indexing_bounds_error_parity` — out-of-bounds integer selections must raise in zarr iff they raise in NumPy (zarr's bounds errors subclass `IndexError`). Catches the silent-wrong-data class — returning garbage where NumPy would raise.
- `test_write_is_unambiguous` — regression anchor for the duplicate-target filter.

**Stateful model-lockstep harness** (slow-gated; runs in the Slow Hypothesis job):
- `_IndexingStateMachine` applies many random reads/writes to one array while a NumPy model tracks it in lockstep, asserting equality after every step — catching read-modify-write / chunk-boundary / shard-merge bugs the single-shot tests can't see. Runs over regular, sharded, and 3-D geometries.

These borrow the strongest ideas from TensorStore's `driver_testutil` (model lockstep) and ndindex's `check_same` (error parity + dtype-strict comparison).

## Forward-compatibility (no lazy dependency)
The read oracle is consumer-agnostic — `assert_read_matches_numpy(target, ...)` where `target` is any `zarr.Array` — so once lazy indexing lands, the lazy PR can pass a lazy view through this exact harness instead of duplicating it.

Verified on `main`: full `tests/test_properties.py` green (16 passed, 3 skipped [the slow state machines], 1 xfailed); the Slow Hypothesis job exercises the state machines; ruff/mypy clean.


tests/test_properties.py
```diff
@@ -448,3 +328,164 @@ def test_array_metadata_meets_spec(meta: ArrayV2Metadata | ArrayV3Metadata) -> N
...
+    # write, mirroring the historical per-mode guards:
+    #   - vindex setitem is chunk-dependent for repeated coords (skipped)
+    #   - oindex/mask writes to sharded arrays are unsupported (GH2834)
+    #   - oindex writes with repeated indices are unspecified
+    if mode == "vindex":
+        return
+    if mode in ("oindex", "mask") and zarray.shards is not None:
+        return
+    if mode == "oindex" and _has_repeated_indices(npsel):
+        return
```
dcherian: this is why i didn't do this haha. 

From a quick glance I can't tell what makes it through.

d-v-b: yeah i hate this! i'm looking for something less dirty

src/zarr/testing/strategies.py
```diff
@@ -598,6 +599,68 @@ def orthogonal_indices(
...
+    return tuple(out)
+
+
+@st.composite
+def indexers(
+    draw: st.DrawFn, *, mode: IndexMode, shape: tuple[int, ...]
+) -> tuple[Selection, Selection]:
+    """A ``(zarr_selection, numpy_selection)`` pair for ``mode`` on ``shape``.
+
+    One strategy covering every indexing mode, so a test can be written once and
```
maxrjones: why leave off block indexing?

d-v-b: this strategy is designed to generate numpy array-compatible indexing routines, which doesn't include block indexing. block indexing examples are generated by a different strategy

d-v-b: I had the docstring updated to make this more clear

dcherian (commented):
to be honest, I find this extremely hard to read with all the extra indirection. The duplication in the previous version made it far more readable. And given that this is all agent-generated, why not just duplicated and optimize for readability?

tests/test_properties.py
```diff
@@ -448,3 +329,365 @@ def test_array_metadata_meets_spec(meta: ArrayV2Metadata | ArrayV3Metadata) -> N
...
+    shards: tuple[int, ...] | None = None
+
+    def __init__(self) -> None:
+        super().__init__()
+        n = int(np.prod(self.shape, dtype=int))
+        self.model = np.arange(n, dtype="i4").reshape(self.shape)
+        self.zarray = zarr.create_array(
+            {}, shape=self.shape, chunks=self.chunks, shards=self.shards, dtype="i4"
+        )
+        self.zarray[...] = self.model
```
dcherian: Why not use the generator here? and set `self.model = self.array[:]`

d-v-b:
> to be honest, I find this extremely hard to read with all the extra indirection. The duplication in the previous version made it far more readable. And given that this is all agent-generated, why not just duplicated and optimize for readability?

noted! the reason for the indirection is to facilitate testing of lazy indexing, which supports the numpy-style indexing semantics, and so requires the exact same set of tests. Without this indirection, I worried that we would have to write  `_lazy` versions of every indexing method test, and this invites mistakes.

d-v-b:
per @dcherian's recommendation, I'm going to go back to duplicated per-mode tests. We can deal with the cost of duplication (if any) when we start wiring up lazy indexing.

tests/test_properties.py
d-v-b: do you mean get our array values from a generator? I'm not sure that we want this for checking sequences of array indexing operations, since our assumption here is that indexing is insensitive to the scalar values. a deterministic sequence seems like a better set of initial values?

d-v-b:
per-indexing-mode tests added back in, now broken down separately for reads and writes.

d-v-b:
this is too complicated and probably pointless now that `zarr-indexing` is published. closing!

Issue #1806, opened 2024-04-22
Support for batched Store API in v3


akshaysubr:
The `Store` API in v3 supports fetching a single key at a time or partial values of multiple keys.
```python
class Store(ABC):
    @abstractmethod
    async def get(
        self, key: str, byte_range: Optional[Tuple[int, Optional[int]]] = None
    ) -> Optional[bytes]:
        ...

    @abstractmethod
    async def get_partial_values(
        self, key_ranges: List[Tuple[str, Tuple[int, int]]]
    ) -> List[bytes]:
        ...

```

The new `BatchedCodecPipeline`, for instance, currently fetches data for all requested chunks concurrently by using an asyncio thread pool with each task calling the `get` method of the `Store`:
```python
            chunk_bytes_batch = await concurrent_map(
                [(byte_getter,) for byte_getter, _, _, _ in batch_info],
                lambda byte_getter: byte_getter.get(),
                runtime_configuration.concurrency,
            )
```

Since there have been some concerns about scalability of asyncio for a large number of tasks, would it make sense to move this batched fetch into the `Store` itself? This would allow another `Store` implementation to potentially use a more performant asynchronous framework for the batched fetch, say in C++ or Rust, and can look like a single asyncio task to zarr-python.

This is a feature that currently exists in v2 through the [`getitems` Store API](https://github.com/zarr-developers/zarr-python/pull/1131) which is used to enable GPU Direct Storage in Zarr [through kvikIO](https://github.com/rapidsai/kvikio/blob/branch-24.06/python/kvikio/zarr.py#L129).

A similar feature is also now added (provisionally) to [v3 codecs](https://github.com/zarr-developers/zarr-python/blob/batched-codec-pipeline/src/zarr/v3/abc/codec.py#L84) that support an `encode_batch` and a `decode_batch` with a default implementation that ships tasks off to an asyncio thread pool but allows a codec to override that in favor or another approach if needed.

rabernat:
Do queues have a role to play here?

To me it makes sense that the store and the codec pipeline would communicate via a queue. The store puts compressed chunks into a queue. The codec pipeline pulls items out and processes them. Either one could do this either via single iterms or in batches. If the queue is full, the store could stop fetching chunks until there is room (backpressure).

jhamman:
I've also been thinking that a producer/consumer queue is the way forward here but I'm not clear at all where in the stack these queues would go. IMO, this is the part of the equation that is missing for us to take action here.

[...]

PR #4112, opened 2026-07-01
Add Store.get_many bulk fetch API

TomNicholas:
Adds a public, overridable `Store.get_many` for fetching many values — each a whole key or a `(key, byte_range)` pair — in a single call, so a store can batch/coalesce reads that land in the same underlying object instead of issuing one `get` per key.

It generalizes `Store.get_ranges` (many ranges of one key) to many keys, yielding `(request_index, Buffer | None)` batches in completion order. The ABC default is a concurrent fan-out over `get`; `FsspecStore` overrides it to coalesce via fsspec's `cat_ranges`. Coalescing tuning is left to each store rather than exposed on the interface.

Motivation — xref #1806 (batched Store API), #1758 (request coalescing), and zarr-developers/VirtualiZarr#947 (files-as-shards / consolidating small reads): a custom store such as VirtualiZarr's `ManifestStore` or icechunk's `IcechunkStore` can override `get_many` to merge many small chunk reads into fewer requests.

Draft — feedback on the signature welcome.


src/zarr/abc/store.py
```diff
@@ -237,6 +237,71 @@ async def get_partial_values(
...
 
+    async def get_many(
+        self,
+        requests: Sequence[tuple[str, ByteRequest | None] | str],
+        *,
+        prototype: BufferPrototype,
+    ) -> AsyncIterator[Sequence[tuple[int, Buffer | None]]]:
+        """Retrieve many values, possibly from different keys, at once.
+
+        This is the bulk counterpart to :meth:`get`: the whole set of requests
```
d-v-b, line 248: rst -> mkdocs

src/zarr/testing/store.py
```diff
@@ -407,6 +407,41 @@ async def test_get_partial_values(
...
+        ]
+        collected: dict[int, Buffer | None] = {}
+        async for batch in store.get_many(requests, prototype=default_buffer_prototype()):
+            for index, value in batch:
+                assert index not in collected  # reported exactly once
+                collected[index] = value
+
+        assert set(collected) == set(range(len(requests)))
+        assert collected[4] is None  # missing key -> None (not omitted)
+        expected = {0: b"aaaaa", 1: b"bb", 2: b"aa", 3: b"cccc"}  # index 2 is "aaaaa"[1:3]
```
d-v-b, line 432: if we define `expected` here, why don't we just assert that `expected == collected`?

d-v-b:
@TomNicholas ok if I push to your branch? that might move faster than our current approach

PR #4113, opened 2026-07-01
Use Store.get_many for whole-chunk reads in BatchedCodecPipeline

TomNicholas:
Builds on #4112. `BatchedCodecPipeline.read` now fetches a whole (non-sharded) request with a single `Store.get_many` call instead of one `get` per chunk, so a store can batch/coalesce the underlying reads — independently of `codec_pipeline.batch_size`, which still governs only decode batching.

The sharding codec's partial-decode path is unchanged, and stores without a specialized `get_many` fall back to the previous concurrent per-chunk behavior.

Motivation — xref #1758 (request coalescing), #1806 (batched Store API), and zarr-developers/VirtualiZarr#947 (files-as-shards / consolidating small reads).

Stacked on #4112 — its commit is the first one here; review after it. Draft.

### DIFF
--- a/src/zarr/abc/store.py
+++ b/src/zarr/abc/store.py
@@ -237,6 +237,71 @@ class Store(ABC):
         """
         ...
 
+    async def get_many(
+        self,
+        requests: Sequence[tuple[str, ByteRequest | None] | str],
+        *,
+        prototype: BufferPrototype,
+    ) -> AsyncIterator[Sequence[tuple[int, Buffer | None]]]:
+        """Retrieve many values, possibly from different keys, at once.
+
+        This is the bulk counterpart to :meth:`get`: the whole set of requests
+        is handed to the store in a single call, so an implementation can fetch
+        them together — for example by coalescing reads that land in the same
+        underlying object into fewer requests — rather than one at a time. It
+        generalizes :meth:`get_ranges` (which reads many ranges from a *single*
+        key) to many keys, each with an optional byte range.
+
+        Yields one batch per underlying I/O operation, each a sequence of
+        ``(request_index, Buffer | None)`` tuples where ``request_index`` is the
+        position of the request in ``requests``. Every request is reported
+        exactly once across all batches; a ``None`` buffer means that key is
+        absent. Batches arrive in completion order, not request order, so
+        callers use the indices to reassemble results.
+
+        The default implementation fetches each request concurrently with
+        :meth:`get`, so every store gets a working version for free; stores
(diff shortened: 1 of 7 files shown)


ilan-gold:
Can you highlight what the relation of this to https://github.com/zarr-developers/zarr-python/pull/3925 is?

TomNicholas:
Yes good question - `get_ranges` is only for a single key. 

For my [use case](https://github.com/zarr-developers/VirtualiZarr/issues/947#issuecomment-4859947806) (fetching virtual chunks that are separate keys but happen to be part of the same netCDF file object) I need a method for bulk fetching multiple keys. Then my store (`virtualizarr.ManifestStore`/`IcechunkStore`) can use its manifests at runtime to find which chunks actually happen to live close to each other in the same object, then fetch those in one coalesced request. See https://github.com/zarr-developers/VirtualiZarr/pull/1033 for an example implementation of `get_many`. I've seen 10x speedups for some queries with this addition.

src/zarr/abc/store.py
```diff
@@ -237,6 +237,71 @@ async def get_partial_values(
...
 
+    async def get_many(
+        self,
+        requests: Sequence[tuple[str, ByteRequest | None] | str],
+        *,
+        prototype: BufferPrototype,
+    ) -> AsyncIterator[Sequence[tuple[int, Buffer | None]]]:
+        """Retrieve many values, possibly from different keys, at once.
+
+        This is the bulk counterpart to :meth:`get`: the whole set of requests
```
d-v-b, line 248: rst-style docstring -> mkdocs-style docstring

src/zarr/abc/store.py
```diff
@@ -237,6 +237,71 @@ async def get_partial_values(
...
+
+        Yields
+        ------
+        Sequence[tuple[int, Buffer | None]]
+            One batch per underlying I/O operation, each a sequence of
+            ``(request_index, Buffer | None)`` tuples.
+        """
+        # Local imports to avoid an import cycle at module load time.
+        from zarr.core.common import concurrent_map
+        from zarr.core.config import config
```
d-v-b, line 289: what if the concurrency is a plain keyword-only parameter for this function?

ilan-gold:
> See https://github.com/zarr-developers/VirtualiZarr/pull/1033 for an example implementation of get_many. I've seen 10x speedups for some queries with this addition.

Ah nice! Now I get what you mean by "is handed to the store in a single call, so an implementation can fetch..."

PR #4116, opened 2026-07-02
fix: keep element order when writing F-contiguous chunks with vlen codecs

### LINKED ISSUE #3558: Wrong array write or read after writing numpy data converted from string[pyarrow]
### Zarr version

v3.1.3

### Numcodecs version

v0.16.3

### Python Version

3.12.7, 3.13.9

### Operating System

Windows

### Installation

pip into virtual environment

### Description

I have a pandas DataFrame storing string data with string[pyarrow] datatype.

When converting this data to numpy array with either "df.values" or "df.to_numpy()" it produces a numpy "object" array (dtype=="O").

When storing this array in a chunked zarr-array there seems to be a strange "off-by-one" error either when writing or when reading. The result of the array read back from zarr differs from the original array.

When the DataFrame was stored as normal (python) string, this does not happen even though the numpy array looks exactly the same.

### Steps to reproduce

```python
# /// script
# requires-python = ">=3.12"
# dependencies = [
#   "pandas",
#   "pyarrow",
# ]
# ///
#

import zarr
# your reproducer code
import numpy as np
import pandas as pd

# Number of rows and columns
num_rows = 11
num_cols = 11

# Generate incrementing strings
incrementing_strings = [f"S{i:05}" for i in range(1, num_rows * num_cols + 1)]

npdata = np.array(incrementing_strings, dtype="str").reshape(num_rows, num_cols)

df = pd.DataFrame(npdata)
dfpa = df.astype("string[pyarrow]")

data = df.values
datapa = dfpa.values

# According to numpy, the converted data is identical:
assert data.dtype == datapa.dtype  # dtype == 'O'
assert data.shape == datapa.shape
assert data.size == datapa.size
assert np.all(data == datapa)

# Now store both in a zarray (using the same array here but doesn't matter if I use a fresh one)
store = zarr.storage.MemoryStore()  # Or LocalStore, doesn't matter

array = zarr.create_array(
                store,
                name="s",
                shape=data.shape,
                chunks=(10,10),
                fill_value="<NA>",
                dtype=str,
            )

# This works as expected
array[:,:] = data
zdata = array[:,:]
assert np.all(zdata == data)

# This doesn't
array[:,:] = datapa
zdata = array[:,:]
print(f"written={datapa[0,:4]}")  # ['S00001' 'S00002' 'S00003' 'S00004']
print(f"read={zdata[0,:4]}")        # ['S00001' 'S00012' 'S00023' 'S00034']
assert np.all(zdata == datapa)  # Fails
```

### Additional output

_No response_

oldrobotdev:
## Summary

Fixes #3558. Writing an F-contiguous object-dtype array to a chunked string or bytes array scrambles the stored data whenever a chunk is covered exactly by the selection: the write path passes the chunk view to numcodecs without a copy, numcodecs flattens it with order='A' (column-major for F-contiguous input), and decode reshapes in C order. The wrappers in codecs/vlen_utf8.py now pass np.ascontiguousarray(...) to the codec, a no-op for C-contiguous chunks and a copy of the object-pointer array otherwise. Regression tests cover vlen-utf8 and vlen-bytes at both the exact-cover and partial-chunk shapes; the exact-cover cases fail on current main.

## For reviewers

The one design question is the fix layer: this could live in numcodecs instead (change the flatten to C order), but that touches the v2 out= decode path where the old behavior was self-consistent for F-ordered arrays, so I kept the change on the zarr side where decode already commits to C order. The repro and codec-level probe are in #3558.

## Author attestation

## TODO

* [ ] Add docstrings and API docs for any new/modified user-facing classes and functions (no user-facing API change)
* [ ] New/modified features documented in `docs/user-guide/*.md` (not applicable)

### DIFF
--- a/changes/4116.bugfix.md
+++ b/changes/4116.bugfix.md
@@ -1,1 +1,1 @@
-Fixed writing Fortran-ordered (F-contiguous) arrays to variable-length string and bytes arrays. The vlen codec wrappers now pass a C-contiguous chunk to numcodecs, whose `order='A'` flattening previously stored the elements of an F-contiguous chunk in transposed order.
+Fixed writing Fortran-ordered (F-contiguous) arrays through the variable-length string and bytes codecs and through numcodecs array-array filters such as `Delta`, `FixedScaleOffset` and `PackBits`. Chunks are now passed to numcodecs as C-contiguous arrays, so elements are no longer stored in transposed order.
--- a/changes/4116.bugfix.md
+++ b/changes/4116.bugfix.md
@@ -1,0 +1,1 @@
+Fixed writing Fortran-ordered (F-contiguous) arrays to variable-length string and bytes arrays. The vlen codec wrappers now pass a C-contiguous chunk to numcodecs, whose `order='A'` flattening previously stored the elements of an F-contiguous chunk in transposed order.
--- a/src/zarr/codecs/numcodecs/_codecs.py
+++ b/src/zarr/codecs/numcodecs/_codecs.py
@@ -162,7 +162,9 @@ class _NumcodecsArrayArrayCodec(_NumcodecsCodec, ArrayArrayCodec):
         return chunk_spec.prototype.nd_buffer.from_ndarray_like(out.reshape(chunk_spec.shape))
 
     async def _encode_single(self, chunk_data: NDBuffer, chunk_spec: ArraySpec) -> NDBuffer:
-        chunk_ndarray = chunk_data.as_ndarray_like()
+        # numcodecs codecs flatten with order="A", so an F-contiguous chunk
+        # would be encoded in transposed element order (gh-3558)
+        chunk_ndarray = np.ascontiguousarray(chunk_data.as_ndarray_like())
         out = await asyncio.to_thread(self._codec.encode, chunk_ndarray)
         return chunk_spec.prototype.nd_buffer.from_ndarray_like(out)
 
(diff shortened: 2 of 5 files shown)


d-v-b:
thanks for this fix! Could you check if the other numcodecs codecs (delta, fixedscaleoffset, etc) are affected by the same bug?

oldrobotdev:
Checked, and yes: Delta, FixedScaleOffset and PackBits all round-trip transposed data for an F-contiguous chunk that exactly covers the selection, both when calling the numcodecs codec directly and through a zarr array with the filter configured. The path is the generic adapter in codecs/numcodecs/_codecs.py, which hands chunk_data.as_ndarray_like() to numcodecs unchanged while decode reshapes in C order, so every array-array filter goes through the same exposed path.

I pushed the same ascontiguousarray treatment for the array-array and array-bytes adapter encode paths, with regression tests for the three filters above and the changelog entry updated to match. One probe gotcha worth mentioning for review: a test pattern that is symmetric under transpose passes even on broken code (arange % 3 on a 16-wide array is symmetric, since 16 ≡ 1 mod 3), so the tests use arange and a lower-triangular bool mask.

I'll open the numcodecs issue about the underlying order='A' behavior next, since direct numcodecs users stay exposed either way.

d-v-b (approved):
pending CI, this looks great! thanks for the fix.

PR #4117, opened 2026-07-02
fix: skip object store prefix markers in list_dir

### LINKED ISSUE #4032: ObjectStore.list_dir("g") raises ValueError when an object keyed "g/" exists

AMBRA7592:
## Summary

Fixes #4032.

`ObjectStore.list_dir` can receive a directory-marker entry that matches the requested non-root prefix itself, such as path `"g"` for prefix `"g"`. That entry represents the prefix, not a child of the prefix, so `_transform_list_dir` now skips exact prefix-marker entries before relativizing returned paths.

This also avoids yielding empty relative paths and adds regression coverage for both `"g"` and `"g/"` prefix inputs.

## For reviewers

Please check whether handling the exact-prefix marker in `_transform_list_dir` is the right level. I kept the change local to `ObjectStore` because this marker entry is an object-store delimiter-listing artifact.

## Author attestation

## TODO

## Validation

* `python3 ci/check_changelog_entries.py`
* `git diff --check`
* `python3 -m py_compile src/zarr/storage/_obstore.py tests/test_store/test_object.py`

I could not run the targeted pytest locally because this Python environment is missing the project test dependency stack (`pytest`, `obstore`, `numcodecs`, `donfig`).


src/zarr/storage/_obstore.py
```diff
@@ -267,7 +267,11 @@ async def _transform_list_dir(
     for path in chain(
         list_result["common_prefixes"], map(itemgetter("path"), list_result["objects"])
     ):
-        yield _relativize_path(path=path, prefix=prefix)
+        if prefix and path in {prefix, f"{prefix}/"}:
```
d-v-b: I think obstore never returns a prefix ending in a "/", you should double-check if this is possible, and if it's not something obstore does, we can simplify this

tests/test_store/test_object.py
```diff
@@ -97,6 +98,19 @@ async def test_store_getsize_prefix(self, store: ObjectStore[LocalStore]) -> Non
         assert total_size == len(buf) * 2
 
 
+@pytest.mark.parametrize("prefix", ["g", "g/"])
+async def test_transform_list_dir_skips_prefix_marker(prefix: str) -> None:
```
d-v-b: i think the preferred test would be to add an example to the shared store tests that breaks obstore without this fix.

d-v-b:
thanks for this! 

> Please check whether handling the exact-prefix marker in _transform_list_dir is the right level.

We might need this a bit higher up. I think there's a similar bug lurking in [_chunks_initialized](https://github.com/zarr-developers/zarr-python/blob/main/src/zarr/core/array.py#L3979).

AMBRA7592:
Updated in ecef745:

- simplified the `ObjectStore` exact-prefix check to skip only `path == prefix`; the empty-relative-path guard still handles marker-shaped inputs without explicitly checking `f"{prefix}/"`
- moved the regression from the private `_transform_list_dir` helper test into the shared store test contract as `test_list_dir_ignores_prefix_marker`
- updated `MemoryStore.list_dir` to satisfy the same prefix-marker contract
- added a `_shards_initialized` guard and regression coverage for a store entry exactly matching the array path

Validation run locally:
- `git diff --check`
- `python3 -m py_compile src/zarr/storage/_obstore.py src/zarr/storage/_memory.py src/zarr/core/array.py src/zarr/testing/store.py tests/test_store/test_object.py tests/test_array.py`

I could not run the targeted pytest locally because this Python environment does not have `pytest` installed.

AMBRA7592:
Updated in a8f54f1. The shared StoreTests case was too broad: local and in-memory backends cannot represent both a raw directory-marker object (g/) and children beneath it. I replaced it with an ObjectStore S3/moto regression that recreates the reported condition by writing g/ directly through boto3, then verifies both list_dir("g") and list_dir("g/"). The _shards_initialized exact-path guard remains covered separately.

src/zarr/core/array.py
```diff
@@ -3993,7 +3993,9 @@ async def _shards_initialized(
         x async for x in array.store_path.store.list_prefix(prefix=array.store_path.path)
     ]
     store_contents_relative = [
-        _relativize_path(path=key, prefix=array.store_path.path) for key in store_contents
+        _relativize_path(path=key, prefix=array.store_path.path)
+        for key in store_contents
+        if not array.store_path.path or key != array.store_path.path
```
d-v-b: - can we avoid falsy string checks in favor of an explicit check that array.store_path.path is not the empty string
- can we get a comment that explains why we are doing this check? we should explain that obstore includes directory markers that we want to exclude

src/zarr/storage/_obstore.py
```diff
@@ -267,7 +267,11 @@ async def _transform_list_dir(
     for path in chain(
         list_result["common_prefixes"], map(itemgetter("path"), list_result["objects"])
     ):
-        yield _relativize_path(path=path, prefix=prefix)
+        if prefix and path == prefix:
```
d-v-b: replace the falsy string check with an explicit `if preifx != ''`

tests/test_store/test_object.py
```diff
@@ -97,6 +99,38 @@ async def test_store_getsize_prefix(self, store: ObjectStore[LocalStore]) -> Non
         assert total_size == len(buf) * 2
 
 
+@pytest.mark.filterwarnings(
+    re.escape("ignore:datetime.datetime.utcnow() is deprecated:DeprecationWarning")
+)
+async def test_list_dir_ignores_s3_prefix_marker(moto_server: str) -> None:
```
d-v-b, line 105: can this test get a docstring comment that explains why it exists? other developers might be surprised about how obstore handles directory-marker-like prefixes

d-v-b:
@AMBRA7592 I added a few style comments otherwise I think this is ready to go, thanks so much for this fix.

AMBRA7592:
Updated in c43c46ce:\n\n- replaced the falsy prefix checks with explicit empty-string checks\n- added a comment explaining why obstore directory markers matching the listed prefix are excluded before shard-key relativization\n- added a docstring explaining the S3 prefix-marker regression test\n\nValidation:\n- \n- \n\nThanks — ready for re-review.

d-v-b (approved):

d-v-b:
thanks @AMBRA7592!

PR #4119, opened 2026-07-05
feat(zarr-metadata): Zarr metadata model layer

d-v-b:
## Summary

Adds dataclass representations of array + group metadata, and syntactical validation routines, to `zarr-metadata`. These classes are designed to augment the basic JSON representation of zarr metadata documents with the user-friendly affordances of a class, without any pretense of runtime behavior like chunk encoding or whatnot. 

This PR also contains some pydantic integration functionality gated behind an optional dep. installing the package with pydantic exposes the integration point. the pydantic integration allows re-using the zarr metadata types as pydantic fields, with validation.

## For reviewers

These classes are a potential replacement for the metadata classes we currently define in `zarr-python`. Those classes are tightly coupled to zarr-python internals, which IMO has no served us well. For that reason, the classes in this PR are deliberately isolated from zarr-python internals. You should think about whether the classes added in this PR would be useful for zarr-python rely on (since zarr-python is the main consumer I'm writing for)

## Author attestation

## TODO


packages/zarr-metadata/src/zarr_metadata/model/_array.py
```diff
@@ -0,0 +1,466 @@
...
+    codecs: tuple[MetadataFieldModelV3, ...]
+    chunk_key_encoding: MetadataFieldModelV3
+    dimension_names: tuple[str | None, ...] | UNSET
+    attributes: dict[str, JSONValue]
+    storage_transformers: tuple[MetadataFieldModelV3, ...]
+    extra_fields: dict[str, ExtensionFieldV3]
+
+
+@dataclass(frozen=True, slots=True, kw_only=True)
+class ArrayMetadataModelV3:
```
d-v-b: flagging this name as a bikeshedding target. The natural name would be `ArrayMetadataV3`, but that's already taken in this package by the typeddict type. Since array metadata is JSON, IMO this class cannot purport to "be" array metadata, but it can claim to model array metadata. hence this name. But i'm open to alternative suggestions!

d-v-b: for example: we could decide that the dataclass model of the metadata is `ArrayMetadata`, and the typeddict model is `ArrayMetadataJSON`, or `ArrayMetadataMessage`

packages/zarr-metadata/src/zarr_metadata/model/_array.py
```diff
@@ -0,0 +1,466 @@
...
+        return must_understand_subset(self.extra_fields)
+
+    @classmethod
+    def from_key_value(cls, mapping: Mapping[str, bytes]) -> ArrayMetadataModelV3:
+        return cls.from_json(load_store_json(mapping, ARRAY_METADATA_STORE_KEY_V3))
+
+    def to_key_value(self, *, indent: int | str | None = None) -> Mapping[str, bytes]:
+        return {
+            ARRAY_METADATA_STORE_KEY_V3: json.dumps(self.to_json(), indent=indent).encode("utf-8")
+        }
```
d-v-b: flagging this as a bikeshedding target. I think it's useful for the metadata object to know the canonical name(s) of the stored objects it becomes. For zarr v3 metadata, it's `{"zarr.json": <utf8>}`, and for zarr v2 metadata it's `{".zarray": <utf8>}` | `{".zattrs": <utf8>, ".zarray": <utf8>}` | `{".zgroup": <utf8>}` | `{".zattrs": <utf8>, ".zgroup": <utf8>}` ... (ignoring zarr v2 consolidated metadata, which adds another dimension), but I am open to other ideas here re: names, signatures, etc

d-v-b: @LDeakin any insights here?

LDeakin: Seems fine, and a neat API for this metadata package.

w.r.t consolidated metadata (and future Zarr V3 stuff that could live in separate metadata files), I think it is sensible that they do not get modelled here nor automatically deserialised.

packages/zarr-metadata/src/zarr_metadata/model/_array.py
LDeakin: Bikeshedding `ArrayMetadataV3Model` as well. `ArrayMetadataModelV3` reads as the third iteration of the model.

d-v-b: agreed, I'm not a fan of that ambiguity. how about this, where we make the following decisions:

- ZarrV3 / ZarrV2 is at the front of the name
- the data class is the primary entity, and the dict form gets a suffix (JSON or Dict or Message)

```python
ZarrV3ArrayMetadata<JSON | Dict | Message> = TypedDict(...)

class ZarrV3ArrayMetadata:
    shape: tuple[int, ...]
    
    def to_json(self) -> ZarrV3ArrayMetadata<JSON | Dict | Message>: ...

```

d-v-b: @kylebarron you're consuming the typeddict types today, the rename proposed here would be breaking for you. would you like a deprecation cycle?

kylebarron: Zarrista hasn't even had a non-beta release. You can break anything

d-v-b:
I'm self-merging this! it will be live in the next version of zarr-metadata.

PR #4122, opened 2026-07-06
test: consolidate vlen F-contiguous regression tests

d-v-b:

Follow-up test-quality cleanup for the gh-3558 fix merged in #4116. No source changes — the fix itself is already on `main`; this only tidies the regression tests it added.

## What and why

The `test_vlen_string_f_contiguous` / `test_vlen_bytes_f_contiguous` tests added in #4116 had a few issues surfaced in review:

- **The `chunks=(2,2)` case never exercised the F-contiguous path.** With `(3,3)` data, a `(2,2)` chunking produces only partial chunks, which `_merge_chunk_array` recopies into a fresh C-order buffer before the codec sees them — so that parametrization gave false coverage confidence. Only the whole-array (`chunks == shape`) case actually hands an F-contiguous buffer to the codec. This PR keeps just that case.
- **The F-order input was built indirectly** (`np.asarray(np.array(...).reshape(3, 3), order="F")`), so a future "simplification" dropping the outer `np.asarray` could silently stop testing the F path. Now it's built directly with `reshape((3, 3), order="F")` and guarded with `assert data.flags.f_contiguous`.
- **The two tests were near copy-paste**, differing only in dtype/fill_value/data — now one test parametrized over `(dtype, fill_value, elements)`, matching the parametrized style of `test_vlen_string` in the same file.
- **Added a docstring** stating the verified behavior.

## Verification

- `pytest tests/test_codecs/test_vlen.py -k f_contiguous` → 2 passed (string + bytes).
- `prek run mypy --all-files`, `ruff check`, `ruff format --check` all pass.

### DIFF
--- a/tests/test_codecs/test_vlen.py
+++ b/tests/test_codecs/test_vlen.py
@@ -9,7 +9,7 @@ from zarr.abc.codec import Codec, SupportsSyncCodec
 from zarr.abc.store import Store
 from zarr.codecs import ZstdCodec
 from zarr.codecs.vlen_utf8 import VLenBytesCodec, VLenUTF8Codec
-from zarr.core.dtype import VariableLengthBytes, get_data_type_from_native_dtype
+from zarr.core.dtype import get_data_type_from_native_dtype
 from zarr.core.metadata.v3 import ArrayV3Metadata
 from zarr.storage import StorePath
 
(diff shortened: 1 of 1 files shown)


d-v-b:
I'm going to self-merge because this is a small internal cleanup.

PR #4133, opened 2026-07-08
docs: fix CacheStore max_age_seconds docstring default

d-v-b:
## Description

The `CacheStore` docstring documented `max_age_seconds` as `int | None` with a default of `None`, but the actual signature is `max_age_seconds: int | str = "infinity"`, and `None` is rejected by validation. This updates the docstring to match the real signature and default.

While there, also fixed the `cache_store` parameter description: it said "can be any Store implementation", but `__init__` raises `ValueError` for stores that don't support deletes, so that requirement is now documented.

## Author attestation

- [x] I confirm that I understand the full content of this pull request, and that its content complies with the contributing guide.


d-v-b:
self-merging when tests finish as this is a tiny cleanup

PR #4136, opened 2026-07-09
ci: stop dependabot updates targeting support/v2

d-v-b:

Removes the `target-branch: support/v2` entry from `.github/dependabot.yml`, so dependabot stops opening weekly github-actions bump PRs against `support/v2` (most recently #4134).

Those PRs can never go green: `support/v2` CI fails for reasons unrelated to the actions bumps — the identical job set failed on the previous bump PR #4098 two weeks earlier. Specifically:

- `test_encode_decode_array_datetime_timedelta` fails under numpy 2.5.1, which escalates the new "'generic' unit for NumPy timedelta is deprecated" DeprecationWarning to an error (fails `minimum_build`, `latest-deps`, and all Linux build jobs; the `numpy==1.24.*` job is only canceled by fail-fast).
- Windows jobs fail all `TestArrayWithABSStore` tests with `ssl.SSLError: [ASN1: NOT_ENOUGH_DATA]` against the Azure Blob emulator.
- The Read the Docs build fails independently.

Since `support/v2` is in maintenance and nobody is going to fix its CI for the sake of actions bumps, the dependabot PRs are pure noise. Dependabot reads its config from the default branch only, so this change on `main` is sufficient; #4134 should be closed manually (it won't be recreated once this lands).


d-v-b:
internal cleanup, going to self-merge

PR #4138, opened 2026-07-09
test: pass a list, not a generator, to parametrize in test_docs

d-v-b:

## What

`tests/test_docs.py::test_docstrings` passes the `find_examples(...)` **generator** directly to `@pytest.mark.parametrize`. Materialize it to a `list`, matching the existing `list(find_examples(...))` usage a few lines up in the same module.

```diff
-@pytest.mark.parametrize("example", find_examples(str(SOURCES_ROOT)), ids=str)
+@pytest.mark.parametrize("example", list(find_examples(str(SOURCES_ROOT))), ids=str)
```

## Why

pytest 9.1 deprecated passing a non-Collection iterable (e.g. a generator) to `parametrize`, emitting `PytestRemovedIn10Warning`. Because this repo sets `filterwarnings = ["error"]`, that deprecation is promoted to a hard error at **collection time**:

```
ERROR collecting tests/test_docs.py
E   pytest.PytestRemovedIn10Warning: Passing a non-Collection iterable to parametrize is deprecated.
E   Test: tests/test_docs.py::test_docstrings, argvalues type: generator
Interrupted: 1 error during collection
```

This is the sole blocker for the pytest `9.0.3 → 9.1.1` bump in the python-dependencies group update (#4137): in that run every job passes except `doctests`, which is the only env that installs `pytest-examples` and therefore actually collects this module (every other env skips it via `importorskip`).

## Verification

Under `pytest==9.1.1` with `pytest-examples` installed:
- **before:** `no tests collected, 1 error in 0.11s` (reproduces the CI failure)
- **after:** `60 tests collected`, no error


d-v-b:
merging this as soon as tests pass

PR #4139, opened 2026-07-10
test: bind moto server to an ephemeral port to fix xdist deadlock

d-v-b:
This is a claude-authored PR that changes our S3 test fixture to use an ephemeral port managed at runtime by moto itself. I discovered this while commissioning claude to find new pytest features that we can use to improve our test suite.

## Problem

Running the full suite locally with `pytest -n auto` deadlocks near the end (~98%), with every xdist worker parked at 0% CPU. It has been this way since the shared moto fixture landed (#4018, 2026-06-02).

**Root cause:** the session-scoped `moto_server` fixture in `tests/conftest.py` bound a fixed port (`MOTO_SERVER_PORT = 5555`). Under pytest-xdist, *session scope means once per worker*, so when two or more workers reach an S3 test concurrently they race to bind the same port. The first wins; every loser blocks **forever** in `ThreadedMotoServer.start()` — its werkzeug thread dies on "Address already in use" before ever setting the ready `Event` that `start()` waits on (no timeout).

Minimal repro: construct two `ThreadedMotoServer(port=5555)` and call `start()` on both — the second never returns.

CI never sees this because `run-coverage` runs the suite single-process, and the one `-nauto` CI target (`run-hypothesis`) doesn't touch S3.

## Fix

- `moto_server` binds `port=0` (ephemeral) and yields the actual endpoint from `server.get_host_and_port()`. This also makes concurrent pytest sessions on one machine safe, not just concurrent workers.
- Since the endpoint is now only known at runtime, `test_fsspec.py`'s module-level `endpoint_url` constant becomes an `endpoint_url` fixture (absorbing the old `s3_base` alias), and its consumers take it as a parameter. `tests/test_docs.py` already consumed the fixture value, so it needs no changes.

## Verification

- `pytest tests/test_store/test_fsspec.py -n 4`: 102 passed in 8.8s (previously a guaranteed deadlock shape — 4 workers, 4 servers).
- Full suite `pytest --ignore tests/benchmarks -n auto`: **6196 passed, 277 skipped, 4 xfailed in 55.54s** on a 10-core M-series machine, where it previously hung indefinitely.
- Diagnosis receipts: sampling the hung workers showed all threads in `__psynch_cvwait`; `kill -ABRT` produced faulthandler dumps pointing at `Event.wait` ← `ThreadedMotoServer.start` ← `moto_server` fixture setup.


d-v-b:
this is a small test-time-scoped quality-of-life improvement to our test suite that does not affect public code. I will self-merge.

PR #4146, opened 2026-07-15
fix: let open_like create arrays by default

### LINKED ISSUE #3352: `zarr.api.asynchronous.open_like` is broken

goutamadwant:
## Summary

Closes #3352.

`zarr.api.asynchronous.open_like` is intended to create or open a persistent array using metadata copied from an existing array. When callers omitted `mode`, or passed `mode=None`, it delegated to `open_array` without a create-capable mode and raised `ArrayNotFoundError` for a missing target path.

This changes `open_like` so an omitted or `None` mode defaults to `"a"` before delegating to `open_array`. Explicit modes are still passed through unchanged.

The regression test covers both Zarr v2 and v3, with `mode` omitted and with `mode=None`.

## For reviewers

The main behavior choice is defaulting only `open_like` to append/create mode. I left `open_array` unchanged so this stays focused on the helper whose purpose is to create a similar persistent array.

Verified with:

- `uv run pytest -q tests/test_api.py::test_open_like_creates_array_by_default tests/test_api.py::test_array_like_creation --tb=short`
- `uv run pytest -q tests/test_api.py --tb=short`
- `uvx prek run --files src/zarr/api/asynchronous.py tests/test_api.py changes/3352.bugfix.md`

## Author attestation

## TODO

- [x] Add unit tests and/or doctests in docstrings
- [x] Add docstrings and API docs for any new/modified user-facing classes and functions - N/A, no new public API
- [x] New/modified features documented in `docs/user-guide/*.md` - N/A, bug fix only
- [x] Changes documented as a new file in `changes/`
- [ ] GitHub Actions have all passed
- [ ] Test coverage is 100% (Codecov passes)


d-v-b:
looks good, thanks for fixing this. 
### requested changes

- can you add a test that probes the intersection between a read-only store and a mutable access mode (like "a")? We expect this to error.
- can you add a note to the changelog that users with a read-only store will need to explicitly declare mode="r" when using these routines
- can you add docstrings to the `open_like` routines that explain what the default `mode` is.

goutamadwant:
> looks good, thanks for fixing this.
> 
> ### requested changes
> * can you add a test that probes the intersection between a read-only store and a mutable access mode (like "a")? We expect this to error.
> * can you add a note to the changelog that users with a read-only store will need to explicitly declare mode="r" when using these routines
> * can you add docstrings to the `open_like` routines that explain what the default `mode` is.

@d-v-b addressed these comments. Let me know if this looks good or any suggestions. Thanks!

d-v-b (approved):
thanks!

PR #4149, opened 2026-07-15
docs: add roadmap page outlining future plans

d-v-b:
Adds a Roadmap page to the documentation describing the goals for the
next major cycle of work ("v4"), the intended changes per theme, and
the three-stream release model (additive minors, deprecations, one
minimal removals-only major), with an option for a `legacy` module.

## Author attestation


d-v-b:
cc @ilan-gold @maxrjones, opening this as a draft for visibility. lmk if there's anything more or less we should say here

ilan-gold:
Screenshot 2026-07-16 at 17 06 53

This doesn't lead me where I thought it would, somehow linking to an old page

docs/roadmap.md
```diff
@@ -0,0 +1,303 @@
...
+7. **Metadata** — pure data documents describing arrays and groups.
+
+Today, Zarr-Python is a monolith that serves every level: a consumer who only
+needs metadata handling has to install the full dependency footprint of the
+whole library, and a faster chunk-decoding implementation cannot plug in
+without re-implementing the layers above it. The v4 direction is to re-shape
+Zarr-Python around the stack, so that each level is something you can depend
+on, conform to, or replace, without buying every other level:
+
+- **A focused package per level** — `zarr-metadata`, `zarr-store`,
```
ilan-gold: Would maybe note the `zarrs` example here

docs/roadmap.md
```diff
@@ -0,0 +1,303 @@
...
+  encodings, etc.
+
+An important design input: [zarrs](https://github.com/zarrs/zarrs) (Rust) and
+[TensorStore](https://github.com/google/tensorstore) (C++) are two independent
+Zarr implementations that have converged on the same architectural patterns —
+sync-first codec APIs, per-codec concurrency budgets, adaptive sharded-read
+strategies, request deduplication, conditional reads. We treat them as
+complementary rather than competitive: Zarr-Python aims to be the best
+pure-Python Zarr implementation *and* the best wrapper around the
+compiled-language implementations, so that users who need native throughput can
```
ilan-gold: I'm really conflicted on this, because I think baking fragmentation into the ecosystem is not great, but the flipside is that languages may have different features/advantages. It would be great to have a concrete reason (or list of reasons) as to why we would design around this level of extensibility

docs/roadmap.md
```diff
@@ -0,0 +1,303 @@
...
+
+Name and specify the layer that sits between the store API (key-agnostic
+bytes) and the user-facing `Array` / `Group` facade, as a small set of typed
+verbs (`read_array_metadata`, `write_chunk`, `list_children`,
+`read_selection`, ...). Alternative engines implement the verbs end-to-end;
+hierarchy-aware caching wraps them; chunk-introspection APIs expose them.
+
+### Codecs
+
+The current codec API wraps every codec in an unnecessary async layer (a
```
ilan-gold: Not anymore!

docs/roadmap.md
d-v-b: good points, I'd say we want the backend to be an extension point for a few reasons:
- we always want a traditional python impl. IMO it's our job as the python implementation, and it keeps the ecosystem healthy to have multiple implementations
- we can't guarantee uniform, complete feature support across all implementations. As long as we allow people to hack on stuff in python, there will be things the python impl does that e.g. a rust impl doesn't support, and that's OK. For now, a compiled backend is effectively an external dep, and we simply can't promise what external deps will do in perpetuity. 
- IMO a full backend is just an refined version of our current model, which requires extenders manage buffer prototypes, data type implementations, codec implementations, and store implementations. Consolidating all of this pluggability on a single surface should make extending Zarr Python easier even for other python backends (like a cupy backend, or a pysparse backend)

docs/roadmap.md
```diff
@@ -0,0 +1,303 @@
...
+constraints, and it inherited many structural patterns from the 2.x
+implementation it replaced. The library has never had a release cycle whose
+primary goal was the *shape* of the internals. The next body of work — which we
+call **"v4"** — is that overdue investment.
+
+## Goals
+
+If the 3.0 goals could be sloganized as "migrate to Zarr V3, and improve cloud
+storage support", the slogan for the v4 goals is:
+**"support a Zarr-based Python ecosystem for chunked arrays"**. Zarr-Python
```
maxrjones:
```suggestion
**"a frictionless Zarr-based Python ecosystem for chunked arrays"**. Zarr-Python
```
I'd argue that there is a zarr-based Python ecosystem for chunked arrays, it just has a fair amount of friction to it. I'd like some adjective that's more descriptive than "support" for the slogan.

docs/roadmap.md
```diff
@@ -0,0 +1,303 @@
...
+should be *foundational* for the growing number of Python packages that work
+with data in the Zarr format. Concretely, that means pushing in these
+directions:
+
+- Give Zarr-Python users excellent performance, out of the box.
+- Make Zarr-Python APIs ergonomic and useful for developers.
+- Expand our scope to cover vital quality-of-life routines like data copying,
+  rechunking, and the like.
+- Support the growth of Python tools across all levels of the Zarr stack.
```
maxrjones:
```suggestion
- Ease the growth of Python tools across all levels of the Zarr stack.
```
it's technically feasible today; we want to make it easier

docs/roadmap.md
```diff
@@ -0,0 +1,303 @@
...
+
+## The Zarr stack
+
+Different applications need different levels of Zarr support: a convention
+validator only needs to read metadata documents; a visualization tool may only
+need read-only array access; other tools need everything. We think of this as a
+"Zarr stack", from most abstract to most concrete:
+
+1. **Conventions** — domain-specific schemas built on top of Zarr (OME-NGFF,
+   GeoZarr, anndata-zarr).
```
maxrjones:
```suggestion
1. **Conventions** — application and/or domain-specific schemas built on top of Zarr (OME-NGFF,
   GeoZarr, anndata-zarr, multiscales).
```
I think it's important to include that conventions need not be oriented around domains. they were designed to support application specific use-cases as well (xref https://github.com/zarr-developers/blog/pull/68/changes; which needs updating and publication and is on me to do)

docs/roadmap.md
```diff
@@ -0,0 +1,303 @@
...
+  The first of these,
+  [`zarr-metadata`](https://pypi.org/project/zarr-metadata/), is already
+  published.
+- **A documented interface per level** — capability protocols for stores, a
+  small stateless codec API, pure-data dtypes.
+- **A conformance suite per level** — so that alternative implementations of a
+  level can verify they behave correctly.
+- **Engine pluggability at the chunk-decoding level** — alternative engines
+  (zarrs, TensorStore) can take over IO without re-implementing hierarchy
+  traversal, indexing, or metadata handling.
```
maxrjones: this seems a bit misaligned with other statements that imply pluggabillity at each level

docs/roadmap.md
```diff
@@ -0,0 +1,303 @@
...
+
+## What we intend to change
+
+Each theme below is backed by a detailed technical proposal; the summaries
+here describe the intended end state.
```
maxrjones: as we discussed at the dev meeting, if the idea is to reference zarr-python-planning for detailed technical proposals, we should move it to the zarr-developers github organization

docs/roadmap.md
```diff
@@ -0,0 +1,303 @@
...
+
+### Device-agnostic IO
+
+Make Zarr-Python's IO surfaces device-agnostic rather than adding GPU support
+as a bolted-on feature: stores and codecs grow APIs for writing into a
+caller-provided buffer (`read_into`, `decode_into`), and the `Array` facade
+returns array-like objects in the user's chosen Array API namespace. GPU
+support falls out once the assumption of CPU destinations is removed, and CPU
+paths get faster too, because pre-allocated output buffers eliminate per-chunk
+allocation.
```
maxrjones, line 152: @TomAugspurger it'd be great to get your thoughts on this component in particular.

docs/roadmap.md
```diff
@@ -0,0 +1,303 @@
...
+`Array.__getitem__` performs IO eagerly and returns NumPy arrays, which makes Zarr
+arrays the odd one out among modern array libraries and blocks compliance with
+the [Python Array API](https://data-apis.org/array-api/) standard. Add an
+opt-in `array.lazy[...]` accessor backed by a stable coordinate-mapping algebra
+(the `IndexTransform` work in
+[#3906](https://github.com/zarr-developers/zarr-python/pull/3906)), plus a
+small query planner that turns chained selections into a single IO plan before
+any chunks are fetched. No new array type is introduced. Whether the *default*
+of bare `array[...]` ever flips from eager to lazy is an explicit, separate
+decision — see [decision points](#decision-points) below.
```
maxrjones:
```suggestion
The Zarr-Python Array API was initially designed to mirror NumPy, with eager
syntax. `Array.__getitem__` performs IO eagerly and returns a NumPy arrays.
That was helpful to the dominant use-case at the time of its creation, but it
means deferred I/O and computation currently require an external library
such as Dask. It means there is no build-in support for representing multi
step reads as a single deferred plan. Further, it means that every chained
selection round-trips to storage independently.

To solve this limitation, Add an opt-in `array.lazy[...]` accessor backed by a
stable coordinate-mapping algebra (the `IndexTransform` work in
[#3906](https://github.com/zarr-developers/zarr-python/pull/3906)), plus a
small query planner that turns chained selections into a single IO plan before
any chunks are fetched. No new array type is introduced. Whether the *default*
of bare `array[...]` ever flips from eager to lazy is an explicit, separate
decision — see [decision points](#decision-points) below.
```
I think the original description was a bit overstated with respect to the eagerness landscape and array API relationship. I suggest these revisions to clarify the motivations a bit.

docs/roadmap.md
```diff
@@ -0,0 +1,303 @@
...
+### Decision points
+
+Flipping the default of bare `array[...]` from eager to lazy is the single
+highest-migration-cost item in the plan, so it is handled as an explicit
+decision, not bundled into the additive work. The opt-in `array.lazy[...]`
+accessor ships first, with no default change. Whether the default ever flips
+hinges on whether Array API conformance at the bare-`__getitem__` surface turns
+out to be a hard requirement; if it does, the flip happens as a long-window
+deprecation with an explicit eager escape hatch and downstream coordination —
+never as a reason to adopt a major version.
```
maxrjones: To some extent, I think it would be better to separate this from the roadmap. Include the commitment to delivering lazy indexing, leave the how to the detailed documents and reference zarr-python-planning.

docs/roadmap.md
maxrjones: It also may be worth separating the "how" out to the detailed document, only setting the "what" here.

maxrjones (commented):
This is awesome! thanks so much for your work on this @d-v-b!  🚀

maxrjones:
> This doesn't lead me where I thought it would, somehow linking to an old page

fyi @d-v-b this can be fixed by removing this redirect: https://github.com/zarr-developers/zarr-python/blob/50b7e016590d76011382771fd52843949782c9c0/mkdocs.yml#L204

docs/roadmap.md
```diff
@@ -0,0 +1,308 @@
...
+for the Zarr V2 and V3 storage formats, storage APIs ergonomic for high-latency
+(cloud) storage, and backwards compatibility with Zarr-Python 2.x where
+possible. Those goals were largely achieved, and more than a year on, the
+2.x → 3.x transition is effectively resolved.
+
+The 3.x redesign was carried out under hard backwards-compatibility
+constraints, and it inherited many structural patterns from the 2.x
+implementation it replaced. The library has never had a release cycle whose
+primary goal was the *shape* of the internals. The next body of work — which we
+call **"v4"** — is that overdue investment.
```
maxrjones:
```suggestion
The 3.x redesign was carried out under hard backwards-compatibility
constraints, and it inherited many structural patterns from the 2.x
implementation it replaced. These structural patterns have made it
challenging to evolve the API for usability (e.g., many argument are
counter-intuitive and lead to user error) and provide excellent performance
across backends (e.g., local and hybrid systems suffer from 
Python overhead with our current design). The goal of **v4** is
to provide a highly usable library that delivers first-class performance
across all backends. We recognize that accomplishing these goals is
only possible via improvements to Zarr Python's internals, and consider
the investment in the Zarr Python internals to be a crucial part of
the **v4*** effort.
```

I'd like this section to be explicitly motivated for reasons beyond just cleaning up the shape of the internals.

docs/roadmap.md
d-v-b: good point, will add something crediting zarrs with this layout

docs/roadmap.md
d-v-b: I made that move: https://github.com/zarr-developers/zarr-python-planning

docs/roadmap.md
d-v-b: agreed, I thinned this section down heavily.

docs/roadmap.md
```diff
@@ -0,0 +1,183 @@
...
+### Coordinated and distributed writes
+
+Give the two patterns that actually produce large Zarr archives — parallel
+disjoint-region writes and append-along-axis growth — a design home: disjoint
+chunk-aligned region writes with alignment *checked* rather than assumed, a
+create-then-hand-out-regions primitive, and single-writer resize/append, all on
+plain Zarr V3. Stronger guarantees (atomicity, reader isolation, concurrent
+appenders) are enabled through the seam a transactional engine such as
+[Icechunk](https://icechunk.io/) builds on, rather than implemented in
+Zarr-Python itself.
```
TomNicholas: I'm curious about this, but I don't actually know what this LLM paragraph is saying.

d-v-b: basically we can introduce some simple synchronization / coordination routines and make writing to the same chunks safer :) I will de-llmify it, or just remove the test altogether

docs/roadmap.md
```diff
@@ -0,0 +1,183 @@
...
+    This roadmap reflects the current thinking of the core developers. It is a
+    statement of direction, not a schedule. We don't know how long these changes
+    will take, only that we are committed to moving the project in the direction outlined
+    here.
+
+## Where we are
+
+The [3.0 release](https://github.com/zarr-developers/zarr-python/releases/tag/v3.0.0)
+was a total redesign of the library's internals, with three goals: full support
+for the Zarr V2 and V3 storage formats, storage APIs ergonomic for high-latency
```
kylebarron:
```suggestion
for the Zarr V2 and V3 storage formats, storage APIs that are ergonomic for high-latency
```

docs/roadmap.md
```diff
@@ -0,0 +1,183 @@
...
+(cloud) storage, and backwards compatibility with Zarr-Python 2.x where
```
kylebarron:
```suggestion
storage (such as cloud storage), and backwards compatibility with Zarr-Python 2.x where
```

docs/roadmap.md
```diff
@@ -0,0 +1,183 @@
...
+possible. Those goals were largely achieved, and more than a year on, the
+2.x → 3.x transition is effectively resolved, in the sense that it the transition
+is no longer the prevailing source of issues and pull requests.
```
kylebarron:
```suggestion
possible.

Those goals were largely achieved! Now, more than a year on, the
2.x → 3.x transition is effectively resolved, in the sense that that transition
is no longer the prevailing source of issues and pull requests.
```

docs/roadmap.md
```diff
@@ -0,0 +1,183 @@
...
+storage support", the slogan for the v4 goals is:
+**"a frictionless Zarr-based Python ecosystem for chunked arrays"**. Zarr-Python
+should be *foundational* for the growing number of Python packages that work
+with data in the Zarr format. Concretely, that means pushing in these
+directions:
+
+!!! note
+  Many of the features described below will not require breaking public 3.x APIs. We can and will
+  ship those features in 3.x releases; at the same time, we consider it clarifying to frame the
+  coherent development direction as vectored at a 4.0 milestone.
```
kylebarron: I'd put this after the bulleted list.

docs/roadmap.md
```diff
@@ -0,0 +1,183 @@
...
+
+- Deliver excellent performance, out of the box, by whatever means necessary (e.g., Rust bindings).
```
kylebarron: I think "by whatever means necessary" is maybe _too_ broad. That could mean 100% vibe coded SIMD assembly, for all the reader knows.

I think we should be more explicit about this being _excellent performance while preserving maintainability_

docs/roadmap.md
```diff
@@ -0,0 +1,183 @@
...
+- Expand our scope to cover vital quality-of-life routines like data copying,
+  rechunking, and the like.
+- Ease the growth of Python tools across all levels of the Zarr stack.
+- Accelerate the implementation of new codecs, chunk grids, chunk key
+  encodings, etc.
+
+An important design input: [zarrs](https://github.com/zarrs/zarrs) (Rust) and
+[TensorStore](https://github.com/google/tensorstore) (C++) are two independent
+Zarr implementations that use architectural patterns we want to learn from.
+We see them a complementary rather than competitive.
```
kylebarron:
```suggestion
We see them as complementary rather than competitive.
```

docs/roadmap.md
```diff
@@ -0,0 +1,183 @@
...
+
+## The Zarr stack
+
+Different applications need different levels of Zarr support: a convention
+validator only needs to read metadata documents; a visualization tool may only
+need read-only array access; other tools need everything. We think of this as a
+"Zarr stack", from most abstract to most concrete:
+
+1. **Conventions** — application and/or domain-specific schemas built on top of Zarr (OME-NGFF,
```
kylebarron, line 74:
```suggestion
1. **Specification Conventions** — application and/or domain-specific schemas built on top of Zarr (OME-NGFF,
```

docs/roadmap.md
```diff
@@ -0,0 +1,183 @@
...
+   GeoZarr, anndata-zarr, multiscales).
+2. **Groups** — Zarr hierarchies, traversal, group-level attributes.
+3. **Arrays** — the user-facing array object, plus indexing and slicing.
+4. **Chunk decoding** — the codec pipeline.
```
kylebarron, line 78:
```suggestion
4. **Chunk encoding/decoding** — the codec pipeline. How typed array data gets translated into bytes that are persisted to a store.
```

docs/roadmap.md
```diff
@@ -0,0 +1,183 @@
...
+5. **Chunk addressing** — chunk grids and key encodings that map array
+   coordinates to store keys.
+6. **Stores** — the key-value layer.
+7. **Metadata** — pure data documents describing arrays and groups.
+
+Today, Zarr-Python is a monolith that serves every level: a consumer who only
+needs metadata handling has to install the full dependency footprint of the
+whole library, and a faster chunk-decoding implementation cannot plug in
+without re-implementing the layers above it. The v4 direction is to re-shape
```
kylebarron, line 87:
```suggestion
without reimplementing the layers above it. The v4 direction is to reshape
```
🤷‍♂️ not sure when to use either version

docs/roadmap.md
```diff
@@ -0,0 +1,183 @@
...
+Zarr-Python around the stack, so that each level is something you can depend
+on, conform to, or replace, without buying every other level:
+
+- **A focused package per level** — `zarr-metadata`, `zarr-storage`,
+  `zarr-codec`, `zarr-dtype`, all composed in the `zarr` package. The Rust `zarrs` library
```
kylebarron:
```suggestion
- **A focused package per level** — `zarr-metadata`, `zarr-storage`,
  `zarr-codec`, `zarr-dtype`, all relied upon by the primary `zarr` entry point. The Rust `zarrs` library
```

docs/roadmap.md
```diff
@@ -0,0 +1,183 @@
...
+  traversal, indexing, or metadata handling.
+
+## What we intend to change
+
+Each theme below is backed by a detailed technical proposal; the summaries
+here describe the intended end state.
+
+### Foundation: swappable backends
+
+Refactor the internals around a *swappable engine* — a single protocol that defines the core routines a
```
kylebarron:
```suggestion
We want to refactor the internals around a *swappable engine* — a single protocol that defines the core routines a
```

docs/roadmap.md
```diff
@@ -0,0 +1,183 @@
...
+Zarr implementation must support. Zarr-Python becomes one user-facing API that can be driven by multiple
+backends, including a Python-heavy backend, but also a Rust-based backend, via bindings to the `zarrs` crate.
```
kylebarron: Do you want to mention zarrista here? It's fine if that's not mentioned in the official roadmap

docs/roadmap.md
```diff
@@ -0,0 +1,183 @@
...
+
+Internally we will branch over two kinds of backends: synchronous and asynchronous. The synchronous backend is suitable for arrays and groups persisted to low-latency storage like in-memory stores, where async scheduling is pure friction. The asynchronous backend will use `async` and provide concurrent APIs where it helps: for arrays and groups persisted to high-latency storage.
```
kylebarron:
```suggestion
Internally we will branch over two kinds of backends: synchronous and asynchronous. The synchronous backend is suitable for arrays and groups persisted to low-latency storage like in-memory stores or local file systems, where async scheduling is pure friction. The asynchronous backend will use Python's `async` support and will provide concurrent APIs where it helps: for arrays and groups persisted to high-latency storage.
```

docs/roadmap.md
```diff
@@ -0,0 +1,183 @@
...
+step reads as a single deferred plan. Further, it means that every chained
+selection round-trips to storage independently.
+
+We can fix this by introducing an API for lazy indexing. Under this model, an indexing operation
+like `array[::2]`desugars to a declarative state like `(array, selection)`. Chained selections like
+ `array[10:100][::2]` are fused immediately, and we defer actual IO for the time when the result of
+ indexing is needed. TensorStore is an excellent role model for Zarr-Python here, and we can deliver
+ this functionality without breaking ordinary indexing behavior. See this
+ [classic discussion](https://github.com/zarr-developers/zarr-python/discussions/1603) for more
+ background.
```
kylebarron:
```suggestion
`array[10:100][::2]` are fused immediately, and we defer actual IO for the time when the result of
indexing is needed. TensorStore is an excellent role model for Zarr-Python here, and we can deliver
this functionality without breaking ordinary indexing behavior. See this
[classic discussion](https://github.com/zarr-developers/zarr-python/discussions/1603) for more
background.
```

nit

docs/roadmap.md
```diff
@@ -0,0 +1,183 @@
...
+
+### Data types
+
+First-class support for ML-specific dtypes — `bfloat16`, the `float8`
+variants, packed `int4`/`uint4` — via
+[`ml_dtypes`](https://github.com/jax-ml/ml_dtypes), using the exact identifiers
+registered in `zarr-extensions` so the data stays readable by other
+implementations. Ragged arrays, variable-length strings, and an investigation
```
kylebarron:
```suggestion
implementations. 

Ragged arrays, variable-length strings, and an investigation
```

docs/roadmap.md
kylebarron: Actually all the below sections don't have a subject so this is fine

docs/roadmap.md
```diff
@@ -0,0 +1,183 @@
...
+paths get faster too, because pre-allocated output buffers eliminate per-chunk
+allocation.
+
+### Configuration, registries, and plugins
+
+Move configuration from "global mutable state read implicitly" to "typed data
+passed explicitly": a typed config object replacing the untyped global `donfig`
+dict, array-scoped runtime config passed at open time, a registry redesign that
+addresses implementations by stable identity and resolves plugin name-conflicts
+deliberately, and named profiles replacing global mutators.
```
kylebarron, line 160: FWIW I would like to prototype using a context manager in Zarrista for scope-local configuration. 

See https://github.com/developmentseed/zarrista/issues/153. In particular the idea is to borrow from [`rasterio.Env`](https://rasterio.readthedocs.io/en/stable/api/rasterio.env.html#rasterio.env.Env), which injects settings only for that local scope. 

```py
with rasterio.Env(GDAL_CACHEMAX=128000000) as env:
    # All drivers are registered, GDAL's raster block cache
    # size is set to 128 MB.
    # Commence processing...
    ...
    # End of processing.

# At this point, configuration options are set to their
# previous (possible unset) values.
```

kylebarron (approved):

docs/roadmap.md
d-v-b: yeah, we will mention zarrista. It's an indicator that `zarrs` bindings are available

docs/roadmap.md
d-v-b: This kind of thing relies on two routines or objects being the same, modulo the configuration. So why not make this explicit? for functions its easy, we declare their configuration in a parameter. for objects we can do something similar, by initializing the object with a configuration, and copying the object with a different configuration as needed, e.g. `new_object = StatefulObject.with_config(new_config)`. IMO this is cleaner because it means runtime internals don't need to know about environment variables. they just inspect local data -- function parameters or the attributes of `self` -- for configuration info.

kylebarron: > This kind of thing relies on two routines or objects being the same, modulo the configuration. So why not make this explicit? 

We do currently make this explicit as a parameter into each function. The context manager would be an abstraction to automatically pass down the config into all zarrista calls within that scope.

I'm still not sure whether I like the idea though.

kylebarron: > don't need to know about environment variables.

Oh but to be clear, I didn't have in mind environment variables. Despite the name of `rasterio.Env`, I don't think it's setting global environment variables, because that wouldn't be scope-local.

I had in mind a context manager that would hold config as a Python object, and that local config could be queried. But I don't think Zarrista should touch shell environment variables at all

d-v-b: >  that local config could be queried.

this is a pattern I don't really like. if an object like a function or class instance depends on some other data, then that should be formalized as part of the function signature or class structure. zarr-python today has a lot of places where code reaches for a free-floating config object, when IMO the simpler solution is to add parameters to functions / classes.

d-v-b:
I gave the text a sweep, folks should have another look.

kylebarron (approved):

d-v-b:
@zarr-developers/python-core-devs please have a look!

d-v-b:
I'd like to get this merged soon, so I will _self-merge today_ unless anyone raises objections to specific content here, in which case we can refine things and then merge soon. Otherwise I recommend we continue iterating on the roadmap in follow up PRs.

docs/roadmap.md
maxrjones: I prefer keeping it as just "conventions" since specification may confuse folks who think of the core zarr spec

maxrjones (commented):
I agree with everything in this roadmap. I also support merging this since there's been a long window for comments, and folks can still open PRs against the roadmap.

docs/roadmap.md
maxrjones: this suggestion was based off an earlier version of the PR; I didn't realize it'd still be included in my review. It probably isn't still necessary, but I'll leave it.

PR #4157, opened 2026-07-17
Fix/memorystore buffer aliasing

d-v-b:
@maxrjones ran a regression check on our pre-3.3.0 release state, and found a few regressions that need fixing. Here's the first fix, which ensures that memory storage always copies buffers, instead of hanging on to aliases of arrays owned by the caller.

this PR was written by claude code and opened by me; to view claude's original output, see https://github.com/d-v-b/zarr-python/pull/224


maxrjones (approved):
this looks right to me, thanks @d-v-b!

FWIW I also really like the way you've structured your PR with the link to the raw work here.

PR #4165, opened 2026-07-20
fix(store): don't close a shared filesystem in FsspecStore.close()

d-v-b:
#4003 added logic to zarr-python for closing a fsspec-managed filesystem when closing an instance of `FsspecStore`. This is unsafe because fsspec filesystems are _shared_, and so there's no way for a single `FsspecStore` instance to have the information required to only close filesystems in use by that particular store instance. That means a copy of a store instance could close, bring down the underlying fsspec filesystem, and break functionality for all other stores using that filesystem with no recourse. undesirable.

This PR ensures that zarr-python does not close fsspec filesystems. That's left for routines with a broader context.

The changes in this PR were written by Claude, based on a report max commissioned that revealed released blockers.
The source PR from claude is here: https://github.com/d-v-b/zarr-python/pull/226

## Author attestation

## TODO


changes/226.bugfix.md
```diff
@@ -0,0 +1,11 @@
...
+network session. fsspec caches and shares filesystem instances across callers,
+so a store cannot know whether it is the only user, and closing a shared
+session broke sibling stores built from the same host — and left the closed
+filesystem in fsspec's cache for later callers. The filesystem's lifecycle now
+stays with whoever created it; use fsspec's own tools (for example
+`AbstractFileSystem.clear_instance_cache()`) to release it.
+
+This removes the store-owns-the-filesystem behavior added in gh-4003. Code that
+relied on `store.close()` releasing the filesystem's session must now close the
+filesystem itself.
```
maxrjones: did this bug precede the last release? I was under the impression that this was a regression since the last release, in which case a changelog entry isn't necessary.

maxrjones (approved):
Thanks, Davis!

changes/226.bugfix.md
maxrjones: if this changelog entry does stay in, the filename needs fixing to 4165 rather than 226

d-v-b: good point, this is reverting unreleased changes so we can omit the changelog entirely

d-v-b: fixed in 6160c9eb1976a83cd1f678adb3a8e82cf7886d1d

PR #4172, opened 2026-07-21
perf(indexing): speed up sorted 1-D coordinate selections

### LINKED ISSUE #4170: arr_1d[np.sort(idx)] can be much faster (CoordinateIndexer fast path)
### Describe the new feature you'd like

Hi,

Fancy-indexing a 1-D Zarr array with a sorted integer array (`arr[sorted_idx]`) is dominated by `CoordinateIndexer.__init__`. It runs several checks over the flat selection. For a sorted selection this is avoidable: the per-chunk grouping can be computed with a single searchsorted over the touched chunk boundaries in `O(#chunks_touched · log n)`.

`arr[idx]`, `vindex[idx]`, and `get_coordinate_selection` all construct a `CoordinateIndexer`, so this covers the natural indexing syntax. And we have batched reads of CSR data stored in zarr for ML training, where each batch is a large sorted gather. And it is definetely a hot path.

## Evidence

Index construction only (no I/O), regular sharded 1-D array, ~1450-element runs (table written in md with AI):

| runs | elements | CoordinateIndexer build |
|-----:|---------:|------------------------:|
| 1024 |    1.48M |         8.7 ms → 0.5 ms |
| 4096 |    5.94M |          35 ms → 2.1 ms |
| 8192 |    11.9M |          71 ms → 4.5 ms |

Minimal repro (only needs `zarr` + `numpy`):

```python
import numpy as np, zarr, time
from zarr.core.indexing import CoordinateIndexer

n = 130 * 89_351_824
z = zarr.create_array(
    shape=(n,),
    chunks=(91_549,),
    shards=(89_351_824,),
    dtype="f4"
)
cg = z._async_array._chunk_grid
starts = np.sort(np.random.default_rng(0).choice(n // 1450, 4096, replace=False))
coords = np.concatenate([np.arange(s * 1450, (s + 1) * 1450) for s in starts])

t = time.perf_counter()
CoordinateIndexer((coords,), (n,), cg)
print((time.perf_counter() - t) * 1e3, "ms")
```

### Suggestion

If we added this codeblock around here in the constructor: https://github.com/zarr-developers/zarr-python/blob/ecc2d77718dbc4343da052347c4377945246d4fa/src/zarr/core/indexing.py#L1208

```python
if (
    isinstance(g0, FixedDimension)
    and g0.size > 0  # guard the divide below
    and coords.ndim == 1
    and coords.size > 0
    and coords[0] >= 0
    and coords[-1] < shape[0]
    and bool((np.diff(coords) >= 0).all())  # sorted ascending -> grouped by chunk
):
    size = g0.size
    first = int(coords[0]) // size
    last = int(coords[-1]) // size
    
    # count selected points per chunk in [first, last] via boundary searchsorted
    edges = np.arange(first, last + 2, dtype=np.intp) * size
    counts = np.diff(np.searchsorted(coords, edges))
    chunk_rixs = (first + np.nonzero(counts)[0]).astype(np.intp)
    
    chunk_nitems = np.zeros(nchunks, dtype=np.intp)
    chunk_nitems[first : last + 1] = counts
    chunk_nitems_cumsum = np.cumsum(chunk_nitems)

    object.__setattr__(self, "sel_shape", coords.shape or (1,))
    object.__setattr__(self, "selection", (coords,))
    object.__setattr__(self, "sel_sort", None)
    object.__setattr__(self, "chunk_nitems_cumsum", chunk_nitems_cumsum)
    object.__setattr__(self, "chunk_rixs", chunk_rixs)
    object.__setattr__(self, "chunk_mixs", (chunk_rixs,))
    object.__setattr__(self, "dim_grids", dim_grids)
    object.__setattr__(self, "shape", coords.shape or (1,))
    object.__setattr__(self, "drop_axes", ())
    return
```
kinda related convo: https://github.com/zarrs/zarrs-python/pull/147#issuecomment-3907264401

selmanozleyen:
## Summary

Speeds up `CoordinateIndexer` for sorted, in-bounds, one-dimensional integer selections on regular chunk grids.

Because the coordinates are already grouped by chunk, this uses `searchsorted` over the touched chunk boundaries instead of repeatedly processing the full selection.
| Runs | Elements | `main` | This PR |
|---:|---:|---:|---:|
| 1,024 | 1.48M | 8.7 ms | 0.5 ms |
| 4,096 | 5.94M | 35 ms | 2.1 ms |
| 8,192 | 11.9M | 71 ms | 4.5 ms |

Closes #4170

## Author attestation

## TODO

- [x] Add unit tests and/or doctests in docstrings
- [x] Add docstrings and API docs — not applicable; no public API changes
- [x] Update the user guide — not applicable; behavior is unchanged
- [x] Changes documented in `changes/`
- [x] GitHub Actions have all passed
- [x] Test coverage is 100% (Codecov passes)


src/zarr/core/indexing.py
```diff
@@ -1206,6 +1207,49 @@ def __init__(
...
+            # the normalization turned ints/lists into arrays. Only the sorted-1D-over-regular-grid
+            # shape is special-cased; everything else falls through to the general path below.
+            if (
+                isinstance(g0, FixedDimension)
+                and g0.size > 0  # guard the divide below
+                and coords.ndim == 1
+                and coords.size > 0
+                and coords[0] >= 0
+                and coords[-1] < shape[0]
+                and bool((np.diff(coords) >= 0).all())  # sorted ascending -> grouped by chunk
```
d-v-b: wont this wrap if coords is an unsigned int?

changes/4172.misc.md
```diff
@@ -0,0 +1,7 @@
+Added a fast path to `CoordinateIndexer` for sorted, in-bounds, one-dimensional integer coordinate
```
d-v-b: this needs to be `4172.misc.md`. a perf improvement is not really a feature.

src/zarr/core/indexing.py
selmanozleyen: Thank you!

This is a very nice catch. I made an update to it and now it's actually even faster this way :D 

```
For 11.9 million sorted int64 coordinates:

bool((np.diff(coords) >= 0).all()) | 11.26 ms
bool((coords[:-1] <= coords[1:]).all()) | 2.04 ms
```
Turns out it's because it doesn't create this intermediate int array before the `>=`.

src/zarr/core/indexing.py
```diff
@@ -1206,6 +1207,50 @@ def __init__(
...
+                last = int(coords[-1]) // size
+                # count selected points per chunk in [first, last] via boundary searchsorted
+                edges = np.arange(first, last + 2, dtype=np.intp) * size
+                counts = np.diff(np.searchsorted(coords, edges))
+                chunk_rixs = (first + np.nonzero(counts)[0]).astype(np.intp)
+                chunk_nitems = np.zeros(nchunks, dtype=np.intp)
+                chunk_nitems[first : last + 1] = counts
+                chunk_nitems_cumsum = np.cumsum(chunk_nitems)
+
+                object.__setattr__(self, "sel_shape", coords.shape or (1,))
```
d-v-b: when can coords.shape be falsy in this branch?

changes/4172.misc.md
d-v-b: this should be renamed to 4172.misc.md (after this PR number)

src/zarr/core/indexing.py
selmanozleyen: you are right, it can't be

selmanozleyen:
I did another performance and boundary-condition pass and found two additional cases worth handling:

- The boundary after the last selected chunk could overflow `np.intp`. I changed the calculation to search only internal chunk boundaries and derive the first and last counts from the selection length.
- Sorted but very sparse selections could be slower because the new implementation enumerated every chunk between the first and last coordinates. I added a cost check based on `chunk_span * log2(n)` so these selections fall through to the existing implementation.

For a selection spanning one million chunks, this removes the regression:

| | Time |
|---|---:|
| New implementation before the guard | 6.88 ms |
| Existing implementation | 3.31 ms |
| With the guard | 4.14 ms |
| Forced existing implementation in the same run | 4.23 ms |

The original large-gather workload still uses the new implementation by a wide margin. I also added regression tests for both cases.


Runs | Elements | Old CoordinateIndexer | Current implementation | Speedup
-- | -- | -- | -- | --
1,024 | 1.48M | 10.36 ms | 0.32 ms | 32.3×
4,096 | 5.94M | 49.15 ms | 1.59 ms | 30.9×
8,192 | 11.9M | 97.59 ms | 3.04 ms | 32.1×

d-v-b (approved):
looks good, thank you!

PR #4177, opened 2026-07-22
fix: informative errors for None in chunk specifications

d-v-b:
## Summary

Passing `None` as a `chunks` parameter results in an uninformative `TypeError`, when it should raise an informative `ValueError` instead. This PR makes that change, along with some improvements to the signature of the chunk normalization routines. These routines are now shaped like regular normalization functions, with a wide input type (`object`) and a narrow output type.

These changes were written by claude.

See https://github.com/d-v-b/zarr-python/pull/236 for the original claude-authored PR.

## Author attestation

## TODO


d-v-b:
i think this is a candidate for a self-merge. I plan to do so in ~8-16 hours

src/zarr/core/array.py
```diff
@@ -4461,7 +4461,17 @@ async def init_array(
...
             )
 
-    # Normalize the user's chunks into canonical ChunksTuple form
+    # Normalize the user's chunks into canonical ChunksTuple form.
+    # Auto-chunking is an API-level concept, so the guidance toward it is
+    # raised here rather than in the mechanical normalizer. Validate through
+    # an object-typed view: None/True are outside ChunksLike but reachable
+    # from untyped callers.
+    chunks_input: object = chunks
+    if chunks_input is None or chunks_input is True:
```
mkitti, line 4470: What happens if `chunks_input is False`?

mkitti: Rather than checking for a few invalid values, should the check be if chunks is either "auto", an int, or a tuple of ints?

mkitti, line 4470:
```suggestion
if not (
    chunks_input == "auto"
    or (isinstance(chunks_input, int) and not isinstance(chunks_input, bool))
    or (isinstance(chunks_input, tuple) and all(isinstance(v, int) and not isinstance(v, bool) for v in chunks_input))
):
```

Perhaps we can actually verify if the value is valid or not

mkitti (changes requested):

src/zarr/core/array.py
mkitti: Claude suggested something like this.

```python
def _is_valid_chunks(value):
    if value == "auto":
        return True
    if isinstance(value, int) and not isinstance(value, bool):
        return True
    if isinstance(value, tuple) and all(
        isinstance(v, int) and not isinstance(v, bool) for v in value
    ):
        return True
    return False


if not _is_valid_chunks(chunks_input):
    raise ValueError(
        f'{chunks_input!r} is not a valid chunk input. Use chunks="auto" or omit the chunks '
        "argument for automatic chunking, or pass an int / tuple of ints."
    )
```

d-v-b: > What happens if `chunks_input is False`?

chunks=false is valid for 2.x compatibility. it creates a single chunk for the whole array 🙃 

> Rather than checking for a few invalid values, should the check be if chunks is either "auto", an int, or a tuple of ints?

that leaves out iterables of ints, numpy arrays, etc. we have lower-level normalization for that.

d-v-b: we already have type narrowing routines for this. we just need to handle `"auto"` and deprecated values before calling those functions

d-v-b:
closing as redundant via upstream changes. some of these changes -- QOL improvements to the error messages for invalid types -- might re-appear in another PR

PR #4179, opened 2026-07-22
fix: FusedCodecPipeline falls back to async path for sharded arrays with async-only inner codecs

### LINKED ISSUE #4178: Unergonomic error for when using the FusedCodecPipeline + Sharding with an inner async-only codec
MRE:
```python
"""Reproducer: FusedCodecPipeline crashes on sharded arrays with async-only inner codecs.

A third-party codec that implements only the async codec interface
(`_encode_single` / `_decode_single`, no `SupportsSyncCodec`) works fine under
the default BatchedCodecPipeline, and under FusedCodecPipeline when used at the
top level of the codec chain. But as an *inner* codec of a sharded array under
FusedCodecPipeline, writing (and reading) raises:

    TypeError: All codecs must implement SupportsSyncCodec. The following do not: AsyncOnlyNoop

Run: python repro_fused_async_inner.py
"""

import numpy as np

import zarr
from zarr.abc.codec import BytesBytesCodec


class AsyncOnlyNoop(BytesBytesCodec):
    """No-op bytes-to-bytes codec implementing only the async codec interface."""

    is_fixed_size = True

    def to_dict(self):
        return {"name": "async-only-noop", "configuration": {}}

    @classmethod
    def from_dict(cls, data):
        return cls()

    def compute_encoded_size(self, input_byte_length, _spec):
        return input_byte_length

    async def _encode_single(self, chunk_bytes, chunk_spec):
        return chunk_bytes

    async def _decode_single(self, chunk_bytes, chunk_spec):
        return chunk_bytes


data = np.arange(64, dtype="i4").reshape(8, 8)

for pipeline in ("BatchedCodecPipeline", "FusedCodecPipeline"):
    with zarr.config.set({"codec_pipeline.path": f"zarr.core.codec_pipeline.{pipeline}"}):
        try:
            z = zarr.create_array(
                store={},
                shape=(8, 8),
                chunks=(2, 2),
                shards=(4, 4),
                dtype="i4",
                compressors=(AsyncOnlyNoop(),),  # inner codec chain of the shard
            )
            z[:] = data
            ok = np.array_equal(z[:], data)
            print(f"{pipeline}: round-trip {'OK' if ok else 'CORRUPTED'}")
        except TypeError as e:
            print(f"{pipeline}: {type(e).__name__}: {e}")
```

To fix this and raise an error at the correct time (codecpipeline construction time) we need to make some changes to the fused codec pipeline. those will come in a PR.

d-v-b:
## Summary

This PR adds a private method to the sharding codec to allow it to report whether the members of its `codecs` attributes support synchronous execution. I think this is an unfortunate move, and we should more formally approach the sync vs async nature of codecs. But that requires new codec classes, which is a lot of work. Until then, this change is sufficient to avoid a confusing error at chunk encode / decode time.

closes #4178 

The original PR by claude is here: https://github.com/d-v-b/zarr-python/pull/237

## For reviewers

## Author attestation

## TODO


d-v-b:
this is a usability improvement for the unreleased, opt-in fused codec pipeline. it's also spending some tech debt by adding new private methods to the codec API. I plan to self-merge this today.

d-v-b:
i'm self-merging this as part of a larger suite of fixes to the fused codec pipeline prior to the 3.3.0 release.

PR #4183, opened 2026-07-23
fix(ArraySpec): proper and robust equality semantics for ArraySpec by checking fill_value for byte-identicality, fixes #3054.

### LINKED ISSUE #3054: sharding codec use of lru caching fails with numpy void scalars
numpy void scalars are not hashable, which means that a class with a np.void attribute cannot be hashed by `lru_cache`, which is used by the sharding codec class [here](https://github.com/zarr-developers/zarr-python/blob/7584b96f22b5d517cf61f13b415961d7be99b428/src/zarr/codecs/sharding.py#L358).

I ran into this In #2874. Any ideas for fixing this would be appreciated! IMO the ideal fix would be to resolve the basic performance problems we are solving with this caching layer.

sehoffmann:
## Summary

`ArraySpec` describes the exact layout for an array. The previously auto-generated `__eq__` and `__hash__` methods from `@dataclass` did not reflect this correctly due to the `fill_value` attribute.

Specifically:
* For IEEE float nans:  `nan != nan` and thus `ArraySpec(**kwargs) != ArraySpec(**kwargs)` if the fill value is nan (although it they are in fact identical as they have the same fill). 
* Same issue if numpy's `np.datetime64('NaT')` which is also not self-equal
* On the other hand, `float(-0.0) == float(0.0)` despite these two fill_value's being different float numbers. They would result in different arrays in memory and on disk.
* Most importantly, if fill_value is a `np.void`, such as from a structured dtype, `hash(ArraySpec(...))` failed outright, because (mutable) `np.void`s are not hashable. (Issue #3054).

This PR fixes all of the above by comparing `fill_value` using byte-identity, which should be the correct semantics for a fill value. It also adds a comprehensive test suite for the equality and hash semantics of `ArraySpec` which was non-existent before.

Before this fix/PR, the following (newly added) tests failed:
```python
FAILED tests/test_array_spec.py::**TestArraySpecHashEq::test_hashable[structured-void]** - TypeError: unhashable type: 'writeable void-scalar'
FAILED tests/test_array_spec.py::**TestArraySpecHashEq::test_equal_specs_hash_equal[float64-nan]** - AssertionError: assert ArraySpec(shape=(4, 4), dtype=Float64(endianness='little'), fill_value=np.float64(nan), config=ArrayConfig(order='C', ...otype=BufferPrototype(buffer=<class 'zarr.core.buffer.cpu.B...
FAILED tests/test_array_spec.py::**TestArraySpecHashEq::test_equal_specs_hash_equal[datetime64-NaT]** - AssertionError: assert ArraySpec(shape=(4, 4), dtype=DateTime64(endianness='little', scale_factor=1, unit='s'), fill_value=np.datetime64('NaT...otype=BufferPrototype(buffer=<class 'zarr.core.buffer.cpu.B...
FAILED tests/test_array_spec.py::**TestArraySpecHashEq::test_equal_specs_hash_equal[structured-void]** - TypeError: unhashable type: 'writeable void-scalar'
FAILED tests/test_array_spec.py::**TestArraySpecHashEq::test_distinct_specs_unequal[shape-structured-void]** - TypeError: unhashable type: 'writeable void-scalar'
FAILED tests/test_array_spec.py::**TestArraySpecHashEq::test_distinct_specs_unequal[order-structured-void]** - TypeError: unhashable type: 'writeable void-scalar'
FAILED tests/test_array_spec.py::**TestArraySpecHashEq::test_distinct_specs_unequal[prototype-structured-void]** - TypeError: unhashable type: 'writeable void-scalar'
FAILED tests/test_array_spec.py::**TestArraySpecHashEq::test_signed_zero_fills_are_distinct[float16]** - AssertionError: assert ArraySpec(shape=(4, 4), dtype=Float16(endianness='little'), fill_value=np.float16(-0.0), config=ArrayConfig(order='C',...otype=BufferPrototype(buffer=<class 'zarr.core.buffer.cpu.B...
FAILED tests/test_array_spec.py::**TestArraySpecHashEq::test_signed_zero_fills_are_distinct[float32]** - AssertionError: assert ArraySpec(shape=(4, 4), dtype=Float32(endianness='little'), fill_value=np.float32(-0.0), config=ArrayConfig(order='C',...otype=BufferPrototype(buffer=<class 'zarr.core.buffer.cpu.B...
FAILED tests/test_array_spec.py::**TestArraySpecHashEq::test_signed_zero_fills_are_distinct[float64]** - AssertionError: assert ArraySpec(shape=(4, 4), dtype=Float64(endianness='little'), fill_value=np.float64(-0.0), config=ArrayConfig(order='C',...otype=BufferPrototype(buffer=<class 'zarr.core.buffer.cpu.B...
FAILED tests/test_array_spec.py::**TestArraySpecHashEq::test_signed_zero_fills_are_distinct[complex128-both]** - AssertionError: assert ArraySpec(shape=(4, 4), dtype=Complex128(endianness='little'), fill_value=np.complex128(-0-0j), config=ArrayConfig(ord...otype=BufferPrototype(buffer=<class 'zarr.core.buffer.cpu.B...
FAILED tests/test_array_spec.py::**TestArraySpecHashEq::test_signed_zero_fills_are_distinct[complex128-imag]** - AssertionError: assert ArraySpec(shape=(4, 4), dtype=Complex128(endianness='little'), fill_value=np.complex128(-0j), config=ArrayConfig(order...otype=BufferPrototype(buffer=<class 'zarr.core.buffer.cpu.B...
FAILED tests/test_array_spec.py::**TestArraySpecHashEq::test_signed_zero_fills_are_distinct[complex128-real]** - AssertionError: assert ArraySpec(shape=(4, 4), dtype=Complex128(endianness='little'), fill_value=np.complex128(-0+0j), config=ArrayConfig(ord...otype=BufferPrototype(buffer=<class 'zarr.core.buffer.cpu.B...
FAILED tests/test_array_spec.py::**TestArraySpecHashEq::test_signed_zero_fills_are_distinct[structured]** - AssertionError: assert ArraySpec(shape=(4, 4), dtype=Struct(fields=(('a', Float64(endianness='little')),)), fill_value=np.void((-0.0,), dtype...otype=BufferPrototype(buffer=<class 'zarr.core.buffer.cpu.B...
```

With the fix applied, all tests pass. Also, before this fix, the following valid zarr code would throw an exception:
```python
t = np.dtype([("a", "i4"), ("b", "f4")])

arr = zarr.create_array(
    MemoryStore(),
    shape=(8,),
    chunks=(2,),
    shards=(4,),
    dtype=dt,
    fill_value=(0, 0.0),
)
arr[:] = np.ones(8, dtype=dt)  # -> TypeError: unhashable type: 'writeable void-scalar'
```
This PR fixes the above.

### Behavioral Changes (summarized):
* Structured / np.void fills: hash(spec) no longer raises TypeError; such specs are now hashable and usable as dict/set/cache keys. (Fixes #3054.)
* nan fills: two specs with a NaN fill now compare equal and hash equal; before they compared unequal (nan != nan).
* NaT fills: same change as NaN: now equal, were unequal.
* Signed zero (+0.0 vs -0.0): now compare unequal (distinct fills, distinct bytes); before they compared equal.
* Comparison basis: numpy-scalar fills are now compared by exact bit pattern rather than numeric ==, so equality no longer inherits numpy's value-promotion quirks. (e.g. see https://github.com/numpy/numpy/issues/32071)

## For reviewers

#3054 discusses getting rid of the lru cache entirely. This is an orthogonal design decision in my opinion. While this PR fixes the exception, it does not remove the lru cache. Instead, it ensures that `ArraySpec` has robust hash semantics in the first place which is a good thing regardless of #3054 and whether an lru cache is used or not.

## Author attestation

## TODO


sehoffmann:
Should I add a test case that non-ArraySpec comparison raises as requested by the codecov plugin? IMO this is a trivial case..

d-v-b:
Thanks, this is a great fix. I will leave a few comments, but IMO this is good to go

> Should I add a test case that non-ArraySpec comparison raises as requested by the codecov plugin? IMO this is a trivial case..

only if you feel like it.

tests/test_array_spec.py
```diff
@@ -0,0 +1,182 @@
...
+
+
+MUTATIONS = [
+    pytest.param(_grow_shape, id="shape"),
+    pytest.param(_flip_order, id="order"),
+    pytest.param(_swap_prototype, id="prototype"),
+]
+
+
+class TestArraySpecHashEq:
```
d-v-b: i don't know why we need a class here. IMO plain functions and data are simpler

tests/test_array_spec.py
```diff
@@ -0,0 +1,182 @@
...
+        id="structured-void",
+    ),
+    pytest.param({"native_dtype": "U5", "fill_value": "hello"}, id="fixed-string"),
+    pytest.param({"shape": ()}, id="scalar-shape"),
+    pytest.param({"shape": (0,)}, id="zero-size"),
+    pytest.param({"order": "F"}, id="order-F"),
+]
+
+
+# Mutations: each mutate kwargs to an uneqal version
```
d-v-b:
```suggestion
# Mutations: each mutate kwargs to an unequal version
```

tests/test_array_spec.py
```diff
@@ -0,0 +1,182 @@
...
+    def test_distinct_specs_unequal(
+        self,
+        mutate: Callable[[dict[str, Any]], dict[str, Any]],
+        kwargs: dict[str, Any],
+    ) -> None:
+        """Changing one dtype-independent field makes a spec unequal to its base."""
+        base = _make_spec(**kwargs)
+        variant = _make_spec(**{**kwargs, **mutate(kwargs)})
+        assert base != variant
+        assert hash(base) != hash(variant)
```
d-v-b: hashes collisions between unequal objects can occur, right? so maybe we just need to check that they are unequal, and not that the hashes differ?

d-v-b:
can you update the stale comment in `sharding.py` or re-enable the LRU cache?

tests/test_array_spec.py
sehoffmann: I agree, a != b => hash(a) != hash(b) is not guaranteed. Will be fixed

tests/test_array_spec.py
sehoffmann: Will be adjusted

sehoffmann:
@d-v-b Thanks for the positive feedback. All your points should have been addressed by the recent changes.

d-v-b:
thanks @sehoffmann, do you mind if I push a commit to deal with the codecov issue? and we need a changelog entry (i'm also happy to write that if you don't mind)

sehoffmann:
@d-v-b Yes, for sure. I would be very glad if you could do that for me, thanks :) Also wrt. the changelog entry.

d-v-b (approved):

PR #4189, opened 2026-07-27
fix: allow `require_array` to accept a `ZDType`

### LINKED ISSUE #3377: `group.require_array` is not compatible with new dtypes
[this line](https://github.com/zarr-developers/zarr-python/blob/c21d1f92f7f2648a4c55b164510c524a951ac11d/src/zarr/core/group.py#L1271) will trigger a failure if you attempt to use a `ZDType` instance, which is incorrect and should be fixed.

cc @TomNicholas

JOhnsonKC201:
## Summary

Closes #3377.

`Group.require_array` could not accept a Zarr data type, even though every other
array-creation method on the group already does. Requiring an array that exists
raised:

```
TypeError: Cannot interpret 'Int32(endianness='little')' as a data type
```

`AsyncGroup.require_array` was annotated `npt.DTypeLike | None` and normalised
its argument with `np.dtype(dtype)`, which cannot consume a `ZDType`. Its
siblings (`create_array` and friends) take `ZDTypeLike`. So the "array already
exists" branch always blew up for a `ZDType`, while the string and NumPy forms
worked.

The fix widens the annotation to `ZDTypeLike | None` and normalises with
`parse_data_type(...).to_native_dtype()`. `AsyncArray.dtype` already returns the
native dtype, so the `exact` / `np.can_cast` comparisons below are unchanged.

Two follow-on details: `parse_data_type(None)` resolves to `float64`, which is
exactly what `np.dtype(None)` gave, so the default path is unchanged; and
removing the last real use of `numpy.typing` left the import unused, so it is
dropped (the three remaining `npt.` mentions in this file are in docstrings).

## For reviewers

The bit worth a second look is the `None` case, since `np.dtype(None)` silently
means `float64` and I wanted to preserve that rather than start rejecting it. I
checked all four input forms round-trip to the same array dtype: `ZDType`,
`np.dtype`, `"int32"`, and `None`.

I am confident about the rest: it is a normalisation swap, and behavior for
string and NumPy dtypes is identical.

## Author attestation

## TODO

## Test plan

- `tests/test_group.py::test_require_array_zdtype`, covering `ZDType`,
  `np.dtype` and `str`. It fails on main with the `TypeError` above.
- `tests/test_group.py`: 529 passed, 15 skipped.
- `tests/test_api.py tests/test_array.py`: 1654 passed, 44 skipped.
- `ruff check` and `ruff format --check` clean.


tests/test_group.py
```diff
@@ -1366,6 +1367,19 @@ async def test_require_array(store: Store, zarr_format: ZarrFormat) -> None:
         await root.require_array("bar", shape=(10,), dtype="int8")
 
 
+async def test_require_array_zdtype(store: Store, zarr_format: ZarrFormat) -> None:
```
d-v-b: can you parametrize this test over (input, expected output) pairs, and ensure that the `dtype=None` case is included? Pairs like `[("int32", Int32()), (None, Int64()]`

d-v-b: correction: the pair for the `None` case would be `(None, Float64())`

d-v-b:
hi @JOhnsonKC201 is it OK if I push some changes to your branch to get some of the remaining issues resolved? we just need a few changes before this can be merged

JOhnsonKC201:
Sure, go ahead.

I just pushed 98376b1, which parametrizes `test_require_array_zdtype` over (input, expected) pairs with the `(None, Float64())` case included, since that looked like the open review comment. It landed a few minutes after you asked, so it may not have been there when you looked.

Happy for you to push whatever else is needed to get this over the line.

d-v-b (approved):

d-v-b:
thanks @JOhnsonKC201!

PR #4194, opened 2026-07-28
fix: keep FusedCodecPipeline compute off the event-loop thread

d-v-b:
## Summary

`FusedCodecPipeline.read` and `FusedCodecPipeline.write` run on the event loop thread, which means CPU-bound code it spins up can choke out concurrent awaitables queued in the event loop. This PR instead makes `read` and `write` send work to a thread via `asyncio.to_thread`, which offers a performance win for the `FusedCodecPipeline`. 

Although the perf we get here is good, the pattern is problematic: the codec pipeline is taking charge of how its resource-intensive routines are being scheduled in a manner that is opaque to a caller. There are a few better ways of solving the fundamental problem here, but they require classes that look different. 

## For reviewers

Please check to ensure that this change isn't problematic from a performance POV.

## Author attestation

## TODO

based on work by claude: see https://github.com/d-v-b/zarr-python/pull/247/


d-v-b:
this PR is important for the unreleased fused codec pipeline, which is itself an opt-in feature, so I'm going to self-merge this.

PR #4198, opened 2026-07-29
feat(sharding): manage inner chunks as a semi-regular grid (sharding_indexed v1.1)

d-v-b:
## Summary

Implements a proposed extension to the sharding codec spec which allows semi-regular inner chunk grids. Meaning, with this sharding codec implementation, the outer chunk shape can be `(10,)`, and the inner chunk shape can be _declared_ as `(8,)`, resulting in actual inner chunks sized `(8, 2)`, i.e. 10 partitioned into chunks sized 8, with the remainder. This makes the sharding codec much more useful for arrays with rectilinear chunk grids, or arrays with shapes made of prime numbers.

Implementation-wise, the biggest change is a layer of dynamism for the per-chunk models used by the sharding codec. chunk spec objects have to be generated dynamically per-chunk, because the inner chunk grid depends on the outer chunk shape.

## For reviewers

This is a proof of concept intended to demonstrate the impact of implementing the proposed sharding codec change. Widening our model of the sharding codec has ecosystem-wide implications, so we will not merge this without either ecosystem-wide coordination or an explicit "I accept that I am going off-road" mechanism for zarr-python users to opt into. So do not consider this review-ready until I advertise that one or both of those conditions are met. 

## Author attestation

## TODO

these changes were written with claude. See the claude-authored PR here: https://github.com/d-v-b/zarr-python/pull/251


d-v-b:
see https://github.com/zarr-developers/zarr-specs/pull/370

PR #4201, opened 2026-07-29
fix: fused pipeline falls back for partial-mixin codecs without sync partial methods

d-v-b:
This PR avoids a crash mid-decode in the fused codec pipeline when a codec advertises public partial encode / decode mixin without supporting the sync methods. This PR ensures that we catch this case early and fall back to full chunk decoding / encoding explicitly.

based on this PR: https://github.com/d-v-b/zarr-python/pull/254

## Summary

## For reviewers

## Author attestation

## TODO

### DIFF
--- a/src/zarr/core/codec_pipeline.py
+++ b/src/zarr/core/codec_pipeline.py
@@ -1039,10 +1039,14 @@ class FusedCodecPipeline(CodecPipeline):
 
         # Partial-decode fast path: the AB codec owns IO (read only the
         # byte ranges needed for the requested selection). Same condition
-        # and dispatch as BatchedCodecPipeline.read_batch.
-        if self.supports_partial_decode:
-            codec = self.array_bytes_codec
-            assert hasattr(codec, "_decode_partial_sync")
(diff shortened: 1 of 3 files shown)


src/zarr/core/codec_pipeline.py
```diff
@@ -1039,10 +1039,14 @@ def read_sync(
...
-            codec = self.array_bytes_codec
-            assert hasattr(codec, "_decode_partial_sync")
+        # and dispatch as BatchedCodecPipeline.read_batch, plus a gate on the
+        # sync partial method: the public partial-decode contract
+        # (`ArrayBytesCodecPartialDecodeMixin`) only requires the async
+        # `_decode_partial_single`, so a codec may support partial decode
+        # without `_decode_partial_sync` — such codecs take the full-chunk
+        # path below instead.
+        codec = self.array_bytes_codec
+        if self.supports_partial_decode and hasattr(codec, "_decode_partial_sync"):
```
d-v-b, line 1049: a very cursed blend of two separate ways to advertise similar capabilities. this will need a deep rethink of our codec API to fix

d-v-b:
I'm self-merging this!

Issue #4174, opened 2026-07-22
set_coordinate_selection runs out-of-memory for large arrays


sehoffmann:
### Zarr version

v3.2.1

### Numcodecs version

v0.16.5

### Python Version

3.12.11

### Operating System

Linux

### Installation

uv .venv

### Description

Cause is the `np.bincount` call which will allocate a very large array even if we only set a handful of indices:
```
Traceback (most recent call last):
  File "bug.py", line 29, in <module>
    arr.set_coordinate_selection((rows, cols), vals)  # set_coordinate_selection: MemoryError
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "site-packages/zarr/core/array.py", line 3584, in set_coordinate_selection
    indexer = CoordinateIndexer(selection, self.shape, self._chunk_grid)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "site-packages/zarr/core/indexing.py", line 1252, in __init__
    chunk_nitems = np.bincount(chunks_raveled_indices, minlength=nchunks)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
numpy._core._exceptions._ArrayMemoryError: Unable to allocate 1.01 TiB for an array with shape (138540000000,) and data type int64
```
It makes scaling O(chunks) as well

### Steps to reproduce

```python
# /// script
# requires-python = ">=3.12"
# dependencies = [
#   "zarr@git+https://github.com/zarr-developers/zarr-python.git@main",
# ]
# ///
#
# This script automatically imports the development branch of zarr to check for issues

import zarr
import numpy as np

# chunks = 10_000_000 * 13_854
arr = zarr.create_array(
    zarr.storage.MemoryStore(),
    shape=(10_000_000, 13_854),
    chunks=(1, 1),
    dtype='int32',
    fill_value=0,
)

# we write 3 points
rows = np.array([0, 1, 2])
cols = np.array([5, 5, 5])
vals = np.array([1, 2, 3], dtype='int32')

arr[0, 5] = 1                                     # set_basic_selection: OK
arr.set_orthogonal_selection((rows, 5), vals)     # set_orthogonal_selection: OK
[...]

PR #4218, opened 2026-07-30
fix(chunk-grids): make array creation O(1) in per-dimension chunk count

d-v-b:
## Summary

this is a claude-authored fix of #https://github.com/zarr-developers/zarr-python/issues/4174. in `ChunkLayout`,  instead of materializing all chunk specifications, even regular ones, as numpy array, we specifically model regular chunk grid dimensions as a single shape + extent. This means `ChunksTuple` can just disappear, and we can re-use our chunk grid abstractions (`FixedDimension` and `VaryingDimension`), which is nice and simple. This avoids o(num chunks) memory problems and avoids some unnecessary abstractions.

this is basically what @maxrjones suggested here: https://github.com/zarr-developers/zarr-python/pull/3899#discussion_r3111900128. My instinct to make the API "uniform" came at the cost of an inefficient representation, so I'm pulling back on the commitment to "uniformity" here.

Here's claude's summary: 

Chunk normalization now returns a ChunkGrid whose uniform dimensions are stored as FixedDimension (size + extent) instead of one array entry per chunk, so create_array(shape=(2**62,), chunks=(1,)) succeeds instantly instead of raising "array is too big", and chunks=(1, 1) on a (2**31, 2**31) array no longer allocates ~17 GB per dimension.

Explicit per-chunk lists collapse to FixedDimension when they describe a regular grid; genuinely irregular lists become VaryingDimension. Both variants bind chunk sizes to their extent, so the two forms carry the same invariants. The intermediate ChunksTuple type and as_regular_shape helper are removed; create_chunk_grid_metadata consumes the grid directly and serializes uniform dimensions of mixed rectilinear grids as the spec's bare-int step-size shorthand.

Creation-time counterpart of the gh-4174 indexing fix.

## For reviewers

Check if this purported simplification is in fact simpler, and confirm that we don't see any unpleasant performance side effects.

## Author attestation

## TODO


d-v-b:

Rebased onto current `main` (7729d924) and added a fix for #4272 on top. Summary of what changed:

### Rebase

The conflict was with #4257 ("accept numpy integers as chunk sizes"), which landed after this branch was cut and rewrote the same region of `normalize_chunks_1d`. Its semantics are preserved inside the new `FixedDimension`/`VaryingDimension` code:

- scalars are narrowed with `numbers.Integral` (not `int`) and coerced via `int()`, so numpy integer scalars take the uniform-chunk path;
- the `-1` sentinel check lives inside the integral branch, so numpy-array chunk specs never hit an ambiguous-truth-value error on `chunks == -1`;
- the improved `TypeError` for non-integer, non-iterable specs is kept, along with #4257's tests (translated to the `FixedDimension`/`VaryingDimension` expectation format).

`normalize_chunks_1d` no longer needs main's `span == 0` special case — `FixedDimension` handles zero extents natively. In `zarr/testing/stateful.py` the branch's `chunks=`-reconstruction hunk was dropped in favor of main's newer `zarr.from_array(..., write_data=False)` approach; `chunks_param_from_rectilinear` stays, since the `arrays()` strategy still uses it.

### New commit: explicit per-chunk lists stay rectilinear (fixes #4272)

The explicit-list branch of `normalize_chunks_1d` previously collapsed uniform (or uniform-plus-short-tail) lists to `FixedDimension`, so an explicitly rectilinear spec like `[[10, 10, 4]]` was silently stored as a regular grid — which changes resize semantics (a regular grid grows by extending the uniform pattern; a rectilinear grid appends an edge chunk), breaking append-oriented layouts like `(168,) * 13 + (24,)`. That collapse is now removed: the grid kind follows the input syntax, matching 3.2.x — scalar specs (ints, numpy integers, `-1`) produce `FixedDimension` → regular grid (the O(1) path this PR is about), explicit lists always produce `VaryingDimension` → rectilinear grid. Uniform dimensions of mixed scalar/list specs still serialize as the spec's bare-int step-size shorthand. New tests cover the syntax→kind mapping and the resize divergence (the rectilinear append touches exactly one new chunk; the regular pattern-extension straddles two).

### Relation to #4290

#4290 fixes the same issue on current `main` by threading a `requested_rectilinear` flag from the raw input into `create_chunk_grid_metadata`, because main's `ChunksTuple` normalization erases the form of the request. With this PR's representation the flag isn't needed — `FixedDimension` vs `VaryingDimension` records the request's form, so `create_chunk_grid_metadata` dispatches on the grid with no call-site plumbing, and rectilinear `shards=` producing a rectilinear outer grid falls out automatically. End behavior matches #4290 (including uniform nested specs now hitting the `array.rectilinear_chunks` gate). Credit to #4290 for the touched-chunk-keys assertion style, which the resize regression test here adopts.

d-v-b:
@maxrjones could you review this when you get a chance? It fixes two issues and so I consider it high priority for our next release

maxrjones (commented):
So much better ! Thanks @d-v-b! Also sorry for my lack of communication about a review timeline.

On your unpleasant surprises question, below are three tests that I think need to pass before the PR is merged. Everything else I came up with are nits that can go in follow-ups.


**Test 1 — the default `from_array` must work on a rectilinear source** (currently: `NotImplementedError` on `.chunks` from the data-copy path; the PR's own test hides this behind `write_data=False`):

```python
def test_from_array_rectilinear_default_copies_data() -> None:
    src = zarr.create_array(MemoryStore(), shape=(6,), chunks=[[2, 2, 2]], dtype="uint8")
    src[:] = 7
    dst = zarr.from_array(MemoryStore(), data=src, name="0")  # write_data=True default
    assert isinstance(dst.metadata, ArrayV3Metadata)
    assert dst.metadata.chunk_grid == src.metadata.chunk_grid
    assert (dst[:] == src[:]).all()
```

**Test 2 — `chunks="keep"` must preserve the bare-int shorthand** (currently fails with `assert ((3, 3, 3), (10, 20)) == (3, (10, 20))` — the shorthand is expanded per-chunk, which is the O(nchunks) path and changes the copy's resize semantics):

```python
def test_from_array_keep_preserves_bare_int_shorthand() -> None:
    src = zarr.create_array(MemoryStore(), shape=(9, 30), chunks=(3, [10, 20]), dtype="uint8")
    dst = zarr.from_array(MemoryStore(), data=src, name="0", write_data=False)
    grid = dst.metadata.chunk_grid
    assert isinstance(grid, RectilinearChunkGridMetadata)
    assert grid.chunk_shapes == (3, (10, 20))
```

Deliberately small: a `shape=(2**62, 30)` variant would pin the O(1) property too, but before the fix it hangs rather than fails, which makes a bad red test. After fixing, consider using `shape=(2**62, 30)`

**Test 3 — rectilinear-sharded arrays: `.info` works and sharding survives `from_array`** (currently fails at `src.info` with `NotImplementedError` on `.shards`; the tail assertions then catch the silent sharding drop):

```python
def test_from_array_keep_preserves_rectilinear_sharding() -> None:
    src = zarr.create_array(
        MemoryStore(), shape=(100,), chunks=(10,), shards=[[50, 50]], dtype="int32"
    )
    assert src.info is not None  # _info must not raise on the shards accessor
    dst = zarr.from_array(MemoryStore(), data=src, name="0", write_data=False)
    assert any(isinstance(c, ShardingCodec) for c in dst.metadata.codecs)
    assert dst.metadata.chunk_grid == src.metadata.chunk_grid
```

Another option is to raise on this test, but either way sharding shouldn't be *silently* dropped.

d-v-b:
Thanks max, I'll get fixes in on this PR and explore how our tests allowed them to slip through

d-v-b:
@maxrjones have a look at the latest changes. `from_array` now handles the three cases you flagged, and I had our tests upgraded to ensure that `from_array` runs through a similar test matrix as `create_array`.

one important change: `.chunks` now works for rectilinear chunk grids that use sharding, because the inner chunk grid of sharding is regular (today).

d-v-b:
@ilan-gold do you have time to review this?

changes/4272.bugfix.md
```diff
@@ -0,0 +1,22 @@
+Explicit per-chunk size lists now always produce a rectilinear chunk grid,
+even when the sizes happen to describe a regular grid (all equal, or all equal
+with a smaller trailing chunk). Previously such input was silently collapsed to
```
ilan-gold, line 3: Just to be clear, this is actually gated behind the feature flag, right?

docs/user-guide/arrays.md
```diff
@@ -674,8 +674,11 @@ z_regular = zarr.create_array(
 print(z_regular.write_chunk_sizes)
 ```
 
-Note that the `.chunks` property is only available for regular chunk grids. For
-rectilinear arrays, use `.write_chunk_sizes` (or `.read_chunk_sizes`) instead.
+Note that the `.chunks` property is not available for non-sharded rectilinear
+arrays, since there is no single uniform chunk shape — use `.write_chunk_sizes`
+(or `.read_chunk_sizes`) instead. Sharded arrays always have `.chunks`: it
+returns the inner chunk shape, which is regular even when the shard grid is
+rectilinear (see [Rectilinear shard boundaries](#rectilinear-shard-boundaries)).
```
ilan-gold: I actually don't know this - what if the inner chunking is rectilinear? Is this even possible? I would guess not given how sharding works with byte offsets

src/zarr/core/array.py
```diff
@@ -4765,6 +4778,33 @@ async def create_array(
...
+    """The array's sharding codec, or None if the array is not sharded.
+
+    Zarr format 2 arrays are never sharded.
+    """
+    if isinstance(metadata, ArrayV3Metadata):
+        return metadata.sharding_codec
+    return None
+
+
+def _stored_rectilinear_grid(metadata: ArrayMetadata) -> RectilinearChunkGridMetadata | None:
```
ilan-gold:
```suggestion
def _stored_rectilinear_grid_or_none(metadata: ArrayMetadata) -> RectilinearChunkGridMetadata | None:
```
I personally find it confusing when a function is called something but then might not do that thing.

ilan-gold: I can also be a bit verbose with naming

maxrjones (commented):
A few more road bumps:

---

`from_array` with the default `chunks="keep"` on a rectilinear source that has been shrunk via `resize`. Shrinking legally leaves trailing edges beyond the extent (`update_shape` keeps them on purpose), but the keep path feeds the stored spec into the strict parser, which requires the edges to sum to the span:
```python
def test_from_array_keep_after_shrink_resize() -> None:
    src = zarr.create_array(MemoryStore(), shape=(25,), chunks=[[5, 10, 10]], dtype="uint8")
    src[:] = np.arange(25, dtype="uint8")
    src.resize((18,))  # stored edges legally remain (5, 10, 10)
    dst = zarr.from_array(MemoryStore(), data=src, name="0")
    assert dst.metadata.chunk_grid == src.metadata.chunk_grid
    np.testing.assert_array_equal(dst[:], src[:])
```

   Currently raises `ValueError: Chunk sizes [5, 10, 10] do not sum to span 18` (`_parse_keep_array_attr` → `normalize_chunks_nd`); on `main` this copy succeeded via the clipped `write_chunk_sizes`. The `shards="keep"` branch fails the same way (`shards=[[50, 50]]`, `resize((80,))`). The `normalize_chunks_nd` docstring already says stored metadata belongs on the tolerant `from_sizes`/`from_metadata` rules — the keep path just needs to follow its own advice (exact preservation as asserted above, or at minimum clipping to the extent like `main` did). Worth a `resize`-then-copy entry in `FROM_ARRAY_KEEP_CASES`.

---

Out-of-bounds selections on `_iter_shard_regions` failing loudly, consistent with `_iter_shard_keys`. The switch from `_iter_regions`/`_iter_grid` to `ChunkGrid.iter_chunk_regions` dropped the bounds check, so the two sibling APIs now disagree on identical inputs:

```python
def test_iter_shard_regions_bounds_check() -> None:
    arr = zarr.create_array(MemoryStore(), shape=(30,), chunks=(10,), dtype="uint8")
    with pytest.raises(IndexError):
        list(arr._iter_shard_keys(origin=(1,), selection_shape=(5,)))
    with pytest.raises(IndexError):
        list(arr._iter_shard_regions(origin=(1,), selection_shape=(5,)))
```

   Currently the second call silently yields 2 regions instead of raising. Since these regions are documented as the safe concurrent-write partitions, a miscomputed selection now silently writes fewer regions rather than failing.

---

3. Mixed per-dimension specs hitting the intended error messages regardless of dimension order. The widened `ChunksLike` makes `(5, [10, 20, 70])` official, but `_is_rectilinear_chunks` only samples the first element, so the v2 and chunks+shards gates are order-dependent:

```python
@pytest.mark.parametrize("chunks", [(5, [10, 20, 70]), ([10, 20, 70], 5)])
def test_mixed_chunks_with_shards_raises(chunks) -> None:
    shape = (30, 100) if isinstance(chunks[0], int) else (100, 30)
    with pytest.raises(ValueError, match="Rectilinear chunks with sharding is not supported"):
        zarr.create_array(MemoryStore(), shape=shape, chunks=chunks, shards=(10, 10), dtype="uint8")
    with pytest.raises(ValueError, match="Zarr format 2 does not support rectilinear chunk grids"):
        zarr.create_array(MemoryStore(), shape=shape, chunks=chunks, zarr_format=2, dtype="uint8")
```

   Currently the bare-int-first parametrization dies later with the internal `chunk_shape is only available for regular chunk grids. Use grid[coords] for per-chunk sizes.` — advice naming a private API. Checking `any(...)` across dimensions instead of the first element fixes both gates at once.

---

Two smaller notes, no test required:

* The changelog line "creation-time counterpart of the indexing fix in #4172" oversells the pairing — #4172 only covers sorted 1-D coordinate selections, and the original #4174 repro still hits the 1.01 TiB `np.bincount` allocation at `set_coordinate_selection` time on this branch (creation itself is now instant, as promised). Suggest rewording so #4174 stays open for the indexing half.
* `_nchunks_initialized` now walks every shard coordinate even for regular grids, where the old `nshards_initialized * chunks_per_shard` multiply is exact and O(1) after the listing (~4× slower at 10k shards, linear beyond). Keeping the multiply for regular grids and iterating only the initialized keys for rectilinear ones would restore that, and in the spirit of preferring small tests that fail over big ones that hang, an assertion that the two computations agree on a regular sharded grid would pin it.

d-v-b:
thanks for spotting these! fixes inbound

src/zarr/core/chunk_grids.py
```diff
@@ -717,32 +700,35 @@ def _guess_regular_chunks(
...
             # matching the auto-chunking clamp in _guess_regular_chunks.
-            return np.array([max(span, 1)], dtype=np.int64)
+            return FixedDimension(size=max(span, 1), extent=span)
         if chunk_size <= 0:
             raise ValueError(f"Chunk size must be positive, got {chunk_size}")
-        if span == 0:
-            return np.array([chunk_size], dtype=np.int64)
-        n = ceildiv(span, chunk_size)
-        return np.full(n, chunk_size, dtype=np.int64)
+        return FixedDimension(size=chunk_size, extent=span)
```
ilan-gold, line 747: I have no idea why GitHub won't let me suggest a fix but to be clearer:

```python
        if chunk_size < -1 or chunk_size == 0:
            raise ValueError(f"Chunk size must be positive or -1, got {chunk_size}")
        if chunk_size == -1:
            # A zero-length span still gets chunk size 1 (chunk sizes must be positive),
            # matching the auto-chunking clamp in _guess_regular_chunks.
            return FixedDimension(size=max(span, 1), extent=span)
        return FixedDimension(size=chunk_size, extent=span)

ilan-gold (commented):
I'm not too familiar with this part of the codebase but it was nice to see it. One thing I don't get even by reading the docs is the difference in `{read,write}_chunks_sizes` which is pretty relevant here.

https://zarr.readthedocs.io/en/stable/api/zarr/array/#zarr.Array.read_chunk_sizes
https://zarr.readthedocs.io/en/stable/api/zarr/array/#zarr.Array.write_chunk_sizes

These have identical code examples in terms of chunk sizes so I don't fully get it.

changes/4272.bugfix.md
d-v-b: yes, but IMO we should consider removing the feature flag and making it on by default soon

docs/user-guide/arrays.md
d-v-b: inner chunking cannot be rectilinear yet. IMO that's something we need to fix at the spec level. lachy has [a proposal](https://github.com/zarr-developers/zarr-specs/pull/370) to allow semi-regular inner chunks, which isn't hard for us to support, but it's blocked by the challenge it poses for other implementations.

d-v-b:
@maxrjones the issues you raised should be addressed in the latest commits

src/zarr/core/chunk_grids.py
d-v-b: good idea

src/zarr/core/array.py
d-v-b: ill take it

docs/user-guide/arrays.md
d-v-b: docs get updated here

d-v-b:
this is pretty important to get out before our next release, which I would like to get out ASAP, so unless I hear from yall in the next day or so I'm going to self-merge and we can circle back in other PRs.

maxrjones:
> this is pretty important to get out before our next release, which I would like to get out ASAP, so unless I hear from yall in the next day or so I'm going to self-merge and we can circle back in other PRs.

Sounds good. I probably won't have time for another round (trying to take some PTO), and feel confident in your ability to resolve the prior issues.

PR #4219, opened 2026-07-30
fix: reject malformed chunk keys in DefaultChunkKeyEncoding

d-v-b:
# Summary

This PR fixes the `decode_chunk_key` method for the `DefaultChunkKeyEncoding` class. It didn't validate inputs, leading to garbage outputs.

Here's claude's summary:

🤖 AI text below 🤖

decode_chunk_key stripped a single leading character and split the rest, so any key at all decoded to something. "0/1" silently became (1,) -- the "0" was eaten as if it were the "c" prefix -- and a key written with one separator decoded wrongly under an encoding configured with the other.

Validate the "c<separator>" prefix and raise ValueError when it is absent, so a key that is not a chunk key for this encoding is reported rather than silently misread.

Adds the tests this method never had, covering the round trip for both separators and each way a key can fail to carry the prefix.

## For reviewers

## Author attestation

## TODO


d-v-b:
this is a bugfix for a method with no callers today (this bug was discovered while working on `zarr-http-server`) so I feel comfortable self-merging this when tests pass.

Issue #3989, opened 2026-05-20
test cleanup / consolidation


d-v-b:
Our test suite is slow. In addition to solving this by making the library faster, we should remove any redundant tests.

PR #4093, opened 2026-06-22
test: pytest cleanups / modernization

d-v-b:
_I had claude do a review of our test suite. The [tracking issue](https://github.com/d-v-b/zarr-python/issues/195) is on my fork of zarr-python._

Pure test-suite cleanup (no library/behavior changes) adopting current pytest idioms. Part of the test-suite audit.

### Changes
- **Migrate `tmpdir`/`LEGACY_PATH` → `tmp_path`.** Replace the legacy `py.path.local` `tmpdir` fixture and the private `from _pytest.compat import LEGACY_PATH` import with the stdlib `tmp_path: pathlib.Path` fixture across `tests/conftest.py`, `tests/test_group.py`, `tests/test_store/test_core.py`, `tests/test_store/test_local.py`, and `tests/test_api.py`. Also fixes an inverted annotation (`tmp_path: LEGACY_PATH`). The only remaining `tmpdir` references are inside commented-out dead code in `tests/test_api.py`.
- **Remove redundant `@pytest.mark.asyncio` decorators.** With `asyncio_mode = "auto"`, `async def test_*` are collected automatically, so the ~10 explicit markers in `test_buffer.py`, `test_properties.py`, and `test_api.py` are unnecessary. Also drop the now-unused `asyncio` and `s3` entries from the `markers` list in `pyproject.toml` (the `s3` marker was never applied anywhere).
- **Stabilize parametrize ids.** `test_docstring_consistent_parameters` previously used `ids=str`, which stringified function objects to non-deterministic `<function ... at 0x...>` ids. Cases are now wrapped in `pytest.param(..., id=...)` with stable, human-readable ids.
- **Use `monkeypatch.setenv`.** The `docs_s3_backend` fixture in `tests/test_docs.py` now sets `AWS_ENDPOINT_URL` via `monkeypatch.setenv` (auto-restored) instead of a manual `os.environ` save/restore.

### Verification
- `pytest tests/test_store/ tests/test_group.py tests/test_buffer.py tests/test_api/test_synchronous.py tests/test_api.py -q` → **1737 passed, 214 skipped**.
- `pytest --co -q` collects cleanly (**6479 tests**, no errors). ruff, mypy, numpydoc pre-commit pass.

Part of [#195](https://github.com/d-v-b/zarr-python/issues/195)

PR #4224, opened 2026-08-01
test: fix broken `test_docstring_consistent_parameters` and mark `xfail`

Aniketsy:
## Summary

Towards #3989 

- ~Removed `test_docstring_consistent_parameters` from `tests/test_api/test_synchronous.py`. as i find its already covered here~
- also `test_docstrings_match` checks almost the exact same thing (name/type/description)- for docstring coverage

https://github.com/zarr-developers/zarr-python/blob/a994a4fc972fed428eab6a26d4f14bb95d22c144/tests/test_api.py#L1437

https://github.com/zarr-developers/zarr-python/blob/a994a4fc972fed428eab6a26d4f14bb95d22c144/tests/test_group.py#L1792

please feel free to point out if i'm missing something

## For reviewers

## Author attestation

## TODO

### DIFF
--- a/tests/test_api/test_synchronous.py
+++ b/tests/test_api/test_synchronous.py
@@ -1,11 +1,15 @@
 from __future__ import annotations
 
-from typing import Final
+from typing import TYPE_CHECKING, Any, Final
 
 import pytest
 from numpydoc.docscrape import NumpyDocString
 
+import zarr
 from zarr.api import asynchronous, synchronous
+
+if TYPE_CHECKING:
+    from collections.abc import Callable
 
 MATCHED_EXPORT_NAMES: Final[tuple[str, ...]] = tuple(
     sorted(set(synchronous.__all__) | set(asynchronous.__all__))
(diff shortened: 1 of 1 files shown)


d-v-b:

### Code review

Found 1 issue:

1. The redundancy claim doesn't hold: the deleted test checks something none of the cited tests check. `test_docstring_consistent_parameters` compared the *type* and *description* of a shared parameter **across different routines** (`create` vs `create_array` vs `create_group` vs `Group.create_array`):

https://github.com/zarr-developers/zarr-python/blob/a994a4fc972fed428eab6a26d4f14bb95d22c144/tests/test_api/test_synchronous.py#L112-L128

The three tests cited as covering it all compare a **sync/async pair of the same callable** instead:

- `test_docstrings_match` — sync vs async docstring of one callable: https://github.com/zarr-developers/zarr-python/blob/a994a4fc972fed428eab6a26d4f14bb95d22c144/tests/test_api/test_synchronous.py#L21-L27
- `tests/test_api.py:1437` and `tests/test_group.py:1793` are `test_consistent_signatures`, which compare `inspect.signature()`, not docstring prose: https://github.com/zarr-developers/zarr-python/blob/a994a4fc972fed428eab6a26d4f14bb95d22c144/tests/test_api.py#L1436-L1444

`grep -n docstring tests/test_api.py tests/test_group.py` returns nothing — there are no docstring tests in either file, and `NumpyDocString` is used nowhere else in the suite. So the cross-routine coverage is not duplicated; it disappears with this PR.

That said, the test as written was vacuously passing: every `parameter_name` parametrize value is a tuple (e.g. `("store", "path")`), while `param_dict` is keyed by individual parameter-name strings, so `if parameter_name in param_dict` is always `False` and both asserts reduce to `0 <= 1`.

https://github.com/zarr-developers/zarr-python/blob/a994a4fc972fed428eab6a26d4f14bb95d22c144/tests/test_api/test_synchronous.py#L45-L52

So nothing is lost in terms of tests that were actually running, but the intended coverage — added deliberately in #3390 and maintained as recently as #4093 — is real and unimplemented. Repairing the test (iterate over the tuple, one parameter at a time) looks like the better fix than deleting it. If the intent is specifically to speed up the suite per #3989, this test contributes essentially nothing to runtime either way.

Aniketsy:
@d-v-b thanks for the review, I've fixed the broken test but that leads to failures, so  i've marked test as `xfail` for now and opened issue for the failure. please let me know if you have any thoughts related to this.

tests/test_api/test_synchronous.py
```diff
@@ -41,6 +41,23 @@ def test_docstrings_match(callable_name: str) -> None:
...
 
 
+# NOTE: this test used to always pass silently due to a bug — it compared a
+# tuple like ("store", "path") against a dict keyed by single names, so the
+# check never actually ran. Fixed by looping over each name separately.
+# Now that it runs for real, it fails: the docstrings really are inconsistent
+# across create/create_array/create_group/Group.create_array. Marked xfail
+# for now so CI stays green but the issue is tracked, not forgotten.
+# See #4225
+@pytest.mark.xfail(
```
d-v-b, line 51: instead of xfailing, can we make the test work properly, e.g. by fixing it, and fixing any docstring failures the fixed test reveals?

tests/test_api/test_synchronous.py
```diff
@@ -114,15 +132,16 @@ def test_docstring_consistent_parameters(
...
-            val = param_dict[parameter_name]
-            if tuple(val.desc) in descs:
-                descs[tuple(val.desc)] = descs[tuple(val.desc)] + (key,)
-            else:
-                descs[tuple(val.desc)] = (key,)
-            if val.type in types:
-                types[val.type] = types[val.type] + (key,)
-            else:
-                types[val.type] = (key,)
+        for name in parameter_name:
```
d-v-b, line 135: iterating this way aggregates all parameters in the same dict, and then checks if they have the same description / type, which won't be true. we need to break down the aggregation per-parameter.

tests/test_api/test_synchronous.py
Aniketsy: > instead of xfailing, can we make the test work properly, e.g. by fixing it, and fixing any docstring failures the fixed test reveals?

make sense, i'll be back with this shortly, maybe 1-2 days and fix this.

tests/test_api/test_synchronous.py
Aniketsy: thanks for the pointer

PR #4229, opened 2026-08-03
docs: show package tag instead of monorepo release in subpackage docs

maxrjones:
## Summary

This PR patches mkdocs to use the subpackage's tag instead of Zarr-Python core's in the docs header.

| Before | After |
| --- | --- |
| image  | image |

## For reviewers

Opinions on whether we should duplicate the patch on each sub-package (anti-DRY) or symlink (anti-windows).

## Author attestation

## TODO


d-v-b:
this is awesome max! short term i'm leaning anti DRY (pro duplication) but longer term this is something our docs dep should handle without modifications. we should investigate upstreaming this fix

maxrjones:
> this is awesome max! short term i'm leaning anti DRY (pro duplication) but longer term this is something our docs dep should handle without modifications. we should investigate upstreaming this fix

thanks for the prompt feedback. I consolidated both the patch script and the mkdocs configuration files.

docs/contributing.md
```diff
@@ -221,6 +221,26 @@ Hatch can also be used to serve continuously updating version of the documentati
...
+
+Each site is hosted as its own Read the Docs project. To set one up for a new sub-package:
+
+1. Create a new project on [readthedocs.org](https://app.readthedocs.org) importing the `zarr-python` repository, named after the package (e.g. `zarr-metadata`).
+2. In the project's admin settings, set the configuration file path to `packages/<name>/.readthedocs.yaml`. That file also cancels pull request builds that don't touch the package.
+3. Add an automation rule matching the package's release tags (custom match `^zarr_<name>-v`, note the underscore) with the action "Activate version". Automation rules only apply to versions detected after the rule is created, so activate any earlier release tags manually from the versions list.
+4. When activating a version, edit its slug to the bare version number (`0.4.0`, not `v0.4.0`). RTD keeps a version record for every tag in the monorepo, including inactive ones not shown in the dashboard's versions list, and all of zarr-python's own release tags are `v`-prefixed — so a `v`-prefixed slug collides with zarr-python's release history ("A version with that slug already exists"), while bare version numbers cannot collide.
```
maxrjones, line 242: @d-v-b supposedly (according to Fable) this will address the issues discussed in [#Zarr-Python > zarr-metadata docs](https://ossci.zulipchat.com/#narrow/channel/423692-Zarr-Python/topic/zarr-metadata.20docs/with/613597110). You're welcome to add me to RTD if you'd like any help with maintainence.

packages/zarr-indexing/uv.lock
```diff
@@ -0,0 +1,698 @@
+version = 1
```
maxrjones, line 1: I committed uv.lock since the lockfile is committed for zarr-metadata

docs/contributing.md
d-v-b: I did add that automation rule, but didn't get the expected results! Happy adding you to rtd so you can tinker

packages/source-version.js
```diff
@@ -0,0 +1,73 @@
...
+  const newestFirst = (a, b) => {
+    const va = parse(a), vb = parse(b)
+    for (let i = 0; i < Math.max(va.length, vb.length); i++)
+      if ((vb[i] || 0) !== (va[i] || 0))
+        return (vb[i] || 0) - (va[i] || 0)
+    return 0
+  }
+
+  async function latestPackageTag() {
+    const response = await fetch(
```
d-v-b, line 38: this fires on every page load, and I think github has a rate limit on this endpoint, maybe we want to cache this internally?

packages/source-version.js
```diff
@@ -0,0 +1,73 @@
...
+    return versions[0]
+  }
+
+  function patchDom(version) {
+    for (const el of document.querySelectorAll(".md-source__fact--version"))
+      el.textContent = version
+  }
+
+  function patchCache(version) {
+    const facts = __md_get("__source", sessionStorage)
```
d-v-b, line 60: we probably need to watch this code carefully when / if we update material

maxrjones: we probably won't be updating material 🫠 https://github.com/squidfunk/mkdocs-material/issues/8523

packages/source-version.js
maxrjones: good catch, thanks! fixed in https://github.com/zarr-developers/zarr-python/pull/4229/commits/cb1f8b3dd5187871848a6c7d08cb1b03a05f71de

d-v-b:
this needs some minor conflict resolution, and a few other changes, see https://github.com/maxrjones/zarr-python/compare/docs/subpackage-header-version...d-v-b:zarr-python:docs/subpackage-header-version-fixes

lmk if it's OK to push to this branch?

maxrjones:
> this needs some minor conflict resolution, and a few other changes, see [maxrjones/zarr-python@docs/subpackage-header-version...d-v-b:zarr-python:docs/subpackage-header-version-fixes](https://github.com/maxrjones/zarr-python/compare/docs/subpackage-header-version...d-v-b:zarr-python:docs/subpackage-header-version-fixes)
> 
> lmk if it's OK to push to this branch?

yes, totally fine

Issue #4237, opened 2026-08-05
"ValueError: Input array must be contiguous" when *cast_value* codec next to *transpose* codec


raphaeljolivet:
### Zarr version

v3.3.1.dev3+g976be695

### Numcodecs version

v0.16.5

### Python Version

3.12.11

### Operating System

Debian

### Installation

uv sync

### Description

When combined with the **transpose** codec **cast_value** fails because the array is not continugous.

According to [the specification](https://zarr-specs.readthedocs.io/en/latest/v3/codecs/transpose/),
 the transpose codec is expected to produce "views".

Thus, it should be the responsibility of **every** codec to ensure it makes it contiguous, if needed.

If my interpretation is correct, I will prepare a PR to fix this, with proper unit test.
I guess we should test the behavior on *every* other codec ... 

### Steps to reproduce

```python

def test_enforce_contiguous_arrays() :
    """
    Transpose codec produces non continguous arrays.
    Ensure cast_value make them contiguous before processing.
    """
    data = np.arange(20, dtype=np.float32).reshape(5, 2, 2)

    def make_array(filters):
        return zarr.create_array(
            store=MemoryStore(),
            shape=data.shape,
            dtype=data.dtype,
            chunks=data.shape,
            filters=filters,
            serializer=zarr.codecs.BytesCodec(endian="little"),
            compressors=None,
            zarr_format=3,
        )

    # Cast before tranpose : may fail on read.
    array = make_array([
        zarr.codecs.CastValue(data_type="uint16"),
        zarr.codecs.TransposeCodec(order=(1, 2, 0)),
    ])

    # Write should be ok
    array[:] = data

    # Read may fail
    assert_array_equal(array[:], data)

    # Cast after tranpose : may fail on write
    array = make_array([
        zarr.codecs.TransposeCodec(order=(1, 2, 0)),
        zarr.codecs.CastValue(data_type="uint16"),
    ])

    # Write may fail
    array[:] = data

```
    # Read should be ok
    assert_array_equal(array[:], data)

d-v-b:
thanks for the report, this looks like a bug in cast_value. it should be changed to handle non-contiguous arrays.

raphaeljolivet:
I just did [a PR](https://github.com/zarr-developers/zarr-python/compare/main...raphaeljolivet:zarr-python:bugfix/4237-cast-value-transpose-input-aray-should-be-contiguous)

The "lint" CI fails while prek works fine locally.
I am not sure why.

PR #4238, opened 2026-08-05
Fix of #4237 : 'ValueError: Input array must be contiguous' when *cas…

raphaeljolivet:
Fix of #4237 with unit test.


src/zarr/codecs/cast_value.py
```diff
@@ -359,7 +359,9 @@ def _encode_sync(
         target_native = self.dtype.to_native_dtype()
 
         result = self._do_cast(
```
d-v-b, line 363: can you move the fix into `_do_cast`?

tests/test_codecs/test_cast_value.py
```diff
@@ -4,10 +4,13 @@
 
 import numpy as np
 import pytest
+from numpy.ma.testutils import assert_array_equal
```
d-v-b:
```suggestion
from numpy.testing import assert_array_equal
```

tests/test_codecs/test_cast_value.py
```diff
@@ -477,3 +480,52 @@ def test_parse_scalar_map(case: Expect[Any, Any]) -> None:
...
+
+    # Cast before transpose: may fail on read.
+    array = make_array(
+        [
+            CastValue(data_type="uint16"),
+            TransposeCodec(order=(1, 2, 0)),
+        ]
+    )
+
+    # Write should be ok
```
d-v-b: not sure we need these comments -- the entire test should pass, right?

d-v-b:
thank you for this! can you add a release note?

src/zarr/codecs/cast_value.py
```diff
@@ -317,7 +317,7 @@ def _do_cast(
             to_tgt = int if np.issubdtype(target_dtype, np.integer) else float
             scalar_map_entries = {to_src(k): to_tgt(v) for k, v in scalar_map.items()}
         return cast_array_rs(  # type: ignore[no-any-return]
```
d-v-b, line 321:
```suggestion
        # NOTE: the use of np.ascontiguousarray below makes a copy when 
        # the input is not contiguous.
        return cast_array_rs(  # type: ignore[no-any-return]
```

tests/test_codecs/test_cast_value.py
```diff
@@ -477,3 +480,45 @@ def test_parse_scalar_map(case: Expect[Any, Any]) -> None:
...
     assert parse_scalar_map(case.input) == case.output
+
+
+@requires_cast_value_rs
+def test_enforce_contiguous_arrays() -> None:
+    """
+    Transpose codec produces non-contiguous arrays.
+    Ensure cast_value makes them contiguous before processing.
+    """
+    data = np.arange(20, dtype=np.float32).reshape(5, 2, 2)
```
d-v-b, line 491: can we parametrize this test over the shape of the array? Specifically, we need to test 0-dimensional arrays, because I _think_ the current fix will not work for 0-dim arrays, and we need to do `np.asarray(order="C")` instead of `np.ascontiguousarray`.

d-v-b:
@raphaeljolivet can you check the failing example against the branch in this PR: https://github.com/zarr-developers/zarr-python/pull/4260? If that fixes the bug, then maybe this PR can simplify to just adding tests?

d-v-b:
sorry for the churn @raphaeljolivet but I think some of these changes are superseded by #4260, please have a look

d-v-b:
this was superseded by https://github.com/zarr-developers/zarr-python/pull/4260, thanks for the work in this PR

PR #4252, opened 2026-08-12
fix(docs): build subpackage docs from the package directory

d-v-b:
My description: this unbreaks docs builds for subpackages by ensuring that RTD builds in the subpackage directory. Claude wrote the fix.

AI description:

Read the Docs pointed `-f` at each package's mkdocs.yml from the repo root. mkdocs resolves some settings relative to the current working directory rather than to the config file, so building from elsewhere looks for them in the wrong place -- and silently, because the paths are valid, just wrong.

zarr-indexing hit this at v0.2.0: `pymdownx.snippets` has a relative `base_path` of `[docs, examples]`, so `--8<-- "snippets/canonical_slice.py"` resolved against the repo root and searched zarr-python's own docs/ rather than the package's. The build failed with SnippetMissingError while `just docs-check` passed, because that runs from the package directory.

Building from the package directory makes the Read the Docs invocation identical to the local and CI ones, so a green build there means a green build here. $READTHEDOCS_OUTPUT is absolute, so the cd does not affect where the site lands.

Applied to all three packages. Only zarr-indexing is failing today; zarr-metadata and zarr-http-server do not use snippets, so for them this is preventive -- the hazard is any config resolved against the working directory, and it would show up only on Read the Docs.

Assisted-by: ClaudeCode:claude-opus-5

## For reviewers

## Author attestation

## TODO

### DIFF
--- a/packages/zarr-http-server/.readthedocs.yaml
+++ b/packages/zarr-http-server/.readthedocs.yaml
@@ -24,7 +24,20 @@ build:
       - pip install ./packages/zarr-http-server --group packages/zarr-http-server/pyproject.toml:docs
     build:
       html:
-        - mkdocs build --strict -f packages/zarr-http-server/mkdocs.yml --site-dir $READTHEDOCS_OUTPUT/html
+        # Build from inside the package rather than pointing `-f` at its config
+        # from the repo root. mkdocs resolves some settings relative to the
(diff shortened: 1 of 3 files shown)


d-v-b:
I'm self-merging this as soon as it's green because we need this fix for `zarr-indexing` docs builds.

PR #4260, opened 2026-08-13
deps: bump cast-value.rs to >= 0.4.2

### LINKED ISSUE #4237: "ValueError: Input array must be contiguous" when *cast_value* codec next to *transpose* codec

d-v-b:
## Summary

closes #4237 by updating cast-value.rs. The underlying issue was fixed in https://github.com/zarr-developers/cast-value.rs/pull/28.

I did this by hand.

## For reviewers

I'm going to self-merge this when green because it's fixing a data corruption issue.

## Author attestation

## TODO


d-v-b:
this PR brings in work by @raphaeljolivet on the testing side, augments those tests, and adds a guard to ensure that folks don't use the problematic version of `cast-value.rs` with zarr-python

raphaeljolivet:
Thanks !

Sorry I did not follow up on my previous PR. 
I was switching to other tasks.

PR #4261, opened 2026-08-13
fix(build): use an sdist allowlist so the zarr sdist stops shipping subpackages

d-v-b:
## Summary

This is an AI-written PR that changes our sdist contents from blocklist-defined to allowlist-defined. That keeps new stuff out by default, which is easier when we are adding new subpackages. Without changing _something_ we are bundling the contents of `packages/` in the sdist, which is not correct. If we add each subpackage to the blocklist, we have an unsafe default. Better to have the safe default, where new subpackages don't need to be explicitly excluded. 

This change does raise the possibility that we add a new file or path that is required for sdist consumers like conda, but we forget to add it to the allowlist. We should consider addressing this in CI with some kind of "conda can consume this" check, and/or normalize release candidate releases (not sure if conda forge picks those up).

This was written by claude, and the original PR is here: https://github.com/d-v-b/zarr-python/pull/289

## Author attestation

## TODO


d-v-b:
> This change does raise the possibility that we add a new file or path that is required for sdist consumers like conda, but we forget to add it to the allowlist. We should consider addressing this in CI with some kind of "conda can consume this" check, and/or normalize release candidate releases (not sure if conda forge picks those up).

I will try to address that concern in this PR with some new CI jobs.

d-v-b:
I'm self-merging this, because IMO it's a release blocker and we can always revisit the particulars later.

PR #4264, opened 2026-08-14
fix(zarr-metadata): make JSONValue's array arm covariant

d-v-b:
## Summary

`list[T] | tuple[T]` is invariant but `Sequence[T]` is covariant. In the context of our `JSONValue` type, we want covariance, to support intuitive propagation of the "isintance" operator when narrowing `JSONValue`. That means we need to change our `JSONValue` union. This is technically a breaking change and so the next release of zarr-metadata will indicate this. Why is it breaking? because strings inhabit `Sequence[str]`, but not `list[JSONValue] | tuple[JSONValue]`. Routines handling `JSONValue`-shaped inputs will need to guard for strings where previously they did not. Low blast area but a teeny blast nonetheless.

cc @kylebarron i doubt this affects zarrista but I will do a check in any case.

Written by claude, with an original PR here: https://github.com/d-v-b/zarr-python/pull/295

## Author attestation

## TODO


d-v-b:
SMWG (self-merge when green)

kylebarron:
I've also hit similar things before, and I think you do want this to be covariant.

PR #4265, opened 2026-08-14
fix: accept universal-pathlib UPath as a StoreLike value

### LINKED ISSUE #4244: Universal-pathlib support for zarr.open
### Describe the new feature you'd like

In the past I'd use code as follows:

```sh
import xarray as xr
from upath import UPath

s3_path = UPath("s3://my-bucket/my-key.zarr/")
ds = xr.open_dataarray(s3_path, engine="zarr", chunks={})
```

This recently stopped working (I think because upath.S3Path no longer implements `__fspath__`, not entirely sure though).

Right now xarray just passes the upath directly to `zarr.open()`. Personally I think it would make sense that the `zarr.open` supports a UPath as a parameter, seeing that universal-pathlib is already an optional dependency.

d-v-b:
## Summary

closes #4244 

Get `UPath` working again as a `StoreLike` value. This broke because we only handled `UPath` instances implicitly via isinstance checks against a parent class, and a recent release of `UPath` changed the inheritance structure of those objects.

claude wrote this, source PR is here: https://github.com/d-v-b/zarr-python/pull/280

## Author attestation

## TODO


d-v-b:
@mark-boer could you have a look

src/zarr/storage/_utils.py
```diff
@@ -6,7 +6,9 @@
 from urllib.parse import urlparse
 
 if importlib.util.find_spec("upath"):
-    from upath.core import UPath
+    # Re-exported for zarr.storage._common, which needs it to recognize UPath store_like values.
+    # The redundant-looking alias is the explicit re-export mypy requires under strict mode.
+    from upath.core import UPath as UPath  # noqa: PLC0414
```
mark-boer, line 11: [Useless import alias](https://docs.astral.sh/ruff/rules/useless-import-alias/)

```suggestion
    from upath.core import UPath
```

src/zarr/storage/_fsspec.py
```diff
@@ -171,8 +171,12 @@ def from_upath(
         -------
         FsspecStore
         """
+        # A UPath hands back a filesystem in whatever mode it was constructed with, which is
+        # synchronous unless the caller passed asynchronous=True. Route it through _make_async so
+        # that sync-mode instances of async filesystems are re-created in async mode, and
+        # genuinely synchronous filesystems are wrapped.
         return cls(
-            fs=upath.fs,
+            fs=_make_async(upath.fs),
```
mark-boer, line 179: Is this an unrelated issue / optimization?

src/zarr/storage/_common.py
```diff
@@ -381,6 +382,17 @@ async def make_store(
...
+        # This must be checked before Path: in universal-pathlib < 0.3 every UPath, including
+        # remote ones like S3Path, subclasses pathlib.Path, and would otherwise be misrouted to a
+        # LocalStore. Local UPaths get a LocalStore so that UPath("/data") and Path("/data") agree,
+        # mirroring how the equivalent strings are routed below.
+        if store_like.protocol in ("", "file"):
+            return await make_store(
+                Path(store_like.path), mode=mode, storage_options=storage_options
+            )
+        return FsspecStore.from_upath(store_like, read_only=_read_only)
+
```
mark-boer, line 395: It seems that upaths `LocalPath` still inherits from `pathlib.Path`, so you could if you want move the `elif` back behind the `elif isinstance(store_like, Path)` and the `store_like.protocol in ("", "file")` is not longer neccesary.

Not sure if it makes it more readable though ;-)

mark-boer (commented):
Looks good! Made a couple of remarks / questions.

src/zarr/storage/_common.py
mark-boer: Seems there exists a LocalPath as well:

https://github.com/fsspec/universal_pathlib#local-paths-and-url-paths

> All local UPath types are os.PathLike, but only the PosixUPath and WindowsUPath are subclasses of pathlib.Path.

src/zarr/storage/_fsspec.py
d-v-b: unfortunately we need this logic any time we create an instance of `FsspecStore` from an existing fsspec filesystem.

src/zarr/storage/_utils.py
d-v-b: the comment above explains why we use the alias. it's to satisfy mypy.

src/zarr/storage/_common.py
d-v-b: i think the current sequencing is OK for now

d-v-b:
self-merging since this is a safe fix.

PR #4266, opened 2026-08-14
chore(zarr-metadata): build 0.5.0 changelog

d-v-b:

Release PR for `zarr-metadata` 0.5.0. Runs `towncrier build --version 0.5.0` in `packages/zarr-metadata/`, consuming the pending news fragments into `CHANGELOG.md`.

**Why minor, not patch:**

- #4264 widened `JSONValue`'s array arm from the invariant `list["JSONValue"] | tuple["JSONValue", ...]` to the covariant `Sequence["JSONValue"]`. That changes the meaning of a published type for every consumer that annotates against it — narrower element types (`list[str]`, `Sequence[float]`) now count as JSON values, and TypedDicts with precisely-typed fields become assignable to `Mapping[str, JSONValue]`. Filed as a bugfix fragment, but it is an interface change.
- #4232 removed the version-last SCREAMING_SNAKE constant spellings (`ARRAY_METADATA_STORE_KEY_V2` → `ZARR_V2_ARRAY_METADATA_STORE_KEY`, and eight siblings). The old names are removed, not aliased.

**Two commits:**

1. `note the sdist allowlist, and let misc entries speak` — #4248 (the sdist allowlist for this package) merged without a news fragment, so a user-visible packaging change was about to miss the notes; this adds one. Filing it as `misc` exposed that towncrier's built-in `misc` type sets `showcontent = false`, so the entry would have rendered as a bare PR link. This package's towncrier config now restates all five types from the `changes/README.md` menu — declaring any type replaces the built-in set — as the defaults verbatim except `misc`, which shows its content.
2. `build 0.5.0 changelog` — the towncrier run itself.

No source changes: the version comes from the `zarr_metadata-v*` tag via hatch-vcs, so tagging `zarr_metadata-v0.5.0` after merge is what cuts the release.

**Verification:** `just test` in `packages/zarr-metadata/` — 595 passed.


d-v-b:
sorry for the raw bot content but for chores like this I think there's a very small audience right now

d-v-b:
merging cause I need this fixed JSONValue type downstream

PR #4270, opened 2026-08-18
fix: skip exact byte comparison for gzip codec due to timestamp variance

### LINKED ISSUE #4254: avoid comparing exact output of gzip encoding in codec pipeline tests
https://github.com/zarr-developers/zarr-python/actions/runs/31607404847/job/94150195454?pr=4253 shows a test failure caused by two gzip-encoded bytestreams differ. from what an llm told me, this is due to gzip including timestamp info in compressed payloads, which makes two gzip streams differ even if they decode to the exact same bytes. so tests for exact byte consistency should be avoided here in favor of round-trip fidelity.

sheikhayaan:
Fixes #4254

The test_async_chunk_transform_matches_sync[bb] test was comparing 
raw gzip-encoded bytes between async and sync paths. Since gzip 
embeds a wall-clock timestamp in compressed output, two compressions 
of the same data at different times produce different bytes even 
though the data is identical.

The fix skips the raw byte comparison when a GzipCodec is present 
in the codec chain. Round-trip fidelity is still verified by the 
decode assertions that follow.


d-v-b:
Thanks for this fix. I think skipping the comparison entirely isn't as good as comparing everything except the timestamp. You could add a new utility routine to the gzip module for taking a gzip encoded stream and returning a timestamp, remainder pair. We would then compare the remainder. That's my first idea but I'm sure there's an even better one.

d-v-b:
a marginally better idea: add a private boolean function to the gzip module that compares two gzip streams for "everything but timestamp" equality. There's probably a name for this kind of comparison (comparing two outputs only on the basis of the parameters under control of the producers). internally, we can just compare bytestream a and bytestream b on everything except the four bytes starting at byte 4, which is the mtime field (if I am reading the gzip spec correctly).

sheikhayaan:
Thanks for the suggestion. I updated the test to compare the gzip streams while ignoring only the 4-byte MTIME field, and added a focused test for the comparison.

d-v-b (approved):

d-v-b:
thanks!

PR #4279, opened 2026-08-21
fix: reject a string-valued zero scale in the scale_offset codec

dylanpulver:
## Summary

`ScaleOffset.validate` compared `self.scale` against `0` before parsing it. `scale` is documented and typed as `int | float | str`, and no Python string equals `0`, so `"0"` and `"0.0"` walked past the guard, as did the spec's hex form `"0x0000000000000000"`.

On float dtypes the array is created, every chunk is written as zero and read back as NaN with no error. The zero scale is persisted into `zarr.json`, so reopening the store reproduces it. On integer dtypes it surfaces as an unhandled `ZeroDivisionError` instead of the `ValueError` the guard promises.

The comparison now runs on the value returned by `from_json_scalar`. That reorders one thing: given a bad `offset` and a zero `scale` together, the representability error fires first. Nothing asserted the old order.

Narrowing the annotation to drop `str` would also close it, but a string `scale` is a documented input and the suite already exercises `offset="NaN"`.

Residual: `scale="NaN"` and `scale=float("inf")` are still accepted on float dtypes and also produce unusable reads. That reaches purely numeric input as well, so it reads as a separate question.

## For reviewers

The reorder is the only behavior change beyond the new rejection. `test_rejects_zero_scale` is now parametrized; the three string cases fail on `main` and pass here, and the two numeric cases pass either way.

Full suite: 6908 passed, 461 skipped, 4 xfailed. ruff 0.16.0 check and format clean. mypy 2.3.0 clean.

## Author attestation

## TODO


d-v-b:
looks good, thanks for the fix and sorry about the bug

d-v-b (approved):

PR #4283, opened 2026-08-25
docs: add spec-style description of the consolidated metadata format

d-v-b:

Adds a specification-style description of Zarr-Python's consolidated metadata encoding and reader behavior, linked from the usage guide and navigation. It covers V2 `.zmetadata`, V3 inline metadata, path and child-marker conventions, ordering, and `use_consolidated` behavior.

This documents implementation behavior and interoperability guidance; it is not an adopted Zarr specification. At the factual review, [zarr-specs#309](https://github.com/zarr-developers/zarr-specs/pull/309) remains open and unmerged. Excerpts are attributed to Tom Augspurger and the repository's CC-BY-4.0 license.

The sort-order correction is depth first, then Unicode NFKC normalization and case folding. For equal-depth keys, `a` precedes `B`, unlike raw Unicode code-point ordering. The previous example `B` before `x` demonstrated no difference. Equal normalized keys retain input order, so this does not establish universally byte-identical output.

The page distinguishes current reader acceptance from writer recommendations, including mixed-format entries, ignored `must_understand`, and V2 child-group markers. It also documents that deletion can resave consolidated metadata while other mutations need not keep snapshots synchronized. No runtime behavior changes are included.

The factual follow-up clarifies the document's status; applicable commit hooks pass. The earlier strict documentation build is historical validation.


d-v-b:
@TomAugspurger you should probably also have a look, since this is an LLM summarizing your work

docs/user-guide/consolidated_metadata_format.md
```diff
@@ -0,0 +1,517 @@
...
+!!! warning "Status"
+    Consolidated metadata is **not part of the Zarr format 3 core
+    specification**. The format described here for Zarr format 3 follows the
+    proposal in [zarr-specs#309](https://github.com/zarr-developers/zarr-specs/pull/309)
+    by Tom Augspurger, which is still open at the time of writing. Until that
+    (or a successor) proposal is accepted, Zarr-Python's behaviour is the
+    de facto reference, and it may change to track the specification.
+
+    Consolidated metadata for Zarr format 2 follows the format established by
+    Zarr-Python 2.x, with one deviation noted [below](#zarr-format-2).
```
TomAugspurger: Pet-peeve: I don't like writing non-evergreen documentation. Maybe it's OK for this type of doc, but I'd prefer to remove it.

```suggestion
```

docs/user-guide/consolidated_metadata_format.md
```diff
@@ -0,0 +1,517 @@
...
+> number of nodes. This can be costly, especially for large hierarchies served
+> HTTP from a remote storage system (like Blob Storage).
+
+## Concepts
+
+A hierarchy is **consolidated at** a group (the *consolidating group*). The
+consolidating group's own metadata document gains a copy of the metadata of
+every node below it, at every depth. Nothing about the child nodes' own
+metadata documents changes: they remain in the store and remain authoritative
+for writers.
```
TomAugspurger:
```suggestion
metadata documents changes.
```

TomAugspurger:
> I'm pretty busy these days with a newborn baby so I can't give this careful review.

Congrats! I'll take a look when I get a chance.

Do you have any more info on the motivation for this document. I gather it's primarily for other implementations looking to interoperate with what we write, and am trying to balance this approach vs. telling them to read the spec + source code :)

normanrz:
I asked for a spec on Zulip and Davis was kind enough to create one. The motivation is that we picked up the consolidated metadata in the newly-formed Zarr Format Working Group and I wanted to get a better understanding of the current zarr-python implementation. Having a spec here would be great, but even better would be if we could eventually bring this document into the Zarr 3 spec.

d-v-b:
> Do you have any more info on the motivation for this document. I gather it's primarily for other implementations looking to interoperate with what we write, and am trying to balance this approach vs. telling them to read the spec + source code :)

IMO we don't really have a place for a consolidated metadata spec in the `zarr-specs` repo until we unblock the blockage that prevented your original PR from getting merged in some form or another. For folks curious about the normative structure of consolidated metadata, but also zarr-python's particular implementation choices, I think some kind of document in our docs is a good play. Long term our goal should be to replace an actual spec with a link to a spec defined elsewhere.

normanrz:
> IMO we don't really have a place for a consolidated metadata spec in the `zarr-specs` repo until we unblock the blockage that prevented your original PR from getting merged in some form or another.

Right. With @joshmoore, I am working on a proposal for the ZFWG governance and I could imagine using consolidated metadata as testcase for the governance process.

docs/user-guide/consolidated_metadata.md
```diff
@@ -11,6 +11,10 @@ entire hierarchy, especially when the metadata is being served over a network.
 Consolidated metadata essentially stores all the metadata for a hierarchy in the
 metadata of the root Group.
 
+This page describes how to use consolidated metadata from Python. For a precise
```
TomAugspurger, line 14: I have some hesitancy about linking to this document from user-facing docs. I think the spec (or the PR for the spec) should be sufficient for users, and if it isn't then the docs should be improved there.

docs/user-guide/consolidated_metadata.md
```diff
@@ -84,9 +88,9 @@ print(output.getvalue())
 !!! info "Added in version 3.1.1"
     The keys in the consolidated metadata are sorted prior to writing. Keys are
     sorted in ascending order by path depth, where a path is defined as a sequence
-    of strings joined by `"/"`. For keys with the same path length, lexicographic
-    order is used to break the tie. This behavior ensures deterministic metadata
-    output for a given group.
+    of strings joined by `"/"`. For keys with the same depth, the tie is broken by
+    comparing the paths after Unicode NFKC normalization and case-folding. This
+    behavior ensures deterministic metadata output for a given group.
```
TomAugspurger, line 93: This seems worse for causal readers.

Let's at least keep "lexicographic" (since it isn't incorrect, right?) And if we want to be more specific we can, but let's link to the Python docs on normalization.

docs/user-guide/consolidated_metadata_format.md
```diff
@@ -0,0 +1,517 @@
...
+> This PR adds a new optional field to the core metadata for consolidating all
+> the metadata of all child nodes under a single object. The motivation is
+> similar to consolidated metadata in Zarr V2: without consolidated metadata,
+> the time to load metadata for an entire hierarchy scales linearly with the
+> number of nodes. This can be costly, especially for large hierarchies served
+> HTTP from a remote storage system (like Blob Storage).
+
+## Concepts
+
+A hierarchy is **consolidated at** a group (the *consolidating group*). The
```
TomAugspurger, line 40: "consolidating group" isn't a term in https://github.com/zarr-developers/zarr-specs/pull/309/. If we want to make it one, let's propose it there.

docs/user-guide/consolidated_metadata_format.md
```diff
@@ -0,0 +1,517 @@
...
+    full path of a node relative to the consolidating group. This is the
+    persisted (on-disk) form for both Zarr formats.
+
+Nested
+:   Each group holds only its *immediate* children, and child groups hold
+    their own children recursively. This is Zarr-Python's in-memory form
+    (`GroupMetadata.consolidated_metadata.metadata`). It is never written to a
+    store and is mentioned here only because it leaks into the on-disk form in
+    one place (the [empty child marker](#child-groups-carry-an-empty-marker)).
+
```
TomAugspurger, line 58: I don't think "leaks into" is accurate here, or it conveys the wrong impression. That's deliberate, saying that there aren't any children.

docs/user-guide/consolidated_metadata_format.md
```diff
@@ -0,0 +1,517 @@
...
+Converting between the two is lossless. `ConsolidatedMetadata.flattened_metadata`
+goes nested → flat; `ConsolidatedMetadata._flat_to_nested` goes flat → nested.
+
+## Paths
+
+A **path** is the name of a node relative to the consolidating group:
```
TomAugspurger, line 64: IIRC paths are already defined in the spec. Link to that.

docs/user-guide/consolidated_metadata_format.md
```diff
@@ -0,0 +1,517 @@
...
+  B/
+    y
+    C/
+```
+
+consolidating at `A` produces the paths `B`, `B/C`, `B/y`, `x`; consolidating
+at `B` produces `C`, `y`; consolidating at `C` produces no paths (an empty
+mapping, which is still written).
+
+!!! note "Difference from the zarr-specs#309 text"
```
TomAugspurger, line 86: Ha, I had a review pending from April 2025 pointing out this issue. Submitted that: https://github.com/zarr-developers/zarr-specs/pull/309#discussion_r2044434546 and we can fix it in the spec.

docs/user-guide/consolidated_metadata_format.md
```diff
@@ -0,0 +1,517 @@
...
+| `kind`            | MUST write `"inline"`.                  | MUST reject any value other than `"inline"` (`ValueError`). |
+| `must_understand` | MUST write `false`.                     | Not checked. |
+| `metadata`        | MUST write a JSON object (possibly empty). | MUST reject a non-object (`TypeError`). Each value MUST be a JSON object (`TypeError` otherwise). |
+
+### Values of `metadata`
+
+Each value is a complete node metadata document, i.e. exactly what would be
+found in that node's own `zarr.json`, with the following rules:
+
+* The document MUST contain `zarr_format`. Readers discriminate on this first;
```
TomAugspurger, line 155: I'm guessing this isn't intended / well-tested.

docs/user-guide/consolidated_metadata_format.md
```diff
@@ -0,0 +1,517 @@
...
+Every group entry in `metadata` MUST itself carry a `consolidated_metadata`
+key whose value is the empty marker:
+
+```json
+{"kind": "inline", "must_understand": false, "metadata": {}}
+```
+
+This is the one place the nested in-memory form shows through. The marker
+does **not** mean the child group has no children: the child's descendants are
+still listed in the flat mapping at the consolidating group. It exists so
```
TomAugspurger, line 179: This is wrong (or I'm misunderstanding something). But AFAIK the presence of `metadata: {}` definitely does indicate a group with no children.

TomAugspurger (commented):
Thanks for the context. With that in mind, I'll view this as a somewhat temporary document. Though just merging the zarr-specs PR feels infinitely better.

And for those interested in the implementation, rereading the discussion in https://github.com/zarr-developers/zarr-python/pull/2113 might also be fruitful.

docs/user-guide/consolidated_metadata_format.md
d-v-b: agreed!

PR #4288, opened 2026-08-26
fix: keep fill value and attributes in from_array

### LINKED ISSUE #4287: `from_array` silently discards the source array's `fill_value` and `attributes`
### Zarr version

3.3.1.dev34+gd44f9f92

### Numcodecs version

0.16.5

### Python Version

3.12

### Operating System

Linux

### Installation

uv run with zarr installed from git main

### Description

`zarr.from_array` documents both `fill_value` and `attributes` as defaulting to the
source array's values, but it discards both, using the dtype's default scalar and `{}`
instead.

### Expected vs. actual

Copying an array created with `fill_value=42` and `attributes={"units": "K"}`:

|              | source            | `from_array` copy | expected          |
| ------------ | ----------------- | ----------------- | ----------------- |
| `fill_value` | `42`              | `0`               | `42`              |
| `attributes` | `{"units": "K"}`  | `{}`              | `{"units": "K"}`  |

Every other keep-parameter already defaults to the source
array. These two are the exception.

### Affected versions

The `fill_value` defect is a **regression** from #3198 (released in **3.1.0**), which
introduced the `DEFAULT_FILL_VALUE` sentinel.  The line is unchanged since, so
**3.1.0 through `main` are affected**.

The `attributes` defect is *not* a regression — it has been present since `from_array`
was added.

### Steps to reproduce

```python
# /// script
# requires-python = ">=3.12"
# dependencies = [
# ]
# ///
#

import zarr

src = zarr.create_array({}, shape=(4,), dtype="int32", fill_value=42, attributes={"units": "K"})

copy = zarr.from_array({}, data=src)
print(f"fill_value: {copy.fill_value}      expected 42")
print(f"attrs:      {dict(copy.attrs)}     expected {{'units': 'K'}}")

# The fill value is not just metadata: it decides what unwritten regions read.
meta_only = zarr.from_array({}, data=src, write_data=False)
print(f"write_data=False reads: {meta_only[:]}   expected [42 42 42 42]")

# An explicit None *does* keep the source's fill value -- the exact inverse.
explicit_none = zarr.from_array({}, data=src, fill_value=None)
print(f"fill_value=None:  {explicit_none.fill_value}      keeps 42, inverted")

assert copy.fill_value == 42, "BUG: fill_value not kept"
```

### Additional output

```
fill_value: 0      expected 42
attrs:      {}     expected {'units': 'K'}
write_data=False reads: [0 0 0 0]   expected [42 42 42 42]
fill_value=None:  42      keeps 42, inverted
AssertionError: BUG: fill_value not kept
```

paraseba:
## Summary

`from_array` discards both `fill_value` and `attributes`.

An explicit `fill_value=None` now means the dtype's default scalar (v3) or a null fill value (v2), consistent with `create_array`, and an empty `attributes` dict opts out of copying the source's attributes.

Fixes #4287

## Author attestation

### DIFF
--- a/changes/4288.bugfix.md
+++ b/changes/4288.bugfix.md
@@ -1,0 +1,3 @@
+`zarr.from_array` now defaults to the fill value and the attributes of the source array. Previously both were silently discarded: the array was created with the data type's default scalar and no attributes.
+
+An explicit `fill_value=None` now selects the data type's default scalar (Zarr format 3) or a null fill value (Zarr format 2), consistently with `create_array`, and an empty `attributes` dict creates the array with no attributes.
--- a/src/zarr/api/synchronous.py
+++ b/src/zarr/api/synchronous.py
@@ -1105,6 +1105,8 @@ def from_array(
     fill_value : Any, optional
         Fill value for the array.
         If not specified, defaults to the fill value of the data array.
+        Pass `None` explicitly to use the default scalar of the data type
+        (Zarr format 3) or a null fill value (Zarr format 2) instead.
     order : {"C", "F"}, optional
         The memory order of the array (default is "C").
         For Zarr format 2, this parameter sets the memory order of the array.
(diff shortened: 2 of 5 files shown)


d-v-b:
thanks @paraseba, this looks good. 

long term we will need to make breaking changes around our use of `None` as a sentinel, because some data types (like a JSON data type) would use `None` as a scalar value. But that's a future problem.

d-v-b (approved):

PR #4290, opened 2026-08-26
fix: honor rectilinear chunk-grid requests with uniform edges

### LINKED ISSUE #4272: A rectilinear chunk spec with a short trailing chunk is silently normalized to a regular grid, changing resize semantics
### Zarr version

v3.3.0

### Numcodecs version

v0.16.5

### Python Version

3.12.13

### Operating System

Mac

### Installation

Using Pixi into virtual environment

### Description

3.3.0 picks between RegularChunkGridMetadata and RectilinearChunkGridMetadata from the values of the chunk edges rather than the form of the request. Since is_regular_1d counts "all chunks equal, last one smaller" as regular, an explicitly rectilinear spec whose edges happen to be uniform-plus-a-short-tail is stored as a regular grid.

For the array as created the two are equivalent. They diverge under resize: a FixedDimension grows by extending the uniform pattern, a VaryingDimension by appending an edge. The caller gets a grid that behaves differently from the one they asked for, with no warning and nothing in the metadata recording the substitution.

### Why this matters
The use case is an append-only time series written in two regimes: a coarse backfill that lays down large chunks, then an incremental forward-fill that appends small ones. Rectilinear chunks make each appended window its own chunk, so the forward-fill never rewrites an edge chunk — that invariant is the entire reason for using them.

The collapse makes the invariant contingent on whether the initial backfill plan happens to look uniform. Edges of (168,) * 13 + (24,) become regular and every later append straddles a boundary; genuinely varied edges stay rectilinear and work. Same code, same intent, different outcome. With concurrent writers the shared-chunk rewrite is also a commit conflict rather than a clean append.

### Origin
Introduced by #3899. 3.2.1's resolve_chunks dispatched on input syntax (_is_rectilinear_chunks), so a nested sequence always yielded a rectilinear grid; create_chunk_grid_metadata now dispatches on is_regular_nd after normalization to ChunksTuple. This reads as a side effect of consolidating the two normalization paths rather than an intended change — it's absent from the PR's "User-visible behavior changes" table, from changes/3899.bugfix.md, and from the 3.3.0 release notes.

### Possible resolutions
There's no way today to request a rectilinear grid independently of the edge values; init_array calls create_chunk_grid_metadata(outer_chunks) with no override.

1. Honor the requested kind — a nested-sequence chunks yields a rectilinear grid even when the edges are uniform. Restores 3.2.x behavior.
2. Accept explicit grid metadata (or a chunk_grid="rectilinear" hint) in create_array.
3. Add an explicit chunk-extension mode to resize/append, per #3870, which already raises this exact question.

Filing as design feedback against #4025, before rectilinear chunks leave experimental.

### Steps to reproduce

```python
# /// script
# requires-python = ">=3.12"
# dependencies = [
# ]
# ///
#
import hashlib
import zarr
zarr.config.set({"array.rectilinear_chunks": True})
store: dict[str, object] = {}
a = zarr.create_array(
    store=store, shape=(24,), chunks=[[10, 10, 4]], dtype="float32", zarr_format=3
)
print("at create:   ", a.metadata.chunk_grid)
def touched(start: int, stop: int, val: int) -> list[str]:
    """Chunk keys created or modified by writing a[start:stop]."""
    def snap() -> dict[str, str]:
        return {
            k: hashlib.md5(bytes(v.to_bytes())).hexdigest()
            for k, v in store.items()
            if k.startswith("c/")
        }
    before = snap()
    a[start:stop] = val
    after = snap()
    return sorted(k for k in after if after[k] != before.get(k))
print("write [20:24]", touched(20, 24, 2))
a.resize((34,))  # append a fourth 10-wide chunk
print("after resize:", a.metadata.chunk_grid)
appended = touched(24, 34, 3)
print("write [24:34]", appended)
assert appended == ["c/3"], (
    f"expected the appended region to be exactly one new chunk, got {appended}"
)
```

On main:
> at create:    RegularChunkGridMetadata(chunk_shape=(10,))
> write [20:24] ['c/2']
> after resize: RegularChunkGridMetadata(chunk_shape=(10,))
> write [24:34] ['c/2', 'c/3']
> AssertionError: expected the appended region to be exactly one new chunk, got ['c/2', 'c/3']

Swapping the dependency to zarr==3.2.1 passes, and shows the grid the spec asked for:
> at create:    RectilinearChunkGridMetadata(chunk_shapes=((10, 10, 4),))
> write [20:24] ['c/2']
> after resize: RectilinearChunkGridMetadata(chunk_shapes=((10, 10, 4, 10),))
> write [24:34] ['c/3']

### Additional output

_No response_

CAOShurong:
### Zarr version

3.3.1.dev34+gd44f9f92 (current `main` at the time of this PR)

### Numcodecs version

0.16.5

### Python Version

3.12.14 (local verification environment)

### Operating System

Windows 11 x64

### Installation

Editable install from a local clone of `main` (`d44f9f9`), venv via uv.

### Description

Fixes #4272. A nested-sequence `chunks` spec is an explicit rectilinear request, but since #3899 the stored grid kind was chosen from the edge **values** (`is_regular_nd`) rather than the form of the request, so a spec like `[[10, 10, 4]]` was silently collapsed to a regular grid. This implements resolution 1 from the issue ("honor the requested kind"), which @d-v-b endorsed there as the direction to take now.

#### Root cause

`normalize_chunks_nd` normalizes a nested sequence into a canonical `ChunksTuple`, which loses whether the user *requested* rectilinear; `create_chunk_grid_metadata` then dispatches on `is_regular_nd(chunks)` — and `is_regular_1d` counts "all chunks equal, last one smaller" as regular.

The two grids behave identically at creation time but diverge under `resize`: a regular grid extends the uniform pattern while a rectilinear grid appends an edge, so an append-only workload gets a different (chunk-rewriting) layout from the one it asked for, with no warning and nothing in the metadata recording the substitution.

#### Changes

- `create_chunk_grid_metadata` gains a keyword-only `requested_rectilinear: bool | None = None`. When set, it decides the grid kind; when `None`, behavior is unchanged (inferred from edge values) for callers that only have a normalized `ChunksTuple` (auto-chunking, etc.).
- Both v3 creation call sites derive the flag from the **raw** user input:
  - a nested-sequence `chunks=` requests rectilinear,
  - in `init_array`, a nested-sequence `shards=` also makes the *outer* chunk-grid metadata rectilinear (the grid describes the outer layout; without this, rectilinear shards with uniform-plus-tail edges would have regressed),
  - flat and `"auto"` specs keep inferring from values.
- Regression test `test_nested_sequence_request_stays_rectilinear_when_edges_uniform`: creates `[[10, 10, 4]]`, asserts a rectilinear grid, and asserts the post-resize append touches exactly one new chunk (`c/3`) instead of straddling `c/2 + c/3`; plus unit tests pinning the new flag's three modes and the unchanged flat/auto inference.
- Changelog fragment `changes/4272.bugfix.md`.

### Verification

Local run of the issue's reproducer on Windows/CPython 3.12:

Before (matches the issue's "On main" output):

```
at create:    RegularChunkGridMetadata(chunk_shape=(10,))
write [20:24] ['c/2']
after resize: RegularChunkGridMetadata(chunk_shape=(10,))
write [24:34] ['c/2', 'c/3']
AssertionError
```

After (matches 3.2.1 semantics from the issue):

```
at create:    RectilinearChunkGridMetadata(chunk_shapes=((10, 10, 4),))
write [20:24] ['c/2']
after resize: RectilinearChunkGridMetadata(chunk_shapes=((10, 10, 4, 10),))
write [24:34] ['c/3']
PASS
```

Test suites on this branch (all green): `tests/test_unified_chunk_grid.py` + `tests/test_metadata` + `tests/test_chunk_grids.py` = **612 passed**, `tests/test_array.py` = **1307 passed / 35 skipped**. The new regression test fails on pristine `main` and passes with this patch.

### Documentation

No docs page states the old collapse behavior; the changelog fragment documents the fix.

### AI

Development assisted by AI (Claude), disclosed per project policy.


d-v-b:
hi @CAOShurong I think this PR is superseded by #4218, let me know what you think?

d-v-b:
seperseded by https://github.com/zarr-developers/zarr-python/pull/4218

PR #4299, opened 2026-08-31
chore(deps): raise numcodes >= 0.16.3 for gcc>=15 support

romanc:
## Summary

zarr-python depends on numcodes. Numcodes - in turn - has a dependency on c-blosc. c-blosc had an issue with GCC 15's new default `--std=gnu23`. This issue got fixed in c-blosc [1] and numcodecs picked it up in [2] by updating the submodule. numcodecs 0.16.3 is the first release to ship with gcc 15+ support.

While the `uv.lock` file updated `numcodecs` to 0.16.5, the current minimal version of `numcodecs` in `zarr-python` remained `0.14` in the `pyproject.toml`. However, the `uv.lock` file is only read by `uv` and if users were to install `zarr-python` by any other means (e.g. plain `pip install`), on might get a version of `numcodecs` that is incompatible with GCC 15+.

This PR thus proposes to raise the minimal version of `numcodecs` in the `pyproject.toml` of `zarr-python` to 0.16.3 to ensure all users of `zarr-python` get support for GCC 15+.

[1]: https://github.com/Blosc/c-blosc/issues/393
[2]: https://github.com/zarr-developers/numcodecs/pull/772

## For reviewers

This is really just raising the minimal version of `numcodecs`. Related issue: #4300.

## Author attestation

## TODO

  N/A
  N/A
  N/A


d-v-b:
Thanks for this! I don't anticipate any problems with this change

romanc:
> Thanks for this! I don't anticipate any problems with this change

Happy to hear reassuring words. Would be awesome if this landed before the 3.3.1 release. I'll put it up for review as soon as tests are green.

d-v-b:
> if users were to install zarr-python by any other means (e.g. plain pip install), on might get a version of numcodecs that is incompatible with GCC 15+.

as I understand it, this gcc issue only applies to people who are compiling numcodecs from source, not people installing things via pip. is it possible you are working on a platform that doesn't have numcodecs wheels available? 

asking because bumping the minimum numcodecs version just because some environments cannot compile it from source might be disruptive for anyone zarr user who pinned numcodecs to a lower version for feature-related reasons

romanc:
> is it possible you are working on a platform that doesn't have numcodecs wheels available?

You make an excellent point. For some reason, `uv` decided to install `numcodecs` from source, triggering the whole issue. I am on linux (OpenSUSE 16.0) on an x86_64 arch with glibc 2.40 and python 3.14.7. Imo this should trigger a wheel download instead of building from source. So this might just be a me/setup issue. I can't seem to repro this standalone ... it just downloads a pre-built wheel as expected. Very much confused ...

Feel free to reject this PR if you are concerned about bumping the minimal version of numcodecs.

PR #4305, opened 2026-09-01
Fix infinite loop in _guess_num_chunks_per_axis_shard for 0-dimensional arrays

### LINKED ISSUE #4304: `shards="auto"` on a 0-d array hangs when `array.target_shard_size_bytes` is set
**Zarr version:** 3.3.0 (also on 3.2.1, and the loop is unchanged on `main`)

**Numcodecs version:** 0.16.5

**Python Version:** 3.13.3

**Operating System:** Windows

**Installation:** pip into a fresh venv

---

## Description

`shards="auto"` on a 0-d array works fine and gives you `shards=()`. Set
`array.target_shard_size_bytes` (the key added in #3568) and the same call hangs forever. No error,
no warning, nothing printed, it just spins in pure Python until you kill it.

So a config key that's only meant to affect shard size ends up deciding whether the call returns at
all.

`_guess_num_chunks_per_axis_shard` in `src/zarr/core/chunk_grids.py` only gets called when
`shards="auto"` and the key is set. Without the key the caller just uses 2, so you never reach it.
For a 0-d array both loop conditions are constant:

```python
bytes_per_chunk = math.prod(chunk_shape) * item_size   # math.prod(()) == 1, so == item_size
if max_bytes < bytes_per_chunk:                        # won't fire for any budget >= item_size
    return 1
num_axes = len(chunk_shape)                            # 0
chunks_per_shard = 1
while (bytes_per_chunk * ((chunks_per_shard + 1) ** num_axes)) <= max_bytes and all(
    c * (chunks_per_shard + 1) <= a for c, a in zip(chunk_shape, array_shape, strict=True)
):
    chunks_per_shard += 1
```

`(chunks_per_shard + 1) ** 0` is 1, so the first check is just `item_size <= max_bytes` and never
changes. `zip((), ())` is empty so `all()` is True. Neither depends on the counter, so it loops
forever.

Should return 1. A 0-d array has no axes to grow a shard along, which is what the no-budget path
already works out on its own when it gives you `shards=()`.

Fix is right after `num_axes = len(chunk_shape)`:

```python
if num_axes == 0:
    return 1
```

## Steps to reproduce

```python
# /// script
# requires-python = ">=3.11"
# dependencies = ["zarr==3.3.0"]
# ///
"""`shards="auto"` on a 0-d array never terminates when `array.target_shard_size_bytes` is set.

Each case runs in a subprocess with a 15 s timeout, so this script always exits.
"""

import subprocess
import sys
import warnings

CASES = {"none": None, "128MiB": 128 * 1024 * 1024}

if len(sys.argv) > 1:  # child: run one case
    import zarr

    warnings.simplefilter("ignore")  # "shard shape inference is experimental"
    budget = CASES[sys.argv[1]]
    with zarr.config.set({"array.target_shard_size_bytes": budget} if budget else {}):
        arr = zarr.create_array(store={}, shape=(), dtype="int64", shards="auto")
    print(f"returned, shards={arr.shards}")
else:  # parent: run both, bounded
    import zarr

    print(f"zarr {zarr.__version__}\n")
    for name in CASES:
        print(f"target_shard_size_bytes = {name}:")
        try:
            out = subprocess.run(
                [sys.executable, __file__, name], capture_output=True, text=True, timeout=15
            )
            print(f"    {out.stdout.strip() or out.stderr.strip().splitlines()[-1]}")
        except subprocess.TimeoutExpired:
            print("    DID NOT TERMINATE after 15 s")
```

Gives:

```
zarr 3.3.0

target_shard_size_bytes = none:
    returned, shards=()
target_shard_size_bytes = 128MiB:
    DID NOT TERMINATE after 15 s
```

Calling it directly shows the same thing, and that the early return still works:

```python
from zarr.core.chunk_grids import _guess_num_chunks_per_axis_shard

_guess_num_chunks_per_axis_shard(chunk_shape=(), item_size=8, max_bytes=4, array_shape=())
# 1, early return fires

_guess_num_chunks_per_axis_shard(chunk_shape=(), item_size=8, max_bytes=128 * 1024 * 1024, array_shape=())
# never returns
```

## Notes

Not a dup of #3751 ("Cannot write a 0 dimensional array with sharding", closed). That one was an
`IndexError` out of `_ShardIndex.get_chunk_slices_vectorized` on `arr[()] = 42.0` with an explicit
`shards=()`, fixed by #3793. This is array creation under `shards="auto"`, and it hangs instead of
raising.

Where I ran into it: anndata writes zarr v3 arrays with `shards="auto"` and reads the same config
key, so anything that sets a shard budget and then writes a 0-d array ends up here. anndata dodges
it at the moment because its scalar writers don't go through the sharding path, so nobody's actually
broken by this today. You can still get there from plain zarr, as above.

#4218 is refactoring `chunk_grids.py` but doesn't touch this loop.

Tomatokeftes:
Closes #4304.

Cause: for a rank-0 array, neither condition of the while loop in `_guess_num_chunks_per_axis_shard` depends on the counter (`(n+1) ** 0 == 1` and `zip((), ())` is vacuously true), so creating a 0-d array with `shards=auto` and `array.target_shard_size_bytes` set spins forever with no error.

Fix: return 1 when `chunk_shape` is empty. The caller builds the shard shape by zipping over the (empty) axes, so the result is `shards=()`, which is exactly what the no-budget path already produces. Budget and no-budget paths now agree at rank 0.

Includes a unit test on the helper and a regression test through `create_array`. Note: the repo has no pytest-timeout, so if the bug were reintroduced these tests would stall until the 600 s faulthandler timeout kills them with a traceback pointing at the loop, rather than failing fast.

Happy to move the guard into `resolve_outer_and_inner_chunks` instead if you would rather keep it out of the helper.

### DIFF
--- a/changes/4305.bugfix.md
+++ b/changes/4305.bugfix.md
@@ -1,0 +1,1 @@
+Fixed an infinite loop when creating a 0-dimensional array with `shards="auto"` while the `array.target_shard_size_bytes` config option is set. Such arrays now resolve to `shards=()`, matching the behavior when no shard size target is configured.
--- a/src/zarr/core/chunk_grids.py
+++ b/src/zarr/core/chunk_grids.py
@@ -865,6 +865,8 @@ def _guess_num_chunks_per_axis_shard(
     if max_bytes < bytes_per_chunk:
         return 1
     num_axes = len(chunk_shape)
+    if num_axes == 0:
+        return 1
     chunks_per_shard = 1
     # First check for byte size, second check to make sure we don't go bigger than the array shape
     while (bytes_per_chunk * ((chunks_per_shard + 1) ** num_axes)) <= max_bytes and all(
--- a/tests/test_chunk_grids.py
+++ b/tests/test_chunk_grids.py
@@ -3,14 +3,17 @@ from typing import Any
 import numpy as np
 import pytest
 
+import zarr
 from tests.conftest import Expect, ExpectFail
 from zarr.core.chunk_grids import (
     ChunkLayout,
+    _guess_num_chunks_per_axis_shard,
     _guess_regular_chunks,
     normalize_chunks_1d,
     normalize_chunks_nd,
     resolve_outer_and_inner_chunks,
 )
+from zarr.errors import ZarrUserWarning
 
 
 def _assert_chunks_equal(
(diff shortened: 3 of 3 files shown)


d-v-b (approved):
thanks!

PR #4310, opened 2026-09-03
feat(zarr-indexing): factor chunk plans into a columnar GridPartition

d-v-b:
## Summary

This is an update to the data structures used in `zarr-indexing` for relating a selection on a chunked array to a per-chunk plan. We currently generate objects per chunk eagerly, which is inefficient due to object creation overhead. FWIW I investigated pushing this down to rust but crossing the py03 boundary for a ton of tiny objects ends up pretty expensive. 

a better solution is to model the chunk plan as the result of a product of per-axis plans, which is what this PR adds. the end result restores a lot of performance that had been lost with the eager object creation.

based on a claude-authored PR here: https://github.com/d-v-b/zarr-python/pull/316

## Author attestation

## TODO


d-v-b:
this got a lot of review / improvement from claude and codex. we are pretty close to zarr's current indexing performance on the eager path. merging.

PR #4311, opened 2026-09-03
docs(subpackages): render subpackage changelogs as docs pages

d-v-b:
This is a Claude-authored PR that updates how our subpackages render their release notes in the docs. It's a self-merge situation.

The zarr-indexing docs sidebar and landing page linked to `CHANGELOG.md` on the GitHub `main` branch, so the docs for any version showed the changelog of `main`, and the link left the site entirely.

This adds `docs/release-notes.md`, which includes `CHANGELOG.md` via `pymdownx.snippets` with the package root appended to `base_path`. The root file stays where towncrier, PyPI, and GitHub expect it, and each docs build now renders the release notes for its own version. `tests/test_doc_examples.py` mirrors the new base path so the include-graph test keeps validating it.

The page is named `release-notes.md` rather than `changelog.md` because on a case-insensitive filesystem (macOS) `docs/changelog.md` shadows `CHANGELOG.md` and the include resolves to the page itself, rendering empty.

Verified locally with `mkdocs build --strict`, `tests/test_doc_examples.py`, and ruff.

`packages/zarr-metadata` has the same external changelog link and could get the same treatment in a follow-up.

### DIFF
--- a/packages/zarr-http-server/changes/4311.doc.md
+++ b/packages/zarr-http-server/changes/4311.doc.md
@@ -1,0 +1,2 @@
+The docs now render the release notes as a page instead of linking to
+`CHANGELOG.md` on GitHub, so each docs version shows its own changelog.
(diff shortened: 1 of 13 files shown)


d-v-b:
the PR description is out of date. It poses changing subpackages outside of zarr-indexing as a follow-up. but I'm also doing that here.

Issue #2834, opened 2025-02-14
bug with setitem with oindex and sharding


dcherian:
### Zarr version

v3.0.2

### Numcodecs version

?

### Python Version

3.12

### Operating System

any

### Installation

any

### Description

Discovered in https://github.com/zarr-developers/zarr-python/pull/2825

### Steps to reproduce

```python
import zarr
import numpy as np

store = zarr.storage.MemoryStore()
group = zarr.group(store)
array = group.create_array(
    name="zoo",
    shape=(1,2,1),
    chunks=(1,2,1),
    shards=(1,2,1),
    dtype=np.int32,
)
zindexer = (np.array([0]), np.array([0, 0]), np.array([0]))
new_data = np.full(array.oindex[zindexer].shape, fill_value=1)
array.oindex[zindexer] = new_data
```

raises the error:
```
ValueError: shape mismatch: value array of shape (1,2,1) could not be broadcast to indexing result of shape (2,)
```

It succeeds with `shards=None`.

brokkoli71:
I think the problem is that the information that the selection is an ``OrthogonalSelection`` gets lost on the way.

In `ShardingCodec._encode_partial_single` a new indexer is created to index the inner chunks:
```python
indexer = list(
    get_indexer(
        selection, shape=shard_shape, chunk_grid=RegularChunkGrid(chunk_shape=chunk_shape)
    )
)
```
There, `get_indexer` guesses what indexing should be done based on the selection tuple, and in this case guesses `CoordinateIndexer` instead of `OrthogonalIndexer`

moradology:
Yeah, `get_indexer` is unable to distinguish between the desired selectors. [This](https://github.com/zarr-developers/zarr-python/blob/main/src/zarr/core/indexing.py#L1377) and [this](https://github.com/zarr-developers/zarr-python/blob/main/src/zarr/core/indexing.py#L1386) both return true, so the latter case is never reached

brokkoli71:
One thing I would suggest as a consequence is to rename `get_indexer` to something like `guess_indexer`. 

Furthermore, we would need to pass the indexer or at least information about the type of indexer to the codecs and thus to the sharding codec so it would not need to guess the indexer.

brokkoli71:
Hey @moradology,
I just stumbled across this again — did you have any luck making progress on it? Happy to help if there’s anything I can do to move it forward.

moradology:
[...]

PR #4316, opened 2026-09-04
fix(sharding): flatten any multi-dim value for coordinate selections in partial writes

### LINKED ISSUE #4315: bug(indexing): oindex with mixed numpy array and integer indexing fails
this looks good, i'm going to approve + merge. there's a remaining bugged case that I will fix in a follow-up PR:

```python
import numpy as np, zarr
from zarr.codecs import ShardingCodec, BytesCodec

a = zarr.create_array(
    MemoryStore(), shape=(4, 4, 4), chunks=(2, 4, 4), dtype="int32",
    serializer=ShardingCodec(chunk_shape=(2, 2, 2), codecs=(BytesCodec(),)),
    compressors=None, fill_value=0,
)
a[:] = np.arange(64, dtype="int32").reshape(4, 4, 4)
a.oindex[np.array([3, 1, 2]), 1, np.array([0, 2])] = np.arange(6, dtype="int32").reshape(3, 2)
# ValueError: shape mismatch: value array of shape (1,2) could not be broadcast to indexing result of shape (1,)
```

_Originally posted by @d-v-b in https://github.com/zarr-developers/zarr-python/issues/4284#issuecomment-5537831542_

d-v-b:
## Summary

Fix a dumb singleton dimension bug that breaks oindex with mixed np.array / integer components with the sharding codec.

Closes #4315 

Claude wrote this, see the original PR here: https://github.com/d-v-b/zarr-python/pull/320

## Author attestation

## TODO


d-v-b:

Two follow-up commits, addressing the review finding on mask assignment and the test gap it exposed.

## `49f8872` narrows the flatten

The first version ravelled any multi-dimensional value once the shard-level indexer was a `CoordinateIndexer`. That is too broad: the shard-level selection cannot tell orthogonal from mask indexing, so a mask write with a `(2, 2)` value for four selected elements, or an orthogonal write with a spurious trailing axis like `(3, 2, 1)`, was silently accepted on a sharded array while raising `ValueError` on an unsharded one (and in numpy).

The value shape can tell them apart. An `np.ix_` selection has an N-D `sel_shape`, and the caller's value is that shape with only length-1 axes removed, which is exactly where `np.ix_` puts integer indices. The helper now ravels only when `value.shape` is `sel_shape` minus zero or more unit axes (`_drops_only_unit_axes`). A mask arrives with a 1-D `sel_shape`, so a 2-D value no longer qualifies and fails downstream the same way it does unsharded; an extra axis fails to qualify for the same reason. `test_sharding_set_rejects_value_with_wrong_rank` covers both on both pipelines and fails against the previous commit.

## `aaf30c8` makes the property tests able to find this

The question of how this got past Hypothesis had a plain answer: the case was unreachable.

- `test_oindex`, `test_mask_indexing` and `test_block_indexing` skipped their set half on sharded arrays with `assume(zarray.shards is None)`, added in #2825 when #2834 was filed and never lifted. `test_vindex` had its set half commented out.
- `orthogonal_indices` wrapped every bare integer as a one-element array, so zarr never received an integer index and the dropped-axis path in `OrthogonalIndexer` was unreachable. `basic_indices(min_dims=1)` never yields an integer either, so that branch was dead.
- `arrays()` only drew a shard shape when every axis had a chunk strictly between 1 and the axis length. With the v3 and regular-grid draws on top, 2 of 500 `test_oindex` examples were sharded.

The commit lifts the skips, draws integers explicitly (the numpy indexer gets a 0-d index so its result drops the axis as `oindex` does), enables the vindex write behind a duplicate-point filter, and lets any chunk that fits the array be sharded (33 of 500 now, with an `event()` so `--hypothesis-show-statistics` shows the split). With these changes `test_oindex` fails against the code before this PR with the mixed-integer shape mismatch at the default 300 examples, and passes with it. The full property module and the stateful store and hierarchy machines pass under the default and nightly profiles.

The mask leniency is still outside what the property tests express, since they only draw correctly shaped values; the unit error test carries that.

d-v-b:
it's going in! this unblocks a longstanding hole in our array indexing coverage for sharded arrays.

PR #4323, opened 2026-09-09
docs: state the rectilinear-flag consequence of #4272 and qualify nchunks_initialized

d-v-b:
This PR updates the documentation about changes to the parsing of nested sequences of ints in the context of different chunk grids, documenting specifically that when the rectilinear chunks feature is disabled, then nested sequences of ints that describe a regular grid will be parsed as a regular _rectilinear_ grid and error. on 3.3.0 this input generated a regular chunk grid, but on earlier versions it was an error. 

This documents a very minor breaking change, which is arguably reverting an wart in the 3.3.0 release.

## Summary

Two documentation corrections surfaced by a pre-release review of everything merged since v3.3.0, ahead of the 3.4.0 changelog build (#4256).

`changes/4272.bugfix.md` describes the fix that makes explicit per-chunk size lists always request a rectilinear grid, but it never says what that means for users who leave `array.rectilinear_chunks` off: a uniform nested sequence such as `((4,), (4,))` or `[[3, 3, 1]]` — the shape of a dask array's `.chunks` — was silently accepted as a regular grid in 3.3.0 and now raises the "experimental and disabled by default" error, as it did in 3.2.x. The fragment now states that explicitly and points at the integer-per-dimension form to use instead. Its closing "arrays with regular chunk grids are unaffected" sentence is qualified accordingly.

The `nchunks_initialized` docstrings (both `Array` classes and the module-level helper) were rewritten in #4218 to read as an exact count, but each initialized shard is counted at its full declared size, so an edge shard that overhangs the array contributes chunks that lie entirely outside it and the result can exceed `nchunks` (e.g. `shape=(10,)`, `chunks=(2,)`, `shards=(8,)` fully written reports 8 with `nchunks == 5`). That behavior is unchanged from 3.3.0; the docstrings now say so.

## For reviewers

No code changes — only a changelog fragment and three docstrings. The claim that the nested-sequence input raised in 3.2.x was verified against an installed 3.2.1.

## Author attestation

## TODO

* [x] Add unit tests and/or doctests in docstrings (n/a — documentation only)
* [x] Changes documented as a new file in `changes/` (edits the existing #4272 fragment)
* [x] Test coverage is 100% (Codecov passes) (n/a)

### DIFF
--- a/changes/4272.bugfix.md
+++ b/changes/4272.bugfix.md
@@ -10,6 +10,17 @@ regular grid, nested sequences produce a rectilinear grid. Rectilinear grids
 remain gated behind `zarr.config.set({"array.rectilinear_chunks": True})`.
 See #4174 for the accompanying O(1) chunk normalization change.
 
+One consequence for users who never enable rectilinear chunks: because a
+nested sequence now always requests a rectilinear grid, a per-dimension
+sequence of edge lengths that happens to be uniform — for example the
+`((4,), (4,))` or `[[3, 3, 1]]` form that a dask array's `.chunks` attribute
+produces — is no longer quietly accepted as a regular grid when the
+`array.rectilinear_chunks` option is off. Such input raises
+`ValueError: Rectilinear chunk grids are experimental and disabled by default`,
+exactly as it did in 3.2.x; the silent acceptance existed only in 3.3.0. Pass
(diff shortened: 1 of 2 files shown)


d-v-b:
I will be self-merging this when it's green

PR #4307, opened 2026-09-02
fix: resolve chunks=-1 to chunk size 1 on zero-length axes

d-v-b:
## Summary

follow up for #4304, ensures that we handle 0-length axes correctly in auto chunk size logic. written by claude, claude's PR is [here](https://github.com/d-v-b/zarr-python/pull/315) 

## Author attestation

- [x] I am a human, these are my changes, and I have reviewed and understood every change and can explain why each is correct.

## TODO

* [ ] Add unit tests and/or doctests in docstrings
* [ ] Add docstrings and API docs for any new/modified user-facing classes and functions
* [ ] New/modified features documented in `docs/user-guide/*.md`
* [ ] Changes documented as a new file in `changes/`
* [ ] GitHub Actions have all passed
* [ ] Test coverage is 100% (Codecov passes)

PR #4328, opened 2026-09-09
fix: resolve chunks=False to chunk size 1 on zero-length axes

d-v-b:
This AI-authored PR ensures that `chunks=False` (a pattern I dislike very much, but which we keep around for backwards compatibility), doesn't create 0-length chunks when the array shape is 0.

## Summary

`chunks=False` means "one chunk covering the whole array", but on a zero-length axis it produced a chunk of size 0. For Zarr format 3 this was rejected by `RegularChunkGridMetadata` (`integer chunk edge length must be >= 1, got 0`); with `shards="auto"` it raised `ZeroDivisionError`; and for Zarr format 2 it silently wrote `chunks: [0]`, which corrupted reads once the axis was appended to and resized — the degenerate grid reports zero chunks, so reads return uninitialized memory or zeros.

#4307 already fixed the equivalent `chunks=-1` sentinel by clamping to `max(span, 1)`, matching `chunks="auto"`. This PR routes the `False` branch of `normalize_chunks_nd` through that same `-1` path so there is a single clamp; both spellings now yield chunk size 1 on an empty axis.

The #4307 regression test is extended to cover `False` alongside `-1` for `(0,)`, `(0, 4)` and `(4, 0)` on both formats, with and without `shards="auto"` (and with a shard-size budget), asserting the in-memory chunks/shards, the stored chunk grid, and that `append`/`resize` along the empty axis round-trip data. Without the source change exactly the 14 `False` parametrizations fail and every `-1` case passes.

## For reviewers

The source change is a two-line fallthrough (`chunks = -1`) replacing the direct `FixedDimension` construction in the `False` branch, on top of #4218's chunk-grid rewrite. The v2 behavior change is intentional: the previous `chunks: [0]` metadata was invalid and returned wrong data after a resize, and no existing test depended on it.

## Author attestation

## TODO

* [x] Add docstrings and API docs for any new/modified user-facing classes and functions (n/a)


d-v-b:
this is a simple bugfix, so i'm self-merging

PR #4329, opened 2026-09-09
feat: accept numpy arrays as chunks/shards in create_array, Group.create_array and from_array

d-v-b:
AI-authored PR that lets `create_array` and callers use numpy arrays to declare chunks and shards, which brings `create_array` into parity with `create`, which already accepted numpy arrays for chunks. 

## Summary

`zarr.create_array`, `Group.create_array`, `zarr.from_array` and the entry points built on them now accept a numpy array as the `chunks` or `shards` specification. This is a **feature**, not a fix: that API has never accepted numpy arrays in any 3.x release (verified against 3.0.10, 3.1.6, 3.2.1, 3.3.0 and `main`). Each entry point compares the specification to a string sentinel before normalizing it — `chunks == "auto"` in `init_array`, `shard_shape == "auto"` in `resolve_outer_and_inner_chunks`, and `chunks` / `shards == "keep"` on both branches of `_parse_keep_array_attr` — and for a numpy array that comparison is elementwise, so the `if` raises numpy's ambiguous-truth-value `ValueError` before the normalizer is reached.

The change adds `_is_auto` / `_is_keep` helpers in `chunk_grids.py` that guard the string comparison with `isinstance(spec, str)` and return `TypeIs[Literal[...]]` so mypy's narrowing in the negative branch is unchanged, and routes the six sentinel checks through them. Strings, tuples, lists, numpy integer scalars, and the `ChunkGridMetadata` values allowed by #4218 behave exactly as before. It brings this API in line with the legacy `zarr.create` / `zarr.array` / `zarr.open_array` functions, which have accepted numpy arrays since 2.x (they broke in 3.3.0 with the numpy-integer regression and were restored by #4257). A parametrized test covers four entry points × three spec forms × shards on/off; without the source change exactly the eight numpy-array cases fail.

This PR also corrects the changelog entry for #4257, which said "numpy arrays are now also accepted as chunk specifications" without qualification. That was true only of the internal normalizer and the legacy functions; the entry now says so and points here for the `create_array` family.

## For reviewers

The helpers return `TypeIs` rather than `bool` because the bare `== "auto"` literal comparison was implicitly narrowing `chunks` / `shards` for mypy in the `else` branches. A grep of `src/` for `"auto"` / `"keep"` comparisons on chunk/shard names finds only these six sites; the filters/compressors/serializer sentinel comparisons are deliberately left alone since those never receive array-likes.

## Author attestation

## TODO

* [x] Add docstrings and API docs for any new/modified user-facing classes and functions (n/a)
* [x] New/modified features documented in `docs/user-guide/*.md` (n/a — accepted input types are documented on the parameters)


d-v-b:
I would like to get this out for 3.4.0, and review times have been a bit long recently so I'm going to self-merge this shortly.

PR #4330, opened 2026-09-09
fix: remove attribute nesting depth limit

d-v-b:
An AI-written PR that reverts a regression introduced in #4063, which limited the contents of `attributes` to a nest depth of 64. 

:robot: ai text below :robot:

## Summary

A Zarr v3 array can be created with attributes nested more than 64 levels deep, but reopening the same array fails with `ValueError: JSON value nesting exceeds the maximum depth of 64`. This restriction was introduced in #4063 and regresses the behavior in 3.3.0.

Remove the recursive attribute validator and its depth limit, restoring the existing attribute handling through the JSON reader/writer and the metadata constructor. Keep msgspec validation for structured metadata fields. Update the unreleased changelog entry and add create/reopen regression coverage for nested objects and arrays at depths 0, 65, and 100, for both Zarr formats.

## For reviewers

Please check that the restored attribute handling is consistent with array creation and group metadata. This draft is awaiting the human author's edits and review before it is marked ready.

Validation on current upstream main plus this fix: 2,188 tests passed and 48 skipped across the JSON parsing, metadata, group, and array suites in the Python 3.12 minimal Hatch environment. Ruff lint and formatting checks and `git diff --check` passed.

## Author attestation

- [ ] I am a human, these are my changes, and I have reviewed and understood every change and can explain why each is correct.

## TODO

- [x] Add regression tests
- [x] Update the existing unreleased changelog entry
- [ ] Human author edits the description and reviews the changes
- [ ] GitHub Actions have all passed
- [ ] Codecov passes

PR #4331, opened 2026-09-09
test: give the sharded coordinate-selection write path real weight in test_oindex

d-v-b:
AI-authored PR that strengthens our property-based tests for indexing to include more sharded arrays.

## Summary

#4284 / #4316 fixed orthogonal writes on sharded arrays where the sharding codec re-derives a `CoordinateIndexer` and must ravel the value, but the property test meant to cover it, `test_oindex`, reached that path in 2 of 300 examples under the derandomized `ci` Hypothesis profile — a fixed outcome, so a green run said almost nothing about the fix. The test draws `one_of(simple_arrays, rectilinear_arrays)`, rectilinear arrays cannot shard, and only ~8.6% of `simple_arrays` draws are sharded.

This adds `zarr.testing.strategies.sharded_arrays`, a strategy that always yields a sharded Zarr format 3 array (chunk shape, shard shape, subchunk write order and inner codec chain are drawn; the `ShardingCodec` construction is factored into a private helper shared with `arrays()` with the same draw order, so existing derandomized sequences are unchanged), and draws it as a third arm in `test_oindex` and `test_vindex`. The value-reshape path is now exercised in ~50 of 300 `ci` examples (13–27% of writes across random seeds) with no measurable change in suite time, and removing the guard in `sharding.py` makes `test_oindex` fail (shrunk example: a `(1, 3, 1)` array with shards `(1, 3, 1)`, selection `(array([0]), slice(1, 3), 0)`, value shape `(1, 2)`). A Hypothesis `event()` at the write site makes the reach visible under `--hypothesis-show-statistics`.

Separately, the docstring of `test_set_selection_rejects_value_with_wrong_rank` in `tests/test_indexing.py` claimed storage layout never changes which writes are rejected; it does — a `(2, 2, 1)` value on `oindex[[3, 1], [0, 2]]` is accepted by a chunked array (each chunk's piece is `(1, 1, 1)`, which numpy broadcasts) and rejected by a sharded one — so the docstring now states exactly what the test pins.

## For reviewers

The design choice worth a look is a dedicated strategy rather than a `sharded=` knob on `arrays()`: Zarr format 2 arrays cannot shard, so a knob would have to silently override `zarr_formats`. Note that the fixed path is reached by a bare-int axis plus one array axis too (`OrthogonalIndexer` routes through `ix_` when `n_array_dims > 1 or drop_axes`), not only by two or more array axes; the `min_dims=2` on the `test_oindex` arm follows from the guard needing a ≥2-D value. `derandomize=True` re-rolls the outcome on every edit to the test body, so the `ci` reach count is only stable while the test is.

## Author attestation

## TODO

* [x] Add docstrings and API docs for any new/modified user-facing classes and functions (`sharded_arrays` is picked up by the existing `zarr.testing.strategies` API page)


src/zarr/testing/strategies.py
```diff
@@ -255,6 +255,19 @@ def shard_shapes(
...
 
+@st.composite
+def _sharding_codecs(draw: st.DrawFn, *, chunk_shape: tuple[int, ...]) -> ShardingCodec:
+    """A ``ShardingCodec`` over ``chunk_shape`` with a drawn subchunk write order and inner codec chain."""
+    subchunk_write_order = draw(subchunk_write_orders)
+    inner_codecs = draw(sharding_inner_codecs, label="sharding inner codecs")
+    return ShardingCodec(
+        subchunk_write_order=subchunk_write_order,
+        codecs=inner_codecs,
+        index_codecs=[BytesCodec(), Crc32cCodec()],
```
d-v-b, line 278: consider supporting recursive sharding. A single recursion is likely enough to catch weird issues.

d-v-b:
this PR also fixes a lint error that snuck in #4330

PR #4334, opened 2026-09-09
fix(chunk-grids): one invariant for zero-length axes across model, clamps, and metadata

d-v-b:
This PR tightens our representation of chunk grids to disallow the representation of 0-length chunks. This has compatibility implications: previous versions of `zarr` could access arrays with 0-length chunks and exercise a subset of the Zarr API against them (reading / writing attributes), but not write chunks. I'm still thinking about the right compatibility shim for this, hence the draft status,

Previous PRs that removed 0-length chunks caused issues downstream so I'd like some eyes on this change (cc @TomNicholas) 

Makes chunk sizes positive while allowing zero-length array axes. `FixedDimension` now rejects `size=0`, and full-span chunk inference uses a shared helper returning `max(span, 1)`. Rectilinear creation accepts a non-empty list of positive chunk sizes on an empty axis and retains those sizes for later growth.

## Compatibility policy

A stored chunk size of 0 on a zero-length axis is read as 1 with a `ZarrUserWarning`. A zero chunk size on a positive-length axis is rejected. This is a compatibility policy for the current grid model; the metadata does not prove whether chunk payloads exist in the store.

This now covers both formats. Zarr format 3 was added after measuring what older releases actually wrote — before, arrays written by 3.0 and 3.1 could not be opened at all.

Both formats get the policy from one routine, `parse_stored_regular_chunk_shape` in `zarr.core.metadata.common`, which checks a stored regular chunk shape against its array's shape: one entry per axis, every chunk size at least 1, and the empty-axis exception above. It is regular-grid-specific by design: another chunk grid, such as the rectilinear grid, is free to define its own semantics for 0-length chunks, so its chunk sizes are never passed to it. `ArrayV2Metadata.__init__` calls it directly, which also replaces V2's separate after-the-fact length check (`parse_metadata`). For Zarr format 3 it is called from `ArrayV3Metadata.__init__` rather than from the chunk grid parser, because chunk grid metadata does not carry the array shape, and only for a grid named `regular` whose `chunk_shape` is all integers; anything else goes to the chunk grid parser untouched.

## What older releases wrote

Measured with real installs, creating an array with a zero-length axis:

| Format | Stored | Written by |
|---|---|---|
| V2 `chunks` | `[0]` | 2.18.7, and every 3.x before 3.4 — 3.3.0 for `chunks=-1` and `chunks=False`, 3.0–3.2 for `chunks=(0,)`/`False` |
| V2 `chunks` | `[false]` | 3.0.x for `chunks=False` |
| V3 `chunk_shape` | `[0]` | 3.0.x and 3.1.x |
| V3 `chunk_shape` | `[false]` | 3.0.x for `chunks=False` |

So this is not only about zarr-python 2.x: 3.3.0 wrote a zero chunk size for an empty Zarr format 2 array a few weeks ago, and 3.4.0 still reads those stores.

## Why this matters beyond validation hygiene

Appending to such a Zarr format 2 array **loses data silently** in 3.4.0. Taking an empty array created by 3.3.0 with `chunks=-1` and appending `[7, 8, 9]` with 3.4.0: the array reports `shape=(3,)`, no chunk file is written, and it reads back `[0, 0, 0]` with no error. With this PR the same store opens with a warning, appends correctly, and re-saves to valid metadata.

Zarr format 3 stores with a zero chunk size fail to open in 3.4.0 rather than losing data, so that half is a loud failure being fixed, not a silent one.

## Re-saving corrected metadata

Both warnings say how to persist the corrected chunk size: open the array writable and call `array.update_attributes({})`, which rewrites the whole metadata document from the parsed metadata. Verified for both formats against real legacy stores: the document is rewritten with chunk size 1 and reopening is warning-free.

## Tests

Tests cover six chunk spellings over empty/scalar shapes and applicable format/sharding combinations; separate cases check invalid chunk edges and positive-extent rejection. The policy itself is tested on `parse_stored_regular_chunk_shape`: one table of valid and legacy inputs (multi-axis shapes, JSON `false`, numpy integers) and one test per rejection (dimension mismatch, zero on a non-empty axis, negative). Wiring is tested end to end: a round-trip test over both formats and both stored spellings (`0` and JSON `false`) that opens a legacy store, appends, checks the data survives, re-saves with `update_attributes({})` and asserts a clean reopen, plus a rejection test for a stored zero chunk size on a positive-length axis in both formats. The Zarr format 3 cases were confirmed to fail without the shim. Full suite: 8904 passed.

## Merge and release notes

- Up to date with `main`. Merges cleanly with #4376 and #4377; with #4375 there is a one-line import conflict in `metadata/v3.py` (adjacent import lines; keep both). All four together, with that resolved: 8953 tests pass, and a real zarr 3.2.1 mixed-grid store (#4374) still reaches #4375's reader because it is not a regular chunk shape.
- Release ordering: per @TomNicholas above, a zarr release containing this should follow the VirtualiZarr release that includes zarr-developers/VirtualiZarr#1089, which is still open. Merging here is not blocked by it.


TomNicholas:
I was slightly ahead of you here - I think as long as I release [my VZ PR](https://github.com/zarr-developers/VirtualiZarr/pull/1089) before you release this then VZ at least will not break as a result of this change in Zarr-python.

PR #4335, opened 2026-09-09
fix: preserve the source Zarr dtype in from_array

d-v-b:
This AI-authored PR ensures that `from_array` uses the Zarr data type object, when available, instead of the numpy-like data type exposed by `array.dtype`, which ensures that data types like vlen-bytes have their data type preserved via from_array.

`zarr.from_array({}, data=source)` fails when `source` is a variable-length bytes array. It passes the source's NumPy `object` dtype to array creation, losing the complete Zarr dtype already available in the source metadata. Inferring a storage type from `object` then raises `ValueError` because it is ambiguous.

Preserve the source's explicit Zarr dtype when creating the copy. Array-like inputs continue to use their native dtype.

A parametrized test covers all 22 registered dtype examples in formats 2 and 3, checking dtype preservation and data when copying and reopening. Before the fix, both variable-length bytes cases failed.

Validation: 1,152 `from_array` tests passed, 720 unsupported format-2 sharding combinations skipped. All repository commit checks passed.


src/zarr/core/group.py
```diff
@@ -1257,11 +1259,15 @@ async def require_array(
...
-            dtype = parse_data_type(
-                "float64" if dtype is None else dtype,
-                zarr_format=self.metadata.zarr_format,
-            ).to_native_dtype()
+            # Existing arrays need a native dtype comparison, not storage-type
+            # inference: object is valid here even though it is ambiguous for creation.
+            dtype = "float64" if dtype is None else dtype
+            try:
+                dtype = np.dtype(cast("npt.DTypeLike", dtype))
+            except (TypeError, ValueError):
```
d-v-b: we should explain what triggers these exceptions in a comment

d-v-b:
simple bugfix, so i'm self-merging

PR #4338, opened 2026-09-10
docs: explain from_array attribute copy recursion limit

d-v-b:
AI-authored PR that clarifies the behavior of `from_array` when it comes to copying `attributes` from the source array to the destination array. The current code uses `copy.deepcopy` which is good insofar as it avoids copying by reference, but bad insofar as `copy.deepcopy` can trigger a recursion error for highly nested dicts. There are a few alternatives we could pursue if this proves problematic, but documenting it is the direction I'm taking for now.

Document that `from_array` deep-copies inherited attributes and can raise `RecursionError` for sufficiently deep documents even when storing and reopening succeeds. The threshold depends on Python's recursion limit and call stack. The API documentation and attribute guide explain how to omit inherited attributes, and the #4325 release note records the limitation.

This documents the accepted limitation for 3.4.0. A consistent policy for the Python-dictionary/JSON boundary will be addressed separately. No runtime behavior changes.

Validation: all repository commit checks passed, including docstring validation, Markdown lint, Ruff, and mypy.


d-v-b:
I'm self-merging this

PR #296, opened 2018-08-31
Add support for shapes in dtype definitions

onalant:
The objective of this PR is to resolve #111. Metadata encoding and
decoding is modified to accept shaped fields of structured dtypes and
shaped arrays of primitives (e.g. `(10,10)f8`).

Tests not utilizing optional requirements pass on Python 2.7 and 3.6. I
am working on setting up my environment with the optional requirements,
and hope to have those tests running very soon. If needed, I can expand
the test suite to more rigorously test the new encoding functionality.

`docs/tutorial.rst` will also be updated to include information about
shaped fields in dtypes.

Thank you for your time and consideration.

TODO:
* [x] Add unit tests and/or doctests in docstrings
* [x] Changes documented in docs/release.rst
* [x] Docs build locally (e.g., run ``tox -e docs``)
* [x] AppVeyor and Travis CI passes
* [x] Test coverage to 100% (Coveralls passes)

PR #4340, opened 2026-09-12
fix(dtype): fix nested v2 round trips and detect unsupported field conversions

d-v-b:

Fix nested structured dtype round-tripping in V2 metadata, detect field features unsupported by the current native dtype conversion, and warn when that conversion changes layout.

- Validate nested V2 field definitions as lists of fields. The baseline can emit nested metadata that its own dtype parser rejects.
- Reject native NumPy field titles and subarray fields, including nested ones, with a ValueError identifying the field. On the baseline, dtype conversion treats string title aliases as additional fields and subarray fields as raw bytes. Array writes using these inputs can consequently fail with unrelated errors.
- Preserve the baseline's conversion of padded layouts to packed layouts, adding a warning that field values are preserved when writing arrays but offsets and itemsize may change. The integration tests cover aligned layouts, nested padding and explicit offsets in both formats.

## Format and implementation boundaries

Subarray rejection is a limitation of the current Zarr-Python dtype implementation, not of V2. The [V2 specification](https://zarr-specs.readthedocs.io/en/latest/v2/v2.0.html#data-type-encoding) explicitly supports `[fieldname, datatype, shape]`. Nested fields and subarrays were clarified in [2018's #296](https://github.com/zarr-developers/zarr-python/pull/296#issuecomment-429141298). A probe using Zarr-Python 2.18.7 successfully writes and reopens a structured subarray field. Restoring that support is outside this change.

For titles, NumPy's `descr` supports `(title, name)` pairs. Zarr adopted `dtype.descr` in [April 2016](https://github.com/zarr-developers/zarr-python/commit/a1e8c5598c37a7c69a3891b69f9b5bec00e99d58), but the JSON decoder did not reconstruct those pairs as tuples. The clarified V2 grammar describes field names as strings. The reviewed history provides no explicit rationale for excluding titles; this PR makes no claim that they are inherently unrepresentable.

The [V3 struct extension](https://github.com/zarr-developers/zarr-extensions/blob/main/data-types/struct/README.md) separately specifies named, fixed-size fields and packed encoding. Its constraints should not be applied retroactively to V2.

## Validation

- 770 dtype, V2 and property tests passed; 5 existing skips.
- New strategies sample built-in dtype classes, bounded parameters and nested layouts. They do not claim exhaustive coverage or support arbitrary third-party dtype constructors. Datetime generation excludes the microsecond alias and normalizes generic-unit scale factors; those normalization issues remain separate.
- The structured-dtype property independently classifies NumPy field descriptors. It requires rejection for the currently unsupported features and successful conversion for the other generated cases, checking packed dtype equality and layout warnings. An injected unrelated ValueError is detected rather than counted as success.
- Compared runtime behavior of Zarr-Python 2.18.7, the PR baseline `ba883a556`, and this branch. The statements above distinguish those versions deliberately.

Supersedes d-v-b/zarr-python#188.


d-v-b:
🎵 self-merging when tests pass 🎵

PR #4352, opened 2026-09-14
fix(metadata): validate codec chains against the threaded chunk spec

d-v-b:
This AI-authored PR ensures that codec chains validate against the chunk shape emitted by the preceding codec, not the original chunk grid.

Codec validation needs the geometry and dtype produced by preceding codecs. In particular, transpose followed by sharding must check divisibility against the transposed chunk sizes. Validating against the original grid can accept chains that read back incorrect data or reject valid chains. The same validation is needed inside sharding.

`evolve_and_validate_codecs` threads chunk specs through evolution, validation, and metadata resolution. Inner sharding chains use the actual inner spec, including its fill value.

### Geometry is threaded as a whole chunk grid

Following [zarrs](https://github.com/zarrs/zarrs) (`encoded_chunk_grid` / chunk-local grids), codecs describe how they map the chunk grid through a new `BaseCodec.resolve_chunk_grid(shape=..., chunk_grid=...)` method, and validation carries the grid through the chain rather than enumerating chunk shapes. Cost is proportional to the number of codecs and dimensions, independent of how many distinct chunk shapes a rectilinear grid has.

- Codecs that do not override `resolve_metadata` declare the identity by default.
- `CastValue`, `ScaleOffset` and the numcodecs `Delta`, `FixedScaleOffset` and `AsType` declare the identity explicitly (they change dtype or fill value, never shape).
- `TransposeCodec` permutes the shape and the per-axis edge lists.
- A declaration is ignored if a subclass overrides `resolve_metadata` without the hook, or if it disagrees with `resolve_metadata` on the representative chunk.

### Trade-off: chunk-local validation

A codec that overrides `resolve_metadata` and returns `None` from `resolve_chunk_grid` (the default) is still validated exactly on a regular grid, since every chunk has one shape. On a **rectilinear** grid it makes the rest of the chain *chunk-local*: later codecs are validated against one representative chunk (largest edge per axis). Rejections are always correct, but acceptance is incomplete: a chain that is invalid only for some other chunk shape is accepted at creation and fails when such a chunk is first encoded or decoded.

To make that failure an error rather than corruption, `ShardingCodec._get_chunks_per_shard` now checks divisibility at run time (previously it floor-divided; with the check removed, a non-dividing shard writes and reads back wrong rows). Codecs whose correctness depends on chunk shape must do the same.

The trade-off is documented in the docstrings of `evolve_and_validate_codecs`, `resolve_chunk_grid` and the sharding check, in a new "Chunk geometry and validation" section of the extending guide, in the rectilinear sharding docs, and in the changelog entry.

### Tests

Accepted and rejected transpose/sharding chains (also behind `Delta`), reshape followed by transpose, nested sharding, rectilinear divisibility, actual fill values, data and metadata round trips. Chunk-local behaviour is pinned from both sides: an undeclared filter on a rectilinear grid still rejects a shard shape the representative chunk fails, accepts one that only a smaller chunk fails and then raises on the first write to that shard, and does not enumerate chunks to find a shape-changing codec's error. A table test covers when a declared grid is trusted or ignored. The bounded-call-count regression now includes `Delta` and an undeclared filter chains at ranks 4 and 12.

Validation: full suite **8806 passed, 1291 skipped, 4 xfailed**; pre-commit hooks incl. mypy pass. A 3-d rectilinear grid with 100 distinct edges per axis and `(Delta, Bytes, Zstd)` took ~3.8 s per metadata construction with the previous streaming fallback; now ~0 s.

Known, separate issue (not fixed here): `create_codec_pipeline` evolves the pipeline against a placeholder all-ones chunk shape on rectilinear grids, which breaks any codec whose `resolve_metadata` depends on chunk shape, on `main` too.

Supersedes d-v-b/zarr-python#303.


d-v-b:
this PR has a performance impact for codec creation with rectilinear chunk grids, where we loop over every chunk shape defined in a grid, checking each one for compatibility with the entire codec chain. I'm working on a mitigation.

d-v-b:
requesting review because this adds to codec API

src/zarr/abc/codec.py
```diff
@@ -148,6 +148,55 @@ def resolve_metadata(self, chunk_spec: ArraySpec) -> ArraySpec:
         """
         return chunk_spec
 
+    def resolve_chunk_grid(
```
d-v-b, line 151: @ilan-gold this is new API on the codec abc. we need the whole chunk grid to be in-scope for resolution because combining a rectilinear chunk grid with e.g. a transpose or reshape codec creates a large number of chunk shapes, and each chunk size needs to be consistent with the subchunk grid of e.g. a sharding codec. The meaning of "consistent" will in theory vary with the downstream codecs, but for sharding we just need to ensure that each incoming chunk shape tiles the sharding chunk grid. 

this would be a lot easier if the sharding codec chunk grid was semi-regular.

kokhlo (commented):
**Verified review** — I duplicated the issue #4354 scenario on `main` (7291469) and on this branch (5a52e85) before reading the diff closely; findings below are all reproduced locally, not read off the code.

## Root cause confirmed independently

On `main`, `ArrayV3Metadata.__init__` validates every codec against the *array-level* `chunk_grid` (`v3.py`, old loop: `c.validate(shape=spec.shape, dtype=spec.dtype, chunk_grid=chunk_grid_parsed)`), and `ShardingCodec.validate` checks every rectilinear edge of that original grid against `chunk_shape`. After a shape/rank-changing array→array codec the geometry no longer matches the original grid, so main produces **axis-confused false rejections** and, worse, would accept chains that read back wrong data. Reproduced on main:

```
zarr 3.4.1.dev12+g7291469 [main]
rectilinear chunks ((30,),(34,)) + transpose(1,0) + sharding(chunk_shape=(17,15))
REJECTED: ValueError: Chunk edge length 30 in dimension 0 is not divisible by the shard's inner chunk size 17.
```

(17 divides the *transposed* dim-0 edge 34, 15 divides the transposed dim-1 edge 30 — the chain is valid; the error is checking the untransposed grid.) Same construction on this branch is accepted and round-trips metadata:

```
zarr 3.4.1.dev26+g5a52e85 [pr]
metadata ACCEPTED
JSON roundtrip of metadata: OK
```

Also verified on the branch: representative-chunk evolution cost stays flat (3-d rectilinear, 100 distinct edges/axis, `Delta+Bytes+Zstd` → 0.0001s metadata construction), the chunk-local accept path ends in the new loud `_get_chunks_per_shard` error rather than silent mislayout, and `tests/test_codecs/test_codec_chain_validation{,_properties}.py` pass (40 passed). My a-priori gap candidates — inner sharding grids and `create_codec_pipeline`'s old all-ones placeholder — are both addressed (`evolve_and_validate_codecs` inside `evolve_from_array_spec`; representative-chunk evolution in `array.py`).

## Trust-boundary probe: declaration spoofing fails on all three paths

I tried to produce a codec that lies via `resolve_chunk_grid` (identity declaration while reshaping at runtime), through subclassing rather than a from-scratch codec, since that's the path the docstring's distrust rules exist for:

1. Subclassing `TransposeCodec` and overriding only `resolve_metadata` → correctly treated as undeclared (`_defining_class` MRO walk: `metadata_owner is not grid_owner` → `None` → chunk-local).
2. Subclassing `TransposeCodec` with an inconsistent `resolve_chunk_grid` override → caught by the repr-disagreement check.
3. Subclassing a no-`resolve_metadata` codec (`GzipCodec`) that reshapes at runtime → inherited identity declaration is *trusted*, but this is the right call: for every shipped codec that inherits the default, the declaration is structurally true, and the runtime divisibility guard catches the damage on the sharding path.

No hole found. One consideration, non-blocking:

- `PackBits.resolve_metadata` returns a 1-d packed shape, so it now drives the chain chunk-local on rectilinear grids even though the *edge structure* is preserved in a flattened form; fine, but if a future grid type could express "flattened edges" it would restore exact validation behind `PackBits`.

## One gap worth closing before merge

**This PR never references #4354** — body, changelog entry (`changes/4352.bugfix.md`), and docs have no mention. It was opened ~1 hour before the issue and links only the fork PR, so GitHub won't cross-link them and the issue has no visible resolution path. Since this branch is the resolution of that issue, adding "#4354" to the changelog entry (or the body) would close the loop.

The design matches zarrs' grid-mapping approach, the chunk-local trade-off is documented on all three surfaces (docstrings, extending guide, changelog) with both failure directions stated, and CI is green. Happy to re-review if the reference is added or anything above needs rework.

src/zarr/codecs/transpose.py
```diff
@@ -95,6 +95,21 @@ def resolve_metadata(self, chunk_spec: ArraySpec) -> ArraySpec:
             prototype=chunk_spec.prototype,
         )
 
+    def resolve_chunk_grid(
+        self, *, shape: tuple[int, ...], chunk_grid: ChunkGridMetadata
+    ) -> tuple[tuple[int, ...], ChunkGridMetadata]:
+        """Permute the array shape and the per-axis chunk edges by `order`."""
+        from zarr.core.metadata.v3 import RectilinearChunkGridMetadata, RegularChunkGridMetadata
```
ilan-gold, line 102: Why does this need to be a local import?

src/zarr/abc/codec.py
```diff
@@ -148,6 +148,55 @@ def resolve_metadata(self, chunk_spec: ArraySpec) -> ArraySpec:
...
+
+        Return `None` when the chunks after this codec cannot be described by
+        a single chunk grid computed from `chunk_grid` alone. On a regular grid
+        that costs nothing, because all chunks have one shape and
+        `resolve_metadata` describes them exactly. On a rectilinear grid it
+        makes the remaining chain "chunk-local": later codecs are validated
+        against one representative chunk only, so a chain that is invalid for
+        some other chunk shape is accepted at creation and fails when such a
+        chunk is first encoded or decoded (see
+        `zarr.core.metadata.v3.evolve_and_validate_codecs`).
```
ilan-gold, line 171: Why do we do not fail completely in this case? To protect people who have already written `Codec`s that don't implement chunk grid resolution from unnecessary errors?

ilan-gold (commented):
zarr has done breaking changes before, and this seems kind of important. It feels like there should be a safer way to do this than allowing users to create technically invalid arrays only to fail at read/write time.

Does that make sense?

PR #4353, opened 2026-09-14
fix(store): run every LocalStore filesystem call off the event loop

d-v-b:
AI-authored PR that does what it says. Keeps expensive `LocalStore` IO off the event loop with `to_thread`. I'm going to benchmark this to ensure that nothing insane happens.

## Summary

`LocalStore.get` and `set` already run their disk I/O in a worker thread via `asyncio.to_thread`, but nine other async methods did not: `open`/`_open` (and so the lazy open reached from `get`/`set`), `clear`, `delete`, `delete_dir`, `list`, `list_prefix`, `list_dir`, `move` and `getsize` all called `pathlib`/`shutil`/`os` directly from the event loop thread. Walking or modifying a directory tree therefore stalled every other task sharing the loop. Ruff's `ASYNC` rules cannot see these, since they cannot infer that `self.root / key` is a `Path`.

Each method now delegates to one small synchronous helper run under `asyncio.to_thread`. Exception types and ordering are unchanged. Two things fell out of that:

- Once `_open` suspends, concurrent lazy opens (for example from `set_many`) hit the ABC's "store is already open" error. `LocalStore._ensure_open` is now race-tolerant; an explicit second `_open()` still raises as the conformance suite requires.
- `list_prefix` materializes its walk in the thread (as `list` already did) and uses `str.removeprefix` instead of `str.replace`, which also fixes a latent key-mangling case when the root path recurs inside a key.

## Tests

`tests/test_store/test_local.py` gains an autouse detector fixture for the `LocalStore` conformance class: it patches the `os` and `io` entry points every `pathlib`/`shutil` helper bottoms out in and records a violation whenever one is reached on the event-loop thread from inside a `_local.py` coroutine. On unfixed `main` it flags all nine methods above. A companion test asserts the hooks really observe `get`/`set` I/O off-loop, so the detector cannot pass vacuously, and `test_concurrent_lazy_open` pins the race.

Full `tests/` suite on the merged branch: 8787 passed, 1291 skipped, 4 xfailed. mypy, ruff and ruff-format clean.

## Interaction with changes that landed on main since this was opened

Both touch the code this PR moves off the loop, so to spell out why neither is weakened:

- **#4358 / #4359** (retry the `LocalStore` rename when the destination is busy). `_move_with_retry` and `_atomic_write` are reached only from `_put`, which this branch already ran under `asyncio.to_thread`, so the retry's `time.sleep` blocks a worker thread rather than the event loop. This is checked rather than assumed: the detector patches `os.rename`, `os.replace`, `os.link` and `os.unlink`, which is exactly what those helpers bottom out in, so the new retry path is policed like everything else. The retry's own tests are module-level and therefore outside the class-scoped detector fixture, so the two do not interfere in either direction.
- **#4363** (replace runtime asserts with explicit checks). The five redundant `assert isinstance(key, str)` it removed from `_local.py` stay removed; one of them sat on the line where this branch switches `get` to `await self._ensure_open()`, which is the only place the two needed reconciling. The `S101` lint that PR enabled is satisfied, since the asserts this branch adds are all under `tests/`, which the rule excludes.

## Benchmarks

CodSpeed on the pre-merge head: one benchmark improved 10.38% (`test_sharded_morton_single_chunk[(30, 30, 30)-memory]`), 112 untouched, no regressions — so the extra `to_thread` hops cost nothing measurable. The run against the merged head should repost on this push.

Split out of https://github.com/d-v-b/zarr-python/pull/191, whose remaining content is lint-rule enablement.


src/zarr/storage/_local.py
```diff
@@ -131,6 +131,58 @@ def _put(path: Path, value: Buffer, exclusive: bool = False) -> int:
...
+        shutil.rmtree(path)
+    else:
+        path.unlink(missing_ok=True)
+
+
+def _delete_dir(path: Path, prefix: str) -> None:
+    if path.is_dir():
+        shutil.rmtree(path)
+    elif path.is_file():
+        raise ValueError(f"delete_dir was passed a {prefix=!r} that is a file")
```
ilan-gold:
```suggestion
    if not path.is_dir():
        raise ValueError(f"delete_dir was passed a {prefix=!r} that is not a directory")
    shutil.rmtree(path)
```

This seems more exhaustive - there are other things besides directories and files?

src/zarr/storage/_local.py
```diff
@@ -131,6 +131,58 @@ def _put(path: Path, value: Buffer, exclusive: bool = False) -> int:
...
+def _ensure_root(root: Path, *, create: bool) -> None:
+    if create:
+        root.mkdir(parents=True, exist_ok=True)
+    if not root.exists():
+        raise FileNotFoundError(f"{root} does not exist")
+
+
+def _clear(root: Path) -> None:
+    shutil.rmtree(root)
+    root.mkdir()
```
ilan-gold: Is `mkdir` a side effect here in this function? Maybe `_clear` -> `_clear_dir_contents`? Also, what if `root` is not a directory?

tests/test_store/test_local.py
```diff
@@ -1,25 +1,165 @@
...
+        )
+
+    async def test_filesystem_calls_are_observed(
+        self, store: LocalStore, filesystem_calls: _FilesystemCalls
+    ) -> None:
+        """The detector must actually see LocalStore's I/O, or its silence means nothing."""
+        await store.set("foo", self.buffer_cls.from_bytes(b"x"))
+        await store.get("foo")
+        assert ("_put", "io.open") in filesystem_calls.off_loop
+        assert ("_get", "io.open") in filesystem_calls.off_loop
```
ilan-gold, line 145: Parametrize and test all ops? Why only do `_put` and `_get`, both in one test?

src/zarr/storage/_local.py
```diff
@@ -211,18 +263,25 @@ async def open(
...
+        await asyncio.to_thread(_ensure_root, self.root, create=not self.read_only)
         return await super()._open()
 
+    async def _ensure_open(self) -> None:
+        # docstring inherited
+        if not self._is_open:
+            await asyncio.to_thread(_ensure_root, self.root, create=not self.read_only)
+            # Concurrent lazy opens (every `set` of a `set_many`, say) all pass the check
+            # above and each verifies the root, which is idempotent; only the first may
+            # flip the flag, since `Store._open` refuses to open an open store.
```
ilan-gold, line 275: The way this reads to me, the fact that `Store._open` refuses to open an open store (and in fact errors) _causes_ the first flipping the flag, but I _think_ it's the other way around. The first call to finish ensuring the root enters the `if not self._is_open` below and flips the flag, causing all subsequent completions not to enter the guard.

Do I have this right? If not, I might need to re-review

ilan-gold (approved):
Seems reasonable!

src/zarr/storage/_local.py
d-v-b: i'll get the prose re-written

PR #4356, opened 2026-09-14
fix(indexing): avoid full-grid allocations for sparse selections

### LINKED ISSUE #4174: set_coordinate_selection runs out-of-memory for large arrays

d-v-b:
this is an AI-authored PR that makes coordinate indexers much more memory efficient, resolving #4174

Fixes #4174. Coordinate reads and writes and orthogonal integer-array selections now store occupied chunk IDs and run-end offsets instead of allocating count arrays for the entire chunk grid. Selecting a few points from an array with billions of chunks therefore avoids the previous full-grid allocation. This covers both the general coordinate path and the sorted 1D fast path; mask selections routed through `CoordinateIndexer` also benefit.

Run-boundary calculation is linear in the number of selected points. Unsorted selections can still require sorting, and dense selections can use additional temporary memory. The legacy dense attributes on `CoordinateIndexer` and `IntArrayDimIndexer` are materialized on first access and cached without deprecation warnings. Repeated reads return the same array, in-place mutations persist, and iteration honors edits to materialized cumulative offsets. The legacy dataclass fields remain available; `dataclasses.asdict()` materializes them, while `repr` omits the dense arrays to keep inspection sparse. Accessing these attributes still requires full-grid memory.

The new `ceildiv_int` helper uses exact Python integer arithmetic without type-based routing. All eight production call sites for chunk counts and slice calculations use it. The existing `zarr.core.common.ceildiv` helper retains its original floating-point behavior.

The orthogonal boolean-axis optimization is deferred: expanding dense masks into integer positions caused a memory and runtime regression. `BoolArrayDimIndexer` retains its existing implementation and attributes.

Validation: 1,195 tests passed across `test_common.py`, `test_indexing.py`, `test_unified_chunk_grid.py`, and `test_codecs/test_sharding.py` in the Hatch Python 3.12 minimal environment, with 2 skips and 2 expected failures. Coverage includes huge sparse grids, sorted and unsorted selections, duplicates, projection reconstruction, lazy attribute caching, mutation behavior, dataclass introspection, and exact large-integer division.


d-v-b:
@sehoffmann have a look

Issue #3522, opened 2025-10-14
Permission denied when multiple Processes write to the same file


Trobolt:
### Zarr version

3.1.3

### Numcodecs version

0.16.3

### Python Version

3.13.4

### Operating System

Windows 11 24H2

### Installation

using pip into virtual environment

### Description

I am encountering an issue when writing to a Zarr array using Python multiprocessing. (Or more specifically: Missuse Dask's Multiprocessing, as i write into a zarr array inside a map_block function)
The problem occurs when the array chunks contain multiple elements. Writing works correctly 
when each chunk contains only a single element, but fails when the chunk size is larger.

Example:

Here group is a zarr group and zarr_array is another zarr array.

My code that writes to b_arr in a multprocessed manner works, when this creates the array:
b_arr=group.create_array(
    name=name,
    shape=zarr_array.cdata_shape,
    chunks=(1, 1),
    dtype=np.uint64,
    fill_value=0,
)

and here not:
b_arr=group.create_array(
    name=name,
    shape=zarr_array.cdata_shape,
    chunks=zarr_array.cdata_shape,
    dtype=np.uint64,
    fill_value=0,
)

I believe that zarr should work when mutliple processes concurrenty write into an array, so i consider it a bug.

### Steps to reproduce

```python
# /// script
# requires-python = ">=3.11"
# dependencies = [
#   "zarr@git+https://github.com/zarr-developers/zarr-python.git@main",
# ]
# ///
#
# This script automatically imports the development branch of zarr to check for issues

import zarr
# your reproducer code
# zarr.print_debug_info()
```

FlynnOConnell:
I am getting this error as well

```
File [~\AppData\Roaming\uv\python\cpython-3.12.9-windows-x86_64-none\Lib\pathlib.py:1376](http://localhost:8888/lab/workspaces/auto-P/tree/dev/~/AppData/Roaming/uv/python/cpython-3.12.9-windows-x86_64-none/Lib/pathlib.py#line=1375), in Path.replace(self, target)
   1366 def replace(self, target):
   1367     """
   1368     Rename this path to the target path, overwriting if that path exists.
   1369 
   (...)   1374     Returns the new Path instance pointing to the target path.
   1375     """
-> 1376     os.replace(self, target)
   1377     return self.with_segments(target)

[...]

PR #4358, opened 2026-09-15
fix(store): retry the LocalStore rename when the destination is busy

Tomatokeftes:
## Summary

Zarr v2 retried `DirectoryStore`'s rename because Windows intermittently refuses
to replace a destination (#597, fixed by #698). Atomic writes arrived in v3's
`LocalStore` in #3412 without that retry, so the failure is back — `_atomic_write`'s
`tmp_path.replace(path)` raises

```
PermissionError: [WinError 5] Access is denied:
  '...zarr.<hex>.partial' -> '...zarr.json'
```

and aborts the write. That's #3522, and @d-v-b already suggested on #597 that #698's
retry should be ported over. This does that.

`_move_with_retry` wraps the final move and retries only the two Windows codes that
mean the destination couldn't be superseded right now. It needs no platform check:
off Windows an `OSError` carries no `winerror`, so the first attempt either succeeds
or raises, and the tests still run everywhere.

The `exclusive` path goes through it too but is unaffected by construction — the
`FileExistsError` it relies on to report an existing node is `ERROR_ALREADY_EXISTS`
(183), which isn't in the retried set, so it still propagates on the first attempt.

On Windows 11, 4,000 group-attr rewrites (each a replace onto an existing
`zarr.json`): 155–171 raised before, 0 after, for 3.68 s → 4.08 s of wall clock on a
workload that is nothing but replace-onto-existing.

Two things that may be worth knowing, since #3522 is framed as a multiprocessing
problem and #698 is titled as an antivirus one:

- it isn't concurrency — it reproduces single-threaded, in one process, with zarr not
  imported, at roughly 4% of replaces onto an existing destination;
- it doesn't look like antivirus either — replacing a *fresh* name never failed here
  (0 in 4,000) while reusing the same name fails ~5%, same write pattern and same
  scanner. That points at NTFS file tunnelling, and there's a
  [Microsoft thread](https://learn.microsoft.com/en-us/answers/questions/5559596/movefileex-can-produce-a-access-is-denied-error-on)
  where it reproduces with Defender disabled and is gone on Server 2025.

Either way the retry is the same fix.

## For reviewers

The bit I'd most value a second look at is routing the `exclusive` branch through
`_move_with_retry` at all. I believe 183 excludes it safely and
`test_move_with_retry_does_not_retry_other_errors` pins that, but if you'd rather it
bypassed the retry entirely that's a one-line change.

I'd also welcome a view on `_RETRY_DELAYS`. It sums to under a second so a genuine
failure still surfaces promptly, and in a 20,000-replace loop 475 of 498 recoveries
needed only the second attempt with the worst needing the fourth — but the tail is
judgement, not measurement.

Behaviour is unchanged off Windows.

## Author attestation

## TODO

* [ ] New/modified features documented in `docs/user-guide/*.md` — n/a, internal behaviour


d-v-b:
this looks good, I'm going to merge as-is then apply some polishing changes in a follow-up PR

d-v-b (approved):

PR #4363, opened 2026-09-16
fix: replace runtime assert statements with explicit checks

### LINKED ISSUE #4362: Bad assert usage
In core/group.py

Line 414:
``    def from_dict(cls, data: dict[str, Any]) -> GroupMetadata:
        data = dict(data)

        assert data.pop("node_type", None) in ("group", None)``

If code is run with -O it does not work correctly

d-v-b:
AI-authored PR that removes runtime assert statements. closes #4362 

## Summary

Removes every `assert` statement from runtime code under `src/zarr/` and enables ruff's `S101` rule so new ones are flagged. Asserts are stripped under `python -O`, so any check that matters for correctness or for narrowing a type must be an explicit `if`/`raise`.

Two of the 47 asserts were load-bearing:

- `GroupMetadata.from_dict` asserted that `node_type` was `"group"`, and `zarr.api.asynchronous.open` caught `AssertionError` to fall back from array to group. Under `-O` that fallback silently disappeared. It now raises `NodeTypeValidationError`, which that call site already catches, and the `AssertionError` catch is gone.
- `make_store` asserted on `mode`, but the real validation lives downstream in `StorePath.open`. A bad user-supplied mode now raises the same `ValueError` there.

The rest fall into three buckets:

- **Deleted (25)**: redundant with type annotations (`isinstance(key, str)` in the stores, `isinstance(chunk_array, NDBuffer)` in codecs) or with an existing check (`Buffer.combine` dtype checks the constructor already enforces).
- **Restructured so mypy narrows on its own (11)**: locals in `_get_loop` and `_validate_scalar_map`, a hoisted `sync_transform` in the codec-pipeline write path, and `get_array_metadata` returning from each branch via two small helpers.
- **Converted to explicit raises (11)**: user-reachable ones raise `ValueError`; unsupported-feature ones in the shard byte getter/setter raise `ValueError`/`NotImplementedError`; impossible-state ones (a `None` shard index, a missing sync transform) raise `RuntimeError`.

`src/zarr/testing/` is excluded from `S101` along with `tests/`: the store conformance suite and hypothesis strategies assert on purpose.

## For reviewers

Behavior is meant to be unchanged for anyone not running under `-O`. Worth a second look:

- The exception types chosen for the converted asserts, especially `RuntimeError` for the impossible-state cases in `sharding.py` and `codec_pipeline.py`.
- The `get_array_metadata` restructure in `array.py`. It is the largest diff and the only one that moves logic around rather than swapping a line.
- The import-time version check in `zarr/__init__.py` now raises `RuntimeError` unconditionally instead of only outside `-O`.

## Author attestation

## TODO

* [x] Add unit tests and/or doctests in docstrings (one test per new user-reachable error)
* [x] Add docstrings and API docs for any new/modified user-facing classes and functions (none)


d-v-b:
this gets merged when it's green

PR #4365, opened 2026-09-16
fix(zarr-metadata): v2 array document is open and filters may be empty

d-v-b:
This AI-authored PR brings `zarr-metadata` closer to the spec by allowing `filters: []` and extra keys in `.zarray` (I didn't know the spec allowed this -- extra keys are forbidden from `.zgroup`)

Two structural rules in `zarr_metadata`'s v2 array validator are stricter than the v2 spec. The shared conformance corpus was updated first (d-v-b/zarr-metadata.js#2, `conformance/v2_array.json` cases 5 and 8), which turns this package's conformance run red (65/67) until it follows — this PR brings it back to 67/67.

| Rule | Spec (`docs/v2/v2.0.rst`) | Was | Now |
|---|---|---|---|
| Extra members in `.zarray` | line 91: "Other keys **SHOULD NOT** be present within the metadata object and **SHOULD be ignored** by implementations" — vs `.zgroup` line 313: "Other keys **MUST NOT** be present" | `invalid_value` at the key | tolerated by `validate_array_metadata_v2`, dropped by the model, allowed by the Pydantic schema |
| `filters: []` | line 77: "A list of JSON objects providing codec configurations, or ``null``" — no minimum | `invalid_value` "expected at least one filter" | accepted (`min_length=1` dropped from the schema too) |

## Changes

- `model/_validation.py`: the v2 array validator no longer calls `_unexpected_keys` (the group validator still does) and no longer rejects an empty filter list.
- `_pydantic_schema.py`: `ZarrV2ArrayMetadataJSON` becomes `extra_items=JSONValue` like the v3 TypedDict; `ZarrV2FilterPipelineJSON` loses `Field(min_length=1)`. `ZarrV2GroupMetadataJSON` stays `closed=True`.
- Unchanged: the on-disk `.zarray` rule in `from_key_value` that `attributes` must not appear there (it lives in `.zattrs`) — that's a different rule, and the test for it is now separate from the "other extras are ignored" test.
- Tests inverted accordingly: `test_v2_filters_may_be_empty`, `test_array_v2_ignores_unknown_document_member`, `test_v2_from_key_value_ignores_zarray_extra_members`, `test_v2_array_schema_allows_empty_filters`, `test_v2_array_schema_allows_unknown_document_members`.

The TypeScript port keeps the SHOULD NOT visible: its (TS-only) v2 semantic layer reports extra `.zarray` members as advisories so consumers can surface them as warnings.

## Test plan

- [x] `just test` in `packages/zarr-metadata`: 595 passed
- [x] `just lint`
- [x] zarr-metadata.js `just conformance <this checkout>`: 67/67 (was 65/67 against the released package)


d-v-b:

Added a docs commit that links every spec statement in `packages/zarr-metadata` (source and tests) to the exact text — commit-pinned permalinks with line ranges into zarr-specs `fc7dd9c` and zarr-extensions `4da7b37`. Page-level `tree/main` links to zarr-extensions are pinned the same way.

While doing that, three statements turned out to be wrong or stale, corrected in the same commit — reviewers may want to look at these specifically:

1. **`v3/codec/zstd.py`** typed `checksum` as required, citing zarr-specs PR #256 ("unmerged at time of writing" — it never merged). The published zarr-extensions entry says `checksum` is optional ("Should be omitted if false") and its `schema.json` requires only `level`. `ZstdCodecConfiguration.checksum` is now `NotRequired[bool]`. Typing-only; no runtime validator inspects zstd configuration.
2. **`v3/consolidated.py`** said the v3 core spec "does not define consolidated metadata" / "not a spec artifact". Since zarr-specs #373 the core spec's *Additional fields* section names `consolidated_metadata` and fixes its envelope (`core/index.rst` L802-L816); the entry format is still a reference-implementation convention. Reworded accordingly.
3. **`v2/array.py`** `ZarrV2ArrayMetadataJSONPartial` docstring referred to the array document's "closed shape", which this PR makes untrue; it now says other `.zarray` keys are ignored per the spec (the `.zgroup` counterpart keeps "closed", with its MUST NOT linked).

`just test` 595 passed; `just lint` and `just typecheck` clean.

PR #4369, opened 2026-09-17
fix(store): don't pin from_url filesystems in fsspec's instance cache

### LINKED ISSUE #4221: FsspecStore.from_url() raises "RuntimeError: Loop is not running" during interpreter shutdown
### Zarr version

v3.3.0

### Numcodecs version

v0.16.5

### Python Version

3.12.11

### Operating System

Ubuntu 26.04

### Installation

uv add zarr

### Environment

Python 3.12.11
zarr 3.3.0
fsspec 2026.7.0
adlfs 2026.5.0
azure-identity 1.25.3

### Description

Creating a FsspecStore backed by Azure Blob Storage succeeds, but during
interpreter shutdown the following exception is emitted:

`RuntimeError: Loop is not running`

The exception is raised from weakref._exitfunc during cleanup.

### Steps to reproduce

```python
# /// script
# requires-python = ">=3.12"
# dependencies = [
# ]
# ///
#

import zarr
from zarr.storage import FsspecStore
from azure.identity import DefaultAzureCredential

FsspecStore.from_url(
    "az://container/example.zarr",
    storage_options={
        "account_name": "<account>",
        "credential": DefaultAzureCredential(),
    },
)

print("OK")
# zarr.print_debug_info()
```

### Additional output

```
  File ".../weakref.py", line ...
  ...
  File ".../fsspec/asyn.py", line 76, in sync
    raise RuntimeError("Loop is not running")
RuntimeError: Loop is not running
```

YuEfSaEDU:
`FsspecStore.from_url()` retained the filesystem it created in fsspec's global instance cache, so backends that register cleanup finalizers on their filesystems kept those instances alive until interpreter shutdown. At exit, finalizers that call fsspec's `sync` machinery (e.g. adlfs) find the event loop gone and emit `RuntimeError: Loop is not running` from `weakref._exitfunc` on every process exit.

`from_url()` now creates its filesystem with `skip_instance_cache=True`, mirroring the `cachable=False` behavior of fsspec's own `AsyncFileSystemWrapper`, so the filesystem's lifetime follows the store's and backend cleanup runs while the process is still executing. Callers who rely on cross-call instance sharing can pass `skip_instance_cache` in `storage_options` to restore it. A subprocess regression test pins the reported symptom against a local async-capable backend, and a towncrier fragment documents the behavior change.

Fixes #4221


src/zarr/storage/_fsspec.py
```diff
@@ -248,7 +252,13 @@ def from_url(
...
 
         opts = storage_options or {}
-        opts = {"asynchronous": True, **opts}
+        # The filesystem created here is owned by this store: instruct fsspec not
+        # to keep it in its global instance cache, so that it is released when
+        # the store is released instead of lingering until interpreter shutdown
+        # with pending cleanup finalizers that touch dead event loops
+        # (https://github.com/zarr-developers/zarr-python/issues/4221). An
+        # explicit ``skip_instance_cache`` in ``storage_options`` wins.
+        opts = {"asynchronous": True, "skip_instance_cache": True, **opts}
```
d-v-b, line 261: this amounts to changing the default `skip_instance_cache` behavior for `url_to_fs`. I think it's better to leave that default to fsspec. Folks who want to skip the instance cache can say so explicitly, using the kwarg, until fsspec changes its default.

PR #4378, opened 2026-09-19
fix(zarr-metadata)!: make entity types assignable to metadata fields; validators return tuples

d-v-b:

Part 1 of a split of d-v-b/zarr-python#296 (reviewed on the fork as d-v-b/zarr-python#317). Small and independent; a follow-up PR adding a composition rules layer builds on it.

- `ZarrV3NamedConfigJSON.name` / `.configuration` are `ReadOnly` (PEP 705) and the envelope is `closed` (PEP 728), so every concrete codec / chunk-grid / chunk-key-encoding / data-type TypedDict is now assignable to the fields it describes (`codecs`, `chunk_grid`, `data_type`, ...). Previously a type checker rejected putting `BloscCodecMetadata` into `codecs`.
- Every concrete `*Object` / `*Configuration` TypedDict is `closed`, and object forms declare `must_understand: NotRequired[bool]`.
- **Breaking:** every `validate_*` in `zarr_metadata.model` returns `tuple[ValidationProblem, ...]` instead of a list, and `MetadataValidationError.problems` is a tuple.
- `load_store_json` returns `object` rather than `Any`; `ANN401` is enforced package-wide.
- `ZarrV2ConsolidatedMetadataJSON.zarr_consolidated_format` is `Literal[1]`.
- `zarr_metadata.pydantic` serializers declare their return schema via the pydantic-facing shadow types.

The structural validator is otherwise unchanged; the `dimension_names` / `chunks` rank checks stay where they are until part 2.

Rebased onto main. The only conflict was with #4365 (v2 array document is open): its semantics are kept, and its two new tests now expect `()` from `validate_array_metadata_v2`.

Verified after rebase: ruff check/format, pyright 1.1.404 (0 errors), 595 package tests.


d-v-b:
this is a zarr-metadata PR that makes a variety of improvements as described in the AI-written PR description. I'm self-merging it.

PR #4379, opened 2026-09-19
feat(zarr-metadata): composition rules layer, shape-exact entity validators

d-v-b:

Part 2 of a split of d-v-b/zarr-python#296 (reviewed on the fork as d-v-b/zarr-python#318). #4378 has merged, so this diff is self-contained against `main`.

## What this adds

`zarr_metadata.rules`: semantic validation for whole documents. `model` checks structure element by element; `rules` judges what structure alone cannot — whether the values are in range, and whether the document's parts agree with one another.

The second layer is **not a rule engine**. Every extension point is a frozen dataclass that answers for itself:

| method | question |
|---|---|
| `coerce` | is this metadata an entity of mine, and if not, why? (type-space) |
| `problems` | which of my own values does the spec disallow? (value-space) |
| `to_json` | what is my simplest equivalent spelling? |
| `fill_value_problems` / `incoming_problems` / `shape_problems` / `transition` / `grid` | the questions that need one part of the document |

Validating a document is composing those answers. Nothing in the document layer knows what `blosc` or `int32` or `rectilinear` is, so **adding an extension is a class and one registry entry**.

Scope is a `Context`: which identifiers are in play at each extension point, as either `CORE` (what the specification defines) or `CORE_AND_EXTENSIONS` (plus what `zarr-extensions` registers and this package models). A name in neither is not rejected — that is extension openness — it is simply not judged. The scope is passed to every `coerce`, and every entity that does not contain another ignores it.

`zarr_metadata.v3.entity` is the public door: the four base classes, `Context` and the scopes, `ArrayParts`/`ChunkGrid`, and the member checks a `member_types` table is built from. `validate_*`, `parse_*` and `canonicalize_array_metadata_v3` take `context=`, so a third party can have its own entity judged without touching this package. `tests/v3/test_extension_api.py` is the proof: a third-party codec and data type, defined, registered and judged, importing nothing private.

## What a codec is judged against

A codec is judged against the array it *receives*, not the document's top-level fields: a shard behind a transpose must divide the transposed chunk, and a `bytes` codec behind a `cast_value` needs an endianness for the target type.

What flows through the chain is `ArrayParts` — the chunk grid itself plus the coerced data type. The plural is load-bearing: one pipeline encodes every chunk, so a chain rule quantifies over all of them. `ChunkGrid` carries its rank and, per dimension, the *set* of lengths that dimension's chunks take: singletons for a regular grid, `{30, 34}` for a rectilinear axis, `None` only where nothing can be read.

That expresses judgments a single shape cannot. A rectilinear grid, a transpose moving its varying axis, and a shard that must divide what arrives: inner extents of 15, 30 and 34 against chunks of 30 and 34 are all rejected, and only a common divisor passes. A grid this package cannot read still pins the rank, so a rank-mismatched transpose behind a third-party grid is caught. Sharding is not a special case — a shard is a nested array, so its inner pipeline is walked by the same function as the document's own, and the shard index is judged against its own derived shape (chunks-per-shard plus a trailing dimension of 2). Neither waits on what reached the codec, so an unmodelled codec upstream — every `numcodecs.*` filter is one — does not silence the shard interior.

Following zarrs: a grid is built from metadata *and* the array shape it divides, because neither determines a grid alone; its dimensionality is total rather than optional; its edge lengths are reported per dimension.

## Boundary changes

- v3 `dimension_names` vs `shape` and v2 `chunks` vs `shape` move from the structural validator into rules. `model` parsers and dataclasses now accept those documents.
- **Pydantic field types run the semantic layer** before normalizing, so they are strictly stronger than before rather than losing the two moved checks.
- **Unknown configuration members are an error.** Whether a `configuration` is closed is unspecified (zarr-developers/zarr-specs#270), and this package takes the strict reading on purpose: such a member is almost always a typo or a setting meant for a different entity. `rules.parse_*` and the whole-document pydantic types reject it; `model.parse_*` and the bare `ZarrV3MetadataField` type — neither of which interprets configurations — accept it. The `unknown_key` kind exists for triage, and the location names the offending key.
- Optional configuration members default to `UNSET`, not `None`, so `{"offset": null}` and `{}` stay different documents.
- A bare `bytes` inside sharding `index_codecs` is rejected: the shard index is `uint64`, so the endianness is required there too.
- `to_key_value()` returns `Mapping[str, bytes]`; the previous literal-key mapping made the advertised round trip fail to type-check.

## Adversarial review

Five reviewers on zarr-python-shaped workloads: reading and writing `zarr.json`, authoring a third-party extension, hunting wrong verdicts, type-checking a strict consumer, and v2/group/consolidated. Eleven defects, all reproduced and fixed here. The one that mattered: a metadata field **nested** inside a configuration — a shard's pipelines, a struct field's data type, a `cast_value` target — was not getting the structural judgment a top-level one gets, so an extra member or a non-object `configuration` was accepted there and then deleted by the canonicalizer.

Two results worth stating:

- Eleven `zarr.json` documents written by this repository's own `create_array` — float32+blosc, sharded, string, datetime64, complex128, structured, transposed — validate clean. No false rejections.
- Against the rule registry this replaces, over one shared corpus of 800 documents: **no verdict differs in the laxer direction, and the 400 valid documents report identically**.

Known limitation, deliberate: an entity's `to_json` returns its own object TypedDict, which mypy will not accept where the envelope type is wanted — it reads every TypedDict as `Mapping[str, object]`, never as `Mapping[str, JSONValue]`. The conversion is sound, because every TypedDict here is `closed` (PEP 728) and pyright implements PEP 728 while mypy does not yet (python/mypy#8994, python/mypy#18439). The annotation stays accurate and the write path needs a `cast` under mypy.

## Testing

Codec chains are drawn from the codec `TypedDict`s with `st.from_type`. A second strategy corrupts one configuration member to arbitrary JSON — the only generated input that reaches an entity holding a member it cannot read. A third builds documents valid by construction, with inner chunk shapes drawn from the divisors of the extents they must divide, over regular and rectilinear grids; it is the only generated coverage of the accept side. Each strategy's reach is asserted rather than assumed.

Tests are type-checked along with the sources (`src` strict, `tests` at a relaxed rule set). They were previously checked by nothing.

## Relative to d-v-b/zarr-python#296

Dropped: the incremental builder, the one-shot `create_*` factories, the `check_*` / `Valid` / `Invalid` family, `rules.is_*`, the extension-point provenance table, `TypeIs` codec guards, and fill-value propagation through codec chains.

Also gone, and the point of the current shape: the rule registry itself — the dispatcher, the `reads`/`reads_optional` gate, nine modules of rules keyed by entity name, the fill-value dispatch table, three frozensets of data-type names, four tables of per-member type checks, and the drift tests that existed to keep those tables in step. The tables cannot drift now because there is one of each fact instead of four.


packages/zarr-metadata/src/zarr_metadata/v3/_entity.py
```diff
@@ -0,0 +1,1291 @@
...
+            value,
+            **{
+                name: canonicalize_nested(field_annotation, getattr(value, name))
+                for name, field_annotation in field_hints(inner).items()
+            },
+        )
+    return value
+
+
+_MISSING_DEFAULT: Final = object()
```
d-v-b: use a sentinel

packages/zarr-metadata/src/zarr_metadata/v3/_entity.py
```diff
@@ -0,0 +1,1291 @@
...
+"""Distinguishes "declared no default" from a default that is None or UNSET."""
+
+
+@dataclass(frozen=True, slots=True)
+class Opaque:
+    """A metadata field this reading did not turn into an entity.
+
+    Carrying the JSON rather than dropping it is what makes the result a
+    real union: `CodecEntity | Opaque` is exhaustive and narrows, where
+    `CodecEntity | object` is just `object` and narrows to nothing.
```
d-v-b: do not document relative to the code this replaced. nobody cares about that. use contrastive statements when theres something useful to contrast to

packages/zarr-metadata/src/zarr_metadata/v3/_entity.py
```diff
@@ -0,0 +1,1291 @@
...
+# body* pointing at the class it replaced. Several entities call `super()`
+# to narrow `to_json` and to adjust `configuration`, so they would each
+# have to spell it `super(Cls, self)`. CPython fixed this in 3.13, so when
+# that is the floor this is worth revisiting; the memory saved is small at
+# document scale, which is why it has not been.
+_DERIVED: Final = ("member_types", "configuration_required", "nested_members", "name_members")
+"""The class variables `_compile_entity` derives; a declaration of one is refused."""
+
+
+def _compile_entity(cls: type[MetadataEntity]) -> None:
```
d-v-b: red flag. never modify a class in place. find another way to solve this that keeps classes immutable after definition.

packages/zarr-metadata/src/zarr_metadata/v3/_entity.py
```diff
@@ -0,0 +1,1291 @@
...
+        inner = get_args(hint)[0] if get_origin(hint) is ClassVar else hint
+        if get_origin(inner) is Literal and getattr(cls, name) not in get_args(inner):
+            return (
+                f"{cls.__name__} sets {name} = {getattr(cls, name)!r}, "
+                f"which is not one of {get_args(inner)!r}"
+            )
+    return None
+
+
+def _array_array_codecs_define_transition(cls: type[MetadataEntity]) -> str | None:
```
d-v-b: what on earth is this. just use an ABC and abstract classes?

packages/zarr-metadata/src/zarr_metadata/v3/_entity.py
```diff
@@ -0,0 +1,1116 @@
...
+    Answers `to_json` and `canonical` as an entity does, so a field typed
+    `CodecEntity | Opaque` is written and simplified without asking
+    which case it holds.
+
+    Built by the reader through `create_unchecked`, from JSON it has
+    refined; the constructor checks one built by hand.
+    """
+
+    json: JSONValue
+    reason: Literal["out_of_scope", "invalid"]
```
d-v-b, line 214: these reasons cant be bare strings. consider a shared literal type, and then unions of useful combinations

d-v-b:
the slop was too strong here, closing this and opening something better soon

PR #4380, opened 2026-09-19
fix(zarr-metadata): reject cast_value wrap on non-integer targets

d-v-b:

**Depends on #4379**, and stacked on it: this targets `main`, so its diff includes #4379's commits. Review only the last commit, or the diff against #4379's branch: https://github.com/d-v-b/zarr-python/compare/zarr-metadata/rules-layer...zarr-metadata/rules-wrap-blosc

## What this adds

`cast_value`'s `out_of_range: "wrap"` is rejected for targets it is not defined on.

The spec permits wrapping only for integral targets with a two's complement representation. A modelled non-integer target reports `invalid_value` at `out_of_range` — `bool` included, since it is not two's complement. A target this reader does not model declines, because it may be an integral extension type:

```
wrap -> int32             accepted
wrap -> uint64            accepted
wrap -> bool              out_of_range 'wrap' requires a two's complement integer data_type
wrap -> float32           out_of_range 'wrap' requires a two's complement integer data_type
wrap -> r24               out_of_range 'wrap' requires a two's complement integer data_type, got 'r24'
wrap -> numpy.datetime64  out_of_range 'wrap' requires a two's complement integer data_type
wrap -> mycorp.bigint     accepted
```

## Where the fact lives

Whether a type wraps is something the **data type** says about itself, `twos_complement`, and `DataTypeEntity` requires it. A data type that does not decide fails at import:

```
TypeError: Undecided does not declare twos_complement
```

Required rather than defaulted on purpose: either default answers for a new data type silently, and one of them accepts a cast the spec does not define.

That replaces what this rule carried before #4379 was rebuilt — two frozensets of data type names, plus a registry test asserting they partition every modelled type so that none could be accepted by omission. There is no longer a table to partition, so there is nothing for a test to keep in step.

## Also here

Entities gain `name`, the spelling a document writes, as distinct from `identifier`, the key they are tabled under. The two differ only for the raw-bytes family, whose identifier is invented and belongs in no message a reader sees — this rule was reporting `got 'r<N>'` for a document that said `r24`.


d-v-b:
this will be superseded by better stuff, so i'm closing it

PR #4382, opened 2026-09-20
perf: cut the fill_value check on the chunk write path

PursuitOfDataScience:
## Summary

Before zarr writes a chunk it asks one question: is this chunk entirely `fill_value`?
If it is, the chunk is skipped and no file is written.

That check is `NDBuffer.all_equal`. It runs on **every chunk write**, because
`write_empty_chunks=False` is the default. On an uncompressed float32 write it was
**62.6% of the total time**.

This PR makes the check cheap. Behaviour is unchanged.

<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://raw.githubusercontent.com/PursuitOfDataScience/zarr-python/bench-assets/bench/speedup-dark.png">
  Chunk write speedup by calling path: zarr whole-array write 2.16x, zarr per-chunk writes 1.74x, xarray.to_zarr 1.63x, xarray.to_zarr with dask 1.45x
</picture>

## Two things were slow

They are independent, and each one helps a different set of users.

### 1. The comparison used a dtype NumPy has no fast path for

When `fill_value` is zero the check compares raw bytes, not values, so that `-0.0` is
not mistaken for `0.0` (#3144). It did that by viewing both sides as `V{itemsize}`, a
void dtype.

NumPy has vectorised comparison loops for integers. It has none for void, so the void
view quietly falls back to a generic element-by-element path.

An unsigned integer of the same width compares exactly the same bits and does have a
fast path. On a 16 MiB all-zero buffer that is 27x faster for float32 and 33x for
float64 on numpy 2.5.3, and 23x and 17x on numpy 2.2.6. Widths with no same-size
integer, longdouble and complex128, keep the void path.

This one helps zarr's own default, where `fill_value` is `0.0`.

### 2. The scan never stopped early

If the first element already differs from `fill_value`, the answer is known. The old
code read the whole chunk anyway, because `np.array_equal` builds a full boolean array
before reducing it. On the write path "not uniform" is the common case, so that cost
was paid nearly every time.

It now walks slices of the leading axis, doubling in length, and returns on the first
slice that differs. A chunk that really is uniform costs the same as before.

**Why the leading axis rather than a flattened view:** a chunk is normally a strided
view into a larger array, so flattening it copies. An earlier version of this patch
required a contiguous buffer, which meant the early exit never actually ran and
measured exactly 1.00x end to end. Slicing axis 0 is a view whatever the layout.
`test_all_equal_strided_chunk_view` pins this.

This one helps xarray, which writes `fill_value=NaN` and so never reached the first
problem at all.

## Results

Paired and interleaved with alternating arm order, bootstrap 90% CI on the median
ratio, Wilcoxon signed-rank on the paired times. AMD EPYC 7702, numpy 2.5.3, 256 MiB
float32, uncompressed. Every point below is significant at p < 1e-11.

<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://raw.githubusercontent.com/PursuitOfDataScience/zarr-python/bench-assets/bench/grid-dark.png">
  Speedup across chunk size and data shape on two filesystems: from 1.56x for 1 MiB random chunks up to 13.1x for 16 MiB all-fill_value chunks
</picture>

The worst single pair anywhere in the run was 1.05x, so no configuration regressed. The
win also survives the default zstd compressor at 2.0x to 3.2x, so it is not an artifact
of testing uncompressed.

## Correctness

3,622 differential comparisons against the old implementation, **0 mismatches**. They
cover 13 dtypes, C, F and strided layouts, `-0.0`, NaN, inf, mismatches at the first,
middle and last element, block boundaries, and the longdouble and complex128 fallbacks.

New tests cover agreement with a whole-buffer scan, `-0.0` staying distinct (#3144),
NaN, non-contiguous buffers, strided chunk views, and `shape[0] == 1`, where there is
no leading axis to slice and the code must fall back.

## For reviewers

This is purely a performance change; behaviour is meant to be identical. Two places I
would most value a second opinion:

1. **Is the unsigned integer view bit-identical to the void view for every dtype it
   accepts?** This is the part that could change results rather than only timings. It
   is gated on itemsize (1, 2, 4, 8) with everything else left on the old path, and the
   differential test covers `-0.0`, NaN and inf, but a second read is worth it.
2. **Is slicing the leading axis the right shape for the early exit?** Flattening would
   read more simply, but copies on the strided views this method actually receives.

I am already confident in the `shape[0] == 1` and empty-array fallbacks, both covered by
tests.

## Author attestation

## TODO


src/zarr/core/buffer/core.py
```diff
@@ -531,32 +540,66 @@ def __len__(self) -> int:
     def __repr__(self) -> str:
         return f"<NDBuffer shape={self.shape} dtype={self.dtype} {self._data!r}>"
 
-    def all_equal(self, other: Any, equal_nan: bool = True) -> bool:
-        """Compare to `other` using np.array_equal."""
-        if other is None:
-            # Handle None fill_value for Zarr V2
-            return False
+    @staticmethod
+    def _compare_all(data: NDArrayLike, other: Any, equal_nan: bool) -> bool:
```
d-v-b, line 544: Instead of a static method we should have a free function that gets used in the body of a plain method

d-v-b:
this looks good, feel free to ignore my request about the static method thing because I can fix that in a follow-up

d-v-b (approved):

